PR #1, opened 2017-02-06
Add KMS alias to cloudformation

jacobwinch:
@paulbrown1982 

This fixes the minor alias issue that you mentioned this morning:
http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-kms-alias.html


paulbrown1982:
👍 Nice, I'll copy that!

PR #2, opened 2017-02-08
Add auth check to validate that request originates from Zuora

jacobwinch:
@paulbrown1982

### DIFF
--- a/src/main/scala/com/gu/autoCancel/Auth.scala
+++ b/src/main/scala/com/gu/autoCancel/Auth.scala
@@ -1,0 +1,27 @@
+package com.gu.autoCancel
+
+import play.api.libs.json.JsValue
+
+object Auth extends Logging {
+
+  def credentialsAreValid(inputEvent: JsValue, username: String, password: String): Boolean = {
+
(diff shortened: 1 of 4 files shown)


src/main/scala/com/gu/autoCancel/Auth.scala
```diff
@@ -0,0 +1,27 @@
...
+
+object Auth extends Logging {
+
+  def credentialsAreValid(inputEvent: JsValue, username: String, password: String): Boolean = {
+
+    /* Using query strings because for Basic Auth to work Zuora requires us to return a WWW-Authenticate
+    header, and API Gateway does not support this header (returns x-amzn-Remapped-WWW-Authenticate instead)
+    */
+    val maybeUsernameValue = (inputEvent \ "queryStringParameters" \ "username").asOpt[String]
+    val maybePasswordValue = (inputEvent \ "queryStringParameters" \ "password").asOpt[String]
```
paulbrown1982: Is this sent through encrypted if it's in the query string?

paulbrown1982:
👍

PR #5, opened 2017-02-13
Remove invoice un-posting logic

jacobwinch:
This PR removes all code related to 'un-posting' invoices, which is now superfluous (following discussions with Stuart and Mark).

I've also taken the opportunity to tidy up some of the logic (other changes are test edits due to this re-factor).

@paulbrown1982 @johnduffell

### DIFF
--- a/src/main/scala/com/gu/autoCancel/Lambda.scala
+++ b/src/main/scala/com/gu/autoCancel/Lambda.scala
@@ -96,20 +96,18 @@ object Lambda extends App with Logging {
 
   def getOverdueUnpaidInvoice(accountSummary: AccountSummary, dateToday: LocalDate): String \/ Invoice = {
     val unpaidAndOverdueInvoices = accountSummary.invoices.filter { invoice => invoiceOverdue(invoice, dateToday) }
-    if (unpaidAndOverdueInvoices.size > 1) {
-      logger.error(s"Found more unpaid invoices than expected for account: ${accountSummary.basicInfo.id}. The invoices we got were: ${unpaidAndOverdueInvoices.map(_.id)}")
-      "Multiple unpaid invoices".left
-    } else {
-      val maybeOverdueInvoice = unpaidAndOverdueInvoices.headOption
-      maybeOverdueInvoice match {
-        case Some(overdueInvoice) => {
-          logger.info(s"Determined that InvoiceId: ${overdueInvoice.id} is the overdue unpaid invoice for Account Id: ${accountSummary.basicInfo.id}")
-          overdueInvoice.right
-        }
-        case None => {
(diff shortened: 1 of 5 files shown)


jacobwinch:
FYI @ajosephides

src/main/scala/com/gu/autoCancel/Lambda.scala
```diff
@@ -115,52 +123,28 @@ object Lambda extends App with Logging {
...
-    activeSubs.flatMap { subs =>
-      {
-        if (subs.size > 1) {
-          // This should be a pretty rare scenario, because the Billing Account to Sub relationship is (supposed to be) 1-to-1
-          logger.error(s"More than one subscription is Active on account: ${accountSummary.basicInfo.id}. Subscription ids are: ${subs.map(_.id)}")
-          "More than one active sub found!".left // Don't continue because we don't know which active sub to cancel
-        } else {
-          val subToCancel = subs.head
+    val activeSubs = accountSummary.subscriptions.filter(_.status == "Active")
+    if (activeSubs.size > 1) {
```
johnduffell: for these situations you could always use pattern matching -
```scala
activeSubs match {
  case sub :: Nil => // there's one
  case Nil => // there's none
  case subs => // there are more than one
}
```

src/main/scala/com/gu/autoCancel/Lambda.scala
```diff
@@ -115,52 +123,28 @@ object Lambda extends App with Logging {
...
   }
 
-  def handleZuoraResults(
-    updateSubscriptionResult: UpdateSubscriptionResult,
-    cancelSubscriptionResult: CancelSubscriptionResult,
-    updateInvoiceResult: UpdateInvoiceResult
-  ): String \/ Unit = {
-    if (!cancelSubscriptionResult.success || !updateInvoiceResult.success || !updateSubscriptionResult.success) {
+  def handleZuoraResults(updateSubscriptionResult: UpdateSubscriptionResult, cancelSubscriptionResult: CancelSubscriptionResult): String \/ Unit = {
+    if (!cancelSubscriptionResult.success || !updateSubscriptionResult.success) {
```
johnduffell: I guess you want it to try to do both, and only return left after that.  this does feel a bit funny doing the conversion to a left later (although I have a suspicion that is what the `|@|` that we love so much is for!)

johnduffell:
looks good to me, just the two pattern match with headOption that can be clearer! 👍 🚀 
also good work for the great messages for diagnosing things 😄

ajosephides:
👍 👍 👍 👍 👍

src/main/scala/com/gu/autoCancel/Lambda.scala
jacobwinch: Thanks John - much neater!

PR #6, opened 2017-02-13
Add mandatory query strings to CloudFormation; make naming consistent

jacobwinch:
@paulbrown1982

### DIFF
--- a/cloudformation.yaml
+++ b/cloudformation.yaml
@@ -149,8 +149,8 @@ Resources:
             HttpMethod: POST
             AuthorizationType: NONE
             RequestParameters:
-              method.request.querystring.apiuser: true
-              method.request.querystring.apipass: true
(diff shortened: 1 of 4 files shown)


cloudformation.yaml
```diff
@@ -146,8 +146,11 @@ Resources:
             AuthorizationType: NONE
             RestApiId: !Ref ZuoraAutoCancelAPI
             ResourceId: !Ref ZuoraAutoCancelProxyResource
-            HttpMethod: ANY
+            HttpMethod: POST
             AuthorizationType: NONE
+            RequestParameters:
+              method.request.querystring.apiuser: true
+              method.request.querystring.apipass: true
```
paulbrown1982: I think this should be 'apiclientid' and 'apitoken' rather than user and pass as it better represents the type of authentication and it being a robot client not a human user.

PR #9, opened 2017-03-13
Handle accounts with multiple unpaid invoices

jacobwinch:
Currently this Lambda will refuse to auto-cancel an account with multiple unpaid invoices.

As soon as we get 100% confirmation that the historical backlog can be cleared down, we will need to start processing accounts in this state. This PR allows us to do that.

Setting the cancellation date == due date of the first unpaid invoice, means that the negative invoice which is subsequently created in Zuora has the correct amount to balance out all outstanding positive charges.

@paulbrown1982 @johnduffell @ajosephides

### DIFF
--- a/src/main/scala/com/gu/autoCancel/Lambda.scala
+++ b/src/main/scala/com/gu/autoCancel/Lambda.scala
@@ -97,17 +97,14 @@ object Lambda extends App with Logging {
 
   def getCancellationDateFromInvoices(accountSummary: AccountSummary, dateToday: LocalDate): String \/ LocalDate = {
     val unpaidAndOverdueInvoices = accountSummary.invoices.filter { invoice => invoiceOverdue(invoice, dateToday) }
-    unpaidAndOverdueInvoices match {
-      case invoices @ (invoice :: _) => {
-        logger.info(s"Found at least one unpaid invoices for account: ${accountSummary.basicInfo.id}. Invoice id(s): ${unpaidAndOverdueInvoices.map(_.id)}")
-        val earliestDueDate = invoices.map(_.dueDate).min
-        logger.info(s"Earliest overdue invoice for account ${accountSummary.basicInfo.id} has due date: $earliestDueDate. Setting this as the cancellation date.")
-        earliestDueDate.right
-      }
-      case Nil => {
-        logger.error(s"Failed to find an unpaid invoice that was overdue. The invoices we got were: ${accountSummary.invoices}")
-        "No unpaid and overdue invoices found!".left
-      }
+    if (unpaidAndOverdueInvoices.isEmpty) {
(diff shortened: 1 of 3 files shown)


jacobwinch:
(This is 'DO NOT MERGE' pending final sign-off)

ajosephides:
👍  (when written sign off given and date decided so business can manage a churn number)

src/main/scala/com/gu/autoCancel/Lambda.scala
```diff
@@ -87,36 +88,34 @@ object Lambda extends App with Logging {
...
   }
 
-  def getOverdueUnpaidInvoice(accountSummary: AccountSummary, dateToday: LocalDate): String \/ Invoice = {
+  def getCancellationDateFromInvoices(accountSummary: AccountSummary, dateToday: LocalDate): String \/ LocalDate = {
     val unpaidAndOverdueInvoices = accountSummary.invoices.filter { invoice => invoiceOverdue(invoice, dateToday) }
     unpaidAndOverdueInvoices match {
-      case invoice :: Nil => {
-        logger.info(s"Determined that InvoiceId: ${invoice.id} is the overdue unpaid invoice for Account Id: ${accountSummary.basicInfo.id}")
-        invoice.right
+      case invoices @ (invoice :: _) => {
```
johnduffell: that's clever, what does that pattern match do?
gets the invoices as a list into `invoices`, only where invoice is the first one?

johnduffell: maybe you don't actually need the `@ (invoice :: _)` bit? might be simpler to just say
```scala
if (unpaidAndOverdueInvoices.isEmpty) {
 // log and return the left
} else {
  // do the stuff
}
```

jacobwinch: Finds a non-empty list of Invoices (first element must be an invoice, but the tail can be anything, including Nil presumably) and sets the full list = invoices so that we can work with it easily.

Taken from this StackOverflow answer:
http://stackoverflow.com/questions/34585248/how-to-use-pattern-match-get-a-nonempty-list-in-scala

johnduffell: fair enough, I'm still not sure if it's matching `invoice` or extracting into it, but you could just use `@ _ :: _` if you really want or use `.nonEmpty` or `isEmpty` which might be less surprising!

jacobwinch: Yeah I agree that if/else is simpler here - thanks!

paulbrown1982:
👍 looks good - will be great to get a lot more invoices cleared down!

johnduffell:
code looks good now! thanks! 👍

PR #10, opened 2017-03-21
Refuse to process auto-cancellation for an account if autopay is not true

jacobwinch:
@paulbrown1982 

This is very ugly but it does the trick.

Note that this will exclude users who pay by cheque from the auto-cancel process @ajosephides

### DIFF
--- a/src/main/scala/com/gu/autoCancel/Lambda.scala
+++ b/src/main/scala/com/gu/autoCancel/Lambda.scala
@@ -45,7 +45,8 @@ object Lambda extends App with Logging {
 
   import org.apache.commons.io.output.NullOutputStream
   val testXML = <callout>
-                  <parameter name="AccountID">12345</parameter>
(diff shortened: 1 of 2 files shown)


src/main/scala/com/gu/autoCancel/Lambda.scala
```diff
@@ -71,11 +72,19 @@ object Lambda extends App with Logging {
     }
   }
 
+  //  FIXME - we should use JSON instead of XML now that Zuora support it
   def parseXML(xmlBody: Elem): String \/ String = {
```
jacobwinch: In case you are wondering why I chose to handle the auto-pay value in this function (as opposed to when handling the account summary):

Unfortunately the response to the getAccountSummary call does not contain the autopay status for the account, and I don't want to make another call just to get this value. However, it's possible to send the auto-pay status as part of the event notification, so that's what I have opted to do.

PR #12, opened 2017-03-31
Disable auto-pay after processing a cancellation

jacobwinch:
This ensures that the customer will not be automatically charged after auto-cancel has run.

@paulbrown1982 @ajosephides

### DIFF
--- a/src/main/scala/com/gu/autoCancel/Lambda.scala
+++ b/src/main/scala/com/gu/autoCancel/Lambda.scala
@@ -100,7 +100,8 @@ object Lambda extends App with Logging {
       cancellationDate <- getCancellationDateFromInvoices(accountSummary, date)
       updateSubscription <- restService.updateCancellationReason(subToCancel)
       cancelSubscription <- restService.cancelSubscription(subToCancel, cancellationDate)
-      result <- handleZuoraResults(updateSubscription, cancelSubscription)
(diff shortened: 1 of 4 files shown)


src/main/scala/com/gu/autoCancel/Lambda.scala
```diff
@@ -145,8 +146,12 @@ object Lambda extends App with Logging {
...
   }
 
-  def handleZuoraResults(updateSubscriptionResult: UpdateSubscriptionResult, cancelSubscriptionResult: CancelSubscriptionResult): String \/ Unit = {
-    if (!cancelSubscriptionResult.success || !updateSubscriptionResult.success) {
+  def handleZuoraResults(
+    updateSubscriptionResult: UpdateSubscriptionResult,
+    cancelSubscriptionResult: CancelSubscriptionResult,
+    updateAccountResult: UpdateAccountResult
+  ): String \/ Unit = {
+    if (!updateSubscriptionResult.success || !cancelSubscriptionResult.success || !updateAccountResult.success) {
```
paulbrown1982: Not something I think you should change now, but FWIW I think OR-NOTs can get a bit confusing. I think it's far safer to say:

```
if (!(updateSubscriptionResult.success && cancelSubscriptionResult.success && updateAccountResult.success)) {
```

PR #13, opened 2017-05-24
Respond to Zuora callouts with appropriate status codes 

jacobwinch:
Zuora callouts (webhooks) are often made for accounts which do not meet the criteria for auto-cancellation (e.g. if AutoPay is false). 

The system is configured to automatically retry in the event of an error, so Zuora continues to send callouts for accounts which will never be accepted by the process.

This PR attempts to use the correct status codes to mitigate this behaviour and prevent retries.

cc @johnduffell @pvighi @michaelwmcnamara

### DIFF
--- a/src/main/scala/com/gu/autoCancel/APIGatewayResponse.scala
+++ b/src/main/scala/com/gu/autoCancel/APIGatewayResponse.scala
@@ -2,14 +2,14 @@ package com.gu.autoCancel
 
 import com.gu.autoCancel.ResponseWriters._
 import java.io.{ OutputStream, OutputStreamWriter }
-import com.gu.autoCancel.ResponseModels.{ Headers, Response }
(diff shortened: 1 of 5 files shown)


johnduffell:
so a left (500?) is causing retries, however if we send a forbidden then it classes it as a successful attempt and leaves it up to us?

johnduffell:
looks good assuming that's what we expected! 👍

jacobwinch:
Yep, I'm hoping that Zuora respect: "The server understood the request, but is refusing to fulfill it. Authorization will not help and the request SHOULD NOT be repeated."

johnduffell:
to be honest it's not really a refusal, it's an acceptance but no action needed.  it's not like you're refusing the fact that it's a payment failure or whatever.

PR #14, opened 2017-05-25
Return 200s if no processing is required

jacobwinch:
@johnduffell it turns out Zuora also retry when they receive a 403. So this PR incorporates your last comment on https://github.com/guardian/zuora-auto-cancel/pull/13 in order to fix this issue properly!


jacobwinch:
I'll need to implement some better monitoring around this process too, so that we know if we get a sudden spike in these types of request - but I'll address that in a separate PR.

johnduffell:
😆 👍 🚀

PR #20, opened 2017-09-15
Temporarily exclude PayPal users from payment failure process

jacobwinch:
Pending a resolution to the increased number of PayPal errors we are seeing, we want to exclude these users from the payment failure process.

This change means that:
1. If a user fails a 1st or 2nd PayPal payment, they won't get an email notification
2. We will switch off autoPay so that no future payments (or retries) are attempted for these users.

Since autoPay is false, the check [here](https://github.com/guardian/zuora-auto-cancel/blob/master/src/main/scala/com/gu/autoCancel/Lambda.scala#L87) means that these users will not be cancelled on the 21 days overdue callout either.

We log the fact that we've done this, and we should also be able to observe this happening in @johnduffell's anomalies email, as well as by running standard reports in Zuora. 

Obviously, this means that we will need to revert this PR and perform some tidy up for these users once we have got to the bottom of this incident.

cc @paulbrown1982 @pvighi

### DIFF
--- a/src/main/scala/com/gu/paymentFailure/Lambda.scala
+++ b/src/main/scala/com/gu/paymentFailure/Lambda.scala
@@ -43,7 +43,7 @@ trait PaymentFailureLambda extends Logging {
                   case -\/(errorResponse) => outputForAPIGateway(outputStream, errorResponse)
                   case \/-(updateAccountResult) if (!updateAccountResult.success) => outputForAPIGateway(outputStream, internalServerError("Failed to switch off AutoPay"))
                   case \/-(updateAccountResult) if (updateAccountResult.success) => {
-                    logger.info(s"$accountId | AutoPay disabled due to ongoing PayPal incident. Don't forget to turn me back on")
+                    logger.info(s"$accountId | AutoPay disabled due to ongoing PayPal incident. Don't forget to turn this setting back on")
                     outputForAPIGateway(outputStream, noActionRequired("payment failure process is currently suspended for PayPal"))
                   }
                 }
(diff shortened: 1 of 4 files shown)


paulbrown1982:
https://en.wikipedia.org/wiki/Speedy_Gonzales

paulbrown1982:
👍 LGTM nice that you already had a method to disable auto-pay!

src/main/scala/com/gu/paymentFailure/Lambda.scala
```diff
@@ -36,9 +36,22 @@ trait PaymentFailureLambda extends Logging {
...
               logger.info(s"received ${loggableData(callout.value)}")
-              enqueueEmail(callout.value) match {
-                case -\/(error) => outputForAPIGateway(outputStream, internalServerError(error))
-                case \/-(_) => outputForAPIGateway(outputStream, successfulCancellation)
+              if (callout.value.paymentMethodType == "PayPal") {
+                val accountId = callout.value.accountId
+                logger.info(s"${accountId} | user pays by PayPal, will not send email due to PayPal/Zuora integration issues")
+                zuoraService.disableAutoPay(accountId) match {
+                  case -\/(errorResponse) => outputForAPIGateway(outputStream, errorResponse)
+                  case \/-(updateAccountResult) if (!updateAccountResult.success) => outputForAPIGateway(outputStream, internalServerError("Failed to switch off AutoPay"))
```
johnduffell: funny that we have 2 types of error

johnduffell:
looks good to me! 👍 quick work 🚀

src/main/scala/com/gu/paymentFailure/Lambda.scala
jacobwinch: Yes this should really all be handled inside the disableAutoPay method - but that's one for another day!

pvighi:
looks ok 👍

PR #22, opened 2017-09-28
Clean up project naming and structure

jacobwinch:
We never really tidied this project up after adding the payment failure emails during the design sprint. This PR is the first attempt at that.

Nearly all of these changes are renaming. I've also moved all code that's shared by both lambdas into a util package.

cc @pvighi 

I was also thinking that we should rename this project in GitHub - to something like 'payment-failure-lambdas'?


src/main/scala/com/gu/util/ApiGatewayResponse.scala
```diff
@@ -0,0 +1,51 @@
...
+  def outputForAPIGateway(outputStream: OutputStream, response: ApiResponse): Unit = {
+    val writer = new OutputStreamWriter(outputStream, "UTF-8")
+    val jsonResponse = Json.toJson(response)
+    logger.info(s"Response will be: \n ${jsonResponse.toString}")
+    writer.write(Json.stringify(jsonResponse))
+    writer.close()
+  }
+
+  val successfulExecution = ApiResponse("200", new Headers, "Success")
+  def noActionRequired(reason: String) = ApiResponse("200", new Headers, s"Auto-cancellation is not required: $reason")
```
pvighi: maybe we should not hard code auto-cancellation here. maybe it could come already concatenated in the reason or something. 
Not really important since we don't use this response for payment failure but still

src/main/scala/com/gu/util/ApiGatewayResponse.scala
```diff
@@ -0,0 +1,51 @@
...
+
+  val unauthorized = ApiResponse("401", new Headers, "Credentials are missing or invalid")
+  val badRequest = ApiResponse("400", new Headers, "Failure to parse XML successfully")
+  def internalServerError(error: String) = ApiResponse("500", new Headers, s"Failed to process auto-cancellation with the following error: $error")
```
pvighi: same here as the other message it should probably not say auto-cancellation

PR #24, opened 2017-10-02
Read config from S3

jacobwinch:
As per https://github.com/guardian/fulfilment-lookup/pull/8.

cc @pvighi 

Pre-merge:

- [x] Upload PROD config file to S3
- [x] Update private conf in Zuora
- [x] Cloudform 

Post-merge:

- [ ] Tidy up old env variables
- [ ] New PR(s) to delete KMS key & references from Cloudformation


cloudformation.yaml
```diff
@@ -45,60 +45,12 @@ Resources:
                           - Effect: Allow
                             Action: sqs:*
                             Resource: !Sub arn:aws:sqs:${AWS::Region}:${AWS::AccountId}:subs-welcome-email
-
```
jacobwinch: We don't use this KMS key anymore...

src/main/scala/com/gu/autoCancel/AutoCancelHandler.scala
```diff
@@ -24,16 +22,26 @@ object AutoCancelHandler extends App with Logging {
...
+          logger.info("Authenticated request successfully...")
+          val xmlBody = extractXmlBodyFromJson(inputEvent)
+          cancellationAttemptForPayload(config.zuoraRestConfig, xmlBody, outputStream)
+        } else {
+          logger.info("Request from Zuora could not be authenticated")
+          outputForAPIGateway(outputStream, unauthorized)
+        }
+      }
+      case Failure(configLoadError) => {
+        outputForAPIGateway(outputStream, internalServerError(s"Failed to execute AutoCancel lambda due to $configLoadError"))
```
pvighi: is there any chance of this configLoad error containing secrets ? If there is we should not return it in the api response

src/main/scala/com/gu/paymentFailure/PaymentFailureHandler.scala
```diff
@@ -1,75 +1,85 @@
...
-      logger.info("Request from Zuora could not be authenticated")
-      outputForAPIGateway(outputStream, unauthorized)
+          } else {
+            logger.info("Request from Zuora could not be authenticated")
+            outputForAPIGateway(outputStream, unauthorized)
+          }
+        }
+      }
+      case Failure(configLoadError) => {
+        outputForAPIGateway(outputStream, internalServerError(s"Failed to execute PaymentFailure lambda due to $configLoadError"))
```
pvighi: We are also returning the configuration load error here.

Also, maybe we could break up this handler into smaller functions to make it more readable instead of having so many nested blocks. 
We could refactor it in another PR if it's too much work but maybe in this one we could move all the current code into a function to avoid it making the handler even bigger:
```
configAttempt match {
  case Success(config) =>  doStuff(config)
  case Failure(configLoadError) =>  outputForAPIGateway(outputStream, internalServerError(s"Failed to execute 
      PaymentFailure lambda due to config error)) 
```

jacobwinch: Although I only opted for the minimal change in this PR, I agree with your comment about this handler function - generally it's getting a bit unwieldy! I will address this properly in a future PR.

PR #26, opened 2017-10-05
Add alarms for PROD stack

jacobwinch:
Very similar to https://github.com/guardian/fulfilment-lookup, with slightly more generous error thresholds. Again, these alarms will be created for the PROD stack only.

This change is useful because we've completely broken this process for direct debit users before (the API was being called with invalid credentials), and no-one noticed for several days because there was no monitoring configured. 

This PR just covers the very basics, but it should allow us to find out if things are fundamentally broken, which seems important as we're likely to make more regular changes to this project in support of KR1.

As with all alarms, the usual 'let's tune these as we go' disclaimers apply...

cc @paulbrown1982 @pvighi @lmath @AWare 

After +1 and before merge:

- [x] Cloudform PROD stack

### DIFF
--- a/cloudformation.yaml
+++ b/cloudformation.yaml
@@ -2,10 +2,6 @@ AWSTemplateFormatVersion: "2010-09-09"
 Description: Handles auto-cancellations for membership and subscriptions, using API Gateway and Lambda
 
 Parameters:
-    App:
-        Description: Application name
-        Type: String
-        Default: zuora-auto-cancel
     Stage:
         Description: Stage name
         Type: String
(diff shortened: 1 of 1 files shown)


cloudformation.yaml
```diff
@@ -2,17 +2,22 @@ AWSTemplateFormatVersion: "2010-09-09"
 Description: Handles auto-cancellations for membership and subscriptions, using API Gateway and Lambda
 
 Parameters:
-    App:
```
jacobwinch: This wasn't being used

paulbrown1982:
👍

PR #27, opened 2017-10-31
Use JSON as input and re-factor

jacobwinch:
Zuora now supports JSON callouts, so this PR changes the code to read a JSON body, rather than XML.

We've also taken the opportunity to have a general tidy-up/refactor of the lambda logic.

cc @johnduffell

### DIFF
--- a/src/main/scala/com/gu/autoCancel/AutoCancelHandler.scala
+++ b/src/main/scala/com/gu/autoCancel/AutoCancelHandler.scala
@@ -94,22 +94,6 @@ object AutoCancelHandler extends App with Logging {
     if (callout.isAutoPay) \/-(()) else -\/(noActionRequired("AutoPay is false"))
   }
 
-  /* When developing, it's best to bypass handleRequest (since this requires actually invoking the Lambda)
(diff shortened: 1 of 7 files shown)


src/main/scala/com/gu/autoCancel/AutoCancelHandler.scala
```diff
@@ -4,105 +4,112 @@ import com.amazonaws.services.lambda.runtime.Context
...
+
+  def parseBody(apiGatewayRequest: ApiGatewayRequest): ApiResponse \/ AutoCancelCallout = {
+    apiGatewayRequest.parsedBody match {
+      case JsSuccess(autoCancelCallout, _) => \/-(autoCancelCallout)
+      case JsError(error) => {
+        logger.error(s"Error when parsing JSON from API Gateway: $error")
+        -\/(badRequest)
+      }
+    }
+  }
```
johnduffell: in future this and `parseApiGatewayInput` could probably be combined

johnduffell:
👍 🚀

PR #28, opened 2017-11-01
add a direct-debit only filter to the url params for auto cancel

johnduffell:
We are now auto cancelling everything at 28 days.  We would like to add a new trigger to cancel DD only at 21 days overdue.

We will use the same lambda for both triggers but we need to let the lambda decide whether to go ahead.

This PR adds a new parameter for the URL (optional) so you can add &onlyCancelDirectDebit=true to the url to get it to not cancel others.  This will be checked against the Account.PaymentMethod.Type parameter, if it's "BankTransfer" we know it's a DD and can be cancelled even in the constrained case.

Next step is to deploy to code and follow the steps in the google doc to test it.  Then we can add the body parameter in prod (this won't break the old lambda), deploy to PROD and then add the 21 day trigger.

@jacobwinch

### DIFF
--- a/src/main/scala/com/gu/autoCancel/AutoCancelCallout.scala
+++ b/src/main/scala/com/gu/autoCancel/AutoCancelCallout.scala
@@ -4,9 +4,11 @@ import play.api.libs.json.Json
 
 case class AutoCancelCallout(
     accountId: String,
-    autoPay: String
+    autoPay: String,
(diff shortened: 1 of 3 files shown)


jacobwinch:
These changes look good to me 👍 I've tested in CODE using Postman and this is behaving as expected (after a bit of confusion on my part - see commit history!)

PR #29, opened 2017-11-14
Call ET directly from the lambda

johnduffell:
This PR changes the payment failure lambda to call ET directly rather than through membership workflow.  This is alright because it happens from a Zuora callout which will retry several times so no queue is needed.

I have also added configuration for additional numbers of retries.

Todo before deployment:
DONE remove the UAT callout on the External Payment made trigger (added for testing) that goes to the CODE lambda
DONE add the new et send ids for the pf3/pf4/overdue29 into this code in PROD - make sure the ignored PROD config test passes locally
add the new callouts for the pf3/pf4/overdue29 into this code in CODE
make sure the extra fields are on the prod auto cancel callout in PROD (needed for the email)

Todo later when the email copy is correct:
add triggers to prod zuora for PF3 and 4
uncomment the line that sends emails after an auto cancel event and deploy again

==
other rough notes include that I've tried a new way of separating out the code to avoid testability issues.  This has a more severe manifestation in the TouchpointBackend situation, but here is a smaller situation to try to do this better.  Hopefully we can apply learnings about this in case we ever get to the point of breaking down touchpoint backend.
The only funny thing I found was to do with building up the Deps tree (i.e. saying what code depends on what other code).  I made a `default` function to make it easily during normal running, but it's still about 3 steps to get between a function call and its default implementation.  With Cake pattern it would be click once to the trait, then go to look in the object for the real dependency, then click into that and find the method.  Or just assume the default implementaiton would be obvious from the trait.
The disadvantage of cake over `.default` is you can't curry the raw dependencies into the dependent function, forcing you to pass http clients etc all the way down (breaking encapulation).

The typeclass pattern works quite well when you have parallel type hierarchies, so there might be a way of making the `default` all implicits and using the typeclass pattern to make it pick them up automatically but I'm not sure I want to go down that route.


src/main/scala/com/gu/paymentFailure/PaymentFailureCallout.scala
```diff
@@ -18,5 +22,21 @@ case class PaymentFailureCallout(
 )
 
 object PaymentFailureCallout {
-  implicit val jf = Json.reads[PaymentFailureCallout]
-}
\ No newline at end of file
+  //  implicit val jf = Json.reads[PaymentFailureCallout]
```
jacobwinch: Delete me

src/main/scala/com/gu/paymentFailure/PaymentFailureSteps.scala
```diff
@@ -0,0 +1,35 @@
...
+import com.gu.util.zuora.Zuora
+import play.api.libs.json.Json
+
+object PaymentFailureSteps extends Logging {
+
+  def loggableData(callout: PaymentFailureCallout) = {
+    s"accountId: ${callout.accountId}, paymentId: ${callout.paymentId}, failureNumber: ${callout.failureNumber}, paymentMethodType: ${callout.paymentMethodType}, currency: ${callout.currency}"
+  }
+
+  def performZuoraAction(deps: PFDeps = defaultPFDeps)(apiGatewayRequest: ApiGatewayRequest): ZuoraOp[Unit] = {
```
jacobwinch: Hmm, in this case I'm not sure I agree with this naming. Since we're sending an email too it's not strictly a ZuoraAction anymore

src/main/scala/com/gu/util/Auth.scala
```diff
@@ -1,36 +1,19 @@
...
-import play.api.libs.json.JsValue
+import com.gu.effects.Logging
+import com.gu.util.apigateway.RequestAuth
 
 object Auth extends Logging {
 
-  def credentialsAreValid(requestAuth: RequestAuth, trustedApiConfig: TrustedApiConfig): Boolean =
-    (requestAuth.apiClientId == trustedApiConfig.apiClientId && requestAuth.apiToken == trustedApiConfig.apiToken)
-
-  def deprecatedCredentialsAreValid(inputEvent: JsValue, trustedApiConfig: TrustedApiConfig): Boolean = {
```
jacobwinch: 🎉

src/main/scala/com/gu/util/ConfigLoader.scala
```diff
@@ -15,6 +26,45 @@ object ZuoraRestConfig {
...
+object ETConfig {
+
+  implicit val mapReads: Reads[Map[Int, String]] = new Reads[Map[Int, String]] {
+
+    override def reads(json: JsValue) = {
+      json match {
+        case JsObject(map) =>
+          val jsResults = map.map({
+            case (num: String, JsString(key)) =>
+              val aaa = Try { Integer.parseInt(num) }
```
jacobwinch: Presumably renaming this is also on your list, but just in case it's not, here's a comment to remind you!

src/main/scala/com/gu/util/ConfigLoader.scala
```diff
@@ -15,6 +26,45 @@ object ZuoraRestConfig {
...
+              aaa match {
+                case Success(num) => JsSuccess((num, key))
+                case Failure(f) => JsError(s"num parse error in config: ${f}")
+              }
+            case other => JsError(s"value of TS key wasn't a string: $other")
```
jacobwinch: What is a TS key?

src/main/scala/com/gu/util/ConfigLoader.scala
```diff
@@ -25,39 +75,29 @@ object TrustedApiConfig {
...
-
   def parseConfig(jsonConfig: String): Try[Config] = {
     Json.fromJson[Config](Json.parse(jsonConfig)) match {
       case validConfig: JsSuccess[Config] => {
         logger.info(s"Successfully parsed JSON config")
         Success(validConfig.value)
       }
       case error: JsError => {
-        logger.error("Failed to parse JSON config")
+        logger.error(s"Failed to parse JSON config: $error")
```
jacobwinch: I think I used to log the actual error here too, but myself and @pvighi were concerned that this could leak the secrets from the config into the logs...

src/main/scala/com/gu/util/apigateway/ApiGatewayHandler.scala
```diff
@@ -0,0 +1,46 @@
...
+import com.gu.util.zuora.Types._
+import play.api.libs.json.Json
+
+import scala.io.Source
+import scala.util.Try
+import scalaz.{ -\/, Reader, \/, \/- }
+
+object ApiGatewayHandler extends Logging {
+
+  /* Entry point for our Lambda - this takes the input event from API Gateway,
```
jacobwinch: This comment can go... perhaps we should replace it by explaining that this is general code for handling requests from API Gateway - but maybe the naming and location of this file are sufficient for making that clear

src/main/scala/com/gu/autoCancel/AutoCancelHandler.scala
```diff
@@ -1,192 +1,12 @@
...
-      apiGatewayRequest <- parseApiGatewayInput(inputStream)
-      auth <- authenticateCallout(apiGatewayRequest.requestAuth, config.trustedApiConfig)
-      _ = logger.info("Authenticated request successfully...")
-      _ = logger.info(s"Body from Zuora was: ${apiGatewayRequest.body}, onlyDirectDebit queryString was: ${apiGatewayRequest.queryStringParameters.onlyCancelDirectDebit}")
-      autoCancelCallout <- parseBody(apiGatewayRequest)
-      _ <- filterInvalidAccount(autoCancelCallout, onlyCancelDirectDebit = apiGatewayRequest.onlyCancelDirectDebit)
-    } yield cancellationAttemptForPayload(config.zuoraRestConfig, autoCancelCallout)
-    outputForAPIGateway(outputStream, response.fold(identity, identity))
-  }
+  def handleRequest = ApiGatewayHandler() {
```
jacobwinch: I think it's still useful to note that this is the entry point for the lambda (especially for any engineers who aren't used to working with lambdas) - it might also be worth noting that the name of this function matters, as it's referenced in CloudFormation. (I've forgotten about this before and broken everything).

src/main/scala/com/gu/paymentFailure/PaymentFailureHandler.scala
```diff
@@ -1,178 +1,11 @@
...
-  }
-
-  def unpaid(itemisedInvoice: ItemisedInvoice): Boolean = {
-    itemisedInvoice.balance > 0 && itemisedInvoice.status == "Posted"
-  }
-
-  def invoiceItemFilter(item: InvoiceItem): Boolean = {
-    item.chargeAmount > 0 // remove discounts, holiday credits and free products
-  }
+  def handleRequest = ApiGatewayHandler() {
```
jacobwinch: Might be worth an 'entry point' comment here too

src/main/scala/com/gu/paymentFailure/PaymentFailureSteps.scala
johnduffell: actually the whole ZuoraOp also includes the whole config now, I need to rethink the name

src/main/scala/com/gu/util/ConfigLoader.scala
johnduffell: ET triggeredSend, I'll update the message

src/main/scala/com/gu/util/ConfigLoader.scala
johnduffell: I can take it out, but if it doesn't go into the logs, where can it go?  maybe warn level would be nicer.
I wouldn't necessarily class it as a leak since Config in S3 is just as secure as in Cloudwatch.

src/main/scala/com/gu/effects/AwsS3.scala
```diff
@@ -0,0 +1,60 @@
+package com.gu.effects
```
johnduffell: anything in the "effects" package is not testable without calling out to something

src/main/scala/com/gu/paymentFailure/PaymentFailureCallout.scala
```diff
@@ -1,11 +1,15 @@
...
+import play.api.libs.functional.syntax._
 import play.api.libs.json._
 
+import scala.util.Try
+
 case class PaymentFailureCallout(
   accountId: String,
   email: String,
-  failureNumber: String,
+  failureNumber: Int,
```
johnduffell: I made this an int, but in reality there are a fair few other things in here that shouldn't be strings

src/main/scala/com/gu/paymentFailure/ToMessage.scala
```diff
@@ -0,0 +1,62 @@
...
+import org.joda.time.LocalDate
+
+import scala.math.BigDecimal.decimal
+
+object ToMessage {
+
+  def apply(paymentFailureCallout: PaymentFailureCallout, paymentFailureInformation: PaymentFailureInformation) = EmailRequest(
+    attempt = paymentFailureCallout.failureNumber,
+    Message(
+      DataExtensionName = dataExtensionNameForAttempt(paymentFailureCallout.failureNumber),
```
johnduffell: not sure if this field is needed - rachel didn't recognise it but we send it in the payload so I'll keep it for now

src/main/scala/com/gu/util/ConfigLoader.scala
```diff
@@ -25,39 +77,29 @@ object TrustedApiConfig {
   )(TrustedApiConfig.apply _)
 }
 
-case class Config(trustedApiConfig: TrustedApiConfig, zuoraRestConfig: ZuoraRestConfig)
+case class Config(stage: String, trustedApiConfig: TrustedApiConfig, zuoraRestConfig: ZuoraRestConfig, etConfig: ETConfig)
 
 object Config extends Logging {
 
   implicit val configReads: Reads[Config] = (
+    (JsPath \ "stage").read[String] and
```
johnduffell: stage is now in the config so that if we accidentally drop the prod config into code or vice versa when doing the S3 copies, we won't be able to start up.  I think that's a good idea, if people agree?

src/main/scala/com/gu/util/ConfigLoader.scala
```diff
@@ -25,39 +77,29 @@ object TrustedApiConfig {
...
         logger.info(s"Successfully parsed JSON config")
         Success(validConfig.value)
-      }
-      case error: JsError => {
-        logger.error("Failed to parse JSON config")
-        Failure(new ConfigFailure(error))
-      }
+      case error: JsError =>
+        logger.error(s"Failed to parse JSON config")
+        logger.warn(s"Failed to parse JSON config: $error")
```
johnduffell: the error itself is not logged at error level just in case we sentry it at some point, does that seems sensible?

src/main/scala/com/gu/util/apigateway/ApiGatewayHandler.scala
```diff
@@ -0,0 +1,75 @@
...
+import scala.io.Source
+import scala.util.Try
+import scalaz.{ -\/, \/, \/- }
+
+object ApiGatewayHandler extends Logging {
+
+  //  code dependencies
+  case class HandlerDeps(
+    parseConfig: String => Try[Config] = Config.parseConfig
+  )
```
johnduffell: I've put the wiring bewtween different bits of code in this cake style object, but the config etc I've used the reader.  I feel like I might have done better to just stick or one or the other though?

src/main/scala/com/gu/util/apigateway/ApiGatewayHandler.scala
```diff
@@ -0,0 +1,75 @@
...
+
+  def loadConfig(rawEffects: RawEffects, deps: HandlerDeps = HandlerDeps()) = {
+    val stage = rawEffects.stage()
+    logger.info(s"${this.getClass} Lambda is starting up in $stage")
+
+    for {
+
+      textConfig <- rawEffects.s3Load(stage).toFailableOp("load config from s3")
+      config <- deps.parseConfig(textConfig).toFailableOp("parse config file")
+      _ <- if (stage == config.stage) { \/-(()) } else { -\/(ApiGatewayResponse.internalServerError(s"running in $stage with config from ${config.stage}")) }
```
johnduffell: stage is now in the config so that if we accidentally drop the prod config into code or vice versa when doing the S3 copies, we won't be able to start up. I think that's a good idea, if people agree?

src/main/scala/com/gu/util/reader/Types.scala
```diff
@@ -0,0 +1,79 @@
...
+  type FailableOp[A] = ApiResponse \/ A
+
+  // unfortunately the trait is needed for the monad transformer to provide a context for the state type I
+  // since the EitherT type constructor first argument is specifically a higher kinded type with arity one but Reader has arity two
+  // see pages such as https://meta.plasm.us/posts/2015/07/11/roll-your-own-scala/
+  // this leaves us with a kind of curried type constructor
+  trait WithDeps[I] {
+    type XReader[A] = Reader[I, A] // needed becuase EitherT first type param is a single arity
+    type FailableOp[A] = EitherT[XReader, ApiResponse, A]
+  }
```
johnduffell: if we don't like the trait, we can use type lambdas to do it all in one go.. I'm not sure it is better though
```scala

  type WithDepsFailableOp[I, A] = EitherT[({ type XReader[AA] = Reader[I, AA] })#XReader, ApiResponse, A]
```
the really horrible bit is that then the below implicit class looks like this:
```scala
  // lets us use the Reader .local function through the EitherT
  // this is important to let us have different dependencies for different functions
  implicit class WithDepsFailableOpOps[A, F](httpFailable: WithDepsFailableOp[F, A]) {

    def local[I](func: I => F): WithDepsFailableOp[I, A] = {
      EitherT[({ type XReader[AA] = Reader[I, AA] })#XReader, ApiResponse, A](httpFailable.run.local[I](func))
    }

  }
```

johnduffell:
so I suddenly realised there's no point doing the callout to trigger an email on the 5th failure as the auto cancel is on the 28th day overdue.  I'm thinking we need to change the auto cancel to trigger it on a successful auto cancellation.

src/main/scala/com/gu/autoCancel/AutoCancelFilter.scala
```diff
@@ -0,0 +1,27 @@
+package com.gu.autoCancel
+
+import com.gu.util.Logging
+import com.gu.util.apigateway.ApiGatewayResponse.noActionRequired
+import com.gu.util.reader.Types.FailableOp
+
+import scalaz.{ -\/, \/- }
+
+object AutoCancelFilter extends Logging {
```
jacobwinch: What about 'AutoCancelInputFilter'...

src/main/scala/com/gu/autoCancel/AutoCancelFilter2.scala
```diff
@@ -0,0 +1,83 @@
...
+import com.gu.util.reader.Types.{ FailableOp, _ }
+import com.gu.util.zuora.Zuora
+import com.gu.util.zuora.ZuoraModels.{ AccountSummary, Invoice, SubscriptionId }
+import com.gu.util.{ Logging, ZuoraRestConfig }
+import okhttp3.{ Request, Response }
+import org.joda.time.LocalDate
+
+import scalaz.Scalaz._
+
+object AutoCancelFilter2 extends Logging {
```
jacobwinch: ... and 'AutoCancelDataCollectionFilter' (or a slightly shorter version of that!)

src/main/scala/com/gu/autoCancel/AutoCancelHandler.scala
```diff
@@ -1,192 +1,31 @@
...
-    } else {
-      logger.info("All Zuora calls completed successfully")
-        .right
-    }
+  // this is the entry point
+  // it's referenced by the cloudformation so make sure you keep it in step
+  // it's the only part you can't test of the handler
+  def handleRequest(inputStream: InputStream, outputStream: OutputStream, context: Context): Unit = {
+    val deps = default(RawEffects.createDefault)
+    deps.agh(inputStream, outputStream, context)
```
jacobwinch: What does 'agh' mean in this context?

src/main/scala/com/gu/autoCancel/AutoCancelSteps.scala
```diff
@@ -0,0 +1,56 @@
...
+    sendEmailRegardingAccount: (String, PaymentFailureInformation => EmailRequest) => FailableOp[Unit]
+  )
+
+  def apply(deps: AutoCancelStepsDeps)(apiGatewayRequest: ApiGatewayRequest): FailableOp[Unit] = {
+    for {
+      autoCancelCallout <- Json.fromJson[AutoCancelCallout](Json.parse(apiGatewayRequest.body)).toFailableOp.withLogging("zuora callout")
+      _ <- AutoCancelFilter(autoCancelCallout, onlyCancelDirectDebit = apiGatewayRequest.onlyCancelDirectDebit)
+      acRequest <- deps.autoCancelFilter2(autoCancelCallout).withLogging(s"auto-cancellation filter for ${autoCancelCallout.accountId}")
+      _ <- deps.autoCancel(acRequest).withLogging(s"auto-cancellation for ${autoCancelCallout.accountId}")
+      //      _ <- PaymentFailureSteps.sendEmailSteps()(paymentFailureCallout(autoCancelCallout))
```
jacobwinch: Can this go?

src/main/scala/com/gu/autoCancel/AutoCancelHandler.scala
johnduffell: 😆  api gateway handler - didn't spot the funny name

src/main/scala/com/gu/paymentFailure/PaymentFailureCallout.scala
jacobwinch: Yes, I think we just made everything a String whilst we were getting this lambda up and running, but never actually went back and improved it afterwards...

src/main/scala/com/gu/util/ConfigLoader.scala
jacobwinch: 👍

src/main/scala/com/gu/util/exacttarget/EmailSendSteps.scala
```diff
@@ -0,0 +1,188 @@
+package com.gu.util.exacttarget
```
jacobwinch: There is quite a bit of commented out code here - can we remove it?

src/main/scala/com/gu/util/exacttarget/EmailSendSteps.scala
```diff
@@ -0,0 +1,188 @@
...
+      _ <- filterTestEmail(request.message.To.Address, prod)
+    } yield ()
+
+  }
+
+  private def isProd(stage: Stage): FailableOp[Boolean] = {
+    \/.right(stage.isProd)
+  }
+
+  private def filterTestEmail(email: String, prod: Boolean): FailableOp[Unit] = {
```
jacobwinch: Nice safety check! It might be worth noting the existence of this method in some sort of testing notes in the README (I can imagine a scenario where future me wastes loads of time trying to test this with my gmail address!)

src/main/scala/com/gu/util/exacttarget/SalesforceAuthenticate.scala
```diff
@@ -0,0 +1,54 @@
...
+import com.gu.util.{ ETConfig, Logging }
+import okhttp3.{ FormBody, Request, Response }
+import play.api.libs.functional.syntax._
+import play.api.libs.json.{ JsPath, JsSuccess, Json, Reads }
+
+import scalaz.{ -\/, \/- }
+
+object SalesforceAuthenticate extends Logging {
+
+  private val authEndpoint = "https://auth.exacttargetapis.com/v1/requestToken"
```
jacobwinch: Presumably these endpoints are the same across all environments? (So it's safe to hardcode them instead of putting them in config files?)

src/main/scala/com/gu/util/exacttarget/EmailSendSteps.scala
johnduffell: good idea, althoughh hopefully the logging would tell you - actually I added this as I was frustrated that the CODE env didn't have any ET sendIDs at all so it just didn't send anything.

src/main/scala/com/gu/util/zuora/Zuora.scala
```diff
@@ -0,0 +1,29 @@
...
+import com.gu.util.reader.Types.WithDepsFailableOp
+import com.gu.util.zuora.ZuoraModels._
+import com.gu.util.zuora.ZuoraReaders._
+import com.gu.util.zuora.ZuoraWriters._
+import org.joda.time.LocalDate
+object Zuora {
+
+  import ZuoraRestRequestMaker._
+
+  def getAccountSummary(accountId: String): WithDepsFailableOp[StageAndConfigHttp, AccountSummary] =
```
jacobwinch: Rename StageAndConfigHttp

src/main/scala/com/gu/autoCancel/AutoCancel.scala
```diff
@@ -0,0 +1,24 @@
...
+import com.gu.util.zuora.Zuora
+import com.gu.util.zuora.Zuora.ZuoraDeps
+import com.gu.util.zuora.ZuoraModels.SubscriptionId
+import org.joda.time.LocalDate
+
+object AutoCancel extends Logging {
+
+  case class AutoCancelRequest(accountId: String, subToCancel: SubscriptionId, cancellationDate: LocalDate)
+
+  def apply(sc: ZuoraDeps)(acRequest: AutoCancelRequest): FailableOp[Unit] = {
```
jacobwinch: We should rename sc to zuoraDeps too!

src/main/scala/com/gu/autoCancel/AutoCancelHandler.scala
```diff
@@ -1,192 +1,31 @@
...
-  ): ApiResponse \/ Unit = {
-    if (!updateSubscriptionResult.success || !cancelSubscriptionResult.success || !updateAccountResult.success) {
-      logger.error(s"Zuora rejected one (or more) of our calls during autoCancellation")
-      internalServerError("Received at least one failure result from Zuora during autoCancellation").left
-    } else {
-      logger.info("All Zuora calls completed successfully")
-        .right
-    }
+  // this is the entry point
+  // it's referenced by the cloudformation so make sure you keep it in step
```
jacobwinch: 👍

src/main/scala/com/gu/autoCancel/AutoCancelSteps.scala
```diff
@@ -0,0 +1,55 @@
...
+  )
+
+  def apply(deps: AutoCancelStepsDeps)(apiGatewayRequest: ApiGatewayRequest): FailableOp[Unit] = {
+    for {
+      autoCancelCallout <- Json.fromJson[AutoCancelCallout](Json.parse(apiGatewayRequest.body)).toFailableOp.withLogging("zuora callout")
+      _ <- AutoCancelInputFilter(autoCancelCallout, onlyCancelDirectDebit = apiGatewayRequest.onlyCancelDirectDebit)
+      acRequest <- deps.autoCancelFilter2(autoCancelCallout).withLogging(s"auto-cancellation filter for ${autoCancelCallout.accountId}")
+      _ <- deps.autoCancel(acRequest).withLogging(s"auto-cancellation for ${autoCancelCallout.accountId}")
+      request <- makeRequest(deps.etSendIds, autoCancelCallout)
+      _ <- deps.sendEmailRegardingAccount(autoCancelCallout.accountId, request)
```
jacobwinch: Do we need to comment this out until Rachel is ready with the ET stuff?

src/main/scala/com/gu/util/zuora/ZuoraObjects.scala
```diff
@@ -21,18 +23,14 @@ object ZuoraModels {
...
-
-  case class UpdateAccountResult(success: Boolean)
-
   case class SubscriptionCancellation(cancellationEffectiveDate: LocalDate)
 
   case class SubscriptionUpdate(cancellationReason: String)
 
   case class AccountUpdate(autoPay: Boolean)
 
+  case class ZuoraCommonFields(success: Boolean)
```
jacobwinch: 👍

src/main/scala/com/gu/util/zuora/ZuoraRestRequestMaker.scala
```diff
@@ -0,0 +1,68 @@
...
+
+  def convertResponseToCaseClass[T](response: Response)(implicit r: Reads[T]): ApiResponse \/ T = {
+    if (response.isSuccessful) {
+      val bodyAsJson = Json.parse(response.body.string)
+      bodyAsJson.validate[ZuoraCommonFields] match {
+        case JsSuccess(ZuoraCommonFields(true), _) =>
+          bodyAsJson.validate[T] match {
+            case success: JsSuccess[T] => success.get.right
+            case error: JsError => {
+              logger.info(s"Failed to convert Zuora response to case case 1 $error. Response body was: \n ${bodyAsJson}")
```
jacobwinch: Should be logger.error

src/main/scala/com/gu/util/zuora/ZuoraRestRequestMaker.scala
```diff
@@ -0,0 +1,68 @@
...
+              internalServerError("Error when converting Zuora response to case class").left
+            }
+          }
+        case JsSuccess(zuoraFailure, _) =>
+          logger.error(s"Zuora rejected our call $bodyAsJson")
+          internalServerError("Received failure result from Zuora during autoCancellation").left
+        case error: JsError => {
+          logger.info(s"Failed to convert Zuora response to case case 2 $error. Response body was: \n ${bodyAsJson}")
```
jacobwinch: As above

src/main/scala/com/gu/autoCancel/AutoCancelHandler.scala
```diff
@@ -1,192 +1,31 @@
...
-      }
-      case subs => {
-        // This should be a pretty rare scenario, because the Billing Account to Sub relationship is (supposed to be) 1-to-1
-        logger.error(s"More than one subscription is Active on account: ${accountSummary.basicInfo.id}. Subscription ids are: ${activeSubs.map(_.id)}")
-        noActionRequired("More than one active sub found!").left // Don't continue because we don't know which active sub to cancel
-      }
-    }
-  }
-
-  def handleZuoraResults(
```
jacobwinch: Glad to see that this logic has been moved somewhere more sensible 💯

src/main/scala/com/gu/util/exacttarget/SalesforceAuthenticate.scala
johnduffell: there's only one ET environment, if there was another, maybe we'd have to add to config or make this a map of PROD -> {these urls}, CODE -> {other urls}

src/main/scala/com/gu/util/zuora/ZuoraRestRequestMaker.scala
jacobwinch: should this be 'to case class case 1'?

src/main/scala/com/gu/util/zuora/ZuoraRestRequestMaker.scala
jacobwinch: and 'to case class case 2'?

src/test/scala/com/gu/paymentFailure/EndToEndHandlerTest.scala
```diff
@@ -0,0 +1,198 @@
...
+object EndToEndData {
+
+  def responses: Map[String, (Int, String)] = Map(
+    ("/transactions/invoices/accounts/2c92c0f85fc90734015fca884c3f04cf", (200, invoices)),
+    ("/v1/requestToken", (200, """{"accessToken":"", "expiresIn":1}""")),
+    ("/messaging/v1/messageDefinitionSends/111/send", (202, ""))
+  )
+
+  val expectedEmailSend =
+    """{"To":{"Address":"john.duffell@guardian.co.uk","SubscriberKey":"john.duffell@guardian.co.uk","ContactAttributes":{"SubscriberAttributes":{"SubscriberKey":"john.duffell@guardian.co.uk","EmailAddress":"john.duffell@guardian.co.uk","subscriber_id":"A-S00071536","product":"Supporter","payment_method":"CreditCardReferenceTransaction","card_type":"Visa","card_expiry_date":"12/2019","first_name":"eSAFaBwm4WJZNg5xhIc","last_name":"eSAFaBwm4WJZNg5xhIc","paymentId":"2c92c0f95fc912eb015fcb2a481720e6","price":"$49.00","serviceStartDate":"17 November 2017","serviceEndDate":"16 November 2018"}}}}"""
```
jacobwinch: You might want to remove this address?

src/test/scala/manualTest/ConfigLoaderSystemTest.scala
```diff
@@ -0,0 +1,49 @@
...
+      b <- Config.parseConfig(a)
+    } yield b
+    val loadedOK = con.map(config => ())
+    withClue(configAttemp) {
+      loadedOK should be(Success(()))
+    }
+  }
+
+  it should "be able to load the local test config successfully" in {
+    //    val requestAuth = Some(RequestAuth(apiClientId = "validUser", apiToken = "ndjashjkhajshs"))
```
jacobwinch: This is commented out a few times throughout the test code - can we remove it?

src/main/scala/com/gu/util/exacttarget/SalesforceAuthenticate.scala
jacobwinch: Ah yes, of course - ignore the above!

jacobwinch:
There's a lot here, but it's looking good to me (with the exception of a few outstanding comments) 👍 

Probably worth updating the PR description based on the latest discussions on how/when to send each email too. Hopefully that'll make it clearer whether we can merge this right away, or whether we have a dependency on Rachel finishing some ET config first (I'm not 100% clear on this at the moment).

src/main/scala/com/gu/autoCancel/AutoCancelSteps.scala
johnduffell: good point

src/main/scala/com/gu/util/zuora/ZuoraRestRequestMaker.scala
johnduffell: it was just to make the messages distinct while I was debugging - I'll review them all here

johnduffell: ok in the end I pulled the different parts into different functions to make it much simpler, and added an extra test

src/test/scala/com/gu/paymentFailure/EndToEndHandlerTest.scala
johnduffell: it doesn't actually send an email, just checks what it woudl have sent.  It's the EmailClientSystemTest that should probably take a parameter or a config file though!

johnduffell:
so during the go live we discovered for the overdue29 event there's no payment and therefore no paymentId.  ET uses that as the primary key.  Now the "cancelled" data extension has been changed to use the invoiceId rather than the paymentId as the key, so I've updated the code to do the same.  A quick look would be good @jacobwinch please!

src/main/scala/com/gu/util/exacttarget/EmailSendSteps.scala
```diff
@@ -24,21 +27,52 @@ case class SubscriberAttributesDef(
...
-  paymentId: String, // ET will filter out multiple emails with the same payment id
+  primaryKey: PrimaryKey,
   price: String,
   serviceStartDate: String,
   serviceEndDate: String
-
 )
 
+sealed trait PrimaryKey {
+  // ET will filter out multiple emails with the same payment id for PF1,2,3,4
```
jacobwinch: 👍

src/test/scala/com/gu/paymentFailure/SqsMessageSerializationTest.scala
```diff
@@ -1,14 +1,14 @@
...
+import com.gu.util.exacttarget._
 import org.scalatest.FlatSpec
 import org.scalatest.Matchers._
 import org.scalatest.mockito.MockitoSugar
 import play.api.libs.json.Json
 
 class SqsMessageSerializationTest extends FlatSpec with MockitoSugar {
 
-  "Message" should "serialize to the correct json format" in {
+  "Message" should "serialize to the correct json format with payemnt id" in {
```
jacobwinch: typo

jacobwinch:
Still looks good 👍

src/test/scala/com/gu/paymentFailure/SqsMessageSerializationTest.scala
johnduffell: never mind that, how about the name of the test? doesn't do anything about SQS now! :P

PR #30, opened 2017-11-22
turn on auto cancel email from ET

johnduffell:
This is to add a step to the auto cancel lambda to send emails from ET once we have auto cancelled someone.  It can go live once the email content has been finalised.

Then we can turn off the Zuora driven email @jacobwinch 😁

### DIFF
--- a/src/main/scala/com/gu/autoCancel/AutoCancelSteps.scala
+++ b/src/main/scala/com/gu/autoCancel/AutoCancelSteps.scala
@@ -43,7 +43,7 @@ object AutoCancelSteps extends Logging {
       acRequest <- deps.autoCancelFilter2(autoCancelCallout).withLogging(s"auto-cancellation filter for ${autoCancelCallout.accountId}")
       _ <- deps.autoCancel(acRequest).withLogging(s"auto-cancellation for ${autoCancelCallout.accountId}")
       request <- makeRequest(deps.etSendIds, autoCancelCallout)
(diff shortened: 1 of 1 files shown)


src/main/scala/com/gu/autoCancel/AutoCancelSteps.scala
```diff
@@ -43,7 +43,7 @@ object AutoCancelSteps extends Logging {
       acRequest <- deps.autoCancelFilter2(autoCancelCallout).withLogging(s"auto-cancellation filter for ${autoCancelCallout.accountId}")
       _ <- deps.autoCancel(acRequest).withLogging(s"auto-cancellation for ${autoCancelCallout.accountId}")
       request <- makeRequest(deps.etSendIds, autoCancelCallout)
-      //      _ <- deps.sendEmailRegardingAccount(autoCancelCallout.accountId, request) TODO add this back once the ET setup is done
+      _ <- deps.sendEmailRegardingAccount(autoCancelCallout.accountId, request) TODO add this back once the ET setup is done
```
jacobwinch: We can remove the TODO!

johnduffell: good point, that's not going to compile well with that!

johnduffell:
this is ready to go
image

paulbrown1982:
👍 once it's confirmed that the CRM team has activated the Triggered Send in ET.

PR #31, opened 2017-12-11
Increase memory (and CPU) allocation

jacobwinch:
This should help us with the timeout issues/alarms whilst we are working on a longer term queuing strategy.

cc @johnduffell @paulbrown1982

### DIFF
--- a/cloudformation.yaml
+++ b/cloudformation.yaml
@@ -74,7 +74,7 @@ Resources:
                 Fn::GetAtt:
                 - ZuoraAutoCancelRole
                 - Arn
-            MemorySize: 512
+            MemorySize: 1536
             Runtime: java8
             Timeout: 300
         DependsOn:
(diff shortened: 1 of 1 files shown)


johnduffell:
thanks!  is it worth understanding the cost change?
probably not as it could only be triple at the most, but firstly it should be quicker and secondly less should "fail" so we won't have as many retries. 
but I'll admit I don't know how much auto cancel/PF lambda costs us. less than an auto scaling group/elb?
👍 🚀

jacobwinch:
I think the trade off between speed of execution vs memory used would be an interesting one if we were using lambdas at scale, but based on what I can work out from the billing console, our current costs for lambda usage are so low that even if we tripled the cost it would be negligible...

jacobwinch:
There's a noticeable drop in the max execution times for these lambdas:

image

johnduffell:
what does each line mean and what is the scale on the vertical?

PR #32, opened 2017-12-18
Disable autoPay before cancelling the sub

jacobwinch:
AutoCancelDataCollection filter stops us from re-processing an account if the sub has already been cancelled.

However, this means that if we fail to disable autopay (i.e. we get as far as making the request, but it does not succeed) during the first attempt, we will never complete this step as part of a retry. 

Swapping the order of the requests in the for comprehension means that we don't attempt to cancel the sub until the other two requests (which are idempotent*) have succeeded.

_*I'm pretty sure this is true_


johnduffell:
is that actually better? we check auto pay in the filter too, so if we turn off auto pay first and then it fails to cancel, they'll just get into the free stuff for life bucket.

jacobwinch:
That's a good point, I'd forgotten about that part of the filtering logic. So no, this is probably worse.

johnduffell:
fair enough - I think maybe just set system autocancel first (as we do) and then if we get called on a sub in that state, just skip the checks and try to do everything (again)

PR #33, opened 2017-12-20
update ET email client local test to send all emails for all products

johnduffell:
reliased I didn't push these changes.

Basically before the unique key on the emails was all payment id.  This was changed to be invoice id for the overdue email, because there's no actual payment for an overdue invoice.  There may be several.
The code was duplicated which was alright, but then I wanted to test over all the possible products, as ET will send different emails.  To avoid the duplication getting worse I just combined into one function and called that number of emails * number of products times, with the right parameters.

This is just a test you'd run locally, but it makes sense to have the latest version pushed.  It won't affect anythign that happens automatically.

@jacobwinch

### DIFF
--- a/src/test/scala/manualTest/EmailClientSystemTest.scala
+++ b/src/test/scala/manualTest/EmailClientSystemTest.scala
@@ -26,7 +26,7 @@ object EmailClientSystemTest extends App {
           payment_method = "paymentMethodValue",
           card_type = "cardTypeValue",
           card_expiry_date = "cardExpiryValue",
-          first_name = s"$hint-firstNameValue",
+          first_name = s"${hint.replaceAll(" ", "-")}-firstNameValue",
           last_name = "lastNameValue",
           primaryKey = key, // must be unique otherwise the email won't arrive
           price = "49.0 GBP",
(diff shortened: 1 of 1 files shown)


src/test/scala/manualTest/EmailClientSystemTest.scala
```diff
@@ -66,15 +55,24 @@ object EmailClientSystemTest extends App {
...
+  } yield Seq(
+    "Supporter",
+    "Digital Pack",
+    "Guardian Weekly Zone A",
+    "Guardian Weekly Zone B",
+    "Guardian Weekly Zone C",
+    "Contributor",
+    "Newspaper Voucher",
+    "Newspaper Delivery"
+  ).flatMap(xxx => five(a, xxx)).foreach {
```
jacobwinch: xxx?

src/test/scala/manualTest/EmailClientSystemTest.scala
```diff
@@ -66,15 +55,24 @@ object EmailClientSystemTest extends App {
     config <- Config.parseConfig(configAttempt)
     deps = EmailSendStepsDeps.default(Stage("CODE"), RawEffects.createDefault.response, config.etConfig)
     a = config.etConfig.etSendIDs
```
jacobwinch: What is 'a'?

src/test/scala/manualTest/EmailClientSystemTest.scala
```diff
@@ -37,27 +37,16 @@ object EmailClientSystemTest extends App {
...
-          card_expiry_date = "cardExpiryValue",
-          first_name = s"firstNameValue overdue",
-          last_name = "lastNameValue",
-          primaryKey = InvoiceId(s"invoiceId$unique"), // must be unique otherwise the email won't arrive
-          price = "49.0 GBP",
-          serviceStartDate = "31 January 2016",
-          serviceEndDate = "31 January 2017"
-        )
-      )
+  def five(a: ETSendIds, product: String) = {
```
jacobwinch: I think this function could have a clearer name

jacobwinch:
Looks good - just a couple of comments about naming (as per Scala Exchange advice!)

johnduffell:
haha 🙈

jacobwinch:
@johnduffell - shall we merge this (👍  from me)?

johnduffell:
need to do the review comments first, sorry

johnduffell:
finally merged this one 😆 🚀

PR #34, opened 2018-01-09
stripe card updated trigger to zuora

johnduffell:
This is needed so that when stripe cards are updated automcatically, we still have the information cached in zuora to say what the expiry and last4 are.  This is useful for display to the customer and for outbound calling for expired cards.

See https://stripe.com/blog/smarter-saved-cards

We can do the follow up changes in a later PR for checking signatures, while the backfill is going on.  This should be fine because the api token will be checked regardless.

The url would have to be of the form 
/CODE/stripe-customer-source-updated?apiToken=SUPERSECRET&stripeAccount=GNM_Membership_AUS

comments please @paulbrown1982 @jacobwinch @lmath

### DIFF
--- a/cloudformation.yaml
+++ b/cloudformation.yaml
@@ -174,16 +174,16 @@ Resources:
         - ZuoraAutoCancelProxyResource
 
 
-    StripeCustomerUpdatedLambda:
+    StripeCustomerSourceUpdatedLambda:
         Type: AWS::Lambda::Function
         Properties:
             Description: Handles auto-cancellations for membership and subscriptions
             FunctionName:
-                !Sub stripe-customer-updated-${Stage}
+                !Sub stripe-customer-source-updated-${Stage}
             Code:
                 S3Bucket: zuora-auto-cancel-dist
                 S3Key: !Sub membership/${Stage}/zuora-auto-cancel/zuora-auto-cancel.jar
-            Handler: com.gu.stripeCustomerUpdated.Lambda::apply
+            Handler: com.gu.stripeCustomerSourceUpdated.Lambda::apply
             Environment:
                 Variables:
                   Stage: !Ref Stage
@@ -197,39 +197,39 @@ Resources:
         DependsOn:
         - ZuoraAutoCancelRole
 
-    StripeCustomerUpdatedAPIPermission:
+    StripeCustomerSourceUpdatedAPIPermission:
         Type: AWS::Lambda::Permission
         Properties:
             Action: lambda:invokeFunction
-            FunctionName: !Sub stripe-customer-updated-${Stage}
+            FunctionName: !Sub stripe-customer-source-updated-${Stage}
             Principal: apigateway.amazonaws.com
-        DependsOn: StripeCustomerUpdatedLambda
+        DependsOn: StripeCustomerSourceUpdatedLambda
 
-    StripeCustomerUpdatedProxyResource:
+    StripeCustomerSourceUpdatedProxyResource:
         Type: AWS::ApiGateway::Resource
         Properties:
             RestApiId: !Ref ZuoraAutoCancelAPI
             ParentId: !GetAtt [ZuoraAutoCancelAPI, RootResourceId]
-            PathPart: stripe-customer-updated
+            PathPart: stripe-customer-source-updated
         DependsOn: ZuoraAutoCancelAPI
 
-    StripeCustomerUpdatedMethod:
+    StripeCustomerSourceUpdatedMethod:
         Type: AWS::ApiGateway::Method
         Properties:
             AuthorizationType: NONE
             RestApiId: !Ref ZuoraAutoCancelAPI
-            ResourceId: !Ref StripeCustomerUpdatedProxyResource
+            ResourceId: !Ref StripeCustomerSourceUpdatedProxyResource
             HttpMethod: POST
             RequestParameters:
               method.request.header.stripe-signature: true
             Integration:
               Type: AWS_PROXY
               IntegrationHttpMethod: POST
-              Uri: !Sub arn:aws:apigateway:${AWS::Region}:lambda:path/2015-03-31/functions/${StripeCustomerUpdatedLambda.Arn}/invocations
+              Uri: !Sub arn:aws:apigateway:${AWS::Region}:lambda:path/2015-03-31/functions/${StripeCustomerSourceUpdatedLambda.Arn}/invocations
         DependsOn:
         - ZuoraAutoCancelAPI
-        - StripeCustomerUpdatedLambda
-        - StripeCustomerUpdatedProxyResource
+        - StripeCustomerSourceUpdatedLambda
+        - StripeCustomerSourceUpdatedProxyResource
 
 
     ZuoraAutoCancelAPI:
--- a/cloudformation.yaml
+++ b/cloudformation.yaml
@@ -221,7 +221,7 @@ Resources:
             ResourceId: !Ref StripeCustomerSourceUpdatedProxyResource
             HttpMethod: POST
             RequestParameters:
-              method.request.header.stripe-signature: true
+              method.request.querystring.apiToken: true
             Integration:
               Type: AWS_PROXY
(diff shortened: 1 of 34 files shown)


cloudformation.yaml
```diff
@@ -119,12 +119,6 @@ Resources:
             Principal: apigateway.amazonaws.com
         DependsOn: PaymentFailureLambda
 
-    ZuoraAutoCancelAPI:
```
johnduffell: I reordered some stuff in the CF so that each endpoint stuff are together, rather than keeping the same resources together. I think that made more sense as it meant I could change the stuff for the new thing collectively, but perhaps it should really be an imported cloudformation or something

src/main/scala/com/gu/autoCancel/AutoCancel.scala
```diff
@@ -2,9 +2,8 @@ package com.gu.autoCancel
 
 import com.gu.util.Logging
 import com.gu.util.reader.Types.FailableOp
-import com.gu.util.zuora.Zuora
```
johnduffell: I did organise imports on everything (sorry!) 💨

src/main/scala/com/gu/effects/AwsS3.scala
```diff
@@ -13,7 +13,7 @@ object ConfigLoad extends Logging {
 
   // if you are updating the config, it's hard to test it in advance of deployment
   // with this, you can upload the new config with a new name
-  val version = "2"
+  val version = "3"
```
johnduffell: @lmath  it's not clear from the comment in the code, but it's not essential to update the version unless you're making a breaking change.  Just adding something to the existing config shouldn't break anything

src/main/scala/com/gu/util/ConfigLoader.scala
```diff
@@ -57,17 +57,36 @@ object ETConfig {
...
+case class StripeSecretKey(key: String) extends AnyVal
+
+object StripeSecretKey {
+  implicit val stripeSecretKeyReads = Json.reads[StripeSecretKey]
+}
+
+case class StripeConfig(ukStripeSecretKey: StripeSecretKey, auStripeSecretKey: StripeSecretKey)
+object StripeConfig {
+  implicit val apiConfigReads: Reads[StripeConfig] = (
+    (JsPath \ "api.key.secret").read[String].map(StripeSecretKey.apply) and
```
paulbrown1982: could we rename this to `uk-membership.key.secret` ?

src/main/scala/com/gu/util/zuora/ZuoraRestRequestMaker.scala
```diff
@@ -71,11 +80,20 @@ object ZuoraRestRequestMaker extends Logging {
     Reader { zuoraDeps: ZuoraDeps =>
       val body = RequestBody.create(MediaType.parse("application/json"), Json.toJson(req).toString)
       val request = buildRequest(zuoraDeps.config)(path).put(body).build()
-      logger.info(s"Attempting to $path with the following command: $req")
+      logger.info(s"Attempting to PUT $path with the following command: $req")
```
paulbrown1982: is $req secure?

src/test/scala/com/gu/autoCancel/AutoCancelHandlerTest.scala
```diff
@@ -64,14 +64,14 @@ class AutoCancelHandlerTest extends FlatSpec {
   }
 
   "authenticateCallout" should "return a left if the credentials are invalid" in {
-    val requestAuth = RequestAuth(apiClientId = "correctId", apiToken = "token")
-    val trustedApiConfig = TrustedApiConfig(apiClientId = "wrongId", apiToken = "token", tenantId = "tenant")
+    val requestAuth = RequestAuth(apiToken = "tokpen")
```
paulbrown1982: Maybe change `tokpen` to `foo` just so the next code reader doesn't double-take like I just did!

paulbrown1982:
This is looking good @johnduffell @lmath. Let me know when you want a final review.

src/main/scala/com/gu/effects/AwsS3.scala
lmath: Oh, I didn't realise! That makes sense though.

src/main/scala/com/gu/util/ConfigLoader.scala
```diff
@@ -57,17 +57,36 @@ object ETConfig {
...
+    (JsPath \ "au-membership.key.secret").read[String].map(StripeSecretKey.apply)
```
johnduffell: since we have a stripeSecretKeyReads we can just do read[StripeSecretKey] rather than read[String] and mapping it, just saves a tiny bit of duplication here

lmath: Ah yes, good point. I'll fix that.

src/main/scala/com/gu/util/zuora/Zuora.scala
```diff
@@ -28,5 +36,134 @@ object Zuora {
   def disableAutoPay(accountId: String): WithDepsFailableOp[ZuoraDeps, Unit] =
     put(AccountUpdate(autoPay = false), s"accounts/$accountId")
 
+  case class Query(queryString: String)
```
johnduffell: I don't think all this should be in this file as they are not all related (other than all happen to be zuora calls), I'm going to separate out the query stuff and possible separate files

src/test/scala/com/gu/autoCancel/AutoCancelHandlerTest.scala
johnduffell: good idea, better still
```scala
val requestAuth = RequestAuth(apiToken = "incorrectRequestToken")
```

src/main/scala/com/gu/util/zuora/ZuoraRestRequestMaker.scala
johnduffell: might do in terms of customer tokens, but the logs shouldn't be retained too long

src/main/scala/com/gu/util/apigateway/ApiGatewayRequest.scala
```diff
@@ -1,25 +1,42 @@
...
+  implicit val reads: Reads[StripeAccount] = JsPath.read[String].map(fromString(_).get)
+}
+case class URLParams(apiToken: Option[String], onlyCancelDirectDebit: Boolean, stripeAccount: Option[StripeAccount])
 
 /* Using query strings because for Basic Auth to work Zuora requires us to return a WWW-Authenticate
   header, and API Gateway does not support this header (returns x-amzn-Remapped-WWW-Authenticate instead)
   */
-case class ApiGatewayRequest(queryStringParameters: Option[URLParams], body: String) {
-  def onlyCancelDirectDebit: Boolean = queryStringParameters.exists(_.onlyCancelDirectDebit.contains("true"))
+case class ApiGatewayRequest(queryStringParameters: Option[URLParams], body: String, headers: Option[Map[String, String]]) {
```
johnduffell: in the URLParams we have modelled the actual parameters we need rather than having to extract them later.  That means the processing code can use named methods rather than picking strings from a map.
Admittedly it's shared between all the readers, but I would still recommend doing the work in the readers as much as possible so we can keep the decision logic simpler.

johnduffell:
So I think we should get this reviewed and deployed (asap this afternoon), then we can add the stripe hook and monitor it.  Then we can leave it on or off overnight depending how we feel.
Then we can do the final tweaks as a further PR while we're doing the backfill hopefully later this week.  Any thoughts @paulbrown1982 @lmath ?

src/main/scala/com/gu/util/ConfigLoader.scala
paulbrown1982: Note the naming convention is the same here as with membership-common: https://github.com/guardian/membership-common/blob/b9ed285395d9cf240ff26e91629dfee4654830bc/conf/touchpoint.PROD.conf#L18

PR #35, opened 2018-01-10
Trigger exp month over to zuora when the card is updated

johnduffell:
Copy and paste error, combined with testing that the code returns what it did when I wrote the test let this one slip though!

@jacobwinch @paulbrown1982 @lmath


paulbrown1982:
🤦‍♂️ 👍

PR #36, opened 2018-01-11
Verify Stripe-Signature

lmath:
In the Stripe Customer Source Updated lambda, we receive events customer.source.updated events from Stripe. Stripe [signs](https://stripe.com/docs/webhooks#signatures) these requests. Here, we check the signature and verify that the requests are coming from Stripe.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -45,7 +45,8 @@ libraryDependencies ++= Seq(
   "org.scalatest" %% "scalatest" % "3.0.1" % "test",
   "org.mockito" % "mockito-core" % "1.9.5" % "test",
   "com.github.nscala-time" % "nscala-time_2.12" % "2.16.0",
-  "org.apache.commons" % "commons-io" % "1.3.2"
+  "org.apache.commons" % "commons-io" % "1.3.2",
+  "com.stripe" % "stripe-java" % "5.28.0"
 )
 
 initialize := {
--- a/src/main/scala/com/gu/stripeCustomerSourceUpdated/SourceUpdatedCallout.scala
+++ b/src/main/scala/com/gu/stripeCustomerSourceUpdated/SourceUpdatedCallout.scala
@@ -1,7 +1,10 @@
 package com.gu.stripeCustomerSourceUpdated
 
+import com.gu.util.StripeConfig
 import play.api.libs.functional.syntax._
 import play.api.libs.json._
+
+case class StripeDeps(config: StripeConfig)
 
 case class EventData(`object`: EventDataObject)
 case class EventDataObject(
--- a/src/main/scala/com/gu/stripeCustomerSourceUpdated/SourceUpdatedCallout.scala
+++ b/src/main/scala/com/gu/stripeCustomerSourceUpdated/SourceUpdatedCallout.scala
@@ -4,8 +4,6 @@ import com.gu.util.StripeConfig
 import play.api.libs.functional.syntax._
 import play.api.libs.json._
 
-case class StripeDeps(config: StripeConfig)
-
 case class EventData(`object`: EventDataObject)
 case class EventDataObject(
   id: StripeSourceId,
--- a/src/main/scala/com/gu/stripeCustomerSourceUpdated/SourceUpdatedSteps.scala
+++ b/src/main/scala/com/gu/stripeCustomerSourceUpdated/SourceUpdatedSteps.scala
@@ -1,6 +1,8 @@
 package com.gu.stripeCustomerSourceUpdated
 
+import com.gu.stripeCustomerSourceUpdated.StripeSignatureChecker._
 import com.gu.util._
+import com.gu.util.apigateway.ApiGatewayResponse.unauthorized
 import com.gu.util.apigateway.{ ApiGatewayRequest, ApiGatewayResponse }
 import com.gu.util.reader.Types._
 import com.gu.util.zuora.CreatePaymentMethod.{ CreateStripePaymentMethod, CreditCardType }
@@ -16,14 +18,26 @@ import scalaz.syntax.std.option._
 object SourceUpdatedSteps extends Logging {
 
   def apply(deps: Deps)(apiGatewayRequest: ApiGatewayRequest): FailableOp[Unit] = {
-    (for {
-      sourceUpdatedCallout <- Json.fromJson[SourceUpdatedCallout](Json.parse(apiGatewayRequest.body)).toFailableOp.withLogging("fromJson SourceUpdatedCallout").pure[WithDeps].toEitherT
-      _ = logger.info(s"from: ${apiGatewayRequest.queryStringParameters.map(_.stripeAccount)}")
-      accountId <- getAccountToUpdate(sourceUpdatedCallout.data.`object`.customer, sourceUpdatedCallout.data.`object`.id)
-      _ <- updatePaymentMethod(accountId, sourceUpdatedCallout.data.`object`)
-    } yield ()).run.run(deps.zuoraDeps)
+    (
+
+      for {
+        body <- bodyIfSignatureVerified(deps, apiGatewayRequest).pure[WithDeps].toEitherT
+        sourceUpdatedCallout <- Json.fromJson[SourceUpdatedCallout](Json.parse(body)).toFailableOp.withLogging("fromJson SourceUpdatedCallout").pure[WithDeps].toEitherT
+        _ = logger.info(s"from: ${apiGatewayRequest.queryStringParameters.map(_.stripeAccount)}")
+        accountId <- getAccountToUpdate(sourceUpdatedCallout.data.`object`.customer, sourceUpdatedCallout.data.`object`.id)
+        _ <- updatePaymentMethod(accountId, sourceUpdatedCallout.data.`object`)
+      } yield ()
+    ).run.run(deps.zuoraDeps)
+
   }
 
+  def bodyIfSignatureVerified(deps: Deps, apiGatewayRequest: ApiGatewayRequest): FailableOp[String] = {
+    val signatureVerified = verifyStripeSignature(stripeConfig = deps.stripeDeps.config, apiGatewayRequest.headers.get, apiGatewayRequest.body)
+    signatureVerified match {
+      case true => \/-(apiGatewayRequest.body)
+      case false => -\/(unauthorized)
+    }
+  }
   def updatePaymentMethod(accountId: AccountId, eventDataObject: EventDataObject): WithDepsFailableOp[ZuoraDeps, Unit] = {
     for {
       // (clear default payment method and autopay off not needed because we are not changing the payment gateway)
@@ -85,11 +99,11 @@ object SourceUpdatedSteps extends Logging {
 
   object Deps {
     def default(response: Request => Response, config: Config): Deps = {
-      Deps(ZuoraDeps(response, config.zuoraRestConfig))
+      Deps(ZuoraDeps(response, config.zuoraRestConfig), StripeDeps(config.stripeConfig))
     }
   }
 
-  case class Deps(zuoraDeps: ZuoraDeps)
+  case class Deps(zuoraDeps: ZuoraDeps, stripeDeps: StripeDeps)
 
 }
 
--- a/src/main/scala/com/gu/stripeCustomerSourceUpdated/SourceUpdatedSteps.scala
+++ b/src/main/scala/com/gu/stripeCustomerSourceUpdated/SourceUpdatedSteps.scala
@@ -1,9 +1,8 @@
 package com.gu.stripeCustomerSourceUpdated
 
-import com.gu.stripeCustomerSourceUpdated.StripeSignatureChecker._
 import com.gu.util._
 import com.gu.util.apigateway.ApiGatewayResponse.unauthorized
-import com.gu.util.apigateway.{ ApiGatewayRequest, ApiGatewayResponse }
+import com.gu.util.apigateway.{ ApiGatewayRequest, ApiGatewayResponse, StripeAccount }
 import com.gu.util.reader.Types._
 import com.gu.util.zuora.CreatePaymentMethod.{ CreateStripePaymentMethod, CreditCardType }
 import com.gu.util.zuora.ZuoraQueryPaymentMethod.{ AccountId, PaymentMethodId }
@@ -14,14 +13,14 @@ import play.api.libs.json.Json
 import scalaz._
 import scalaz.syntax.applicative._
 import scalaz.syntax.std.option._
+import com.gu.stripeCustomerSourceUpdated.StripeRequestSignatureChecker._
 
 object SourceUpdatedSteps extends Logging {
 
   def apply(deps: Deps)(apiGatewayRequest: ApiGatewayRequest): FailableOp[Unit] = {
     (
-
       for {
-        body <- bodyIfSignatureVerified(deps, apiGatewayRequest).pure[WithDeps].toEitherT
+        body <- bodyIfSignatureVerified(deps.stripeDeps, apiGatewayRequest).pure[WithDeps].toEitherT
         sourceUpdatedCallout <- Json.fromJson[SourceUpdatedCallout](Json.parse(body)).toFailableOp.withLogging("fromJson SourceUpdatedCallout").pure[WithDeps].toEitherT
         _ = logger.info(s"from: ${apiGatewayRequest.queryStringParameters.map(_.stripeAccount)}")
         accountId <- getAccountToUpdate(sourceUpdatedCallout.data.`object`.customer, sourceUpdatedCallout.data.`object`.id)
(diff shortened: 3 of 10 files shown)


src/main/scala/com/gu/stripeCustomerSourceUpdated/StripeSignatureChecker.scala
```diff
@@ -0,0 +1,31 @@
...
+import scala.util.Try
+
+object StripeSignatureChecker {
+
+  def verifyStripeSignature(stripeConfig: StripeConfig, headers: Map[String, String], payload: String) =
+    verifyStripeSignatureForAccount(stripeConfig.ukStripeSecretKey.key, headers, payload) || verifyStripeSignatureForAccount(stripeConfig.auStripeSecretKey.key, headers, payload)
+
+  private def verifyStripeSignatureForAccount(secretKey: String, headers: Map[String, String], payload: String) = {
+    val signatureHeader: Option[String] = headers.get("Stripe-Signature")
+    val headerVerified: Try[Boolean] = Try(Signature.verifyHeader(payload, signatureHeader.get, secretKey, 1000l))
```
lmath: 1000l is saying I'm willing for the timestamp to be off by up to 1000ms but I'm not sure what a good value is for this or how configurable it needs to be

paulbrown1982: I think about 5-10 seconds is better than 1 second.

paulbrown1982: Also this is the Stripe API reference: https://static.javadoc.io/com.stripe/stripe-java/5.28.0/com/stripe/net/Webhook.Signature.html

build.sbt
```diff
@@ -45,7 +45,8 @@ libraryDependencies ++= Seq(
   "org.scalatest" %% "scalatest" % "3.0.1" % "test",
   "org.mockito" % "mockito-core" % "1.9.5" % "test",
   "com.github.nscala-time" % "nscala-time_2.12" % "2.16.0",
-  "org.apache.commons" % "commons-io" % "1.3.2"
```
johnduffell: the rest of the dependency list looks worrying

src/main/scala/com/gu/stripeCustomerSourceUpdated/SourceUpdatedSteps.scala
```diff
@@ -16,14 +18,26 @@ import scalaz.syntax.std.option._
...
+    ).run.run(deps.zuoraDeps)
+
   }
 
+  def bodyIfSignatureVerified(deps: Deps, apiGatewayRequest: ApiGatewayRequest): FailableOp[String] = {
+    val signatureVerified = verifyStripeSignature(stripeConfig = deps.stripeDeps.config, apiGatewayRequest.headers.get, apiGatewayRequest.body)
+    signatureVerified match {
+      case true => \/-(apiGatewayRequest.body)
+      case false => -\/(unauthorized)
+    }
```
johnduffell: pattern match on a boolean is more concise as an if statement
```scala
if (signatureVerified)
  \/-(apiGatewayRequest.body)
else 
  -\/(unauthorized)
```
I'm pretty sure there's some special scalaz thing, but I'm not going to work out what it is to recommend it!

src/main/scala/com/gu/stripeCustomerSourceUpdated/SourceUpdatedCallout.scala
```diff
@@ -1,8 +1,11 @@
 package com.gu.stripeCustomerSourceUpdated
 
+import com.gu.util.StripeConfig
 import play.api.libs.functional.syntax._
 import play.api.libs.json._
 
+case class StripeDeps(config: StripeConfig)
```
johnduffell: this should be in the file that requires the dependencies, so StripeSignatureChecker.scala in this case.  I'd be happy to do without it since there's only one value so far, but keeping it is also fine for forward extensibility.

src/main/scala/com/gu/stripeCustomerSourceUpdated/StripeSignatureChecker.scala
```diff
@@ -0,0 +1,31 @@
...
+      case e: SignatureVerificationException => {
+        logger.error(s"something went wrong with sig header: ${e.getSigHeader} D:< with message ${e.getMessage}")
+        false
+      }
+      case e: Throwable => {
+        logger.error(s"something went wrong with verifying the signature header D:< with message ${e.getMessage}")
+        false
+      }
+    }
+    falseIfNotSigned.get //because false if any throwable
```
johnduffell: so rather than separating the recovery code from the get, this means the compiler can't guarantee that everything was handled (and into the future)
Better to pattern match the Try to see if it's Failure or Success, and return false or true from the pattern match.  Then the compiler will complain if any gaps exist.

src/main/scala/com/gu/stripeCustomerSourceUpdated/StripeSignatureChecker.scala
paulbrown1982: See if this Signature.verifyHeader API function can be mocked or dependency injected so that you can test the Lambda without having to generate a real signature that Signature.verifyHeader can verify. If you can't mock it then I think you should put a comment in the code of the Stripe reference material of how to generate a signature.

src/test/scala/com/gu/stripeCustomerSourceUpdated/StripeSignatureCheckerTest.scala
```diff
@@ -0,0 +1,96 @@
...
+
+    val hasher = Mac.getInstance("HmacSHA256")
+    hasher.init(new SecretKeySpec(secretKey.getBytes("UTF8"), "HmacSHA256"))
+    val hash: Array[Byte] = hasher.doFinal(signedPayload.getBytes("UTF8"))
+
+    var result = ""
+    for (b <- hash) {
+      result += Integer.toString((b & 0xff) + 0x100, 16).substring(1)
+    }
+    result
```
paulbrown1982: Ideally you shouldn't have to generate a real signature to test your Lambda, you should be able to inject a verifier that accepts "foo" as the signature. See my comment above. If you can't inject/mock the verifier then I think you should link to the doc which tells you how to generate a signature.

lmath: Agreed. I'm on it!

src/main/scala/com/gu/stripeCustomerSourceUpdated/StripeSignatureChecker.scala
```diff
@@ -0,0 +1,31 @@
...
+import com.gu.util.StripeConfig
+import com.stripe.exception.SignatureVerificationException
+import com.stripe.net.Webhook.Signature
+
+import scala.util.Try
+
+object StripeSignatureChecker {
+
+  def verifyStripeSignature(stripeConfig: StripeConfig, headers: Map[String, String], payload: String) =
+    verifyStripeSignatureForAccount(stripeConfig.ukStripeSecretKey.key, headers, payload) || verifyStripeSignatureForAccount(stripeConfig.auStripeSecretKey.key, headers, payload)
```
johnduffell: for more determinism, we do know which one passed the call in, it's in the callout object as apiGatewayRequest.queryStringParameters.stripeAccount

lmath: Oh nice! I'll make use of that then.

src/test/scala/com/gu/TestData.scala
```diff
@@ -102,7 +104,7 @@ class TestingRawEffects(val isProd: Boolean = false, val defaultCode: Int = 1, r
 
   val stage = Stage(if (isProd) "PROD" else "DEV")
 
-  def basicResults = result.map { request =>
+  def requestsAttempted = result.map { request =>
```
johnduffell: good rename - definitely an improvement

src/test/scala/com/gu/stripeCustomerSourceUpdated/SourceUpdatedStepsTest.scala
```diff
@@ -40,25 +38,24 @@ class SourceUpdatedStepsTest extends FlatSpec with Matchers {
...
-                                |  "records": [
-                                |    {
-                                |      "Id": "defaultPMID",
-                                |      "AccountId": "accid"
-                                |    }
-                                |  ],
-                                |  "size": 1,
-                                |  "done": true
-                                |}""".stripMargin)), //defaultPMID
+                                  |  "records": [
```
johnduffell: my current PR is editing these tests, so it may be worth not doing all these whitespace only changes

src/main/scala/com/gu/stripeCustomerSourceUpdated/SourceUpdatedSteps.scala
```diff
@@ -16,14 +18,26 @@ import scalaz.syntax.std.option._
...
+        _ = logger.info(s"from: ${apiGatewayRequest.queryStringParameters.map(_.stripeAccount)}")
+        accountId <- getAccountToUpdate(sourceUpdatedCallout.data.`object`.customer, sourceUpdatedCallout.data.`object`.id)
+        _ <- updatePaymentMethod(accountId, sourceUpdatedCallout.data.`object`)
+      } yield ()
+    ).run.run(deps.zuoraDeps)
+
   }
 
+  def bodyIfSignatureVerified(deps: Deps, apiGatewayRequest: ApiGatewayRequest): FailableOp[String] = {
+    val signatureVerified = verifyStripeSignature(stripeConfig = deps.stripeDeps.config, apiGatewayRequest.headers.get, apiGatewayRequest.body)
```
johnduffell: this will throw a None.get if there are no headers (perhaps better to change headers to be non optional, and default it to an empty Map in the reader)

src/test/scala/com/gu/stripeCustomerSourceUpdated/StripeSignatureCheckerTest.scala
```diff
@@ -0,0 +1,96 @@
...
+      |    "id": null,
+      |    "idempotency_key": null
+      |  },
+      |  "type": "customer.source.updated"
+      |}
+    """.stripMargin
+
+  def headersWithStripeSignature(timestamp: String, signature: String) = Map(
+    "SomeHeader1" -> "testvalue",
+    "Content-Type" -> "application/json",
```
johnduffell: it looks like we only need Stripe-Signature header.  So we should only pass that in to the code otherwise the tests have unnecessary values needed.
Would be better to follow the same convention as we do in the URLParams class, where the reader understand what fields are what and then wraps in a case class.

src/test/scala/com/gu/stripeCustomerSourceUpdated/SourceUpdatedStepsTest.scala
lmath: Oops I've actually been trying to find where intellij is automagically doing this and disable it

src/main/scala/com/gu/stripeCustomerSourceUpdated/SourceUpdatedSteps.scala
lmath: Good shout. I'm on it.

src/test/scala/com/gu/stripeCustomerSourceUpdated/StripeSignatureCheckerTest.scala
lmath: Yes, good idea. I shall take your suggestion.

src/main/scala/com/gu/stripeCustomerSourceUpdated/StripeSignatureChecker.scala
```diff
@@ -1,31 +1,50 @@
...
-    val falseIfNotSigned = headerVerified recover {
-      case e: SignatureVerificationException => {
-        logger.error(s"something went wrong with sig header: ${e.getSigHeader} D:< with message ${e.getMessage}")
-        false
-      }
-      case e: Throwable => {
-        logger.error(s"something went wrong with verifying the signature header D:< with message ${e.getMessage}")
-        false
+    stripeAccount match {
+      case Some(account) => {
```
johnduffell: not a big difference, but for future reference you can do stripeAccount.exists { account => <line 19-32 here> } rather than pattern match and return false

src/main/scala/com/gu/stripeCustomerSourceUpdated/StripeSignatureChecker.scala
```diff
@@ -1,31 +1,50 @@
...
+        val headerVerified: Try[Boolean] = Try(stripeDeps.signatureChecker.verifySignature(secretKey, payload, signatureHeader, 10000l))
+
+        headerVerified match {
+          case Success(verified) => verified
+          case Failure(e: SignatureVerificationException) => {
+            logger.error(s"Signature header was not validated ${e.getSigHeader} with message ${e.getMessage}")
+            false
+          }
+          case Failure(e: Throwable) => {
+            logger.error(s"something went wrong with verifying the signature header with message ${e.getMessage}")
```
johnduffell: e.toString would be a bit more useful as it says the actual exception.  But actually you should probably just call the two parameter version of .error that takes the whole throwable as the second param.

src/main/scala/com/gu/stripeCustomerSourceUpdated/StripeSignatureChecker.scala
```diff
@@ -1,31 +1,50 @@
...
+
+trait SignatureChecker {
+  def verifySignature(secretKey: StripeSecretKey, payload: String, signatureHeader: Option[String], tolerance: Long): Boolean
+}
+
+class StripeSignatureChecker extends SignatureChecker {
+  def verifySignature(secretKey: StripeSecretKey, payload: String, signatureHeader: Option[String], tolerance: Long): Boolean = {
+    signatureHeader match {
+      case Some(header) => Signature.verifyHeader(payload, header, secretKey.key, tolerance)
+      case None => false
```
johnduffell: you could have used .exists here as well to neaten it up

src/main/scala/com/gu/stripeCustomerSourceUpdated/SourceUpdatedSteps.scala
```diff
@@ -17,13 +19,24 @@ object SourceUpdatedSteps extends Logging {
 
   def apply(deps: Deps)(apiGatewayRequest: ApiGatewayRequest): FailableOp[Unit] = {
     (for {
-      sourceUpdatedCallout <- Json.fromJson[SourceUpdatedCallout](Json.parse(apiGatewayRequest.body)).toFailableOp.withLogging("fromJson SourceUpdatedCallout").pure[WithDeps].toEitherT
+      body <- bodyIfSignatureVerified(deps.stripeDeps, apiGatewayRequest).pure[WithDeps].toEitherT
```
johnduffell: we don't need to use this to get the body, just return Unit if it's valid, then below we can still pick out the body if we need it

johnduffell:
great, I'm down to stylistic nitpicks now, thanks for going through all that and looks great 👍

lmath:
I realised I was using the wrong secret to check the signature and actually there is a signing signature for each webhook, for each stripe account: https://stripe.com/docs/webhooks#signatures.

@paulbrown1982 @johnduffell Would one of you mind taking another quick look since I've made these config changes? I'm planning on having config like

```
"stripe": {
    "customerSourceUpdatedWebhook": {
      "api.key.secret": "...",
      "au-membership.key.secret": "..."
    },
    "api.key.secret": "...",
    "au-membership.key.secret": "..."
  }
```
and then removing both API secret key (stripe.{api or au-membership}.key.secret) for each Stripe account after I merge.

src/test/scala/com/gu/TestData.scala
```diff
@@ -37,14 +37,14 @@ object TestData extends Matchers {
   val fakeZuoraConfig = ZuoraRestConfig("https://ddd", "fakeUser", "fakePass")
   val fakeETSendIds = ETSendIds(ETSendId("11"), ETSendId("22"), ETSendId("33"), ETSendId("44"), ETSendId("can"))
   val fakeETConfig = ETConfig(etSendIDs = fakeETSendIds, "fakeClientId", "fakeClientSecret")
-  val fakeStripeConfig = StripeConfig(StripeSecretKey("ukStripeSecretKey"), StripeSecretKey("auStripeSecretKey"))
+  val fakeStripeConfig = StripeConfig(customerSourceUpdatedWebhook = StripeWebhook(StripeSecretKey("ukcustomerSourceUpdatedSecretKey"), StripeSecretKey("aucustomerSourceUpdatedStripeSecretKey")))
```
jacobwinch: Minor point, but 'ukCustomerSourceUpdatedSecretKey' and 'auCustomerSourceUpdatedStripeSecretKey' is probably what we want here... unless we can think of a more succinct name.

jacobwinch: Sorry for being so pedantic, I was skim-reading and I didn't realise that these were only test strings!

build.sbt
johnduffell: see https://github.com/guardian/zuora-auto-cancel/pull/39

johnduffell:
thanks for getting this shipped!! 🎉 👍

PR #37, opened 2018-01-11
dont take payment for outstanding invoices when stripe expiry update comes in

johnduffell:
So I realised that by creating a fresh payment method with the same stripe ids, this would reset the payment failure count for the default payment method back to zero.
This could cause zuora to resume trying to charge any outstanding invoices where auto pay is left on.

Since we are not sure how many people this would affect, I have changed it to copy the payment failure count to the new payment method object in zuora.  This should prevent the user being charged too many times, although it would bring forward the next retry to the next day, rather than waiting a week between payment attempts.  This may actually not be a bad idea.

Thanks for the help @jacobwinch 
cc @paulbrown1982 @lmath

### DIFF
--- a/src/main/scala/com/gu/stripeCustomerSourceUpdated/SourceUpdatedSteps.scala
+++ b/src/main/scala/com/gu/stripeCustomerSourceUpdated/SourceUpdatedSteps.scala
@@ -4,7 +4,7 @@ import com.gu.util._
 import com.gu.util.apigateway.{ ApiGatewayRequest, ApiGatewayResponse }
 import com.gu.util.reader.Types._
 import com.gu.util.zuora.CreatePaymentMethod.{ CreateStripePaymentMethod, CreditCardType }
-import com.gu.util.zuora.ZuoraQueryPaymentMethod.{ AccountId, PaymentMethodId }
+import com.gu.util.zuora.ZuoraQueryPaymentMethod.{ PaymentMethodFields, PaymentMethodId }
 import com.gu.util.zuora._
 import okhttp3.{ Request, Response }
 import play.api.libs.json.Json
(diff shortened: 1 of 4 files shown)


src/main/scala/com/gu/stripeCustomerSourceUpdated/SourceUpdatedSteps.scala
```diff
@@ -19,41 +19,41 @@ object SourceUpdatedSteps extends Logging {
...
-      paymentMethod <- createPaymentMethod(eventDataObject, accountId).withLogging("createPaymentMethod")
+      paymentMethod <- createPaymentMethod(eventDataObject, paymentMethodFields).withLogging("createPaymentMethod")
       // set payment method as the default - update account defaultpaymentmethodid
-      _ <- SetDefaultPaymentMethod.setDefaultPaymentMethod(accountId, paymentMethod.id).withLogging("setDefaultPaymentMethod")
+      _ <- SetDefaultPaymentMethod.setDefaultPaymentMethod(paymentMethodFields.AccountId, paymentMethod.id).withLogging("setDefaultPaymentMethod")
     } yield ()
   }
 
-  def getAccountToUpdate(customer: StripeCustomerId, source: StripeSourceId): WithDepsFailableOp[ZuoraDeps, AccountId] = {
+  def getAccountToUpdate(customer: StripeCustomerId, source: StripeSourceId): WithDepsFailableOp[ZuoraDeps, PaymentMethodFields] = {
```
paulbrown1982: The name of this needs to be different, it's not just the accountId anymore.

johnduffell:
ready for a +1 when anyone wants to have a poke around! 🚀

paulbrown1982:
👍 LGTM

PR #38, opened 2018-01-17
allow stripe card/cust token to update 0-3 accounts in zuora

johnduffell:
So the trigger of card details to zuora worked brilliantly, apart from a small number of situations where either no details in zuora that matched, or multiple.
I had actually put in a safety check to make sure we weren't either not finding any of the accounts, or were finding hundreds and populating them all.  This was causing a 500 error on purpose so we knew about it.
The backfill could cope with that as it didn't have the safety check, so I have broadened the range of numbers it would do in the trigger from exactly one, to 0-3.

Hopefully that should remove the errors, and I can turn the threshold on the cloudwatch alarm back down (I did a 10x increase before)

@lmath @paulbrown1982 @jacobwinch @AWare see what you all think

edit: I think we now have some for comprehensions worthy of membership-common, however I am happy to adjust them to be more readable if someone would like to help me out

### DIFF
--- a/src/main/scala/com/gu/stripeCustomerSourceUpdated/SourceUpdatedSteps.scala
+++ b/src/main/scala/com/gu/stripeCustomerSourceUpdated/SourceUpdatedSteps.scala
@@ -28,8 +28,8 @@ object SourceUpdatedSteps extends Logging {
       _ = logger.info(s"from: ${apiGatewayRequest.queryStringParameters.map(_.stripeAccount)}")
       _ <- (for {
         defaultPaymentMethod <- ListT(getPaymentMethodsToUpdate(sourceUpdatedCallout.data.`object`.customer, sourceUpdatedCallout.data.`object`.id))
-        _ <- ListT.apply[WithZuoraDepsFailableOp, Unit](createUpdatedDefaultPaymentMethod(defaultPaymentMethod, sourceUpdatedCallout.data.`object`).map(_.pure[List]))
-      } yield ()).run.map({ a: List[Unit] /*prove we're mapping the functor containing the list of units*/ => () })
+        _ <- ListT[WithZuoraDepsFailableOp, Unit](createUpdatedDefaultPaymentMethod(defaultPaymentMethod, sourceUpdatedCallout.data.`object`).map(_.pure[List]))
+      } yield ()).run.map({ list: List[Unit] /*prove we're mapping the functor containing the list of units*/ => () })
     } yield ()).run.run(deps.zuoraDeps)
   }
 
--- a/src/main/scala/com/gu/stripeCustomerSourceUpdated/SourceUpdatedSteps.scala
+++ b/src/main/scala/com/gu/stripeCustomerSourceUpdated/SourceUpdatedSteps.scala
@@ -1,10 +1,12 @@
 package com.gu.stripeCustomerSourceUpdated
 
 import com.gu.util._
+import com.gu.util.apigateway.ResponseModels.ApiResponse
 import com.gu.util.apigateway.{ ApiGatewayRequest, ApiGatewayResponse }
 import com.gu.util.reader.Types._
 import com.gu.util.zuora.CreatePaymentMethod.{ CreateStripePaymentMethod, CreditCardType }
-import com.gu.util.zuora.ZuoraQueryPaymentMethod.{ PaymentMethodFields, PaymentMethodId }
+import com.gu.util.zuora.ZuoraGetAccountSummary.AccountSummary
+import com.gu.util.zuora.ZuoraQueryPaymentMethod.{ AccountPaymentMethodIds, PaymentMethodFields, PaymentMethodId }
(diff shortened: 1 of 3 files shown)


src/main/scala/com/gu/stripeCustomerSourceUpdated/SourceUpdatedSteps.scala
```diff
@@ -1,51 +1,60 @@
...
 import scalaz._
 import scalaz.syntax.applicative._
 import scalaz.syntax.std.option._
+import scalaz.std.list._
+import scalaz.EitherT._
 
 object SourceUpdatedSteps extends Logging {
 
+  type WithZuoraDepsFailableOp[A] = WithDepsFailableOp[ZuoraDeps, A]
+  implicit val mWithZuoraDepsFailableOp: Monad[WithZuoraDepsFailableOp] = eitherTMonad[({ type XReader[AA] = Reader[ZuoraDeps, AA] })#XReader, ApiResponse]
```
johnduffell: seems the compiler can't make proofs using type constructors that take more parameters than the implicit itself, even if there is a type alias in scope.  So I had to call eitherTMonad myself and reproduce the type alias inside the type parameter.
Not sure what better to do - otherwise we can't flatMap over the WithZuoraDepsFailableOp

src/main/scala/com/gu/stripeCustomerSourceUpdated/SourceUpdatedSteps.scala
```diff
@@ -1,51 +1,60 @@
...
   def apply(deps: Deps)(apiGatewayRequest: ApiGatewayRequest): FailableOp[Unit] = {
     (for {
       sourceUpdatedCallout <- Json.fromJson[SourceUpdatedCallout](Json.parse(apiGatewayRequest.body)).toFailableOp.withLogging("fromJson SourceUpdatedCallout").pure[WithDeps].toEitherT
       _ = logger.info(s"from: ${apiGatewayRequest.queryStringParameters.map(_.stripeAccount)}")
-      defaultPaymentMethod <- getPaymentMethodToUpdate(sourceUpdatedCallout.data.`object`.customer, sourceUpdatedCallout.data.`object`.id)
-      _ <- createUpdatedDefaultPaymentMethod(defaultPaymentMethod, sourceUpdatedCallout.data.`object`)
+      _ <- (for {
+        defaultPaymentMethod <- ListT(getPaymentMethodsToUpdate(sourceUpdatedCallout.data.`object`.customer, sourceUpdatedCallout.data.`object`.id))
+        _ <- ListT[WithZuoraDepsFailableOp, Unit](createUpdatedDefaultPaymentMethod(defaultPaymentMethod, sourceUpdatedCallout.data.`object`).map(_.pure[List]))
+      } yield ()).run.map({ list: List[Unit] /*prove we're mapping the functor containing the list of units*/ => () })
```
johnduffell: I want to do all the listy stuff and then fold it back to a single unit to carry on with the overall process.
I am not quite sure why I wanted to check that the nested for comprehension returned a unit, I think it might be useful in case someone wants to aggregate the result of createUpdatedDefaultPaymentMethod?
Maybe better to just take that out, but I just feel a bit funny returning a List[Unit] into an _

src/main/scala/com/gu/stripeCustomerSourceUpdated/SourceUpdatedSteps.scala
paulbrown1982: 😕 I think I understand now!

src/main/scala/com/gu/stripeCustomerSourceUpdated/SourceUpdatedSteps.scala
paulbrown1982: I agree it's probably more readable, if not more reliable, to remove it.

paulbrown1982:
👍 Looks great. It's a nice, pragmatic solution to the update-of-multiple-methods problem.

src/main/scala/com/gu/stripeCustomerSourceUpdated/SourceUpdatedSteps.scala
johnduffell: maybe a better way to say it is
EitherT[F[_], A, B] drills through a monad F[_] to flatmap an Either[A,B] inside the F.
Unfortunately, Reader[D, T] takes 2 type parameters, but EitherT is looking for something that takes one.
This means that when scala is looking for an implicit, it can't work out that it can use Reader[ZuoraDeps, _] in place of F[_].
The solution if it were normal function parameters would be to partially apply the function to end up with a single argument function.
In fact the solution is the same for type constructors, however the syntax for partial application is to define another type that has one type parameter, hard code the other one, and use that type.
so `({ type XReader[AA] = Reader[ZuoraDeps, AA] })#XReader` is a one parameter type constructor that represents a Reader of its parameter.

PR #39, opened 2018-01-18
update dependencies to the latest

johnduffell:
I started adding cloudwatch metrics for when something unexpected happens, but then I realised our deps weren't up to date.

This was the result.  GDPR level up!

We should try to minimise our deps to only things supported by big organisations (e.g. scala, play, AWS libs) then we have half a chance of keeping up to date without binary conflicts.

@jacobwinch @paulbrown1982 @lmath

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -41,7 +41,6 @@ libraryDependencies ++= Seq(
   "org.scalaz" %% "scalaz-core" % "7.2.18",
   "com.typesafe.play" %% "play-json" % "2.6.8",
   "org.scalatest" %% "scalatest" % "3.0.1" % "test",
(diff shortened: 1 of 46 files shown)


johnduffell:
this might be a close due to the comments in the standup about the focus being on members data api

paulbrown1982:
👍

johnduffell:
this seems to pass in prod - I had trouble testing in CODE due to the signature validation, so I added functionality to disable checking with a config setting.
once we have metrics in prod, it might be nice to have one for "signatures not checked in prod" then we can alert on it if we accidentally enable it.

it would be great to get another +1 if someone has time then I can ship next week

PR #40, opened 2018-01-19
Add the customerSourceUpdated lambda to the readme

lmath:
Since we've recently added a new lambda to this repo, I think we should mention this in the README. Any other essential info to be updated about the customerSourceUpdated lambda?
cc @johnduffell @jacobwinch @paulbrown1982


johnduffell:
excellent, good spot! 👍 🥇

PR #41, opened 2018-02-27
Use config from encrypted bucket 

jacobwinch:
cc @pvighi @johnduffell


cloudformation.yaml
```diff
@@ -50,12 +50,18 @@ Resources:
                           - Effect: Allow
                             Action: sqs:*
                             Resource: !Sub arn:aws:sqs:${AWS::Region}:${AWS::AccountId}:subs-welcome-email
-                - PolicyName: ReadPrivateCredentials
+                - PolicyName: ReadOldPrivateCredentials
```
jacobwinch: Needs a separate PR to remove access to the old S3 bucket (which I'll also delete)

PR #43, opened 2018-02-27
Remove access to old S3 bucket

jacobwinch:
Tidy up after https://github.com/guardian/zuora-auto-cancel/pull/41

cc @pvighi


jacobwinch:
This old bucket has now been deleted.

PR #44, opened 2018-02-28
pre work for separate lambda projects in one repo

johnduffell:
I am adding some new lambdas for backfilling identity ids to allow people to manage their subs online by logging in.
This would normally be a nice little script, run locally, but since this is important I am doing it as real code that runs in a production env.
Recently we have bundled together the three lambdas into one repo, build, codebase, dependencies, jar, deploy, api gateway, stack.  Other lambdas have been completely separate in all these things.
I am trying to make a middle ground where we can be flexible in the shared/separate aspects of things. 
 So the aim for this is to add another lambda into one repo and build, and they can share some aspects of their code and the build, but have separate deploys, stacks, dependencies, jars.
This lets us share dependencies and guarantee all apps are updated when we ship a change to something shared, but keeps the development of the non shared code separate.

This does swap one set of problems for another, but I think we need to become smarter with using modules within a project appropriately, and create them and remove them as easily as any other refactoring.  Hopefully we can set a precenent for more flexibility.

I would like opinions on what I've done, I've still left the repo name the same, not attempted to split the existing at all, and added the new one piggy backed.  Of course in time we should split the root module into aggregated projects pretty much along the lines of the 3 packages plus extra modules for any shared code (not called "common" or any other catch-all name!)

Next step will be to actually start the code, at the moment it can deploy successfully .  You can see Identity Backfill and Zuora Auto Cancel separately.  I wasn't sure whether to add the Identity Backfill as another layer below the Zuora Auto Cancel so people know which repo to look in?
image

The single build will build both (we could have two builds if we wanted)
image
You can do project identity-backfill or project root to decide what to build locally.

Comments please!  I've tagged people who hopefully can also help with sbt knowledge :) @jacobwinch @paulbrown1982 @pvighi @davidfurey @lmath @rupertbates


identity-backfill/cfn.yaml
```diff
@@ -0,0 +1,169 @@
...
+                - PolicyName: LambdaPolicy
+                  PolicyDocument:
+                      Statement:
+                          - Effect: Allow
+                            Action:
+                            - logs:CreateLogGroup
+                            - logs:CreateLogStream
+                            - logs:PutLogEvents
+                            - lambda:InvokeFunction
+                            Resource: "*"
```
jacobwinch: It would be nice if we could restrict this policy so that the lambda can only write into specific log group. If possible, it would be better to define the log group in Cloudformation too, so that we can explicitly set a retention policy. I'm not exactly sure how to do this for a lambda (since AWS normally defines the name of the log group), but I've done this for a few Play apps (e.g. https://github.com/guardian/subscriptions-frontend/pull/1063/files)

identity-backfill/cfn.yaml
```diff
@@ -0,0 +1,169 @@
...
+            Integration:
+              Type: AWS_PROXY
+              IntegrationHttpMethod: POST
+              Uri: !Sub arn:aws:apigateway:${AWS::Region}:lambda:path/2015-03-31/functions/${IdentityBackfillLambda.Arn}/invocations
+        DependsOn:
+        - IdentityBackfillAPI
+        - IdentityBackfillLambda
+        - IdentityBackfillProxyResource
+
+
```
jacobwinch: There are a few lines of whitespace here, is that intentional?

identity-backfill/cfn.yaml
```diff
@@ -0,0 +1,169 @@
...
+        Properties:
+            RestApiId: !Ref IdentityBackfillAPI
+            ParentId: !GetAtt [IdentityBackfillAPI, RootResourceId]
+            PathPart: identity-backfill
+        DependsOn: IdentityBackfillAPI
+
+    IdentityBackfillMethod:
+        Type: AWS::ApiGateway::Method
+        Properties:
+            AuthorizationType: NONE
```
jacobwinch: Are you planning to write some code to check auth, or will you enable this in a different PR?

build.sbt
```diff
@@ -22,7 +22,13 @@ scalacOptions ++= Seq(
   "-Ywarn-value-discard"
 )
 
-lazy val root = (project in file(".")).enablePlugins(RiffRaffArtifact)
+// currently the original code is lying in the root, in due course we need to make three separate sub projects for these original lambdas
```
jacobwinch: 👍 - we should probably also rename this repo (and the API which the other 3 lambdas use), but that's for another PR!

johnduffell: yes indeed! definitely we need a more generic name

identity-backfill/cfn.yaml
johnduffell: hasn't really got that far.  should I be deleting parts of the cloud formation just to be on the safe side?

identity-backfill/cfn.yaml
johnduffell: this was previously to separate the stuff there's only one of from the stuff there is one for each endpoint.  but there's only one in this cloudformation.  In future there may be a few.  Normally it'd be mutliple files but I don't use cloudformation enough to know how that would work here.

identity-backfill/cfn.yaml
jacobwinch: You could just add an API key for now? This would allow you to keep testing relatively easily, but would stop anybody else from calling this API successfully.

There's an example here:
https://github.com/guardian/fulfilment-lookup/blob/master/cloudformation.yaml#L102-L134
https://github.com/guardian/fulfilment-lookup/blob/master/cloudformation.yaml#L149

(I can also show you the Postman config I've got if it's not obvious to work out how to call this with an API key).

pvighi:
👍 looks good, if we ever need to share cloudformation resources as well we could also have a base template with exports to be imported in the specific lambda templates

identity-backfill/cfn.yaml
```diff
@@ -43,7 +43,7 @@ Resources:
                             - logs:CreateLogStream
                             - logs:PutLogEvents
                             - lambda:InvokeFunction
-                            Resource: "*"
+                            Resource: !Sub "arn:aws:logs:eu-west-1:865473395570:log-group:/aws/lambda/identity-backfill-${Stage}:log-stream:*"
```
paulbrown1982: This should be replaced with CF parameters and the git history overwritten!

PR #45, opened 2018-03-05
move zuora lambda code into a separate sbt module

johnduffell:
This PR splits the zuora code out from the auto cancel lambdas into a separate sbt module in the same overall sbt build.
This means that both identity backfill and zuora auto cancel family can depend on the zuora stuff, without having to share all their other dependencies.  This should result in cleaner code and smaller jars, without the overhead of separate repos and shipping libraries to maven.  It also means the code can't get out of step between its dependencies.
The flip side is that if you change the zuora code, you will deploy it in all envs, so you have to make sure you are prepared to ship everything.
@paulbrown1982 @pvighi 

I have testing in CODE and it seems to work the same.


build.sbt
```diff
@@ -1,34 +1,48 @@
...
-  "-language:higherKinds",
-  "-language:implicitConversions",
-  "-unchecked",
-  "-Xfatal-warnings",
-  "-Xlint",
-  "-Yno-adapted-args",
-  "-Ywarn-dead-code",
-  "-Ywarn-numeric-widen",
-  "-Ywarn-value-discard"
+val scalaSettings = Seq(
```
johnduffell: settings defined in the sbt file at the top level only apply to that project.
To apply them to other projects, you either need to duplicate them in an SBT file inside the project folder, or put them in a val and do project.settings(.....) in the top level sbt file.
If the settings are shared, for example which version of scala, it's best to just do them once at the top level.
For things that are more local to the library e.g. dependencies it would make more sense to have them defined in the sbt folder.
This does add the possibility of inconsistencies of libraries between projects.  If there are, sbt reports the old favourite error:
```
[warn] Found version conflict(s) in library dependencies; some are suspected to be binary incompatible:
[warn] 
[warn] 	* org.scalaz:scalaz-core_2.12:7.2.18 is selected over 7.1.1
[warn] 	    +- com.gu:zuora_2.12:0.0.1                            (depends on 7.2.18)
[warn] 	    +- com.gu:identity-backfill_2.12:0.0.1                (depends on 7.1.1)
```

identity-backfill/build.sbt
```diff
@@ -3,27 +3,6 @@
 name := "identity-backfill"
 description:= "links subscriptions with identity accounts"
 
-scalaVersion := "2.12.4"
```
johnduffell: these are in scalaSettings at the top level now so they can be shared

src/main/scala/com/gu/autoCancel/AutoCancel.scala
```diff
@@ -18,6 +20,6 @@ object AutoCancel extends Logging {
       _ <- ZuoraCancelSubscription(subToCancel, cancellationDate).withLogging("cancelSubscription")
       _ <- ZuoraDisableAutoPay(accountId).withLogging("disableAutoPay")
     } yield ()
-    zuoraOp.run.run(zuoraDeps)
+    zuoraOp.leftMap(ApiGatewayResponse.fromClientFail).run.run(zuoraDeps)
```
johnduffell: zuora client doesn't (shouldn't) have access to the ApiGatewayResponse any more, so it has its own new `ClientFail` type which is adapted by the caller here to the api gateway class

src/main/scala/com/gu/stripeCustomerSourceUpdated/zuora/ZuoraQueryPaymentMethod.scala
```diff
@@ -1,48 +1,33 @@
...
 import scalaz.{ -\/, NonEmptyList, \/- }
+import com.gu.util.zuora.ZuoraAccount._
+import com.gu.util.zuora.{ ZuoraDeps, ZuoraQuery }
+import com.gu.util.zuora.ZuoraQuery.Query
+import com.gu.util.reader.Types._
 
-object ZuoraQueryPaymentMethod {
+object ZuoraQueryPaymentMethod extends Logging {
 
-  case class SecondTokenId(value: String) extends AnyVal
```
johnduffell: this file didn't move to zuora package as it's quite specific to the stripe operation, but the field wrappers are pretty standard to zuora so have moved out to ZuoraAccount.scala

PR #46, opened 2018-03-07
Adding billing details to Payment failure email

pvighi:
before merging this we have to 

- ~~Once ET is configured test in code the fields are populated correctly in the data extension~~ Done
- ~~Get the correct field names to send to the data extension~~ Done

After merging:
Add the missing fields to the PROD Zuora callout for payment failure 1 to match the UAT configuration.  The billing fields are optional so we should be able to add these after merging the changes


src/test/scala/com/gu/paymentFailure/EndToEndHandlerTest.scala
```diff
@@ -28,9 +28,39 @@ class EndToEndHandlerTest extends FlatSpec with Matchers {
...
+    val responseString = new String(os.toByteArray(), "UTF-8")
+
+    val expectedResponse =
+      s"""
+         |{"statusCode":"200","headers":{"Content-Type":"application/json"},"body":"Success"}
+         |""".stripMargin
+    responseString jsonMatches expectedResponse
+  }
+
+  it should "manage an end to end call with billing details" in {
```
pvighi: This is just a copy and paste of the regular test but with billing details included in the zuora callout.
 I didn't just edit the existing one because I wanted to make sure nothing changes for the callouts that don't include billing details.
I will refactor this try to remove as much duplication as I can before merging this PR

johnduffell: good reasoning on adding an extra test, and good idea about reducing the duplication

johnduffell:
looks good to me!  I'll wait for the do not merge to go away for a final look

src/test/scala/com/gu/paymentFailure/EndToEndHandlerTest.scala
```diff
@@ -17,8 +17,29 @@ class EndToEndHandlerTest extends FlatSpec with Matchers {
     Lambda.default(config.rawEffects)(stream, os, null)
 
     config.resultMap
-      .get(("POST", "/messaging/v1/messageDefinitionSends/111/send")) should be(
-        Some(Some(EndToEndData.expectedEmailSend))) // TODO check the body too
+      .get(("POST", "/messaging/v1/messageDefinitionSends/111/send")).get.get jsonMatches (
```
pvighi: the old way of comparing would fail for equivalent JSON but with different formatting or field order and this caused false test failures when refactoring

paulbrown1982:
👍 Agree the title is better sitting up with the name.

PR #47, opened 2018-03-08
make identity backfill endpoint always return success

johnduffell:
This PR pulls out the API gateway helper code so that we can use it from the identity backfill package.  This means that the identity backfill can now respons with "Success" when you hit it with a request.

Next stage is to
- ___add to the IdentityBackfillSteps to actually access zuora and identity, and do the checking that it can be populated.___ *THIS WILL HAPPEN LATER*
- fetch the idenitity id from the ID API by email address (not including the address - yet) use the credentials we already have in membership-workflow for now
-  do the changes in zuora (add identity ID to account)
- do the changes in SF (add identity ID to SF contact) need a SF credentials in the S3 file
- make sure the address is fine before the sync?
- add a message to the SQS queue (to sync the address from SF to identity)
- the updated account should be logged in SF

PS I am tempted to add a "healthcheck" mode to the ApiGatewayHandler and "EndToEndHandlerTest" so that we can run the test in that mode and it will call the API gateway with the testingraweffects object as a parameter and it will return the requests made as part of the execution with the response.  This means we can have some certainty that the code is working in CODE or PROD without having to wait for the errors to come through.


api-gateway/src/main/scala/com/gu/util/ConfigLoader.scala
```diff
@@ -71,7 +70,7 @@ object StripeConfig {
     (JsPath \ "signatureChecking").readNullable[String].map(!_.contains("false")))(StripeConfig.apply _)
 }
 
-case class Config(
+case class Config[ZuoraRestConfig](
```
johnduffell: `ZuoraRestConfig` is a generic argument name - normally you'd see it called `Z` or something, but that's not helpful.  Any resemblance to the class `ZuoraRestConfig` is coincidental.
I added it because I don't want the API gateway code to depend on the zuora client, or vice versa.

api-gateway/src/main/scala/com/gu/util/apigateway/ApiGatewayHandler.scala
```diff
@@ -8,73 +8,74 @@ import com.gu.util._
...
+import play.api.libs.json.{ Json, Reads }
 
 import scala.io.Source
 import scala.util.Try
 import scalaz.{ -\/, \/, \/- }
 
 object ApiGatewayHandler extends Logging {
 
-  //  code dependencies
-  case class HandlerDeps(
```
johnduffell: I've got rid of the HandlerDeps class and made ApiGatewayHandler into a class with the same parameters and a companion object.  This effectively means that all methods in the class have access to all of the deps.  That wouldn't be ideal from a dependency point of view, as if someone adds one method to the class it gets a potential dependency on all of the deps, but I thought maybe it will confuse people less to have it that way.

build.sbt
```diff
@@ -20,7 +20,9 @@ val scalaSettings = Seq(
     "-Ywarn-dead-code",
     "-Ywarn-numeric-widen",
     "-Ywarn-value-discard"
-  )
+  ),
+  javaOptions in Test += s"""-Dlog4j.configuration=file:${new File(".").getCanonicalPath}/test_log4j.properties""",
```
johnduffell: I had to add this and the fork so that the logging appears when we're running the tests.  Otherwise only the test passed/failed appears, which is not as useful

identity-backfill/cfn.yaml
```diff
@@ -54,7 +59,7 @@ Resources:
             Code:
                 S3Bucket: zuora-auto-cancel-dist
                 S3Key: !Sub membership/${Stage}/identity-backfill/identity-backfill.jar
-            Handler: com.gu.identity_backfill.Handler::apply
+            Handler: com.gu.identityBackfill.Handler::apply
```
johnduffell: changed this, as although java naming conventions say we should use underscores, I don't think we did that elsewhere

identity-backfill/cfn.yaml
```diff
@@ -44,7 +44,12 @@ Resources:
                             - logs:PutLogEvents
                             - lambda:InvokeFunction
                             Resource: !Sub "arn:aws:logs:${AWS::Region}:${AWS::AccountId}:log-group:/aws/lambda/identity-backfill-${Stage}:log-stream:*"
-
+                - PolicyName: ReadPrivateCredentials
+                  PolicyDocument:
+                      Statement:
+                          - Effect: Allow
+                            Action: s3:GetObject
+                            Resource: !Sub arn:aws:s3:::gu-reader-revenue-private/membership/payment-failure-lambdas/${Stage}/*
```
johnduffell: at the moment we are using the same config dir. interesting that the config is payment-failure-lambdas but the repo is called zuora-auto-cancel (and the actual lambdas are doing other things...)

identity-backfill/src/main/scala/com/gu/identityBackfill/IdentityBackfillSteps.scala
```diff
@@ -0,0 +1,31 @@
...
+    //val zuoraDeps = ZuoraDeps(response, config.zuoraRestConfig)
+    new IdentityBackfillSteps().apply
+  }
+
+}
+
+class IdentityBackfillSteps() extends Logging {
+
+  def apply(apiGatewayRequest: ApiGatewayRequest): FailableOp[Unit] = {
+    println("ap")
```
johnduffell: this whole things is a big todo, see the description

build.sbt
```diff
@@ -33,16 +35,39 @@ lazy val zuora = project.settings(scalaSettings).settings(
...
 // currently the original code is lying in the root, in due course we need to make three separate sub projects for these original lambdas
 // they should produce their own self contained jar to reduce the artifact size and startup time.  Any shared code can be
 // a set of projects that is "dependsOn(..)" by the sharing projects.  Don't be afraid to restructure things to keep the code nice!
 lazy val root = (project in file(".")).settings(scalaSettings).enablePlugins(RiffRaffArtifact).aggregate(
   `identity-backfill`,
-  zuora
-).dependsOn(zuora)
+  zuora,
+  effects
+).dependsOn(zuora, `api-gateway`, effects % "compile->compile;test->test")
```
jacobwinch: What is this line for?

effects/src/test/scala/com/gu/effects/TestingRawEffects.scala
```diff
@@ -0,0 +1,94 @@
...
+    val body = buffer.readString(UTF_8)
+    val url = request.url
+    BasicResult(request.method(), url.encodedPath(), body)
+  }
+
+  val response: Request => Response = {
+    req =>
+      result = req :: result
+      val (code, response) = responses.getOrElse(req.url().encodedPath(), (defaultCode, """{"success": true}"""))
+      println(s"request for: ${req.url().encodedPath()} so returning $response")
```
jacobwinch: I think this should just be logged

identity-backfill/cfn.yaml
jacobwinch: I think we wanted to rename everything to 'payment-failure-lambdas', but only got as far as doing the S3 bucket!

identity-backfill/src/test/scala/com/gu/identityBackfill/EndToEndHandlerTest.scala
```diff
@@ -0,0 +1,103 @@
...
+      |    },
+      |    "queryStringParameters": {
+      |        "apiClientId": "a",
+      |        "apiToken": "b"
+      |    },
+      |    "pathParameters": null,
+      |    "stageVariables": null,
+      |    "requestContext": {
+      |        "path": "/CODE/payment-failure",
+      |        "accountId": "865473395570",
```
jacobwinch: we should remove this

src/test/scala/com/gu/stripeCustomerSourceUpdated/SourceUpdatedStepsTest.scala
```diff
@@ -274,7 +275,7 @@ class SourceUpdatedStepsTest extends FlatSpec with Matchers {
       expiry = StripeExpiry(exp_month = 7, exp_year = 2020),
       last4 = StripeLast4("1234"))
 
-    val actual = SourceUpdatedSteps.createUpdatedDefaultPaymentMethod(PaymentMethodFields(PaymentMethodId("PMID"), AccountId("fake"), NumConsecutiveFailures(1)), eventData).run.run(effects.zuoraDeps)
+    val actual = SourceUpdatedSteps.createUpdatedDefaultPaymentMethod(PaymentMethodFields(PaymentMethodId("PMID"), AccountId("fake"), NumConsecutiveFailures(1)), eventData).run.run(TestData.zuoraDeps(effects))
```
jacobwinch: Can we delete the /**/?

api-gateway/src/main/scala/com/gu/util/ConfigLoader.scala
```diff
@@ -1,7 +1,6 @@
 package com.gu.util
```
jacobwinch: Personally I think ConfigLoader would fit more naturally in the effects project (instead of apigateway)

api-gateway/src/main/scala/com/gu/util/Logging.scala
```diff
@@ -1,11 +1,13 @@
 package com.gu.util
```
jacobwinch: Similarly I don't really think the logging trait belongs here

src/main/scala/com/gu/util/ZuoraToApiGateway.scala
```diff
@@ -0,0 +1,11 @@
+package com.gu.util
+
+import com.gu.util.apigateway.ResponseModels.{ ApiResponse, Headers }
+import com.gu.util.zuora.internal.ClientFail
+
+object ZuoraToApiGateway {
+
+  def fromClientFail(clientFail: ClientFail): ApiResponse =
+    ApiResponse(clientFail.statusCode, new Headers, s"zuora client fail: ${clientFail.message}")
```
jacobwinch: Now that we have pulled the API Gateway helper code out, I find it a bit confusing that we are passing ApiResponses around at this level (it would make more sense to me if we just passed some more generic error in a Left, and then let ApiGatewayHandler add the headers etc.)

src/test/scala/com/gu/stripeCustomerSourceUpdated/SourceUpdatedStepsTest.scala
jacobwinch: (it's confusing for me and GitHub apparently!)

api-gateway/src/main/scala/com/gu/util/ConfigLoader.scala
jacobwinch: What about something more generic ('CombinedConfig'?), so it's clear(er) that there is no resemblance to the ZuoraRestConfig class?

johnduffell: good idea, I've gone for StepsConfig
I've left the value name called zuoraRestConfig for now, as it will require refactoring the config file.  When we split something else out and need a new type combined into the class, it would be necessary to really combine them into one level in the config.

build.sbt
johnduffell: I'm adding a sensiubly named value for this
```scala
val effectsDepIncludingTestFolder: ClasspathDependency = effects % "compile->compile;test->test"
```

effects/src/test/scala/com/gu/effects/TestingRawEffects.scala
johnduffell: now that I've enabled logging in tests, that will work!

identity-backfill/src/test/scala/com/gu/identityBackfill/EndToEndHandlerTest.scala
johnduffell: every 'arn' has the account id, as it is the equivalent of the domain of our AWS account
https://github.com/search?q=org%3Aguardian+865473395570&type=Code

src/test/scala/com/gu/stripeCustomerSourceUpdated/SourceUpdatedStepsTest.scala
johnduffell: I'm splitting them into separate objects with separate names.  That is making me think that perhaps they should be separate in the code as well as the tests..

api-gateway/src/main/scala/com/gu/util/ConfigLoader.scala
johnduffell: effects is only for code that actually does something to the outside world, and is untested and should be as small as possible.
as far as I can tell, there are no side effects of config loader, it's all pure functions and unit test coverage seems to reflect that (although it's not 100%)
The side effects happen in the AwsS3.scala (although I think even some of the that code shouldn't be there as it doesn't have side effects.)

api-gateway/src/main/scala/com/gu/util/Logging.scala
johnduffell: true, I think logging is a bit of a side effect, but I don't want any projects other than the handlers to depend on effects package.  Since logging is a fairly safe "write only" operation that we don't tend to test or mock, I would be happy do it anywhere and everywhere.

Other approaches would be to have a logging project which everything can import and use from everywhere, or to have a log function coming from the top handler in RawEffects.default everywhere and pass it down in the same way as the Http response generator.

Any thoughts?

src/main/scala/com/gu/util/ZuoraToApiGateway.scala
johnduffell: this level (the handler) depends on api gateway and zuora, so it's the only place that we could convert from one to the other

the handler should be a generic way of mapping things from one domain to the other (zuora to api gateway), ideally shouldn't have too much logic, it should all be wiring/mapping code.However I don't want to get too carried away with restructuring at this stage, I'm focussing on code that needs to be reused, rather than code that should have dependencies in a certain way.

api-gateway/src/main/scala/com/gu/util/ConfigLoader.scala
jacobwinch: OK, that sounds sensible, I think you're right that this doesn't belong in effects... I guess the problem I was trying to get at is:

Loading config is entirely unrelated to the fact that these lambdas sit behind an API Gateway - so I found it a bit confusing that this code was located in a project called api-gateway

api-gateway/src/main/scala/com/gu/util/Logging.scala
jacobwinch: I guess my feelings about this file are similar to ConfigLoader (i.e. it's not really to do with API Gateway, so why does it live here?). However, I appreciate you're probably reluctant to have a 'common'/'util' project which contains this type of thing as that could easily become incredibly bloated. 

Feel free to leave it here for now, as we should probably have a more general re-think about logging (e.g. adding Sentry + Kibana) and could tackle this part of the re-structure then

api-gateway/src/main/scala/com/gu/util/ConfigLoader.scala
johnduffell: rename to common? 😆

johnduffell: I guess this is more like the "do something from api gateway using our normal standards" which is api key in the url, config in S3, stage CODE or PROD, etc.  Not sure what the name for that is, maybe just "main" or "handler" ?

api-gateway/src/main/scala/com/gu/util/Logging.scala
johnduffell: yes as above I want to avoid any kind of catch all names that anything could go in .  the trouble is as soon as you introduce some specificity in the naming, there are some things you can't put in any more.  I think this includes only things to do with how we as a team like to invoke lambdas from API gateway.  So not only the api gateway stuff but the S3 loader and logging.
Perhaps the s3 and logging would be sharable with lambdas that don't use api gateway so we would split this in future for reuse.
I'm starting to think "handler" would be better than "api gateway", if that's still specific enough

johnduffell:
this is ready for final checking and +1ing please! 🎉 
@jacobwinch @paulbrown1982 see what you think, I think I've either answered/argued back or fixed every comment, if not let me know 😄

johnduffell:
thanks for all the feedback 💯 
the main thing is to always keep the structure under review, adjusting the sbt project structure etc needs to become part of normal refactoring that everyone is comfortable to do
I will ship this tomorrow morning rather than right now!

PR #48, opened 2018-03-12
Document everything

johnduffell:
Suddenly realised I have basically not really documented anything.  This is a first pass, but should be improved and changed over time.

@paulbrown1982 @pvighi @jacobwinch @AWare @lmath

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # zuora-auto-cancel
 
-This is the reader rev lambda API/orchestration layer.  But it has a bad name so far - TODO fix this. 
+This is the reader revenue lambda API/orchestration layer.  But it has a bad name so far - TODO fix this. 
 
 Please keep all the various README in this project up to date, and improve them!
 There should be one in each project and anywhere else you think it's would help.
--- a/README.md
+++ b/README.md
@@ -1,5 +1,11 @@
 # zuora-auto-cancel
-Contains three Scala lambdas behind the same API gateway:
+
+Please keep all the README in this project up to date, and improve them!  There should be one in each project.
+
+The main project aggregates all the sub projects from handlers and lib, so we can build and test them in one go.
+
+## root
+Contains three Scala lambdas behind the same API gateway.  TODO These should be moved into a new set of 3 projects in the handlers folder.
 
 **autoCancel**: 
 Used to cancel subscriptions with overdue invoices, based on an event trigger within Zuora.
--- a/README.md
+++ b/README.md
@@ -1,7 +1,34 @@
 # zuora-auto-cancel
 
-Please keep all the README in this project up to date, and improve them!  There should be one in each project.
+This is actually the start of the reader rev lambda API repo possibly.  But it has a bad name so far. 
 
+Please keep all the various README in this project up to date, and improve them!
+There should be one in each project and anywhere else you think it's would help.
+
+## philosophy
+The general philosophy of this project is to keep things under review and don't be afraid to refactor things
+esepecially the hard parts of the structure when you think they are wrong.
+The PR review at the end will always let you know when people don't like the idea
+but if you miss an opportunity to improve the structure, this hurts more in the long term than a good
+change with poor naming.
+Yes, small PRs are good, but that means small in terms of explanation, rather than small in terms
+of lines touched.  If you split out some code into a separate subproject and rename an existing one,
+that is a small change because I can explain it in one sentence.  Github's failings in terms
+of displaying it concisely are not your failings in making it small!
+
+Anything that isn't a line in the sand should be questioned and changed at will.
+Anything that is should be questioned too.
+
+## Guidelines in the sand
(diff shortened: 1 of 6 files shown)


handlers/identity-backfill/README.md
```diff
@@ -0,0 +1,6 @@
+# identity-backfill
+
+This is a lambda to add identity accounts to zuora and salesforce where appropriate.
+
+It is deployed as Membership Admin::Identity Backfill and this will do the cloudformation at the same time.
+Use postman to post to https://{{autoCancelCloudfrontStage}}/auto-cancel?apiToken={{apiToken}}
```
jacobwinch: I think this is the wrong path for this project

README.md
```diff
@@ -1,7 +1,34 @@
 # zuora-auto-cancel
 
-Please keep all the README in this project up to date, and improve them!  There should be one in each project.
-
+This is actually the start of the reader rev lambda API repo possibly.  But it has a bad name so far.
```
jacobwinch: actually or possibly - make up your mind!

README.md
```diff
@@ -1,7 +1,34 @@
...
-Please keep all the README in this project up to date, and improve them!  There should be one in each project.
-
+This is actually the start of the reader rev lambda API repo possibly.  But it has a bad name so far. 
+
+Please keep all the various README in this project up to date, and improve them!
+There should be one in each project and anywhere else you think it's would help.
+
+## philosophy
+The general philosophy of this project is to keep things under review and don't be afraid to refactor things
+esepecially the hard parts of the structure when you think they are wrong.
```
jacobwinch: typo

README.md
```diff
@@ -1,7 +1,34 @@
...
+Anything that is should be questioned too.
+
+## Guidelines in the sand
+- good naming - a good name promotes cohesion - it says more about what **shouldn't** be in the construct
+than what should be.  If you have a catch all name like "common" or "helpers" in mind, think again.
+- effects separation - to promote good reuse and testability, keep all side effects in one place, and only depend
+on it from the top level handlers.  Effects should be minimal, and the top level handlers should mostly be wiring.
+If there's any code in either that you feel could be unit tested, it should probably be in another project.
+- one jar per lambda - minimise the size of the deployment artifact
+- minimise dependencies (aka liabilities) on external libraries
```
jacobwinch: it might be worth a note on why this is a good/important thing

README.md
```diff
@@ -1,7 +1,34 @@
 # zuora-auto-cancel
 
-Please keep all the README in this project up to date, and improve them!  There should be one in each project.
-
+This is actually the start of the reader rev lambda API repo possibly.  But it has a bad name so far. 
+
+Please keep all the various README in this project up to date, and improve them!
+There should be one in each project and anywhere else you think it's would help.
+
+## philosophy
```
jacobwinch: I agree with pretty much everything you've said here - however, most of it is (read: should be) applicable to every project we work on, so I wonder whether it belongs in this README or whether we be looking for better ways to promote this culture more generally (or both?!)

README.md
johnduffell: it's definitely the start
but I'm only 75% sure of that?

I feel like we need a way of being clear about how definite things are (lines in the sand are 95%, random decisions that were too hard are 25% and best intentions are 60%?

README.md
johnduffell: maybe... there's the central https://github.com/guardian/recommendations repo, but it's too big.  I feel like 4 lines in the sand is too many already.  I think this readme needs simplifying a bit.  It's a start I suppose.

handlers/identity-backfill/README.md
johnduffell: fixed and pushed

README.md
johnduffell: fixed and pushed

README.md
johnduffell: fixed and pushed

johnduffell:
I've tweaked everything, everyone can have (another) look and see what you think please!

README.md
johnduffell: I've fixed this! `This is the reader rev lambda API/orchestration layer.`

README.md
```diff
@@ -1,5 +1,38 @@
 # zuora-auto-cancel
-Contains three Scala lambdas behind the same API gateway:
+
+This is the reader rev lambda API/orchestration layer.  But it has a bad name so far - TODO fix this.
```
pvighi: maybe use "revenue" here instead of "rev". Probably doesn't matter but maybe someone in the future could get confused

README.md
pvighi: I agree with Jacob,  a slightly more general version of this could be linked from pretty much every project we have although there's no harm in having it here

README.md
johnduffell: done!

johnduffell:
thanks for the comments @pvighi and @jacobwinch 
we should keep thinking about these and revise/simplify both the docs and the structure as and when.
merged this as a first pass! 🚀

PR #49, opened 2018-03-16
Add identity lookup to identity backfill lambda

johnduffell:
This adds an identity lookup plus tests to the identity backfill.  This is needed to get the identity id of a user that we want to populate into zuora/sf.

I have also partly undone the refactoring I made to api gateway handler previously that got rid of the Deps object.  I haven't put back the Deps object, but I've just passed the deps in as parameters.  If there are too many deps, then the function is probably doing too much.

There is also a new manualTest, so if you run it locally with valid identity credentials, you can see it working.  Maybe we need some more documentation oh how to do this as it's not ideal to have a "test" that doesn't run automatically.

I have tested in CODE and all the lambdas seem to work in general.

@pvighi @paulbrown1982 @lmath @jacobwinch please have a quick look

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,26 @@ If there's any code in either that you feel could be unit tested, it should prob
 - **one jar per lambda** - minimise the size of the deployment artifact
 - **minimise dependencies (aka liabilities)** on external libraries as we have to keep them up to date, also they increase the size of the artifact
 
+# Howto update config
+At the moment we just use S3 for config.
+
+If you're making an addition, you can just copy the file from S3 for PROD and CODE, then update and upload.
+Check the version in the AwsS3.scala ConfigLoad object.
+`aws s3 cp s3://gu-reader-revenue-private/membership/payment-failure-lambdas/CODE/payment-failure-lambdas.private.v<VERSION>.json /etc/gu/CODE/ --profile membership`
+Then do the reverse to upload again.
+
+If you're making a change that you only want to go live on deployment (and have the ability to roll back
+with riffraff) then you can increment the version.  Make sure you upload the file with the new version,
+otherwise it will break the existing lambda immediately.
+
+Follow the same process to update the prod config.
+
+To check that you have done it correctly, run the ConfigLoaderSystemTest.
+You will have to remove the @Ignore from it before you can run it.  Remember to replace afterwards.
+That will check the latest version can be understood by the current version of the code.
+
+Ideally this test should be automated and your PR shouldn't be mergable until the config is readable.
+
 ## structure
 The main project aggregates all the sub projects from handlers and lib, so we can build and test them in one go.
 
--- a/handlers/identity-backfill/src/main/scala/com/gu/identity/GetByEmail.scala
+++ b/handlers/identity-backfill/src/main/scala/com/gu/identity/GetByEmail.scala
@@ -35,8 +35,7 @@ object GetByEmail {
       case _ => -\/(ApiError("not an OK response from api"))
     }
 
-  def apply(deps: IdentityClientDeps)(email: EmailAddress): ApiError \/ IdentityId = {
-    import deps._
+  def apply(email: EmailAddress)(getResponse: Request => Response, identityConfig: IdentityConfig): ApiError \/ IdentityId = {
 
     val url = HttpUrl.parse(identityConfig.baseUrl + "/user").newBuilder().addQueryParameter("emailAddress", email.value).build()
     val response = getResponse(new Request.Builder().url(url).addHeader("X-GU-ID-Client-Access-Token", "Bearer " + identityConfig.apiToken).build())
@@ -60,7 +59,3 @@ case class IdentityConfig(
 object IdentityConfig {
   implicit val reads: Reads[IdentityConfig] = Json.reads[IdentityConfig]
 }
-
-case class IdentityClientDeps(
-  getResponse: Request => Response,
-  identityConfig: IdentityConfig)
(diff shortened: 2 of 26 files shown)


handlers/identity-backfill/src/main/scala/com/gu/identity/GetByEmail.scala
```diff
@@ -0,0 +1,66 @@
...
+import scalaz.syntax.std.either._
+
+object GetByEmail {
+
+  case class EmailAddress(value: String)
+  case class IdentityId(value: String)
+
+  case class ApiError(message: String)
+
+  object RawWireModel {
```
johnduffell: I have made this into the raw model of strings, so we can use the built in json reads deserialiser.  We can then convery this into a useful object in our domain model using IdentityId.fromUser below.  This reduces the amount of thinknng about how to use applicatives to build the domain object directly from the jsvalue

README.md
```diff
@@ -28,6 +28,26 @@ If there's any code in either that you feel could be unit tested, it should prob
...
+with riffraff) then you can increment the version.  Make sure you upload the file with the new version,
+otherwise it will break the existing lambda immediately.
+
+Follow the same process to update the prod config.
+
+To check that you have done it correctly, run the ConfigLoaderSystemTest.
+You will have to remove the @Ignore from it before you can run it.  Remember to replace afterwards.
+That will check the latest version can be understood by the current version of the code.
+
+Ideally this test should be automated and your PR shouldn't be mergable until the config is readable.
```
johnduffell: we really should sort out some pre-deployment tests.  Since this needs pretty strong AWS access, it's hard to automate.
Same applies to a healthcheck mode.  At the moment I check the lambdas by hitting them with dummy data and checking they fail with the expected error.

handlers/identity-backfill/src/test/scala/manualTest/GetByEmailSystemTest.scala
```diff
@@ -0,0 +1,26 @@
...
+import com.gu.identity.GetByEmail
+import com.gu.identity.GetByEmail.EmailAddress
+import com.gu.identityBackfill.IdentityBackfillSteps.StepsConfig
+import com.gu.util.{ Config, Logging }
+
+import scala.io.Source
+import scala.util.Try
+import scalaz.syntax.std.either._
+
+// run this manually
```
johnduffell: this test has to be run manually.  It could form part of the future health check as it's a read only operation.

src/test/scala/manualTest/ConfigLoaderSystemTest.scala
```diff
@@ -15,32 +17,30 @@ class ConfigLoaderSystemTest extends FlatSpec with Matchers {
...
+      b <- Config.parseConfig[StepsConfig](a)
     } yield b
     val loadedOK = con.map(config => ())
     withClue(configAttemp) {
-      loadedOK should be(Success(()))
-      verificationStatus.foreach {
-        expected => con.get.stripeConfig.signatureChecking should be(expected)
-      }
+      loadedOK should be(\/-(()))
+      con.toOption.get.stage should be(stage)
```
johnduffell: I've made this check that the CODE and PROD config has been uploaded to the right place,

handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/IdentityBackfillSteps.scala
```diff
@@ -1,30 +1,39 @@
...
-  def apply(apiGatewayRequest: ApiGatewayRequest): FailableOp[Unit] = {
+  def apply(config: StepsConfig, getResponse: Request => Response)(apiGatewayRequest: ApiGatewayRequest): FailableOp[Unit] = {
     println("ap")
     for {
-      autoCancelCallout <- Json.fromJson[Seq[String]](Json.parse(apiGatewayRequest.body)).toFailableOp.withLogging("zuora callout")
+      autoCancelCallout <- Json.fromJson[IdentityBackfillRequest](Json.parse(apiGatewayRequest.body)).toFailableOp.withLogging("zuora callout")
       _ = println(autoCancelCallout)
+      aaa = fromRequest(autoCancelCallout)
+      bbb <- GetByEmail(aaa)(getResponse, config.identityConfig).leftMap(a => ApiGatewayResponse.internalServerError(a.toString))
+      _ = println(bbb)
```
pvighi: these aaa and bbb are going to change in the next PR so I'm fine with leaving then in temporarily

handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/IdentityBackfillSteps.scala
```diff
@@ -1,30 +1,39 @@
...
+  def fromRequest(identityBackfillRequest: IdentityBackfillRequest): EmailAddress = {
+    EmailAddress(identityBackfillRequest.emailAddress)
+  }
 
-  def apply(apiGatewayRequest: ApiGatewayRequest): FailableOp[Unit] = {
+  def apply(config: StepsConfig, getResponse: Request => Response)(apiGatewayRequest: ApiGatewayRequest): FailableOp[Unit] = {
     println("ap")
     for {
-      autoCancelCallout <- Json.fromJson[Seq[String]](Json.parse(apiGatewayRequest.body)).toFailableOp.withLogging("zuora callout")
+      autoCancelCallout <- Json.fromJson[IdentityBackfillRequest](Json.parse(apiGatewayRequest.body)).toFailableOp.withLogging("zuora callout")
```
pvighi: this should probably be renamed

johnduffell: I've already started work on the next PR, so I've just renamed in the local changes.  This will def be shipped, so I promise it won't be left like this 🤞

johnduffell:
> There are a lot of functional library changes and I'm a little unsure exactly why some of them were needed.

Not exactly sure what you mean here, but I'd agree I'm not sure why some of the changes are needed either.  I'm just trying to make it so it's easy and clear to write code, but without ending up with excessive interdependencies.

PR #51, opened 2018-03-20
Add identity backfill pre-checks against zuora

johnduffell:
This is part of the overall task to link a never transacted existing identity account with a zuora account having the same email address but a single product.

This PR adds the pre-checks against zuora that we are allowed to backfill an identity account.
These checks come after looking up the identity id by email address.

### Method
We check that there is exactly one account in zuora with that billing work email address.
We then check there are no accounts in zuora with that indeitity id
If either of those checks fail, we return a 404 to basically say there's a client error in suggesting to update that account.
If both pass, then we return a 200 no action needed if we are in dry run mode.  In prod mode we will return a 500 error "TODO".  This is because the actual code to backfill is not written yet.

unfortunately I mixed in a load of reformatting 😵 , so it's not a small size pr. Feel free to tell me to split it into two PRs and I won't be able to say no!
@paulbrown1982 @jacobwinch @pvighi @lmath

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -22,7 +22,14 @@ val scalaSettings = Seq(
     "-Ywarn-value-discard"
   ),
   javaOptions in Test += s"""-Dlog4j.configuration=file:${new File(".").getCanonicalPath}/test_log4j.properties""",
-  fork in Test := true
+  fork in Test := true,
+  {
+    import scalariform.formatter.preferences._
+    scalariformPreferences := scalariformPreferences.value
+      .setPreference(DanglingCloseParenthesis, Force)
+      .setPreference(SpacesAroundMultiImports, false)
+      .setPreference(NewlineAtEndOfFile, true)
+  }
 )
 
 lazy val zuora = (project in file("lib/zuora")).settings(scalaSettings).settings(
--- a/handlers/identity-backfill/src/main/scala/com/gu/identity/GetByEmail.scala
+++ b/handlers/identity-backfill/src/main/scala/com/gu/identity/GetByEmail.scala
@@ -1,11 +1,11 @@
 package com.gu.identity
 
-import com.gu.identity.GetByEmail.RawWireModel.{ User, UserResponse }
-import com.gu.identityBackfill.Types.{ EmailAddress, IdentityId }
-import okhttp3.{ HttpUrl, Request, Response }
-import play.api.libs.json.{ Json, Reads }
+import com.gu.identity.GetByEmail.RawWireModel.{User, UserResponse}
+import com.gu.identityBackfill.Types.{EmailAddress, IdentityId}
+import okhttp3.{HttpUrl, Request, Response}
+import play.api.libs.json.{Json, Reads}
 
-import scalaz.{ -\/, \/, \/- }
+import scalaz.{-\/, \/, \/-}
 import scalaz.syntax.std.either._
 
 object GetByEmail {
@@ -50,7 +50,8 @@ object GetByEmail {
 
 case class IdentityConfig(
   baseUrl: String,
-  apiToken: String)
+  apiToken: String
+)
 
 object IdentityConfig {
   implicit val reads: Reads[IdentityConfig] = Json.reads[IdentityConfig]
--- a/handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/Handler.scala
+++ b/handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/Handler.scala
@@ -31,12 +31,12 @@ object Handler {
     def operation: Config[StepsConfig] => ApiGatewayRequest => FailableOp[Unit] =
       config => IdentityBackfillSteps(
         GetByEmail(rawEffects.response, config.stepsConfig.identityConfig),
-        GetZuoraAccountsForEmail.apply(ZuoraDeps(rawEffects.response, config.stepsConfig.zuoraRestConfig)),
-        CountZuoraAccountsForIdentityId.apply(ZuoraDeps(rawEffects.response, config.stepsConfig.zuoraRestConfig)),
+        GetZuoraAccountsForEmail(ZuoraDeps(rawEffects.response, config.stepsConfig.zuoraRestConfig)),
+        CountZuoraAccountsForIdentityId(ZuoraDeps(rawEffects.response, config.stepsConfig.zuoraRestConfig)),
         UpdateZuoraIdentityId.apply,
         UpdateSalesforceIdentityId.apply
       )
-    ApiGatewayHandler.default[StepsConfig](implicitly)(operation, lambdaIO).run((rawEffects.stage, rawEffects.s3Load(rawEffects.stage)))
+    ApiGatewayHandler.default[StepsConfig](operation, lambdaIO).run((rawEffects.stage, rawEffects.s3Load(rawEffects.stage)))
   }
 
 }
--- a/handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/Handler.scala
+++ b/handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/Handler.scala
@@ -1,17 +1,19 @@
 package com.gu.identityBackfill
 
-import java.io.{ InputStream, OutputStream }
+import java.io.{InputStream, OutputStream}
 
 import com.amazonaws.services.lambda.runtime.Context
 import com.gu.effects.RawEffects
-import com.gu.identity.{ GetByEmail, IdentityConfig }
+import com.gu.identity.{GetByEmail, IdentityConfig}
 import com.gu.identityBackfill.Types._
+import com.gu.identityBackfill.zuora.{CountZuoraAccountsForIdentityId, GetZuoraAccountsForEmail}
 import com.gu.util.Config
 import com.gu.util.apigateway.ApiGatewayHandler.LambdaIO
-import com.gu.util.apigateway.{ ApiGatewayHandler, ApiGatewayRequest, ApiGatewayResponse }
+import com.gu.util.apigateway.{ApiGatewayHandler, ApiGatewayRequest, ApiGatewayResponse}
 import com.gu.util.reader.Types.FailableOp
-import com.gu.util.zuora.ZuoraRestConfig
-import play.api.libs.json.{ Json, Reads }
+import com.gu.util.zuora.{ZuoraDeps, ZuoraRestConfig}
+import play.api.libs.json.{Json, Reads}
+
 import scalaz.syntax.either._
 
 object Handler {
@@ -29,25 +31,14 @@ object Handler {
     def operation: Config[StepsConfig] => ApiGatewayRequest => FailableOp[Unit] =
       config => IdentityBackfillSteps(
         GetByEmail(rawEffects.response, config.stepsConfig.identityConfig),
-        GetZuoraAccountsForEmail.apply,
-        CountZuoraAccountsForIdentityId.apply,
+        GetZuoraAccountsForEmail.apply(ZuoraDeps(rawEffects.response, config.stepsConfig.zuoraRestConfig)),
+        CountZuoraAccountsForIdentityId.apply(ZuoraDeps(rawEffects.response, config.stepsConfig.zuoraRestConfig)),
         UpdateZuoraIdentityId.apply,
-        UpdateSalesforceIdentityId.apply)
+        UpdateSalesforceIdentityId.apply
+      )
     ApiGatewayHandler.default[StepsConfig](implicitly)(operation, lambdaIO).run((rawEffects.stage, rawEffects.s3Load(rawEffects.stage)))
   }
 
-}
-
-object GetZuoraAccountsForEmail {
-  def apply(emailAddress: EmailAddress): FailableOp[List[ZuoraAccountIdentitySFContact]] = {
-    ApiGatewayResponse.internalServerError("todo").left
-  }
-}
-
-object CountZuoraAccountsForIdentityId {
-  def apply(identityId: IdentityId): FailableOp[Int] = {
-    ApiGatewayResponse.internalServerError("todo").left
-  }
 }
 
 object UpdateZuoraIdentityId {
--- a/handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/IdentityBackfillSteps.scala
+++ b/handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/IdentityBackfillSteps.scala
@@ -4,11 +4,11 @@ import com.gu.identity.GetByEmail
 import com.gu.identityBackfill.IdentityBackfillSteps.WireModel.IdentityBackfillRequest
 import com.gu.identityBackfill.Types._
 import com.gu.util.Logging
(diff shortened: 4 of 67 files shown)


build.sbt
```diff
@@ -22,7 +22,14 @@ val scalaSettings = Seq(
...
   javaOptions in Test += s"""-Dlog4j.configuration=file:${new File(".").getCanonicalPath}/test_log4j.properties""",
-  fork in Test := true
+  fork in Test := true,
+  {
+    import scalariform.formatter.preferences._
+    scalariformPreferences := scalariformPreferences.value
+      .setPreference(DanglingCloseParenthesis, Force)
+      .setPreference(SpacesAroundMultiImports, false)
+      .setPreference(NewlineAtEndOfFile, true)
+  }
```
johnduffell: I decided to stop the formatter fighting with intellij, also stopped it moving the trailing brackets to the end of the previous line.  feel free to disagree (I'd rather have a standard I don't like than none at all)
the new line at the end of the file just stops github having those red no entry signs at the bottom of the file

handlers/identity-backfill/src/main/scala/com/gu/identity/GetByEmail.scala
```diff
@@ -1,11 +1,11 @@
 package com.gu.identity
 
-import com.gu.identity.GetByEmail.RawWireModel.{ User, UserResponse }
-import com.gu.identityBackfill.Types.{ EmailAddress, IdentityId }
-import okhttp3.{ HttpUrl, Request, Response }
-import play.api.libs.json.{ Json, Reads }
+import com.gu.identity.GetByEmail.RawWireModel.{User, UserResponse}
```
johnduffell: unfortunately, due to the above change to the formatter, it's gone through all the files I've changed, changing them back again.

handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/Handler.scala
```diff
@@ -29,27 +31,16 @@ object Handler {
     def operation: Config[StepsConfig] => ApiGatewayRequest => FailableOp[Unit] =
       config => IdentityBackfillSteps(
         GetByEmail(rawEffects.response, config.stepsConfig.identityConfig),
-        GetZuoraAccountsForEmail.apply,
-        CountZuoraAccountsForIdentityId.apply,
+        GetZuoraAccountsForEmail.apply(ZuoraDeps(rawEffects.response, config.stepsConfig.zuoraRestConfig)),
+        CountZuoraAccountsForIdentityId.apply(ZuoraDeps(rawEffects.response, config.stepsConfig.zuoraRestConfig)),
```
johnduffell: the handler is "pre-wiring" the dependencies, to allow us to unit test them cleanly

handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/zuora/CountZuoraAccountsForIdentityId.scala
```diff
@@ -0,0 +1,30 @@
...
+    implicit val zaReads = Json.reads[ZuoraAccount]
+
+  }
+
+  def apply(zuoraDeps: ZuoraDeps)(identityId: IdentityId): FailableOp[Int] = {
+    val accounts = for {
+      accountsWithEmail <- ListT(ZuoraQuery.query[ZuoraAccount](ZuoraQuery.Query(s"SELECT Id FROM Account where IdentityId__c='${identityId.value}'")).map(_.records))
+    } yield AccountId(accountsWithEmail.Id)
+
+    accounts.run.run.run(zuoraDeps).bimap(e => ApiGatewayResponse.internalServerError(e.message), l => l.size)
```
johnduffell: I think we should get rid of the Reader some point in future, as it's not buying us a lot and here we have listT of eithert of reader which causes a buildup of `run`s when unpacking it.
We can use the "pre-wiring" technique I used above to get the dependencies in a neater? way.

handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/zuora/GetZuoraAccountsForEmail.scala
```diff
@@ -0,0 +1,40 @@
...
+    } yield ZuoraAccountIdentitySFContact(
+      AccountId(accountsWithEmail.Id),
+      IdentityId(accountsWithEmail.IdentityId__c),
+      SFContactId(accountsWithEmail.sfContactId__c)
+    )
+
+    accounts.run.run.run(zuoraDeps).leftMap(e => ApiGatewayResponse.internalServerError(e.message))
+  }
+
+}
```
johnduffell: I feel like this is quite a neat structure for accessing zuora.  The exact input and output are specific the the operation, so can be left here.  The actual code for getting in touch qith zuora functionality (queries and get/post) are pretty common, so are in the zuora package.
This stops a proliferation of ZuoraService with a load of unrelated operations - here each object is one operation with its own data structures, and returns the response in a way that makes sense in the domain of the code that's calling it.

lib/effects/src/test/scala/com/gu/effects/TestingRawEffects.scala
```diff
@@ -27,8 +32,19 @@ class TestingRawEffects(val isProd: Boolean = false, val defaultCode: Int = 1, r
...
+          responses.get(path) // todo should change to have GETs and POSTs in one list using a Sum type, and have a getUnusedRequests or something at the end to assert they were "used" by the code
+        } else if (req.method() == "POST") {
+          val reqBody = body(req.body)
+          logger.info(s"HTTP POST body is $reqBody")
+          postResponses.get(POSTRequest(path, reqBody)).orElse(responses.get(path)) // use get response
+        } else responses.get(path)
+      val HTTPResponse(code, response) = aaa.getOrElse(
+        HTTPResponse(defaultCode, """{"success": true}""")
+      )
+      logger.info(s"HTTP precanned response is: $code - $response")
```
johnduffell: I had to add support in this area for POST requests, as all queries to zuora are post, so we need to be able to mock them.  This solution is a hack for now, ideally it should be all in the same Map where the key is some kind of MockRequest which would be GET/POST/whatever.

lib/handler/src/main/scala/com/gu/util/Logging.scala
```diff
@@ -22,20 +20,12 @@ trait Logging {
 
   }
 
-  implicit class LogImplicit2[E, A](failableOp: E \/ A) {
+  protected implicit class LogImplicit2[A](op: A) {
```
johnduffell: because we "Extend" Logging, these start appearing here there and everywhere unexpectedly.  I have made them protected to stop that (object oriented++)

lib/zuora/src/main/scala/com/gu/util/zuora/ZuoraRestRequestMaker.scala
```diff
@@ -40,7 +39,7 @@ object ZuoraRestRequestMaker extends Logging {
       val body = response.body.string
       val truncated = body.take(500) + (if (body.length > 500) "..." else "")
       logger.error(s"Request to Zuora was unsuccessful, the response was: \n $response\n$truncated")
-      internalServerError("Request to Zuora was unsuccessful").left
+      ClientFail("Request to Zuora was unsuccessful").left
```
johnduffell: zuora shouldn't really be talking in terms of internal server error, even if under the covers it's just returning its own case class.

lib/zuora/src/main/scala/com/gu/util/zuora/internal/Types.scala
```diff
@@ -22,14 +22,4 @@ object Types {
 
 }
 
-case class ClientFail(statusCode: String, message: String)
-object ClientFail extends Logging {
```
johnduffell: this companion object was what was pulling in the implicit Logging classes and sharing them with any code that used a ClientFail in any way.

src/main/scala/com/gu/autoCancel/AutoCancel.scala
```diff
@@ -1,11 +1,12 @@
 package com.gu.autoCancel
```
johnduffell: Everything after this point is just reformatting the code - this happened automatically with scalariform (sorry!!) 🙈

handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/IdentityBackfillSteps.scala
```diff
@@ -30,16 +31,16 @@ object IdentityBackfillSteps extends Logging {
...
       identityId <- getByEmail(emailAddress).leftMap(a => ApiGatewayResponse.internalServerError(a.toString)).withLogging("GetByEmail")
-      _ <- if (request.dryRun) -\/(ApiGatewayResponse.noActionRequired("DRY RUN requested! skipping to the end")) else \/-(()) // FIXME this will be removed once the later steps actually do something. it was just needed for testing
       zuoraAccountsForEmail <- getZuoraAccountsForEmail(emailAddress)
-      zuoraAccountForEmail <- zuoraAccountsForEmail match { case one :: Nil => \/-(one); case _ => -\/(ApiGatewayResponse.internalServerError("should have exactly one zuora account per email at this stage")) }
+      zuoraAccountForEmail <- zuoraAccountsForEmail match { case one :: Nil => \/-(one); case _ => -\/(ApiGatewayResponse.notFound("should have exactly one zuora account per email at this stage")) }
       zuoraAccountsForIdentityId <- countZuoraAccountsForIdentityId(identityId)
-      _ <- if (zuoraAccountsForIdentityId > 0) \/-(()) else -\/(ApiGatewayResponse.internalServerError("already used that identity id"))
+      _ <- if (zuoraAccountsForIdentityId == 0) \/-(()) else -\/(ApiGatewayResponse.notFound("already used that identity id"))
```
pvighi: Is "not found" the correct status for pre check failures?

handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/zuora/GetZuoraAccountsForEmail.scala
```diff
@@ -0,0 +1,40 @@
...
+      sfContactId__c: String
+    )
+    implicit val zaReads = Json.reads[ZuoraAccount]
+
+  }
+
+  def apply(zuoraDeps: ZuoraDeps)(emailAddress: EmailAddress): FailableOp[List[ZuoraAccountIdentitySFContact]] = {
+    val accounts = for {
+      contactWithEmail <- ListT(ZuoraQuery.query[ZuoraContact](ZuoraQuery.Query(s"SELECT Id FROM Contact where WorkEmail='${emailAddress.value}'")).map(_.records))
+      accountsWithEmail <- ListT(ZuoraQuery.query[ZuoraAccount](ZuoraQuery.Query(s"SELECT Id, IdentityId__c, sfContactId__c FROM Account where BillToId='${contactWithEmail.Id}'")).map(_.records))
```
pvighi: I think it might be a little confusing having both a query def and a query class in the ZuoraQuery object

handlers/identity-backfill/src/test/scala/com/gu/identityBackfill/StepsTest.scala
```diff
@@ -0,0 +1,100 @@
...
+import org.scalatest.{FlatSpec, Matchers}
+
+import scalaz.{-\/, \/-}
+
+class StepsTest extends FlatSpec with Matchers {
+
+  class StepsWithMocks {
+
+    var zuoraUpdate: Option[(AccountId, IdentityId)] = None // !!
+    var salesforceUpdate: Option[(SFContactId, IdentityId)] = None // !!
```
pvighi: 😱

lib/effects/src/test/scala/com/gu/effects/TestingRawEffects.scala
```diff
@@ -27,8 +32,19 @@ class TestingRawEffects(val isProd: Boolean = false, val defaultCode: Int = 1, r
     req =>
       requests = req :: requests
       val path = getPathWithQueryString(req)
-      val (code, response) = responses.getOrElse(path, (defaultCode, """{"success": true}"""))
-      logger.info(s"request for: $path so returning $response")
+      logger.info(s"HTTP ${req.method} request for: $path")
+      val aaa =
```
pvighi: is there a better name for this than aaa ?

handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/IdentityBackfillSteps.scala
johnduffell: I wanted it to be a 4xx error and I wasn't sure what.  It's not really a bad request or unauthorised, it's just a "no, sorry".  That was about the closes I could think of!

handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/zuora/GetZuoraAccountsForEmail.scala
johnduffell: good point (that's old code, so well spotted)  I'll fix

handlers/identity-backfill/src/test/scala/com/gu/identityBackfill/StepsTest.scala
johnduffell: 🙈

johnduffell: another (more FP safe, but less understandable) solution would be to use a Writer monad throughout everything to collect up the responses, or to have a parameterised monad ("tagless final")  otherwise it's difficult to collect up the side effects.

lib/effects/src/test/scala/com/gu/effects/TestingRawEffects.scala
johnduffell: yes 😆

handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/zuora/GetZuoraAccountsForEmail.scala
johnduffell:
```scala
      contactWithEmail <- {
        val contactQuery = Query(s"SELECT Id FROM Contact where WorkEmail='${emailAddress.value}'")
        ListT(ZuoraQuery.getResults[ZuoraContact](contactQuery).map(_.records))
      }
      accountsWithEmail <- {
        val accountQuery = Query(s"SELECT Id, IdentityId__c, sfContactId__c FROM Account where BillToId='${contactWithEmail.Id}'")
        ListT(ZuoraQuery.getResults[ZuoraAccount](accountQuery).map(_.records))
      }
```

lib/effects/src/test/scala/com/gu/effects/TestingRawEffects.scala
johnduffell: `presetResponse`

johnduffell:
tested on CODE and it looks OK! shipping.... thanks for the comments @pvighi

PR #52, opened 2018-03-21
Health checks for lambdas

johnduffell:
I am deploying the lambdas quite often to CODE and PROD.  It's quite hard to know if I broke something unrelated.  Deployment is just copying a JAR file and we only find out if there's a problem when an alarm goes off.

This PR adds another field to the steps/operation object to define a health check operation.  This is called when you hit the lambda's url with isHealthcheck=true in the URL.  You don't need any api token as enforced by the lambda, but you would need the api key for API gateway.

I have added an example health check for identity backfill, which calls identity and zuora to make sure it can do a lookup in each.

A third aspect is a local test which can run the healthcheck against the DEV config.
Also there is a CODE/PROD local test which hits the configured urls in the healthcheck local config file, and makes sure things are working in CODE or PROD.

@jacobwinch will be happy about this quality improvement? ;)

@pvighi @paulbrown1982 @lmath

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,8 @@ This is the reader revenue lambda API/orchestration layer.  But it has a bad nam
 Please keep all the various README in this project up to date, and improve them!
 There should be one in each project and anywhere else you think it's would help.
 
+---
+
 ## philosophy
 The general philosophy of this project is to keep things under review and don't be afraid to refactor things
 especially the hard parts of the structure when you think they are wrong.
@@ -28,7 +30,29 @@ If there's any code in either that you feel could be unit tested, it should prob
 - **one jar per lambda** - minimise the size of the deployment artifact
 - **minimise dependencies (aka liabilities)** on external libraries as we have to keep them up to date, also they increase the size of the artifact
 
-# Howto update config
+---
+
+## Quality matters
+The testing story is not perfect at the moment!  You are responsible for getting your own changes into production safely.  However there are a few common situations you will encounter:
+
+### code only changes
+You can be confident when making code only changes.  Run `sbt test`. They will run on your build of your branch and when you merge to master.  Don't forget to "Run with Coverage" in IntelliJ.
+
+### changes to config
+The system won't safe you for free here.  Once you changed your config (see the howto below), you should run ConfigLoaderSystemTest.  This test does not (YET!) run automatically, so if you deploy to PROD without checking this, the lambdas will deploy but not actually work.
+
+### system integration points
+Traiditonal health checking doesn't happen.  After deploy you can't relax until you know an HTTP request will hit your lambda, and that your lambda can access any external services.
+You can run the health check as a local test by downloading the DEV config, and running HealthCheckSystemTest.
+You can run it in CODE and PROD by downloading the health check config from DEV and running CODEPRODHealthCheck.
+The CODE and PROD health checks are called by RunScope every 5 minutes.
+
+To download the dev config use the following command:
+`aws s3 cp s3://gu-reader-revenue-private/membership/payment-failure-lambdas/DEV/ /etc/gu/ --exclude "*" --include "payment-failure-*" --profile membership --recursive`
+
+---
+
+## Howto update config
 At the moment we just use S3 for config.
 
 If you're making an addition, you can just copy the file from S3 for PROD and CODE, then update and upload.
@@ -48,11 +72,18 @@ That will check the latest version can be understood by the current version of t
 
 Ideally this test should be automated and your PR shouldn't be mergable until the config is readable.
 
+---
+
 ## structure
 The main project aggregates all the sub projects from handlers and lib, so we can build and test them in one go.
 
 ## root
-Contains three Scala lambdas behind the same API gateway.  TODO These should be moved into a new set of 3 projects in the handlers folder.
+Contains three Scala lambdas behind the same API gateway.
+At present these can be deployed to CODE and PROD as MemSub::Membership Admin::Zuora Auto Cancel.
+
+Testing: to get all the emails from ET, run the EmailClientSystemTest against your own email address.  You should get a deluge of emails.
+
+TODO These three lambdas should be moved into a new set of 3 projects in the handlers folder.
 
 **autoCancel**: 
 Used to cancel subscriptions with overdue invoices, based on an event trigger within Zuora.
@@ -74,10 +105,12 @@ Stripe works with card networks so that when a customer's card details are updat
 The full workflow is currently:
 Stripe Webhook > AWS Cloudfront > AWS API Gateway > AWS Lambda
 
-*An additional CloudFront distribution is currently required because callouts do not support SNI, and the default CloudFront distribution (which gets set up in front of API Gateway) seems to require it.
+*An additional CloudFront distribution is currently required because zuora callouts do not support SNI, and the default CloudFront distribution (which gets set up in front of API Gateway) seems to require it.
 
-# Testing
+**Identity backfill**
 
-- Run `sbt test` to execute the unit tests
-- If you need to validate that API Gateway can trigger Lambda execution successfully, deploy your changes to CODE and use Postman (or equivalent) to hit the API Gateway url with a valid payload.
-- For a full integration test of the autoCancel or paymentFailure lambdas, you can trigger the callout from our UAT Zuora environment.
+[identity-backfill](handlers/)
+
+**Libraries**
+
+[shared code](lib/)
--- a/README.md
+++ b/README.md
@@ -38,17 +38,17 @@ The testing story is not perfect at the moment!  You are responsible for getting
 ### code only changes
 You can be confident when making code only changes.  Run `sbt test`. They will run on your build of your branch and when you merge to master.  Don't forget to "Run with Coverage" in IntelliJ.
 
-### changes to config
-The system won't safe you for free here.  Once you changed your config (see the howto below), you should run ConfigLoaderSystemTest.  This test does not (YET!) run automatically, so if you deploy to PROD without checking this, the lambdas will deploy but not actually work.
-
 ### system integration points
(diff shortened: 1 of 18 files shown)


handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/IdentityBackfillSteps.scala
```diff
@@ -32,20 +33,33 @@ object IdentityBackfillSteps extends Logging {
...
+        _ <- updateZuoraIdentityId(zuoraAccountForEmail.accountId, identityId)
+        _ <- updateSalesforceIdentityId(zuoraAccountForEmail.sfContactId, identityId)
+        // need to remember which ones we updated?
+      } yield ()
+
+    def healthcheck() =
+      for {
+        identityId <- getByEmail(EmailAddress("john.duffell@guardian.co.uk")).leftMap(a => ApiGatewayResponse.internalServerError(a.toString)).withLogging("healthcheck getByEmail")
+        _ <- countZuoraAccountsForIdentityId(identityId)
+      } yield ()
```
johnduffell: it's not a complicated health check, but it should be better than nothing!

handlers/identity-backfill/src/test/scala/manualTest/HealthCheckSystemTest.scala
```diff
@@ -0,0 +1,114 @@
+package manualTest
```
johnduffell: I'm a bit nervous about the number of important manual tests I've added, that you have to know something in order to use them.
I'm wondering whether we need another sbt task, where it runs all the manual tests using config from S3.  We would still have to make it prominent in the README.  This wouldn't exactly help when we want to do the "health check" for CODE or PROD, because they would be more appropriate to do after deployment.  Maybe the answer is another lambda that healthchecks everything repeatedly and alarms if any fail?

lib/handler/src/main/scala/com/gu/util/apigateway/ApiGatewayHandler.scala
```diff
@@ -38,12 +38,18 @@ object ApiGatewayHandler extends Logging {
     if (credentialsAreValid(requestAuth, trustedApiConfig)) \/-(()) else -\/(unauthorized)
   }
 
-  def default[StepsConfig: Reads](operation: Config[StepsConfig] => ApiGatewayRequest => FailableOp[Unit], io: LambdaIO): Reader[(Stage, Try[String]), Unit] =
+  case class Operation(steps: ApiGatewayRequest => FailableOp[Unit], healthcheck: () => FailableOp[Unit])
+  object Operation {
+    def noHealthcheck(steps: ApiGatewayRequest => FailableOp[Unit]) =
+      Operation(steps, () => \/-(()))
```
johnduffell: this is more of a legacy thing, the health check always passes assuming the jvm could start up and the config could be read.

lib/handler/src/main/scala/com/gu/util/apigateway/ApiGatewayRequest.scala
```diff
@@ -17,7 +17,12 @@ object StripeAccount {
 
   implicit val reads: Reads[StripeAccount] = JsPath.read[String].map(fromString(_).get)
 }
-case class URLParams(apiToken: Option[String], onlyCancelDirectDebit: Boolean, stripeAccount: Option[StripeAccount])
+case class URLParams(
+  apiToken: Option[String],
+  onlyCancelDirectDebit: Boolean,
+  stripeAccount: Option[StripeAccount],
+  isHealthcheck: Boolean
+)
```
johnduffell: this class is clearly not right, in future it will need rejigging

handlers/identity-backfill/src/test/scala/manualTest/HealthCheckSystemTest.scala
johnduffell: @pvighi suggested runscope might work well for this

jacobwinch: Yes, I think that's a good suggestion - it will monitor and check that we have not fundamentally broken the lambda (i.e. at least we'll know that AWS can load the .jar and find a handler function), which is better than we have now. 

It may also have the added side-affect/benefit of keeping the lambdas warm(ish).

src/test/scala/manualTest/CODEPRODHealthCheck.scala
```diff
@@ -0,0 +1,128 @@
+package manualTest
```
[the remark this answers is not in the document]
jacobwinch: I would suggest that we configure this in Runscope (or similar), and then we can remove this test.

src/test/scala/manualTest/CODEPRODHealthCheck.scala
```diff
@@ -0,0 +1,128 @@
...
+  it should "successfull run the health checks against PROD" in {
+
+    healthcheckForEnv(_.PROD)
+
+  }
+
+  private def healthcheckForEnv(env: HealthChecks => List[HealthCheckConfig]) = {
+    val healthchecks = for {
+      jsonString <- Try {
+        Source.fromFile("/etc/gu/payment-failure-healthcheck.private.json").mkString
```
pvighi: is this something we need to download from the private bucket ? maybe we could mention it in the readme

handlers/identity-backfill/src/test/scala/manualTest/HealthCheckSystemTest.scala
johnduffell: after a chat with Jacob, this might be possible with prout to check when the deploy has happened, then kick off a travis build to health check everything and post it back to the build as a post deploy testing stage

src/test/scala/manualTest/CODEPRODHealthCheck.scala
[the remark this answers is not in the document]
johnduffell: I have added it in runscope in CODE, needs also enabling in PROD once it's shipped
should we really be using runscope to test CODE? or just PROD.  if so how do we test CODE is good.

src/test/scala/manualTest/CODEPRODHealthCheck.scala
johnduffell: good point, I haven't done any readme about this

src/test/scala/manualTest/CODEPRODHealthCheck.scala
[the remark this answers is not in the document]
pvighi: do we need to test CODE ?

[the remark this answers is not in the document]
johnduffell: after you deployed to code, you want to be told that all the integration points are working ok.  That is the purpose of the health check.  Otherwise you might not find out until you get to prod.
However, you don't want the whole team alerted about an issue in CODE, it only should go to the PR or the person who did the deploy.
So I don't think it should be in runscope, more like a special sbt/travis build kicked off from a code deploy that pushes to the PR?

src/test/scala/manualTest/CODEPRODHealthCheck.scala
johnduffell: ok there is plenty of readme now, also the files you need are in S3.  would be worth making it read directly from S3 if you don't have the files locally, then it won't get stale.  ToDo...

src/test/scala/manualTest/CODEPRODHealthCheck.scala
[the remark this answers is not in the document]
johnduffell: image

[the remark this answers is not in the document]
johnduffell: so based on feedback from @pvighi and @jacobwinch the runscope tests will be prod only, and it will be developers responsibility to be sure it's going to work (by running the healthcheck in CODE first)
This is better than the existing situation, and if we start getting nervous, we can add more safety (MVP style)

handlers/identity-backfill/src/test/scala/manualTest/HealthCheckSystemTest.scala
jacobwinch: There's an example of that here: https://github.com/guardian/support-frontend/blob/master/.prout.json#L6-L12

lib/handler/src/main/scala/com/gu/util/apigateway/ApiGatewayHandler.scala
jacobwinch: I actually think even this sanity check is pretty useful

README.md
```diff
@@ -28,7 +30,29 @@ If there's any code in either that you feel could be unit tested, it should prob
...
+The testing story is not perfect at the moment!  You are responsible for getting your own changes into production safely.  However there are a few common situations you will encounter:
+
+### code only changes
+You can be confident when making code only changes.  Run `sbt test`. They will run on your build of your branch and when you merge to master.  Don't forget to "Run with Coverage" in IntelliJ.
+
+### system integration points
+Traiditonal health checking doesn't happen.  After deploy you can't relax until you know an HTTP request will hit your lambda, and that your lambda can access any external services.
+You can run the health check as a local test by downloading the DEV config, and running HealthCheckSystemTest.
+You can run it in CODE and PROD by downloading the health check config from DEV and running CODEPRODHealthCheck.
+The PROD health checks are called by RunScope every 5 minutes.
```
jacobwinch: Link to our Runscope tests here!

src/test/scala/manualTest/CODEPRODHealthCheck.scala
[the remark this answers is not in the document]
jacobwinch: 👍

johnduffell:
image
working in prod and hopefully notifying us if it fails

PR #54, opened 2018-03-23
Add Add identity id to zuora operation

johnduffell:
This PR adds a new operation to update an identity id in zuora.

I have also added a test that runs it against DEV zuora on a ready made account.  So it updates the id and checks the new value is correct.

In the same PR I have added tags to all the Effectful tests, and made them run on the config downloaded from DEV S3 bucket.  Local config is not used any more.

With AWS credentials, you should be able to run all the tests without trouble.  Without, only the standard ones would would.

Any comments or +1s please @jacobwinch @pvighi @paulbrown1982 @lmath

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -32,8 +32,10 @@ val scalaSettings = Seq(
   }
 )
 
-lazy val EffectsTest = config("effectsTest") extend(Test) describedAs("run the egde tests")
-lazy val HealthCheckTest = config("healthCheck") extend(Test) describedAs("run the health checks agains prod/code")
+// fixme this whole file needs splitting down appropriately
+
+lazy val EffectsTest = config("effectsTest") extend(Test) describedAs("run the edge tests")
+lazy val HealthCheckTest = config("healthCheck") extend(Test) describedAs("run the health checks against prod/code")
 val testSettings = inConfig(EffectsTest)(Defaults.testTasks) ++ inConfig(HealthCheckTest)(Defaults.testTasks) ++ Seq(
   testOptions in Test += Tests.Argument("-l", "com.gu.test.EffectsTest"),
   testOptions in Test += Tests.Argument("-l", "com.gu.test.HealthCheck"),
--- a/build.sbt
+++ b/build.sbt
@@ -32,7 +32,27 @@ val scalaSettings = Seq(
   }
 )
 
-lazy val zuora = (project in file("lib/zuora")).settings(scalaSettings).settings(
+lazy val EffectsTest = config("effectsTest") extend(Test) describedAs("run the egde tests")
+lazy val HealthCheckTest = config("healthCheck") extend(Test) describedAs("run the health checks agains prod/code")
+val testSettings = inConfig(EffectsTest)(Defaults.testTasks) ++ inConfig(HealthCheckTest)(Defaults.testTasks) ++ Seq(
+  testOptions in Test += Tests.Argument("-l", "com.gu.test.EffectsTest"),
+  testOptions in Test += Tests.Argument("-l", "com.gu.test.HealthCheck"),
+
+  //  configs(EffectsTest),
+
+  testOptions in EffectsTest -= Tests.Argument("-l", "com.gu.test.EffectsTest"),
+  testOptions in EffectsTest -= Tests.Argument("-l", "com.gu.test.HealthCheck"),
+  testOptions in EffectsTest += Tests.Argument("-n", "com.gu.test.EffectsTest"),
+  //  configs(HealthCheckTest),
+
+  testOptions in HealthCheckTest -= Tests.Argument("-l", "com.gu.test.EffectsTest"),
+  testOptions in HealthCheckTest -= Tests.Argument("-l", "com.gu.test.HealthCheck"),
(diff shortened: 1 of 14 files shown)


build.sbt
```diff
@@ -32,7 +32,27 @@ val scalaSettings = Seq(
   }
 )
```
johnduffell: if anyone understands sbt, this is only way I could get it working.  I have a feeling it's confused intellij a bit too, as it doesn't alays seem to compile things when I run the tests myself.

handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/zuora/GetIdentityIdForAccount.scala
```diff
@@ -0,0 +1,33 @@
+package com.gu.identityBackfill.zuora
+
+import com.gu.identityBackfill.Types.{AccountId, IdentityId}
+import com.gu.identityBackfill.zuora.GetIdentityIdForAccount.WireModel.ZuoraAccount
+import com.gu.util.apigateway.ApiGatewayResponse
+import com.gu.util.reader.Types.FailableOp
+import com.gu.util.zuora.{ZuoraDeps, ZuoraRestRequestMaker}
+import play.api.libs.json.Json
+
```
johnduffell: this operation is only needed for testing.  not sure if I should have put it in the test package instead?

handlers/identity-backfill/src/test/scala/manualTest/AddIdentityIdEffectsTest.scala
```diff
@@ -0,0 +1,37 @@
...
+import org.scalatest.{FlatSpec, Matchers}
+import scalaz.\/-
+import scalaz.syntax.std.either._
+
+import scala.util.Random
+
+// run this manually
+class AddIdentityIdEffectsTest extends FlatSpec with Matchers {
+
+  it should "successfu update the identity id against dev" taggedAs EffectsTest in {
```
jacobwinch: typo

handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/zuora/GetIdentityIdForAccount.scala
jacobwinch: I think this should live in the test package

build.sbt
```diff
@@ -32,7 +32,27 @@ val scalaSettings = Seq(
...
 
-lazy val zuora = (project in file("lib/zuora")).settings(scalaSettings).settings(
+lazy val EffectsTest = config("effectsTest") extend(Test) describedAs("run the egde tests")
```
jacobwinch: We could use the the standard IntegrationTest config for this (https://www.scala-sbt.org/1.x/docs/Testing.html#Integration+Tests)?

build.sbt
```diff
@@ -32,7 +32,27 @@ val scalaSettings = Seq(
...
+lazy val HealthCheckTest = config("healthCheck") extend(Test) describedAs("run the health checks agains prod/code")
+val testSettings = inConfig(EffectsTest)(Defaults.testTasks) ++ inConfig(HealthCheckTest)(Defaults.testTasks) ++ Seq(
+  testOptions in Test += Tests.Argument("-l", "com.gu.test.EffectsTest"),
+  testOptions in Test += Tests.Argument("-l", "com.gu.test.HealthCheck"),
+
+  //  configs(EffectsTest),
+
+  testOptions in EffectsTest -= Tests.Argument("-l", "com.gu.test.EffectsTest"),
```
jacobwinch: It seems really odd that we have to exclude and then include these tagged tests... if you can't get this working nicely with tags, another option would be to use filters based on package name, as per: https://github.com/guardian/support-frontend/blob/master/build.sbt#L21-L29?

jacobwinch:
Looks good 👍

handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/zuora/GetIdentityIdForAccount.scala
johnduffell: thanks for decisiveness 🥇 😆

handlers/identity-backfill/src/test/scala/manualTest/AddIdentityIdEffectsTest.scala
johnduffell: 🙈

build.sbt
johnduffell: I did think about there may be a better way, this whole setup is a bit of a mess and slightly confuses intellij.  However package name is a bit unexpected so I think tagging is slightly preferable, only if it works as you say!  one for a later PR after some thought, thanks for the suggestion

jacobwinch: 👍

build.sbt
johnduffell: we could but I don't know what "integration test" means to everyone.  It could be test everything together including the integration points, or it could be test only the integration points i.e. not unit tested bits in isolation (this is what it actually does)  Although I would not object to someone making one or two of the former type.

PR #58, opened 2018-03-29
update identity backfill handler to push identity id into SF

johnduffell:
This is part of the code to link existing identity accounts into zuora/sf subs.

This PR adds the final step to add the identity id into salesforce.

I have split the zuora rest request maker into a rest request maker and reused most of the code for the SF one (as they are both using http/json)  I also took that chance to get rid of some of the Reader monad

I can deploy to code successfully and the healthchecks are passing

@paulbrown1982 @lmath comments please

### DIFF
--- a/handlers/identity-backfill/src/test/scala/com/gu/identityBackfill/salesforce/updateSFIdentityId/GetSalesforceIdentityId.scala
+++ b/handlers/identity-backfill/src/test/scala/com/gu/identityBackfill/salesforce/updateSFIdentityId/GetSalesforceIdentityId.scala
@@ -1,0 +1,24 @@
+package com.gu.identityBackfill.salesforce.updateSFIdentityId
+
+import com.gu.identityBackfill.Types.{IdentityId, SFContactId}
+import com.gu.identityBackfill.salesforce.SalesforceAuthenticate.SalesforceAuth
+import com.gu.identityBackfill.salesforce.SalesforceRestRequestMaker
+import com.gu.util.apigateway.ApiGatewayResponse
+import com.gu.util.reader.Types.FailableOp
+import okhttp3.{Request, Response}
+import play.api.libs.json.Json
+
+object GetSalesforceIdentityId {
+
+  case class WireResult(IdentityID__c: String)
+  object WireResult {
+    implicit val reads = Json.reads[WireResult]
+  }
+
+  def apply(response: Request => Response)(salesforceAuth: SalesforceAuth)(sFContactId: SFContactId): FailableOp[IdentityId] = {
+    val get = SalesforceRestRequestMaker(salesforceAuth, response)
+    val id = get.get[WireResult](s"/services/data/v20.0/sobjects/Contact/${sFContactId.value}")
+    id.bimap(e => ApiGatewayResponse.internalServerError(e.message), id => IdentityId(id.IdentityID__c))
+  }
+
+}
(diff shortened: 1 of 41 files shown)


handlers/identity-backfill/src/test/scala/com/gu/identityBackfill/salesforce/updateSFIdentityId/GetSalesforceIdentityId.scala
```diff
@@ -0,0 +1,24 @@
+package com.gu.identityBackfill.salesforce.updateSFIdentityId
```
johnduffell: this is in the test package because we only need to do it to make sure the PATCH call worked.

lib/effects/src/test/scala/com/gu/effects/TestingRawEffects.scala
```diff
@@ -36,11 +36,11 @@ class TestingRawEffects(
       val presetResponse =
         if (req.method() == "GET") {
           responses.get(path) // todo should change to have GETs and POSTs in one list using a Sum type, and have a getUnusedRequests or something at the end to assert they were "used" by the code
-        } else if (req.method() == "POST") {
+        } else {
```
johnduffell: now we support PATCH requests so I made this a parameter

lib/restHttp/src/test/scala/com/gu/util/RestServiceTest.scala
```diff
@@ -0,0 +1,126 @@
...
+import scalaz.{-\/, \/-}
+
+class RestServiceTest extends AsyncFlatSpec {
+
+  case class BasicAccountInfo(id: String)
+  implicit val read = Json.reads[BasicAccountInfo]
+  implicit val readUnit = new Reads[Unit] {
+    override def reads(json: JsValue): JsResult[Unit] = JsSuccess(())
+  }
+
```
johnduffell: these are tested in the "run success end to end GET" but maybe we should be testing them separately too

lib/handler/src/main/scala/com/gu/util/Logging.scala
```diff
@@ -1,33 +1,9 @@
 package com.gu.util
 
-import com.gu.util.reader.Types.{WithDepsFailableOp, _}
 import org.apache.log4j.Logger
 
 trait Logging {
 
   val logger = Logger.getLogger(getClass.getName)
 
-  protected implicit class LogImplicit[A, D](configHttpFailableOp: WithDepsFailableOp[D, A]) {
```
johnduffell: moved because this was causing a cyclical dependency.  this logging class shouldn't be containing anything extra

lib/zuora/src/main/scala/com/gu/util/zuora/ZuoraQuery.scala
```diff
@@ -31,6 +32,6 @@ object ZuoraQuery {
 
   // https://www.zuora.com/developer/api-reference/#operation/Action_POSTquery
   def getResults[QUERYRECORD: Reads](query: Query): WithDepsClientFailableOp[ZuoraDeps, QueryResult[QUERYRECORD]] =
-    post(query, s"action/query")
+    Reader { zuoraDeps: ZuoraDeps => ZuoraRestRequestMaker(zuoraDeps).post(query, s"action/query", true): ClientFailableOp[QueryResult[QUERYRECORD]] }.toEitherT
```
johnduffell: really we should get rid of the readers everywhere, but I didn't want to go too far yet

PR #59, opened 2018-03-29
return friendly message if identity account not round

johnduffell:
when identity backfilling, the first check is does that email user have an identity account.

At present, the lambda returns a 500 error containing `Failed to process event due to the following error: ApiError(failed http with 404)`

This turns it into a 404 explaining "user doesn't have identity"
image

@paulbrown1982 @lmath just a small fix

### DIFF
--- a/handlers/identity-backfill/src/test/scala/com/gu/identityBackfill/StepsTest.scala
+++ b/handlers/identity-backfill/src/test/scala/com/gu/identityBackfill/StepsTest.scala
@@ -1,5 +1,6 @@
 package com.gu.identityBackfill
 
+import com.gu.identity.GetByEmail.NotFound
 import com.gu.identityBackfill.StepsData._
 import com.gu.identityBackfill.Types.{IdentityId, _}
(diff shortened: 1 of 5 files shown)


handlers/identity-backfill/src/test/scala/com/gu/identityBackfill/StepsTest.scala
```diff
@@ -109,6 +111,20 @@ class StepsTest extends FlatSpec with Matchers {
...
+
+    val result =
+      getSteps(emailHasIdentity = false)(ApiGatewayRequest(None, identityBackfillRequest(false), None))
+
+    val expectedResult = -\/(ApiGatewayResponse.notFound("user doesn't have identity"))
+    result should be(expectedResult)
+    zuoraUpdate should be(None)
+    salesforceUpdate should be(None)
+  }
+
```
johnduffell: I didn't add it to the end to end component test or the GetByEmail effectsTest as I thought only needed to test the mapping of response code to api error and api error to api gateway response as unit tests.  This is not essential behaviour so it doesn't make as much sense to test in the more expensive tests.

PR #63, opened 2018-04-09
library updates based on snyk scan

johnduffell:
I spotted snyk is complaining about some libraries.  So I thought I'd bump the versions.

There seem to be no binary incompatibilities which is a good thing.. 

@pvighi @paulbrown1982 @lmath


johnduffell:
this seems to healthcheck ok on CODE! 🤞

PR #64, opened 2018-04-09
Digital subscription expiry lambda

pvighi:
This is just the general wiring and the emergency tokens functionality. A future PR will add the missing code to return the expiry dates based on Zuora Subscriptions.

Note that this lambda is not exposed through an api gateway yet.
For that we would have to open a PR in [Digital subscription Authorisation](https://github.com/guardian/digital-subscription-authorisation) to import this new lambda into its API Gateway


handlers/digital-subscription-expiry/src/main/scala/com/gu/digitalSubscriptionExpiry/DigitalSubscriptionExpirySteps.scala
```diff
@@ -0,0 +1,118 @@
...
+import play.api.libs.json.{JsValue, Json}
+import com.gu.digitalSubscriptionExpiry.emergencyToken.EmergencyTokens
+
+import scala.util.{Success, Try}
+import scalaz.{-\/}
+import scalaz.std.option.optionSyntax._
+
+object DigitalSubscriptionExpirySteps extends Logging {
+
+  def getZuoraExpiry(): Option[DigitalSubscriptionExpiryResponse] = {
```
pvighi: This example code will be replaced with the actual implementation in a future PR (it had to be commented out to make the existing tests pass)

handlers/digital-subscription-expiry/src/main/scala/com/gu/digitalSubscriptionExpiry/DigitalSubscriptionExpirySteps.scala
```diff
@@ -0,0 +1,118 @@
...
+        expiryResponse <- (getEmergencyTokenExpiry(expiryRequest.subscriberId, emergencyTokens) orElse getZuoraExpiry()).toRightDisjunction(notFoundResponse)
+      } yield {
+        val responseJson = Json.toJson(expiryResponse)
+        ApiResponse("200", new Headers, Json.prettyPrint(responseJson))
+      }
+      val response = successfulOrErrorResponse.valueOr(identity)
+      -\/(response)
+    }
+
+    //TODO
```
pvighi: There is no need for a health check in this version but we might add it when we add a dependency on zuora

handlers/digital-subscription-expiry/src/main/scala/com/gu/digitalSubscriptionExpiry/DigitalSubscriptionExpiryResponse.scala
```diff
@@ -0,0 +1,42 @@
...
+
+case class AuthResponse(expiryDate: LocalDate)
+
+//TODO do we need these? also do we need to return all the hardcoded values that cas returns?
+object ExpiryType {
+  val SUB = "sub"
+  val FREE = "free"
+  val SPECIAL = "special" // deprecated
+  val DEFAULT = "default"
+  val DEVICE_CONFIGURED = "deviceConfigured"
```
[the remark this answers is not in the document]
johnduffell: I wanted codacy for the code coverage facilities, but this kind of stuff comes for free (but needs refinement clearly!)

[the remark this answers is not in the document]
johnduffell: I have turned this one off for now, unless people prefer it to be on?  it will also flag up fields in case classes that are for reprsenting JSON, which will be annoying

handlers/digital-subscription-expiry/src/main/scala/com/gu/digitalSubscriptionExpiry/DigitalSubscriptionExpirySteps.scala
```diff
@@ -0,0 +1,118 @@
...
+    //      expiryDate = expiryValue,
+    //      expiryType = ExpiryType.SUB,
+    //      subscriptionCode = None,
+    //      provider = Some("test provider")
+    //    )))
+    None
+  }
+
+  def getEmergencyTokenExpiry(subscriberId: String, emergencyTokens: EmergencyTokens): Option[DigitalSubscriptionExpiryResponse] = {
+    import com.gu.digitalSubscriptionExpiry.emergencyToken.TokenPayloadOps._
```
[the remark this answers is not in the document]
johnduffell: I've turned off this rule, probably not so useful (although if we feel the need to scope things down like this, maybe a separate file would be more useful)

handlers/digital-subscription-expiry/src/main/scala/com/gu/digitalSubscriptionExpiry/zuora/GetSubscription.scala
```diff
@@ -0,0 +1,55 @@
...
+      (__ \ "id").read[String].map(SubscriptionName.apply) and
+      (__ \ "ActivationDate__c").readNullable[DateTime](lenientDateTimeReader) and
+      (__ \ "customerAcceptanceDate").readNullable[DateTime](lenientDateTimeReader) and
+      (__ \ "termStartDate").readNullable[DateTime](lenientDateTimeReader) and
+      (__ \ "termEndDate").readNullable[DateTime](lenientDateTimeReader) and
+      (__ \ "ratePlans").read[List[RatePlan]]
+    )(SubscriptionResult.apply _)
+
+  def apply(request: SubscriptionId): WithDepsFailableOp[ZuoraDeps, SubscriptionResult] =
+    Reader { zuoraDeps: ZuoraDeps => ZuoraRestRequestMaker(zuoraDeps).get[SubscriptionResult](s"object/subscriptions/${request.get}").leftMap(fromClientFail) }.toEitherT
```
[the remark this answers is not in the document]
johnduffell: I think it's just got confused!

handlers/digital-subscription-expiry/src/main/scala/com/gu/digitalSubscriptionExpiry/DigitalSubscriptionExpirySteps.scala
```diff
@@ -0,0 +1,117 @@
...
+        jsonRequest <- parseJson(apiGatewayRequest.body).toRightDisjunction(badRequest)
+        expiryRequest <- Json.fromJson[DigitalSubscriptionExpiryRequest](jsonRequest).asOpt.toRightDisjunction(badRequest)
+        expiryResponse <- (getEmergencyTokenExpiry(expiryRequest.subscriberId, emergencyTokens) orElse getZuoraExpiry()).toRightDisjunction(notFoundResponse)
+      } yield {
+        val responseJson = Json.toJson(expiryResponse)
+        ApiResponse("200", new Headers, Json.prettyPrint(responseJson))
+      }
+      val response = successfulOrErrorResponse.valueOr(identity)
+      -\/(response)
+    }
```
johnduffell: you can simplify? this from
```scala
      } yield {
        val responseJson = Json.toJson(expiryResponse)
        ApiResponse("200", new Headers, Json.prettyPrint(responseJson))
      }
      val response = successfulOrErrorResponse.valueOr(identity)
      -\/(response)
```
to
```scala
      _ <- {
        val responseJson = Json.toJson(expiryResponse)
        -\/(ApiResponse("200", new Headers, Json.prettyPrint(responseJson)))
      }
      } yield ()
```

handlers/digital-subscription-expiry/src/main/scala/com/gu/digitalSubscriptionExpiry/DigitalSubscriptionExpirySteps.scala
```diff
@@ -0,0 +1,117 @@
...
+    val notFoundBody =
+      """
+        |{
+        |    "error": {
+        |        "message": "Unknown subscriber",
+        |        "code": -90
+        |    }
+        |}
+      """.stripMargin
+    ApiGatewayResponse.notFound(notFoundBody)
```
johnduffell: I'd be tempted to have a case class ErrorResponse with a field error: ErrorObject which is a case class with a message: String and a code: Integer, then just collapse these down to
```scala
ApiGatewayResponse.notFound(ErrorResponse(ErrorObject("Unknown subscriber", -90)))
```

handlers/digital-subscription-expiry/src/main/scala/com/gu/digitalSubscriptionExpiry/DigitalSubscriptionExpirySteps.scala
pvighi: We decided not to do that as it introduces a warning and we might change the return type of the steps anyway

handlers/digital-subscription-expiry/src/main/scala/com/gu/digitalSubscriptionExpiry/DigitalSubscriptionResponse.scala
```diff
@@ -0,0 +1,65 @@
...
+  implicit val codeWrites = new Writes[SubscriptionCode] {
+    override def writes(code: SubscriptionCode) = JsString(code.toString)
+  }
+  implicit val dateTimeWriter: Writes[DateTime] = JodaWrites.jodaDateWrites("yyyy-MM-dd")
+  implicit val ExpiryWrites = Json.writes[Expiry]
+}
+
+case class AuthResponse(expiryDate: LocalDate)
+
+//TODO do we need these? also do we need to return all the hardcoded values that cas returns?
```
pvighi: I suspect most of these are not being used. We should remove the ones that are still unused after we add the zuora responses

paulbrown1982:
👍 Looks great!

handlers/digital-subscription-expiry/src/main/scala/com/gu/digitalSubscriptionExpiry/zuora/GetSubscription.scala
```diff
@@ -0,0 +1,54 @@
...
+  implicit val reads: Reads[SubscriptionResult] =
+    (
+      (__ \ "subscriptionNumber").read[String].map(SubscriptionId.apply) and
+      (__ \ "id").read[String].map(SubscriptionName.apply) and
+      (__ \ "ActivationDate__c").readNullable[DateTime](lenientDateTimeReader) and
+      (__ \ "customerAcceptanceDate").readNullable[DateTime](lenientDateTimeReader) and
+      (__ \ "termStartDate").readNullable[DateTime](lenientDateTimeReader) and
+      (__ \ "termEndDate").readNullable[DateTime](lenientDateTimeReader) and
+      (__ \ "ratePlans").read[List[RatePlan]]
+    )(SubscriptionResult.apply _)
```
johnduffell: Since json custom readers are quite tricky in terms of syntax, I am generally trying to have a "wire" model that's just strings, and internal to the GetSubscription class, and then have a thing to construct a domain class with the nice IDs etc from the wire model.
The advantage is that all the code is just normal scala rather than using the applicative and map on the Reads which is a bit confusing.

handlers/digital-subscription-expiry/src/main/scala/com/gu/digitalSubscriptionExpiry/zuora/GetSubscription.scala
```diff
@@ -0,0 +1,54 @@
...
+
+  def apply(request: SubscriptionId): WithDepsFailableOp[ZuoraDeps, SubscriptionResult] =
+    Reader { zuoraDeps: ZuoraDeps => ZuoraRestRequestMaker(zuoraDeps).get[SubscriptionResult](s"object/subscriptions/${request.get}").leftMap(fromClientFail) }.toEitherT
```
johnduffell: for new stuff you needn't use a Reader (that was a failed experiment)
Just change from
```scala
def apply(request: SubscriptionId): WithDepsFailableOp[ZuoraDeps, SubscriptionResult] =
    Reader { zuoraDeps: ZuoraDeps => ZuoraRestRequestMaker(zuoraDeps).get[SubscriptionResult](s"object/subscriptions/${request.get}").leftMap(fromClientFail) }.toEitherT
}
```
to something like:
```scala
def apply(zuoraDeps: ZuoraDeps)(request: SubscriptionId): FailableOp[SubscriptionResult] =
    ZuoraRestRequestMaker(zuoraDeps).get[SubscriptionResult](s"object/subscriptions/${request.get}").leftMap(fromClientFail)
}
```

As background info, I need to go through and remove all the remaining ones from the code base.  As a next step, ideally the parameter would be the `RestRequestMaker.Requests` rather than the `ZuoraDeps` but that's a separate refactoring.

handlers/digital-subscription-expiry/src/main/scala/com/gu/digitalSubscriptionExpiry/zuora/GetSubscription.scala
pvighi: We'll keep than in mind for when we actually do the Zuora part.
The only reason this is included in the PR is because I didn't anticipate that I was going to open an initial PR without the Zuora so some Zuora bits got included. 
I don't think it is worth the effort to remove all the Zuora parts from this PR unless there is something seriously wrong with them

johnduffell:
great, and there's plenty of tests which is good. just a few small comments.

something I'd like to see is a single "EndToEndHandlerTest" test that calls `Handler. runWithEffects` uses mocked responses just to check everything is wired together automatically correctly.  This is almost the health check only it doesn't need credentials or use real resources.  It can reuse the mock responses from the unit tests to avoid too much setup.

I would have thought it would also be important to have at least one EffectsTest, this means even if people forget to deploy to CODE and try hitting it, at least there's a good chance the lambda would work.  Not sure what the risks would be, I guess mainly config issues.  Maybe adding a Config effects test would be enough?

Also once it's shipped/live it would be good to add it to the list of things checked by the CODEPRODHeathCheck test and runscope in PROD.

PR #70, opened 2018-04-13
remove all scalaz.Reader from the zuora stuff

johnduffell:
The last layer of readers was still hanging around and making the code confusing.  This purges them all from the zuora layer, and just passes in a Requests (which may be badly named, but it is the thing that lets us call zuora. 
This means the Zuora operations are a bit more clean.
ZuoraDeps is also removed/combined with ZuoraRestRequestMaker, as the request maker is the only reason to need the deps, so we may as well pass that instead.


handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/zuora/AddIdentityIdToAccount.scala
```diff
@@ -17,8 +17,8 @@ object AddIdentityIdToAccount {
     WireRequest(id.value)
   }
 
-  def apply(zuoraDeps: ZuoraDeps)(accountId: AccountId, identityId: IdentityId): FailableOp[Unit] = {
-    val accounts = ZuoraRestRequestMaker(zuoraDeps).put[WireRequest, Unit](reqFromIdentityId(identityId), s"accounts/${accountId.value}")
+  def apply(requests: Requests)(accountId: AccountId, identityId: IdentityId): FailableOp[Unit] = {
+    val accounts = requests.put[WireRequest, Unit](reqFromIdentityId(identityId), s"accounts/${accountId.value}")
 
     accounts.leftMap(e => ApiGatewayResponse.internalServerError(e.message))
```
johnduffell: still need to get rid of the need to repeat this leftMap everywhere.  Adam has shown me the Attempt case class+object that can hide all the mapping code in one place but still be understandable.  This might be a solution https://github.com/guardian/security-hq/blob/master/hq/app/utils/attempt/Attempt.scala

lib/handler/src/main/scala/com/gu/util/apigateway/ApiGatewayHandler.scala
```diff
@@ -77,6 +77,7 @@ object LoadConfig extends Logging {
 
   def default[StepsConfig: Reads]: Reader[(Stage, Try[String]), FailableOp[Config[StepsConfig]]] = loadConfig(Config.parseConfig[StepsConfig] _)
 
+  // FIXME don't use Reader
```
johnduffell: this bit is pretty self contained and not related to the zuora stuff, we can save removing it for a more rainy day

lib/zuora/src/main/scala/com/gu/util/zuora/internal/Types.scala
```diff
@@ -1,23 +0,0 @@
-package com.gu.util.zuora.internal
```
johnduffell: one of the Types objects is gone, but there are still two.  we need to arrange things so that they don't make it hard to find files by just searching the project  this was pointed out by @pvighi .  Perhaps we need a convention, perhaps add the name of the project/package to the class name if it's fairly gneeric?

src/main/scala/com/gu/autoCancel/AutoCancel.scala
```diff
@@ -3,22 +3,23 @@ package com.gu.autoCancel
...
+  def apply(requests: Requests)(acRequest: AutoCancelRequest): FailableOp[Unit] = {
     val AutoCancelRequest(accountId, subToCancel, cancellationDate) = acRequest
     logger.info(s"Attempting to perform auto-cancellation on account: $accountId")
     val zuoraOp = for {
-      _ <- ZuoraUpdateCancellationReason(subToCancel).withLogging("updateCancellationReason")
-      _ <- ZuoraCancelSubscription(subToCancel, cancellationDate).withLogging("cancelSubscription")
-      _ <- ZuoraDisableAutoPay(accountId).withLogging("disableAutoPay")
+      _ <- ZuoraUpdateCancellationReason(requests)(subToCancel).withLogging("updateCancellationReason")
+      _ <- ZuoraCancelSubscription(requests)(subToCancel, cancellationDate).withLogging("cancelSubscription")
+      _ <- ZuoraDisableAutoPay(requests)(accountId).withLogging("disableAutoPay")
```
johnduffell: here we can see we where we had the advantage of the reader monad, because now we have to pass in the requests every time rather than it happening for free when we flatmap!

handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/zuora/CountZuoraAccountsForIdentityId.scala
```diff
@@ -3,22 +3,25 @@ package com.gu.identityBackfill.zuora
...
+  def apply(requests: Requests)(identityId: IdentityId): FailableOp[Int] = {
+    val accountByIdentityQuery = ZuoraQuery.Query(s"SELECT Id FROM Account where IdentityId__c='${identityId.value}'")
     val accounts = for {
-      accountsWithEmail <- ListT(ZuoraQuery.getResults[WireResponse](ZuoraQuery.Query(s"SELECT Id FROM Account where IdentityId__c='${identityId.value}'")).map(_.records))
+      accountsWithEmail <- ListT[ClientFailableOp, WireResponse](
+        ZuoraQuery.getResults[WireResponse](requests)(accountByIdentityQuery).map(_.records)
+      )
     } yield AccountId(accountsWithEmail.Id)
 
-    accounts.run.run.run(zuoraDeps).bimap(e => ApiGatewayResponse.internalServerError(e.message), l => l.size)
```
jacobwinch: nice to see .run.run.run going away!

lib/handler/src/main/scala/com/gu/util/reader/Types.scala
```diff
@@ -82,19 +68,6 @@ object Types extends Logging {
...
-
-      (configHttpFailableOp.run map {
-        _.withLogging(message)
-      }).toEitherT
-
-    }
-
-  }
-
   implicit class LogImplicit2[A](op: A) {
```
jacobwinch: Can we drop the 2 now?

johnduffell: probably should be completely different

PR #73, opened 2018-04-17
Use java.time instead of joda time

pvighi:
Not a PR to merge to master, just to bring my changes into check-zuora-for-subs-expiry
changes:
Use java.time instead of joda time


handlers/digital-subscription-expiry/src/test/scala/com/gu/digitalSubscriptionExpiry/zuora/GetSubscriptionEffectsTest.scala
```diff
@@ -42,18 +42,22 @@ class GetSubscriptionEffectsTest extends FlatSpec with Matchers {
       subscription
     }
 
-    val dateFormatter = DateTimeFormat.forPattern("dd/MM/yyyy")
-    def asDate(str: String) = dateFormatter.parseLocalDate(str)
-    def asDateTime(str: String) = ISODateTimeFormat.dateTimeParser().parseDateTime(str)
-
-    val customerAcceptanceDate = asDate("15/12/2017")
-    val startDate = asDate("29/11/2017")
+    def removeActivationTime(sub: SubscriptionResult) = sub.copy(
```
pvighi: this is to make the test more reliable, we just compare the date portion of the dateTime and set everything to 0:00

if we just change it so that we load the activation date as a string in the subscriptionResult class we would not have to do this.

handlers/digital-subscription-expiry/src/main/scala/com/gu/digitalSubscriptionExpiry/emergencyToken/TokenPayloadOps.scala
```diff
@@ -1,12 +1,15 @@
...
 import com.gu.cas.{Guardian, SevenDay, SubscriptionCode, TokenPayload}
-import org.joda.time.LocalDate
+import java.time.LocalDate
 
 object TokenPayloadImplicits {
 
   implicit class TokenPayloadOps(payload: TokenPayload) {
-    def expiryDate: LocalDate = payload.creationDate.plus(payload.period).plusDays(1).toLocalDate
+    def expiryDate: LocalDate = {
+      val jodaExpiryDate = payload.creationDate.plus(payload.period).plusDays(1)
```
jacobwinch: is this naming intentional / can you clarify it?

handlers/digital-subscription-expiry/src/main/scala/com/gu/digitalSubscriptionExpiry/common/CommonFormatters.scala
```diff
@@ -1,14 +0,0 @@
-package com.gu.digitalSubscriptionExpiry.common
```
pvighi: we don't need these reader and writers anymore. It looks like play json supports java.time and it looks like their default format for LocalDate is already "yyyy-MM-dd"

handlers/digital-subscription-expiry/src/main/scala/com/gu/digitalSubscriptionExpiry/common/CommonFormatters.scala
```diff
@@ -1,14 +0,0 @@
...
-
-import org.joda.time.{DateTime, LocalDate}
-import play.api.libs.json.{JodaReads, JodaWrites, Reads, Writes}
-
-object CommonFormatters {
-  implicit val localDateWriter: Writes[LocalDate] = JodaWrites.jodaLocalDateWrites("yyyy-MM-dd")
-
-  implicit val lenientDateTimeReader: Reads[DateTime] =
-    JodaReads.DefaultJodaDateTimeReads orElse Reads.IsoDateReads.map(new DateTime(_))
```
pvighi: I didn't really replace this so we probably will have problems parsing subscriptions if the activation date is not a consistent format. I think we should just redefine it as Option[String] in subscriptionResult to reflect that it is a String in Zuora

handlers/digital-subscription-expiry/src/test/scala/com/gu/digitalSubscriptionExpiry/zuora/GetSubscriptionEffectsTest.scala
lmath: Yeah, let's change the activation date in the SubscriptionResult case class to be a string. Currently in CAS we're just checking if there's already an existing value before setting an activation date.

handlers/digital-subscription-expiry/src/main/scala/com/gu/digitalSubscriptionExpiry/common/CommonFormatters.scala
lmath: 👍

handlers/digital-subscription-expiry/src/main/scala/com/gu/digitalSubscriptionExpiry/emergencyToken/TokenPayloadOps.scala
pvighi: Yes, although I should note that this is a PR just to put my changes into leigh anne's branch not master.
This is just copied over from [cas proxy](https://github.com/guardian/content-authorisation-proxy/blob/master/src/main/scala/com/gu/subscriptions/cas/model/TokenPayloadOps.scala), basically it gets the creation date of the token and adds the period the token is valid plus one for to get the date when the token expires.

This version  has the additional step of converting from joda date classes to java.time since  token payload is defined in the shared library [content authorisation common](https://github.com/guardian/content-authorisation-common/blob/master/src/main/scala/com/gu/cas/Token.scala#L21) a which still uses joda time.

PR #74, opened 2018-04-17
if identity is not validated, stop processing

johnduffell:
This is a new piece of code to avoid linking identity ids with non validated emails.  This avoids the risk of syncing the address of a real customer to someone who has created a fake account.

@pvighi @paulbrown1982 @lmath @jacobwinch


johnduffell:
the codacy issues were just added complexity in one of the files (I could have split it into more than one) and files changing too often being a churning risk.  I'll keep it in mind in my next PR.

PR #75, opened 2018-04-17
Check zuora for subs expiry

pvighi:
added checking zuora for digipack expiration date.
If the subscription id is not recognised as an emergency token the app will check zuora by following the following steps:
- find a subscription in Zuora with the provided id
- check if the related account data matches the provided password
- get all the active digipack rate plan charges related to the subscription
- if the activation date is not set in zuora set it to today (unless the 'noActivation' query string param was provided to disable this)
- return the effective end date of the charge that ends last ( there should be only one in all/most cases)


handlers/digital-subscription-expiry/src/main/scala/com/gu/digitalSubscriptionExpiry/DigitalSubscriptionExpirySteps.scala
```diff
@@ -1,71 +1,53 @@
...
-        expiryResponse <- (getEmergencyTokenExpiry(expiryRequest.subscriberId) orElse getZuoraExpiry()).toRightDisjunction(notFoundResponse)
-      } yield {
-        apiResponse(expiryResponse, "200")
-      }
-      val response = successfulOrErrorResponse.valueOr(identity)
-      -\/(response)
-    }
-
-    //TODO
-    //    def healthcheck() =
```
pvighi: we decided it's not worth adding a health check since the operation is mostly read only anyway, we can just call /subs as a healthcheck the same way we do it for /auth

handlers/digital-subscription-expiry/src/main/scala/com/gu/digitalSubscriptionExpiry/zuora/GetSubscriptionExpiry.scala
```diff
@@ -0,0 +1,63 @@
...
+    val formattedPassword = format(password)
+
+    candidates.contains(formattedPassword)
+  }
+
+  implicit def dateOrdering: Ordering[LocalDate] = Ordering.fromLessThan(_ isAfter _)
+
+  private val digipackChargeNamePrefixes = List("DIGIPACK", "DIGITAL PACK")
+
+  private def isDigipackName(chargeName: String) = {
```
pvighi: @paulbrown1982 do you think it would be better to just do 'chargeName.startsWith("Digi")' like you mentioned the other day ?

handlers/digital-subscription-expiry/src/test/scala/com/gu/digitalSubscriptionExpiry/zuora/GetSubscriptionReaderTest.scala
```diff
@@ -1,24 +1,36 @@
...
 import com.gu.digitalSubscriptionExpiry.zuora.GetSubscription._
-import org.joda.time.DateTime
+import org.joda.time.format.DateTimeFormat
 import org.scalatest.FlatSpec
 import org.scalatest.Matchers._
 import play.api.libs.json._
 
 class GetSubscriptionReaderTest extends FlatSpec {
+
+  val dateFormatter = DateTimeFormat.forPattern("dd/MM/yyyy")
```
paulbrown1982: Does this test still need the joda DateTimeFormat?

handlers/digital-subscription-expiry/src/test/scala/com/gu/digitalSubscriptionExpiry/zuora/UpdateSubscriptionEffectsTest.scala
```diff
@@ -0,0 +1,47 @@
...
+import com.gu.util.zuora.ZuoraRestRequestMaker
+import com.gu.util.{Config, Stage}
+import java.time.LocalDate
+
+import com.gu.digitalSubscriptionExpiry.zuora.GetAccountSummary.AccountId
+import org.scalatest.{FlatSpec, Matchers}
+import scalaz.syntax.std.either._
+import scalaz.{\/, \/-}
+
+class UpdateSubscriptionEffectsTest extends FlatSpec with Matchers {
```
lmath: This test is not very useful currently so I'm going to see if I can improve it or remove it shortly.

handlers/digital-subscription-expiry/src/main/scala/com/gu/digitalSubscriptionExpiry/zuora/GetSubscriptionExpiry.scala
paulbrown1982: There's only one more name case, the '-bolt on' text so maybe either include that or do the .startsWith as you suggest.

handlers/digital-subscription-expiry/src/main/scala/com/gu/digitalSubscriptionExpiry/DigitalSubscriptionExpirySteps.scala
johnduffell: Fair enough - in general I'd prefer to have a standardised health check for everything, so that we can call them all centrally and consistently by iterating through a list of endpoints rather than having to hard code some parameters.  The calling /subs is also useful as a higher level test.
If we're happy with the default "does it start and load the config" health check then it's fine to leave it like that.
We're going to have to come up with a clever way when we start doing lambdas that don't have endpoints (how about lambdas that read from SQS queues?)

handlers/digital-subscription-expiry/src/main/scala/com/gu/digitalSubscriptionExpiry/zuora/GetSubscriptionExpiry.scala
pvighi: we are doing 'contains' here so it will match "digital pack monthly" or "digital pack-bolt on", right?

handlers/digital-subscription-expiry/src/main/scala/com/gu/digitalSubscriptionExpiry/zuora/GetSubscriptionExpiry.scala
```diff
@@ -0,0 +1,63 @@
...
+    val dateToCheck = if (!subscription.startDate.isAfter(date) && subscription.customerAcceptanceDate.isAfter(date)) subscription.customerAcceptanceDate else date
+
+    def isActiveDigipack(charge: RatePlanCharge) = isDigipackName(charge.name) && !charge.effectiveEndDate.isBefore(dateToCheck) && !charge.effectiveStartDate.isAfter(dateToCheck)
+
+    val activeDigipackCharges = subscription.ratePlans.map(_.ratePlanCharges).flatten.filter(isActiveDigipack)
+
+    if (activeDigipackCharges.isEmpty) None else Some(activeDigipackCharges.map(_.effectiveEndDate).max)
+  }
+
+  def apply(providedPassword: String, subscription: SubscriptionResult, accountSummary: AccountSummaryResult, date: LocalDate = LocalDate.now()): FailableOp[Unit] =
```
johnduffell: shouldn't call LocalDate.now from the code as it is not referentially transparent - just remove the default value and pass in from the handler
```scala
def apply(date: LocalDate)(providedPassword: String, subscription: SubscriptionResult, accountSummary: AccountSummaryResult): FailableOp[Unit] =
```

handlers/digital-subscription-expiry/src/main/scala/com/gu/digitalSubscriptionExpiry/zuora/GetSubscription.scala
```diff
@@ -1,51 +1,54 @@
...
   implicit val reads: Reads[SubscriptionResult] =
     (
       (__ \ "subscriptionNumber").read[String].map(SubscriptionId.apply) and
       (__ \ "id").read[String].map(SubscriptionName.apply) and
-      (__ \ "ActivationDate__c").readNullable[DateTime](lenientDateTimeReader) and
-      (__ \ "customerAcceptanceDate").readNullable[DateTime](lenientDateTimeReader) and
-      (__ \ "termStartDate").readNullable[DateTime](lenientDateTimeReader) and
-      (__ \ "termEndDate").readNullable[DateTime](lenientDateTimeReader) and
+      (__ \ "accountId").read[String].map(AccountId.apply) and
+      (__ \ "ActivationDate__c").readNullable[String] and
```
pvighi: Activation date is now read as an Option[String] rather than Option[DateTime] as it was in CAS since it is defined as a String in Zuora for some reason and it has no guaranteed formatting.
 In this app we only care if it is empty or not so it is not worth the effort to parse it and risk getting an error because of the formatting not being what we expected

handlers/digital-subscription-expiry/src/main/scala/com/gu/digitalSubscriptionExpiry/Handler.scala
```diff
@@ -28,8 +29,18 @@ object Handler extends Logging {
...
         val emergencyTokens = EmergencyTokens(config.stepsConfig.emergencyTokens)
-        DigitalSubscriptionExpirySteps(GetTokenExpiry(emergencyTokens))
+        val zuoraRequests = ZuoraRestRequestMaker(rawEffects.response, config.stepsConfig.zuoraRestConfig)
+
+        DigitalSubscriptionExpirySteps(
+          getEmergencyTokenExpiry = GetTokenExpiry(emergencyTokens),
+          getSubscription = GetSubscription(zuoraRequests),
+          updateSubscription = UpdateSubscription(zuoraRequests),
+          getAccountSummary = GetAccountSummary(zuoraRequests),
+          getSubscriptionExpiry = GetSubscriptionExpiry.apply,
```
johnduffell: this should be partially applies with the daye
```scala
GetSubscriptionExpiry(rawEffects.now)
```
then there's no need to pass it on on the next line any more

handlers/digital-subscription-expiry/src/main/scala/com/gu/digitalSubscriptionExpiry/zuora/UpdateSubscription.scala
```diff
@@ -0,0 +1,31 @@
...
+
+  implicit val writes = Json.writes[UpdateRequestBody]
+
+  def asIsoString(date: LocalDateTime) = date.format(ISO_LOCAL_DATE_TIME)
+
+  def apply(requests: Requests)(subscription: SubscriptionResult, activationDate: String = asIsoString(LocalDateTime.now())): FailableOp[Unit] =
+    subscription.casActivationDate.fold {
+      logger.debug(s"Setting activation date ($activationDate) in subscription ${subscription}")
+      requests.put[UpdateRequestBody, Unit](UpdateRequestBody(activationDate), s"subscriptions/${subscription.id.get}").leftMap(clientFail => ApiGatewayResponse.internalServerError(s"zuora client fail: ${clientFail.message}"))
+    } { date =>
```
johnduffell: really the UpdateSubscription object should just do the update only, not make any decisions.  I would suggest to put a separare step to run before this to do the checks if it's already active (then you can test the check on its own, and then the operation is very simple)

handlers/digital-subscription-expiry/src/main/scala/com/gu/digitalSubscriptionExpiry/zuora/GetSubscriptionExpiry.scala
pvighi: yeah I forgot to remove the default value, but it is already being passed the date from the handler the default is not used

handlers/digital-subscription-expiry/src/main/scala/com/gu/digitalSubscriptionExpiry/zuora/UpdateSubscription.scala
lmath: Oh, yes, good point. Commit incoming.

handlers/digital-subscription-expiry/src/test/scala/com/gu/digitalSubscriptionExpiry/DigitalSubscriptionExpiryHandlerEffectsTest.scala
```diff
@@ -0,0 +1,129 @@
...
+    //execute
+
+    Handler.runWithEffects(rawEffects, LambdaIO(stream, os, null))
+
+    val responseString = new String(os.toByteArray, "UTF-8")
+
+    responseString jsonMatches notFoundResponse
+  }
+
+  it should "return bad request for invalid json" taggedAs EffectsTest in {
```
johnduffell: this is probably not an essential test to run as an EffectsTest, it could be tested as a unit test.  As if we ship with this test broken, it won't be a big issue

handlers/digital-subscription-expiry/src/test/scala/com/gu/digitalSubscriptionExpiry/zuora/GetSubscriptionExpiryTest.scala
```diff
@@ -0,0 +1,133 @@
...
+      startDate = lastYear,
+      endDate = lastWeek,
+      ratePlans = List(RatePlan("Digital Pack", List(RatePlanCharge("Digital Pack Monthly", lastYear, lastWeek))))
+    )
+
+    val actualResponse = GetSubscriptionExpiry("123-sold", expiredDigitalPack, accountSummary)
+    actualResponse shouldEqual notFoundResponse
+  }
+
+}
```
johnduffell: this is fine, but the fact there are a lot of tests indicated that the function is doing a lot.  There's no reason against splitting down a bigger step into another set of steps - a bit like I did with PreReqCheck after your suggestion it was getting complicated.
It does make the wiring a little more complicated (in the handler) but it helps keep the testing concise.

As I say no need to change this, but something to keep in mind that having "Steps" is a concept that can be applied at more than one level.

handlers/digital-subscription-expiry/src/test/scala/com/gu/digitalSubscriptionExpiry/zuora/GetSubscriptionReaderTest.scala
```diff
@@ -1,24 +1,33 @@
...
 import com.gu.digitalSubscriptionExpiry.zuora.GetSubscription._
-import org.joda.time.DateTime
 import org.scalatest.FlatSpec
 import org.scalatest.Matchers._
 import play.api.libs.json._
 
 class GetSubscriptionReaderTest extends FlatSpec {
+
   "GetSubscription" should "deserialise correctly a valid response from Zuora" in {
     val subscriptionStream = getClass.getResourceAsStream("/digitalSubscriptionExpiry/validSubscription.json")
```
johnduffell: this is existing anyway, but just for the record as we discussed together before, I prefer to use triple quotes and a scala string, maybe at the bottom of the file if possible. Then it makes it easier to see and navigate to the test data

lib/handler/src/main/scala/com/gu/util/apigateway/ApiGatewayRequest.scala
```diff
@@ -41,7 +42,8 @@ object URLParams {
     (JsPath \ "apiToken").readNullable[String] and
     (JsPath \ "onlyCancelDirectDebit").readNullable[String].map(_.contains("true")) and
     (JsPath \ "stripeAccount").readNullable[String].map(_.flatMap(StripeAccount.fromString)) and
-    (JsPath \ "isHealthcheck").readNullable[String].map(_.contains("true"))
+    (JsPath \ "isHealthcheck").readNullable[String].map(_.contains("true")) and
+    (JsPath \ "noActivation").readNullable[String].map(_.contains("true"))
```
johnduffell: I guess this is an existing parameter - but if it's new I'd prefer to call it "dontActivate" or "preventActivation" assuming that's what it means

johnduffell: this also reminds me that we need to find a way of not making this list of possible parameters in the API gateway code 🙁

lib/zuora/src/test/scala/com/gu/util/ZuoraRestServiceTest.scala
```diff
@@ -70,7 +70,7 @@ class ZuoraRestServiceTest extends AsyncFlatSpec {
 
   it should "return a left[String] if the body of a successful http response has a zuora failed in it" in {
     val either = ZuoraRestRequestMaker.zuoraIsSuccessful(validFailedUpdateSubscriptionResult)
-    assert(either == -\/(internalServerError("Received failure result from Zuora during autoCancellation")))
+    assert(either == -\/(internalServerError("Received a failure result from Zuora")))
```
johnduffell: 🙈

src/test/scala/com/gu/stripeCustomerSourceUpdated/SourceUpdatedStepsTest.scala
```diff
@@ -327,9 +327,9 @@ class SourceUpdatedStepsUpdatePaymentMethodTest extends FlatSpec with Matchers {
     val actual = SourceUpdatedSteps.createUpdatedDefaultPaymentMethod(
       TestData.zuoraDeps(effects)
     )(
-      PaymentMethodFields(PaymentMethodId("PMID"), AccountId("fake"), NumConsecutiveFailures(1)),
-      eventData
-    )
+        PaymentMethodFields(PaymentMethodId("PMID"), AccountId("fake"), NumConsecutiveFailures(1)),
```
johnduffell: oh yeah mine kept doing that too! I've no idea why, but we might as well let it

handlers/digital-subscription-expiry/src/test/scala/com/gu/digitalSubscriptionExpiry/DigitalSubscriptionExpiryHandlerEffectsTest.scala
```diff
@@ -0,0 +1,129 @@
...
+  val badRequest =
+    """
+      |{
+      |"statusCode":"400",
+      |"headers":{"Content-Type":"application/json"},
+      |"body":"{\n  \"error\" : {\n    \"message\" : \"Mandatory data missing from request\",\n    \"code\" : -50\n  }\n}"
+      |}
+      |""".stripMargin
+
+}
```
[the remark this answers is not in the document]
johnduffell: for some examples of End to end Handler tests without actually being effects tests, see here:
https://github.com/guardian/zuora-auto-cancel/blob/master/handlers/identity-backfill/src/test/scala/com/gu/identityBackfill/EndToEndHandlerTest.scala

handlers/digital-subscription-expiry/src/main/scala/com/gu/digitalSubscriptionExpiry/zuora/GetSubscriptionExpiry.scala
johnduffell: yes (although misleading you called it" Prefixes" but actually using contains not using startsWith)

handlers/digital-subscription-expiry/src/test/scala/com/gu/digitalSubscriptionExpiry/DigitalSubscriptionExpiryHandlerEffectsTest.scala
pvighi: we can just remove these cases as they are already covered by the steps test  [here](https://github.com/guardian/zuora-auto-cancel/blob/fbc2b621c42a33c94b3c8b03266aa5d4692bdcec/handlers/digital-subscription-expiry/src/test/scala/com/gu/digitalSubscriptionExpiry/DigitalSubscriptionExpiryStepsTest.scala#L142)

handlers/digital-subscription-expiry/src/main/scala/com/gu/digitalSubscriptionExpiry/DigitalSubscriptionExpirySteps.scala
```diff
@@ -1,71 +1,50 @@
...
-    //      for {
-    //        identityId <- getByEmail(EmailAddress("john.duffell@guardian.co.uk")).leftMap(a => ApiGatewayResponse.internalServerError(a.toString)).withLogging("healthcheck getByEmail")
-    //        _ <- countZuoraAccountsForIdentityId(identityId)
-    //        _ <- sfAuth()
-    //      } yield ()
+      val nowAsString = LocalDateTime.now().format(ISO_LOCAL_DATE_TIME)
```
pvighi: we should probably get the current date from the handler like we do for getSubscriptionExpiry instead of doing LocalDateTime.now() here.
Although something might have to be refactored as the now function in RawEffects has no time, just date

lib/handler/src/main/scala/com/gu/util/apigateway/ApiGatewayRequest.scala
lmath: Yeah, it is already noActivation in CAS

handlers/digital-subscription-expiry/src/main/scala/com/gu/digitalSubscriptionExpiry/DigitalSubscriptionExpirySteps.scala
johnduffell: oh yes, didn't spot that, the date/time should indeed come from the raweffects via the handler.  This can be preapplied into to `updateSubscription` (it would be `updateSubscriptionToToday` once partially applied in that case I suppose!)

handlers/digital-subscription-expiry/src/test/scala/com/gu/digitalSubscriptionExpiry/zuora/UpdateSubscriptionEffectsTest.scala
```diff
@@ -1,47 +0,0 @@
...
-      customerAcceptanceDate = testDate,
-      ratePlans = Nil
-    )
-    //    val now = LocalDate.now()
-    val actual: \/[io.Serializable, Unit] = for {
-      configAttempt <- ConfigLoad.load(Stage("DEV")).toEither.disjunction
-      config <- Config.parseConfig[StepsConfig](configAttempt)
-
-      zuoraRequests = ZuoraRestRequestMaker(RawEffects.createDefault.response, config.stepsConfig.zuoraRestConfig)
-      update <- UpdateSubscription(zuoraRequests)(testSub)
```
johnduffell: it's a good idea to keep this test (and I would encourage to have a single effectstest for every actual endpoint where possible) just as long as there's no real logic/decisions in the code being tested.
if it's an update call, make sure to add an operation in the test code to do the similar Get to check it really worked

johnduffell: see here for an example:
https://github.com/guardian/zuora-auto-cancel/blob/master/handlers/identity-backfill/src/test/scala/com/gu/identityBackfill/salesforce/updateSFIdentityId/UpdateSalesforceIdentityIdEffectsTest.scala
although I had to make a special Get call just for the testing purposes
https://github.com/guardian/zuora-auto-cancel/blob/master/handlers/identity-backfill/src/test/scala/com/gu/identityBackfill/salesforce/updateSFIdentityId/GetSalesforceIdentityId.scala

handlers/digital-subscription-expiry/src/test/scala/com/gu/digitalSubscriptionExpiry/zuora/GetSubscriptionExpiryTest.scala
```diff
@@ -126,7 +150,7 @@ class GetSubscriptionExpiryTest extends FlatSpec {
       ratePlans = List(RatePlan("Digital Pack", List(RatePlanCharge("Digital Pack Monthly", lastYear, lastWeek))))
     )
 
-    val actualResponse = GetSubscriptionExpiry("123-sold", expiredDigitalPack, accountSummary)
+    val actualResponse = GetSubscriptionExpiry(today)("123-sold", expiredDigitalPack, accountSummary)
```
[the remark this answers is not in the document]
pvighi: I think that is kinda what this test is doing. 
Instead of using the same subscription and calling GetSubscriptionExpiry with a future day which would be the sensible thing to do I created a subscription that expires in the past and called getSubscriptionExpiry with today. I guess It's a more realistic case but it has more setup code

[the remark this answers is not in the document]
johnduffell: ah, sorry, didn't read it right!! 😀 you're right!!

johnduffell:
image
image
great job and thanks for the feedback on the lambda structure too! ship it!

handlers/digital-subscription-expiry/src/test/scala/com/gu/digitalSubscriptionExpiry/zuora/UpdateSubscriptionEffectsTest.scala
johnduffell: feel free to save this test for later, no need to block merging the rest of the stuff waiting for this 👍

PR #76, opened 2018-04-18
make sub expiry fix behave more like CAS

pvighi:
Use the rateplan charges just to see if there is an active digipack in the sub but return the sub term end instead of the effective end date of the charge.
Also add 1 to the date in the response

### DIFF
--- a/handlers/digital-subscription-expiry/src/main/scala/com/gu/digitalSubscriptionExpiry/Handler.scala
+++ b/handlers/digital-subscription-expiry/src/main/scala/com/gu/digitalSubscriptionExpiry/Handler.scala
@@ -32,13 +32,13 @@ object Handler extends Logging {
 
         val emergencyTokens = EmergencyTokens(config.stepsConfig.emergencyTokens)
         val zuoraRequests = ZuoraRestRequestMaker(rawEffects.response, config.stepsConfig.zuoraRestConfig)
-        def today() = rawEffects.now().toLocalDate
+        val today = () => rawEffects.now().toLocalDate
         DigitalSubscriptionExpirySteps(
           getEmergencyTokenExpiry = GetTokenExpiry(emergencyTokens),
           getSubscription = GetSubscription(zuoraRequests),
           setActivationDate = SetActivationDate(zuoraRequests, rawEffects.now),
           getAccountSummary = GetAccountSummary(zuoraRequests),
-          getSubscriptionExpiry = GetSubscriptionExpiry(today _),
+          getSubscriptionExpiry = GetSubscriptionExpiry(today),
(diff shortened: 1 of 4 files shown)


handlers/digital-subscription-expiry/src/main/scala/com/gu/digitalSubscriptionExpiry/zuora/GetSubscriptionExpiry.scala
```diff
@@ -40,9 +40,11 @@ object GetSubscriptionExpiry {
 
     def isActiveDigipack(charge: RatePlanCharge) = isDigipackName(charge.name) && !charge.effectiveEndDate.isBefore(dateToCheck) && !charge.effectiveStartDate.isAfter(dateToCheck)
 
-    val activeDigipackCharges = subscription.ratePlans.map(_.ratePlanCharges).flatten.filter(isActiveDigipack)
+    val hasActiveDigipackCharges = subscription.ratePlans.map(_.ratePlanCharges).flatten.exists(isActiveDigipack)
 
-    if (activeDigipackCharges.isEmpty) None else Some(activeDigipackCharges.map(_.effectiveEndDate).max)
+    if (hasActiveDigipackCharges) {
+      Some(subscription.endDate.plusDays(1))
+    } else None
```
pvighi: this code is trying to replicate [this](https://github.com/guardian/content-authorisation-proxy/blob/master/src/main/scala/com/gu/subscriptions/cas/directives/ProxyDirective.scala#L119) behaviour in CAS PROXY

handlers/digital-subscription-expiry/src/main/scala/com/gu/digitalSubscriptionExpiry/Handler.scala
```diff
@@ -32,7 +32,7 @@ object Handler extends Logging {
 
         val emergencyTokens = EmergencyTokens(config.stepsConfig.emergencyTokens)
         val zuoraRequests = ZuoraRestRequestMaker(rawEffects.response, config.stepsConfig.zuoraRestConfig)
-        def today() = rawEffects.now().toLocalDate
+        val today = () => rawEffects.now().toLocalDate
```
pvighi: I changed this to get rid of the warning "eta-expansion of zero-argument method values is deprecated" when passing this as parameter to GetSubscriptionExpiry

johnduffell:
👍 great!
I presume you are doing some comparator process at the moment just to check both systems return the same?

pvighi:
I'm just calling the two with the same subs and checking that they return the same but I don't really have any formal process or tool or anything

PR #78, opened 2018-04-24
Fetch the catalog from Zuora

jacobwinch:
Adds code to read the product catalog from Zuora.

### DIFF
--- a/handlers/catalog-service/src/main/scala/com/gu/catalogService/Handler.scala
+++ b/handlers/catalog-service/src/main/scala/com/gu/catalogService/Handler.scala
@@ -1,8 +1,16 @@
 package com.gu.catalogService
 
 import com.amazonaws.services.lambda.runtime.Context
-import com.gu.util.Logging
+import com.gu.util.{Config, Logging}
 import java.io.{InputStream, OutputStream}
+
+import com.gu.effects.RawEffects
+import com.gu.util.apigateway.LoadConfig
+import com.gu.util.apigateway.ApiGatewayHandler.LambdaIO
+import com.gu.util.zuora.{ZuoraReadCatalog, ZuoraRestConfig, ZuoraRestRequestMaker}
+import play.api.libs.json.{Json, Reads}
+
+import scalaz.{-\/, \/-}
 
 object Handler extends Logging {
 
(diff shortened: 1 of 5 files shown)


handlers/catalog-service/src/main/scala/com/gu/catalogService/Handler.scala
```diff
@@ -1,15 +1,47 @@
...
+        // Upload to S3
+      }
+    }
+
+    LoadConfig.default[StepsConfig].map {
+      failableConfig =>
+        failableConfig.map {
+          config => operation(config)
+        }
+    }.run((rawEffects.stage, rawEffects.s3Load(rawEffects.stage)))
```
johnduffell: I think we need to try to keep the "wiring" (to ZuoraRestRequestMaker) on the outside and have the logic/decision code (that we want to test) somewhere that can be just having minimal parameters and return values.
having said that, I'm not sure how much of this is actually making decisions!  Perhaps the only thing it has is a more end to end test?

johnduffell: not sure if now is a good moment to sort out the TODO about getting rid of the Reader in LoadConfig too. It shouldn't be hard.  I am then sure we can get rid of some of these lines of code.

johnduffell:
great start, another small PR out quickly 🎉 thanks.  I agree with Patricio about the tests need to come soon, I was also thinking further into the future we should also think about monitoring, as a health check wouldn't make sense in the traditional way.
This will turn out to be a nice pattern for scheduled lambdas, as all the existing ones are simply api gateway.
I think we need to move the config stuff out from the "apigateway" lib, make anew one., as you don't want to depend on all the api gateway stuff for this.
We can add that and potentially removing the Reader in this PR or the next one.  But I think we should put the zuora operation class in this handler, also separate the "wiring" from the "doing" code in this PR if possible.  At least a separate method even if you don't feel a separate object or file is appropriate.

lib/zuora/src/main/scala/com/gu/util/zuora/Zuora.scala
```diff
@@ -132,3 +132,10 @@ object ZuoraDisableAutoPay {
 
 }
 
+object ZuoraReadCatalog {
```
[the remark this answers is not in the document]
jacobwinch: For the first pass I'm going to copy the whole lot, but I agree that there's a case to be made for working out what we actually care about and just storing that in the future.

jacobwinch:
@johnduffell would you mind taking another look? I'd prefer to tackle the re-factor separately as I made a tiny start before realising that doing it properly will require fairly significant changes (which probably deserve their own PR).

johnduffell:
looking now, I might have a play locally as this wasn't quite what I meant

johnduffell:
I have araised a PR against this branch from a new branch with a few suggestions, hopefully that would make more sense rather than trying to comment, but it doesn't have to be merged https://github.com/guardian/zuora-auto-cancel/pull/79

jacobwinch:
@johnduffell - thanks for the suggestions, they look good - I've merged them in.

johnduffell:
there are a couple of files with trivial changes, not sure if it's worth changing them back to simplify the PR, but regardless it's good to go!
👍

PR #79, opened 2018-04-26
some refactoring to the catalog lambda

johnduffell:
all: Please NOTE!  Not a PR to "master", it's some suggestions to @jacobwinch .

I have had a play and thought the easiest way to demonstrate would be a PR to your branch.  Feel free to look through and decide whether it makes sense and copy what you need or merge it in as a starting point.  I'll add some comments below as I think needed.

### DIFF
--- a/handlers/catalog-service/src/main/scala/com/gu/catalogService/Handler.scala
+++ b/handlers/catalog-service/src/main/scala/com/gu/catalogService/Handler.scala
@@ -33,14 +33,7 @@ object Handler extends Logging {
       zuoraRequests = ZuoraRestRequestMaker(response, config.stepsConfig.zuoraRestConfig)
       s3Uploader = s3Write(config.stage).andThen(_.toFailableOp("write to s3"))
       fetchCatalogAttempt <- ZuoraReadCatalog(zuoraRequests).withLogging("loaded catalog")
-      _ <- ValidateCatalog(fetchCatalogAttempt).withLogging("catalog validation")
       _ <- s3Uploader(fetchCatalogAttempt).withLogging("uploaded to s3")
     } yield ()
 
-}
-
-object ValidateCatalog {
-  // testable code, makes a decision.  simplest signature (could even return a boolean?)
-  //TODO
-  def apply(catalog: String): FailableOp[Unit] = ???
 }
--- a/handlers/catalog-service/src/main/scala/com/gu/catalogService/Handler.scala
+++ b/handlers/catalog-service/src/main/scala/com/gu/catalogService/Handler.scala
@@ -1,51 +1,46 @@
 package com.gu.catalogService
 
-import com.gu.util.Logging
 import com.gu.effects.RawEffects
 import com.gu.util.apigateway.LoadConfig
-import com.gu.util.zuora.RestRequestMaker.ClientFailableOp
+import com.gu.util.reader.Types._
 import com.gu.util.zuora.{ZuoraRestConfig, ZuoraRestRequestMaker}
-import play.api.libs.json.{JsValue, Json, Reads}
-import scalaz.{-\/, \/-}
+import com.gu.util.{Config, Logging, Stage}
+import okhttp3.{Request, Response}
+import play.api.libs.json.{Json, Reads}
+
+import scala.util.Try
 
 object Handler extends Logging {
-
-  // this is the entry point
-  // it's referenced by the cloudformation so make sure you keep it in step
-  def apply(): Unit = {
-    runWithEffects(RawEffects.createDefault)
-  }
-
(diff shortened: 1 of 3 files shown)


handlers/catalog-service/src/main/scala/com/gu/catalogService/Handler.scala
```diff
@@ -1,51 +1,46 @@
...
+      zuoraRequests = ZuoraRestRequestMaker(response, config.stepsConfig.zuoraRestConfig)
+      s3Uploader = s3Write(config.stage).andThen(_.toFailableOp("write to s3"))
+      fetchCatalogAttempt <- ZuoraReadCatalog(zuoraRequests).withLogging("loaded catalog")
+      _ <- ValidateCatalog(fetchCatalogAttempt).withLogging("catalog validation")
+      _ <- s3Uploader(fetchCatalogAttempt).withLogging("uploaded to s3")
+    } yield ()
 
 }
 
-object ScheduledLambdaRunner {
```
johnduffell: nothing special about a scheduled runner, it just loads config which already exists, so I don't think we need a separate object

handlers/catalog-service/src/main/scala/com/gu/catalogService/Handler.scala
```diff
@@ -1,51 +1,46 @@
...
 object Handler extends Logging {
 
+  case class StepsConfig(zuoraRestConfig: ZuoraRestConfig)
+  implicit val stepsConfigReads: Reads[StepsConfig] = Json.reads[StepsConfig]
+
   // this is the entry point
   // it's referenced by the cloudformation so make sure you keep it in step
   def apply(): Unit = {
-    runWithEffects(RawEffects.createDefault)
+    runWithEffects(RawEffects.response, RawEffects.stage, RawEffects.s3Load, RawEffects.s3Write)
```
johnduffell: I decided since we need to add s3 writing function to the raw effects, the concept of passing that object around everywhere was bad (eventually it'd end up a touchpoint backend)  We don't want to pass in anything more anywhere than is actually essential for the method.   Hence splitting into explicit params.

handlers/catalog-service/src/main/scala/com/gu/catalogService/ZuoraReadCatalog.scala
```diff
@@ -1,11 +1,11 @@
...
 import com.gu.util.zuora.RestRequestMaker.{ClientFailableOp, Requests}
-import play.api.libs.json.JsValue
+import play.api.libs.json.{JsValue, Json}
 
 object ZuoraReadCatalog {
 
-  def apply(requests: Requests): ClientFailableOp[JsValue] =
-    requests.get[JsValue]("catalog/products?pageSize=40")
+  def apply(requests: Requests): ClientFailableOp[String] =
+    requests.get[JsValue]("catalog/products?pageSize=40").map(Json.stringify)
```
johnduffell: this is a funny one because we don't want to look inside really?  but we do want to make sure it looks vaguely like a catalog.

handlers/catalog-service/src/main/scala/com/gu/catalogService/Handler.scala
```diff
@@ -1,51 +1,46 @@
...
+    stage: Stage,
+    s3Load: Stage => Try[String],
+    s3Write: Stage => String => Try[Unit]
+  ): Unit =
+    for {
+      config <- LoadConfig.default[StepsConfig].run((stage, s3Load(stage))).withLogging("loaded config")
+      zuoraRequests = ZuoraRestRequestMaker(response, config.stepsConfig.zuoraRestConfig)
+      s3Uploader = s3Write(config.stage).andThen(_.toFailableOp("write to s3"))
+      fetchCatalogAttempt <- ZuoraReadCatalog(zuoraRequests).withLogging("loaded catalog")
+      _ <- ValidateCatalog(fetchCatalogAttempt).withLogging("catalog validation")
```
johnduffell: I guessed we'd need some catalog validation, just to avoid overwriting a valid catalog with something broken that came back or an empty string or something

jacobwinch: We confirm here that the API call was successful: https://github.com/guardian/zuora-auto-cancel/blob/471ae40e57940886b76bbf1a82e906d9df3546da/lib/zuora/src/main/scala/com/gu/util/zuora/ZuoraObjects.scala#L27 - not sure if we want additional validation on top of that?

johnduffell: maybe not, I just wanted to add some "testable" step 😆 we can destroy it

PR #80, opened 2018-04-27
update libraries

pvighi:
also removed unused library

### DIFF
--- a/handlers/digital-subscription-expiry/build.sbt
+++ b/handlers/digital-subscription-expiry/build.sbt
@@ -19,6 +19,7 @@ libraryDependencies ++= Seq(
   "com.squareup.okhttp3" % "okhttp" % "3.9.1",
   "com.typesafe.play" %% "play-json" % "2.6.9",
   "org.scalatest" %% "scalatest" % "3.0.1" % "test",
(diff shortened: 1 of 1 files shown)


handlers/digital-subscription-expiry/build.sbt
```diff
@@ -16,9 +16,8 @@ libraryDependencies ++= Seq(
   "com.amazonaws" % "aws-lambda-java-log4j" % "1.0.0",
   "com.amazonaws" % "aws-lambda-java-core" % "1.2.0",
   "com.amazonaws" % "aws-java-sdk-s3" % "1.11.265",
-  "log4j" % "log4j" % "1.2.17",
```
pvighi: This one doesn't seem to be needed. 
@johnduffell Is there we had it ? if not we should remove it from all lambdas in this project

johnduffell: maybe not, I thought it was somehow needed by the lambda library

johnduffell:
👍

pvighi:
addresses snyk report by overriding jackson-databind, same approach as https://github.com/guardian/zuora-auto-cancel/pull/65

johnduffell:
👍

PR #81, opened 2018-04-30
Upload the catalog to S3

jacobwinch:
This PR:

- Adds permissions to write to an S3 bucket
- Adds code to upload the catalog to S3
- Adds basic mocks and test code
- Causes the lambda to throw an exception on failure (so that we can add some simple monitoring which checks for failed lambda invocations)


handlers/catalog-service/src/main/scala/com/gu/catalogService/S3UploadCatalog.scala
```diff
@@ -0,0 +1,30 @@
...
+
+import scalaz.{-\/, \/, \/-}
+
+object S3UploadCatalog extends Logging {
+
+  def apply(stage: Stage, catalog: String,
+            localFileWrite: FileConstructor => Try[File],
+            s3Write: PutObjectRequest => Try[PutObjectResult]): String \/ PutObjectResult = {
+    logger.info("Uploading catalog to S3...")
+    val path = "/tmp/catalog.json" // Must use /tmp when running in a lambda
```
pvighi: will this try to write to the lambda's local file system ?

jacobwinch: Yes - see https://aws.amazon.com/lambda/faqs/ (What if I need scratch space on disk for my AWS Lambda function?)

pvighi: Maybe I'm misunderstanding how this works but why cant we just write to s3 directly without saving on disk first?

handlers/catalog-service/src/main/scala/com/gu/catalogService/Handler.scala
```diff
@@ -18,19 +21,29 @@ object Handler extends Logging {
...
+    val attempt = for {
+      config <- LoadConfig.default[StepsConfig].run((stage, s3Load(stage))).withLogging("loaded config").leftMap(_.body)
       zuoraRequests = ZuoraRestRequestMaker(response, config.stepsConfig.zuoraRestConfig)
-      fetchCatalogAttempt <- ZuoraReadCatalog(zuoraRequests).withLogging("loaded catalog")
-    } yield ()
+      fetchCatalogAttempt <- ZuoraReadCatalog(zuoraRequests).leftMap(_.message)
+      uploadCatalogAttempt <- S3UploadCatalog(stage, fetchCatalogAttempt, localFileWrite, s3Write)
+    } yield uploadCatalogAttempt
+
+    attempt.fold(failureReason => throw CatalogServiceException(failureReason), _ => Unit)
```
johnduffell: where does throwing end up? I'd prefer not to set a precendent for throwing, I'd rather we had a raweffects.fail(...) which would throw/create a metric/log as appropriate so that we can alert on it in due course

johnduffell: or just return it to the apply method and do the fold there, then at least it's at the edge for testing purposes - a fail method may well be overkill at this stage

lib/effects/src/test/scala/com/gu/effects/TestingRawEffects.scala
```diff
@@ -69,6 +72,23 @@ class TestingRawEffects(
     //verify
     requests.map(req => (req.method, req.url.encodedPath) -> Option(req.body).map(body)).toMap
   }
+
+  val dummyFile = new File("blah", "/tmp/blah")
+  val fakeRequest = new PutObjectRequest(s"gu-fake-bucket-for-testing", "fake-catalog.json", dummyFile)
+
+  val successfulS3Upload = {
```
johnduffell: these could actually be in a companion object of TestingRawEffects as they don't depend on any of the instance values

johnduffell:
looksgreat, a few small comments but good work 👍 
not suggesting you need to here, but bear in mind if you have any concern about the s3 upload/file save having trouble, you can write an EffectsTest to hit that code.  It would need a test area in a suitable bucket though.
Same goes for an end to end test - you could if you wanted make an effects test that would run through the whole lambda with DEV config, locally, to save deploying to CODE.  This would give people more confidence when they merge other code in this project that they didn't have any chance of breaking this code.

handlers/catalog-service/src/main/scala/com/gu/catalogService/Handler.scala
jacobwinch: Throwing fails the lambda invocation (this means it shows up as an error via the default AWS metrics) - so the idea was that we will have a metric to alert on without needing to write extra code or create custom metrics.

johnduffell: makes sense, sensible to reuse the functionality in that case

handlers/catalog-service/src/test/scala/com/gu/catalogService/S3UploadCatalogTest.scala
```diff
@@ -0,0 +1,15 @@
...
+import com.gu.effects.{LocalFile, UploadToS3}
+import com.gu.test.EffectsTest
+import com.gu.util.Stage
+import org.scalatest.FlatSpec
+
+class S3UploadCatalogEffectsTest extends FlatSpec {
+
+  "S3UploadCatalog" should "upload a file" taggedAs EffectsTest in {
+    val attempt = S3UploadCatalog(Stage("EffectsTest"), """{"catalog":"myProducts"}""", LocalFile.create, UploadToS3.putObject)
+    assert(attempt.isRight)
```
johnduffell: should check the file really arrived in S3

handlers/catalog-service/src/main/scala/com/gu/catalogService/S3UploadCatalog.scala
johnduffell: I agree, it would be nice for simplicity to get rid of the extra step

handlers/catalog-service/src/test/scala/com/gu/catalogService/S3UploadCatalogTest.scala
```diff
@@ -1,15 +1,29 @@
...
+    val readBefore = getLatestCatalogTimestamp
+    val upload = S3UploadCatalog(Stage("EffectsTest"), """{"catalog":"myProducts"}""", UploadToS3.putObject)
+    val readAfter = getLatestCatalogTimestamp
+    val compareDates = for {
+      beforeUpload <- readBefore
+      afterUpload <- readAfter
+    } yield afterUpload.after(beforeUpload)
+
+    compareDates.fold(
+      ex => assert(false), timestampUpdated => assert(timestampUpdated)
```
johnduffell: nice!  although I'd love to see how helpful the message is when it fails

johnduffell: "false was not true on line 24"

johnduffell:
great, nice PR 👍 🚀

PR #83, opened 2018-04-30
support accounts with null postcodes in Digital subscription expiry calls

pvighi:
There are several Zuora accounts with null postcodes. The /subs call in digitalSubscriptionExpiry will fail for those.

We initially missed this since clearing the postcode value from the Zuora console will result in a "" value instead of null. We could reproduce this in dev and UAT by setting the value explicitly to null in the zuora API.


johnduffell:
yes I hit exactly this just now running the effectstests, thanks for fixing! 👍

johnduffell:
your PR quality is ok, but for some reason there's a @jacobwinch  version of the organisation which is more strict!  Actually should look at this one not the one above: https://app.codacy.com/app/gu-sx/zuora-auto-cancel/pullRequest?prid=1597641

pvighi:
@johnduffell sorry I set the postcode to null in the sub we use in the effects tests on purpose just to have something to run while developing

PR #88, opened 2018-05-08
Split config into separate package

jacobwinch:
This PR moves code related to config into a separate package.

### DIFF
--- a/handlers/digital-subscription-expiry/src/test/scala/com/gu/digitalSubscriptionExpiry/zuora/GetAccountSummaryEffectsTest.scala
+++ b/handlers/digital-subscription-expiry/src/test/scala/com/gu/digitalSubscriptionExpiry/zuora/GetAccountSummaryEffectsTest.scala
@@ -3,10 +3,10 @@ import java.io
 
 import com.gu.digitalSubscriptionExpiry.Handler.StepsConfig
 import com.gu.digitalSubscriptionExpiry.zuora.GetAccountSummary.{AccountId, AccountSummaryResult}
-import com.gu.effects.{ConfigLoad, RawEffects}
+import com.gu.effects.{S3ConfigLoad, RawEffects}
 import com.gu.test.EffectsTest
 import com.gu.util.zuora.ZuoraRestRequestMaker
-import com.gu.util.{Config, Stage}
+import com.gu.util.config.{LoadConfig, Stage}
 import org.scalatest.{FlatSpec, Matchers}
 import scalaz.syntax.std.either._
 import scalaz.{\/, \/-}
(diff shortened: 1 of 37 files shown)


lib/handler/src/main/scala/com/gu/util/config/LoadConfig.scala
```diff
@@ -0,0 +1,44 @@
...
+    s3Load: Try[String]
+  ): FailableOp[Config[StepsConfig]] = {
+    logger.info(s"${this.getClass} Lambda is starting up in $stage")
+    for {
+      textConfig <- s3Load.toFailableOp("load config from s3")
+      config <- parseConfig(textConfig).toFailableOp("parse config file")
+      _ <- if (stage == config.stage)
+        \/-(())
+      else
+        -\/(ApiGatewayResponse.internalServerError(s"running in $stage with config from ${config.stage}"))
```
jacobwinch: We still need another small PR to make this return type non-API Gateway specific.

lib/handler/src/main/scala/com/gu/util/config/LoadConfig.scala
```diff
@@ -0,0 +1,44 @@
+package com.gu.util.config
```
johnduffell: the api gateway stuff shouldn't be in the jar if it's not needed - perhaps rename the lib/handler to lib/apigateway and move this to a new lib/config
Then the (lack of) interdependencies will be enforced by the compiler, and the structure will be clearer.
I'm interested in using something like https://www.pantsbuild.org/ to make the process of creating separate source folders easier, or learning to use sbt better.

johnduffell: of course that could all be in a second PR which depends on a fix for the below comment about the API gateway speciific type!

handlers/digital-subscription-expiry/src/test/scala/com/gu/digitalSubscriptionExpiry/zuora/GetAccountSummaryEffectsTest.scala
```diff
@@ -16,8 +16,8 @@ class GetAccountSummaryEffectsTest extends FlatSpec with Matchers {
     val testAccountId = AccountId("2c92c0f86078c4d4016079e1402d6536")
 
     val actual: \/[io.Serializable, AccountSummaryResult] = for {
-      configAttempt <- ConfigLoad.load(Stage("DEV")).toEither.disjunction
-      config <- Config.parseConfig[StepsConfig](configAttempt)
+      configAttempt <- S3ConfigLoad.load(Stage("DEV")).toEither.disjunction
+      config <- LoadConfig.parseConfig[StepsConfig](configAttempt)
```
johnduffell: good name improvements

johnduffell:
great, good work! 👍 🚀 
It's a shame we have pretty much the same 2 lines in all the effects tests, I'm sure we could tidy that up, but that's nothing new.

PR #89, opened 2018-05-10
Make the catalog service multi-environment 

jacobwinch:
This PR creates a separate lambda function for each environment. This is so that a failing lambda (due to DEV/UAT Zuora downtime) does not affect the PROD service.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -119,8 +119,10 @@ lazy val root = all(project in file(".")).enablePlugins(RiffRaffArtifact).aggreg
   `identity-backfill`,
   `digital-subscription-expiry`,
   `catalog-service`,
+  effects,
+  handler,
+  restHttp,
   zuora,
-  effects
 ).dependsOn(zuora, handler, effectsDepIncludingTestFolder, testDep)
 
 lazy val `identity-backfill` = all(project in file("handlers/identity-backfill")) // when using the "project identity-backfill" command it uses the lazy val name
--- a/handlers/catalog-service/cfn.yaml
+++ b/handlers/catalog-service/cfn.yaml
@@ -150,6 +150,7 @@ Resources:
 
     CatalogS3Bucket:
         Type: "AWS::S3::Bucket"
+        Condition: CreateProdOnlyResources
         Properties:
             AccessControl: Private
             BucketName: gu-zuora-catalog
--- a/handlers/catalog-service/cfn.yaml
+++ b/handlers/catalog-service/cfn.yaml
@@ -14,14 +14,21 @@ Mappings:
     StageVariables:
         CODE:
             ReadableS3Buckets:
+              - arn:aws:s3:::gu-reader-revenue-private/membership/payment-failure-lambdas/DEV/*
               - arn:aws:s3:::gu-reader-revenue-private/membership/payment-failure-lambdas/CODE/*
             WriteableS3Buckets:
               - arn:aws:s3:::gu-zuora-catalog/CODE/*
         PROD:
             ReadableS3Buckets:
+              - arn:aws:s3:::gu-reader-revenue-private/membership/payment-failure-lambdas/DEV/*
+              - arn:aws:s3:::gu-reader-revenue-private/membership/payment-failure-lambdas/CODE/*
               - arn:aws:s3:::gu-reader-revenue-private/membership/payment-failure-lambdas/PROD/*
             WriteableS3Buckets:
               - arn:aws:s3:::gu-zuora-catalog/PROD/*
+
+Conditions:
+    CreateCodeOnlyResources: !Equals [ !Ref Stage, CODE ]
+    CreateProdOnlyResources: !Equals [ !Ref Stage, PROD ]
 
 Resources:
     CatalogServiceRole:
@@ -57,12 +64,13 @@ Resources:
                             Action: s3:PutObject
                             Resource: !FindInMap [StageVariables, !Ref Stage, WriteableS3Buckets]
 
-    CatalogServiceLambda:
+    PRODCatalogServiceLambda:
         Type: AWS::Lambda::Function
+        Condition: CreateProdResources
         Properties:
-            Description: Reads the product catalog from Zuora and stores it in S3
+            Description: Reads the product catalog from PROD Zuora and stores it in S3
             FunctionName:
-                !Sub catalog-service-${Stage}
+                !Sub catalog-service-${Stage}-zuora-prod
             Code:
                 S3Bucket: subscriptions-dist
                 S3Key: !Sub subscriptions/${Stage}/catalog-service/catalog-service.jar
@@ -70,6 +78,7 @@ Resources:
             Environment:
                 Variables:
                   Stage: !Ref Stage
+                  ZuoraEnvironment: PROD
             Role:
                 !GetAtt CatalogServiceRole.Arn
             MemorySize: 1536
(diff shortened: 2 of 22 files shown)


build.sbt
```diff
@@ -119,8 +119,10 @@ lazy val root = all(project in file(".")).enablePlugins(RiffRaffArtifact).aggreg
   `identity-backfill`,
   `digital-subscription-expiry`,
   `catalog-service`,
+  effects,
+  handler,
```
jacobwinch: I noticed that tests weren't being run for handler or restHttp which prompted me to add them here.

lib/restHttp/src/test/scala/com/gu/util/RestRequestMakerTest.scala
```diff
@@ -17,7 +17,7 @@ class RestRequestMakerTest extends AsyncFlatSpec {
   }
 
   "buildRequest" should "set the headers and url correctly" in {
-    val hdr = Map("name" -> "value")
+    val hdr = Map("apiSecretAccessKey" -> "fakePassword")
```
jacobwinch: This was broken, but not being run

handlers/catalog-service/cfn.yaml
```diff
@@ -78,28 +85,64 @@ Resources:
         DependsOn:
         - CatalogServiceRole
 
-    CatalogServiceLambdaInvokePermission:
+    DEVCatalogServiceLambda:
```
pvighi: Probably a stupid question but why don't you have different stacks for DEV, PROD and UAT ?

jacobwinch: We want a production service to (reliably) upload the correct catalogs for each Zuora environment. This means that the apps (and developers running locally) can always trust the catalogs in PROD/ZuoraPROD, PROD/ZuoraUAT and PROD/ZuoraDev, even if someone ships a broken version of the catalog service lambda to CODE.

jacobwinch: It's not a stupid question, I did toy with lots of different ways of doing this - I hated this one the least 😜

pvighi: of course now that I read it it is al little confusing that we use  the name PROD for both the backend and the stage

build.sbt
johnduffell: oops

handlers/catalog-service/cfn.yaml
pvighi: yet another option:
Having just one lambda that takes UAT, PROD or DEV as parameter ?

jacobwinch: I don't think DEV, PROD and two UAT stacks really solves the problem... If we break the DEV catalog, we stop everybody from developing locally (assuming we move to a model where all apps read catalogs from S3), so we still want to read a Zuora DEV catalog that has been generated by a reliable production lambda.

lib/restHttp/src/test/scala/com/gu/util/RestRequestMakerTest.scala
johnduffell: multi project builds are great until... 🙈

handlers/catalog-service/cfn.yaml
jacobwinch: A parameterised lambda is a good suggestion. 

Downsides: it makes monitoring/logging more complicated because you have executions and errors related to non-PROD environments mixed in.

Upsides: we don't have 3 lambdas which are essentially the same.

jacobwinch: Any thoughts @johnduffell?

lib/restHttp/src/test/scala/com/gu/util/RestRequestMakerTest.scala
jacobwinch: When I add new tests, I often make them fail deliberately to check that they're actually being run as expected!

handlers/catalog-service/cfn.yaml
johnduffell: I see it as we can push more division into AWS which gives it more knowledge of our stuff hence the built in logging and isolation are better.  Or we can use AWS more generically and keep the stuff there simpler, and parameterise things which makes them easier to test locally.

Personally I don't like the idea of uploading several times (and having 3x as many lambdas to warm), however I don't think cost and speed are a big issue for this specific lambda.  There is also the clutter issue in the aws console which affects developers using it more.
Another developer issue is ability to have fine grained alerts - we already have issues with false alarms and things grouped too broadly, so being able to do it more separately is quite important.

So I'd go down on the side of making one for each, and making it so the alarm is "if the alert goes off, we need to take action".

Hope that helps!! 😆

handlers/catalog-service/cfn.yaml
```diff
@@ -109,3 +152,50 @@ Resources:
...
+                  ZuoraEnvironment: PROD
+            Role:
+                !GetAtt CatalogServiceRole.Arn
+            MemorySize: 1536
+            Runtime: java8
+            Timeout: 300
+        DependsOn:
+        - CatalogServiceRole
+
+    PRODCatalogServiceLambdaInvokePermission:
```
johnduffell: re DRY, it would have been nice if we could use a "sub/nested"cfn script or something, where we just pass in parameters to create 3 times.  But I'm guessing that isn't easy?

lib/effects/src/main/scala/com/gu/effects/AwsS3.scala
```diff
@@ -24,11 +26,11 @@ object S3ConfigLoad extends Logging {
     load(stage, key)
   }
 
-  def load(stage: Stage, key: String): Try[String] = {
+  def load(stage: Stage, key: String): ConfigFailure \/ String = {
     logger.info(s"Attempting to load config in $stage")
     val bucket = s"gu-reader-revenue-private/membership/payment-failure-lambdas/${stage.value}"
     val request = new GetObjectRequest(bucket, key)
-    GetFromS3.fetchString(request)
+    GetFromS3.fetchString(request).fold(_ => -\/(ConfigFailure("Failed to load config from S3")), conf => \/-(conf))
```
johnduffell: do we need to at least log the exception?

johnduffell: I'd be tempted to pattern match instead of fold, for clarity (but I am not a fold fan!)

handlers/catalog-service/src/main/scala/com/gu/catalogService/Handler.scala
```diff
@@ -19,22 +20,25 @@ object Handler extends Logging {
...
     stage: Stage,
-    s3Load: Stage => Try[String],
+    zuoraEnvironment: ZuoraEnvironment,
+    s3Load: Stage => ConfigFailure \/ String,
     s3Write: PutObjectRequest => Try[PutObjectResult]
   ): Unit = {
 
     val attempt = for {
-      config <- LoadConfig.default[StepsConfig](implicitly)(stage, s3Load(stage)).withLogging("loaded config").leftMap(_.body)
+      config <- LoadConfig.default[StepsConfig](implicitly)(stage, s3Load(zuoraEnvironment.stageToLoad), false)
```
johnduffell: I'm a bit confused by this - so we choose the application stage config based on the zuora stage instead but we skip the check it is correct for the deployment stage (as it's likely not)?  we are really blurring the lines between our stages and zuora stages.  We should still always cross check the config had what we expected, there's no need to skip that as we should always know what we expect to find.

perhaps similar to the standard apps, we should have one section in the file for each zuora config (touch point backend) ie DEV/PROD/UAT and just pick the right part.  Or to avoid duplication, separate files only for zuora config, and load each one as appropriate.  That might make key rotation simpler too if we don't migrate to the secret manager?

lib/handler/src/main/scala/com/gu/util/config/LoadConfig.scala
```diff
@@ -1,43 +1,45 @@
...
     for {
-      textConfig <- s3Load.toFailableOp("load config from s3")
-      config <- parseConfig(textConfig).toFailableOp("parse config file")
-      _ <- if (stage == config.stage)
-        \/-(())
-      else
-        -\/(ApiGatewayResponse.internalServerError(s"running in $stage with config from ${config.stage}"))
+      textConfig <- s3Load
+      config <- parseConfig(textConfig)
+      _ <- safetyCheck(shouldCrossCheckStage, stage, config.stage)
```
johnduffell: assuming we do want to skip the check sometimes (see above comment regarding that), I would prefer to structure this slightly differently.
Instead of passing in a boolean to say whether to do A then B or just A, I'd split out B as you have, but I'd also split out A, and in the "skip" situation just call A directly.  Then it reduces the number of unneeded parameters passed around
In this case, A is s3Load+parseConfig, and B is safetyCheck

lib/restHttp/src/test/scala/com/gu/util/RestRequestMakerTest.scala
johnduffell: yes, that's a great idea and finds this kind of issue more often than you'd hope.  Thanks for spotting that.

handlers/catalog-service/cfn.yaml
jacobwinch: I can't think of a trivial way to do it... one approach would be to have [fragments](https://github.com/guardian/support-workers/blob/master/cloud-formation/src/templates/lambda.yaml) and 'build' them with some sort of script, as we do [here](https://github.com/guardian/support-workers/blob/master/cloud-formation/src/build-cfn.js).

handlers/catalog-service/src/main/scala/com/gu/catalogService/Handler.scala
jacobwinch: The config for this project works on the assumption that:

1. If my stage is DEV, I want Zuora (and SF) DEV
2. If my stage is CODE, I want Zuora (and SF) UAT
3. If my stage is PROD, I want Zuora (and SF) PROD

And we have a separate file for each.

This lambda is the first one which breaks this convention, meaning that the check for 'does my stage match the stage listed at the top of the config file?' is meaningless in this context.

jacobwinch: I agree that this is a bit confusing. Generally the config for this project has become a bit unwieldy as more lambdas have been added to the repo. For example, all lambdas download a file which contains Stripe config, but most of them simply choose not to parse the Stripe values.

Perhaps encrypted environment variables / smaller file fragments / secrets manager would be a better way to manage this going forwards (that way each lambda can be restricted to accessing the most minimal set of config possible)

johnduffell: > This lambda is the first one which breaks this convention, meaning that the check for 'does my stage match the stage listed at the top of the config file?' is meaningless in this context.

Actually, the check is "have I uploaded the wrong file into the wrong directory" so it would still be possible without much change.

Agree about the config being a mess, this is something we need to deal with in the medium term.  Definitely things that don't need stripe (etc) should be forbidden from accessing it.

handlers/catalog-service/cfn.yaml
johnduffell: I was meaning more like this: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/using-cfn-nested-stacks.html

But I don't know if that is heavy enough to be over kill for us.  I did try to do that for dotcom back in the day, and got most of the way there, but it was a bit cumbersome then.  Leave it for now, at least it's divided well between lambdas (even though there's duplication between them.

handlers/catalog-service/src/main/scala/com/gu/catalogService/Handler.scala
jacobwinch: Ah thanks - I'd misunderstood what the check was trying to do!

johnduffell:
go for it! 🚀 👍

johnduffell:
PS the codacy is just extra duplication, we should keep that in mind as tech debt and come back to it.  Coverage is up due to the tests actually being run.

PR #91, opened 2018-05-17
Handle GET requests

jacobwinch:
So far all of our API Gateway > Lambda wiring assumes that the request is a POST. This PR allows us to cope with requests which don't contain a body.

### DIFF
--- a/handlers/digital-subscription-expiry/src/main/scala/com/gu/digitalSubscriptionExpiry/DigitalSubscriptionExpirySteps.scala
+++ b/handlers/digital-subscription-expiry/src/main/scala/com/gu/digitalSubscriptionExpiry/DigitalSubscriptionExpirySteps.scala
@@ -1,20 +1,16 @@
 package com.gu.digitalSubscriptionExpiry
 
-import com.gu.digitalSubscriptionExpiry.common.CommonApiResponses._
+import com.gu.digitalSubscriptionExpiry.common.CommonApiResponses
 import com.gu.digitalSubscriptionExpiry.zuora.GetAccountSummary.{AccountId, AccountSummaryResult}
 import com.gu.digitalSubscriptionExpiry.zuora.GetSubscription.{SubscriptionId, SubscriptionResult}
 import com.gu.util.Logging
 import com.gu.util.apigateway.ApiGatewayHandler.Operation
 import com.gu.util.apigateway.{ApiGatewayRequest, URLParams}
-import com.gu.util.reader.Types.{FailableOp, _}
+import com.gu.util.reader.Types._
 import main.scala.com.gu.digitalSubscriptionExpiry.DigitalSubscriptionExpiryRequest
-import play.api.libs.json.{JsValue, Json}
 import scalaz.\/-
 
-import scala.util.Try
 object DigitalSubscriptionExpirySteps extends Logging {
-
-  def parseJson(input: String): Option[JsValue] = Try(Json.parse(input)).toOption
 
   def apply(
     getEmergencyTokenExpiry: String => FailableOp[Unit],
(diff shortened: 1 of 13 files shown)


lib/handler/src/main/scala/com/gu/util/apigateway/ApiGatewayRequest.scala
```diff
@@ -28,13 +33,31 @@ case class URLParams(
...
+case class ApiGatewayRequest(queryStringParameters: Option[URLParams], body: Option[String], headers: Option[Map[String, String]]) {
+
   def onlyCancelDirectDebit: Boolean = queryStringParameters.exists(_.onlyCancelDirectDebit)
   def requestAuth: Option[RequestAuth] =
     for {
       queryStringParameters <- queryStringParameters
       apiToken <- queryStringParameters.apiToken
     } yield RequestAuth(apiToken)
+
+  def parseBody[A](failureResponse: ApiResponse = ApiGatewayResponse.badRequest)(implicit reads: Reads[A]): FailableOp[A] = {
```
jacobwinch: We allow a failure response to be passed all of the way through, because some lambdas (digital subscription expiry) return bespoke 400 responses

lib/handler/src/main/scala/com/gu/util/apigateway/ApiGatewayRequest.scala
```diff
@@ -28,13 +33,31 @@ case class URLParams(
 /* Using query strings because for Basic Auth to work Zuora requires us to return a WWW-Authenticate
   header, and API Gateway does not support this header (returns x-amzn-Remapped-WWW-Authenticate instead)
   */
-case class ApiGatewayRequest(queryStringParameters: Option[URLParams], body: String, headers: Option[Map[String, String]]) {
+case class ApiGatewayRequest(queryStringParameters: Option[URLParams], body: Option[String], headers: Option[Map[String, String]]) {
+
   def onlyCancelDirectDebit: Boolean = queryStringParameters.exists(_.onlyCancelDirectDebit)
```
pvighi: this is unrelated to this PR but at some point we should probably move this def somewhere else since this class is supposed to be generic now.

jacobwinch: 👍- we probably need several PRs which just re-structure this project and add no new functionality

src/main/scala/com/gu/stripeCustomerSourceUpdated/SourceUpdatedSteps.scala
```diff
@@ -27,7 +27,7 @@ object SourceUpdatedSteps extends Logging {
   def apply(zuoraRequests: Requests, stripeDeps: StripeDeps): Operation = Operation.noHealthcheck({ apiGatewayRequest: ApiGatewayRequest =>
     for {
       _ <- if (stripeDeps.config.signatureChecking) bodyIfSignatureVerified(stripeDeps, apiGatewayRequest) else \/-(())
-      sourceUpdatedCallout <- Json.fromJson[SourceUpdatedCallout](Json.parse(apiGatewayRequest.body)).toFailableOp.withLogging("fromJson SourceUpdatedCallout")
+      sourceUpdatedCallout <- apiGatewayRequest.parseBody[SourceUpdatedCallout]()
       _ = logger.info(s"from: ${apiGatewayRequest.queryStringParameters.map(_.stripeAccount)}")
```
pvighi: I'm not exactly how this is supposed to work but wouldn't the body be parsed  when bodyIfSignatureVerified is called but the result ignored and then parsed again in the next line?

jacobwinch: Yes, thanks, that's much cleaner 👍

handlers/digital-subscription-expiry/src/main/scala/com/gu/digitalSubscriptionExpiry/DigitalSubscriptionExpirySteps.scala
```diff
@@ -27,14 +23,13 @@ object DigitalSubscriptionExpirySteps extends Logging {
...
-        jsonRequest <- parseJson(apiGatewayRequest.body).toFailableOp(badRequest)
-        expiryRequest <- Json.fromJson[DigitalSubscriptionExpiryRequest](jsonRequest).asOpt.toFailableOp(badRequest)
+        expiryRequest <- apiGatewayRequest.parseBody[DigitalSubscriptionExpiryRequest](CommonApiResponses.badRequest)
         _ <- getEmergencyTokenExpiry(expiryRequest.subscriberId)
         subscriptionId = SubscriptionId(expiryRequest.subscriberId.trim.dropWhile(_ == '0'))
         subscriptionResult <- getSubscription(subscriptionId)
         _ <- if (skipActivationDateUpdate(apiGatewayRequest.queryStringParameters, subscriptionResult)) \/-(()) else setActivationDate(subscriptionResult.id)
         accountSummary <- getAccountSummary(subscriptionResult.accountId)
-        password <- expiryRequest.password.toFailableOp(notFoundResponse)
+        password <- expiryRequest.password.toFailableOp(CommonApiResponses.notFoundResponse)
```
johnduffell: `#$%*($#ApiResponses` ? 😱

johnduffell: someone sneaked in a "Common" when I wasn't looking

lib/handler/src/main/scala/com/gu/util/apigateway/ApiGatewayRequest.scala
```diff
@@ -28,13 +33,31 @@ case class URLParams(
...
+        Try(Json.parse(requestBody)) match {
+          case Success(js) =>
+            Json.fromJson[A](js).toFailableOp(failureResponse)
+          case Failure(ex) =>
+            logger.warn(s"Tried to parse JSON but it was invalid")
+            -\/(failureResponse)
+        }
+      case None =>
+        logger.warn(s"Attempted to access response body but there was none")
+        None.toFailableOp(ApiGatewayResponse.internalServerError("attempted to parse body when handling a GET request"))
```
johnduffell: random thought is it would be nice if this was a compile error, but on its own it shouldn't confuse people too much.

johnduffell:
looks good, I'd prefer the json stuff not to be in the api gateway request object if that's simple to do? as that's the model we're deserialising and json parsing seems a bit heavy weight, but looks good and great work splitting into small PR 👍

PR #92, opened 2018-05-18
Add minimal logic for identity-retention API

jacobwinch:
This allows us to search Zuora and confirm if an identity id is linked to an account.

Currently we serve a 200 if we find the user, and a 404 if we don't. In the future we need to have different classes of success (e.g. we queried successfully and found the user, but all of their accounts are cancelled OR we queried Zuora and they still have an active account). Currently we haven't got any cancelled Zuora accounts, so I'd prefer to get something that the identity team can use into production before worrying too much about all of the different cases.

All of the response bodies still need to be JSONified - that's coming in the next PR.

### DIFF
--- a/handlers/identity-retention/cfn.yaml
+++ b/handlers/identity-retention/cfn.yaml
@@ -86,7 +86,7 @@ Resources:
         Properties:
             RestApiId: !Ref IdentityRetentionAPI
             ParentId: !GetAtt [IdentityRetentionAPI, RootResourceId]
-            PathPart: identity-retention
+            PathPart: user
         DependsOn: IdentityRetentionAPI
 
     IdentityRetentionMethod:
@@ -97,6 +97,8 @@ Resources:
             RestApiId: !Ref IdentityRetentionAPI
             ResourceId: !Ref IdentityRetentionProxyResource
             HttpMethod: GET
+            RequestParameters:
+              method.request.querystring.identityId: true
             Integration:
               Type: AWS_PROXY
               IntegrationHttpMethod: POST
--- a/handlers/identity-retention/cfn.yaml
+++ b/handlers/identity-retention/cfn.yaml
@@ -86,7 +86,7 @@ Resources:
         Properties:
             RestApiId: !Ref IdentityRetentionAPI
             ParentId: !GetAtt [IdentityRetentionAPI, RootResourceId]
-            PathPart: user
+            PathPart: /retention-status
         DependsOn: IdentityRetentionAPI
 
     IdentityRetentionMethod:
--- a/handlers/identity-retention/src/main/scala/com.gu.identityRetention/ActiveZuoraCheck.scala
+++ b/handlers/identity-retention/src/main/scala/com.gu.identityRetention/ActiveZuoraCheck.scala
@@ -1,0 +1,33 @@
+package com.gu.identityRetention
+
+import com.gu.util.apigateway.ApiGatewayResponse
+import com.gu.util.reader.Types.FailableOp
+import com.gu.util.zuora.ZuoraQuery
+import com.gu.util.zuora.ZuoraQuery.ZuoraQuerier
+import play.api.libs.json.Json
+import scalaz.{-\/, \/-}
+
+object ActiveZuoraCheck {
+
+  case class IdentityQueryResponse(Id: String, Status: String)
+  implicit val reads = Json.reads[IdentityQueryResponse]
+
+  def apply(identityId: String, zuoraQuerier: ZuoraQuerier): FailableOp[Unit] = {
+
+    def searchForAccounts = {
+      val identityQuery = ZuoraQuery.Query(s"select id, status from account where IdentityId__c = '$identityId'")
+      zuoraQuerier[IdentityQueryResponse](identityQuery)
+    }
+
+    searchForAccounts match {
+      case \/-(result) if result.size > 0 =>
+        \/-(())
+      case \/-(result) if result.size == 0 =>
+        -\/(ApiGatewayResponse.notFound("Identity user has no linked Zuora accounts"))
+      case -\/(error) =>
+        -\/(ApiGatewayResponse.internalServerError("Failed to retrieve the identity user's from Zuora"))
+    }
+
+  }
+
+}
--- a/handlers/identity-retention/src/main/scala/com.gu.identityRetention/ActiveZuoraCheck.scala
+++ b/handlers/identity-retention/src/main/scala/com.gu.identityRetention/ActiveZuoraCheck.scala
@@ -25,7 +25,7 @@ object ActiveZuoraCheck {
       case \/-(result) if result.size == 0 =>
         -\/(ApiGatewayResponse.notFound("Identity user has no linked Zuora accounts"))
       case -\/(error) =>
-        -\/(ApiGatewayResponse.internalServerError("Failed to retrieve the identity user's from Zuora"))
+        -\/(ApiGatewayResponse.internalServerError("Failed to retrieve the identity user's details from Zuora"))
     }
 
   }
--- a/handlers/identity-retention/src/main/scala/com.gu.identityRetention/Handler.scala
+++ b/handlers/identity-retention/src/main/scala/com.gu.identityRetention/Handler.scala
@@ -8,7 +8,7 @@ import com.gu.util.apigateway.ApiGatewayHandler.{LambdaIO, Operation}
 import com.gu.util.config.ConfigReads.ConfigFailure
 import com.gu.util.config.{Config, LoadConfig, Stage}
 import com.gu.util.reader.Types._
-import com.gu.util.zuora.ZuoraRestConfig
+import com.gu.util.zuora.{ZuoraQuery, ZuoraRestConfig, ZuoraRestRequestMaker}
 import okhttp3.{Request, Response}
 import play.api.libs.json.{Json, Reads}
 import scalaz.\/
@@ -26,14 +26,18 @@ object Handler {
 
   def runWithEffects(response: Request => Response, stage: Stage, s3Load: Stage => ConfigFailure \/ String, lambdaIO: LambdaIO) = {
 
-    def operation: Config[StepsConfig] => Operation = config => IdentityRetentionSteps.apply
+    def operation: Config[StepsConfig] => Operation = config => {
+      val zuoraRequests = ZuoraRestRequestMaker(response, config.stepsConfig.zuoraRestConfig)
+      val zuoraQuerier = ZuoraQuery(zuoraRequests)
+      IdentityRetentionSteps.apply(zuoraQuerier)
+    }
 
     ApiGatewayHandler[StepsConfig](lambdaIO)(for {
       config <- LoadConfig.default[StepsConfig](implicitly)(stage, s3Load(stage))
         .toFailableOp("load config")
       configuredOp = operation(config)
-
     } yield (config, configuredOp))
+
   }
 
 }
--- a/handlers/identity-retention/src/main/scala/com.gu.identityRetention/IdentityRetentionSteps.scala
+++ b/handlers/identity-retention/src/main/scala/com.gu.identityRetention/IdentityRetentionSteps.scala
@@ -11,8 +11,10 @@ object IdentityRetentionSteps extends Logging {
 
   def apply(zuoraQuerier: ZuoraQuerier): Operation = Operation.noHealthcheck({
     apiGatewayRequest: ApiGatewayRequest =>
-      extractIdentityId(apiGatewayRequest.queryStringParameters)
-      ActiveZuoraCheck.apply("123", zuoraQuerier)
+      for {
+        identityId <- extractIdentityId(apiGatewayRequest.queryStringParameters)
+        _ <- ActiveZuoraCheck.apply(identityId, zuoraQuerier)
+      } yield ()
   }, false)
 
   def extractIdentityId(queryStringParams: Option[URLParams]): FailableOp[String] = {
--- a/handlers/identity-retention/src/main/scala/com.gu.identityRetention/IdentityRetentionSteps.scala
+++ b/handlers/identity-retention/src/main/scala/com.gu.identityRetention/IdentityRetentionSteps.scala
@@ -2,17 +2,28 @@ package com.gu.identityRetention
 
 import com.gu.util.Logging
 import com.gu.util.apigateway.ApiGatewayHandler.Operation
-import com.gu.util.apigateway.ApiGatewayRequest
-import com.gu.util.reader.Types._
-import scala.util.Success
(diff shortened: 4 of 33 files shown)


handlers/identity-retention/cfn.yaml
```diff
@@ -97,6 +97,8 @@ Resources:
             RestApiId: !Ref IdentityRetentionAPI
             ResourceId: !Ref IdentityRetentionProxyResource
             HttpMethod: GET
```
johnduffell: if this is GET, should we be thinking about cache headers? (probably not, but I thought I'd put the thought there)

handlers/identity-retention/src/main/scala/com.gu.identityRetention/ActiveZuoraCheck.scala
```diff
@@ -0,0 +1,33 @@
+package com.gu.identityRetention
```
johnduffell: the directory is called com.gu.identityRetention, normally it would be 3 nested folders.  is this right to do it that way?

handlers/identity-retention/src/main/scala/com.gu.identityRetention/ActiveZuoraCheck.scala
```diff
@@ -0,0 +1,33 @@
...
+import com.gu.util.zuora.ZuoraQuery.ZuoraQuerier
+import play.api.libs.json.Json
+import scalaz.{-\/, \/-}
+
+object ActiveZuoraCheck {
+
+  case class IdentityQueryResponse(Id: String, Status: String)
+  implicit val reads = Json.reads[IdentityQueryResponse]
+
+  def apply(identityId: String, zuoraQuerier: ZuoraQuerier): FailableOp[Unit] = {
```
johnduffell: try
`def apply(zuoraQuerier: ZuoraQuerier)(identityId: String): FailableOp[Unit]`
then you can not have the "steps" having to worry about ZuoraQuerier, they just need to know a way to do the check. `String => FailableOp[Unit]` where the string is an IdentityId.

handlers/identity-retention/src/main/scala/com.gu.identityRetention/ActiveZuoraCheck.scala
```diff
@@ -0,0 +1,33 @@
+package com.gu.identityRetention
+
+import com.gu.util.apigateway.ApiGatewayResponse
+import com.gu.util.reader.Types.FailableOp
+import com.gu.util.zuora.ZuoraQuery
+import com.gu.util.zuora.ZuoraQuery.ZuoraQuerier
+import play.api.libs.json.Json
+import scalaz.{-\/, \/-}
+
+object ActiveZuoraCheck {
```
johnduffell: what is the check really doing?  can we use a name such as `HasActiveZuoraAccounts`?

handlers/identity-retention/cfn.yaml
```diff
@@ -86,7 +86,7 @@ Resources:
         Properties:
             RestApiId: !Ref IdentityRetentionAPI
             ParentId: !GetAtt [IdentityRetentionAPI, RootResourceId]
-            PathPart: identity-retention
+            PathPart: user
```
johnduffell: not totally keen on either of these names, how about /user-outside-retention?identityId=2345 or something a bit more purpose decided?

handlers/identity-retention/src/main/scala/com.gu.identityRetention/ActiveZuoraCheck.scala
```diff
@@ -0,0 +1,33 @@
...
+
+  case class IdentityQueryResponse(Id: String, Status: String)
+  implicit val reads = Json.reads[IdentityQueryResponse]
+
+  def apply(identityId: String, zuoraQuerier: ZuoraQuerier): FailableOp[Unit] = {
+
+    def searchForAccounts = {
```
johnduffell: would be ok to have this as a val, but either was as you wish

handlers/identity-retention/src/main/scala/com.gu.identityRetention/ActiveZuoraCheck.scala
```diff
@@ -0,0 +1,33 @@
...
+      val identityQuery = ZuoraQuery.Query(s"select id, status from account where IdentityId__c = '$identityId'")
+      zuoraQuerier[IdentityQueryResponse](identityQuery)
+    }
+
+    searchForAccounts match {
```
johnduffell: this match statement is pretty simple logic, but if it were actually "doing" something that you might want to test, I'd make it a separate Step.

handlers/identity-retention/src/main/scala/com.gu.identityRetention/Handler.scala
```diff
@@ -26,14 +26,18 @@ object Handler {
 
   def runWithEffects(response: Request => Response, stage: Stage, s3Load: Stage => ConfigFailure \/ String, lambdaIO: LambdaIO) = {
 
-    def operation: Config[StepsConfig] => Operation = config => IdentityRetentionSteps.apply
+    def operation: Config[StepsConfig] => Operation = config => {
+      val zuoraRequests = ZuoraRestRequestMaker(response, config.stepsConfig.zuoraRestConfig)
+      val zuoraQuerier = ZuoraQuery(zuoraRequests)
+      IdentityRetentionSteps.apply(zuoraQuerier)
```
johnduffell: no need the ".apply"

handlers/identity-retention/src/main/scala/com.gu.identityRetention/IdentityRetentionSteps.scala
```diff
@@ -2,17 +2,30 @@ package com.gu.identityRetention
...
+import scalaz.{-\/, \/-}
 
 object IdentityRetentionSteps extends Logging {
 
-  def apply: Operation = Operation.noHealthcheck(steps, false)
+  def apply(zuoraQuerier: ZuoraQuerier): Operation = Operation.noHealthcheck({
+    apiGatewayRequest: ApiGatewayRequest =>
+      for {
+        identityId <- extractIdentityId(apiGatewayRequest.queryStringParameters)
+        _ <- ActiveZuoraCheck.apply(identityId, zuoraQuerier)
```
johnduffell: can remove the ".apply" here as well

handlers/identity-retention/src/main/scala/com.gu.identityRetention/IdentityRetentionSteps.scala
```diff
@@ -2,17 +2,30 @@ package com.gu.identityRetention
...
-  def steps(apiGatewayRequest: ApiGatewayRequest): FailableOp[Unit] = {
-    logger.info("Running fake identity retention steps")
-    Success(()).toFailableOp("always succeeds")
+  def extractIdentityId(queryStringParams: Option[URLParams]): FailableOp[String] = {
+    val identityId = for {
+      queryStrings <- queryStringParams
+      id <- queryStrings.identityId
+    } yield id
+    identityId match {
+      case Some(user) => \/-(user)
```
johnduffell: could consider wrapping the String into an IdentityId object to improve the type safety (at the expense of a little more code)

handlers/identity-retention/src/test/scala/com/gu/identityRetention/ActiveZuoraCheckTest.scala
```diff
@@ -0,0 +1,83 @@
...
+class ActiveZuoraCheckTest extends FlatSpec with Matchers {
+
+  val noZuoraAccounts = QueryResult[IdentityQueryResponse](Nil, 0, true, None)
+
+  val singleZuoraAccount = QueryResult[IdentityQueryResponse](List(IdentityQueryResponse("123", "Active")), 1, true, None)
+
+  it should "return a left(404) if the identity id is not linked to any Zuora accounts" in {
+    val effects = new TestingRawEffects(postResponses = ZuoraQueryMocks.postResponse(Nil))
+    val requestMaker = ZuoraRestRequestMaker(effects.response, ZuoraRestConfig("https://zuora", "user", "pass"))
+    val zuoraQuerier = ZuoraQuery(requestMaker)
```
johnduffell: could simplify this (unit) test by mocking a ZuoraQuery rather than the whole Request->Response and everything between

handlers/identity-retention/src/test/scala/com/gu/identityRetention/IdentityRetentionStepsTest.scala
```diff
@@ -0,0 +1,21 @@
+package com.gu.identityRetention
+
+import com.gu.util.apigateway.{ApiGatewayResponse, URLParams}
+import org.scalatest.{FlatSpec, Matchers}
+import scalaz.{-\/, \/-}
+
+class IdentityRetentionStepsTest extends FlatSpec with Matchers {
+
+  it should "return a bad response if the identity id query string cannot be found" in {
+    val result = IdentityRetentionSteps.extractIdentityId(Some(URLParams(None, false, None, false, false, None)))
```
johnduffell: When writing a test, it's a bit of a line in the sand if you have to pass in something not needed for the test.  As it makes the tests harder to understand and means the code has deps it shouldn't have.
  So we should not really have to pass in all those extra values just to keep the compiler happy.  I guess this is kind of type wrangling code, maybe  it really means we need to fix the URLParams system instead.

lib/handler/src/main/scala/com/gu/util/apigateway/ApiGatewayRequest.scala
```diff
@@ -27,7 +27,8 @@ case class URLParams(
   onlyCancelDirectDebit: Boolean,
   stripeAccount: Option[StripeAccount],
   isHealthcheck: Boolean,
-  noActivation: Boolean
+  noActivation: Boolean,
+  identityId: Option[String]
```
johnduffell: I have added sorting the url params to our list of tech debt 😁

handlers/identity-retention/src/main/scala/com.gu.identityRetention/ActiveZuoraCheck.scala
```diff
@@ -0,0 +1,33 @@
...
+
+object ActiveZuoraCheck {
+
+  case class IdentityQueryResponse(Id: String, Status: String)
+  implicit val reads = Json.reads[IdentityQueryResponse]
+
+  def apply(identityId: String, zuoraQuerier: ZuoraQuerier): FailableOp[Unit] = {
+
+    def searchForAccounts = {
+      val identityQuery = ZuoraQuery.Query(s"select id, status from account where IdentityId__c = '$identityId'")
```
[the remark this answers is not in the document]
johnduffell: excellent spot!
we are concatenating untrusted and trusted data into a higher level command that runs with privileges not otherwise available, classic security hole!  We need a proper way to build (zuora) queries that doesn't involve string interpolation.
We should review the rest of the code in case we did that elsewhere.
Lucky none of it is used by untrusted external actors (due to api keys)

handlers/identity-retention/cfn.yaml
jacobwinch: I'd agree with the probably not - I'm not sure we'd gain much from caching here...

handlers/identity-retention/src/main/scala/com.gu.identityRetention/ActiveZuoraCheck.scala
jacobwinch: Nope, I made a mistake

handlers/identity-retention/src/main/scala/com.gu.identityRetention/ActiveZuoraCheck.scala
jacobwinch: Good point - this was one of those WIP names that never got revisited!

handlers/identity-retention/src/main/scala/com.gu.identityRetention/ActiveZuoraCheck.scala
jacobwinch: Yep, I think when we start processing the contents of the query (e.g. looking at Active and Cancelled accounts), this will warrant its own step

handlers/identity-retention/cfn.yaml
jacobwinch: 👍

handlers/identity-retention/src/main/scala/com.gu.identityRetention/IdentityRetentionSteps.scala
jacobwinch: thanks 👍

handlers/identity-retention/src/main/scala/com.gu.identityRetention/ActiveZuoraCheck.scala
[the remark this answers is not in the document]
jacobwinch: Great spot @pvighi - I've added some basic protection [here](https://github.com/guardian/zuora-auto-cancel/pull/92/commits/70ab91f9157de05cb0c59eb956c40b3eb8e1ce44#diff-9fc36403376f06fb34dc749091833d8aR34).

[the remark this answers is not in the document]
johnduffell: good improvement, I have added the general case as tech debt

handlers/identity-retention/src/main/scala/com/gu/identityRetention/HasActiveZuoraAccounts.scala
```diff
@@ -0,0 +1,34 @@
...
+      zuoraQuerier[IdentityQueryResponse](identityQuery)
+    }
+
+    searchForAccounts match {
+      case \/-(result) if result.size > 0 =>
+        \/-(())
+      case \/-(result) if result.size == 0 =>
+        -\/(ApiGatewayResponse.notFound("Identity user has no linked Zuora accounts"))
+      case -\/(error) =>
+        -\/(ApiGatewayResponse.internalServerError("Failed to retrieve the identity user's details from Zuora"))
```
johnduffell: do we need to log `error` here?

johnduffell:
looks good, the only thing I'd really like to see addressed before shipping is the test that's a lot more complicated than it needs to be - not sure if the comment was missed: https://github.com/guardian/zuora-auto-cancel/pull/92#discussion_r189333497
apart from that it is good for me 👍

handlers/identity-retention/src/main/scala/com/gu/identityRetention/HasActiveZuoraAccounts.scala
jacobwinch: On the version you reviewed, yes, but now that I've merged in master this will get logged [here](https://github.com/guardian/zuora-auto-cancel/blob/master/lib/handler/src/main/scala/com/gu/util/apigateway/ApiGatewayResponse.scala#L78).

PR #93, opened 2018-05-21
Return JSON response bodies

jacobwinch:
As we'll need this for the identity team's retention lookup / API call...

Also note that more specific messages about why the operation failed are now logged, rather than returned in the response body (it seemed a bit odd that we were leaking details of the failed operation via the response).

This PR also renames the 'common' package/objects in Digital Subscription Expiry.

### DIFF
--- a/handlers/digital-subscription-expiry/src/main/scala/com/gu/digitalSubscriptionExpiry/DigitalSubscriptionExpirySteps.scala
+++ b/handlers/digital-subscription-expiry/src/main/scala/com/gu/digitalSubscriptionExpiry/DigitalSubscriptionExpirySteps.scala
@@ -1,6 +1,6 @@
 package com.gu.digitalSubscriptionExpiry
 
-import com.gu.digitalSubscriptionExpiry.common.CommonApiResponses
+import com.gu.digitalSubscriptionExpiry.responses.DigitalSubscriptionApiResponses
 import com.gu.digitalSubscriptionExpiry.zuora.GetAccountSummary.{AccountId, AccountSummaryResult}
 import com.gu.digitalSubscriptionExpiry.zuora.GetSubscription.{SubscriptionId, SubscriptionResult}
 import com.gu.util.Logging
@@ -23,13 +23,13 @@ object DigitalSubscriptionExpirySteps extends Logging {
 
     def steps(apiGatewayRequest: ApiGatewayRequest): FailableOp[Unit] = {
       for {
-        expiryRequest <- apiGatewayRequest.bodyAsCaseClass[DigitalSubscriptionExpiryRequest](CommonApiResponses.badRequest)
+        expiryRequest <- apiGatewayRequest.bodyAsCaseClass[DigitalSubscriptionExpiryRequest](DigitalSubscriptionApiResponses.badRequest)
         _ <- getEmergencyTokenExpiry(expiryRequest.subscriberId)
         subscriptionId = SubscriptionId(expiryRequest.subscriberId.trim.dropWhile(_ == '0'))
         subscriptionResult <- getSubscription(subscriptionId)
         _ <- if (skipActivationDateUpdate(apiGatewayRequest.queryStringParameters, subscriptionResult)) \/-(()) else setActivationDate(subscriptionResult.id)
         accountSummary <- getAccountSummary(subscriptionResult.accountId)
-        password <- expiryRequest.password.toFailableOp(CommonApiResponses.notFoundResponse)
+        password <- expiryRequest.password.toFailableOp(DigitalSubscriptionApiResponses.notFoundResponse)
         subscriptionEndDate <- getSubscriptionExpiry(password, subscriptionResult, accountSummary)
       } yield {}
 
--- a/handlers/identity-backfill/src/test/scala/com/gu/identityBackfill/EndToEndHandlerTest.scala
+++ b/handlers/identity-backfill/src/test/scala/com/gu/identityBackfill/EndToEndHandlerTest.scala
@@ -21,9 +21,13 @@ class EndToEndHandlerTest extends FlatSpec with Matchers {
     val (responseString, requests): (String, List[TestingRawEffects.BasicRequest]) = getResultAndRequests(identityBackfillRequest(true))
 
     val expectedResponse =
-      s"""
-         |{"statusCode":"200","headers":{"Content-Type":"application/json"},"body":"Processing is not required: DRY RUN requested! skipping to the end"}
-         |""".stripMargin
+      """{
+        |"statusCode":"200",
+        |"headers":{"Content-Type":"application/json"},
+        |"body":"{\n  \"message\" : \"Processing is not required: DRY RUN requested! skipping to the end\"\n}"
+        |}
+        |""".stripMargin
+
     responseString jsonMatches expectedResponse
     requests should be(List(
       BasicRequest("GET", "/services/data/v20.0/sobjects/Contact/00110000011AABBAAB", ""),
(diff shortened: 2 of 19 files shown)


lib/handler/src/main/scala/com/gu/util/apigateway/ApiGatewayResponse.scala
```diff
@@ -43,13 +43,49 @@ object ApiGatewayResponse extends Logging {
...
   }
 
-  val successfulExecution = ApiResponse("200", new Headers, "Success")
-  def noActionRequired(reason: String) = ApiResponse("200", new Headers, s"Processing is not required: $reason")
-
-  val unauthorized = ApiResponse("401", new Headers, "Credentials are missing or invalid")
-  def notFound(message: String) = ApiResponse("404", new Headers, message)
-  def badRequestWithBody(body: String) = ApiResponse("400", new Headers, body)
-  val badRequest = badRequestWithBody("Failure to parse JSON successfully")
-  def internalServerError(error: String) = ApiResponse("500", new Headers, s"Failed to process event due to the following error: $error")
```
jacobwinch: The error that's passed through here is now logged, rather than returned in the response body.

handlers/digital-subscription-expiry/src/main/scala/com/gu/digitalSubscriptionExpiry/DigitalSubscriptionExpirySteps.scala
```diff
@@ -1,6 +1,6 @@
 package com.gu.digitalSubscriptionExpiry
 
-import com.gu.digitalSubscriptionExpiry.common.CommonApiResponses
+import com.gu.digitalSubscriptionExpiry.responses.DigitalSubscriptionApiResponses
```
johnduffell: 😄

handlers/identity-backfill/src/test/scala/com/gu/identityBackfill/EndToEndHandlerTest.scala
```diff
@@ -21,9 +21,13 @@ class EndToEndHandlerTest extends FlatSpec with Matchers {
...
     val expectedResponse =
-      s"""
-         |{"statusCode":"200","headers":{"Content-Type":"application/json"},"body":"Processing is not required: DRY RUN requested! skipping to the end"}
-         |""".stripMargin
+      """{
+        |"statusCode":"200",
+        |"headers":{"Content-Type":"application/json"},
+        |"body":"{\n  \"message\" : \"Processing is not required: DRY RUN requested! skipping to the end\"\n}"
+        |}
+        |""".stripMargin
```
johnduffell: much better

lib/handler/src/main/scala/com/gu/util/apigateway/ApiGatewayResponse.scala
johnduffell: good spot.  this is another case of trusting the caller too much, which is not a sustainable assumption

lib/handler/src/main/scala/com/gu/util/apigateway/ApiGatewayResponse.scala
```diff
@@ -43,13 +43,49 @@ object ApiGatewayResponse extends Logging {
...
+  case class ResponseBody(message: String)
+
+  implicit val responseBodyWrites = Json.writes[ResponseBody]
+
+  def toJsonBody(responseBody: ResponseBody) = Json.prettyPrint(Json.toJson(responseBody))
+
+  val successfulExecution = ApiResponse(
+    "200",
+    new Headers,
```
johnduffell: do we ever use the `new Headers` for anything other than empty?  I feel like they just hang around

lib/handler/src/main/scala/com/gu/util/apigateway/ApiGatewayResponse.scala
```diff
@@ -43,13 +43,49 @@ object ApiGatewayResponse extends Logging {
...
-  val unauthorized = ApiResponse("401", new Headers, "Credentials are missing or invalid")
-  def notFound(message: String) = ApiResponse("404", new Headers, message)
-  def badRequestWithBody(body: String) = ApiResponse("400", new Headers, body)
-  val badRequest = badRequestWithBody("Failure to parse JSON successfully")
-  def internalServerError(error: String) = ApiResponse("500", new Headers, s"Failed to process event due to the following error: $error")
+  case class ResponseBody(message: String)
+
+  implicit val responseBodyWrites = Json.writes[ResponseBody]
+
+  def toJsonBody(responseBody: ResponseBody) = Json.prettyPrint(Json.toJson(responseBody))
```
johnduffell: not a big issue to pretty print, although it will make the responses bigger, and any decent browser or postman should be able to format it nicely even if it's not sent nicely

src/main/scala/com/gu/paymentFailure/PaymentFailureSteps.scala
```diff
@@ -57,7 +56,7 @@ object ZuoraEmailSteps {
       invoiceTransactionSummary <- getInvoiceTransactions(accountId).leftMap(ZuoraToApiGateway.fromClientFail)
       paymentInformation <- GetPaymentData(accountId)(invoiceTransactionSummary)
       message = toMessage(paymentInformation)
-      _ <- sendEmail(message).leftMap(resp => resp.copy(body = s"email not sent for account ${accountId}"))
+      _ <- sendEmail(message).leftMap(_ => internalServerError(s"email not sent for account ${accountId}"))
```
johnduffell: I think this is assuming too much that all Left are an error, we should make sendEmail to return something other than a FailableOp e.g. an either[string, unit]

src/test/scala/com/gu/TestData.scala
```diff
@@ -39,8 +39,25 @@ object TestData extends Matchers {
     stripeConfig = StripeConfig(customerSourceUpdatedWebhook = StripeWebhook(ukStripeSecretKey = StripeSecretKey("abc"), auStripeSecretKey = StripeSecretKey("def")), true)
   )
 
-  val missingCredentialsResponse = """{"statusCode":"401","headers":{"Content-Type":"application/json"},"body":"Credentials are missing or invalid"}"""
-  val successfulResponse = """{"statusCode":"200","headers":{"Content-Type":"application/json"},"body":"Success"}"""
+  val missingCredentialsResponse = """{
```
johnduffell: interesting indentation, I'd be tempted to go for 2 space if it's multi line where possible, something more along the lines of
```scala
  val missingCredentialsResponse =
    """{
      |"statusCode":"401",
      |"headers":{"Content-Type":"application/json"},
      |"body":"{\n  \"message\" : \"Credentials are missing or invalid\"\n}"
      |}
      |""".stripMargin
  
```
etc

johnduffell:
great stuff, a few comments for you, the only one I'd like to fix before shipping is the one where we assume all Left from sendEmail are an internal server error.  Just change it to return Either[String,Unit] would be a sensible fix I guess

johnduffell:
👍

jacobwinch:
Thanks for the comments @johnduffell - I've addressed the majority of them (I've left prettyPrinting as it is for now).

src/main/scala/com/gu/paymentFailure/PaymentFailureSteps.scala
jacobwinch: Good catch - I hadn't realised I was enforcing this assumption. I decided to preserve the existing behaviour in the end. We now wrap the message in a ResponseBody in order to return valid JSON, but retain the response code from sendEmail.

johnduffell:
great work!

PR #94, opened 2018-05-22
Fix CFN mistake

jacobwinch:
The "/" breaks CFN updates for this stack.


johnduffell:
we have CODE env  (if only it was easier to use)

jacobwinch:
So far this only exists in CODE (although I still should have checked I could deploy the branch before merging to master!)

PR #95, opened 2018-05-22
Return more information on a user's relationship status

jacobwinch:
This PR adds more specific responses for the client (i.e. the identity team's account-deletion Step Function).

There are a number of scenarios to deal with here:

1. The identity id is not linked to a Zuora account (or it's some kind of anomaly i.e. it's linked to a Zuora account but has no subscriptions attached). In this case we return a 404, effectively confirming that the identity account can be deleted.
2. The identity id is linked to a Zuora account, but it has been inactive for longer than our retention period.  In this case we return a 404, effectively confirming that the identity account can be deleted. Note: this is not yet implemented, as I'm unsure how these accounts will be identified on our side.
3. The identity id is linked to a Zuora account and there is at least one active sub. We return a 200 and confirm that there is an ongoing relationship, effectively asking that the identity account is retained. The client updates the 'last activity date' (in identity) with today's date, which means that they won't retry again for x* months.
4. The identity id is linked to a Zuora account, there are only cancelled subs, but the user hasn't yet exceeded our retention period. The client updates the 'last activity' date (in identity) with the date that the service will stop. Assuming there is no other login activity, the account will become a candidate deletion for again x* months from the service end date, and the client will call this API again. 

*defined by the identity team

cc @jfsoul @rahilb


handlers/identity-retention/src/main/scala/com/gu/identityRetention/IdentityRetentionResponses.scala
```diff
@@ -0,0 +1,50 @@
...
+
+  sealed trait IdentityRetentionResponse
+
+  case class SuccessResponse(ongoingRelationship: Boolean, serviceEndDate: Option[LocalDate]) extends IdentityRetentionResponse
+
+  object SuccessResponse {
+    implicit val successResponseWrites = Json.writes[SuccessResponse]
+  }
+
+  case class NotFoundResponse(previousRelationship: Boolean) extends IdentityRetentionResponse
```
jacobwinch: This enables us to differentiate between scenarios 1 and 2 (see PR description), if we ever need to.

handlers/identity-retention/src/main/scala/com/gu/identityRetention/QueryZuoraSubscriptions.scala
```diff
@@ -0,0 +1,38 @@
...
+import play.api.libs.json.Json
+import scalaz.{-\/, \/-}
+
+object QueryZuoraSubscriptions {
+
+  case class SubscriptionsQueryResponse(Id: String, Status: String, Name: String, TermEndDate: LocalDate)
+  implicit val reads = Json.reads[SubscriptionsQueryResponse]
+
+  def buildQuery(accountsToQuery: List[AccountId]): String = {
+    val accountIdClause = "accountId = '" + accountsToQuery.map(_.value).mkString("' or accountId = '") + "'"
```
jacobwinch: This gets the latest version of the sub for all accounts linked to the identity id.

handlers/identity-retention/src/main/scala/com/gu/identityRetention/QueryZuoraSubscriptions.scala
```diff
@@ -0,0 +1,38 @@
...
+import java.time.LocalDate
+import com.gu.identityRetention.Types.AccountId
+import com.gu.util.apigateway.ApiGatewayResponse
+import com.gu.util.reader.Types.FailableOp
+import com.gu.util.zuora.ZuoraQuery
+import com.gu.util.zuora.ZuoraQuery.ZuoraQuerier
+import play.api.libs.json.Json
+import scalaz.{-\/, \/-}
+
+object QueryZuoraSubscriptions {
```
johnduffell: SubscriptionsForAccount

handlers/identity-retention/src/main/scala/com/gu/identityRetention/QueryZuoraSubscriptions.scala
```diff
@@ -0,0 +1,38 @@
...
+
+  case class SubscriptionsQueryResponse(Id: String, Status: String, Name: String, TermEndDate: LocalDate)
```
johnduffell: 4 fields is a fair number, consider
```scala
case class SubscriptionsQueryResponse(
  Id: String,
  Status: String,
  Name: String,
  TermEndDate: LocalDate
)
```

handlers/identity-retention/src/main/scala/com/gu/identityRetention/QueryZuoraSubscriptions.scala
```diff
@@ -0,0 +1,38 @@
...
+  implicit val reads = Json.reads[SubscriptionsQueryResponse]
+
+  def buildQuery(accountsToQuery: List[AccountId]): String = {
+    val accountIdClause = "accountId = '" + accountsToQuery.map(_.value).mkString("' or accountId = '") + "'"
+    s"select id, name, status, termEndDate from subscription where status != 'Expired' and $accountIdClause"
```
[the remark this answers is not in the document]
johnduffell: we don't need to select the Status as I don't think we use it.  same for the Id I think?  but maybe it returns it regardless

handlers/identity-retention/src/main/scala/com/gu/identityRetention/SubscriptionResults.scala
```diff
@@ -0,0 +1,24 @@
+package com.gu.identityRetention
+
+import java.time.LocalDate
+import com.gu.identityRetention.QueryZuoraSubscriptions.SubscriptionsQueryResponse
+import com.gu.util.reader.Types.FailableOp
+import scalaz.{-\/, \/-}
+
+object SubscriptionResults {
```
johnduffell: RelationshipForSubscriptions perhaps?

handlers/identity-retention/src/main/scala/com/gu/identityRetention/SubscriptionResults.scala
```diff
@@ -0,0 +1,24 @@
...
+
+  implicit def ordering: Ordering[LocalDate] = Ordering.by(_.toEpochDay)
+
+  def apply(subscriptions: List[SubscriptionsQueryResponse]): FailableOp[Unit] = {
```
johnduffell: this is some important logic, so perhaps we can simplify the return type to simplify testing/make it clearer. 
given a list of status and end dates, we want to define the retention status. 
So the return type could be an ApiResponse, or even better a sealed trait with 3 case objects/classes (notFoundInZuora, ongoingRelationship and cancelledRelationship).
Then we do need a type wrangling to wrap into a Left of Api response, but at least we can test the logic clearly as that's the interesting part.

handlers/identity-retention/src/main/scala/com/gu/identityRetention/SubscriptionResults.scala
```diff
@@ -0,0 +1,24 @@
...
+    subscriptions match {
+      case subs if (subs.size == 0) =>
+        -\/(IdentityRetentionApiResponses.notFoundInZuora)
+      case subs if (subs.map(_.Status).contains("Active")) =>
```
johnduffell: `subs.exists(_.Status == "Active")` might be slightly simpler

handlers/identity-retention/src/main/scala/com/gu/identityRetention/SubscriptionResults.scala
```diff
@@ -0,0 +1,24 @@
...
+import com.gu.util.reader.Types.FailableOp
+import scalaz.{-\/, \/-}
+
+object SubscriptionResults {
+
+  implicit def ordering: Ordering[LocalDate] = Ordering.by(_.toEpochDay)
+
+  def apply(subscriptions: List[SubscriptionsQueryResponse]): FailableOp[Unit] = {
+    subscriptions match {
+      case subs if (subs.size == 0) =>
```
johnduffell: if you think it's more readable, we could do `case Nil =>`

handlers/identity-retention/src/main/scala/com/gu/identityRetention/SubscriptionResults.scala
```diff
@@ -0,0 +1,24 @@
...
+        -\/(IdentityRetentionApiResponses.notFoundInZuora)
+      case subs if (subs.map(_.Status).contains("Active")) =>
+        -\/(IdentityRetentionApiResponses.ongoingRelationship)
+      case _ =>
```
johnduffell: might be worth a comment saying "has subs but they are all cancelled"?

johnduffell:
great stuff, a few comments mostly about naming for your discretion.  I would like to see the function return types for the critical logic as simple as possible before shipping, but other than that it's good to go from my point of view 👍

handlers/identity-retention/src/main/scala/com/gu/identityRetention/IdentityRetentionResponses.scala
jacobwinch: Fair question, any thoughts on this @jfsoul @paulbrown1982?

handlers/identity-retention/src/main/scala/com/gu/identityRetention/QueryZuoraSubscriptions.scala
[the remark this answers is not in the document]
jacobwinch: Good point re 'Name' - the lower casing works fine

handlers/identity-retention/src/test/scala/com/gu/identityRetention/RelationshipForSubscriptionsTest.scala
```diff
@@ -1,45 +1,45 @@
...
+class RelationshipForSubscriptionsTest extends FlatSpec with Matchers {
 
-  val activeSubResponse = SubscriptionsQueryResponse("id123", "Active", "A-S123", LocalDate.now().plusMonths(4))
-  def cancelledSubResponse(date: LocalDate, subName: String) = SubscriptionsQueryResponse("id123", "Cancelled", subName, date)
+  val activeSubResponse = SubscriptionsQueryResponse("id123", "Active", LocalDate.now().plusMonths(4))
+  def cancelledSubResponse(date: LocalDate) = SubscriptionsQueryResponse("id123", "Cancelled", date)
 
   it should "return a 404 if no subscriptions are found" in {
-    val subResults = SubscriptionResults.apply(List())
+    val subResults = RelationshipForSubscriptions.apply(List())
```
johnduffell: could do away with all the .apply in here

handlers/identity-retention/src/test/scala/com/gu/identityRetention/SubscriptionsForAccountsTest.scala
```diff
@@ -0,0 +1,20 @@
...
+
+  it should "build a valid query for a single account" in {
+    val query = SubscriptionsForAccounts.buildQuery(List(AccountId("acc123")))
+    val expectedQuery = s"select id, name, status, termEndDate from subscription where status != 'Expired' and accountId = 'acc123'"
+    query should be(expectedQuery)
+  }
+
+  it should "build a valid query for multiple accounts" in {
+    val query = SubscriptionsForAccounts.buildQuery(List(AccountId("acc123"), AccountId("acc321"), AccountId("acc456")))
+    val expectedQuery = s"select id, name, status, termEndDate from subscription where status != 'Expired' and accountId = 'acc123' or accountId = 'acc321' or accountId = 'acc456'"
```
johnduffell: codacy is also picking this up as a long line - which is a fair point, a bit of stripmargin or similar would probably make it easier to understand.
alternatively we could just test the where clause generation rather than the whole query

handlers/identity-retention/src/test/scala/com/gu/identityRetention/IdentityRetentionHandlerTest.scala
```diff
@@ -20,20 +18,33 @@ class IdentityRetentionHandlerTest extends FlatSpec with Matchers {
 
     val actualResponse = new String(os.toByteArray, "UTF-8")
 
-    actualResponse jsonMatches (notFoundResponse)
+    actualResponse jsonMatches (noPreviousRelationship)
```
johnduffell: codacy is picking up a lot of duplciation in this test, perhaps we can have a getActual function that takes a dummyRequest(...) and handles all the wiring, then it's clearer what's actually different in the tests

handlers/identity-retention/src/main/scala/com/gu/identityRetention/SubscriptionsForAccounts.scala
```diff
@@ -0,0 +1,43 @@
...
+    Id: String,
+    Status: String,
+    TermEndDate: LocalDate
+  )
+
+  implicit val reads = Json.reads[SubscriptionsQueryResponse]
+
+  def buildQuery(accountsToQuery: List[AccountId]): String = {
+    val accountIdClause = "accountId = '" + accountsToQuery.map(_.value).mkString("' or accountId = '") + "'"
+    s"select id, name, status, termEndDate from subscription where status != 'Expired' and $accountIdClause"
```
johnduffell: @jacobwinch fyi I was adding an effects test for this and I noticed a problem with the query
https://knowledgecenter.zuora.com/DC_Developers/K_Zuora_Object_Query_Language
"The and operator takes priority in the order of operation over the or operator."
I found this query actually returns expired subscriptions belonging to the second and subsequent account passed in.
I am fixing it as part of stopping us building queries directly in the code.

PR #96, opened 2018-05-23
Add custom domain name for API Gateway

jacobwinch:
I need to add the Route 53 record in a separate PR, as the target DNS generated by this change cannot be accessed in CFN yet.

### DIFF
--- a/handlers/identity-retention/cfn.yaml
+++ b/handlers/identity-retention/cfn.yaml
@@ -9,10 +9,6 @@ Parameters:
             - PROD
             - CODE
         Default: CODE
-    DomainName:
-        Type: String
-    CertificateArn:
-        Type: String
 
 Conditions:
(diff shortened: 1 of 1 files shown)


handlers/identity-retention/cfn.yaml
```diff
@@ -9,6 +9,10 @@ Parameters:
             - PROD
             - CODE
         Default: CODE
+    DomainName:
```
johnduffell: can we put this in the stage mappings, or use !Sub, so we don't have to pass it in?

johnduffell: talked to @pvighi and it's just in case we want to keep it secret because it's the internal endpoint of a service that goes via fastly, so we could put in the mappings in this case

johnduffell:
see above, then 👍 as discussed we should roughly write down the steps needed to get a new endpoint going to see where it is hard

PR #98, opened 2018-05-24
Exclude accounts which are outside of the retention period

jacobwinch:
... so that we can allow the associated identity accounts to be deleted.

Note that this PR also removes a distinction between users who don't appear in Zuora at all, and those who are in Zuora (but are outside of the relevant retention period), because nobody can think of a use case for this extra detail in the response.

### DIFF
--- a/handlers/identity-retention/src/main/scala/com/gu/identityRetention/HasActiveZuoraAccounts.scala
+++ b/handlers/identity-retention/src/main/scala/com/gu/identityRetention/HasActiveZuoraAccounts.scala
@@ -11,13 +11,25 @@ import scalaz.{-\/, \/-}
 
 object HasActiveZuoraAccounts {
 
-  case class IdentityQueryResponse(Id: String, Status: String)
+  case class IdentityQueryResponse(Id: String)
   implicit val reads = Json.reads[IdentityQueryResponse]
 
   def apply(identityId: IdentityId, zuoraQuerier: ZuoraQuerier): FailableOp[List[AccountId]] = {
 
     def searchForAccounts = {
-      val identityQuery = ZuoraQuery.Query(s"select id, status from account where IdentityId__c = '${identityId.value}'")
+
+      val commonConditions = s"IdentityId__c = '${identityId.value}' and status != 'Canceled'"
+
+      /*
+      Todo simplify this once we can rely on Account.Status
+      Unfortunately Zuora do not support parentheses, and != doesn't pick up nulls
+      https://knowledgecenter.zuora.com/DC_Developers/K_Zuora_Object_Query_Language#Syntax
+      */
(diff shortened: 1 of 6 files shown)


handlers/identity-retention/src/main/scala/com/gu/identityRetention/IdentityRetentionResponses.scala
```diff
@@ -14,7 +14,7 @@ object IdentityRetentionResponseModels {
     implicit val successResponseWrites = Json.writes[SuccessResponse]
   }
 
-  case class NotFoundResponse(previousRelationship: Boolean) extends IdentityRetentionResponse
+  case class NotFoundResponse(message: String = "identity account can be deleted") extends IdentityRetentionResponse
```
paulbrown1982: I don't agree with this message text - it's not up to us to advise Identity to delete, but to advise that we have no active relationship for the customer it sent to us.

handlers/identity-retention/src/main/scala/com/gu/identityRetention/HasActiveZuoraAccounts.scala
```diff
@@ -11,13 +11,25 @@ import scalaz.{-\/, \/-}
...
-  case class IdentityQueryResponse(Id: String, Status: String)
+  case class IdentityQueryResponse(Id: String)
   implicit val reads = Json.reads[IdentityQueryResponse]
 
   def apply(identityId: IdentityId, zuoraQuerier: ZuoraQuerier): FailableOp[List[AccountId]] = {
 
     def searchForAccounts = {
-      val identityQuery = ZuoraQuery.Query(s"select id, status from account where IdentityId__c = '${identityId.value}'")
+
+      val commonConditions = s"IdentityId__c = '${identityId.value}' and status != 'Canceled'"
```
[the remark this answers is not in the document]
johnduffell: how about `nonCancelledForIdentityId` ?

handlers/identity-retention/src/main/scala/com/gu/identityRetention/HasActiveZuoraAccounts.scala
```diff
@@ -30,7 +42,7 @@ object HasActiveZuoraAccounts {
       case \/-(result) if result.size > 0 =>
         \/-(result.records.map(account => AccountId(account.Id)))
       case \/-(result) if result.size == 0 =>
-        -\/(IdentityRetentionApiResponses.notFoundInZuora)
+        -\/(IdentityRetentionApiResponses.canBeDeleted)
```
johnduffell: shoudl have changed `canBeDeleted` to be `noActiveRelationships` to match

johnduffell:
yep looks good, just a couple of naming things 👍

PR #99, opened 2018-05-29
Zuora reports initial PR

pvighi:
This is the initial PR for a generic zuora reports project.
This will eventually contain the lambdas to run Zuora Aqua queries and store the results in S3 once they are done.

this PR only has the initial project and a fully functioning querier lambda.


lib/zuora/src/main/scala/com/gu/util/zuora/ZuoraDeps.scala
```diff
@@ -1,7 +1,6 @@
 package com.gu.util.zuora
```
pvighi: Initially I though I needed to change the zuora configuration but It turns out I don't. 
I reverted my changes but I left this refactor in if anyone has any reason why the old way was better I could just leave it as it

handlers/zuora-reports/cfn.yaml
```diff
@@ -0,0 +1,63 @@
+AWSTemplateFormatVersion: "2010-09-09"
+Description: Zuora reporter
```
johnduffell: might be worth a bit more description?

lib/zuora/src/main/scala/com/gu/util/zuora/ZuoraDeps.scala
johnduffell: definitely better, no need to define our own reads for that 👍

handlers/zuora-reports/src/test/scala/com/gu/zuora/reports/QuerierEffectsTest.scala
```diff
@@ -0,0 +1,30 @@
...
+
+import com.gu.effects.{RawEffects, S3ConfigLoad}
+import com.gu.test.EffectsTest
+import com.gu.util.config.{LoadConfig, Stage}
+import com.gu.zuora.reports.AquaLambda.StepsConfig
+import org.scalatest.{FlatSpec, Matchers}
+import scalaz.{-\/, \/}
+import scalaz.syntax.std.either._
+
+//TODO SEE HOW TO DO EFFECT TESTS FOR REPORT QUERIES SINCE THEY ALL DEPEND ON THIS ONE
```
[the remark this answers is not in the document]
johnduffell: as discussed the EndToEndHandlerTest and GetByEmailTest with the fake TestingRawEffects might be a useful template for tests with the responses copy and pasted from Post man.

This gives confidence that we can correctly deal with the existing responses so is probably worthwhile.

Of course the effects tests can provide an extra level of confidence although at significantly greater effort of maintenance and execution.  So it would be down to a judgment of cost to benefit for each layer of testing.

handlers/zuora-reports/src/main/scala/com/gu/zuora/reports/Querier.scala
```diff
@@ -0,0 +1,67 @@
...
+  implicit val writes = Json.writes[QuerierResponse]
+}
+
+case class Query(name: String, query: String)
+
+case class AquaQuery(name: String, query: String, `type`: String = "zoqlexport")
+
+trait ZuoraAquaRequest
+
+case class QuerierRequest(name: String, queries: Seq[Query]) extends ZuoraAquaRequest
```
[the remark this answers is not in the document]
johnduffell: yes, I used the convention of "wire" for the ones that go over the wire to zuora, and a more normal sounding name for the ones we use internally in the code.
However I am happy with any sensible convention and we can rename the existing ones I called "wire" to make them consistent.

handlers/zuora-reports/src/main/scala/com/gu/zuora/reports/ReportsLambda.scala
```diff
@@ -0,0 +1,75 @@
...
+object ReportsLambda extends Logging {
+
+  case class StepsConfig(zuoraRestConfig: ZuoraRestConfig)
+
+  implicit val stepsConfigReads: Reads[StepsConfig] = Json.reads[StepsConfig]
+
+  trait LambdaException extends Exception
+
+  object LambdaException {
+    def apply(message: String, root: Throwable) = new Exception(message, root) with LambdaException
```
johnduffell: maybe don't need this version any more

handlers/zuora-reports/src/main/scala/com/gu/zuora/reports/aqua/ZuoraAquaRequestMaker.scala
```diff
@@ -0,0 +1,46 @@
...
+  def zuoraIsSuccessful(bodyAsJson: JsValue): ClientFailableOp[Unit] = {
+
+    bodyAsJson.validate[ZuoraAquaResponse] match {
+      case JsSuccess(ZuoraAquaResponse("error", name, errorCode, message, _, _), _) => {
+        logger.error(s"Zuora Aqua Api rejected our call $bodyAsJson")
+        val codePart = errorCode.map(c => s"error code $c:")
+        val messagePart = message.getOrElse("No error message")
+        GenericError(s"$codePart $messagePart").left
+      }
+      case JsSuccess(ZuoraAquaResponse(_, _, _, _, _, _), _) => ().right
```
johnduffell: `case JsSuccess(ZuoraAquaResponse(_, _, _, _, _, _), _) => ().right`
might work in a shorter way as:
`case JsSuccess(_ :ZuoraAquaResponse, _) => ().right`

handlers/zuora-reports/src/test/scala/com/gu/zuora/reports/ReportsLambdaEndToEndTest.scala
```diff
@@ -0,0 +1,152 @@
...
+ |"clientSecret": "someSecret"
+ |},
+ |"stripe": {
+ |"customerSourceUpdatedWebhook": {
+ |"api.key.secret": "secret",
+ |"au-membership.key.secret": "anotherSecret"
+ |}
+ |}
+ |}
+  """.stripMargin
```
johnduffell: the indentation went a bit weird here

handlers/zuora-reports/src/test/scala/com/gu/zuora/reports/ReportsLambdaEndToEndTest.scala
```diff
@@ -0,0 +1,152 @@
...
+      |    ],
+      |    "status": "submitted",
+      |    "name": "verificationQuery3",
+      |    "id": "aquaJobId",
+      |    "version": "1.0",
+      |    "format": "CSV"
+      |}
+    """.stripMargin
+
+  val aquaQueryRequest = """{"format":"csv","version":"1.0","name":"TestQuery","encrypted":"none","useQueryLabels":"true","dateTimeUtc":"true","queries":[{"name":"query1","query":"select something from somethingElse","type":"zoqlexport"}]}"""
```
johnduffell: could pretty print this one to match the others as well if you're in htere

johnduffell:
great work now the coverage is great and the test will be useful to keep things working in future👍 🛥

handlers/zuora-reports/src/test/scala/com/gu/zuora/reports/ReportsLambdaEndToEndTest.scala
pvighi: I think I if I pretty print it the test would fail since this would not match anymore:
https://github.com/guardian/zuora-auto-cancel/blob/master/lib/effects/src/test/scala/com/gu/effects/TestingRawEffects.scala#L47

PR #100, opened 2018-05-30
remove extra logging from catalog service

pvighi:
Tested in CODE, config file is not logged and catalog file is updated in S3.
@jacobwinch anything else we need to check ?


johnduffell:
👍 thanks!

pvighi:
💯

PR #101, opened 2018-06-01
Add alarms for identity-retention API

jacobwinch:
This PR adds basic monitoring for 4XX and 5XX response codes.


johnduffell:
👍 great

PR #102, opened 2018-06-01
Zuora reports 2 : added query results lambda

pvighi:
Checks whether a Zuora AQuA job is done and if it is returns the file ids with the results

### DIFF
--- a/handlers/zuora-reports/src/main/scala/com/gu/zuora/reports/GetJobResult.scala
+++ b/handlers/zuora-reports/src/main/scala/com/gu/zuora/reports/GetJobResult.scala
@@ -9,14 +9,14 @@ import scalaz.{-\/, \/-}
 object GetJobResult {
   def apply(zuoraRequester: Requests, JobResultRequest: JobResultRequest): ClientFailableOp[JobResultResponse] = {
     val zuoraAquaResponse = zuoraRequester.get[ZuoraAquaResponse](s"batch-query/jobs/${JobResultRequest.jobId}")
-    toCheckerResponse(zuoraAquaResponse)
+    toJobResultResponse(zuoraAquaResponse)
   }
 
   def toBatch(aquaBatch: aqua.Batch): Option[Batch] = aquaBatch.fileId.map {
     fileId => Batch(name = aquaBatch.name, fileId = fileId)
   }
 
-  def toCheckerResponse(aquaResponse: ClientFailableOp[ZuoraAquaResponse]): ClientFailableOp[JobResultResponse] = {
+  def toJobResultResponse(aquaResponse: ClientFailableOp[ZuoraAquaResponse]): ClientFailableOp[JobResultResponse] = {
     aquaResponse match {
       case \/-(ZuoraAquaResponse(status, name, _, _, aquaBatches, _)) if (status.isCompleted) =>
         val batches = aquaBatches.map(toBatch)
--- a/handlers/zuora-reports/src/main/scala/com/gu/zuora/reports/GetJobResult.scala
+++ b/handlers/zuora-reports/src/main/scala/com/gu/zuora/reports/GetJobResult.scala
@@ -7,8 +7,8 @@ import play.api.libs.json._
 import scalaz.{-\/, \/-}
 
 object GetJobResult {
-  def apply(zuoraRequester: Requests, JobResultRequest: JobResultRequest): ClientFailableOp[JobResult] = {
-    val zuoraAquaResponse = zuoraRequester.get[AquaJobResponse](s"batch-query/jobs/${JobResultRequest.jobId}")
+  def apply(zuoraRequester: Requests, jobResultRequest: JobResultRequest): ClientFailableOp[JobResult] = {
+    val zuoraAquaResponse = zuoraRequester.get[AquaJobResponse](s"batch-query/jobs/${jobResultRequest.jobId}")
     toJobResultResponse(zuoraAquaResponse)
   }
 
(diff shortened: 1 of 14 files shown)


handlers/zuora-reports/src/main/scala/com/gu/zuora/reports/GetJobResult.scala
```diff
@@ -0,0 +1,74 @@
+package com.gu.zuora.reports
+
+import com.gu.util.zuora.RestRequestMaker
+import com.gu.util.zuora.RestRequestMaker.{ClientFailableOp, GenericError, Requests}
+import com.gu.zuora.reports.aqua.ZuoraAquaResponse
+import play.api.libs.json._
+import scalaz.{-\/, \/-}
+
+object GetJobResult {
+  def apply(zuoraRequester: Requests, JobResultRequest: JobResultRequest): ClientFailableOp[JobResultResponse] = {
```
johnduffell: GetJobResult, thinking about it on its own I would expect to take a "JobID" rather than a JobResultRequest as that just sounds very vague.  If we can, we should be as specific as possible (we can always make a positive decision to rename later).
It looks like we want it to return a JobStatus rather than just a generic JobResultResponse.
At the moment the idea of these rest functions is they are really simple/specific, self contained and close to what you would do in postman, so we needn't try to make them too generic. 
We could even rename the GetJobResult to be GetJobStatus for added specificity.

handlers/zuora-reports/src/main/scala/com/gu/zuora/reports/GetJobResult.scala
```diff
@@ -0,0 +1,74 @@
...
+        if (batches.contains(None)) {
+          -\/(RestRequestMaker.GenericError(s"file Id missing from response : $aquaResponse"))
+        } else {
+          \/-(Completed(name, batches.flatten))
+        }
+
+      case \/-(ZuoraAquaResponse(status, _, _, _, _, _)) if (status.isPending) => \/-(Pending)
+      case \/-(zuoraResponse) => -\/(GenericError(s"unexpected status in zuora response: $zuoraResponse"))
+      case -\/(error) => -\/(error)
+    }
```
johnduffell: we never look at the errorCode and message field in this response so we can remove them from the class to simplify this code.
I noticed it is shared with the ZuoraAquaRequestMaker, which doesn't read the name, batches and id.  Since only the status is shared, it's probably not worth trying to share the case class between the two?

handlers/zuora-reports/src/main/scala/com/gu/zuora/reports/GetJobResult.scala
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  implicit class StringStatusOps(s: String) {
+    def isCompleted = s.equalsIgnoreCase("completed")
+
+    def isPending = List("pending", "executing").exists(_.equalsIgnoreCase(s))
+
+    def isFailure = !isCompleted && !isPending
```
johnduffell: I don't think isFailure is used

handlers/zuora-reports/src/main/scala/com/gu/zuora/reports/GetJobResult.scala
```diff
@@ -0,0 +1,74 @@
...
+        }
+
+      case \/-(ZuoraAquaResponse(status, _, _, _, _, _)) if (status.isPending) => \/-(Pending)
+      case \/-(zuoraResponse) => -\/(GenericError(s"unexpected status in zuora response: $zuoraResponse"))
+      case -\/(error) => -\/(error)
+    }
+  }
+
+  implicit class StringStatusOps(s: String) {
+    def isCompleted = s.equalsIgnoreCase("completed")
```
johnduffell: does zuora sometimes return wrong case or is this just general zuora belt and braces?
I know they vary success/Success between the "rest" and "object" apis

handlers/zuora-reports/src/main/scala/com/gu/zuora/reports/GetJobResult.scala
```diff
@@ -0,0 +1,74 @@
...
+
+object Batch {
+  implicit val writes = Json.writes[Batch]
+}
+
+case class Completed(name: String, batches: Seq[Batch]) extends JobResultResponse
+
+object Completed {
+  val writes = Json.writes[Completed]
+}
```
johnduffell: Althogh overall it's easy to understand the code when it's together, I feel like it's a bit blurry/mixed what is the API to zuora, and what is communicating outwards via the lambda.  I would prefer to see this broken down as
1. a request and response object for the API gateway with the default implicit reads and writes
1. one scala object with its own minimal types for doing the zuora call,
1. one for processing the response to a nicer type to return to the client i.e. the toJobResultResponse stuff
Then a top level method can connect the parts together with a for comprehension or just calling them consecutively, but everything is clearly separate for testing and minimal dependencies on other code.

johnduffell:
looks good, code is all great, I just have a little nervousness about the code being a little bit more combined than it should.  If we can keep each piece small and dedicated, and group into files and objects (and packages even) of code that is cohesive (in terms of connections, rather than their type) then it will be easier to separate in future.  The more that things end up mixed (on any level) the less likely things would be to be refactoable in future.
I have left some comments about that topic, have a think about that in general, but don't let that block things too much 👍

handlers/zuora-reports/src/main/scala/com/gu/zuora/reports/GetJobResult.scala
pvighi: this was getJobStatus originally I changed it because It also gets the results (file ids) if the status is completed but I guess I'm not bothered calling it either way.

I like the idea of not calling it JobResultRequest but the problem with calling it job Id is that it still needs a jobId field inside so the json request is deserialised properly. So we would end up doing jobId.jobId

handlers/zuora-reports/src/main/scala/com/gu/zuora/reports/GetJobResult.scala
johnduffell: patricio explained the code should be thought of more like a proxy server with some liight filtering/error handling rather than a lambda that takes some input, decides to call some zuora stuff, and then decides what result to return. Since we should keep the code as simple as meets the concepts, it makes sense to share data structured between the zuora and lambda facing stuff.

So this comment is not really relevant in that case! 💣

johnduffell:
👍

PR #103, opened 2018-06-01
Add simple monitoring for catalog service

jacobwinch:
This PR means that we will be alerted if we fail to update the Zuora catalog(s) in S3 for a sustained period of time:

PROD - an hour and a half
UAT - 3 hours
DEV - 6 hours

Based on https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/AlarmThatSendsEmail.html#alarm-evaluation

There's no point having super-sensitive alerts because they will just spam us (even more) if the Zuora API is down; these alerts are just intended to catch obvious production bugs/misconfigurations with this handler.

### DIFF
--- a/handlers/catalog-service/cfn.yaml
+++ b/handlers/catalog-service/cfn.yaml
@@ -207,11 +207,11 @@ Resources:
 
     InvocationErrorsAlarm:
       Type: AWS::CloudWatch::Alarm
-      # Condition: CreateProdOnlyResources
+      Condition: CreateProdOnlyResources
       Properties:
         AlarmActions:
         - !Sub arn:aws:sns:${AWS::Region}:${AWS::AccountId}:subscriptions_dev
-        AlarmName: High error rate for PROD Zuora Catalog Service
+        AlarmName: High error rate for Zuora PROD Catalog Service
         ComparisonOperator: GreaterThanThreshold
         Dimensions:
           - Name: FunctionName
@@ -219,7 +219,7 @@ Resources:
         EvaluationPeriods: 9
         MetricName: Errors
         Namespace: AWS/Lambda
-        Period: 3600
+        Period: 600
         Statistic: Sum
         Threshold: 1
         TreatMissingData: notBreaching
--- a/handlers/catalog-service/cfn.yaml
+++ b/handlers/catalog-service/cfn.yaml
@@ -205,18 +205,56 @@ Resources:
         DependsOn:
         - PRODCatalogServiceLambda
 
-    InvocationErrorsAlarm:
+    StaleProdCatalogAlarm:
       Type: AWS::CloudWatch::Alarm
       Condition: CreateProdOnlyResources
       Properties:
         AlarmActions:
         - !Sub arn:aws:sns:${AWS::Region}:${AWS::AccountId}:subscriptions_dev
-        AlarmName: High error rate for Zuora PROD Catalog Service
+        AlarmName: High error rate when fetching PROD Zuora Catalog
         ComparisonOperator: GreaterThanThreshold
         Dimensions:
           - Name: FunctionName
             Value: !Ref PRODCatalogServiceLambda
         EvaluationPeriods: 9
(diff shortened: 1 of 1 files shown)


jacobwinch:
The downsides of this approach (as opposed to monitoring files in the bucket for staleness) are that we won't notice if the lambdas don't run at all (i.e. if we break/delete the scheduling) or if the lambdas are failing silently. 

However this approach is quicker, simpler (I think the other approach will require another lambda to do the monitoring) and should catch the majority of mistakes.

handlers/catalog-service/cfn.yaml
```diff
@@ -203,4 +203,61 @@ Resources:
...
+        Dimensions:
+          - Name: FunctionName
+            Value: !Ref DEVCatalogServiceLambda
+        EvaluationPeriods: 36
+        MetricName: Errors
+        Namespace: AWS/Lambda
+        Period: 600
+        Statistic: Sum
+        Threshold: 1
+        TreatMissingData: notBreaching
```
pvighi: I assumed we only wanted an alarm when the prod catalog failed. If we are getting an alarm even if the dev one fails to update why did we bother having multiple lambdas ?

jacobwinch: I originally just set up a PROD alarm, but then I thought we should probably still know if the DEV or UAT versions are fundamentally broken. Otherwise someone is going to waste a lot of time one day debugging it...

jacobwinch: Open to other suggestions though (e.g. if you don't think we ever care about alerts for the DEV catalog)

jacobwinch:
@pvighi I would quite like a general service for monitoring things in S3. Being able to easily send an alert if file y is more than x hours old would be useful for cases like this, and also for ensuring that retention policies are being applied correctly in buckets which contain PII.

johnduffell:
that's a great idea about monitoring things in S3.  I know on dotcom there is a thing to email when things *are* changed (switches file) but I'm not sure if we could alert by treating missing data as breaching, and then saying if it's breaching 120 times in the last 120 minutes then alarm.

handlers/catalog-service/cfn.yaml
```diff
@@ -203,4 +203,61 @@ Resources:
...
+        - !Sub arn:aws:sns:${AWS::Region}:${AWS::AccountId}:subscriptions_dev
+        AlarmName: High error rate when fetching DEV Zuora Catalog
+        ComparisonOperator: GreaterThanThreshold
+        Dimensions:
+          - Name: FunctionName
+            Value: !Ref DEVCatalogServiceLambda
+        EvaluationPeriods: 36
+        MetricName: Errors
+        Namespace: AWS/Lambda
+        Period: 600
```
johnduffell: why a 10 min period and 36 periods rather than 1 minute and 360 periods?  Not saying it should be changed, I just wonder if it breaks once will we get an email up to 10 minutes later, regardless of if zuora is down? or will it make sure it's failed at least once in every 10 minutes period for a full 360 mins before it emails?

johnduffell:
thanks! 👍

handlers/catalog-service/cfn.yaml
jacobwinch: It alerts if it's failed at least once in every period for the number of evaluation periods specified. So in this case it has to fail consistently for 6 hours before we get the alert.

I chose 10 minutes as the period because that's currently the [frequency](https://github.com/guardian/zuora-auto-cancel/blob/master/handlers/catalog-service/cfn.yaml#L201) for the schedule.

jacobwinch: Interestingly, when I tested this I noticed that there are automatic retries when this function fails (as it's an async invocation): 

https://docs.aws.amazon.com/lambda/latest/dg/invoking-lambda-function.html#supported-event-source-cloudwatch-events
https://docs.aws.amazon.com/lambda/latest/dg/retries-on-errors.html

PR #104, opened 2018-06-01
Improve catalog-service readme

jacobwinch:
The multi-environment nature of this project is not particularly intuitive, so I've documented the reasoning behind it here.


johnduffell:
thanks that's very helpful 👍

PR #106, opened 2018-06-04
added results fetcher lambda to riff-raff.yaml

pvighi:
I forgot to add it in my last PR


johnduffell:
so many things to remember!! 😢 👍

PR #107, opened 2018-06-05
Zuora reports fetch files and some extra stuff

pvighi:
* added file fetcher lambda : downloads query results from Zuora and stores them in S3
* improved getJobStatus response to be easier to parse: it now contains a status that could be pending or completed
* added manual testing to run the lambda code in dev without having to deploy it
* added some documentation
* added export statements to cloudformation template

### DIFF
--- a/handlers/zuora-reports/README.md
+++ b/handlers/zuora-reports/README.md
@@ -69,11 +69,14 @@ Here we get the results of the job submitted in the previous example and get a c
 }
 ```
 ## zuora-reports-fileFetcher
-This lambda is used to fetch the results of a query using the file ids retrieved from the response of jobResul.
+This lambda is used to fetch the results of a query using the file ids retrieved from the response of JobResult.
 
 ### sample usage
-The input contains the the file id and the name we want the file to have when it saved to s3 ( .csv will be appended to the end of it).
-The output will contain the fileId and the path in s3 where the file was saved.
+The input contains the the file id and the location we want to full within the zuora-reports-[STAGE] bucket.
+
+The output will contain the uri of the file with the results of the query.
+This uri will remain valid for up to a day to ensure complicance with GDPR.
+ 
  #### input
 ```
 {
--- a/handlers/zuora-reports/README.md
+++ b/handlers/zuora-reports/README.md
@@ -38,7 +38,7 @@ This Lambda is used to check the status of an Aqua job (normally submitted using
 The response will include whether the job is completed or not and the file id that can be used to fetch the results.
 ### sample usage
 Here we get the results of the job submitted in the previous example and get a completed response with the file ids we will need to download the query results: 
-####input
+#### input
 ```
 {
   "jobId": "23049832094823098403284039284"
@@ -68,10 +68,10 @@ Here we get the results of the job submitted in the previous example and get a c
   "status" : "pending"
 }
 ```
-##zuora-reports-fileFetcher
+## zuora-reports-fileFetcher
 This lambda is used to fetch the results of a query using the file ids retrieved from the response of jobResul.
 
-###sample usage
+### sample usage
 The input contains the the file id and the name we want the file to have when it saved to s3 ( .csv will be appended to the end of it).
 The output will contain the fileId and the path in s3 where the file was saved.
  #### input
--- a/handlers/zuora-reports/README.md
+++ b/handlers/zuora-reports/README.md
@@ -1,1 +1,90 @@
 # zuora-reports
+This project consists of 4 lambdas used for zuora report generation:
+* zuora-reports-querier: Submits Zuora AQuA query jobs
+* zuora-reports-jobResult: Checks the status of submitted Zuora AQuA query jobs 
+* zuora-reports-fileFetcher: Fetches csv result files from completed jobs and saves them in S3
+
+These lambdas are exported to be used as part of step functions that require zuora reports.
+## zuora-reports-querier
+Used to submit query jobs to Zuora AQuA. The lambda takes an array of AQuA queries and returns a jobId that can later be used to get the results.
+### sample usage
+Here is an example of a successful request which returns the job id that can be passed to the jobResult lambda and check the status of the job.
+#### Input
+```
+{
+  "name": "someQueryJob",
+  "queries": [
+    {
+      "name": "subs",
+      "query": "SELECT Id FROM Subscription WHERE  Name='1234'"
+    },
+    {
+      "name": "accounts",
+      "query": "SELECT Id FROM Account WHERE  Name='4321'"
+    }    
+  ]
+}
+```
+#### Ouput
+```
+{
+  "name": "someQueryJob",
+  "jobId": "23049832094823098403284039284"
+}
+```
+
+##zuora-reports-jobResult
+This Lambda is used to check the status of an Aqua job (normally submitted using the querier lambda).
+The response will include whether the job is completed or not and the file id that can be used to fetch the results.
+### sample usage
+Here we get the results of the job submitted in the previous example and get a completed response with the file ids we will need to download the query results: 
+####input
+```
+{
+  "jobId": "23049832094823098403284039284"
+}
+```
+#### completed job output
+```
+{
+  "name": "someQueryJob",
+  "status": "completed",
+  "batches": [
+    {
+      "fileId": "10293821098320983091283012",
+      "name": "subs"
+    },
+    {
+      "fileId": "20341983092183092180392810",
+      "name": "accounts"
+    }
+  ]
+}
+```
+#### pending job output
+```
+{
+  "name": "someQueryJob",
+  "status" : "pending"
+}
+```
+##zuora-reports-fileFetcher
+This lambda is used to fetch the results of a query using the file ids retrieved from the response of jobResul.
+
+###sample usage
+The input contains the the file id and the name we want the file to have when it saved to s3 ( .csv will be appended to the end of it).
+The output will contain the fileId and the path in s3 where the file was saved.
+ #### input
+```
+{
+  "fileId": "20341983092183092180392810",
+  "name": "accounts"
+}
+```
+#### output
+```
+{
+  "fileId": "20341983092183092180392810",
+  "S3Path": "zuora-reports-code/accounts.csv"
+}
+```
--- a/handlers/zuora-reports/README.md
+++ b/handlers/zuora-reports/README.md
@@ -78,13 +78,13 @@ The output will contain the fileId and the path in s3 where the file was saved.
 ```
 {
   "fileId": "20341983092183092180392810",
-  "name": "accounts"
+  "saveLocation": "myReports/accounts.csv"
 }
 ```
 #### output
 ```
 {
   "fileId": "20341983092183092180392810",
-  "S3Path": "zuora-reports-code/accounts.csv"
+  "uri": "s3://zuora-reports-code/myReports/accounts.csv"
 }
 ```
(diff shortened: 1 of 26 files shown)


handlers/zuora-reports/src/main/scala/com/gu/zuora/reports/S3ReportUploader.scala
```diff
@@ -0,0 +1,34 @@
...
+    Stage("DEV") -> "zuora-reports-dev"
+  )
+
+  def apply(stage: Stage, s3Write: PutObjectRequest => Try[PutObjectResult])(downloadStream: DownloadStream, queryName: String): ClientFailableOp[String] = {
+
+    val metadata = new ObjectMetadata()
+    metadata.setContentLength(downloadStream.lengthBytes)
+
+    val destBucket = buckets(stage)
+    val destKey = s"${queryName}.csv" //todo do we want any type of directory structure to save the files ?
```
pvighi: Should we have some sort of folder structure in the results bucket or should everything be thrown together in the same place. I'm now thinking that with the code as it is reports with the same name will overwrite each other so we might need to do something if we want to avoid that

handlers/zuora-reports/src/main/scala/com/gu/zuora/reports/S3ReportUploader.scala
```diff
@@ -0,0 +1,34 @@
...
+    val putObjectRequest = new PutObjectRequest(destBucket, destKey, downloadStream.stream, metadata)
```
pvighi: for now we can use the regular s3.write function passing a stream and the size returned from zuora en the content-length header to avoid loading the whole file in memory. 
We still have a limit of 5 minutes of total execution time to transfer the files so if we need to fetch large files like we do for the data lake we might need to use s3 multipart upload to move the file in multiple lambda executions

handlers/zuora-reports/README.md
```diff
@@ -1 +1,90 @@
...
+{
+  "name": "someQueryJob",
+  "queries": [
+    {
+      "name": "subs",
+      "query": "SELECT Id FROM Subscription WHERE  Name='1234'"
+    },
+    {
+      "name": "accounts",
+      "query": "SELECT Id FROM Account WHERE  Name='4321'"
```
johnduffell: just a random thought, should we in future provide a prepared statement style interface here? so people don't have to build the query as a string ( equals risky)

handlers/zuora-reports/README.md
```diff
@@ -1 +1,90 @@
...
+{
+  "fileId": "20341983092183092180392810",
+  "name": "accounts"
+}
+```
+#### output
+```
+{
+  "fileId": "20341983092183092180392810",
+  "S3Path": "zuora-reports-code/accounts.csv"
```
johnduffell: should be s3Path to match camelcase

handlers/zuora-reports/cfn.yaml
```diff
@@ -9,8 +9,37 @@ Parameters:
...
+          LifecycleConfiguration:
+            Rules:
+              - Id: DeleteAllOldFiles
+                Prefix: ""
+                Status: Enabled
+                ExpirationInDays: 365
+              - Id: DeleteOldFiles
+                Prefix: ""
+                Status: Enabled
+                ExpirationInDays: 14
```
johnduffell: why two rules, one is longer than the other?

handlers/zuora-reports/src/main/scala/com/gu/zuora/reports/FetchFile.scala
```diff
@@ -0,0 +1,28 @@
+package com.gu.zuora.reports
+
+import com.gu.util.zuora.RestRequestMaker.{ClientFailableOp, DownloadStream, GenericError, Requests}
+import play.api.libs.json._
+
+object FetchFile {
+  def apply(
+    uploader: (DownloadStream, String) => ClientFailableOp[String],
```
johnduffell: minor - this is probably "upload" (verb) rather than "uploader" (noun)

handlers/zuora-reports/README.md
pvighi: We should think about how hard to implement and restrictive that could be.
These are input parameters to the lambda, not an api gateway or anything user facing.
This lambda will be called from apps with hard coded queries so I don't think validation is that much of a priority.
The idea is that if you can call the lambda manually you probably have S3 access that allows you to just read the zuora secrets.
If someone somehow manages to get access to running the lambda they would still need access to the encrypted s3 bucket where the results are stored. ( I don't think the [AQuA api](https://knowledgecenter.zuora.com/DC_Developers/T_Aggregate_Query_API/AA_AQuA_API_Introduction) provides support for any state changes, just querying data)

handlers/zuora-reports/cfn.yaml
pvighi: because I'm stupid and copied and pasted the code from fulfilment and forgot to delete the 365 day rule

handlers/zuora-reports/src/main/scala/com/gu/zuora/reports/Handlers.scala
```diff
@@ -5,14 +5,34 @@ import java.io.{InputStream, OutputStream}
 import com.amazonaws.services.lambda.runtime.Context
 import com.gu.effects.RawEffects
 import com.gu.util.apigateway.ApiGatewayHandler.LambdaIO
+import com.gu.util.config.Config
+import com.gu.zuora.reports.ReportsLambda.StepsConfig
+import com.gu.zuora.reports.aqua.ZuoraAquaRequestMaker
 
 object Handlers {
 
+  def requestMaker(config: Config[StepsConfig]) = ZuoraAquaRequestMaker(RawEffects.response, config.stepsConfig.zuoraRestConfig)
```
johnduffell: this function only needs to take a ZuoraRestConfig not the whole Config[...]

handlers/zuora-reports/src/main/scala/com/gu/zuora/reports/S3ReportUploader.scala
johnduffell: good idea, I like your idea before about getting the caller to pass in the full name by convention to include the job id as the first part

lib/effects/src/main/scala/com/gu/effects/AwsS3.scala
```diff
@@ -1,12 +1,16 @@
 package com.gu.effects
 
+import java.io.{BufferedInputStream, InputStream}
+import java.util.Scanner
```
johnduffell: were these import changes just left over from some experimentation?

lib/effects/src/main/scala/com/gu/effects/RawEffects.scala
```diff
@@ -1,5 +1,7 @@
 package com.gu.effects
 
+import java.io.{InputStream, OutputStream}
```
johnduffell: is this import needed?

lib/restHttp/src/main/scala/com/gu/util/zuora/RestRequestMaker.scala
```diff
@@ -75,6 +81,21 @@ object RestRequestMaker extends Logging {
...
+    }
+
+    def getDownloadStream(path: String): ClientFailableOp[DownloadStream] = {
+      val request = buildRequest(headers, baseUrl + path, _.get())
+      val response = getResponse(request)
+      for {
+        _ <- httpIsSuccessful(response)
+        contentlength <- extractContentLength(response)
+      } yield DownloadStream(response.body.byteStream, contentlength)
+    }
```
johnduffell: is it worth adding tests for these new methods?  I'm not sure if the rest of this file is actually tested specifically, it's probably just running by chance due to the other tests.

handlers/zuora-reports/src/test/scala/com/gu/zuora/reports/ReportsManualEffectsTest.scala
```diff
@@ -0,0 +1,62 @@
...
+  def querierTest {
+
+    val expected = QuerierResponse(
+      jobId = "something",
+      name = "bla"
+    )
+    val response = for {
+      configAttempt <- S3ConfigLoad.load(Stage("DEV")).toEither.disjunction
+      config <- LoadConfig.parseConfig[StepsConfig](configAttempt)
+      zuoraRequests = ZuoraAquaRequestMaker(RawEffects.response, config.stepsConfig.zuoraRestConfig)
```
johnduffell: the above 3 lines are basically the same for all the 3 functions, could be split into one function

handlers/zuora-reports/src/test/scala/com/gu/zuora/reports/ReportsManualEffectsTest.scala
```diff
@@ -0,0 +1,62 @@
...
+import com.gu.test.EffectsTest
+import com.gu.util.config.{LoadConfig, Stage}
+import com.gu.zuora.reports.ReportsLambda.StepsConfig
+import com.gu.zuora.reports.aqua.ZuoraAquaRequestMaker
+import org.scalatest.{FlatSpec, Matchers}
+import scalaz.\/-
+import scalaz.syntax.std.either._
+
+object ReportsManualEffectsTest extends App {
+  def querierTest {
```
johnduffell: codacy is mentioning a problem here, I thought the correct sytax was `def querierTest = {` but I'm not sure if this is a short hand for when it's not returning something?

handlers/zuora-reports/README.md
johnduffell: agree no one could directly attack, it jsut provides a layer where we could force people to specify the query in a more structured way.
I'm imagining a system where you can do a SAR through a web form and download your zuora data as a zip.  If we have identity id as a parameter, we would have to build the query (correctly) in the thing that calls this.  Rather if we only accepted queries in a decomposed format, there wouldn't be the opportinuity to accidentally build `1234' OR identityId__c != '1234` into the query, as this code inside the lambda would resolve that centrally.

We just need to keep vigilant about how we're using things - I don't actually think it is important to do it here and now, but it's easy forgotten.

handlers/zuora-reports/src/main/scala/com/gu/zuora/reports/Handlers.scala
```diff
@@ -5,14 +5,29 @@ import java.io.{InputStream, OutputStream}
...
+  private def defaultWiring[REQ, RES](call: Requests => AquaCall[REQ, RES])(config: Config[StepsConfig]) = call(ZuoraAquaRequestMaker(RawEffects.response, config.stepsConfig.zuoraRestConfig))
+
+  def queryHandler(inputStream: InputStream, outputStream: OutputStream, context: Context) =
+    ReportsLambda[QuerierRequest, QuerierResponse](RawEffects.stage, RawEffects.s3Load, LambdaIO(inputStream, outputStream, context), defaultWiring(Querier.apply))
+
+  def fetchResultsHandler(inputStream: InputStream, outputStream: OutputStream, context: Context) =
+    ReportsLambda[JobResultRequest, JobResult](RawEffects.stage, RawEffects.s3Load, LambdaIO(inputStream, outputStream, context), defaultWiring(GetJobResult.apply))
+
+  def fetchFileHandler(inputStream: InputStream, outputStream: OutputStream, context: Context) = {
+    def customWiring(config: Config[StepsConfig]) = {
```
johnduffell: wiring is quite a good idea for a name, seems clear

johnduffell:
great work! 👍 I think this will be very useful to take away the hassle from querying zuora.

PR #110, opened 2018-06-06
clean up the zuora library

johnduffell:
In preparation for making zuora querier not force people to turn queries into a string, I noticed that the zuora lib had a lot of baggage left over from when it was extracted from the original lambdas.

I decided to do the tidyup as a standalone PR so that the way is clear for the querier improvements.

I have tidied lambda specific stuff out into the original lambdas, and also done a bit of other spring cleaning along the way.
@jacobwinch @pvighi @twrichards @paulbrown1982


handlers/catalog-service/src/test/scala/com/gu/catalogService/CatalogServiceStepsTest.scala
```diff
@@ -29,7 +29,7 @@ class CatalogServiceStepsTest extends FlatSpec with Matchers {
         failureResponseEffects.response,
         successfulResponseEffects.stage,
         successfulResponseEffects.zuoraEnvironment,
-        failureResponseEffects.rawEffects.s3Load,
```
johnduffell: I have got rid of the combined "raweffects" case class, now everything is independent in the raweffects object

handlers/digital-subscription-expiry/src/main/scala/com/gu/digitalSubscriptionExpiry/zuora/SetActivationDate.scala
```diff
@@ -1,21 +1,24 @@
...
 
 object SetActivationDate extends Logging {
 
   case class UpdateRequestBody(ActivationDate__c: String)
 
   implicit val writes = Json.writes[UpdateRequestBody]
 
+  implicit val unitReads: Reads[Unit] =
+    Reads(_ => JsSuccess(()))
+
```
johnduffell: some put or post requests don't care about the response.  Since by the type definition we have to deserialise something, we can define a `Unit` reads.
Previously we just had a shared one in the zuora lib, but it didn't really belong there.   This is now everywhere, and another illustration of the lack of my inspiration over how to fit HTTP stuff into the type system neatly.

src/main/scala/com/gu/stripeCustomerSourceUpdated/SourceUpdatedSteps.scala
```diff
@@ -54,20 +54,23 @@ object SourceUpdatedSteps extends Logging {
   }
 
   def getPaymentMethodsToUpdate(
-    requests: Requests,
+    requests: Requests
```
johnduffell: removing the comma unleashed scalariform on the rest of the file

lib/zuora/src/main/scala/com/gu/util/zuora/Zuora.scala
```diff
@@ -1,133 +0,0 @@
-package com.gu.util.zuora
```
johnduffell: all the operations from here moved to the separate files in the root src folder (near the bottom of this diff)

handlers/digital-subscription-expiry/src/test/scala/com/gu/digitalSubscriptionExpiry/DigitalSubscriptionExpiryStepsTest.scala
```diff
@@ -74,7 +74,7 @@ class DigitalSubscriptionExpiryStepsTest extends FlatSpec with Matchers {
       setActivationDate = setActivationDate,
       getAccountSummary = getAccount,
       getSubscriptionExpiry = getSubExpiry,
-      skipActivationDateUpdate = skipActivationDateUpdate,
+      skipActivationDateUpdate = skipActivationDateUpdate
```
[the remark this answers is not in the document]
pvighi: they really confuse me as well

handlers/digital-subscription-expiry/src/main/scala/com/gu/digitalSubscriptionExpiry/zuora/SetActivationDate.scala
pvighi: maybe we can put it somewhere else  instead of creating multiple copies of it

src/main/scala/com/gu/util/zuora/ZuoraDisableAutoPay.scala
```diff
@@ -0,0 +1,22 @@
...
+import play.api.libs.json.{JsSuccess, Json, Reads, Writes}
+
+object ZuoraDisableAutoPay {
+
+  case class AccountUpdate(autoPay: Boolean)
+
+  implicit val accountUpdateWrites = new Writes[AccountUpdate] {
+    def writes(accountUpdate: AccountUpdate) = Json.obj(
+      "autoPay" -> accountUpdate.autoPay
+    )
```
pvighi: why not  this? :
 ```
implicit val writes = Json.writes[AccountUpdate]
```

johnduffell: didn't even look at the code, just moved it 😆

handlers/digital-subscription-expiry/src/main/scala/com/gu/digitalSubscriptionExpiry/zuora/SetActivationDate.scala
johnduffell: agree, I just couldn't think where.  It isn't really part of the zuora library, it's kind of call specific. (ie. the call specifically doesn't want anything)

PR #111, opened 2018-06-08
rename FailableOp to ApiGatewayOp

johnduffell:
We have a vague concept of FailableOp which is actually a `scalaz.\/` and is actually a way for API Gateway handlers to send back a result.

I have changed it to use ApiGatewayOp and made it a wrapper  so we can use it as a proper object rather than a type alias.  Here is the interesting bit: https://github.com/guardian/zuora-auto-cancel/pull/111/files#diff-2595522d2b3616dd6bfbf98bff7d3782R14

Naming us up for grabs, I have given it an initial name for now.

@pvighi @jacobwinch @twrichards @paulbrown1982


lib/handler/src/main/scala/com/gu/util/reader/Types.scala
```diff
@@ -5,32 +5,71 @@ import com.gu.util.apigateway.ApiGatewayResponse
...
+  }
+
+  implicit val apiGatewayOpM: Monad[ApiGatewayOp] = {
+    val a = implicitly[Monad[({ type Q[X] = scalaz.\/[ApiResponse, X] })#Q]]
+    new Monad[ApiGatewayOp] {
+      override def bind[A, B](fa: ApiGatewayOp[A])(f: A => ApiGatewayOp[B]): ApiGatewayOp[B] = ApiGatewayOp(a.bind(fa.underlying)(f.andThen(_.underlying)))
+
+      override def point[A](a: => A): ApiGatewayOp[A] = ApiGatewayOp(scalaz.\/-(a))
+    }
+  }
```
johnduffell: this is needed for the traverseU that I found for checking reader types for Agent subs.

lib/handler/src/main/scala/com/gu/util/reader/Types.scala
```diff
@@ -5,32 +5,71 @@ import com.gu.util.apigateway.ApiGatewayResponse
...
+
+  implicit class ApiResponseOps(apiGatewayOp: ApiGatewayOp[ApiResponse]) {
+
+    def apiResponse: ApiResponse =
+      apiGatewayOp.underlying.fold(identity, identity)
```
johnduffell: this function is only available if you have returned an ContinueProcessing(ApiResponse) and want to return the ApiResponse.  Hopfully this is a bit better than returning FailableOp[Unit] and magically making it a success.

handlers/digital-subscription-expiry/src/main/scala/com/gu/digitalSubscriptionExpiry/emergencyToken/GetTokenExpiry.scala
```diff
@@ -4,19 +4,19 @@ import java.time.LocalDate
...
+import com.gu.util.reader.Types.ApiGatewayOp
+import ApiGatewayOp.{ReturnWithResponse, ContinueProcessing}
 import com.gu.digitalSubscriptionExpiry.responses.DigitalSubscriptionApiResponses._
 import com.gu.digitalSubscriptionExpiry.responses.{Expiry, ExpiryType, SuccessResponse}
 
 import scala.util.{Success, Try}
 
 object GetTokenExpiry {
-  def apply(emergencyTokens: EmergencyTokens, today: () => LocalDate)(subscriberId: String): FailableOp[Unit] = {
+  def apply(emergencyTokens: EmergencyTokens, today: () => LocalDate)(subscriberId: String): ApiGatewayOp[Unit] = {
```
johnduffell: this still seems unclear, even though my change was supposed to help with this.. maybe there's a further improvement needed:
Given my guideline that a function should have the signature that makes mose sense for its purpose, if someone asked you for a function to work out whether a token is a valid emergency token and if so return the date of expiry it would be nothing to do with ApiGateway (or FailableOp)
`def getEmergencyTokenExpiry: SubscriberId => Option[ExpiryDate]`

handlers/digital-subscription-expiry/src/main/scala/com/gu/digitalSubscriptionExpiry/zuora/GetAccountSummary.scala
```diff
@@ -27,7 +25,7 @@ object GetAccountSummary {
       (__ \ "soldToContact" \ "zipCode").readNullable[String]
     )(AccountSummaryResult.apply _)
 
-  def apply(requests: Requests)(accountId: AccountId): FailableOp[AccountSummaryResult] =
-    requests.get[AccountSummaryResult](s"accounts/${accountId.value}/summary").leftMap(clientFail => ApiGatewayResponse.internalServerError(s"zuora client fail: ${clientFail.message}"))
+  def apply(requests: Requests)(accountId: AccountId): ClientFailableOp[AccountSummaryResult] =
```
johnduffell: in future we should probably fix `ClientFailableOp` similarly to be `RestResponse`, and maybe make it a proper case class rather than an either.

handlers/digital-subscription-expiry/src/main/scala/com/gu/digitalSubscriptionExpiry/zuora/GetSubscriptionExpiry.scala
```diff
@@ -47,9 +47,9 @@ object GetSubscriptionExpiry {
     } else None
   }
 
-  def apply(today: () => LocalDate)(providedPassword: String, subscription: SubscriptionResult, accountSummary: AccountSummaryResult): FailableOp[Unit] =
+  def apply(today: () => LocalDate)(providedPassword: String, subscription: SubscriptionResult, accountSummary: AccountSummaryResult): ApiResponse =
```
johnduffell: this now returns ApiResponse, but it's actually suffering a similar problem to my comment on `GetTokenExpiry` above.  It should relly return an Option[Expiry] or similar and something else can turn that into whatever kind of response needed for Api gateway or otherwise.

src/test/scala/com/gu/autoCancel/AutoCancelHandlerTest.scala
```diff
@@ -24,7 +24,7 @@ class AutoCancelHandlerTest extends FlatSpec {
   "filterInvalidAccount" should "return a left if AutoPay = false" in {
     val autoCancelCallout = fakeCallout(false)
     val either = apply(autoCancelCallout, false)
```
pvighi: i think this is not an either anymore

src/test/scala/com/gu/autoCancel/AutoCancelDataCollectionFilterTest.scala
```diff
@@ -40,42 +40,42 @@ class AutoCancelDataCollectionFilterTest extends FlatSpec {
 
   "getCancellationDateFromInvoices" should "return a left if no overdue invoices are found" in {
     val either = getCancellationDateFromInvoices(AccountSummary(basicInfo, List(subscription), List(invoiceZeroBalance, invoiceNotDue, invoiceNotPosted)), LocalDate.now)
```
pvighi: this file as well has vals call either

johnduffell:
seems to pass the health checks in code, so we'll see what happens in prod!

PR #112, opened 2018-06-08
make zuora reports a lib instead of a project

pvighi:
Because of problems implementing the old approach of sharing lambdas we will now just share the code and let every project have its own lambda.

To make the diff more manageable this first PR just moves zuora reports to a lib and creates a zuora retention placeholder project.

No code was changed to keep the diff simple.


handlers/zuora-retention/cfn.yaml
```diff
@@ -12,18 +12,18 @@ Parameters:
 Mappings:
   StageVariables:
     CODE:
-      reportsBucketName : "zuora-reports-code"
+      retentionBucketName : "zuora-retention-code"
```
pvighi: this change will cause the placeholder retention project from being able to fetch files as the destination bucket is hard coded to zuora-reports-[STAGE]. 
In a future PR the file fetcher lambda will be refactored to enable custom bucket names.

johnduffell:
great 👍

PR #113, opened 2018-06-12
moved SalesforceAuthenticate into 'lib'

twrichards:
Moved SalesforceAuthenticate (and accompanying tests) into 'lib' to make available for re-use (initially for a new lambda for case creation, for new cancellation flow)

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -29,9 +29,9 @@ val scalaSettings = Seq(
       .setPreference(NewlineAtEndOfFile, true)
   },libraryDependencies += "com.lihaoyi" %% "acyclic" % "0.1.7" % "provided",
 
-    autoCompilerPlugins := true,
+  autoCompilerPlugins := true,
 
-addCompilerPlugin("com.lihaoyi" %% "acyclic" % "0.1.7"),
+  addCompilerPlugin("com.lihaoyi" %% "acyclic" % "0.1.7"),
   scalacOptions += "-P:acyclic:force"
 )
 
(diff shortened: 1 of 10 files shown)


build.sbt
```diff
@@ -56,41 +56,46 @@ def all(theProject: Project) = theProject.settings(scalaSettings, testSettings).
 
 // ==== START libraries ====
 
-lazy val zuora = all(project in file("lib/zuora")).dependsOn(restHttp).settings(
-  libraryDependencies ++= Seq(
-    okhttp3, logging, scalaz, playJson, scalatest,
-    "com.fasterxml.jackson.core" % "jackson-databind" % "2.8.11.1"
+lazy val test = all(project in file("lib/test"))
```
johnduffell: given my experience doing a similar thing after you did today, I don't think we need to change the order if we don't want (but it doesn't hurt anything)

johnduffell: it has completely confused github's diff though

build.sbt
```diff
@@ -56,41 +56,46 @@ def all(theProject: Project) = theProject.settings(scalaSettings, testSettings).
...
+  )
 
-// to aid testability, only the actual handlers called as a lambda can depend on this
-lazy val effects = all(project in file("lib/effects")).dependsOn(handler).settings(
-  libraryDependencies ++= Seq(
-    okhttp3, logging, scalaz, playJson, scalatest,
-    "com.amazonaws" % "aws-java-sdk-s3" % "1.11.311",
-    "com.fasterxml.jackson.core" % "jackson-databind" % "2.8.11.1"
+lazy val salesforce = all(project in file("lib/salesforce"))
+  .dependsOn(restHttp, handler, effects % "test->test", testDep)
```
johnduffell: random comment - as mentioned earlier I think testDep and effects%test would always be used together, so maybe be combined one way or another in future.

project/Dependencies.scala
```diff
@@ -7,6 +7,7 @@ object Dependencies {
   val scalaz = "org.scalaz" %% "scalaz-core" % "7.2.18"
   val playJson = "com.typesafe.play" %% "play-json" % "2.6.9"
   val scalatest = "org.scalatest" %% "scalatest" % "3.0.1" % "test"
-
+  val jacksonDatabind = "com.fasterxml.jackson.core" % "jackson-databind" % "2.8.11.1"
+  val awsS3 = "com.amazonaws" % "aws-java-sdk-s3" % "1.11.311"
```
johnduffell: good extracting there 🏅

johnduffell:
good solid improvement, very good work splitting into such an easy to review chunk (so clearly I shouldn't have taken too long to look...) just a minor comment about the reordering but it's good to go 👍 🚀  keep things tidy 😆

PR #114, opened 2018-06-13
build zuora queries in a fundamentally safe way

johnduffell:
At the moment we leave it up to callers to construct a zuora query by concatenating together strings.  This is a fundamentally broken and dangerous way to work - no application code should build "strings" which have higher level meaning, it should only happen in specifically written serialisation code, and well tested as such.

This PR introduces a query serialiser as a custom string interpolator.  It builds a Safe query object which can then be further serialised into Json by the normal js writes.  All of the inserts are classed as string literals, and as such it will insert the appropriate single quotes, plus escaping of such.

As an interesting aside which would be particularly relevant to @pvighi , this code is designed for the zoql query interface, where backslash is used to escape any special characters: https://knowledgecenter.zuora.com/DC_Developers/K_Zuora_Object_Query_Language/Filter_Statements#Reserved_and_Escaped_Characters
However, in the "zoql export" api, which takes pretty much the same types of queries, in typical zuora style they have gone for escaping single quotes by doubling them up.  So we will need a separate serialiser for that at some point https://knowledgecenter.zuora.com/DC_Developers/M_Export_ZOQL/A_Select_Statement#Reserved_and_Escaped_Characters 

This is about ready to go, please take a (another) look @pvighi @paulbrown1982 @twrichards @jacobwinch   Expecially interested in ease of use/clarity and safety.  There are a fair few concepts in there I'd rather avoid, apart from the readabilirty.  Custom string interpolation, varargs extracted separately with type classes on each arg, there's a lot of weirdness going on.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -64,13 +64,23 @@ lazy val test = all(project in file("lib/test"))
 val testDep = test % "test->test"
 
 lazy val zuora = all(project in file("lib/zuora"))
-  .dependsOn(restHttp)
+  .dependsOn(
+    restHttp,
+    testDep,
+    handler % "test->test",//only for the config, which should ideally be split out
+    effects % "test->test"
+  )
   .settings(
     libraryDependencies ++= Seq(okhttp3, logging, scalaz, playJson, scalatest, jacksonDatabind)
   )
 
 lazy val salesforce = all(project in file("lib/salesforce"))
-  .dependsOn(restHttp, handler, effects % "test->test", testDep)
+  .dependsOn(
+    restHttp,
+    handler,// % "test->test" TODO make this dep only in test - SF client shouldn't depends on ApiGateway
+    effects % "test->test",
+    testDep
+  )
   .settings(
     libraryDependencies ++= Seq(okhttp3, logging, scalaz, playJson, scalatest)
   )
--- a/handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/zuora/CountZuoraAccountsForIdentityId.scala
+++ b/handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/zuora/CountZuoraAccountsForIdentityId.scala
@@ -2,7 +2,7 @@ package com.gu.identityBackfill.zuora
 
 import com.gu.identityBackfill.Types.{AccountId, IdentityId}
 import com.gu.util.zuora.RestRequestMaker.ClientFailableOp
-import com.gu.util.zuora.ZuoraQuery
+import com.gu.util.zuora.ZuoraQuery.Query
 import com.gu.util.zuora.ZuoraQuery.ZuoraQuerier
 import play.api.libs.json.Json
 import scalaz.ListT
@@ -13,7 +13,7 @@ object CountZuoraAccountsForIdentityId {
   implicit val reads = Json.reads[WireResponse]
 
   def apply(zuoraQuerier: ZuoraQuerier)(identityId: IdentityId): ClientFailableOp[Int] = {
-    val accountByIdentityQuery = ZuoraQuery.Query(s"SELECT Id FROM Account where IdentityId__c='${identityId.value}'")
+    val accountByIdentityQuery = zoql"SELECT Id FROM Account where IdentityId__c=${identityId.value}"
     val accounts = for {
       accountsWithEmail <- ListT[ClientFailableOp, WireResponse](
         zuoraQuerier[WireResponse](accountByIdentityQuery).map(_.records)
--- a/handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/zuora/CountZuoraAccountsForIdentityId.scala
+++ b/handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/zuora/CountZuoraAccountsForIdentityId.scala
@@ -2,7 +2,8 @@ package com.gu.identityBackfill.zuora
 
 import com.gu.identityBackfill.Types.IdentityId
 import com.gu.util.zuora.RestRequestMaker.ClientFailableOp
-import com.gu.util.zuora.ZuoraQuery.{Query, ZuoraQuerier}
+import com.gu.util.zuora.SafeQueryBuilder.Implicits._
+import com.gu.util.zuora.ZuoraQuery.ZuoraQuerier
 import play.api.libs.json.Json
 
 object CountZuoraAccountsForIdentityId {
--- a/handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/zuora/CountZuoraAccountsForIdentityId.scala
+++ b/handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/zuora/CountZuoraAccountsForIdentityId.scala
@@ -1,26 +1,19 @@
 package com.gu.identityBackfill.zuora
 
-import com.gu.identityBackfill.Types.{AccountId, IdentityId}
+import com.gu.identityBackfill.Types.IdentityId
 import com.gu.util.zuora.RestRequestMaker.ClientFailableOp
-import com.gu.util.zuora.ZuoraQuery.Query
-import com.gu.util.zuora.ZuoraQuery.ZuoraQuerier
+import com.gu.util.zuora.ZuoraQuery.{Query, ZuoraQuerier}
 import play.api.libs.json.Json
-import scalaz.ListT
 
 object CountZuoraAccountsForIdentityId {
 
   case class WireResponse(Id: String)
   implicit val reads = Json.reads[WireResponse]
 
-  def apply(zuoraQuerier: ZuoraQuerier)(identityId: IdentityId): ClientFailableOp[Int] = {
-    val accountByIdentityQuery = zoql"SELECT Id FROM Account where IdentityId__c=${identityId.value}"
-    val accounts = for {
-      accountsWithEmail <- ListT[ClientFailableOp, WireResponse](
-        zuoraQuerier[WireResponse](accountByIdentityQuery).map(_.records)
-      )
-    } yield AccountId(accountsWithEmail.Id)
-
-    accounts.run.map(_.size)
-  }
+  def apply(zuoraQuerier: ZuoraQuerier)(identityId: IdentityId): ClientFailableOp[Int] =
+    for {
+      accountByIdentityQuery <- zoql"SELECT Id FROM Account where IdentityId__c=${identityId.value}"
+      noOfAccountsWithID <- zuoraQuerier[WireResponse](accountByIdentityQuery).map(_.size)
+    } yield noOfAccountsWithID
 
 }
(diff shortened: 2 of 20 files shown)


handlers/identity-retention/src/main/scala/com/gu/identityRetention/SubscriptionsForAccounts.scala
```diff
@@ -17,15 +16,22 @@ object SubscriptionsForAccounts {
...
+  def buildQuery(accountsToQuery: List[AccountId]): SanitisedQuery = {
+    zoql"""select
+       | id,
+       | name,
+       | status,
+       | termEndDate
+       | from subscription
+       | where ${accountsToQuery.map(acc => zoql"status != 'Expired' and accountId = ${acc.value}").or}
+       |"""
+      .stripMarginAndNewline
```
johnduffell: not sure whether the `.or` is a bit hidden away, maybe it would be better as a proper function?

lib/zuora/src/main/scala/com/gu/util/zuora/ZuoraQuery.scala
```diff
@@ -1,18 +1,94 @@
...
+  }
+
+  def conv[A](a: A)(implicit aString: Conv[A]) = aString(a)
+
+  @implicitNotFound("implicitNotFound: ZuoraQuery: can only insert a string literal or a sanitised query into a parent query")
+  trait Conv[A] {
+    def apply(a: A): ClientFailableOp[String]
+  }
+
+  implicit val aaa = new Conv[SanitisedQuery] {
```
pvighi: temp val name ?

lib/zuora/src/test/scala/com/gu/util/ZuoraQueryTest.scala
```diff
@@ -0,0 +1,83 @@
...
+
+import com.gu.util.zuora.RestRequestMaker.{ClientFailableOp, GenericError}
+import com.gu.util.zuora.ZuoraQuery
+import com.gu.util.zuora.ZuoraQuery.{OrTraverse, Query, SanitisedQuery}
+import org.scalatest._
+import scalaz.{-\/, \/-}
+
+class WireQueryEscapeTest extends FlatSpec with Matchers {
+
+  it should "escape signel quotes" in {
```
pvighi: single ?

lib/zuora/src/test/scala/com/gu/util/ZuoraQueryTest.scala
```diff
@@ -0,0 +1,83 @@
...
+    val actual = ZuoraQuery.stringInsertToQueryLiteral("""a very "nice" query""")
+    actual should be(\/-("""'a very \"nice\" query'"""))
+  }
+
+  it should "escape backslashes" in {
+    val actual = ZuoraQuery.stringInsertToQueryLiteral("""a very \ query""")
+    actual should be(\/-("""'a very \\ query'"""))
+  }
+
+  it should "escape signel quotes double check the length" in {
```
pvighi: single again

lib/zuora/src/test/scala/com/gu/util/ZuoraQueryTest.scala
```diff
@@ -0,0 +1,83 @@
...
+    val actual = ZuoraQuery.stringInsertToQueryLiteral("""'""")
+    actual.map(_.length) should be(\/-(4))
+  }
+
+  it should "escape double quotes double check the length" in {
+    val actual = ZuoraQuery.stringInsertToQueryLiteral(""""""")
+    actual.map(_.length) should be(\/-(4))
+  }
+
+  it should "escape backslasgh double check the length" in {
```
pvighi: backslash

lib/zuora/src/main/scala/com/gu/util/zuora/SafeQueryBuilder.scala
```diff
@@ -0,0 +1,89 @@
...
+          .flatMap { orig =>
+            if (orig.replaceAll("""\p{Cntrl}""", "") == orig)
+              \/-(orig)
+            else
+              -\/(GenericError(s"control characters can't be inserted into a query: $orig"))
+          }
+          .map(_.replaceAll("""\\""", """\\\\""")
+            .replaceAll("'", """\\'""")
+            .replaceAll(""""""", """\\""""))
+        sanitised.map(sanitised => s"'$sanitised'")
```
pvighi: this is a little confusing when everything is called sanitised

build.sbt
```diff
@@ -64,13 +64,23 @@ lazy val test = all(project in file("lib/test"))
 val testDep = test % "test->test"
 
 lazy val zuora = all(project in file("lib/zuora"))
-  .dependsOn(restHttp)
+  .dependsOn(
+    restHttp,
+    testDep,
+    handler % "test->test",//only for the config, which should ideally be split out
+    effects % "test->test"
```
johnduffell: these extra deps are needed because I wanted to add an effects test.  ideally we would only need one dep for an effectstest.

handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/zuora/CountZuoraAccountsForIdentityId.scala
```diff
@@ -1,26 +1,20 @@
...
-      )
-    } yield AccountId(accountsWithEmail.Id)
-
-    accounts.run.map(_.size)
-  }
+  def apply(zuoraQuerier: ZuoraQuerier)(identityId: IdentityId): ClientFailableOp[Int] =
+    for {
+      accountByIdentityQuery <- zoql"SELECT Id FROM Account where IdentityId__c=${identityId.value}"
+      noOfAccountsWithID <- zuoraQuerier[WireResponse](accountByIdentityQuery).map(_.size)
+    } yield noOfAccountsWithID
```
johnduffell: the .map(_.size) on line 17 could be just a .size on line 18

handlers/identity-retention/src/main/scala/com/gu/identityRetention/Types.scala
```diff
@@ -2,7 +2,7 @@ package com.gu.identityRetention
 
 object Types {
 
-  case class IdentityId(value: Long)
+  case class IdentityId(value: String)
```
johnduffell: IdentityId is stored as a string in zuora, and we can no longer automtically convert a number to a string literal in a query

lib/zuora/src/main/scala/com/gu/util/zuora/ZuoraQuery.scala
```diff
@@ -1,18 +1,18 @@
...
 
-  case class Query(queryString: String)
-
   case class QueryLocator(value: String) extends AnyVal
 
   case class QueryResult[QUERYRECORD](records: List[QUERYRECORD], size: Int, done: Boolean, queryLocator: Option[QueryLocator])
 
-  implicit val queryW: Writes[Query] = Json.writes[Query]
+  // can't generate from macro as it needs an apply method for some reason which we'd rather not expose
+  implicit val queryW: Writes[SafeQuery] = (JsPath \ "queryString").write[String].contramap(_.queryString)
```
johnduffell: contramap is map the input parameter, which means we're going backwards (map goes forwards)
so contramap from String to SafeQuery needs a function from SafeQuery to String

lib/zuora/src/test/scala/com/gu/util/SafeQueryBuilderTest.scala
```diff
@@ -0,0 +1,85 @@
...
+    val actual = makeSafeStringIntoQueryLiteral(""""""")
+    actual.map(_.length) should be(\/-(4))
+  }
+
+  it should "escape backslash double check the length" in {
+    val actual = makeSafeStringIntoQueryLiteral("""\""")
+    actual.map(_.length) should be(\/-(4))
+  }
+
+  it should "remove control chars - this is not 100% safe - we should reject completely" in {
```
pvighi: this test description seems outdated since it now rejects control chars

lib/zuora/src/main/scala/com/gu/util/zuora/SafeQueryBuilder.scala
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    implicit val makeSafeAlreadySafe: MakeSafe[SafeQuery] = new MakeSafe[SafeQuery] {
+      override def apply(safeQuery: SafeQuery): ClientFailableOp[String] =
+        \/-(safeQuery.queryString)
+    }
+
+    implicit val makeSafeStringIntoQueryLiteral: MakeSafe[String] = new MakeSafe[String] {
+      override def apply(untrusted: String): ClientFailableOp[String] = {
+        if (untrusted.replaceAll("""\p{Cntrl}""", "") != untrusted) {
```
pvighi: minor: we can probably change this to untrusted.replaceFirst right?
there is probably a slightly different regex we can use to do untrusted.match([some_other_regex]) but I hate regex so I don't know what that is without some lengthy investigation

lib/zuora/src/test/scala/com/gu/util/SafeQueryBuilderTest.scala
johnduffell: not only that but the test shoudl check each one separately

lib/zuora/src/main/scala/com/gu/util/zuora/ZuoraQuery.scala
pvighi: is there any reason we cannot do something more explicit like this:
https://github.com/guardian/zuora-auto-cancel/blob/0be5e9bbefa97456a157d57759a672bae3068f2a/lib/handler/src/main/scala/com/gu/util/apigateway/ApiGatewayResponse.scala#L26-L32

I think it would be easier to understand

lib/zuora/src/test/scala/com/gu/util/SafeQueryBuilderTest.scala
```diff
@@ -0,0 +1,85 @@
...
+    val ids = List("anna", "bill")
+    val actual = for {
+      insert <- OrTraverse(ids)({ id => zoql"""id = $id""" })
+      wholeQuery <- zoql"""select hi from table where $insert"""
+    } yield wholeQuery
+    actual.map(_.queryString) should be(\/-("""select hi from table where id = 'anna' or id = 'bill'"""))
+  }
+
+  //POST query - SELECT Id, promotioncode__c FROM Subscription where PromotionCode__c = 'qwerty\"asdf\'zxcv\\1234'
+  // NOTE for "zoql export" we don't escape anything, just double up on single quotes only. tested all june 2018
```
pvighi: are these two comment lines meant to be here ?

lib/zuora/src/main/scala/com/gu/util/zuora/ZuoraQuery.scala
johnduffell: great spot, intellij even make an extra simplification
```scala
  implicit val queryW: Writes[SafeQuery] =
    (response: SafeQuery) => Json.obj(
      "queryString" -> response.queryString
    )
```

lib/zuora/src/main/scala/com/gu/util/zuora/ZuoraQuery.scala
```diff
@@ -6,13 +6,52 @@ import play.api.libs.json._
...
+  object Or {
+    def apply(queries: List[SanitisedQuery]): SanitisedQuery = {
+      new SanitisedQuery(queries.map(_.queryString).mkString(" or "))
+    }
+  }
+
+  object SanitisedQuery {
+
+    def sanitise(input: String): String = {
+      val sanitised = input
```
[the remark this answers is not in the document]
johnduffell: I guess this comment just took a long time to appear?

johnduffell:
OK thanks both of you, I think I've fixed everything up, all the comments are collapsed, effects tests are passing, codacy is happy, so I'll ship this afternoon unless any new comments before then!

PR #115, opened 2018-06-14
zuora retention pr 2

pvighi:
this PR adds a Zuora retention state machine that does generates a csv file with a list of accounts to set to "DoNotProcess". in a future pr another step will be added to read the file and actually update all accounts.


handlers/zuora-retention/src/main/scala/com/gu/zuora/retention/Diff.scala
```diff
@@ -0,0 +1,42 @@
...
+this class was needed because iterators won't allow peeking without advancing
+ */
+case class SortedCrmIdIterator(crmIdIterator: Iterator[String]) {
+  var currentCrmId: Option[String] = None
+
+  def nextGreaterOrEqual(crmId: String): Option[String] = {
+    /*salesforce 18 character ids have lower and upper case letters but are actually case insensitive, see : https://help.salesforce.com/articleView?id=000004383&language=en_US&type=1
+      we need to do case insensitive comparsion to match the alphabetical order in Zuora query results
+    */
+    while (crmIdIterator.hasNext && !currentCrmId.exists(_.toLowerCase >= crmId.toLowerCase)) {
```
pvighi: @johnduffell do you think a tail recursive function would be more clear here?

handlers/zuora-retention/src/main/scala/com/gu/zuora/retention/Diff.scala
```diff
@@ -0,0 +1,42 @@
...
+      }
+    }
+    List(candidatesHeader).iterator ++ valueRows
+  }
+}
+/*
+this class was needed because iterators won't allow peeking without advancing
+ */
+case class SortedCrmIdIterator(crmIdIterator: Iterator[String]) {
+  var currentCrmId: Option[String] = None
```
pvighi: this whole file is not very FP because the iterator is not immutable so it kinda makes it harder to avoid things like using var

handlers/zuora-retention/src/main/scala/com/gu/zuora/retention/FilterCandidates.scala
```diff
@@ -0,0 +1,100 @@
...
+object FilterCandidates {
+
+  def apply(inputStream: InputStream, outputStream: OutputStream, context: Context): Unit = {
+    val buckets = Map(
+      "PROD" -> "zuora-retention-prod",
+      "CODE" -> "zuora-retention-code",
+      "DEV" -> "zuora-retention-dev"
+    )
+
+    val wiredS3Upload = uploadToS3(RawEffects.s3Write, buckets(RawEffects.stage.value.toUpperCase), "doNotProcessList.csv") _
```
pvighi: I will probably change this in either this or a future PR to make every execution of the state machine upload files to a new subdirectory and / or file name. The way it is now leaves the opportunity for one execution picking up the files of another.

handlers/zuora-retention/src/main/scala/com/gu/zuora/retention/FilterCandidates.scala
```diff
@@ -0,0 +1,100 @@
...
+
+  def uploadToS3(
+    s3Write: PutObjectRequest => Try[PutObjectResult],
+    bucket: String,
+    key: String
+  )(filteredCandidates: Iterator[String]) = {
+    val uploadLocation = s"s3://$bucket/$key"
+    logger.info(s"uploading do do not process list to $uploadLocation")
+
+    val stringData = filteredCandidates.toList.mkString("\n")
```
pvighi: I didn't find a way to write from an Iterator[String] with the file lines directly to s3 without putting everything in memory first as s3 needs the content size to be specified first.
hopefully this is not a huge problem since this is the final result file so it should not be that big

handlers/zuora-retention/src/main/scala/com/gu/zuora/retention/Diff.scala
```diff
@@ -0,0 +1,42 @@
+package com.gu.zuora.retention
+
+object Diff {
```
johnduffell: maybe we can think of a more descriptive name at some point

handlers/zuora-retention/src/main/scala/com/gu/zuora/retention/Diff.scala
```diff
@@ -0,0 +1,42 @@
...
+  /**
+   * Returns an iterator for the lines in candidateLines with crmIds that are not in the exclusionLines
+   * The candidates and exclusion iterators are expected to iterate ascending CrmId order.
+   * The point of this is to avoid loading the whole exclusionLines in memory.
+   */
+  def apply(candidateLines: Iterator[String], exclusionLines: Iterator[String]): Iterator[String] = {
+    exclusionLines.next //skip header
+    val exclusionCrmIds = SortedCrmIdIterator(exclusionLines)
+    val candidatesHeader = candidateLines.next()
+    val crmidLocation = candidatesHeader.split(",").indexOf(crmIdColName)
```
johnduffell: it would be nice to separate out the code that deals with CSV from the logic.
even if we don't want to fully parse into a nice case class, we could still abstract the picking of the crmid.
even one function that takes an Iterator[String] and produces a Iterator[HandyClass(crmid:String, originalCSVLine:String)] would be ok.  Then we don't need to mention commas and splitting anwehere in the trickly part of the code

handlers/zuora-retention/src/main/scala/com/gu/zuora/retention/Diff.scala
```diff
@@ -0,0 +1,42 @@
...
+    val valueRows = candidateLines.filterNot { line =>
+      line.trim.isEmpty || {
+        val crmId = line.split(",")(crmidLocation).trim
+        val comp = exclusionCrmIds.nextGreaterOrEqual(crmId)
+        comp.exists(_.toLowerCase == crmId.toLowerCase)
```
johnduffell: since we only care about no case, I would map the iterators both to lower case ahead of time so we don't care about it any more in the logic

handlers/zuora-retention/src/main/scala/com/gu/zuora/retention/Diff.scala
johnduffell: I spent a while playing to see whether it's possible to use recursion instead of var to deal with a non recursive data structure and it's not.  I ended up more than doubling the size of the file, although separated out the concerns better in some ways.
I committed my changes but I haven't tidied up the naming but it might have some good ideas to get.  At least all the tests are still passing!
Have a look at https://github.com/guardian/zuora-auto-cancel/blob/diff-sorted-csvs/handlers/zuora-retention/src/main/scala/com/gu/zuora/retention/Diff.scala

handlers/zuora-retention/src/main/scala/com/gu/zuora/retention/FilterCandidates.scala
johnduffell: if it gets too big, by that time the chunked api would be able to handle it, so someone can add support for that.

handlers/zuora-retention/src/main/scala/com/gu/zuora/retention/Diff.scala
johnduffell: 😵

handlers/zuora-retention/src/main/scala/com/gu/zuora/retention/FilterCandidates.scala
johnduffell: is the case of stage not consistent?  we could make the map from Stage->bucket to simplify this wiring

handlers/zuora-retention/src/main/scala/com/gu/zuora/retention/FilterCandidates.scala
```diff
@@ -0,0 +1,100 @@
...
+  implicit val reads = Json.reads[FilterCandidatesRequest]
+}
+
+object FilterCandidatesResponse {
+  implicit val writes = Json.writes[FilterCandidatesResponse]
+}
+
+object FilterCandidates {
+
+  def apply(inputStream: InputStream, outputStream: OutputStream, context: Context): Unit = {
```
johnduffell: usually I find it worth just adding a comment mentioning this is referenced from cloudformation, just in case of refactoring

handlers/zuora-retention/src/main/scala/com/gu/zuora/retention/FilterCandidates.scala
```diff
@@ -0,0 +1,100 @@
...
+    val result = for {
+      request <- ParseRequest[FilterCandidatesRequest](lambdaIO.inputStream)
+      exclusionsUri <- getUri(request.fetched, ToAquaRequest.exclusionQueryName)
+      exclusionsIterator <- s3Iterator(exclusionsUri)
+      candidatesUri <- getUri(request.fetched, ToAquaRequest.candidatesQueryName)
+      candidatesIterator <- s3Iterator(candidatesUri)
+      filteredCandidates = diff(candidatesIterator, exclusionsIterator)
+      putResult <- uploadToS3(filteredCandidates)
+    } yield (putResult)
+    result match {
```
johnduffell: if it's a standard pattern of ParseRequest as json and SerialiseResponse as json or throw an exception, it would be worth making into a handler a bit like ApiGatewayHandler.  Then all the function for each lambda needs to do is return a Try, and the handler would do the talking to the lambda.  I think I've seen this "json or exception" pattern in another PR

handlers/zuora-retention/src/main/scala/com/gu/zuora/retention/Handlers.scala
```diff
@@ -0,0 +1,11 @@
+package com.gu.zuora.retention
+
+import com.gu.zuora.reports.ReportHandlers
+import com.gu.zuora.retention.query.{ToAquaRequest, RetentionQueryRequest}
+
+object Handlers extends ReportHandlers[RetentionQueryRequest] {
```
johnduffell: I'd rather avoid cake pattern/inheritance of code classes if possible, just due to the fact that things get "fat" easily.  I can see that it's equivalent to just passing in these parameters to a partially applied method, but in that case it becomes extemely clear/painful if you have too many dependencies, wheras with cake pattern there's no visual reminder of lots of deps.

handlers/zuora-retention/src/main/scala/com/gu/zuora/retention/query/RetentionQueryRequest.scala
```diff
@@ -0,0 +1,63 @@
...
+           | Subscription
+           |WHERE
+           | Account.CrmId != '' AND
+           | Status != 'Expired' AND
+           | Status != 'Draft'
+           |GROUP BY
+           | Account.CrmId
+           |HAVING
+           |  MAX(Status) = 'Cancelled' AND
+           |  (MIN(Status) = 'Active' OR MAX(SubscriptionEndDate) >= '$dateStr')
```
johnduffell: great, I assume brackets do work in aqua.  Also, this field is pretty safe, but we should be careful in future about untrusted strings as per the other PR

handlers/zuora-retention/src/main/scala/com/gu/zuora/retention/Diff.scala
pvighi: I would prefer to not map to lowercase the filtered file we generate at the end, although that would probably not break anything. But I will try to refactor this make this better while preserving the original crmIds in the results file

handlers/zuora-retention/src/test/scala/com/gu/zuora/retention/DiffTest.scala
```diff
@@ -0,0 +1,157 @@
...
+      """Account.Id,Account.CrmId,BillToContact.Id,SoldToContact.Id""".stripMargin
+
+    Diff(candidates.lines, excluded.lines).mkString("\n") shouldBe expected
+  }
+  it should "not skip ahead when ascii sorting differs from zuora order" in {
+    val candidates =
+      """Account.CrmId
+        |6
+        |a_lowercase
+        |Z_uppercase
```
johnduffell: I think this method of including IDs really helps document the tests and the purpose.  would it be feasible to do the same on the other tests, instead of crmId4 we could put crmId4_duplicated etc

handlers/zuora-retention/src/test/scala/com/gu/zuora/retention/FilterCandidatesEffectsTest.scala
```diff
@@ -0,0 +1,39 @@
...
+
+    val testInputStream = new ByteArrayInputStream(input.getBytes)
+    val files = List(
+      FetchedFile("asdasd", "candidatesQuery", "s3://zuora-reports-dev/someFile.csv"),
+      FetchedFile("asdasd", "exclusionQuery", "s3://zuora-reports-dev/exclusionQuery.csv")
+    )
+
+    val request = FilterCandidatesRequest(files)
+    FilterCandidates(testInputStream, null, null)
+    "bla" shouldBe ("bla")
```
johnduffell: is this test finished? it looks  like an end to end test but it's not checking anything, and it's called "do something"

pvighi: oops I didn't mean to push that

handlers/zuora-retention/src/main/scala/com/gu/zuora/retention/Diff.scala
pvighi: thanks for going though all that trouble, I'll steal some parts of that

handlers/zuora-retention/src/main/scala/com/gu/zuora/retention/Diff.scala
johnduffell: sounds good, as discussed since the tests are quite comprehensive and this code is very self contained, so it's not going to cause a wider issue with the codebase or become worse/spread over time.  So no need to spend time thinking about it

handlers/zuora-retention/src/test/scala/com/gu/zuora/retention/DiffTest.scala
```diff
@@ -9,93 +9,66 @@ class DiffTest extends AsyncFlatSpec {
...
-        |31,crmId3,33,34
-        |31,crmId3,33,34""".stripMargin
-
-    Diff(candidates.lines, excluded.lines).mkString("\n") shouldBe expected
+        |21,C,23,24
+        |31,E_duplicate,33,34
+        |31,E_duplicate,33,34""".stripMargin
+    val ss = Diff(candidates.lines, excluded.lines).mkString("\n")
+    println(ss)
+    ss shouldBe expected
```
johnduffell: probably don't need the ss and the println

handlers/zuora-retention/src/test/scala/com/gu/zuora/retention/DiffTest.scala
```diff
@@ -120,9 +93,9 @@ class DiffTest extends AsyncFlatSpec {
 
     val excluded =
       """Account.CrmId
-        |crmId1
-        |crmId4
-        |crmId5
+        |B
+        |H
+        |L
```
johnduffell: yep you're right these letters make it a lot easier to know what's going on 😄

johnduffell:
great I think we've disucssed most things and decided most of it isn't urgent enough, thanks for splitting it down 👍 🚀

handlers/zuora-retention/src/main/scala/com/gu/zuora/retention/FilterCandidates.scala
johnduffell: as discussed this will probably be a future enhancement when we do the next one that needs this pattern

handlers/zuora-retention/src/main/scala/com/gu/zuora/retention/Handlers.scala
johnduffell: as discussed this shared trait creates a bit of a cross dependency between different lambdas also, and could be modelled as separate .  the reason it's a trait in this PR is to allow the method that AWS lambda expects, to have the right signature, but still be shared between lambdas.  In the other lambdas it tends to just appear everywhere (together with the wiring code) .  This has the benefit of being able to have different libraryDependencies and a separate JAR for each lambda.
This will probably not need splitting for this lambda, as the different lambdas by defition all need exactly the same dependencies and the code is simple.  But we will keep in mind if we start adding more.

PR #116, opened 2018-06-20
Zuora retention 

pvighi:
Zuora retention state machine. This should execute a couple of AquaQueries and combine the results to get a list of accounts to mark with ProcessingAdvice= 'DoNotProcess'.

statemachine


handlers/zuora-retention/src/main/scala/com/gu/zuora/retention/updateAccounts/UpdateAccounts.scala
```diff
@@ -0,0 +1,34 @@
...
+
+object UpdateAccounts {
+  val oneMinuteMillis = 60000
+
+  def apply(
+    uri: String,
+    setDoNotProcess: String => ClientFailableOp[Unit],
+    getRemainingTimeInMillis: () => Int
+  )(accountIdsIterator: AccountIdIterator) = {
+
```
pvighi: the cod ein this method is not functional at all because the iterator is mutable. I could probably turn this while loop into a tail recursive function it would probably not make much sense to do that

handlers/zuora-retention/README.md
```diff
@@ -1 +1,98 @@
...
+    },
+    {
+      "fileId": "2c92c08663f7f0170164183fc2a237a4",
+      "name": "candidatesQuery",
+      "uri": "s3://zuora-retention-code/2c92c0f863f7f0400164183fbd8f10b6/candidatesQuery.csv"
+    }
+  ],
+  "done": true,
+  "dryRun": true
+}
```
johnduffell: just a thought, would it be possible to put this code in a test and link it from here, so that it's guaranteed to be up to date?

handlers/zuora-retention/src/main/scala/com/gu/zuora/retention/filterCandidates/UploadToS3.scala
```diff
@@ -0,0 +1,29 @@
...
+
+import scala.util.Try
+
+object UploadToS3 {
+
+  def apply(
+    s3Write: PutObjectRequest => Try[PutObjectResult],
+    bucket: String
+  )(filteredCandidates: Iterator[String], key: String) = {
+    val uploadLocation = s"s3://$bucket/$key"
```
johnduffell: since I just did all the zuora query builder, I should point out that we're building a path from strings here.  bucket is hard coded, I guess job id came from somewhere trusted, and maybe s3 doesn't allow ../ or other exploits anyway?  but we should remain vigilant about building things from strings.

handlers/zuora-retention/src/main/scala/com/gu/zuora/retention/updateAccounts/Handler.scala
```diff
@@ -0,0 +1,76 @@
...
+    getRemainingTimeInMsec: () => Int,
+    updateAccounts: (String, SetDoNotProcess, GetRemainingTime) => AccountIdIterator => Try[UpdateAccountsResponse]
+  )(request: UpdateAccountsRequest): Try[UpdateAccountsResponse] = for {
+    config <- toTry(LoadConfig.default[StepsConfig](implicitly)(stage, s3Load(stage)))
+    zuoraRequests = ZuoraRestRequestMaker(response, config.stepsConfig.zuoraRestConfig)
+    linesIterator <- s3Iterator(request.uri)
+    accountIdsIterator <- AccountIdIterator(linesIterator, request.skipTo.getOrElse(0))
+    setDoNotProcess = SetDoNotProcess(zuoraRequests) _
+    wiredUpdateAccounts = UpdateAccounts(request.uri, setDoNotProcess, getRemainingTimeInMsec) _
+    response <- wiredUpdateAccounts(accountIdsIterator)
```
johnduffell: this name is already used in this scope - I think you were previously suggesting we could use getResponse instead  for the Request=>Response

handlers/zuora-retention/src/main/scala/com/gu/zuora/retention/updateAccounts/Handler.scala
```diff
@@ -0,0 +1,76 @@
...
+  case class StepsConfig(zuoraRestConfig: ZuoraRestConfig)
+
+  implicit val stepsConfigReads: Reads[StepsConfig] = Json.reads[StepsConfig]
+  type SetDoNotProcess = String => ClientFailableOp[Unit]
+  type GetRemainingTime = () => Int
+
+  def operation(
+    response: Request => Response,
+    stage: Stage,
+    s3Load: Stage => ConfigFailure \/ String,
```
johnduffell: I'm a bit concerned by the high number of dependencies of this operation function, I think these first 3 could be turned into a single "zuoraRequests" parameter, then deal with the config-getting in another function that turns those three into a zuoraRequests which would be passed in here.  do you think it's worth to split or better to keep a single for comprehension?

handlers/zuora-retention/README.md
```diff
@@ -1 +1,98 @@
...
+  "jobId": "2c92c0f863f7f0400164183fbd8f10b6",
+  "uri": "s3://zuora-retention-code/2c92c0f863f7f0400164183fbd8f10b6/doNoProcess.csv",
+  "dryRun": true
+}
+```
+### Update account  
+The file generated in the previous step is iterated over and each account is updated with ProcessingAdvice= 'DoNotProcess'.
+
+Similarly to the fetch files step, in the actual state machine this is represented by two states in order to accommodate work loads that take longer than the limit on a single lambda execution time:
+* updateAccounts step: this will update as many accounts as possible until it's within a minute of the lambda execution time limit. if it runs out of time it will return done=false and skipTo set to the first line of the file that was not processed
```
johnduffell: is skipTo the standard kind of terminology?  I think the REST pagination tends to work by saying "nextPage" uri or similar, since we are saying done=false should we be saying nextIndex? As skipto doesn't quite make sense in the response, only in the request.  I couldn't find an official document but this document claims next is an "Internet-wide standard" for API pagination... https://knpuniversity.com/screencast/rest-ep2/pagination

handlers/zuora-retention/README.md
```diff
@@ -1 +1,98 @@
...
+* checkRemainingAccounts step: just a choice state to re-run updateAccounts until it returns done = true
```
johnduffell: I think this might be useful for many things if it works well in practice!  A kind of batch queue processing system in S3.

handlers/zuora-retention/src/main/scala/com/gu/zuora/retention/updateAccounts/Handler.scala
```diff
@@ -0,0 +1,76 @@
...
+
+import scala.util.{Failure, Success, Try}
+
+object Handler {
+
+  case class StepsConfig(zuoraRestConfig: ZuoraRestConfig)
+
+  implicit val stepsConfigReads: Reads[StepsConfig] = Json.reads[StepsConfig]
+  type SetDoNotProcess = String => ClientFailableOp[Unit]
+  type GetRemainingTime = () => Int
```
johnduffell: good idea giving these a name.  if you think a naming convention would be useful then feel free.  I'm not sure whether or not (I usually just cmd-f search when in github anyway)

handlers/zuora-retention/src/main/scala/com/gu/zuora/retention/updateAccounts/Handler.scala
```diff
@@ -0,0 +1,76 @@
...
+    response <- wiredUpdateAccounts(accountIdsIterator)
+    _ <- validateProgress(request, response)
+  } yield (response)
+
+  def toTry(res: ConfigFailure \/ Config[StepsConfig]) = res match {
+    case -\/(configError) => Failure(LambdaException(configError.error))
+    case \/-(config) => Success(config)
+  }
+
+  def validateProgress(request: UpdateAccountsRequest, response: UpdateAccountsResponse): Try[Unit] =
```
johnduffell: can we call this something like `failIfNoProgress`?

handlers/zuora-retention/src/main/scala/com/gu/zuora/retention/updateAccounts/UpdateAccounts.scala
```diff
@@ -0,0 +1,34 @@
...
+import com.gu.util.zuora.RestRequestMaker.ClientFailableOp
+
+import scala.util.Try
+
+object UpdateAccounts {
+  val oneMinuteMillis = 60000
+
+  def apply(
+    uri: String,
+    setDoNotProcess: String => ClientFailableOp[Unit],
```
johnduffell: this takes a String, but it could be worth making a case class AccountId and making the AccountIdIterator return one and this take one, just so the types are clearer.  otherwise I thought it might be a uri.

handlers/zuora-retention/src/main/scala/com/gu/zuora/retention/updateAccounts/UpdateAccounts.scala
```diff
@@ -0,0 +1,34 @@
...
+    getRemainingTimeInMillis: () => Int
+  )(accountIdsIterator: AccountIdIterator) = {
```
johnduffell: generally it would be nice to have return types on the main functions if it's easy enough, to make reviews slightly easier

handlers/zuora-retention/src/main/scala/com/gu/zuora/retention/updateAccounts/UpdateAccounts.scala
johnduffell: agree, I never did play with Stream to see if it's possible, however as per my talk, if it's just a local problem and the external api is functional it shouldn't cause a wider issue!
I think using an iterator on a timer is probably the least functional programming task you can do, so what you did is pretty good!

handlers/zuora-retention/src/main/scala/com/gu/zuora/retention/updateAccounts/ws.sc
```diff
@@ -0,0 +1,17 @@
+import scala.util.control.TailCalls.TailRec
```
johnduffell: is this just a temporary file?

handlers/zuora-retention/src/test/scala/com/gu/zuora/retention/filterCandidates/FilterCandidatesEffectsTest.scala
```diff
@@ -0,0 +1,36 @@
...
+    |  ],
+    |"dryRun" : false
+    |  }""".stripMargin
+
+    val testInputStream = new ByteArrayInputStream(input.getBytes)
+    val testOutputStream = new ByteArrayOutputStream()
+
+    FilterCandidates(testInputStream, testOutputStream, null)
+
+    testOutputStream.toString shouldBe "{\"jobId\":\"testJobId\",\"uri\":\"s3://zuora-retention-dev/testJobId/doNoProcess.csv\",\"dryRun\":false}"
```
johnduffell: if it helps I think we have a jsonMatches somewhere in other tests that might help with the layout (it ignores the layout so you can lay out as you please)
failing that, if you triple quote the string it won't backslash all the quotes.

handlers/zuora-retention/src/test/scala/com/gu/zuora/retention/filterCandidates/FilterCandidatesTest.scala
```diff
@@ -0,0 +1,78 @@
...
+  val filteredIt = List("filtered1").iterator
+
+  private def s3Iterator(uri: String) = uri match {
+    case "s3://candidatesUri" => Success(candidatesIt)
+    case "s3://exclusionsUri" => Success(exclusionsIt)
+    case _ => Failure(LambdaException("some error"))
+  }
+
+  private def uploadToS3(filteredCandidates: Iterator[String], key: String) = {
+    if (filteredCandidates == filteredIt)
```
johnduffell: does `==` work correctly on iterators?

handlers/zuora-retention/src/test/scala/com/gu/zuora/retention/filterCandidates/FilterCandidatesTest.scala
```diff
@@ -0,0 +1,78 @@
...
+  val candidatesFetchedFile = FetchedFile("id", "candidatesQuery", "s3://candidatesUri")
+  val exclusionsFetchedFile = FetchedFile("id", "exclusionQuery", "s3://exclusionsUri")
+  val candidatesIt = List("candidate1", "candidate2").iterator
+  val exclusionsIt = List("exclusion1", "exclusion2").iterator
+  val filteredIt = List("filtered1").iterator
+
+  private def s3Iterator(uri: String) = uri match {
+    case "s3://candidatesUri" => Success(candidatesIt)
+    case "s3://exclusionsUri" => Success(exclusionsIt)
+    case _ => Failure(LambdaException("some error"))
```
johnduffell: fair enough, "some error" I guess is "file not found in s3"

handlers/zuora-retention/src/test/scala/com/gu/zuora/retention/filterCandidates/UploadToS3Test.scala
```diff
@@ -0,0 +1,26 @@
...
+
+import scala.util.{Failure, Success, Try}
+
+class UploadToS3Test extends FlatSpec with Matchers {
+
+  def s3Write(req: PutObjectRequest) = {
+
+    val uploadData = scala.io.Source.fromInputStream(req.getInputStream).mkString
+    val dataLength = uploadData.getBytes.length
+    if (req.getBucketName == "testBucket" && uploadData == "some\ndata" && req.getMetadata.getContentLength == dataLength)
```
johnduffell: it's fine, but maybe something we can think about in general:  the only thing that confuses me about these tests is the actual comparison occurs not in the test definition.  I always think of the test is composed of expected result, input to the function and output, and the comparison code, so it can be confusing when some of the comparison and expected is mixed with the input to the function.

handlers/zuora-retention/src/test/scala/com/gu/zuora/retention/updateAccounts/HandlerTest.scala
```diff
@@ -0,0 +1,29 @@
...
+    val noProgressResponse = UpdateAccountsResponse(uri = "someUri", skipTo = None, done = false)
+    val noProgressError = Failure(LambdaException("no accounts processed in execution!"))
+
+    Handler.validateProgress(request, noProgressResponse) shouldBe (noProgressError)
+  }
+
+  it should "validate when all accounts processed" in {
+    val request = UpdateAccountsRequest(uri = "someUri", skipTo = None)
+    val noProgressResponse = UpdateAccountsResponse(uri = "someUri", skipTo = None, done = true)
+    Handler.validateProgress(request, noProgressResponse) shouldBe (Success(()))
```
johnduffell: this says it's the HandlerTest but actually it's testing validateProgress.  Consider splitting that to a separate file and calling the test by that too.

lib/handler/src/main/scala/com/gu/util/handlers/JsonHandler.scala
```diff
@@ -0,0 +1,31 @@
...
+import play.api.libs.json.{Reads, Writes}
+
+import scala.util.{Failure, Success, Try}
+
+object JsonHandler extends Logging {
+
+  def apply[REQUEST, RESPONSE](
+    lambdaIO: LambdaIO,
+    operation: REQUEST => Try[RESPONSE]
+  )(implicit r: Reads[REQUEST], w: Writes[RESPONSE]): Unit = {
```
johnduffell: nice, Tom and I were trying to think about when longer type parameter names would work and I think this is a great example.

lib/handler/src/main/scala/com/gu/util/handlers/JsonHandler.scala
```diff
@@ -0,0 +1,31 @@
...
+      case Failure(exception) => {
+        logger.error("terminating lambda with error ", exception)
+        throw exception
+      }
+      case Success(successResponse) => SerialiseResponse(lambdaIO.outputStream, successResponse)
+    }
+
+  }
+
+}
```
johnduffell: good yes I think this function is a really neat way to extract this 💯

lib/zuora-reports/README.md
```diff
@@ -58,36 +57,56 @@ Here we get the results of the job submitted in the previous example and get a c
...
+  "done": false,
+  "dryRun": false
+}
 }
-```
\ No newline at end of file
+```
+
+
+##dryRun parameter
```
johnduffell: not suggesting to change, but I always slightly regret calling it "dryRun" in the othe rlambda because it feels like a double negative when I say dryRun = false!

lib/zuora-reports/src/main/scala/com/gu/zuora/reports/ReportHandlers.scala
```diff
@@ -11,10 +11,9 @@ import com.gu.zuora.reports.ReportsLambda.{AquaCall, StepsConfig}
 import com.gu.zuora.reports.aqua.{AquaQueryRequest, ZuoraAquaRequestMaker}
 import play.api.libs.json.Reads
 
-trait ReportHandlers[QUERY_REQUEST] {
+trait ReportHandlers[QUERY_REQUEST <: QuerierRequest] {
```
johnduffell: I don't have a suggestion, but I'm still concerned about a trait causing a focal point for making all the lambdas depend on each other whether they like it or not.  I think I will have to encounter this situation myself before I can make a proper opinion on it.  This comes to the same question we were discussing yesterday with Tom.
I agree that splitting them would cause a lot of unnecessary boiler plate, which seems expensive when they are so closely related and so tightly coupled in behaviour anyway.

johnduffell:
looks great, I've make a load of comments but they're 90% nit picks or general observations.  I think it's great work on the testing as they are mostly clear. 👍and codacy is happy re coverage I probably won't be around tomorrow to discuss but hopefully you will get the gist - don't let it slow you down 🏎

handlers/zuora-retention/src/main/scala/com/gu/zuora/retention/filterCandidates/UploadToS3.scala
pvighi: I could change this to bucket and key instead of returning a uri because i think all the steps that come after this will parse the it to get the bucket and key anyway.

handlers/zuora-retention/src/main/scala/com/gu/zuora/retention/updateAccounts/UpdateAccounts.scala
pvighi: sorry it's just the new intellij that shows me the types anyway so sometimes I forget to actually write them in the code

handlers/zuora-retention/src/test/scala/com/gu/zuora/retention/filterCandidates/FilterCandidatesTest.scala
pvighi: well depends on the definition of working correctly. I just wanted to make sure it's the exact same instance and it seems to work (tested it by replacing one of the iterators) .
I didn't do any research though

handlers/zuora-retention/src/test/scala/com/gu/zuora/retention/filterCandidates/UploadToS3Test.scala
pvighi: Yeah I'm not 100% sure I should put this in here. I could just be happy to check the final result and have all the fake dependencies be as dumb as possible.
I just went for the extra check that dependencies are called with the right params.  Some people might say that is an internal detail and we should not look at that in the unit test. You can also think of it as having more realistic fakes to make sure the response you are asserting is more realistic as well.

In this case I could just look at the output of the object under test but in other cases  where interact with code with side effects checking that something was called with the right params is all we can do.

@jacobwinch what do you think ?

handlers/zuora-retention/src/main/scala/com/gu/zuora/retention/query/RetentionQueryRequest.scala
```diff
@@ -16,7 +17,9 @@ object ToAquaRequest {
   val exclusionQueryName = "exclusionQuery"
   val candidatesQueryName = "candidatesQuery"
   def apply(request: RetentionQueryRequest): AquaQueryRequest = {
-    val dateStr = request.cutOffDate.format(DateTimeFormatter.ISO_LOCAL_DATE)
+
+    val filterDate = request.cutOffDate getOrElse LocalDate.now.minusMonths(30)
```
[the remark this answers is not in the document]
pvighi: yeah.. it's all tested passing an explicit date. I just added the default value because this will end up being triggered from a cloudwatch rule on a schedule and I don't think you can (or at least not easily) generate the correct date from the rule itself. Anyway I'll see if I can find another way to this, maybe involving passing a now dependency somewhere

handlers/zuora-retention/README.md
```diff
@@ -1 +1,98 @@
...
+```
+{
+    "cutOffDate": "YYYY-MM-DD",
+    "dryRun" : true
+}
+```
+
+The optional cutOffDate param determines the latest date a subscription can be cancelled and still be considered 'Old'. If this parameter is omitted it defaults to 30 months ago.
+
+If the dryRun param is set to true then the account updating step will be skipped. This is used for debugging by inspecting the doNotProcess.csv file to see what accounts would be updated if this wasn't a dry run exectuion.
```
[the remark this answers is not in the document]
pvighi: most of these requests and responses are just simple case classes with regular json reads/writes. 
Maybe the json inputs and outputs are not really necessary here as this is not an api but a state machine so probably the only json format most people would need to know is the initial json input to the first lambda

lib/zuora-reports/src/main/scala/com/gu/zuora/reports/ReportHandlers.scala
pvighi: The trait only exists to avoid duplication with other future users of zuora-reports. Given that the only place this is used is zuora-retention I could get rid of it and put all of this in an object in zuora retention.
 I can also split this into 3 small objects to wire each lambda individually since we are not really avoiding any code duplication by putting them all in the same trait after the last few refactors.

handlers/zuora-retention/src/test/scala/com/gu/zuora/retention/filterCandidates/UploadToS3Test.scala
jacobwinch: I'm not hugely opinionated about this to be honest. In a simple case like this one I find it easy enough to understand. 

I'd probably make the message of the exception slightly more explicit so that it's as easy as possible for someone to debug this (i.e. it'd be nice to know what had broken from the sbt test output, without having to dig into the mock)

handlers/zuora-retention/src/main/scala/com/gu/zuora/retention/ZuoraRetentionFetchFileHandler.scala
```diff
@@ -0,0 +1,14 @@
+package com.gu.zuora.retention
+
+import java.io.{InputStream, OutputStream}
+
+import com.amazonaws.services.lambda.runtime.Context
+import com.gu.zuora.reports.handlers.FetchFileHandler
+
+object ZuoraRetentionFetchFileHandler {
+  // this is the entry point
+  // it's referenced by the cloudformation so make sure you keep it in step
```
pvighi: my first attempt was to do something like this: 
```scala
def apply = QueryHandler[RetentionQueryRequest](
    reportsBucketPrefix = "zuora-retention",
    toQueryRequest = ToAquaRequest(RawEffects.now().toLocalDate),
    queryReads = RetentionQueryRequest.reads
  )
```
but aws would not find the handler so unless I was making a mistake it looks like its not supported.
So in the end I have to define it in the old plain way passing the parameters and everything

handlers/zuora-retention/src/main/scala/com/gu/zuora/retention/filterCandidates/UploadToS3.scala
johnduffell: fair enough, if that is the case then serialising and deserialising to a single string would be an unnecessary risk

handlers/zuora-retention/src/test/scala/com/gu/zuora/retention/filterCandidates/FilterCandidatesTest.scala
johnduffell: fair point and I agree identity would be correct in this case and equality wouldn't be possible.  I definitly feel iterators just made me remember all my old java knowledge!

handlers/zuora-retention/src/test/scala/com/gu/zuora/retention/filterCandidates/UploadToS3Test.scala
johnduffell: I wasn't meaning we shouldn't do the check, more that the test made fundamentally slightly more complicated by the intermediate call.  I don't think that's because how you did it, we have exactly the same issue with testing the http calls too.  I am wondering if there is a general pattern to do this without resorting to passing in a monad or something to store the result and check later.  I will keep it in my mind!

handlers/zuora-retention/src/main/scala/com/gu/zuora/retention/ZuoraRetentionFetchFileHandler.scala
johnduffell: agree the handler has to appear as a method when it gets to java, rather than a scala Function3, although scala mostly hides the distinction from you, it can fall through the gaps

handlers/zuora-retention/src/main/scala/com/gu/zuora/retention/updateAccounts/Handler.scala
```diff
@@ -27,21 +28,21 @@ object Handler {
...
+    s3Load: Stage => ConfigFailure \/ String
+  ): Try[Requests] = for {
+    config <- toTry(LoadConfig.default[StepsConfig](implicitly)(stage, s3Load(stage)))
+  } yield ZuoraRestRequestMaker(response, config.stepsConfig.zuoraRestConfig)
+
+  def operation(
     s3Iterator: String => Try[Iterator[String]],
-    getRemainingTimeInMsec: () => Int,
-    updateAccounts: (String, SetDoNotProcess, GetRemainingTime) => AccountIdIterator => Try[UpdateAccountsResponse]
+    updateAccounts: (String, AccountIdIterator) => Try[UpdateAccountsResponse]
```
johnduffell: much better the complexity of the function signature 👍

handlers/zuora-retention/src/main/scala/com/gu/zuora/retention/updateAccounts/SetDoNotProcess.scala
```diff
@@ -1,19 +1,20 @@
...
   case class UpdateRequestBody(ProcessingAdvice__c: String = "DoNotProcess")
 
-  implicit val updateRequestBodyWrites = Json.writes[UpdateRequestBody]
-  implicit val unitReads: Reads[Unit] = Reads(_ => JsSuccess(()))
+  object UpdateRequestBody {
+    implicit val updateRequestBodyWrites = Json.writes[UpdateRequestBody]
+  }
 
-  def apply(zuoraRequests: Requests)(accountId: String) = {
+  def apply(put: (UpdateRequestBody, String) => ClientFailableOp[Unit])(accountId: AccountId) = {
```
johnduffell: agree about this being a bit funny that it gets the implicits during the wiring, but I think the whole "Requests" class is not quite how it should be. I will think of a solution for this eventually as it's very hard to pass in mock ones.

handlers/zuora-retention/src/test/scala/com/gu/zuora/retention/updateAccounts/AccountIdIteratorTest.scala
```diff
@@ -0,0 +1,75 @@
...
+  it should "return error if no account id column in header" in {
+    val noAccountIdData =
+      """col1, col2, col3
+        |data1,data2,data3
+      """.stripMargin
+
+    AccountIdIterator(linesIterator(noAccountIdData), 0) shouldBe errorResponse("No Account.Id column found in csv file header")
+  }
+
+  it should "return error skip to the correct starting position" in {
```
johnduffell: what does return error mean?

johnduffell:
great work 👍 thanks
I checked and most of the codacy warnings are lines >140 and functions >35 lines, which are just nice to have not essentials!  go for it 🚀

PR #117, opened 2018-06-22
Zuora retention fixes

pvighi:
I after testing in PROD I realised that I forgot to add a retry to wait until the queries in are done in Zuora.
I also added a retries parameter to getJobResults to keep track of how many times it has been run and make the whole execution fail if it retries too many times.
I can't really test this in CODE as the query always returns instantly so I will have to run it in PROD and see if it works.
updated state machine:
statemachine

### DIFF
--- a/lib/zuora-reports/src/test/scala/com/gu/zuora/reports/GetJobResultTest.scala
+++ b/lib/zuora-reports/src/test/scala/com/gu/zuora/reports/GetJobResultTest.scala
@@ -27,37 +27,32 @@ class GetJobResultTest extends AsyncFlatSpec {
     )
   ))
   it should "decrease the amount of retries left on each execution" in {
-    def get(path :String) = ZuoraResponseWithStatus("pending")
-    val jobResultRequest = JobResultRequest(jobId= "someJobId", dryRun = false, retries = Some(7))
+    def get(path: String) = ZuoraResponseWithStatus("pending")
+    val jobResultRequest = JobResultRequest(jobId = "someJobId", dryRun = false, tries = Some(7))
     GetJobResult(get)(jobResultRequest) shouldBe \/-(Pending("testResponse", "someJobId", false, 6))
   }
 
-  it should "return -1 retires when called with retries = 0" in {
-    def get(path :String) = ZuoraResponseWithStatus("pending")
-    val jobResultRequest = JobResultRequest(jobId= "someJobId", dryRun = false, retries = Some(0))
-    GetJobResult(get)(jobResultRequest) shouldBe \/-(Pending("testResponse", "someJobId", false, -1))
-  }
-
-  it should "return error if called with a negative number of retries" in {
-    def get(path :String) = ZuoraResponseWithStatus("pending")
-    val jobResultRequest = JobResultRequest(jobId= "someJobId", dryRun = false, retries = Some(-1))
-    GetJobResult(get)(jobResultRequest) shouldBe -\/(GenericError("too many retries!"))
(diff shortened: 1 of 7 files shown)


lib/zuora-reports/src/test/scala/com/gu/zuora/reports/GetJobResultTest.scala
```diff
@@ -26,15 +26,38 @@ class GetJobResultTest extends AsyncFlatSpec {
...
+
+  it should "return -1 retires when called with retries = 0" in {
+    def get(path :String) = ZuoraResponseWithStatus("pending")
+    val jobResultRequest = JobResultRequest(jobId= "someJobId", dryRun = false, retries = Some(0))
+    GetJobResult(get)(jobResultRequest) shouldBe \/-(Pending("testResponse", "someJobId", false, -1))
+  }
+
+  it should "return error if called with a negative number of retries" in {
+    def get(path :String) = ZuoraResponseWithStatus("pending")
+    val jobResultRequest = JobResultRequest(jobId= "someJobId", dryRun = false, retries = Some(-1))
```
johnduffell: can we call it tries instead of retries, then we can just say 3 tries and when we're called with zero we throw up, otherwise I feel like my brain has an off by one error if it's 10 retries does that mean 11 tries in total or 10 in total?

lib/zuora-reports/src/test/scala/com/gu/zuora/reports/GetJobResultTest.scala
```diff
@@ -26,15 +26,38 @@ class GetJobResultTest extends AsyncFlatSpec {
...
-    GetJobResult.toJobResultResponse(ZuoraResponseWithStatus("executing"), true, "someJobId") shouldBe \/-(Pending("testResponse", "someJobId", true))
+    def get(path :String) = ZuoraResponseWithStatus("executing")
+    val jobResultRequest = JobResultRequest("someJobId", true, None)
+    GetJobResult(get)(jobResultRequest) shouldBe \/-(Pending("testResponse", "someJobId", true, 9))
   }
   it should "return error if zuora response status is an unexpected value " in {
-    GetJobResult.toJobResultResponse(ZuoraResponseWithStatus("aborted"), false, "someJobId") shouldBe -\/(GenericError("unexpected status in zuora response: AquaJobResponse(aborted,testResponse,List(Batch(completed,batch1,Some(fileId1)), Batch(completed,batch2,Some(fileId2))),None)"))
+    def get(path :String) = ZuoraResponseWithStatus("aborted")
+    val jobResultRequest = JobResultRequest("someJobId", false, None)
+    GetJobResult(get)(jobResultRequest) shouldBe -\/(GenericError("unexpected status in zuora response: AquaJobResponse(aborted,testResponse,List(Batch(completed,batch1,Some(fileId1)), Batch(completed,batch2,Some(fileId2))),None)"))
```
johnduffell: agree the passing through of all the fields is obscuring the real intent quite a lot.  do yo u think we could do something like MyRequest => Try[ResultAugmenter] so we only pull out what we need, but the JsonHelper remembers the whole request, and uses the result augmenter to add or amend the request parameters?
in a future PR I mean, not in this one.  I think that would help, but it depends whether the ResultAugmenter can be simple enough.  Not sure we can get away with just doing the union of the existing and new, or whether we need a way to delete fields we don't want to carry through.

johnduffell:
great, just a small comment about (re)tries name, looks great 💯 
also you might want to run the compile over it again to make sure the formatting, as I think one of the tests got the colons funny.
👍

PR #118, opened 2018-06-25
exclude accounts with no crmId

pvighi:
There is one account in the candidates results in prod that has no crmId. This crashes the filter when trying to find accounts in the exclusion list with a matching crmId.

This means that accounts with no crmId will never be marked doNotProcess by this. @paulbrown1982 do we need to support this case or is it expected that all accounts should have a crmId?


pvighi:
we should fix this in another way as we should not exclude accounts with no crmId

PR #119, opened 2018-06-25
include candidates with no crmId in results

pvighi:
normally the way this works is by finding accounts with old subs and then excluding from the results the accounts with a crmId that is also present in active or "not old enough" accounts.
If a candidate account has no crmId then it is included in the results as there is nothing to look for in the exclusion list.

### DIFF
--- a/handlers/zuora-retention/src/main/scala/com/gu/zuora/retention/Diff.scala
+++ b/handlers/zuora-retention/src/main/scala/com/gu/zuora/retention/Diff.scala
@@ -5,10 +5,10 @@ object Diff {
   val crmIdColName = "Account.CrmId"
 
   /**
-   * Returns an iterator for the lines in candidateLines with crmIds that are not in the exclusionLines
-   * The candidates and exclusion iterators are expected to iterate ascending CrmId order.
-   * The point of this is to avoid loading the whole exclusionLines in memory.
-   */
+    * Returns an iterator for the lines in candidateLines with crmIds that are not in the exclusionLines
+    * The candidates and exclusion iterators are expected to iterate ascending CrmId order.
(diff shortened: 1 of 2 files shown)


handlers/zuora-retention/src/main/scala/com/gu/zuora/retention/Diff.scala
```diff
@@ -5,25 +5,29 @@ object Diff {
...
     exclusionLines.next //skip header
     val exclusionCrmIds = SortedCrmIdIterator(exclusionLines)
     val candidatesHeader = candidateLines.next()
     val crmidLocation = candidatesHeader.split(",").indexOf(crmIdColName)
     val valueRows = candidateLines.filterNot { line =>
       line.trim.isEmpty || {
-        val crmId = line.split(",")(crmidLocation).trim
-        val comp = exclusionCrmIds.nextGreaterOrEqual(crmId)
-        comp.exists(_.toLowerCase == crmId.toLowerCase)
+        val crmId = line.split(",", -1)(crmidLocation).trim
```
pvighi: split with by default exclude empty trailing strings so
```
 "1".split(",")  == List(1) 
```
the second parameter (limit) makes it so : 
```
 "1".split(",", -1)  == List(1, ) 
```
for more information see the [String javadoc](https://docs.oracle.com/javase/7/docs/api/java/lang/String.html#split(java.lang.String,%20int)) ( i still don't fully understand all the uses of this second param but it seems to wok)

johnduffell: makes sense, the documentation about split is atrocious

johnduffell:
yes good spot 👍

PR #120, opened 2018-06-25
exclude subscriptions with weird end dates from retention query results

pvighi:
there's some subscriptions in PROD with end dates like "0018-02-05" this PR excludes those from the results


johnduffell:
👍 🔢

PR #121, opened 2018-06-25
exclude accounts with no crmId from zuora retention results

pvighi:
we cannot skip the filtering step for subscriptions from Accounts with no crmId.
There might be accounts with multiple subscriptions only one of which is 'old', in this case we would not want to mark the account 'do_not_process'
Since there's only 14 accounts with no crmId we will just leave them alone as it is not worth it to add accountId based filtering to this code


johnduffell:
👍

PR #122, opened 2018-06-26
trigger zuora retention state machine every day at 3:00 GMT

pvighi:
The time was pretty much selected at random (trying to not overlap with the fulfilment queries) @paulbrown1982  do you think we should trigger this at a different time of the day / with a different frequency?
Another thing we could change is keep the Zuora Query results around for longer. Right now they are deleted automatically after one day, but we could use them in case we want to see which accounts were updated or undo the changes in postman.

### DIFF
--- a/handlers/zuora-retention/cfn.yaml
+++ b/handlers/zuora-retention/cfn.yaml
@@ -482,8 +482,8 @@ Resources:
     ScheduleRule:
         Type: "AWS::Events::Rule"
         Properties:
-          Description: "triggerZuoraRetention"
-          ScheduleExpression: "cron(00 3 ? * * *)"
+          Description: "Trigger the Zuora Retention state machine every day at 03:00 GMT"
+          ScheduleExpression: "cron(0 3 ? * * *)"
           State: "ENABLED"
           Targets:
             -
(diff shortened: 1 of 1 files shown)


handlers/zuora-retention/cfn.yaml
```diff
@@ -459,3 +479,18 @@ Resources:
             filterArn: !GetAtt [ ZuoraRetentionFilter, Arn ],
             accountUpdaterArn: !GetAtt [ ZuoraRetentionAccountUpdater, Arn ]
           }
+    ScheduleRule:
+        Type: "AWS::Events::Rule"
+        Properties:
+          Description: "triggerZuoraRetention"
+          ScheduleExpression: "cron(00 3 ? * * *)"
```
paulbrown1982: Maybe use a single-zero? and comment what the cron means in English.

pvighi: I'll change it we do have others with double zero, like the weekly fulfilment one:
https://github.com/guardian/fulfilment-lambdas/blob/master/cloudformation/cloudformation.yaml#L459

johnduffell:
agree with paul, cron expressions are one of my heavily disliked things (along with regular expressions etc) as they are prone to error and impossible to understand in place.

PR #123, opened 2018-06-26
new lambda(s) for SalesForce Case creation/update for cancellation flow

twrichards:
Created two new lambda functions (under a single handler) [via API Gateway]...

- `cancellation-sf-cases-raise-` POST
- `cancellation-sf-cases-update-` PATCH

... for creating and updating SalesForce 'Cases'. Called by the new cancellation flow as per https://docs.google.com/document/d/1FcIj8-gY4cNoo7hjESBaZIX4CVU1Vw1A10iCKXf7m1o/edit?usp=sharing

**NOTES**
- There is some urgency to ship (as blocking release of new cancellation flow this quarter), so quite a few things have been added to the [Lambdas Tech Debt Document](https://docs.google.com/document/d/1A4IyEWhABoGUw01fiuS0JtuuYn2IfQFZdTA5mtrcNho/edit?usp=sharing), _including Tests_ but most points will be addressed imminently
- DISABLED api key based authorisation, but does use identity cookie to resolve to identity id which is then used for the linked Contact in the resulting SF case.


handlers/cancellation-sf-cases/README.md
```diff
@@ -0,0 +1,5 @@
+# cancellation-sf-cases
+Create/update SalesForce cases for self service cancellation tracking
+
+## Handling Multiple Environments 
+todo
```
johnduffell: yeop would be nice to fill this file out a bit more in general before it's forgotten, even rough notes

handlers/cancellation-sf-cases/build.sbt
```diff
@@ -0,0 +1,29 @@
...
+  "com.squareup.okhttp3" % "okhttp" % "3.9.1",
+  "com.typesafe.play" %% "play-json" % "2.6.9",
+  "org.scalatest" %% "scalatest" % "3.0.1" % "test",
+  "com.fasterxml.jackson.core" % "jackson-databind" % "2.8.11.1",
+  "com.gu.identity" %% "identity-cookie" % "3.160"
+)
+
+resolvers ++= Seq(
+  Resolver.sonatypeRepo("releases")
+)
```
johnduffell: why was the resolver needed? needed for the identity-cookie library?

handlers/cancellation-sf-cases/build.sbt
```diff
@@ -0,0 +1,29 @@
...
+riffRaffPackageType := assembly.value
+riffRaffUploadArtifactBucket := Option("riffraff-artifact")
+riffRaffUploadManifestBucket := Option("riffraff-builds")
+riffRaffManifestProjectName := "MemSub::Subscriptions::Lambdas::Cancellation SF Cases"
+riffRaffArtifactResources += (file("handlers/cancellation-sf-cases/cfn.yaml"), "cfn/cfn.yaml")
+
+libraryDependencies ++= Seq(
+  "com.amazonaws" % "aws-lambda-java-log4j" % "1.0.0",
+  "com.amazonaws" % "aws-lambda-java-core" % "1.2.0",
+  "com.amazonaws" % "aws-java-sdk-s3" % "1.11.311",
```
johnduffell: some of these are in Dependencies.scala so we can reference from there (ideally most of the ones that are used in many lambdas should be referenced from there)

handlers/cancellation-sf-cases/cfn.yaml
```diff
@@ -0,0 +1,278 @@
...
+    CancellationSFRaiseCaseLambda:
+        Type: AWS::Lambda::Function
+        Properties:
+            Description: account-frontend used to create SalesForce cases for self service cancellation tracking
+            FunctionName:
+                !Sub cancellation-sf-cases-raise-${Stage}
+            Code:
+                S3Bucket: zuora-auto-cancel-dist
+                S3Key: !Sub subscriptions/${Stage}/cancellation-sf-cases/cancellation-sf-cases.jar
+            Handler: com.gu.cancellation.sf_cases.Handler::raiseCase
```
johnduffell: just an observation (no point to change): the official way is indeed underscore, but I noticed in the past we usually use camelCase for some reason.

handlers/cancellation-sf-cases/cfn.yaml
```diff
@@ -0,0 +1,278 @@
...
+            RestApiId: !Ref CancellationSFCasesAPI
+            ParentId: !GetAtt [CancellationSFCasesAPI, RootResourceId]
+            PathPart: case
+        DependsOn: CancellationSFCasesAPI
+
+    CancellationSFRaiseCaseMethod:
+        Type: AWS::ApiGateway::Method
+        Properties:
+            AuthorizationType: NONE
+            ApiKeyRequired: false
```
johnduffell: do we want to add a default rate limit for the "no api key" user (now or in a later PR)? to avoid having to serve a high rate of requests due to an error or attack?

handlers/cancellation-sf-cases/src/main/scala/com/gu/cancellation/sf_cases/Handler.scala
```diff
@@ -0,0 +1,112 @@
...
+    case class RaiseRequestBody(
+      product: String,
+      reason: String,
+      subscriptionName: String
+    )
+    implicit val reads = Json.reads[RaiseRequestBody]
+
+    implicit val writes = Json.writes[RaiseCaseResponse]
+
+    def embellishRaiseRequestBody(raiseRequestBody: RaiseRequestBody, sfSubscriptionIdContainer: ResponseWithId, sfContactId: ResponseWithId) = NewCase(
```
johnduffell: may be ok in intellij but if it's a bit wide consider writing this more like
```scala
def embellishRaiseRequestBody(
  raiseRequestBody: RaiseRequestBody,
  sfSubscriptionIdContainer: ResponseWithId,
  sfContactId: ResponseWithId) =
    NewCase(
```
codacy is set to mark a new issue if it's over 140 characters and it helps in github UI, although it's not strict and I know @pvighi prefers width ways more than length ways, so make your own opinion!

johnduffell: also, as a separate point, would there be a more precise word than "embellish" to describe that this is doing?

handlers/cancellation-sf-cases/cfn.yaml
twrichards: most definitely - its already in the tech debt doc

handlers/cancellation-sf-cases/src/main/scala/com/gu/cancellation/sf_cases/Handler.scala
```diff
@@ -0,0 +1,112 @@
...
+    )
+
+    def steps(sfRequests: LazySalesforceAuthenticatedReqMaker)(apiGatewayRequest: ApiGatewayRequest) =
+      (for {
+        identityId <- IdentityCookieToIdentityId(apiGatewayRequest.headers)
+        sfRequests <- sfRequests()
+        raiseCaseDetail <- apiGatewayRequest.bodyAsCaseClass[RaiseRequestBody]()
+        sfContactId <- SalesforceGenericIdLookup(sfRequests)("Contact", "IdentityID__c", identityId).toApiGatewayOp("lookup SF contact from identityID")
+        sfSubscriptionIdContainer <- SalesforceGenericIdLookup(sfRequests)("SF_Subscription__c", "Name", raiseCaseDetail.subscriptionName).toApiGatewayOp("lookup SF subscription ID")
+        raiseCaseResponse <- SalesforceCase.Raise(sfRequests)(embellishRaiseRequestBody(raiseCaseDetail, sfSubscriptionIdContainer, sfContactId)).toApiGatewayOp("raise sf case")
```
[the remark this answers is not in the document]
twrichards: i will look to do this when i add tests (as mentioned in the tech debt doc)

handlers/cancellation-sf-cases/src/main/scala/com/gu/identity/IdentityCookieToIdentityId.scala
```diff
@@ -0,0 +1,45 @@
...
+
+  def apply(headersOption: Option[Map[String, String]]): ApiGatewayOp[String] = headersOption match {
+    case Some(headers) => processHeaders(headers)
+    case None => {
+      logger.debug("no headers")
+      badRequest
+    }
+  }
+
+  private def processHeaders(headers: Map[String, String]): ApiGatewayOp[String] =
```
johnduffell: we could call this `getIdentityIdFromHeaders` I think?

handlers/cancellation-sf-cases/src/main/scala/com/gu/identity/IdentityCookieToIdentityId.scala
```diff
@@ -0,0 +1,45 @@
...
+    headers get "Cookie" match {
+      case Some(cookie) => processCookie(cookie, new IdentityCookieDecoder(new PreProductionKeys)) // TODO review keys selection
+      case None => {
+        logger.debug("no cookie")
+        badRequest
+      }
+    }
+
+  private def processCookie(cookieHeaderValue: String, cookieDecoder: IdentityCookieDecoder) =
```
johnduffell: avoid using the word process in general, try to be specific
`getIdentityIdFromCookie` ?

handlers/cancellation-sf-cases/src/main/scala/com/gu/cancellation/sf_cases/Handler.scala
```diff
@@ -0,0 +1,132 @@
...
+
+    case class UpdateCasePathParams(caseId: String)
+    implicit val pathParamReads = Json.reads[UpdateCasePathParams]
+
+    def steps(sfRequests: LazySalesforceAuthenticatedReqMaker)(apiGatewayRequest: ApiGatewayRequest) =
+      (for {
+        _ <- IdentityCookieToIdentityId(apiGatewayRequest.headers) // TODO verify case belongs to identity user
+        sfRequests <- sfRequests()
+        pathParams <- apiGatewayRequest.pathParamsAsCaseClass[UpdateCasePathParams]()
+        requestBody <- apiGatewayRequest.bodyAsCaseClass[JsValue]()
```
johnduffell: so we are directly proxying the json passed in without worrying about the exact fields, I'm not sure if we should be completely comfortable with that, or if we should white list the fields we care about (by going via a case class rather than JsValue)?

lib/handler/src/main/scala/com/gu/util/apigateway/ApiGatewayRequest.scala
```diff
@@ -34,7 +34,8 @@ case class URLParams(
 /* Using query strings because for Basic Auth to work Zuora requires us to return a WWW-Authenticate
   header, and API Gateway does not support this header (returns x-amzn-Remapped-WWW-Authenticate instead)
   */
-case class ApiGatewayRequest(queryStringParameters: Option[URLParams], body: Option[String], headers: Option[Map[String, String]]) {
+case class ApiGatewayRequest(queryStringParameters: Option[URLParams], body: Option[String],
+  headers: Option[Map[String, String]], pathParameters: Option[JsValue] = None) {
```
johnduffell: if you're reformatting params onto more than one line, I'd prefer to completely switch to the one-per-line format just for consistency

lib/handler/src/main/scala/com/gu/util/apigateway/ApiGatewayRequest.scala
```diff
@@ -59,6 +60,12 @@ case class ApiGatewayRequest(queryStringParameters: Option[URLParams], body: Opt
     }
   }
 
+  def pathParamsAsCaseClass[T](failureResponse: ApiResponse = ApiGatewayResponse.badRequest)(implicit reads: Reads[T]): ApiGatewayOp[T] =
+    pathParameters match {
+      case Some(pathParamsJSON) => Json.fromJson[T](pathParamsJSON).toApiGatewayOp(failureResponse)
+      case None => ReturnWithResponse(failureResponse)
+    }
+
```
johnduffell: this is good to have a deserialiser, @pvighi is doing a similar thing for the UrlParams at the moment, and I notice we don't have the qeuivalent for Headers.  the thing we just refactored for getting the cookie header and then reading the ideneitty id, the first lines of the for comprehension are purely just getting hold of the actual value of the header and confirming it is there, perhaps that should also be in here so that everyone who uses headers doesn't have to deal with an Option[Map[...]] can just use a case class instead?

johnduffell:
yep looks good thanks, a couple of comments above that could go into the phase 2 cleanup PR as you wish, I don't see any reason not to go ahead and ship otherwise 👍 🚀

handlers/cancellation-sf-cases/src/main/scala/com/gu/cancellation/sf_cases/Handler.scala
twrichards: fair - added to tech debt

lib/handler/src/main/scala/com/gu/util/apigateway/ApiGatewayRequest.scala
twrichards: good idea - added to tech debt doc

PR #87, opened 2018-05-03
Fix warnings

johnduffell:
https://github.com/guardian/zuora-auto-cancel/pull/85#discussion_r185253279
@jacobwinch 
> A general note: we have started deprecating quite a few methods in this repo and we should make some time to tidy them up soon...

(this also includes the version bumps that are in the previous PR for SBT and scala, because I based this branch on that one, but I will merge separately)

@pvighi @paulbrown1982 @twrichards

PR #124, opened 2018-06-27
Typo on #87

kant:


johnduffell:
thanks but it's probably not worth shipping for no substantial change

PR #125, opened 2018-06-27
Refactor Url Params

pvighi:
All tests are passing but I still need to test in code to verify that all lambdas still work.

### DIFF
--- a/handlers/digital-subscription-expiry/src/main/scala/com/gu/digitalSubscriptionExpiry/DigitalSubscriptionExpirySteps.scala
+++ b/handlers/digital-subscription-expiry/src/main/scala/com/gu/digitalSubscriptionExpiry/DigitalSubscriptionExpirySteps.scala
@@ -10,20 +10,16 @@ import com.gu.util.reader.Types._
 import main.scala.com.gu.digitalSubscriptionExpiry.DigitalSubscriptionExpiryRequest
 import ApiGatewayOp.ContinueProcessing
 import com.gu.util.apigateway.ResponseModels.ApiResponse
-import play.api.libs.json.{JsResult, JsValue, Json, Reads}
+import play.api.libs.json.{Json, Reads}
 
 case class UrlParams(noActivation: Boolean)
 object UrlParams {
 
-  case class UrlParamsWire(noActivation: String = "false") {
-    def toUrlParams = UrlParams(noActivation == "true")
+  case class UrlParamsWire(noActivation: Option[String]) {
+    def toUrlParams = UrlParams(noActivation.contains("true"))
   }
-
-  implicit val wireReads = Json.using[Json.WithDefaultValues].reads[UrlParamsWire]
-
-  implicit val urlParamsReads = new Reads[UrlParams] {
-    override def reads(json: JsValue): JsResult[UrlParams] = wireReads.reads(json).map(_.toUrlParams)
-  }
+  val wireReads = Json.reads[UrlParamsWire]
+  implicit val urlParamsReads: Reads[UrlParams] = json => wireReads.reads(json).map(_.toUrlParams)
 }
 object DigitalSubscriptionExpirySteps extends Logging {
 
--- a/handlers/digital-subscription-expiry/src/main/scala/com/gu/digitalSubscriptionExpiry/DigitalSubscriptionExpirySteps.scala
+++ b/handlers/digital-subscription-expiry/src/main/scala/com/gu/digitalSubscriptionExpiry/DigitalSubscriptionExpirySteps.scala
@@ -5,12 +5,26 @@ import com.gu.digitalSubscriptionExpiry.zuora.GetAccountSummary.{AccountId, Acco
 import com.gu.digitalSubscriptionExpiry.zuora.GetSubscription.{SubscriptionId, SubscriptionResult}
 import com.gu.util.Logging
 import com.gu.util.apigateway.ApiGatewayHandler.Operation
-import com.gu.util.apigateway.{ApiGatewayRequest, URLParams}
+import com.gu.util.apigateway.ApiGatewayRequest
 import com.gu.util.reader.Types._
 import main.scala.com.gu.digitalSubscriptionExpiry.DigitalSubscriptionExpiryRequest
 import ApiGatewayOp.ContinueProcessing
 import com.gu.util.apigateway.ResponseModels.ApiResponse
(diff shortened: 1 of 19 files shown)


handlers/digital-subscription-expiry/src/main/scala/com/gu/digitalSubscriptionExpiry/DigitalSubscriptionExpirySteps.scala
```diff
@@ -5,20 +5,34 @@ import com.gu.digitalSubscriptionExpiry.zuora.GetAccountSummary.{AccountId, Acco
...
+import play.api.libs.json.{JsResult, JsValue, Json, Reads}
 
+case class UrlParams(noActivation: Boolean)
+object UrlParams {
+
+  case class UrlParamsWire(noActivation: String = "false") {
+    def toUrlParams = UrlParams(noActivation == "true")
+  }
+
+  implicit val wireReads = Json.using[Json.WithDefaultValues].reads[UrlParamsWire]
```
johnduffell: since we have the two levels of parsing (wire and normal) with a conversion function, would it be simpler to have it as Option in the Wire model if it's optional, then use .contains("true") in toUrlParams?

handlers/digital-subscription-expiry/src/main/scala/com/gu/digitalSubscriptionExpiry/DigitalSubscriptionExpirySteps.scala
```diff
@@ -5,20 +5,34 @@ import com.gu.digitalSubscriptionExpiry.zuora.GetAccountSummary.{AccountId, Acco
...
+
+  implicit val urlParamsReads = new Reads[UrlParams] {
+    override def reads(json: JsValue): JsResult[UrlParams] = wireReads.reads(json).map(_.toUrlParams)
+  }
```
johnduffell: didn't think about this standard pattern of doing the reads, seems quite nice (at the expense of having a slightly custom reads)

I think you can use the "single abstract method" pattern to save a little boiler plate.  Intellij has a light bulb fix to do that automatically if you prefer.

lib/handler/src/test/scala/com/gu/util/apigateway/ApiGatewayHandlerReadsTest.scala
```diff
@@ -73,4 +82,38 @@ class ApiGatewayHandlerReadsTest extends FlatSpec {
     event should be(expected)
   }
 
+  case class TestParams(testQueryParam: Option[String] = None)
```
johnduffell: is the default None used, or can we remove?

lib/handler/src/test/scala/com/gu/util/apigateway/ApiGatewayHandlerReadsTest.scala
```diff
@@ -73,4 +82,38 @@ class ApiGatewayHandlerReadsTest extends FlatSpec {
...
+    noQueryParamsRequest.queryParamsAsCaseClass[TestParams]() shouldBe ContinueProcessing(TestParams(None))
+  }
+
+  it should "deserialise query string to case class" in {
+    val queryParams = Some(Map(
+      "ignoredParam" -> "ignoredValue",
+      "testQueryParam" -> "testValue"
+    ))
+    val noQueryParamsRequest = ApiGatewayRequest(queryStringParameters = queryParams, body = None, headers = None)
+    noQueryParamsRequest.queryParamsAsCaseClass[TestParams]() shouldBe ContinueProcessing(TestParams(Some("testValue")))
```
johnduffell: I felt a bit uneasy about the `*AsCaseClass` functions to be on ApiGatewayRequest itself - this confirms that because we have to set `body` and `headers` to None even though they are ireelevant to this test.
Not sure if it's worth defining it elsewhere and passing in the apiGatewayRequest (or worse, doing it as an implicit class)?

lib/handler/src/test/scala/com/gu/util/apigateway/AuthTest.scala
```diff
@@ -29,26 +29,15 @@ class AuthTest extends FlatSpec {
...
   }
-
   "credentialsAreValid" should "return true for correct credentials" in {
     val requestAuth = Some(RequestAuth(apiToken = "correctPassword"))
     assert(credentialsAreValid(requestAuth, trustedApiConfig) == true)
   }
 
   "credentialsAreValid" should "return false for an incorrect password" in {
-    val requestAuth = Some(RequestAuth(apiToken = "ndjashjkhajshs"))
+    val requestAuth = Some(RequestAuth(apiToken = "incorrectPassword"))
```
johnduffell: good rename

johnduffell:
also to @jacobwinch as a thinking point: 
Regarding the testing, I'm not sure how many end to end tests we have whether effects tests or standard.  This is a gap affecting our confidence to ship refactoring and we don't want to rely on writing loads of end to end tests (as it won't always happen).  I wonder if in general we should have a mode where we can call the lambda in a special mode, and all the RawEffects record their interactions to an S3 bucket, then we can replay those as a realistic test.  Although these would fail if we are specifically changing the behaviour of a lambda, that wouldn't be the purpose - we'd have to regenerate but we'd understand what we're doing.  This would specifically give us more confidence for general refactors where we are not intending to change the behaviour of lambdas we don't understand.  Any thoughts ?

johnduffell:
excellent, that's really squashed that horrible interdependency, looks spot on. Thanks for doing the extra opportunistic tidyup along the way 🏅 
I have made a few general comments and spotted a few related things that would be nice to tidy up, but nothing that would stop this shipping assuming it works.  The main thing is that it's decoupled now - big win. 👍 🚀

lib/zuora-reports/src/test/scala/com/gu/zuora/reports/FetchFileTest.scala
```diff
@@ -15,7 +15,7 @@ class FetchFileTest extends AsyncFlatSpec {
   def getFakeDownloadStream(path: String): ClientFailableOp[DownloadStream] = {
     val fakeContent = "some fake content"
     val inputStream = new ByteArrayInputStream(fakeContent.getBytes)
-    \/-(DownloadStream(inputStream, fakeContent.getBytes.length))
+    \/-(DownloadStream(inputStream, fakeContent.getBytes.length.toLong))
```
pvighi: this is just to get rid of a "implicit numeric widening" warning

PR #126, opened 2018-06-28
Add contact merge lambda plus bonus json matcher

johnduffell:
This PR adds a new template lambda for merging contacts.

I also removed all the unneeded deps from the handler as they are transitively imported from the libs.  Saves a lot of duplication.  We should do likewise in the other lambdas.

I also added a better way of comparing JSON from the end to end tests, I hope, thoughts appreciated though!  It was a lot more messy than I expected.

Ready for a review apart from the cloudformation which I need to test in CODE and make it work properly.  Take a look at the rest especially at the json matcher stuff see if it's worthwhile @pvighi @jacobwinch @twrichards @paulbrown1982

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -99,7 +99,7 @@ lazy val handler = all(project in file("lib/handler"))
 lazy val effects = all(project in file("lib/effects"))
   .dependsOn(handler)
   .settings(
-    libraryDependencies ++= Seq(okhttp3, logging, scalaz, playJson, scalatest, awsS3, jacksonDatabind)
+    libraryDependencies ++= Seq(okhttp3, logging, scalaz, playJson, scalatest, awsS3, jacksonDatabind, awsLambda)
   )
 
 val effectsDepIncludingTestFolder: ClasspathDependency = effects % "compile->compile;test->test"
--- a/build.sbt
+++ b/build.sbt
@@ -121,6 +121,7 @@ lazy val root = all(project in file(".")).enablePlugins(RiffRaffArtifact).aggreg
   `catalog-service`,
   `identity-retention`,
   `zuora-retention`,
+  `sf-contact-merge`,
   `cancellation-sf-cases`,
   effects,
   handler,
@@ -149,6 +150,10 @@ lazy val `identity-retention` = all(project in file("handlers/identity-retention
 lazy val `zuora-retention` = all(project in file("handlers/zuora-retention"))
   .enablePlugins(RiffRaffArtifact)
   .dependsOn(`zuora-reports`, handler, effectsDepIncludingTestFolder, testDep)
+
+lazy val `sf-contact-merge` = all(project in file("handlers/sf-contact-merge"))
+  .enablePlugins(RiffRaffArtifact)
+  .dependsOn(zuora, handler, effectsDepIncludingTestFolder, testDep)
 
 lazy val `cancellation-sf-cases` = all(project in file("handlers/cancellation-sf-cases"))
   .enablePlugins(RiffRaffArtifact)
--- a/build.sbt
+++ b/build.sbt
@@ -92,14 +92,14 @@ lazy val restHttp = all(project in file("lib/restHttp"))
 
 lazy val handler = all(project in file("lib/handler"))
   .settings(
-    libraryDependencies ++= Seq(okhttp3, logging, scalaz, playJson, scalatest)
+    libraryDependencies ++= Seq(okhttp3, logging, scalaz, playJson, scalatest, awsLambda)
   )
 
 // to aid testability, only the actual handlers called as a lambda can depend on this
 lazy val effects = all(project in file("lib/effects"))
   .dependsOn(handler)
   .settings(
-    libraryDependencies ++= Seq(okhttp3, logging, scalaz, playJson, scalatest, awsS3, jacksonDatabind, awsLambda)
+    libraryDependencies ++= Seq(okhttp3, logging, scalaz, playJson, scalatest, awsS3, jacksonDatabind)
   )
 
 val effectsDepIncludingTestFolder: ClasspathDependency = effects % "compile->compile;test->test"
--- a/handlers/sf-contact-merge/README.md
+++ b/handlers/sf-contact-merge/README.md
@@ -1,6 +1,6 @@
 # sf-contact-merge
 
-This lambda is used to merge together several contacts in SF.  It as initiated by running a report in SF to find out the
+This lambda is used to merge together several contacts in SF.  It is initiated by running a report in SF to find out the
 duplicate sets and which should be the "winner".  Then this is called with the zuora account ids and corrected salesforce
 crmids and contact ids.  Then this will overwrite in zuora.
 
--- a/handlers/sf-contact-merge/README.md
+++ b/handlers/sf-contact-merge/README.md
@@ -1,0 +1,7 @@
+# sf-contact-merge
+
+This lambda is used to merge together several contacts in SF.  It as initiated by running a report in SF to find out the
+duplicate sets and which should be the "winner".  Then this is called with the zuora account ids and corrected salesforce
+crmids and contact ids.  Then this will overwrite in zuora.
+
+It is called manually as part of the ongoing process to get the data quality up.
--- a/handlers/sf-contact-merge/build.sbt
+++ b/handlers/sf-contact-merge/build.sbt
@@ -1,0 +1,24 @@
+// "Any .sbt files in foo, say foo/build.sbt, will be merged with the build definition for the entire build, but scoped to the hello-foo project."
+// https://www.scala-sbt.org/0.13/docs/Multi-Project.html
+name := "sf-contact-merge"
+description:= "Merges together the salesforce account referenced by a set of zuora accounts"
+
+scalacOptions += "-Ypartial-unification"
+
+assemblyJarName := "sf-contact-merge.jar"
+riffRaffPackageType := assembly.value
+riffRaffUploadArtifactBucket := Option("riffraff-artifact")
+riffRaffUploadManifestBucket := Option("riffraff-builds")
+riffRaffManifestProjectName := "MemSub::Membership Admin::SF Contact Merge"
+riffRaffArtifactResources += (file("handlers/sf-contact-merge/cfn.yaml"), "cfn/cfn.yaml")
+
+libraryDependencies ++= Seq(
+  "com.amazonaws" % "aws-lambda-java-log4j" % "1.0.0",
+  "com.amazonaws" % "aws-lambda-java-core" % "1.2.0",
+  "com.amazonaws" % "aws-java-sdk-s3" % "1.11.311",
+  "log4j" % "log4j" % "1.2.17",
+  //"com.squareup.okhttp3" % "okhttp" % "3.9.1",
+  "com.typesafe.play" %% "play-json" % "2.6.9",
+  "org.scalatest" %% "scalatest" % "3.0.1" % "test",
+  "com.fasterxml.jackson.core" % "jackson-databind" % "2.8.11.1"
+)
--- a/handlers/sf-contact-merge/build.sbt
+++ b/handlers/sf-contact-merge/build.sbt
@@ -12,13 +12,4 @@ riffRaffUploadManifestBucket := Option("riffraff-builds")
 riffRaffManifestProjectName := "MemSub::Membership Admin::SF Contact Merge"
 riffRaffArtifactResources += (file("handlers/sf-contact-merge/cfn.yaml"), "cfn/cfn.yaml")
 
-libraryDependencies ++= Seq(
-  "com.amazonaws" % "aws-lambda-java-log4j" % "1.0.0",
-  "com.amazonaws" % "aws-lambda-java-core" % "1.2.0",
-  "com.amazonaws" % "aws-java-sdk-s3" % "1.11.311",
-  "log4j" % "log4j" % "1.2.17",
-  //"com.squareup.okhttp3" % "okhttp" % "3.9.1",
-  "com.typesafe.play" %% "play-json" % "2.6.9",
-  "org.scalatest" %% "scalatest" % "3.0.1" % "test",
-  "com.fasterxml.jackson.core" % "jackson-databind" % "2.8.11.1"
-)
+libraryDependencies ++= Seq()
--- a/handlers/sf-contact-merge/cfn.yaml
+++ b/handlers/sf-contact-merge/cfn.yaml
@@ -18,12 +18,11 @@ Mappings:
         CODE:
             ApiName: sf-contact-merge-api-CODE
             DomainName: sf-contact-merge-code.membership.guardianapis.com
-            CertificateId: e4e6431e-08b9-4315-b997-75f2f9569d6c
+            # need to cloudform, then find out the RegionalDomainName of the api gateway stage manually and replace it here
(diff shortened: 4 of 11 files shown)


build.sbt
```diff
@@ -92,7 +92,7 @@ lazy val restHttp = all(project in file("lib/restHttp"))
 
 lazy val handler = all(project in file("lib/handler"))
   .settings(
-    libraryDependencies ++= Seq(okhttp3, logging, scalaz, playJson, scalatest)
```
johnduffell: this was picking up v1.1.0 instead of v1.2.0 for compilation, but the lambdas were actually putting 1.2.0 on at assembly time.

handlers/sf-contact-merge/build.sbt
```diff
@@ -0,0 +1,15 @@
...
+scalacOptions += "-Ypartial-unification"
+
+assemblyJarName := "sf-contact-merge.jar"
+riffRaffPackageType := assembly.value
+riffRaffUploadArtifactBucket := Option("riffraff-artifact")
+riffRaffUploadManifestBucket := Option("riffraff-builds")
+riffRaffManifestProjectName := "MemSub::Membership Admin::SF Contact Merge"
+riffRaffArtifactResources += (file("handlers/sf-contact-merge/cfn.yaml"), "cfn/cfn.yaml")
+
+libraryDependencies ++= Seq()
```
johnduffell: we no longer need any deps on this as they are all imported transitively.

handlers/sf-contact-merge/src/test/scala/com/gu/sfContactMerge/HandlerEffectsTest.scala
```diff
@@ -0,0 +1,100 @@
...
+class HandlerEffectsTest extends FlatSpec with Matchers {
+
+  import TestData._
+
+  case class ApiResponse(statusCode: String, body: JsEmbeddded[Message], headers: Map[String, String] = Map("Content-Type" -> "application/json"))
+
+  case class Message(message: String)
+
+  implicit val mF: OFormat[Message] = Json.format[Message]
+  implicit val apiF: OFormat[ApiResponse] = Json.format[ApiResponse]
```
johnduffell: we need a format (i.e. reads and writes) rather than just a reads, because we actually convert the string Json=>jsvalue=>ApiResponse=>jsvalue and then we check the jsvalues match (i.e. no extra fields that would go uncompared) before returning the ApiResponse for equality checking.

handlers/sf-contact-merge/src/test/scala/com/gu/sfContactMerge/HandlerEffectsTest.scala
```diff
@@ -0,0 +1,100 @@
...
+import com.gu.sfContactMerge.JsonMatchers._
+import com.gu.test.EffectsTest
+import org.scalatest.{FlatSpec, Matchers}
+import play.api.libs.json._
+
+class HandlerEffectsTest extends FlatSpec with Matchers {
+
+  import TestData._
+
+  case class ApiResponse(statusCode: String, body: JsEmbeddded[Message], headers: Map[String, String] = Map("Content-Type" -> "application/json"))
```
johnduffell: the JsEmbedded just informs the json de/serialiser that it's actually a json object itself serialised into a JsString, rather than normal json.

handlers/sf-contact-merge/src/test/scala/com/gu/sfContactMerge/JsonMatchers.scala
```diff
@@ -0,0 +1,60 @@
+package com.gu.sfContactMerge
```
johnduffell: this is some new test matcher for json, but if it's a good idea we could probably move it somewhere we can use it for all the tests that check against some json.

handlers/sf-contact-merge/src/test/scala/com/gu/sfContactMerge/JsonMatchers.scala
```diff
@@ -0,0 +1,70 @@
...
+
+  implicit def weW[C: OFormat]: Writes[WithoutExtras[C]] = (o: WithoutExtras[C]) =>
+    Json.toJson(o.c)
+
+  implicit def weR[C: OFormat]: Reads[WithoutExtras[C]] = (jsValue: JsValue) =>
+    implicitly[Reads[C]].reads(jsValue).flatMap { c =>
+      val backToJsValue: JsValue = Json.toJsObject(c)
+      if (backToJsValue == jsValue)
+        JsSuccess(WithoutExtras(c))
+      else JsError("extra fields")
```
johnduffell: maybe we should have a second Map[...] field on the WithoutExtras to contain the extras, then we can deserialise and check those properly rather than just failing with this basic message?

handlers/sf-contact-merge/cfn.yaml
```diff
@@ -0,0 +1,227 @@
+AWSTemplateFormatVersion: "2010-09-09"
```
johnduffell: still have to test and refine the cloudformation before it's ready to ship, but the rest of the PR is ready to look at for comments!

handlers/sf-contact-merge/src/test/scala/com/gu/sfContactMerge/JsonMatchers.scala
```diff
@@ -0,0 +1,70 @@
...
+
+    def jsonMatches[C: OFormat](expected: C): Assertion = {
+      val expectedJson: JsResult[WithoutExtras[C]] = JsSuccess(WithoutExtras(expected))
+      val actualJson: JsResult[WithoutExtras[C]] = Json.parse(actual).validate[WithoutExtras[C]]
+      actualJson should be(expectedJson)
+    }
+
+  }
+
+  // *** //
```
jacobwinch: This can be deleted

handlers/sf-contact-merge/README.md
```diff
@@ -0,0 +1,7 @@
+# sf-contact-merge
+
+This lambda is used to merge together several contacts in SF.  It as initiated by running a report in SF to find out the
```
jacobwinch: Small typo here "It **is** initiated"

handlers/sf-contact-merge/src/test/scala/com/gu/sfContactMerge/JsonMatchers.scala
jacobwinch: I think it would be useful to know what the extra fields are. But maybe if we get this JSError we can just log the full JSON that we're trying to parse?

handlers/sf-contact-merge/src/test/scala/com/gu/sfContactMerge/JsonMatchers.scala
```diff
@@ -0,0 +1,70 @@
...
+
+  /*
+  this marks a field as being a JsString which contains serialised json.  This is common where the body
+  is json within a json api gateway json descriptor.
+  When we deserialise, we store the serialised form to use when we serialise again, so that the higher level
+  JsValue comparison will still match exactly.
+  The structure of the embedded value will be checked when it in turn is deseralised.
+   */
+  case class JsEmbeddded[C](c: C, origIgnoredInComparison: Option[String] = None) {
```
pvighi: is  origIgnoredInComparison just the original value but we are also encoding in the name the fact that it is ignored  in comparisons  ?
Because if we are I would just call it orig, and maybe turn this into a regular class instead of a case class

handlers/sf-contact-merge/src/test/scala/com/gu/sfContactMerge/JsonMatchers.scala
```diff
@@ -0,0 +1,70 @@
...
+      case _ => false
+    }
+
+    override def hashCode(): Int = c.hashCode
+
+    override def toString: String = s"JsEmbeddded(${c.toString})"
+  }
+
+  implicit def eW[C: OFormat]: Writes[JsEmbeddded[C]] = (o: JsEmbeddded[C]) =>
+    JsString(o.origIgnoredInComparison.getOrElse(s"missing orig (salt ${scala.util.Random.nextInt})"))
```
pvighi: why do we need this "salt [randomInt]" part?

handlers/sf-contact-merge/src/test/scala/com/gu/sfContactMerge/JsonMatchers.scala
pvighi: Also, I think should be  Jsembedded  not JsEmbeddded

handlers/sf-contact-merge/src/test/scala/com/gu/sfContactMerge/JsonMatchers.scala
pvighi: I'm not sure I'm understanding this correctly but wouldn't this force the test to provide a reads for C ? 
In most cases when we check output json we only have writes for those classes

handlers/sf-contact-merge/src/test/scala/com/gu/sfContactMerge/JsonMatchers.scala
johnduffell: __I was debating whether the above should even be in the same file as the below, makes imports easier but it's not totally right

handlers/sf-contact-merge/src/test/scala/com/gu/sfContactMerge/JsonMatchers.scala
johnduffell: yes it's just the raw string because we need to not accidentally cause a difference in the JsValue when checking for extra fields.  In reality I shouldn't use `if (backToJsValue == jsValue)` I should use one that just checks the structure (for extra fields) rather than the contents of each field.
I will see if that makes more sense.

handlers/sf-contact-merge/src/test/scala/com/gu/sfContactMerge/JsonMatchers.scala
johnduffell: deleted this now

handlers/sf-contact-merge/README.md
johnduffell: done

handlers/sf-contact-merge/src/test/scala/com/gu/sfContactMerge/JsonMatchers.scala
johnduffell: I have solved both of your comments by writing a == function called sameStructure that only checks the structure was recreated rather than the contents (as the contents will be checked by the assert in the test)

handlers/sf-contact-merge/src/test/scala/com/gu/sfContactMerge/HandlerEffectsTest.scala
```diff
@@ -0,0 +1,104 @@
...
+import com.gu.sfContactMerge.JsonMatchers._
+import com.gu.test.EffectsTest
+import org.scalatest.{FlatSpec, Matchers}
+import play.api.libs.json._
+
+class HandlerEffectsTest extends FlatSpec with Matchers {
+
+  import TestData._
+
+  case class TestJsonFormat(
```
johnduffell: I have renamed this to something obviously unlreated to the "real" objects, to make it clear this is simply a nicer way to test against Json in the tests.

handlers/sf-contact-merge/cfn.yaml
```diff
@@ -0,0 +1,226 @@
...
+                !Sub sf-contact-merge-${Stage}
+            Code:
+                S3Bucket: zuora-auto-cancel-dist
+                S3Key: !Sub membership/${Stage}/sf-contact-merge/sf-contact-merge.jar
+            Handler: com.gu.sfContactMerge.Handler::apply
+            Environment:
+                Variables:
+                  Stage: !Ref Stage
+            Role:
+                Fn::GetAtt:
```
jacobwinch: I find the short form a bit easier to read: 
_!Get SfContactMergeRole.Arn_

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-getatt.html

johnduffell: Ill fix - I just copied that from another - we really need to work out how to not duplicate the cfn everywhere 😞

handlers/sf-contact-merge/cfn.yaml
```diff
@@ -0,0 +1,224 @@
...
+
+Conditions:
+  CreateProdMonitoring: !Equals [ !Ref Stage, PROD ]
+
+Mappings:
+    StageMap:
+        CODE:
+            ApiName: sf-contact-merge-api-CODE
+            DomainName: sf-contact-merge-code.membership.guardianapis.com
+            # need to cloudform, then find out the RegionalDomainName of the api gateway stage manually and replace it here
```
pvighi: it doesn't really matter but doesn't this comment belong in the prod version of this ?

handlers/sf-contact-merge/src/test/scala/com/gu/sfContactMerge/JsonMatchers.scala
```diff
@@ -0,0 +1,69 @@
...
+      val actualJson: JsResult[WithoutExtras[C]] = Json.parse(actual).validate[WithoutExtras[C]]
+      actualJson should be(expectedJson)
+    }
+
+  }
+
+  case class JsEmbedded[C](c: C)
+
+  implicit def eW[C: OFormat]: Writes[JsEmbedded[C]] = (o: JsEmbedded[C]) =>
+    JsString(Json.prettyPrint(Json.toJsObject(o.c)))
```
pvighi: just a reminder of the conversation we had: maybe we should have a doc somewhere with all the shorthand / syntactic sugar to help people encountering it for the first time. For example in this line unless you know the names of what you are looking for it is kind of hard to google

handlers/sf-contact-merge/src/test/scala/com/gu/sfContactMerge/JsonMatchers.scala
```diff
@@ -0,0 +1,69 @@
...
+    case (o1: JsObject, o2: JsObject) =>
+      if (o1.keys == o2.keys) {
+        // check what's inside
+        val keysMatch = o1.keys.map { key =>
+          (for {
+            v1 <- o1.value.get(key)
+            v2 <- o2.value.get(key)
+          } yield sameStructure(v1, v2)).getOrElse(false /*won't happen*/ )
+        }
+        println(s"keysMatch $keysMatch")
```
pvighi: is this println left in on purpose?

handlers/sf-contact-merge/src/test/scala/com/gu/sfContactMerge/JsonMatchers.scala
```diff
@@ -0,0 +1,69 @@
...
+      val withExtrasCheck = maybeC.flatMap { c =>
+        val backToJsValue: JsValue = Json.toJsObject(c)
+        if (sameStructure(backToJsValue, jsValue))
+          JsSuccess(c)
+        else JsError(s"extra fields, $backToJsValue == $jsValue")
+      }
+      withExtrasCheck.map(c => JsEmbedded(c))
+    }
+
+  def sameStructure: (JsValue, JsValue) => Boolean = {
```
pvighi: why do we need this sameStructure? can't we check that the two JsValues are the equal ?

PR #127, opened 2018-07-02
Separate configuration loading example

pvighi:
The idea is that instead of all lambdas sharing the same configuration file, we would have multiple files and each lambda will load the secrets from several of them. My initial idea is to have files for the secrets that are accessed by multiple lambdas (like Zuora, Salesforce, maybe Stripe) and lambda specific files for all the non shared secrets.

Along with these changes we would have to change the cloudformation template to restrict each lambda to just the configuration files it needs to access.

In this PR there's an implementation of the alternative config loading code and some unit tests to show how it would be used. No actual lambdas were modified to use it yet.
If we decide to go ahead in this direction we would have to refactor all the lambdas to load their configuration in this way.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -92,7 +92,7 @@ lazy val restHttp = all(project in file("lib/restHttp"))
 
 lazy val handler = all(project in file("lib/handler"))
   .settings(
-    libraryDependencies ++= Seq(okhttp3, logging, scalaz, playJson, scalatest)
+    libraryDependencies ++= Seq(okhttp3, logging, scalaz, playJson, scalatest, awsS3)
   )
 
 // to aid testability, only the actual handlers called as a lambda can depend on this
--- a/build.sbt
+++ b/build.sbt
@@ -92,7 +92,7 @@ lazy val restHttp = all(project in file("lib/restHttp"))
 
 lazy val handler = all(project in file("lib/handler"))
   .settings(
-    libraryDependencies ++= Seq(okhttp3, logging, scalaz, playJson, scalatest, awsS3)
+    libraryDependencies ++= Seq(okhttp3, logging, scalaz, playJson, scalatest, awsLambda, awsS3)
   )
 
 // to aid testability, only the actual handlers called as a lambda can depend on this
@@ -121,6 +121,7 @@ lazy val root = all(project in file(".")).enablePlugins(RiffRaffArtifact).aggreg
   `catalog-service`,
   `identity-retention`,
   `zuora-retention`,
+  `sf-contact-merge`,
   `cancellation-sf-cases`,
   effects,
   handler,
@@ -149,6 +150,10 @@ lazy val `identity-retention` = all(project in file("handlers/identity-retention
 lazy val `zuora-retention` = all(project in file("handlers/zuora-retention"))
   .enablePlugins(RiffRaffArtifact)
   .dependsOn(`zuora-reports`, handler, effectsDepIncludingTestFolder, testDep)
+
+lazy val `sf-contact-merge` = all(project in file("handlers/sf-contact-merge"))
+  .enablePlugins(RiffRaffArtifact)
+  .dependsOn(zuora, handler, effectsDepIncludingTestFolder, testDep)
 
 lazy val `cancellation-sf-cases` = all(project in file("handlers/cancellation-sf-cases"))
   .enablePlugins(RiffRaffArtifact)
--- a/lib/handler/src/main/scala/com/gu/util/config2/LoadConfig2.scala
+++ b/lib/handler/src/main/scala/com/gu/util/config2/LoadConfig2.scala
@@ -13,28 +13,22 @@ case class ConfigLocation[CONFIG](val path: String, version: Int)
 
 class LoadConfig2(stage: Stage, fetchString: GetObjectRequest => Try[String]) extends Logging {
 
-  //todo maybe this is a good chance to change the directory we load the config from
-  val basePath = s"membership/payment-failure-lambdas/${stage.value}"
   val bucketName = "gu-reader-revenue-private"
 
   def apply[CONF](implicit configLocation: ConfigLocation[CONF], reads: Reads[CONF]): ConfigFailure \/ CONF = {
-    logger.info(s"Attempting to load config in $stage")
-    val relativePath = if (stage.value == "DEV") configLocation.path else pathWithVersion(configLocation)
-    val request = new GetObjectRequest(bucketName, s"$basePath/$relativePath")
+    //todo maybe this is a good chance to change the directory we load the config from
+    val basePath = s"membership/payment-failure-lambdas/${stage.value}"
 
+    logger.info(s"Attempting to load config in $stage")
+    val versionString = if (stage.value == "DEV") "" else s".v${configLocation.version}"
+    val relativePath = s"${configLocation.path}$versionString.json"
+    val request = new GetObjectRequest(bucketName, s"$basePath/$relativePath")
     for {
       configStr <- toDisjunction(fetchString(request))
       jsValue <- toDisjunction(Try(Json.parse(configStr)))
-      _ <- validateStage(jsValue)
+      _ <- validateStage(jsValue, stage)
       config <- toDisjunction(Json.fromJson[CONF](jsValue))
     } yield config
-  }
-
-  def pathWithVersion(configLocation: ConfigLocation[_]) = {
-    val path = configLocation.path
-    val versionString = s".v${configLocation.version}"
-    val lastDotPosition = path.lastIndexOf(".")
-    if (lastDotPosition > 0) path.substring(0, lastDotPosition) + versionString + path.substring(lastDotPosition) else path + versionString
   }
 
   case class ConfigWithStage(stage: String)
@@ -43,10 +37,10 @@ class LoadConfig2(stage: Stage, fetchString: GetObjectRequest => Try[String]) ex
     implicit val reads = Json.reads[ConfigWithStage]
   }
 
-  def validateStage(jsValue: JsValue): ConfigFailure \/ Unit = {
+  def validateStage(jsValue: JsValue, expectedStage: Stage): ConfigFailure \/ Unit = {
     jsValue.validate[ConfigWithStage] match {
-      case JsSuccess(ConfigWithStage(stage.value), _) => \/-(())
-      case JsSuccess(ConfigWithStage(otherStage), _) => -\/(ConfigFailure(s"Expected to load ${stage.value} config, but loaded $otherStage config"))
+      case JsSuccess(ConfigWithStage(expectedStage.value), _) => \/-(())
+      case JsSuccess(ConfigWithStage(otherStage), _) => -\/(ConfigFailure(s"Expected to load ${expectedStage.value} config, but loaded $otherStage config"))
       case JsError(error) => -\/(ConfigFailure(s"could not parse stage in configuration file: ${error}"))
     }
   }
--- a/lib/handler/src/main/scala/com/gu/util/config2/LoadConfig2.scala
+++ b/lib/handler/src/main/scala/com/gu/util/config2/LoadConfig2.scala
@@ -19,6 +19,8 @@ object ConfigWithStage {
 
 object LoadConfig2 extends Logging {
 
+  val bucketName = "gu-reader-revenue-private"
+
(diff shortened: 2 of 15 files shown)


lib/handler/src/test/scala/com/gu/util/config2/LoadConfig2Test.scala
```diff
@@ -0,0 +1,109 @@
...
+
+  val prodStage = Stage("PROD")
+
+
+  case class TestConfig(someValue: String, someOtherValue: Int)
+
+  object TestConfig {
+    implicit val reads = Json.reads[TestConfig]
+    implicit val location = ConfigLocation[TestConfig](path = "someDir/filename.json", version = 2)
+  }
```
pvighi: This TestConfig is just an example of a configuration case class.

We could mostly reuse the ones we already have, we just need to add the implicit ConfigLocation to declare the file name and version to use when loading this configuration class

lib/handler/src/main/scala/com/gu/util/config2/LoadConfig2.scala
```diff
@@ -0,0 +1,71 @@
...
+import com.gu.util.config.ConfigReads.ConfigFailure
+import com.gu.util.config.Stage
+import play.api.libs.json._
+import scalaz.{-\/, \/, \/-}
+
+import scala.util.{Failure, Success, Try}
+
+case class ConfigLocation[CONFIG](val path: String, version: Int)
+
+class LoadConfig2(stage: Stage, fetchString: GetObjectRequest => Try[String]) extends Logging {
```
pvighi: We define this as a class instead of an object like everything else in order to be able to provide all its dependencies without having to fix the generic CONF param.
 The idea is to wire all the dependencies in the handler by creating an instance of this class with a stage and s3Loading function. The apply method of this class can now be used with different generic params to load different configurations

lib/handler/src/main/scala/com/gu/util/config2/LoadConfig2.scala
```diff
@@ -0,0 +1,71 @@
...
+
+  //todo maybe this is a good chance to change the directory we load the config from
+  val basePath = s"membership/payment-failure-lambdas/${stage.value}"
```
pvighi: we should probably agree on a new name for the directory where the config files will live, it would be a good idea if this new name matched the future name for this repo

lib/handler/src/main/scala/com/gu/util/config2/LoadConfig2.scala
```diff
@@ -0,0 +1,71 @@
...
+    val versionString = s".v${configLocation.version}"
+    val lastDotPosition = path.lastIndexOf(".")
+    if (lastDotPosition > 0) path.substring(0, lastDotPosition) + versionString + path.substring(lastDotPosition) else path + versionString
+  }
+
+  case class ConfigWithStage(stage: String)
+
+  object ConfigWithStage {
+    implicit val reads = Json.reads[ConfigWithStage]
+  }
```
pvighi: just to preserve existing functionality every config file should contain a stage parameter which will be verified to make sure we don't accidentally copy config files in the wrong place. All config files will be first deserialised into this class and if the stage check passes they will be deserialised again into the custom class specified in the CONF param

lib/handler/src/main/scala/com/gu/util/config2/LoadConfig2.scala
jacobwinch: 👍

lib/handler/src/main/scala/com/gu/util/config2/LoadConfig2.scala
```diff
@@ -0,0 +1,71 @@
...
+
+import scala.util.{Failure, Success, Try}
+
+case class ConfigLocation[CONFIG](val path: String, version: Int)
+
+class LoadConfig2(stage: Stage, fetchString: GetObjectRequest => Try[String]) extends Logging {
+
+  //todo maybe this is a good chance to change the directory we load the config from
+  val basePath = s"membership/payment-failure-lambdas/${stage.value}"
+  val BucketName = "gu-reader-revenue-private"
```
jacobwinch: bucketName?

lib/handler/src/main/scala/com/gu/util/config2/LoadConfig2.scala
```diff
@@ -0,0 +1,71 @@
...
+
+    for {
+      configStr <- toDisjunction(fetchString(request))
+      jsValue <- toDisjunction(Try(Json.parse(configStr)))
+      _ <- validateStage(jsValue)
+      config <- toDisjunction(Json.fromJson[CONF](jsValue))
+    } yield config
+  }
+
+  def pathWithVersion(configLocation: ConfigLocation[_]) = {
```
jacobwinch: Would this new approach allow us to drop the config versioning stuff? (I.e. if we add new secrets, presumably that will be a separate file, and we'll need a specific code change to pick up that file?)

lib/handler/src/main/scala/com/gu/util/config2/LoadConfig2.scala
pvighi: well..  I did this on purpose because I wanted to use the naming convention for constants in scala according to [this thing](https://docs.scala-lang.org/style/naming-conventions.html#constants-values-variable-and-methods). 
But now that I think about it it doesn't make any sense since in the end it's just a regular val ( and now that I look at the doc more closely it doesn't even match their definition of a constant in scala).
I'll change it back, I think I was just regressed to Java thinking when I did this

lib/handler/src/main/scala/com/gu/util/config2/LoadConfig2.scala
pvighi: I'm not sure I understand the question so I'll say something related to versions hoping I answer it by accident.
We specify a version for each file in the ConfigLocation instance that we take as an implicit param when loading a config file.  We can modify the version in the code to make sure the lambda will load a compatible version of the configuration
you can add new secrets in a new file or in an existing file without increasing the version, but a code change will be needed regardless  (either to load a new file or just adding the new fields in the existing config class)

build.sbt
```diff
@@ -92,7 +92,7 @@ lazy val restHttp = all(project in file("lib/restHttp"))
 
 lazy val handler = all(project in file("lib/handler"))
   .settings(
-    libraryDependencies ++= Seq(okhttp3, logging, scalaz, playJson, scalatest)
```
johnduffell: we should probably separate the Config stuff from the Api gateway stuff some time too
this means the s3 would go to the config and the okhttp would go to the api gateway

lib/handler/src/main/scala/com/gu/util/config2/LoadConfig2.scala
johnduffell: good, just remember that everything between the braces of this class gets stage and fetchString as parameters automatically.  Anything that doesn't require both should be moved to a companion object. Personally I think we would only have the apply method as a strict rule, and everything else should be explicitly parameterised separately.  As we are only using the class as a vehicle to pre apply without the type parameters, so we shouldn't also use it for code organisation.

lib/handler/src/main/scala/com/gu/util/config2/LoadConfig2.scala
```diff
@@ -0,0 +1,71 @@
...
+    val path = configLocation.path
+    val versionString = s".v${configLocation.version}"
+    val lastDotPosition = path.lastIndexOf(".")
+    if (lastDotPosition > 0) path.substring(0, lastDotPosition) + versionString + path.substring(lastDotPosition) else path + versionString
```
johnduffell: I don't object enough to say it's important, but I'm wondering if it's worth the hassle to mess with the string (string manipulation makes me nervous), just to avoid people having to change it in dev.  what about we just make the ConfigLocation class have a "dev" and "nondev" or even we just append the .json automatically so we don't have to search for it?

lib/handler/src/main/scala/com/gu/util/config2/LoadConfig2.scala
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  case class ConfigWithStage(stage: String)
```
johnduffell: this is just the `StageFromConfig` or similar now

lib/handler/src/test/scala/com/gu/util/config2/LoadConfig2Test.scala
```diff
@@ -0,0 +1,106 @@
...
+import com.gu.util.config.Stage
+import org.scalatest.{FlatSpec, Matchers}
+import play.api.libs.json.Json
+import scalaz.{-\/, \/-}
+
+import scala.util.Try
+
+class LoadConfig2Test extends FlatSpec with Matchers {
+
+  def fakeS3Load(responsesForStage: Map[String, String])(req: GetObjectRequest): Try[String] = Try {
```
johnduffell: if we only ever pass in one item inside the map, we could just pass two parameters instead

johnduffell:
looks great to me, I've made one important comment about not putting anything in the class unless it's the specific method we're trying to partially apply, other than that looks great (just a few minor things)
my main concern would be how easy it is to copy the config files into place, I tend to get into a bit of a mess with S3, but I think that might be a one off task.  I guess once you have the right `AWS s3 cp` locally you can get everything sorted out in this PR, and then in future people will only be tweaking one file at a time.
if this gets the goahead it would be worth thinking about whether to split the config out one piece at a time or whether to split all the files in one go and do one deploy to changeover.

that's about all my thoughts, I think splitting will be great for being able to rotate the keys and improve the permissions further.
👍 🥇

lib/handler/src/main/scala/com/gu/util/config2/LoadConfig2.scala
pvighi: any name suggestions ? supporter-experience-lambdas ?

johnduffell:
> any name suggestions ? supporter-experience-lambdas ?

I think your original comment was hidden by github.
I think naming it after the dev/tech side might be better than naming after the huddle.   maybe reader-revenue-lambdas would be good, if we're allowed to be more general? or reader-revenue-inlife-lambdas? @jacobwinch @paulbrown1982 @twrichards  any ideas?

lib/handler/src/main/scala/com/gu/util/config2/LoadConfig2.scala
```diff
@@ -0,0 +1,65 @@
...
+    val request = new GetObjectRequest(bucketName, s"$basePath/$relativePath")
+    for {
+      configStr <- toDisjunction(fetchString(request))
+      jsValue <- toDisjunction(Try(Json.parse(configStr)))
+      _ <- validateStage(jsValue, stage)
+      config <- toDisjunction(Json.fromJson[CONF](jsValue))
+    } yield config
+  }
+
+  case class ConfigWithStage(stage: String)
```
johnduffell: I noticed you did a commit but didn't do what I expected, not sure if you meant to move everything expect apply into the object instead of the class or disagreed or my comment wasn't clear?

pvighi: I didn't understand your comment about StageFromConfig but I forgot to ask you about it.
What did you mean by 'this is just the StageFromConfig or similar now' ?

the changes in this last commit were just to remove references to stage outside the apply method and simplify the location manipulation by just appending to the end.

johnduffell: oh I just meant it's got a funny name, it used to be "config with a stage" but now it's just "a stage" so should probabyly rename it.

I am not too worried about that though, I'm more worried about the extra functions etc embedded inside the class, because that causes a hidden dependency on the class variables (that are probably not used)

pvighi: I could also move all the functions to be embedded in the apply method

jacobwinch:
> I think naming it after the dev/tech side might be better than naming after the huddle.

I agree with this - surely we're due a team name change soon?

> reader-revenue-inlife

Perhaps this (dropping the word 'lambdas' so that we can adopt new shiny things when AWS release them without having to change the name again!)

johnduffell:
great, the new changes look great 👍 glad we've got a name decided too!

PR #130, opened 2018-07-04
Identity Cookie Decoding now uses environment specific keys

twrichards:
The 'stage' is now passed into the identity cookie checking to determine whether to use `ProductionKeys` or `PreProductionKeys` for decoding the cookie (in order to extract the identity id).

### DIFF
--- a/handlers/cancellation-sf-cases/src/main/scala/com/gu/cancellation/sf_cases/Handler.scala
+++ b/handlers/cancellation-sf-cases/src/main/scala/com/gu/cancellation/sf_cases/Handler.scala
@@ -31,7 +31,13 @@ object Handler extends Logging {
   type LazySalesforceAuthenticatedReqMaker = () => ApiGatewayOp[RestRequestMaker.Requests]
 
   def raiseCase(inputStream: InputStream, outputStream: OutputStream, context: Context): Unit =
-    runWithEffects(RaiseCase.steps, RawEffects.response, RawEffects.stage, RawEffects.s3Load, LambdaIO(inputStream, outputStream, context))
+    runWithEffects(
+      RaiseCase.steps(RawEffects.stage),
+      RawEffects.response,
+      RawEffects.stage,
+      RawEffects.s3Load,
+      LambdaIO(inputStream, outputStream, context)
+    )
 
   object RaiseCase {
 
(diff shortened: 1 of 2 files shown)


handlers/cancellation-sf-cases/src/main/scala/com/gu/cancellation/sf_cases/Handler.scala
```diff
@@ -70,9 +70,9 @@ object Handler extends Logging {
           .toApiGatewayOp("raise sf case")
       } yield raiseCaseResponse
 
-    def steps(sfRequests: LazySalesforceAuthenticatedReqMaker)(apiGatewayRequest: ApiGatewayRequest) =
+    def steps(stage:Stage)(sfRequests: LazySalesforceAuthenticatedReqMaker)(apiGatewayRequest: ApiGatewayRequest) =
```
johnduffell: small style thing - should have a space after the colon.  I think when you sbt compile it should auto format things like that, but sometimes that doesn't happen due to ordering

handlers/cancellation-sf-cases/src/main/scala/com/gu/cancellation/sf_cases/Handler.scala
```diff
@@ -100,7 +100,7 @@ object Handler extends Logging {
   }
 
   def runWithEffects(
-    steps: LazySalesforceAuthenticatedReqMaker => ApiGatewayRequest => ApiResponse,
+    steps: Stage => LazySalesforceAuthenticatedReqMaker => ApiGatewayRequest => ApiResponse,
```
johnduffell: since we are passing in Stage anyway to this function, can we pass in before we call runWithEffects? (I realise s3Load has exactly the same issue)

handlers/cancellation-sf-cases/src/main/scala/com/gu/identity/IdentityCookieToIdentityId.scala
```diff
@@ -1,19 +1,25 @@
...
-  def apply(headersOption: Option[Map[String, String]]): ApiGatewayOp[String] = for {
-    headers <- headersOption.toApiGatewayContinueProcessing(badRequest, "no headers")
-    cookieHeader <- headers.get("Cookie").toApiGatewayContinueProcessing(badRequest, "no cookie")
-    scGuU <- extractCookieHeaderValue(cookieHeader, "SC_GU_U")
-    cookieDecoder = new IdentityCookieDecoder(new PreProductionKeys) // TODO review keys selection
-    userFromScGuU <- cookieDecoder.getUserDataForScGuU(scGuU).toApiGatewayContinueProcessing(unauthorized)
-  } yield userFromScGuU.getId
+  def apply(headersOption: Option[Map[String, String]], stage: Stage): ApiGatewayOp[String] = {
+
+    lazy val keys = if(stage.isProd) new ProductionKeys else new PreProductionKeys
```
johnduffell: no reason not to do this on the for comprehension between lines 18 and 19

johnduffell:
yep looks fine, just a few small pointers for style etc but nothing much 👍

PR #131, opened 2018-07-05
Config refactor

pvighi:
I still need to delete the old loadConfig and write some effects tests to check that the files exist and are parseable in CODE and PROD. 
Changes:
* Refactored all lambdas to use the new modular configuration loading
* Restricted lambdas permissions to only access the configuration files they need

This PR contains mostly the same changes repeated over and over again.
There might be some opportunities to refactor further ( like making sure only the lambdas that actually need it are loading the TrustedApi config) but those will be addressed in future PR as this one is big enough already.


cloudformation.yaml
```diff
@@ -55,8 +55,11 @@ Resources:
                       Statement:
                           - Effect: Allow
                             Action: s3:GetObject
-                            Resource: !Sub arn:aws:s3:::gu-reader-revenue-private/membership/payment-failure-lambdas/${Stage}/*
-
+                            Resource:
```
jacobwinch: Probably a change for another PR, but we should swap this to use the Readable/Writeable buckets pattern in the future

handlers/digital-subscription-expiry/src/main/scala/com/gu/digitalSubscriptionExpiry/Handler.scala
```diff
@@ -54,13 +45,13 @@ object Handler extends Logging {
...
-
-    ApiGatewayHandler[StepsConfig](lambdaIO)(for {
-      config <- LoadConfig.default[StepsConfig](implicitly)(stage, s3Load(stage))
-        .toApiGatewayOp("load config")
-      configuredOp = operation(config)
-
-    } yield (config, configuredOp))
+    val loadConfig = LoadConfigModule(stage, fetchString)
+    ApiGatewayHandler(lambdaIO)(for {
+      zuoraConfig <- loadConfig[ZuoraRestConfig].toApiGatewayOp("load zuora config")
```
jacobwinch: Out of interest, did we compare execution times before/after loading multiple config fragments?

handlers/digital-subscription-expiry/src/test/scala/com/gu/digitalSubscriptionExpiry/zuora/GetSubscriptionEffectsTest.scala
```diff
@@ -2,24 +2,22 @@ package com.gu.digitalSubscriptionExpiry.zuora
...
 import org.scalatest.{FlatSpec, Matchers}
 import scalaz.{-\/, \/-}
 
 class GetSubscriptionEffectsTest extends FlatSpec with Matchers {
 
   private def actual(testSubscriptionId: SubscriptionId): ApiGatewayOp[SubscriptionResult] = for {
-    configAttempt <- S3ConfigLoad.load(Stage("DEV")).toApiGatewayOp("couldn't load")
-    config <- LoadConfig.parseConfig[StepsConfig](configAttempt).toApiGatewayOp("couldn't parse")
-    zuoraRequests = ZuoraRestRequestMaker(RawEffects.response, config.stepsConfig.zuoraRestConfig)
+    zuoraRestConfig <- LoadConfigModule(Stage("DEV"), GetFromS3.fetchString)[ZuoraRestConfig].toApiGatewayOp("coun't load config")
```
jacobwinch: small typo in the logging here

handlers/zuora-retention/cfn.yaml
```diff
@@ -193,7 +193,7 @@ Resources:
                         Statement:
                             - Effect: Allow
                               Action: s3:GetObject
-                              Resource: !Sub arn:aws:s3:::gu-reader-revenue-private/membership/payment-failure-lambdas/${Stage}/*
+                              Resource: !Sub "arn:aws:s3:::gu-reader-revenue-private/membership/support-service-lambdas/${Stage}/zuoraRest-${Stage}*.json"
```
jacobwinch: Not introduced by this PR (i.e. feel free to ignore this comment), but it might be cleaner to define this policy once and reference it 4 times?

build.sbt
```diff
@@ -67,7 +67,7 @@ lazy val zuora = all(project in file("lib/zuora"))
   .dependsOn(
     restHttp,
     testDep,
-    handler % "test->test",//only for the config, which should ideally be split out
+    handler,// TODO only for the config, which should ideally be split out
```
johnduffell: I'd probably increase the strength of that comment at the time! (replace the "should ideally" with "needs to")

cloudformation.yaml
```diff
@@ -55,8 +55,11 @@ Resources:
...
+                            - !Sub arn:aws:s3:::gu-reader-revenue-private/membership/support-service-lambdas/${Stage}/zuoraRest-${Stage}.*.json
+                            - !Sub arn:aws:s3:::gu-reader-revenue-private/membership/support-service-lambdas/${Stage}/trustedApi-${Stage}.*.json
+                            - !Sub arn:aws:s3:::gu-reader-revenue-private/membership/support-service-lambdas/${Stage}/exactTarget-${Stage}.*.json
+                            - !Sub arn:aws:s3:::gu-reader-revenue-private/membership/support-service-lambdas/${Stage}/stripe-${Stage}.*.json
```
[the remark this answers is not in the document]
johnduffell: maybe add to the "renaming" section of the tech debt doc?

lib/handler/src/main/scala/com/gu/util/config/LoadConfigModule.scala
```diff
@@ -18,27 +15,29 @@ object ConfigWithStage {
 
 object LoadConfigModule extends Logging {
 
+  type StringFromS3 = (String, String) => Try[String]
```
johnduffell: it's great improvement to have a name, but I wonder if it would be better or worse to go further - wrap the strings just to make it clearer what they are? e.g. 
```scala
case class S3Location(bucket: String, path: String)
type StringFromS3 = S3Location => Try[String]
```

cloudformation.yaml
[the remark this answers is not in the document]
jacobwinch: We're already changing the bucket location (to fix the naming) in this PR, so I'd prefer to do it as part of this change (or to decide we don't care and leave it under membership forever)

handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/Handler.scala
```diff
@@ -29,28 +28,23 @@ object Handler {
...
     response: Request => Response,
     lambdaIO: LambdaIO
   ): Unit = {
-    def operation: Config[StepsConfig] => Operation =
-      config => {
-        val zuoraRequests = ZuoraRestRequestMaker(response, config.stepsConfig.zuoraRestConfig)
+    def operation: (ZuoraRestConfig, SFAuthConfig, IdentityConfig) => Operation =
+      (zuoraRestConfig: ZuoraRestConfig,
+        sfConfig: SFAuthConfig,
+        identityConfig: IdentityConfig) => {
```
johnduffell: I'd probably just make this a normal function to avoid duplicating all the types, not sure why I did it like this originally (may be historical reasons or to avoid having to put an `_` after when calling it)

lib/handler/src/main/scala/com/gu/util/apigateway/ApiGatewayHandler.scala
```diff
@@ -65,24 +65,24 @@ object ApiGatewayHandler extends Logging {
...
+      (trustedApiConf, operation) = configOp
       apiGatewayRequest <- parseApiGatewayRequest(inputStream)
       queryParams <- apiGatewayRequest.queryParamsAsCaseClass[ApiGatewayHandlerParams]()
       response <- if (queryParams.isHealthcheck)
         ContinueProcessing(operation.healthcheck()).withLogging("healthcheck")
       else
         for {
-          _ <- isAuthorised(operation.shouldAuthenticate, queryParams.apiToken.map(RequestAuth(_)), config.trustedApiConfig)
+          _ <- isAuthorised(operation.shouldAuthenticate, queryParams.apiToken.map(RequestAuth(_)), trustedApiConf)
             .toApiGatewayContinueProcessing(unauthorized).withLogging("authentication")
```
johnduffell: so as discussed , in a future PR, this whole line in the for comprehension will probably be an optional parameter, so that non url param keyed lambdas don't even need to consider shouldAuthenticate as a concept

handlers/digital-subscription-expiry/src/main/scala/com/gu/digitalSubscriptionExpiry/Handler.scala
pvighi: I didn't but I'll check in CODE before merging.
I expect config loading to be a tiny part of the execution time of most lambdas compared to Sf or Zuora calls

lib/handler/src/main/scala/com/gu/util/config/ConfigLocation.scala
```diff
@@ -0,0 +1,3 @@
+package com.gu.util.config
+
+case class ConfigLocation[CONFIG](val path: String, version: Int)
```
johnduffell: what does val do in this context? I thought it was the default for case classes

lib/handler/src/main/scala/com/gu/util/config/LoadConfigModule.scala
```diff
@@ -18,27 +15,29 @@ object ConfigWithStage {
...
+
   val bucketName = "gu-reader-revenue-private"
 
   //we need this extra class here because otherwise we cannot partially apply the LoadConfig apply method without specifying the generic param
-  class PartialApply(stage: Stage, fetchString: GetObjectRequest => Try[String]) {
+  class PartialApply(stage: Stage, fetchString: StringFromS3) {
     def apply[CONF](implicit configLocation: ConfigLocation[CONF], reads: Reads[CONF]): ConfigFailure \/ CONF = {
```
johnduffell: as discussed we don't really have any good idea about how to fake these deps with typeclasses, something to think about (we have a similar problem but worse with the Requests (get/put/post) object)

lib/handler/src/main/scala/com/gu/util/config/ConfigLocation.scala
pvighi: i don't think it does anything, this used to be a regular class and I forgot to remove the val from there

src/main/scala/com/gu/paymentFailure/Lambda.scala
```diff
@@ -3,49 +3,52 @@ package com.gu.paymentFailure
...
-        config.etConfig.etSendIDs,
-        config.trustedApiConfig
+        etConfig.etSendIDs,
+        trustedApiConfig
       )
+    val loadConfigModule = LoadConfigModule(stage, fetchString)
+    ApiGatewayHandler(lambdaIO)(for {
+      zuoraRestConfig <- loadConfigModule[ZuoraRestConfig].toApiGatewayOp("load zuora config")
+      etconfig <- loadConfigModule[ETConfig].toApiGatewayOp("load et config")
+      trustedApiConfig <- loadConfigModule[TrustedApiConfig].toApiGatewayOp("load trusted Api config")
```
johnduffell: just a thought not a suggestion:
now that we have split the config by functionality rather than grouped together by the fact that it's configuration, I think we have a future opportunity to group the "loading zuora config" together with the "make zuora request maker" etc.  e.g. We would have a function from something like ConfigLoadModule=>ZuoraRequests 
This would mean we might be able to simplify the wiring and reduce duplication between lambdas. Something we can think about in future.

src/test/scala/com/gu/TestData.scala
```diff
@@ -30,7 +29,8 @@ object TestData extends Matchers {
   val fakeETSendIds = ETSendIds(ETSendId("11"), ETSendId("22"), ETSendId("33"), ETSendId("44"), ETSendId("can"))
   val fakeETConfig = ETConfig(etSendIDs = fakeETSendIds, "fakeClientId", "fakeClientSecret")
   val fakeStripeConfig = StripeConfig(customerSourceUpdatedWebhook = StripeWebhook(StripeSecretKey("ukCustomerSourceUpdatedSecretKey"), StripeSecretKey("auCustomerSourceUpdatedStripeSecretKey")), true)
-
+  //todo see what to do with this
```
johnduffell: I think this is the think you're looking into anyway

src/test/scala/manualTest/ConfigLoaderSystemTest.scala
```diff
@@ -1,13 +1,15 @@
...
-import com.gu.stripeCustomerSourceUpdated.SourceUpdatedSteps.StepsConfig
 import com.gu.test.EffectsTest
 import com.gu.util.config.ConfigReads.ConfigFailure
 import com.gu.util.config.{LoadConfig, Stage}
+import com.gu.util.zuora.ZuoraRestConfig
 import org.scalatest.{FlatSpec, Matchers}
+import play.api.libs.json.{Json, Reads}
 import scalaz.{\/, \/-}
 
+//TODO WE SHOULD WE FIX THIS TO USE THE NEW CONFIG LOADING CODE
```
johnduffell: I think you are looking at this already

handlers/digital-subscription-expiry/src/main/scala/com/gu/digitalSubscriptionExpiry/Handler.scala
jacobwinch: Agreed, I'm not particularly concerned - just intrigued!

lib/s3ConfigValidator/src/test/scala/com/gu/test/S3ConfigFilesEffectsTest.scala
```diff
@@ -0,0 +1,84 @@
...
+    PROD -> wireConfigLoader(PROD),
+    CODE -> wireConfigLoader(CODE)
+
+  )
+  def validate[CONF](stage: String)(implicit r: Reads[CONF], loc: ConfigLocation[CONF]) = {
+    val configLoader = configLoaders(stage)
+    val config = configLoader[CONF]
+    config.isRight shouldBe (true)
+  }
+}
```
johnduffell: great, this test does make me a bit cross eyed, but it's good!

johnduffell:
👍

cloudformation.yaml
johnduffell: if this is important do you want to add to the tech debt doc? I think this is to avoid a global allow rule making a mockery of the normal rules

PR #134, opened 2018-07-10
update reference to config files effect test

pvighi:
Thanks @johnduffell  for updating the Readme after my PR, I totally missed it. 
I didn't get a chance to review the Readme PR but I noticed that the reference to the config loading test also needed to be updated


johnduffell:
thanks didn't spot that. 👍 
I wonder if we split the `handler` lib into apigateway and config we should move this section into the config library and just link it, then we'd be more likely to think about it

PR #135, opened 2018-07-10
'cancellation-sf-cases' lambda : added support for 'test-users'

twrichards:
If identity is that of a test-user will load UAT SalesForce config instead

SalesForce cases are raised linked to subscription, therefore this lambda and members-data-api need to be referring to the correct touchpoint backend. This change brings parity with members-data-api in this regard.


handlers/cancellation-sf-cases/src/main/scala/com/gu/cancellation/sf_cases/Handler.scala
```diff
@@ -115,26 +117,41 @@ object Handler extends Logging {
...
+            case \/-(sfUatConf) => {
+              logger.info("using UAT SalesForce config")
+              SalesforceAuthenticate(response, sfUatConf)
+            }
+            case -\/(_) => SalesforceAuthenticate(response, sfConfig)
+          }
+        } else
+          SalesforceAuthenticate(response, sfConfig)
+
+      }
```
johnduffell: since the `SalesforceAuthenticate(response, *)` occurs 3 times, I would prefer this to return a config, and then connect with the authenticate in a separate bit of code.  This reduces the dependencies as this code no longer depends on SalesforceAuthenticate, so the dependency tree is flatter.

handlers/cancellation-sf-cases/src/main/scala/com/gu/identity/IsIdentityTestUser.scala
```diff
@@ -0,0 +1,25 @@
...
+import play.api.libs.json.{Json, Reads}
+
+case class IdentityTestUserConfig(secret: String)
+
+object IsIdentityTestUser {
+
+  implicit val configReads: Reads[IdentityTestUserConfig] = Json.reads[IdentityTestUserConfig]
+  implicit val configLocation = ConfigLocation[IdentityTestUserConfig](path = "identityTestUsers", version = 1)
+
+  def apply(loadConfig: LoadConfigModule.PartialApply)(identityUser: IdentityUser): Boolean =
```
johnduffell: thinking generally about what we think is needed to decide if someone is an identity test user, I'd say having a way to load any configuration is too much information.  I'd say passing in what the secret is would make more sense.  Then the top level code can have the knowledge that it happens to be in config files.

In broad terms I think the Handler.apply should know how to connect to the outside world (RawEffects)

 the Handler.runWithEffects should know that there are config files and Json is in them and other "admin" tasks, and

 the steps should know what things we need to use to actually achieve the task (there may be zero or more layers of steps but they're pretty much just connecting simple functions together)

then at the bottom there are simple functions that don't really depend on each other as much as possible, and don't know about anything above about the outside world - api gateway and aws etc.

handlers/cancellation-sf-cases/src/main/scala/com/gu/cancellation/sf_cases/Handler.scala
```diff
@@ -62,7 +63,8 @@ object Handler extends Logging {
         Subject = "Online Cancellation Attempt"
       )
 
-    def raiseCase(lookup: TSalesforceGenericIdLookup, raiseCase: RaiseCase)(identityId: String, raiseCaseDetail: RaiseRequestBody) =
+    def raiseCase(lookup: TSalesforceGenericIdLookup, raiseCase: RaiseCase)
+                 (identityId: String, raiseCaseDetail: RaiseRequestBody) =
```
johnduffell: generally I use 2 space indent and bracket on its own line for consistency with the indentation we used for braces normally.  This just means when we rename the function it won't affect the indentation, and changes/additional params don't influence the diff for other lines as much (never mind the commas)
```scala
def raiseCase(
  lookup: TSalesforceGenericIdLookup,
  raiseCase: RaiseCase
)(
  identityId: String,
  raiseCaseDetail: RaiseRequestBody
) =
```

handlers/cancellation-sf-cases/src/main/scala/com/gu/identity/IdentityCookieToIdentityUser.scala
```diff
@@ -6,17 +6,22 @@ import com.gu.util.apigateway.ApiGatewayResponse._
 import com.gu.util.config.Stage
 import com.gu.util.reader.Types._
 
-object IdentityCookieToIdentityId extends Logging {
+object IdentityCookieToIdentityUser extends Logging {
 
-  def apply(headersOption: Option[Map[String, String]], stage: Stage): ApiGatewayOp[String] =
+  case class IdentityUser(id: String, displayName: Option[String] = None)
```
johnduffell: do we need the default None?

twrichards: not anymore, will remove

handlers/cancellation-sf-cases/src/main/scala/com/gu/identity/IsIdentityTestUser.scala
```diff
@@ -0,0 +1,19 @@
...
+case class IdentityTestUserConfig(secret: String)
+
+object IsIdentityTestUser {
+
+  def apply(identityTestUsersConfig: IdentityTestUserConfig)(identityUser: IdentityUser): Boolean =
+    (for {
+      displayName <- identityUser.displayName
+      firstName <- displayName.split(' ').headOption
+      testUsernames = TestUsernames(Encoder.withSecret(identityTestUsersConfig.secret), Duration.ofDays(2))
+    } yield testUsernames.isValid(firstName)).getOrElse(false)
```
johnduffell: if you think it's clearer, it might be worth making the last line 16 of the for comprehension into part of the yield (since it's only an `=`) The yield statement can be a block with braces just as a normal bit of code.

johnduffell:
just needs some testing and then it's good to go 👍 
I'd recommend even if there's no real logic to unit test, you need a single "EndToEnd" test calling runWithEffects with fake side effects passed in.  Then you can do an assertion that the expected http request is made to raise the case.  That gives a simple check that anyone else refactoring hasn't broken the fundamentals.
@pvighi  added a test that loads each config file separately but I think you added that before his changes so it should be there?

johnduffell:
here is the config loading test: https://github.com/guardian/zuora-auto-cancel/blob/master/lib/s3ConfigValidator/src/test/scala/com/gu/test/S3ConfigFilesEffectsTest.scala
and here is an example EndToEnd non effects test..
this one is relatively complicated as it's checking that it doesn't authenticate with SF at the wrong time:
https://github.com/guardian/zuora-auto-cancel/blob/master/handlers/identity-backfill/src/test/scala/com/gu/identityBackfill/EndToEndHandlerTest.scala

handlers/cancellation-sf-cases/src/main/scala/com/gu/identity/IsIdentityTestUser.scala
twrichards: personally i don't think that will be any cleaner/clearer, but good to know that yield can be a block for future reference

README.md
```diff
@@ -49,13 +49,16 @@ These tests do not (YET!) run automatically.
...
-This would need suitable AWS credentials to get hold of the config as well as config file under .aws with the content:
+This would need suitable AWS credentials to get hold of the config as well as file called `config` under `~/.aws` directory with the content:
 ```
 [default]
 region = eu-west-1
 ```
 You can tag your integration new tests with `taggedAs EffectsTest`.  This ignores them from the standard `sbt test`.
 
+NOTE: You may notice that IntelliJ doesn't compile the effects tests on-the-fly (as it does with source files) you can re-enable this behaviour by commenting out the contents of the `Seq(...)` at the end of the `val testSettings` assignent in the top-level `build.sbt` .
+This is also means you can run them with the IntelliJ debugger :)
```
johnduffell: 🏅

handlers/cancellation-sf-cases/src/main/scala/com/gu/cancellation/sf_cases/Handler.scala
```diff
@@ -79,25 +86,24 @@ object Handler extends Logging {
...
-    )(
-      sfRequestsFollowingIdentityCheck: IdentityUser => ApiGatewayOp[RestRequestMaker.Requests]
-    )(
-      apiGatewayRequest: ApiGatewayRequest
-    ) =
+    def steps(headersToIdentityAndSfRequests: HeadersToIdentityAndSfRequests)(apiGatewayRequest: ApiGatewayRequest) =
       (for {
-        identityUser <- IdentityCookieToIdentityUser(apiGatewayRequest.headers, stage)
-        sfRequests <- sfRequestsFollowingIdentityCheck(identityUser)
+        identityAndSfRequests <- headersToIdentityAndSfRequests(apiGatewayRequest.headers)
```
johnduffell: from this naming it's not totally clear why the headers affects the sf requests, but I can't think of a better name
not sure if you prefer something like `sfBackendForIdentityCookie` ? if not leave as it is

handlers/cancellation-sf-cases/src/main/scala/com/gu/cancellation/sf_cases/Handler.scala
johnduffell: this comment has been resolved, thanks

handlers/cancellation-sf-cases/src/main/scala/com/gu/salesforce/cases/SalesforceCase.scala
```diff
@@ -40,4 +40,10 @@ object SalesforceCase extends Logging {
 
   }
 
+  object GetById {
+
+    def apply(sfRequests: Requests)(caseId: String): ClientFailableOp[JsValue] =
```
johnduffell: by returning the JsValue and passing that on to the API gateway client,, we are simply returning all the fields to the client.  This is a sensible idea, but only if we are sure it's not going to reveal anything sensitive that also might come back.  Just want to confirm that's the case.

handlers/cancellation-sf-cases/src/test/scala/com/gu/cancellation/sf_cases/EndToEndHandlerEffectsTest.scala
```diff
@@ -0,0 +1,148 @@
...
+       |        "httpMethod": "${httpMethod}"
+       |    },
+       |    "body": "${bodyString}",
+       |    "isBase64Encoded": false
+       |}
+    """.stripMargin
+
+  val createCaseRequest: String = ApiGatewayRequestPayloadBuilder(
+    httpMethod = "POST",
+    bodyString = "{\\\"reason\\\":\\\"mma_editorial\\\",\\\"product\\\":\\\"Membership\\\",\\\"subscriptionName\\\":\\\"A-S00045062\\\"}"
```
johnduffell: you can use triple quotes here to get away from the backslashes

johnduffell:
great, just a few observations that may be useful plus some simplifications possible, the only thing I don't think we should put code in the Jar that's just for the tests.
further observations - It might be worth splitting the Handler down a bit at some point - at the moment there is getting to be a lot in the same file.
re coverage -  I guess you increased the amount of code therefore making the percentage lower.  This would be resolved when you add some non-effects tests.
thanks for going through this 👍

handlers/cancellation-sf-cases/src/test/scala/com/gu/cancellation/sf_cases/EndToEndHandlerEffectsTest.scala
twrichards: is actually needed in this case - we're not quite sure why

PR #137, opened 2018-07-11
stop us building a zuora query with an empty OR

johnduffell:
It turned out the query builder I wrote the other week handled OR by generating an empty string if there are no clauses in the OR.  Call it lack of testing.

zuora was just kicking out the query as the WHERE clause was empty.
```
{ "queryString": "SELECT WorkEmail FROM Contact WHERE" }
```
"You have an error in your ZOQL syntax"

It doesn't really make sense to have an OR clause with no clauses inside (unless you're a philosopher, in which case it means the condition is never satisfied)

I thought of a couple of approaches, from just trusting the input and putting in a 1=0 default clause (zuora won't accept antthing like that unfortauntely) to getting the individual operations to filter it (more boiler plate to remember) and in the end I decided to try making it a compile error (OrTraverse takes a non empty list)

I wasn't sure whether to use the scalaz one, so I just used the normal list.  The drawback is that the standard operators like to upcast to a normal list at every opportunity.

See what you think, the result is below.

@jacobwinch @pvighi @paulbrown1982 @twrichards

### DIFF
--- a/handlers/identity-retention/src/main/scala/com/gu/identityRetention/GetActiveZuoraAccounts.scala
+++ b/handlers/identity-retention/src/main/scala/com/gu/identityRetention/GetActiveZuoraAccounts.scala
@@ -9,12 +9,12 @@ import com.gu.util.zuora.ZuoraQuery.{QueryResult, ZuoraQuerier}
 import play.api.libs.json.Json
 import scalaz.{-\/, \/-}
 
-object HasActiveZuoraAccounts {
+object GetActiveZuoraAccounts {
 
   case class IdentityQueryResponse(Id: String)
   implicit val reads = Json.reads[IdentityQueryResponse]
 
-  def apply(identityId: IdentityId, zuoraQuerier: ZuoraQuerier): ApiGatewayOp[List[AccountId]] = {
+  def apply(zuoraQuerier: ZuoraQuerier)(identityId: IdentityId): ApiGatewayOp[List[AccountId]] = {
 
     def searchForAccounts = {
 
--- a/handlers/identity-retention/src/main/scala/com/gu/identityRetention/SubscriptionsForAccounts.scala
+++ b/handlers/identity-retention/src/main/scala/com/gu/identityRetention/SubscriptionsForAccounts.scala
@@ -20,7 +20,7 @@ object SubscriptionsForAccounts {
 
   implicit val reads = Json.reads[SubscriptionsQueryResponse]
 
-  def buildQuery(activeAccounts: List[AccountId]): ClientFailableOp[SafeQuery] =
+  def buildQuery(activeAccounts: ::[AccountId]): ClientFailableOp[SafeQuery] =
     for {
       or <- OrTraverse(activeAccounts) { acc =>
         zoql"""
@@ -38,7 +38,7 @@ object SubscriptionsForAccounts {
         """
     } yield subscriptionsQuery
 
-  def apply(zuoraQuerier: ZuoraQuerier)(activeAccounts: List[AccountId]): ApiGatewayOp[List[SubscriptionsQueryResponse]] = {
+  def apply(zuoraQuerier: ZuoraQuerier)(activeAccounts: ::[AccountId]): ApiGatewayOp[List[SubscriptionsQueryResponse]] = {
 
     def searchForSubscriptions =
       for {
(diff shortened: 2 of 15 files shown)


handlers/identity-retention/src/main/scala/com/gu/identityRetention/SubscriptionsForAccounts.scala
```diff
@@ -20,7 +20,7 @@ object SubscriptionsForAccounts {
 
   implicit val reads = Json.reads[SubscriptionsQueryResponse]
 
-  def buildQuery(activeAccounts: List[AccountId]): ClientFailableOp[SafeQuery] =
+  def buildQuery(activeAccounts: ::[AccountId]): ClientFailableOp[SafeQuery] =
```
johnduffell: a list can either be Nil or ::, so this just forces the front of the list to be ::.  Looks a bit weird, we could use scalaz or similar if people prefer.

handlers/sf-contact-merge/src/test/scala/com/gu/sf_contact_merge/GetContactsTest.scala
```diff
@@ -14,7 +15,7 @@ class GetContactsTest extends FlatSpec with Matchers {
   it should "work" in {
 
     val getContacts = GetZuoraEmailsForAccounts.GetContacts(ZuoraQuery(ZuoraRestRequestMaker(mock.response, ZuoraRestConfig("http://server", "user", "pass"))))_
-    val actual = getContacts(List("2c92c0f9624bbc5f016253e573970b16", "2c92c0f8644618e30164652a558c6e20").map(AccountId.apply))
+    val actual = getContacts(ToNel.literal(AccountId("2c92c0f9624bbc5f016253e573970b16"), AccountId("2c92c0f8644618e30164652a558c6e20")))
```
johnduffell: if you map a :: it turns back into a List, which loses the knowledge that it has something in.

handlers/sf-contact-merge/src/main/scala/com/gu/sf_contact_merge/Handler.scala
```diff
@@ -50,10 +51,11 @@ object Handler {
   )
   implicit val readWireSfContactRequest = Json.reads[WireSfContactRequest]
 
-  def steps(getZuoraEmails: List[AccountId] => ClientFailableOp[List[Option[EmailAddress]]])(req: ApiGatewayRequest): ResponseModels.ApiResponse =
+  def steps(getZuoraEmails: ::[AccountId] => ClientFailableOp[List[Option[EmailAddress]]])(req: ApiGatewayRequest): ResponseModels.ApiResponse =
```
paulbrown1982: I think :: is should say what it actually is, a NonEmptyList.

lib/zuora/src/main/scala/com/gu/util/zuora/SafeQueryBuilder.scala
```diff
@@ -64,9 +64,20 @@ object SafeQueryBuilder {
 
   }
 
+  object ToNel {
```
paulbrown1982: Also I think you should rename this to ToNonEmptyList.

handlers/sf-contact-merge/src/main/scala/com/gu/sf_contact_merge/Handler.scala
johnduffell: ok I have used the proper scalaz one which makes it clearer (although more scalaz buy in)

lib/zuora/src/main/scala/com/gu/util/zuora/SafeQueryBuilder.scala
johnduffell: I called it MaybeNonEmptyList as it would be an Option

johnduffell:
thanks for looking, I have made the changes, feel free to check again and I will merge tomorrow.

handlers/identity-retention/src/main/scala/com/gu/identityRetention/GetActiveZuoraAccounts.scala
```diff
@@ -9,12 +9,12 @@ import com.gu.util.zuora.ZuoraQuery.{QueryResult, ZuoraQuerier}
 import play.api.libs.json.Json
 import scalaz.{-\/, \/-}
 
-object HasActiveZuoraAccounts {
+object GetActiveZuoraAccounts {
```
johnduffell: I just did a bit of opportunistic renaming some things I noticed

PR #138, opened 2018-07-12
add zuora contact update logic to sf contact merge lambda

johnduffell:
This adds the final step to the contact updating.

The overall piece of work is to merge together several SF contacts where they have the same email address etc.  This will help the call centre to find the reader, and also help with retention of data where we still have a need to keep it.

The matching logic happens in a salesforce custom report, then this lambda will take the list of IDs, check that they really do all have the same email address, and then update them all to have the supplied SF contact and account ids.

Note that these IDs are not validated by the lambda, even to check the format.  Not sure if that's classed as a risk?

@pvighi @paulbrown1982 @abhichawla @jacobwinch please comment!

### DIFF
--- a/handlers/sf-contact-merge/src/test/scala/com/gu/sf_contact_merge/UpdateAccountSFLinksTest.scala
+++ b/handlers/sf-contact-merge/src/test/scala/com/gu/sf_contact_merge/UpdateAccountSFLinksTest.scala
@@ -1,0 +1,38 @@
+package com.gu.sf_contact_merge
+
+import com.gu.sf_contact_merge.GetZuoraEmailsForAccounts.AccountId
+import com.gu.zuora.fake.FakeRequestsPut
+import org.scalatest.{FlatSpec, Matchers}
+import play.api.libs.json._
+import scalaz.\/-
+
+class UpdateAccountSFLinksTest extends FlatSpec with Matchers {
+
+  it should "handle an email and a missing email with a fake querier" in {
+
+    val expectedUrl = """accounts/1234"""
+
+    val response = JsObject(Map("Success" -> JsBoolean(true)))
+    val expectedInput = Json.parse(
+      """
+        |{
+        |    "sfContactId__c": "johnjohn_c",
+        |    "crmId": "crmIdjohn"
+        |    }
+      """.stripMargin
+    )
+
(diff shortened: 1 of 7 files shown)


handlers/sf-contact-merge/src/test/scala/com/gu/sf_contact_merge/UpdateAccountSFLinksTest.scala
```diff
@@ -0,0 +1,38 @@
...
+    val expectedInput = Json.parse(
+      """
+        |{
+        |    "sfContactId__c": "johnjohn_c",
+        |    "crmId": "crmIdjohn"
+        |    }
+      """.stripMargin
+    )
+
+    val (requestsMade, fake) = FakeRequestsPut(expectedUrl, expectedInput, response)
```
pvighi: maybe fake could be called something more descriptive here, like fakePutRequest or something

handlers/sf-contact-merge/src/test/scala/com/gu/zuora/fake/FakeRequestsPut.scala
```diff
@@ -0,0 +1,22 @@
...
+
+import com.gu.util.zuora.RestRequestMaker.{ClientFailableOp, GenericError, RequestsPUT}
+import play.api.libs.json.{JsValue, Json, Reads, Writes}
+import scalaz.-\/
+import scalaz.syntax.std.either._
+
+object FakeRequestsPut {
+  def apply(expectedUrl: String, expectedInput: JsValue, response: JsValue): (() => List[JsValue], RequestsPUT) = {
+    var requestsMade: List[JsValue] = List()
+    (() => requestsMade, new RequestsPUT() {
```
pvighi: maybe we could declare the new RequestsPUT() {... bit outside instead of nesting the definition here. It might me easier to read

handlers/sf-contact-merge/src/test/scala/com/gu/sf_contact_merge/UpdateAccountSFLinksTest.scala
johnduffell: good idea, will do, fakePutter

johnduffell:
tested in code, it updated a pair of accounts correctly! merged to prod.
error handling may not be super explanatory but it should all end up in the logs.

PR #139, opened 2018-07-12
Csr acquisitions endpoint scaffolding

pvighi:
also some minor refactoring to avoid duplication


lib/handler/src/main/scala/com/gu/util/apigateway/ApiGatewayResponse.scala
```diff
@@ -49,6 +49,11 @@ object ApiGatewayResponse extends Logging {
 
   def toJsonBody(responseBody: ResponseBody) = Json.prettyPrint(Json.toJson(responseBody))
 
+  def apply[B](statusCode: String, body: B)(implicit writes: Writes[B]) = {
+    val bodyTxt = Json.prettyPrint(Json.toJson(body))
+    ApiResponse(statusCode, bodyTxt)
+  }
```
pvighi: I was going to duplicate this code and I thought that It might be useful to provide a generic way of creating an ApiGateway response with a specific case class as the body. I don't think we had a way to do this but I might be forgetting about something.
Anyway, this is what causes most the random refactors around the code

handlers/new-product-api/README.md
```diff
@@ -0,0 +1 @@
+# add-product-api
```
johnduffell: TODO

handlers/new-product-api/build.sbt
```diff
@@ -0,0 +1,24 @@
...
+
+libraryDependencies ++= Seq(
+  "com.amazonaws" % "aws-lambda-java-log4j" % "1.0.0",
+  "com.amazonaws" % "aws-lambda-java-core" % "1.2.0",
+  "com.amazonaws" % "aws-java-sdk-s3" % "1.11.311",
+  "log4j" % "log4j" % "1.2.17",
+  "com.squareup.okhttp3" % "okhttp" % "3.9.1",
+  "com.typesafe.play" %% "play-json" % "2.6.9",
+  "org.scalatest" %% "scalatest" % "3.0.1" % "test",
+  "com.fasterxml.jackson.core" % "jackson-databind" % "2.8.11.1"
```
johnduffell: a lot of these are in Dependencies.scala as vals.  Also sbt gives you them transitively, so you can almost certainly get rid of all of them.  See sf-contact-merge for the latest (I didn't go back refactoring the rest...)

handlers/new-product-api/cfn.yaml
```diff
@@ -0,0 +1,226 @@
...
+
+Mappings:
+    StageMap:
+        CODE:
+            ApiName: new-product-api-CODE
+            DomainName: new-product-api-code.membership.guardianapis.com
+            ApiGatewayTargetDomainName: d-ecyddyj7nk.execute-api.eu-west-1.amazonaws.com
+        PROD:
+            ApiName: new-product-api-PROD
+            DomainName: new-product-api-prod.membership.guardianapis.com
```
johnduffell: fyi in sf-contact-merge I got rid of the ApiName and DomainName and just built it from the stage, not sure if it's better

handlers/new-product-api/cfn.yaml
```diff
@@ -0,0 +1,226 @@
...
+    AddSubscriptionLambda:
+        Type: AWS::Lambda::Function
+        Properties:
+            Description: add subscription to account
+            FunctionName:
+                !Sub add-subscription-${Stage}
+            Code:
+                S3Bucket: zuora-auto-cancel-dist
+                S3Key: !Sub membership/${Stage}/new-product-api/new-product-api.jar
+            Handler: com.gu.newProductApi.addSubscription.Handler::apply
```
johnduffell: after tom did it and I checked the coding standards, I have started using underscores in the package names (although that just makes the issue of different kinds of names worse for if we do templating)

handlers/new-product-api/cfn.yaml
```diff
@@ -0,0 +1,226 @@
...
+      Properties:
+        UsagePlanName: !Sub new-product-api-usage-plan-${Stage}
+        ApiStages:
+        - ApiId: !Ref NewProductApi
+          Stage: !Ref NewProductApiStage
+      DependsOn:
+      - NewProductApi
+      - NewProductApiStage
+
+    NewProductUsagePlankey:
```
johnduffell: minor but I don't know if you intended upper case Key instead of key

handlers/new-product-api/src/main/scala/com/gu/newProductApi/addSubscription/AddSubscriptionRequest.scala
```diff
@@ -0,0 +1,49 @@
...
+
+import scala.util.{Failure, Success, Try}
+
+case class AddSubscriptionRequest(
+                                   zuoraAccountId: String,
+                                   contractEffectiveDate: LocalDate,
+                                   acquisitionSource: String,
+                                   createdByCSR: String,
+                                   amountMinorUnits: Int,
+                                 )
```
johnduffell: intellij went a bit generous on the indentation there!

handlers/new-product-api/src/main/scala/com/gu/newProductApi/addSubscription/AddSubscriptionRequest.scala
```diff
@@ -0,0 +1,49 @@
...
+
+
+object AddSubscriptionRequest {
+
+  case class AddSubscriptionRequestWire(
+                                         zuoraAccountId: String,
+                                         contractEffectiveDate: String,
```
johnduffell: at this stage if we want to pass in the start date I'm happy to do that even though I think it's not needed, but I don't think we should use the zuora terminology of contractEffectiveDate, we should use the customer terminology (startDate would be fine with me)

handlers/new-product-api/src/main/scala/com/gu/newProductApi/addSubscription/AddSubscriptionRequest.scala
```diff
@@ -0,0 +1,49 @@
...
+                                         acquisitionSource: String,
+                                         createdByCSR: String,
+                                         amountMinorUnits: Int,
+                                       ) {
+    def toAddSubscriptionRequest = {
+
+
+      val maybeParsedRequest = Try(LocalDate.parse(contractEffectiveDate)).map { parsedEffectiveDate =>
```
johnduffell: I was going to suggest making a simple reads for LocalDate as it's not doing anything custom to us (that we might tesT), but I guess it's likely we will add custom wrappers for some of the strings later

handlers/new-product-api/src/main/scala/com/gu/newProductApi/addSubscription/AddSubscriptionResponse.scala
```diff
@@ -0,0 +1,25 @@
...
+}
+
+object AddSubscriptionResponse {
+  implicit val writes: Writes[AddSubscriptionResponse] = { response =>
+    response match {
+      case e: Error => Error.writes.writes(e)
+      case s: AddedSubscription => AddedSubscription.writes.writes(s)
+    }
+
+  }
```
johnduffell: maybe the SAM syntax won't allow it, but if it was a function you could just 
```scala
implicit val writes: Writes[AddSubscriptionResponse] = _ match {
  case e: Error => Error.writes.writes(e)
  case s: AddedSubscription => AddedSubscription.writes.writes(s)
}
```
and the `_ match ` might even be optional.  Would be neat, but may not work!

handlers/new-product-api/src/main/scala/com/gu/newProductApi/addSubscription/Handler.scala
```diff
@@ -0,0 +1,49 @@
...
+  def addSubscriptionSteps(apiGatewayRequest: ApiGatewayRequest): ApiResponse = {
+    (for {
+      request <- apiGatewayRequest.bodyAsCaseClass[AddSubscriptionRequest]()
+      _ = logger.info(s"parsed request as $request")
+    } yield ApiGatewayResponse(body = AddedSubscription("A-S00045523"), statusCode = "200")).apiResponse
+  }
+
+  def runWithEffects(response: Request => Response, stage: Stage, fetchString: StringFromS3, lambdaIO: LambdaIO) = {
+
+    def operation: ZuoraRestConfig => Operation = zuoraRestConfig => {
```
johnduffell:
```scala
def operation(zuoraRestConfig: ZuoraRestConfig): Operation = {
```
(I think that weird sytax crept in after I refactored something part of the way)

johnduffell: or feel free to inline this operation function into the for comprehension below

lib/handler/src/main/scala/com/gu/util/apigateway/ApiGatewayResponse.scala
johnduffell: noticed that, sounds like a good idea, good idea to do a quick tidyup 🥇

johnduffell:
great, the only blocker for me is the date (we can discuss at standup if necessary)
the others are just for guidance!

handlers/new-product-api/cfn.yaml
pvighi: where did you find that? I remember seeing that and I searching on the internet but all I found was [this](https://docs.scala-lang.org/style/naming-conventions.html#packages) that just says that java package names should be followed and the [java doc](https://docs.oracle.com/javase/tutorial/java/package/namingpkgs.html) that just says to write all words in lower case without any separator for different words

handlers/new-product-api/cfn.yaml
```diff
@@ -0,0 +1,226 @@
...
+            Value: !FindInMap [StageMap, !Ref Stage, ApiName]
+          - Name: Stage
+            Value: !Sub ${Stage}
+        EvaluationPeriods: 1
+        MetricName: 5XXError
+        Namespace: AWS/ApiGateway
+        Period: 3600
+        Statistic: Sum
+        Threshold: 5
+        TreatMissingData: notBreaching
```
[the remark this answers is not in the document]
pvighi: 👍 added the additional alarm down there ⬇️ for some reason Github didn't hide this comment

handlers/new-product-api/cfn.yaml
johnduffell: actually as you pointed out I didn't - the only way to do it would be to use a parameter with a default value, which may be a bit too strange!

PR #140, opened 2018-07-13
added api gateway target domain name for prod new product api

pvighi:
adding the domain name that can't be referenced directly from cloudformation


johnduffell:
👍

johnduffell:
if you can, make sure this is up to date:
https://github.com/guardian/zuora-auto-cancel/blob/master/handlers/HOWTO-create-lambda.md

PR #141, opened 2018-07-13
remove all the duplicated deps from the handlers

johnduffell:
@pvighi and I were discussing about how we did a lot of improvements to the latest lambdas, but due to the high level of duplication we often don't go back and apply the same changes elsewhere.
Of course it would be nice to properly reduce that duplication, so I decided to make a start with one that we mentioned.

I tried clean compile assembly and it seemed to work and run all the tests.  Once this branch builds I'll deploy to CODE. and try some.

@paulbrown1982 @jacobwinch


johnduffell:
thanks I will ship this on monday
Also, I will look to reduce the build.sbt needed for a new lambda to as close as possible to the name only.

PR #142, opened 2018-07-14
use a nicely named type for http client ops rather than scalaz.\/

johnduffell:
we previously turned ApiGatewayOp into its own type and it seemed to help a lot with naming as `\/` was confusing us.
I was having a think about the Op we use for the rest http client, and I thought that probably would be clearer not as a `\/` as well.
There seems to be a bit more type adjusting needed, some places we have to convert from this to a disjunction and then from disjunction to api gateway op.
I actually started to do this as a bit of thinking about how to augment an http request maker without passing it in as a function (which makes it hard to unit test - needs a var, and is unclear)  And also how to string them together where there are multiple requests.  What we technically need is a profunctor and a monoid in that profunctor, but I'd rather not use those words in the code base, so maybe we can have a clearer version.

this is the interesting bit:
https://github.com/guardian/zuora-auto-cancel/pull/142/files#diff-b9a4f68ade1549df685346a63f683230

@paulbrown1982 @pvighi @jacobwinch @twrichards thoughts? worth merging? or better to stick with the generic type?


handlers/zuora-retention/src/main/scala/com/gu/zuora/retention/updateAccounts/Handler.scala
```diff
@@ -63,7 +64,7 @@ object Handler {
 
     def wiredOperation(updateAccountsRequest: UpdateAccountsRequest): Try[UpdateAccountsResponse] = for {
       zuoraRequests <- getZuoraRequestMaker(RawEffects.response, RawEffects.stage, GetFromS3.fetchString)
-      setDoNotProcess = SetDoNotProcess(zuoraRequests.put) _
+      setDoNotProcess = SetDoNotProcess(zuoraRequests.put[UpdateRequestBody, Unit]) _
```
johnduffell: this started complaining about the implicit reads for `Nothing` being ambigous so I added the specific types

handlers/zuora-retention/src/test/scala/com/gu/zuora/retention/updateAccounts/UpdateAccountsTest.scala
```diff
@@ -45,7 +44,8 @@ class UpdateAccountsTest extends FlatSpec with Matchers {
     def getRemainingTime() = remainingMsValues.next()
 
     val response = UpdateAccounts(fakeZuoraUpdate, getRemainingTime _)(testUri, AccountIdIterator(linesIterator, 0).get)
-    response shouldBe Failure(LambdaException("error response from lambda -\\/(GenericError(something failed!))"))
+    // this test is a bit specific, no need to check the exact message
+    response shouldBe Failure(LambdaException("error response from lambda GenericError(something failed!)"))
```
johnduffell: this test broke due to checking the exact string, this isn't actually necessary in such detail

lib/zuora-reports/src/test/scala/com/gu/zuora/reports/GetJobResultTest.scala
```diff
@@ -95,6 +95,7 @@ class GetJobResultTest extends AsyncFlatSpec {
     def get(path: String) = responseWithMissingFileId
     val jobResultRequest = JobResultRequest("someJobId", false, None)
 
-    GetJobResult(get)(jobResultRequest) shouldBe -\/(GenericError("file Id missing from response : \\/-(AquaJobResponse(completed,testResponse,List(Batch(completed,batch1,Some(fileId1)), Batch(completed,batch2,None)),None))"))
+    // test is too specific, shoudlnt' check so much detail
+    GetJobResult(get)(jobResultRequest) shouldBe GenericError("file Id missing from response : ClientSuccess(AquaJobResponse(completed,testResponse,List(Batch(completed,batch1,Some(fileId1)), Batch(completed,batch2,None)),None))")
```
johnduffell: this was another test that checked the string exactly, probably don't need to check the whole thing

lib/restHttp/src/main/scala/com/gu/util/zuora/RestRequestMaker.scala
```diff
@@ -4,41 +4,110 @@ import java.io.InputStream
...
 
 object RestRequestMaker extends Logging {
 
-  sealed trait ClientFail {
+  import Types._
+
+  sealed trait ClientFail extends ClientFailableOp[Nothing] {
+    override def toDisjunction: ClientFail \/ Nothing = -\/(this)
+
+    val isFail = true
```
jacobwinch: very minor point but I'd call these _ClientFailure_ and _isFailure_

lib/restHttp/src/main/scala/com/gu/util/zuora/RestRequestMaker.scala
```diff
@@ -4,41 +4,110 @@ import java.io.InputStream
...
+  }
+
+  object Types {
+
+    implicit class UnderlyingOps[A](theEither: scalaz.\/[ClientFail, A]) {
+
+      def toClientFailableOp: ClientFailableOp[A] =
+        theEither match {
+          case scalaz.\/-(success) => ClientSuccess(success)
+          case scalaz.-\/(finished) => finished
```
pvighi: wouldn't a left in this context mean failure rather than finished ?

pvighi:
I think the benefits are less clear in this case compared to ApiGatewayOp refactor because the disjunction is being used in the expected way: left=error right=success. But it still makes it more explicit.
what do you think @jacobwinch ?

johnduffell:
for the  record I chatted a bit with Jacob about this and suggested since this change is primarily naming related rather than structure, it should be judged more heavily on what happens when a new person joins, perhaps with no scala experience, and try to do a typical bit of work like adding a new endpoint that triggers off a callout, does a couple of REST calls and returns a suitable result.
The extra cruft in the for comprehensions was a negative point to offset the naming improvement.  The custom one off logic in the file I linked above less so, but still needs to be understandable.

handlers/cancellation-sf-cases/src/main/scala/com/gu/cancellation/sf_cases/TypeConvert.scala
```diff
@@ -0,0 +1,12 @@
...
+import com.gu.util.reader.Types._
+import com.gu.util.zuora.RestRequestMaker.ClientFailableOp
+
+object TypeConvert {
+
+  implicit class TypeConvertClientOp[A](theEither: ClientFailableOp[A]) {
+    def toApiGatewayOp = theEither.toDisjunction.toApiGatewayOp(_)
+  }
+
+}
```
johnduffell: @paulbrown1982 suggested that it was better to keep the for comprehensions neat.
this code needs to go somewhere that depends on both restHttp and apigatewayHandler packages, but I feel this is too small to justify another lib!  So this file is just duplicated in all the handlers. see what you think.

handlers/identity-retention/src/main/scala/com/gu/identityRetention/IdentityRetentionSteps.scala
```diff
@@ -42,7 +42,7 @@ object IdentityRetentionSteps extends Logging {
     Try(input.toLong) match {
       case Success(validUserId) => Some(IdentityId(validUserId.toString))
       case Failure(ex) =>
-        logger.error(s"Invalid identity id provided", ex)
+        logger.info(s"Invalid identity id provided: $input")
```
jacobwinch: Did you intend to downgrade this to info? (If so, why?)

lib/zuora/src/main/scala/com/gu/util/zuora/ZuoraRestRequestMaker.scala
```diff
@@ -25,15 +25,18 @@ object ZuoraRestRequestMaker extends Logging {
 
     bodyAsJson.validate[ZuoraResponse] match {
       case JsSuccess(ZuoraSuccess, _) =>
-        ().right
+        ClientSuccess(())
       case JsSuccess(ZuoraErrorResponse(reasons), _) => {
         logger.error(s"Zuora rejected our call $bodyAsJson")
-        if (reasons.exists(_.code.toString.endsWith("40"))) NotFound("Received a 'not found' response from Zuora").left else GenericError("Received a failure result from Zuora").left
+        if (reasons.exists(_.code.toString.endsWith("40")))
```
jacobwinch: 📈 📖 👍

jacobwinch:
Overall I think this makes the codebase easier to understand, so 👍 from me

handlers/identity-retention/src/main/scala/com/gu/identityRetention/IdentityRetentionSteps.scala
johnduffell: I changed it from the whole stack trace because it was logging it in the test and scaring me
no idea why I changed to info so good spot (ideally we should return errors, and error log at that point, but I wasn't planning to make that kind of change) I will change back!

johnduffell:
thanks for looking @jacobwinch  and suggestions from @paulbrown1982 , I will merge once the build passes!

PR #143, opened 2018-07-16
added payment method details api call

pvighi:


handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/zuora/GetPaymentMethodDetails.scala
```diff
@@ -0,0 +1,28 @@
+package com.gu.newproduct.api.addsubscription.zuora
+
+import com.gu.util.zuora.RestRequestMaker.{ClientFailableOp, RequestGet}
+import play.api.libs.json.{JsError, JsSuccess, Json, Reads}
+
```
pvighi: If we will never use anything other than the paymentMethodStatus from this call we could rename it getPaymentMethodStatus but we might use the payment failure information

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/zuora/GetPaymentMethodDetails.scala
```diff
@@ -0,0 +1,28 @@
...
+object GetPaymentMethodDetails {
+
+  sealed trait PaymentMethodStatus
+  object Active extends PaymentMethodStatus
+  object Closed extends PaymentMethodStatus
+  case class PaymentMethodDetails(status: PaymentMethodStatus)
```
pvighi: I originally thought that there were more possible status values. it turns out it's just Active or Closed ( see [zuora doc](https://www.zuora.com/developer/api-reference/#operation/Object_GETPaymentMethod)). So I could just deserialise this into Active:Boolean or PaymentMethodStatusActive : Boolean to avoid confusion with the Active boolean field in the api response that sometimes is false even if PaymentMethodStatus = "Active"



```

File: handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/zuora/GetPaymentMethodDetails.scala (hunk 1 above)
Comment (johnduffell) [reply handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/zuora/GetPaymentMethodDetails.scala]: I'm happier to make it specific, not a big deal eitther way though

File: handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/zuora/GetPaymentMethodDetails.scala [hunk 3]
```diff
@@ -0,0 +1,28 @@
...
+
+
+  object PaymentMethodDetails {
+    case class PaymentMethodWire(PaymentMethodStatus: String)
+    val wireReads = Json.reads[PaymentMethodWire]
+    implicit val reads : Reads[PaymentMethodDetails] = x => wireReads.reads(x) flatMap {
+      case PaymentMethodWire("Active") => JsSuccess(PaymentMethodDetails(Active))
+      case PaymentMethodWire("Closed") => JsSuccess(PaymentMethodDetails(Closed))
+      case PaymentMethodWire(unknown) => JsError(s"unknown payment method status: '$unknown'")
```
johnduffell: would be nice to split this into at least a def, to separate out that weird flatMap stuff, so people can understand the conversion on its own

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/zuora/GetPaymentMethodDetails.scala
johnduffell: I like the descriptive name, boolean is very prone to negative logic and backwardsness

handlers/new-product-api/src/test/scala/com/gu/newproduct/api/addsubscription/zuora/GetPaymentDetailsEffectsTest.scala
```diff
@@ -0,0 +1,24 @@
...
+class GetPaymentDetailsEffectsTest extends FlatSpec with Matchers {
+
+  it should "get payment details" taggedAs EffectsTest in {
+    val actual = for {
+      zuoraRestConfig <- LoadConfigModule(Stage("DEV"), GetFromS3.fetchString)[ZuoraRestConfig]
+      zuoraDeps = ZuoraRestRequestMaker(RawEffects.response, zuoraRestConfig)
+      res <- GetPaymentMethodDetails(zuoraDeps)("2c92c0f8649cc8a60164a2bfd475000c")
+    } yield {
+      res
+    }
```
johnduffell: tiny (and probably a copy and paste) but this could be simply `} yield res`

lib/restHttp/src/main/scala/com/gu/util/zuora/RestRequestMaker.scala
```diff
@@ -49,23 +49,27 @@ object RestRequestMaker extends Logging {
     def put[REQ: Writes, RESP: Reads](req: REQ, path: String): ClientFailableOp[RESP]
   }
 
+  trait RequestGet {
+    def get[RESP: Reads](path: String, skipCheck: Boolean = false): ClientFailableOp[RESP]
+  }
```
johnduffell: good idea adding this specific type

johnduffell:
great! 👍 🚀

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/zuora/GetPaymentMethodStatus.scala
```diff
@@ -0,0 +1,27 @@
...
+      case "Active" => \/-(Active)
+      case "Closed" => \/-(Closed)
+      case unknown => -\/(GenericError(s"Unknown payment method status: '$unknown'"))
+    }
+  }
+  implicit val wireReads = Json.reads[PaymentMethodWire]
+
+  //todo use john's case class for the payment method id
+  def apply(get: (String, Boolean) => ClientFailableOp[PaymentMethodWire])(paymentMethodId: String): ClientFailableOp[PaymentMethodStatus] =
+    get(s"object/payment-method/$paymentMethodId", true).flatMap(_.toPayMentMethodStatus)
```
pvighi: it is a little annoying that i cannot do skipCheck=true here because the get params have no name defined

johnduffell: good point, and to name the parameters we would need a trait or case class.  With the RequestsPUT it gave us that but the extra type parameters were also not applied.  not sure the solution for this.  Would we be better to not use a boolean, but make our own - a sealed trait Check with objects WithCheck and WithoutCheck.  Then at least we won't just see "true"

handlers/new-product-api/src/test/scala/com/gu/newproduct/api/addsubscription/zuora/GetPaymentMethodStatusTest.scala
```diff
@@ -0,0 +1,48 @@
...
+
+  def fakeGet(path:String, skipCheck: Boolean) = \/-(fakeResponses(path))
+
+  it should "get active payment status" taggedAs EffectsTest in {
+
+    val actual = for {
+     res <- GetPaymentMethodStatus(fakeGet)("activeAccountId")
+    } yield {
+      res
+    }
```
johnduffell: I think the cfor comprehension could be simplified to
```scala
val actual = GetPaymentMethodStatus(fakeGet)("activeAccountId")
```

handlers/new-product-api/src/test/scala/com/gu/newproduct/api/addsubscription/zuora/GetPaymentMethodStatusTest.scala
```diff
@@ -0,0 +1,48 @@
...
+import org.scalatest.{FlatSpec, Matchers}
+import scalaz.{-\/, \/-}
+
+class GetPaymentMethodStatusTest extends FlatSpec with Matchers {
+
+  val fakeResponses = Map(
+    "object/payment-method/activeAccountId" -> PaymentMethodWire("Active"),
+    "object/payment-method/closedAccountId" -> PaymentMethodWire("Closed"),
+    "object/payment-method/unexpected" -> PaymentMethodWire("unexpected"),
+  )
```
johnduffell: I'd rather define the test data in the relevant test if possible, rather than splitting out the test data at the top.

johnduffell:
yep, couple of suggestions to simply the tests a bit and improve naming, but all in all nothing to stop it 👍

PR #145, opened 2018-07-17
add data types and ids for contributions plan ids

johnduffell:
We need the plan ids in the checks (to make sure the contributors don't have an existing contribution) and also to subscribe them to the corect plan.  So we are going to work separately on the two parts, but add this shared code first so we can make progress separately.
Since the zuora IDs are not secret, it makes more sense to let the compiler validate the object rather than trying to deal with JSON and messing with config that's only checked at deploy/run time.
@pvighi @david-pepper

### DIFF
--- a/handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/Handler.scala
+++ b/handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/Handler.scala
@@ -84,7 +84,7 @@ object Handler {
       Stage("CODE") -> RecordTypeId("012g0000000DZmNAAW"),
       Stage("DEV") -> RecordTypeId("STANDARD_TEST_DUMMY")
     )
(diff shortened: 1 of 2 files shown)


handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/Handler.scala
```diff
@@ -84,7 +84,7 @@ object Handler {
       Stage("CODE") -> RecordTypeId("012g0000000DZmNAAW"),
       Stage("DEV") -> RecordTypeId("STANDARD_TEST_DUMMY")
     )
-    mappings.get(stage).toApiGatewayContinueProcessing(ApiGatewayResponse.internalServerError(s"missing config for stage $stage"))
+    mappings.get(stage).toApiGatewayContinueProcessing(ApiGatewayResponse.internalServerError(s"missing standard record type for stage $stage"))
```
johnduffell: just noticed this mistake while I was in

PR #146, opened 2018-07-17
add operation to create subscriptions in an existing zuora account

johnduffell:
This adds a new operation to call zuora with a new subscription.  This gets the appropriate product ids and account ids and can make the right subscription.
@pvighi @david-pepper

### DIFF
--- a/handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/zuora/CreateSubscription.scala
+++ b/handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/zuora/CreateSubscription.scala
@@ -75,7 +75,11 @@ object CreateSubscription {
   def apply(
     planAndCharge: PlanAndCharge,
     post: RequestsPost[WireCreateRequest, WireSubscription]
-  )(createSubscription: CreateReq): ClientFailableOp[SubscriptionName] =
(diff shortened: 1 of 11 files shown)


handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/zuora/CreateSubscription.scala
```diff
@@ -0,0 +1,81 @@
...
+    acquisitionCase: CaseId
+  )
+
+  case class SubscriptionName(value: String) extends AnyVal
+
+  def apply(
+    planAndCharge: PlanAndCharge,
+    post: RequestsPost[WireCreateRequest, WireSubscription]
+  )(createSubscription: CreateReq): ClientFailableOp[SubscriptionName] =
+    post(createRequest(createSubscription, planAndCharge), s"subscriptions", WithCheck).map(id => SubscriptionName(id.subscriptionNumber))
```
pvighi: minor: maybe it should be called wireSub instead of id since id.subscriptionNumber looks strange

johnduffell: good spot, I've tidied it up a bit in general too

PR #147, opened 2018-07-17
prerequisites validation for new product api

pvighi:
Added code to validate that:
* the account has an identity id
* the account has an active payment method
* the account has no existing active recurring contribution subscriptions
* the account has autopay enabled
* the account has balance = 0
will add in a future PR:
* validation that payment method type is one of the accepted ones
* start date should be today


handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/PrerequesiteCheck.scala
```diff
@@ -0,0 +1,39 @@
...
+
+import com.gu.newproduct.api.addsubscription.Handler.ProductRatePlanId
+import com.gu.newproduct.api.addsubscription.zuora.GetAccount.{Account, PaymentMethodId}
+import com.gu.newproduct.api.addsubscription.zuora.GetAccountSubscriptions.Subscription
+import com.gu.newproduct.api.addsubscription.zuora.GetAccountSubscriptions.{Active => ActiveSub}
+import com.gu.newproduct.api.addsubscription.zuora.GetPaymentMethodStatus.{Active => ActivePaymentMethod, PaymentMethodStatus}
+
+import com.gu.util.zuora.RestRequestMaker.{ClientFailableOp, GenericError}
+import scalaz.{-\/, \/, \/-}
+
```
pvighi: now that I look at it, this object just returns ClientFailableOp[Unit] so It cannot decide what http status code to return, so if there's an api error the user will see the same type of error as a validation error.
I should probably change it so that It returns an ApiGateway op with the correct status code

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/PrerequesiteCheck.scala
```diff
@@ -0,0 +1,39 @@
+package com.gu.newproduct.api.addsubscription
+
+import com.gu.newproduct.api.addsubscription.Handler.ProductRatePlanId
+import com.gu.newproduct.api.addsubscription.zuora.GetAccount.{Account, PaymentMethodId}
+import com.gu.newproduct.api.addsubscription.zuora.GetAccountSubscriptions.Subscription
+import com.gu.newproduct.api.addsubscription.zuora.GetAccountSubscriptions.{Active => ActiveSub}
+import com.gu.newproduct.api.addsubscription.zuora.GetPaymentMethodStatus.{Active => ActivePaymentMethod, PaymentMethodStatus}
```
pvighi: probably should rename this to avoid naming conflicts in a nicer way

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/PrerequesiteCheck.scala
```diff
@@ -0,0 +1,42 @@
...
+      _ <- check(account.accountBalanceMinorUnits.value == 0, ifFalseReturn = "Zuora account balance is not zero")
+      paymentMethodId <- extract(account.paymentMethodId, ifNoneReturn = "Zuora account has no default payment method")
+      paymentMethodStatus <- getPaymentMethodStatus(paymentMethodId).toApiGatewayOp("load payment method status from Zuora")
+      _ <- check(paymentMethodStatus == ActivePaymentMethod, "Default payment method status in Zuora account is not active")
+      subscriptions <- getAccountSubscriptions(accountId).toApiGatewayOp("load subscriptions for Zuora account")
+      _ <- check(!subscriptions.filter(_.status == ActiveSub).exists(hasContributions), ifFalseReturn = "Zuora account already has an active recurring contribution subscription")
+    } yield ()
+  }
+
+  def errorResponse(msg: String) = ReturnWithResponse(ApiGatewayResponse.messageResponse("422", msg))
```
johnduffell: I had a quick search and it seems that 422 is for WebDAV , so maybe that's why none of us heard of it!

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/PrerequesiteCheck.scala
```diff
@@ -0,0 +1,42 @@
...
+
+  def check(condition: Boolean, ifFalseReturn: String): ApiGatewayOp[Unit] = if (condition) ContinueProcessing(()) else errorResponse(ifFalseReturn)
```
johnduffell: I noticed earlier we already have an implicit class to do .toApiGatewayOp on a boolean, here https://github.com/guardian/zuora-auto-cancel/blob/d6aaa637a548390a0ee7ea88c1dc4aa171a527a3/lib/handler/src/main/scala/com/gu/util/reader/Types.scala#L132
I remember us generally debating about separating the "type converting" code from the actual decision logic that people want to look at (mostly in the context of EitherT and things), and implicit classes seemed OK.
Then the for comprehension lines would be more like
```scala
for {
...
  _ <- (paymentMethodStatus == ActivePaymentMethod).toApiGatewayContinueProcessing(unprocessable("Default payment method status in Zuora account is not active"))
...
} ....
```
from a functional point of view, I prefer how you did it, but I like keeping the interesting parts of the code Left justified too!

See what you think anyway.  Same applies to extract - we have a toApiGatewayOp on an Option in the same file.

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/PrerequesiteCheck.scala
```diff
@@ -0,0 +1,42 @@
...
+    for {
+      account <- getAccount(accountId).toApiGatewayOp("load account from Zuora")
+      _ <- check(account.identityId.isDefined, ifFalseReturn = "Zuora account has no Identity Id")
+      _ <- check(account.autoPay.value, ifFalseReturn = "Zuora account has autopay disabled")
+      _ <- check(account.accountBalanceMinorUnits.value == 0, ifFalseReturn = "Zuora account balance is not zero")
+      paymentMethodId <- extract(account.paymentMethodId, ifNoneReturn = "Zuora account has no default payment method")
+      paymentMethodStatus <- getPaymentMethodStatus(paymentMethodId).toApiGatewayOp("load payment method status from Zuora")
+      _ <- check(paymentMethodStatus == ActivePaymentMethod, "Default payment method status in Zuora account is not active")
+      subscriptions <- getAccountSubscriptions(accountId).toApiGatewayOp("load subscriptions for Zuora account")
+      _ <- check(!subscriptions.filter(_.status == ActiveSub).exists(hasContributions), ifFalseReturn = "Zuora account already has an active recurring contribution subscription")
```
johnduffell: maybe split line 31 to 2 lines for clarity (with optional implicit class - use whichever type conversion you like)
```scala
  maybeExistingContribution = subscriptions.find { sub =>
    sub.status == ActiveSub && hasContributions(sub)
  }
  _ <- maybeExistingContribution.isEmpty.toApiGatewayContinueProcessing(unprocessable("Zuora account already has an active recurring contribution subscription"))
```

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/zuora/GetAccount.scala
```diff
@@ -9,39 +9,41 @@ object GetAccount {
...
-    paymentMethodId: PaymentMethodId,
+  case class Account(
+    identityId: Option[IdentityId],
+    paymentMethodId: Option[PaymentMethodId],
     autoPay: AutoPay,
     accountBalanceMinorUnits: AccountBalanceMinorUnits
   )
 
-  def apply(get: RequestsGet[ZuoraAccount])(accountId: ZuoraAccountId): ClientFailableOp[AccountSummary] =
+  def apply(get: RequestsGet[ZuoraAccount])(accountId: ZuoraAccountId): ClientFailableOp[Account] =
```
johnduffell: thanks for renaming all these and changing to option.  feel free to also rename anything needed to avoid clashing names

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/zuora/GetAccountSubscriptions.scala
```diff
@@ -0,0 +1,48 @@
...
+
+    case class ZuoraSubscription(
+      subscriptionNumber: String,
+      status: String,
+      ratePlans: List[ZuoraRatePlan]
+    )
+
+    def fromWire(zuoraSubscription: ZuoraSubscription): Subscription = Subscription(
+      status = if (zuoraSubscription.status == "Active") Active else NotActive,
+      productRateplanIds = zuoraSubscription.ratePlans.map(rp => ProductRatePlanId(rp.productRatePlanId)).toSet
```
johnduffell: toSet is a smart idea - wouldn't have thought of that

handlers/new-product-api/src/test/scala/com/gu/newproduct/api/addsubscription/PrerequisiteCheckTest.scala
```diff
@@ -0,0 +1,108 @@
...
+    val testAccounts = Map(
+      "validAccount" -> validAccount,
+      "noIdentityAccount" -> noIdentityAccount,
+      "noPaymentAccount" -> noPaymentAccount,
+      "noAutoPayAccount" -> noAutoPayAccount,
+      "disabledPaymentMethodAccount" -> disabledPaymentAccount,
+      "balanceNotZeroAccount" -> balanceNotZeroAccount,
+      "monthlyContributingAccount" -> validAccount,
+      "annualContributingAccount" -> validAccount
+    )
```
johnduffell: as discussed for the general case - it's not necessary to hold things up refactoring now:

it would be nice to keep the tests self contained by default (shared bits by exception), just for ease of understanding.  having to share lots of scaffolding could indicate we are testing at the wrong level of abstraction.

handlers/new-product-api/src/test/scala/com/gu/newproduct/api/addsubscription/zuora/GetAccountSubscriptionsTest.scala
```diff
@@ -0,0 +1,62 @@
...
+
+  def fakeGet(path: String, skipCheck: IsCheckNeeded) = \/-(fakeResponses(path))
+
+  it should "get subscriptions from Account" taggedAs EffectsTest in {
+
+    val actual = for {
+      res <- GetAccountSubscriptions(fakeGet)(ZuoraAccountId("account1"))
+    } yield {
+      res
+    }
```
johnduffell: ```val actual = GetAccountSubscriptions(fakeGet)(ZuoraAccountId("account1"))```
😆 slipped in there again

handlers/new-product-api/src/test/scala/com/gu/newproduct/api/addsubscription/zuora/GetAccountTest.scala
```diff
@@ -9,24 +9,44 @@ import scalaz.{-\/, \/-}
...
+    Balance = 24.55
+  )
+
+  val missingFieldsAccount: ZuoraAccount = ZuoraAccount(
+    IdentityId__c = None,
+    DefaultPaymentMethodId = None,
+    AutoPay = true,
+    Balance = 11.99
+  )
+  val accF: RequestsGet[ZuoraAccount] = {
```
johnduffell: same as the other test I commented - I wouldn't necessarily say we improve the structure by sharing accF rather than keeping them in context.  just means we are coupling the different tests together (e.g. if we delete one of the tests, do we need to find all the other places that are relevant to it?)

johnduffell:
looks great, I've thrown in a load of mostly general comments and food for thought, but nothing to block this, have a read of course at least for future reference/debate, but feel free to ship either way
👍

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/PrerequesiteCheck.scala
pvighi: does that mean that we should not use it ? 
should we use 400 bad request ?

johnduffell:
👍 🚀

PR #151, opened 2018-07-18
refactored validation code

pvighi:


handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/validation/PrerequisiteCheck.scala
```diff
@@ -0,0 +1,21 @@
+package com.gu.newproduct.api.addsubscription.validation
```
pvighi: we could just delete this whole file and just add there 3 lines to the for comprehension that would call this instead

johnduffell: yes, alternatively as a simplification, looking at this more closely, I realise that we are not going to test this PrerequisiteCheck on its own, so we could reasonably hard code the links from here to the checks, rather than doing the injection.  originally I thought we should be injecting everywhere, but actually it's only necessary to make the code more flexible

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/validation/ValidateAccount.scala
```diff
@@ -0,0 +1,30 @@
...
+      case NotFound(_) => ReturnWithResponse(errorResponse(IfNotFoundReturn))
+      case GenericError(e) =>
+        logger.error(s"error while getting zuora account: $e")
+        ReturnWithResponse(internalServerError("error while getting zuora account"))
+      case ClientSuccess(ac) => ContinueProcessing(ac)
+    }
+
+    for {
+      account <- accountFromZuora(id = accountId, IfNotFoundReturn = "Zuora account id is not valid")
+      _ <- check(account.identityId.isDefined, ifFalseReturn = "Zuora account has no Identity Id")
```
pvighi: I forgot to change these lines. Before merging I will refactor the for comprehensions in the validators to look like : 
```scala
_ <- account.identityId.isDefined.ifFalseReturn("Zuora account has no Identity Id")
```

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/validation/ValidateSubscriptions.scala
```diff
@@ -0,0 +1,24 @@
...
+import com.gu.newproduct.api.addsubscription.zuora.GetAccountSubscriptions.{Active, Subscription}
+import com.gu.util.reader.Types._
+import com.gu.util.resthttp.Types.ClientFailableOp
+
+object ValidateSubscriptions extends Validation {
+  def apply(
+    getAccountSubscriptions: ZuoraAccountId => ClientFailableOp[List[Subscription]],
+    contributionRatePlanIds: List[ProductRatePlanId]
+  )(accountId: ZuoraAccountId): ApiGatewayOp[Unit] = {
+    def hasActiveContributions(s: Subscription) = s.status == Active && s.productRateplanIds.exists(contributionRatePlanIds.contains(_))
```
johnduffell: OK, but when nesting code like this just be careful of unseen dependencies - hasActiveContributions seems to only need a `Subscription`, but it magically gets access to `contributionRatePlanIds` which it needs, plus `getAccountSubscriptions` and `accountId` which it doesn't.

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/validation/Validation.scala
```diff
@@ -0,0 +1,11 @@
+package com.gu.newproduct.api.addsubscription.validation
+
+import com.gu.util.reader.Types._
+import com.gu.util.apigateway.ApiGatewayResponse
+import com.gu.util.reader.Types.ApiGatewayOp
+
+trait Validation {
```
johnduffell: single level trait and an object with  import Validation._ are indeed the same in terms of what you get

well you said I wouldn't like it, so I had a think why 💣 here goes:

so with a trait it's all-or-nothing, whereas with an object, you can choose to import parts if necessary. (minor benefit)
however with a trait it saves the hassle of messing with underscores in the imports
but does `check` need to be always bound to `extract`?
to separate we could have three traits with errorResponse at the top! 😆 - but you can't dependency inject a super-trait - unless you make its method abstract like with RequestsGet trait or override it!! - but better hope there's only one method in there!
but even if we start with one function in each having those braces seems to invite more functions in, for the future.
and what's more with a trait it's only too easy to slip in an abstract method, then we're into inheritance and all that entails.
with imports they are not transitive, with traits you can have a whole extra dimension of traits extending other traits and even multiple inheritance. at first it's fine but that could be dangerous as a habit when we end up having a super trait with everything hanging off somewhere.  also it's another dimension to keep in your head.

I just don't see a strong reason to use a trait.  most of what I see is extra structural risk.

happy to debate further and get other opinions though!

handlers/new-product-api/src/test/scala/com/gu/newproduct/api/addsubscription/validation/ValidateAccountTest.scala
```diff
@@ -0,0 +1,68 @@
...
+    ValidateAccount(getAccount)(ZuoraAccountId("validAccountId")) shouldBe validationError("Zuora account has autopay disabled")
+  }
+  it should "fail if account balance is not zero" in {
+    val balanceNotZeroAccount = validAccount.copy(accountBalanceMinorUnits = AccountBalanceMinorUnits(1000))
+
+    def getAccount = fakeGetAccount(response = balanceNotZeroAccount) _
+
+    ValidateAccount(getAccount)(ZuoraAccountId("validAccountId")) shouldBe validationError("Zuora account balance is not zero")
+  }
+}
```
johnduffell: these tests are nice and clear 🥇 
the components of an ideal functional test are

- name/description
- input data
- operation
- expected output data
- comparison/reporting logic

these tests are not too far off that ideal - it's only the fact that the input data is split between getAccount and the input parameters, and the output data comes both into getAccount and also returned from the function.

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/validation/Validation.scala
pvighi: I just forgot to refactor this, I will remove this trait and replace it with implicit classes for Option and Boolean to add .ifNoneReturn(errorMsg) and .ifFalseReturn(errorMsg)

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/validation/ValidateAccount.scala
johnduffell: fair enough, I wasn't going to comment on that :)

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/validation/Validation.scala
johnduffell: fair enough - it was still useful for me to think about what a trait really means for the codebase (because no point having rules based on dogma)

johnduffell:
I think this is an excellent improvement for the tests, thanks!  I think you are already planning to tweak all the stuff I was concerned about, so once that's done go for it 👍

handlers/new-product-api/src/test/scala/com/gu/newproduct/api/addsubscription/validation/ValidateAccountTest.scala
```diff
@@ -0,0 +1,68 @@
...
+
+  it should "succeed with valid account" in {
+    def getAccount = fakeGetAccount(response = validAccount) _
+
+    ValidateAccount(getAccount)(ZuoraAccountId("validAccountId")) shouldBe ContinueProcessing(PaymentMethodId("activePaymentMethod"))
+  }
+  it should "fail if account cannot be loaded" in {
+    def errorGetAccount(zuoraAccountId: ZuoraAccountId) = GenericError("zuora is down")
+
+    ValidateAccount(errorGetAccount)(ZuoraAccountId("validAccountId")) shouldBe ReturnWithResponse(ApiGatewayResponse.internalServerError("some logging message"))
```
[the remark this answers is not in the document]
pvighi: It's a little confusing but there param to internalServerError is just a log message, the actual response is hard coded to "Internal server error". I wrote "some logging message" in attempt to make this clear but it obviously it's not. My other thought was to use "ignored value" instead, maybe it was better but I ended up deciding against it.

handlers/new-product-api/src/test/scala/com/gu/newproduct/api/addsubscription/validation/ValidateSubscriptionsTest.scala
```diff
@@ -0,0 +1,75 @@
...
+
+  def fakeGetSubs(response: List[Subscription])(id: ZuoraAccountId) = {
+    id.value shouldBe "zuoraAccountId"
+    ClientSuccess(response)
+  }
+
+  it should "fail if account already has an active recurring contribution subscription" in {
+    val subs = List(
+      sub(active = true, rateplans = Set("somePlan", "someOtherPlan")),
+      sub(active = true, rateplans = Set("monthlyRatePlanId", "someOtherPlan"))
```
pvighi: I should change "monthlyRateplanId" to "montlyContributionRateplanId". I tried to make the string shorter in the worst possible way and removed the one part of the word that identifies the recurring rate plan in the tests

johnduffell:
@pvighi I have merged in master now, be sure to pull before continuing! 🔥

PR #152, opened 2018-07-19
tech debt: separate "Auth by url param" from "API gateway handler"

johnduffell:
This is a bit of surgery left over from tech debt to separate out the Auth by URL param (needed for zuora and stripe as they don't allow x-api-key header)
I have done a little bit of other tidy at the same time, but hopefully nothing too much!
I need to test in code, please leave comments regarding things to fix before go live, and also things not needed to fix straight away as we can always add to the tech debt.
@pvighi @jacobwinch @paulbrown1982 @twrichards

### DIFF
--- a/handlers/cancellation-sf-cases/src/main/scala/com/gu/cancellation/sf_cases/Handler.scala
+++ b/handlers/cancellation-sf-cases/src/main/scala/com/gu/cancellation/sf_cases/Handler.scala
@@ -36,7 +36,7 @@ object Handler extends Logging {
   case class SfRequests(normal: LazySalesforceAuthenticatedReqMaker, test: LazySalesforceAuthenticatedReqMaker)
 
   def raiseCase(inputStream: InputStream, outputStream: OutputStream, context: Context): Unit =
-    runWithEffectsDontTestAtThisLevel(
+    runForLegacyTestsSeeTestingMd(
       IdentityCookieToIdentityUser.defaultCookiesToIdentityUser(RawEffects.stage.isProd),
       RaiseCase.steps,
       RawEffects.response,
@@ -102,7 +102,7 @@ object Handler extends Logging {
   implicit val pathParamReads = Json.reads[CasePathParams]
 
   def updateCase(inputStream: InputStream, outputStream: OutputStream, context: Context): Unit =
-    runWithEffectsDontTestAtThisLevel(
+    runForLegacyTestsSeeTestingMd(
       IdentityCookieToIdentityUser.defaultCookiesToIdentityUser(RawEffects.stage.isProd),
       UpdateCase.steps,
       RawEffects.response,
@@ -125,7 +125,7 @@ object Handler extends Logging {
 
   }
 
-  def runWithEffectsDontTestAtThisLevel(
+  def runForLegacyTestsSeeTestingMd(
     cookieValuesToIdentityUser: CookieValuesToIdentityUser,
     steps: Steps,
     response: Request => Response,
--- a/handlers/cancellation-sf-cases/src/main/scala/com/gu/cancellation/sf_cases/Handler.scala
+++ b/handlers/cancellation-sf-cases/src/main/scala/com/gu/cancellation/sf_cases/Handler.scala
@@ -36,7 +36,7 @@ object Handler extends Logging {
   case class SfRequests(normal: LazySalesforceAuthenticatedReqMaker, test: LazySalesforceAuthenticatedReqMaker)
 
   def raiseCase(inputStream: InputStream, outputStream: OutputStream, context: Context): Unit =
-    runWithEffects(
+    runWithEffectsDontTestAtThisLevel(
       IdentityCookieToIdentityUser.defaultCookiesToIdentityUser(RawEffects.stage.isProd),
       RaiseCase.steps,
       RawEffects.response,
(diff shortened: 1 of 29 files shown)


handlers/cancellation-sf-cases/src/main/scala/com/gu/cancellation/sf_cases/Handler.scala
```diff
@@ -36,7 +36,7 @@ object Handler extends Logging {
   case class SfRequests(normal: LazySalesforceAuthenticatedReqMaker, test: LazySalesforceAuthenticatedReqMaker)
 
   def raiseCase(inputStream: InputStream, outputStream: OutputStream, context: Context): Unit =
-    runWithEffects(
+    runWithEffectsDontTestAtThisLevel(
```
jacobwinch: I preferred runWithEffects 🙈

handlers/cancellation-sf-cases/src/main/scala/com/gu/cancellation/sf_cases/Handler.scala
```diff
@@ -125,15 +125,23 @@ object Handler extends Logging {
...
     steps: Steps,
     response: Request => Response,
     stage: Stage,
     fetchString: StringFromS3,
     lambdaIO: LambdaIO
-  ) = {
+  ) =
+    ApiGatewayHandler(lambdaIO)(operationForEffects(cookieValuesToIdentityUser, steps, response, stage, fetchString))
 
+  def operationForEffects(
```
jacobwinch: I prefer this though 👍

handlers/cancellation-sf-cases/src/main/scala/com/gu/cancellation/sf_cases/Handler.scala
johnduffell: I will add a testing.md and we can document our approach to testing there.  then this can point there.

lib/handler/src/test/scala/com/gu/util/apigateway/AuthTest.scala
```diff
@@ -29,16 +28,16 @@ class AuthTest extends FlatSpec {
   }
 
   "deprecatedCredentialsAreValid" should "return false if the username query string is missing" in {
```
jacobwinch: It would be good to remove this completely - perhaps one for the tech debt backlog?

lib/handler/src/main/scala/com/gu/util/apigateway/ApiGatewayHandler.scala
```diff
@@ -46,46 +43,41 @@ object ApiGatewayHandler extends Logging {
...
-    requestAuth: Option[RequestAuth],
-    trustedApiConfig: TrustedApiConfig
-  ): Boolean = {
-    !shouldAuthenticate || credentialsAreValid(requestAuth, trustedApiConfig)
+  case class Operation(
+    steps: ApiGatewayRequest => ApiResponse,
+    healthcheck: () => ApiResponse
+  ) {
+    def prependValidationStep(validate: ApiGatewayRequest => ApiGatewayOp[Unit]): Operation = {
+      def newSteps(request: ApiGatewayRequest): ApiResponse =
```
jacobwinch: postValidationSteps?

lib/handler/src/main/scala/com/gu/util/apigateway/ApiGatewayHandler.scala
```diff
@@ -46,46 +43,41 @@ object ApiGatewayHandler extends Logging {
...
+        (for {
+          _ <- validate(request)
+          result = steps(request)
+        } yield result).apiResponse
+      Operation(newSteps, healthcheck)
+    }
```
jacobwinch: Might be worth adding a unit test for this?

jacobwinch:
Looks good to me - I've added a few small suggestions 👍

lib/handler/src/main/scala/com/gu/util/apigateway/ApiGatewayHandler.scala
johnduffell: probably not.
the passed in `validate` will run before the `steps`
`newSteps` is the steps with the validation included.
however you're right the naming is bad, I'll try to improve it!

lib/handler/src/test/scala/com/gu/util/apigateway/AuthTest.scala
johnduffell: I'm not sure why it's deprecated - I did remove it, but then another test failed because if the user forgets to provide an api key, it was giving a bad request rather than a unauthorised.  So this code does need to deal with missing key itself

johnduffell: ok I found it, it's a correct test with a historical name https://github.com/guardian/zuora-auto-cancel/pull/27/files#diff-018b8686af0b77efdcf4d46962705e6bR31
I'll edit

lib/handler/src/main/scala/com/gu/util/apigateway/ApiGatewayHandler.scala
jacobwinch: Ah yes, sorry, my brain was obviously not fully engaged this morning!

johnduffell:
I accidentally merged into `wired-prereq-check` branch

PR #154, opened 2018-07-19
add logic to pass through case and source

johnduffell:
the AcquisitionSource__c and CreatedByCSR__c were not being populated by the service layer when a sub was created despite them being passed in.  This PR fixes that.
image

I also moved the ClientFailableOp logging object as it was unused and I needed it to see the sub id returned from the test.

@pvighi @david-pepper


johnduffell:
@pvighi we should ship this in time for the release of the SF screens tomorrow morning 11am, hopefully shouldn't take long to review

PR #156, opened 2018-07-19
add healthcheck for the new subscription (only works in PROD)

johnduffell:
make the new product healthcheck actually hit prod zuora in the process

this will only work in prod - not sure if we should keep some other test accounts in a Map?

@pvighi @jacobwinch @paulbrown1982 @twrichards 
cc @david-pepper


handlers/new-product-api/src/test/scala/com/gu/newproduct/api/addsubscription/HealthCheckEffectsTest.scala
```diff
@@ -0,0 +1,25 @@
...
+import com.gu.util.reader.Types._
+import com.gu.util.zuora.{ZuoraRestConfig, ZuoraRestRequestMaker}
+import org.scalatest.{FlatSpec, Matchers}
+
+class HealthCheckEffectsTest extends FlatSpec with Matchers {
+
+  it should "pass" taggedAs EffectsTest in {
+
+    val actual = (for {
+      zuoraConfig <- LoadConfigModule(Stage("PROD"), GetFromS3.fetchString)[ZuoraRestConfig].toApiGatewayOp("load zuora config")
```
johnduffell: 🔥 💣 
not sure if this is a good idea.  I feel like I should make the healthcheck work in all stages!  but it would be nice to be able to run all health checks locally against all stages?

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/HealthCheck.scala
```diff
@@ -0,0 +1,21 @@
...
+import com.gu.util.apigateway.ApiGatewayResponse
+import com.gu.util.apigateway.ResponseModels.ApiResponse
+import com.gu.util.resthttp.Types.ClientFailableOp
+
+object HealthCheck {
+
+  def apply(getAccount: ZuoraAccountId => ClientFailableOp[GetAccount.Account]): ApiResponse =
+    (for {
+      account <- getAccount(ZuoraAccountId("2c92a0fb4a38064e014a3f48f1663ad8")).toApiGatewayOp("get test account from zuora")
+      isCorrect = account.identityId.contains(IdentityId("13552794"))
```
pvighi: we could just have a UAT accountId and expected identity id as to make this work on CODE as well. Are there any downsides to that ?

handlers/new-product-api/src/test/scala/com/gu/newproduct/api/addsubscription/HealthCheckEffectsTest.scala
johnduffell: I will change this to DEV!  this would be safe, but still a bad habit

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/HealthCheck.scala
johnduffell: good idea, there's no downside to having a Map

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/HealthCheck.scala
```diff
@@ -1,21 +1,39 @@
...
     } yield if (isCorrect)
       ApiGatewayResponse.successfulExecution
     else
       ApiGatewayResponse.internalServerError("check identity id in health check")).apiResponse
 
 }
+
+object AccountIdentitys {
+
+  case class AccountIdentity(zuoraAccountId: ZuoraAccountId, identityId: IdentityId)
```
pvighi: minor:
maybe we could rename this to be a little more descriptive like:
```scala
case class HealthCheckData(testAccountId: ZuoraAccountId, expectedIdentityId: IdentityId)
```

handlers/new-product-api/src/test/scala/com/gu/newproduct/api/addsubscription/HealthCheckTest.scala
```diff
@@ -9,18 +9,19 @@ class HealthCheckTest extends FlatSpec with Matchers {
 
   it should "pass" in {
     def getAccount(dontcare: ZuoraAccountId): Types.ClientFailableOp[Account] = {
-      ClientSuccess(Account(Some(IdentityId("13552794")), None, AutoPay(false), AccountBalanceMinorUnits(0)))
+      dontcare should be(ZuoraAccountId("accacc"))
```
pvighi: seems like we do care about the dontcare param now

PR #157, opened 2018-07-20
Additional validations

pvighi:
added validation for:
* min and max amount values by currency
* startdate must be today
* Payment type whitelist


handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/Handler.scala
```diff
@@ -37,12 +38,12 @@ object Handler extends Logging {
...
 
   def addSubscriptionSteps(
-    prerequesiteCheck: ZuoraAccountId => ApiGatewayOp[Unit],
+    prerequesiteCheck: (ZuoraAccountId, AddSubscriptionRequest) => ApiGatewayOp[Unit],
     createMonthlyContribution: CreateReq => ClientFailableOp[SubscriptionName]
   )(apiGatewayRequest: ApiGatewayRequest): ApiResponse = {
     (for {
       request <- apiGatewayRequest.bodyAsCaseClass[AddSubscriptionRequest]().withLogging("parsed request")
-      _ <- prerequesiteCheck(request.zuoraAccountId)
+      _ <- prerequesiteCheck(request.zuoraAccountId, request)
```
pvighi: this makes no sense. I should only pass the request or check exactly what I need from it (but since it goes into the requestvalidator it would make sense to pass in the request)

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/validation/ValidateRequest.scala
```diff
@@ -0,0 +1,37 @@
...
+
+import com.gu.newproduct.api.addsubscription.AddSubscriptionRequest
+import com.gu.util.reader.Types._
+import Validation.BooleanValidation
+import com.gu.i18n.Currency
+import com.gu.i18n.Currency._
+
+object ValidateRequest {
+  def apply(
+    now: () => LocalDateTime
```
johnduffell: this only needs  a LocalDate, not a LocalDateTime, so we could do the toLocalDate when we wire the dependency at the beginning.  Then for testing it only needs a LocalDate.of(2018,7,20) rather than the whole time too.

pvighi: 🤦‍♂️

pvighi: I guess I wanted a function that returns the current time, not a val with the current time in it. Given that this is a lambda and everything gets recreated every time it's called it doesn't make much of a difference

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/zuora/GetPaymentMethod.scala
```diff
@@ -0,0 +1,60 @@
...
+  case class PaymentMethod(status: PaymentMethodStatus, paymentMethodType: PaymentMethodType)
+
+  def apply(get: RequestsGet[PaymentMethodWire])(paymentMethodId: PaymentMethodId): ClientFailableOp[PaymentMethod] =
+    get(s"object/payment-method/${paymentMethodId.value}", WithoutCheck).map(_.toPaymentMethod)
+}
+
+sealed trait PaymentMethodType
+
+object PaymentMethodType {
+  val all = List(PayPal, CreditCard, CreditCardReferenceTransaction, BankTransfer, Other)
```
johnduffell: this is exactly the kind of list I forget to add things to! I don't know if we can help that though.

only idea I have it to maybe put it at the end where people are more likely to add new values?

handlers/new-product-api/src/test/scala/com/gu/newproduct/api/addsubscription/StepsTest.scala
```diff
@@ -25,7 +25,7 @@ class StepsTest extends FlatSpec with Matchers {
       ClientSuccess(SubscriptionName("well done"))
     }
 
-    def fakeCheck(accountId: ZuoraAccountId): ApiGatewayOp[Unit] =
+    def fakeCheck(accountId: ZuoraAccountId, request: AddSubscriptionRequest): ApiGatewayOp[Unit] =
       if (accountId.value == "acccc") ContinueProcessing(())
```
johnduffell: presume there's no need to check the request here?

handlers/new-product-api/src/test/scala/com/gu/newproduct/api/addsubscription/zuora/GetAccountTest.scala
```diff
@@ -41,8 +44,18 @@ class GetAccountTest extends FlatSpec with Matchers {
...
-      AccountBalanceMinorUnits(2455)
+      AccountBalanceMinorUnits(2455),
+      GBP
     ))
   }
+  it should "return error if Account has unknown currencyu" in {
+    val unknownCurrencyAccount = acc.copy(Currency = "unknown currency code here")
+    val accF: RequestsGet[ZuoraAccount] = {
+      case ("object/account/id", WithoutCheck) => ClientSuccess(unknownCurrencyAccount)
+      case in => GenericError(s"bad request: $in")
```
johnduffell: we could decided to go with the alternative 
```scala
val accF: RequestsGet[ZuoraAccount] = { in =>
  in should be("object/account/id", WithoutCheck)
  ClientSuccess(unknownCurrencyAccount)
}
```

handlers/new-product-api/src/test/scala/com/gu/newproduct/api/addsubscription/StepsTest.scala
```diff
@@ -25,9 +25,9 @@ class StepsTest extends FlatSpec with Matchers {
       ClientSuccess(SubscriptionName("well done"))
     }
 
-    def fakeCheck(accountId: ZuoraAccountId, request: AddSubscriptionRequest): ApiGatewayOp[Unit] =
-      if (accountId.value == "acccc") ContinueProcessing(())
-      else ReturnWithResponse(ApiGatewayResponse.internalServerError(s"whoops: $accountId was wrong for prereq check"))
+    def fakeCheck(request: AddSubscriptionRequest): ApiGatewayOp[Unit] =
+      if (request.zuoraAccountId.value == "acccc") ContinueProcessing(())
```
johnduffell: I commented here but it got overridden by github - basically I was wondering if we should be checking the whole contents rather than just the account id.  
Also we can use the new "request... should be(blah)" style instead of returning a left

johnduffell:
👍 👍 👍 🚀 only nit picks really

handlers/new-product-api/src/test/scala/com/gu/newproduct/api/addsubscription/validation/ValidatePaymentMethodTest.scala
```diff
@@ -11,24 +14,38 @@ class ValidatePaymentMethodTest extends FlatSpec with Matchers {
...
+    val internalServerError = ReturnWithResponse(ApiGatewayResponse.internalServerError("some log message"))
+    ValidatePaymentMethod(getPaymentMethodStatus)(paymentMethodId) shouldBe internalServerError
   }
 
+  it should "fail if payment method is of an unknown type" in {
+    val paymentMethod = PaymentMethod(ActivePaymentMethod, Other)
+    def getPaymentMethodStatus = fakeGetPaymentMethodStatus(ClientSuccess(paymentMethod)) _
+    val validPaymentTypes = PaymentMethodType.all.filterNot(_ == Other)
+    val expectedMessage = s"Invalid payment method type in Zuora account, must be one of ${validPaymentTypes.mkString(",")}"
+    ValidatePaymentMethod(getPaymentMethodStatus)(paymentMethodId) shouldBe validationError(expectedMessage)
```
johnduffell: as discussed, for another PR:
we could make this return some kind of customised Option type and then have a conversion from that to ApiGatewayOp in the Prereq Steps.  Then these tests would be simpler.

johnduffell:
still 👍 although you now have a small conflict, sorry!

johnduffell:
👍 🚀

PR #159, opened 2018-07-25
When raising SF case, will hand back Id of applicable recent case if possible (i.e. resume case)

twrichards:
As anticipated, based on some users changing cancellation reason in the online flow, cases are being created which look like 'saves' when in fact the user goes on to cancel with a different reason. 

This change will address this problem. A 'suitable recent case' is as follows...
- same contactId as the authenticated user (obviously)
- same subject as the starting subject for new cases ("Online Cancellation Attempt")
- origin is self service
- case was created in the last 3 days
- same subscription as the raise case request

PLUS some refactoring and some old TODOs (update case now verifies case belongs to authenticated user)

### DIFF
--- a/handlers/cancellation-sf-cases/src/main/scala/com/gu/cancellation/sf_cases/Handler.scala
+++ b/handlers/cancellation-sf-cases/src/main/scala/com/gu/cancellation/sf_cases/Handler.scala
@@ -2,26 +2,26 @@ package com.gu.cancellation.sf_cases
 
 import java.io.{InputStream, OutputStream}
 
+import ai.x.play.json.Jsonx
 import com.amazonaws.services.lambda.runtime.Context
 import com.gu.cancellation.sf_cases.TypeConvert._
 import com.gu.effects.{GetFromS3, RawEffects}
-import com.gu.identity.IdentityCookieToIdentityUser.{CookieValuesToIdentityUser, IdentityUser}
+import com.gu.identity.IdentityCookieToIdentityUser.{CookieValuesToIdentityUser, IdentityId, IdentityUser}
 import com.gu.identity.{IdentityCookieToIdentityUser, IdentityTestUserConfig, IsIdentityTestUser}
 import com.gu.salesforce.SalesforceGenericIdLookup
-import com.gu.salesforce.SalesforceGenericIdLookup.{ResponseWithId, SalesforceGenericIdLookupParams, TSalesforceGenericIdLookup}
+import com.gu.salesforce.SalesforceGenericIdLookup.{FieldName, LookupValue, ResponseWithId, SfObjectType, TSalesforceGenericIdLookup}
 import com.gu.salesforce.auth.SalesforceAuthenticate
 import com.gu.salesforce.auth.SalesforceAuthenticate.{SFAuthConfig, SFAuthTestConfig}
 import com.gu.salesforce.cases.SalesforceCase
-import com.gu.salesforce.cases.SalesforceCase.GetMostRecentCaseByContactId.{GetMostRecentCaseByContactIdParams, TGetMostRecentCaseByContactId}
-import com.gu.salesforce.cases.SalesforceCase.Raise.{NewCase, RaiseCase}
-import com.gu.salesforce.cases.SalesforceCase.{CaseWithId, GetMostRecentCaseByContactId}
+import com.gu.salesforce.cases.SalesforceCase.Create.NewCase
+import com.gu.salesforce.cases.SalesforceCase.GetMostRecentCaseByContactId.TGetMostRecentCaseByContactId
+import com.gu.salesforce.cases.SalesforceCase.{CaseId, CaseOrigin, CaseSubject, CaseWithId, ContactId, SubscriptionId}
 import com.gu.util.Logging
 import com.gu.util.apigateway.ApiGatewayHandler.{LambdaIO, Operation}
 import com.gu.util.apigateway.ResponseModels.ApiResponse
 import com.gu.util.apigateway.{ApiGatewayHandler, ApiGatewayRequest, ApiGatewayResponse}
 import com.gu.util.config.LoadConfigModule.StringFromS3
 import com.gu.util.config._
-import com.gu.util.reader.Types.ApiGatewayOp.{ContinueProcessing, ReturnWithResponse}
 import com.gu.util.reader.Types._
 import com.gu.util.resthttp.Types.ClientFailableOp
 import com.gu.util.resthttp.{RestRequestMaker, Types}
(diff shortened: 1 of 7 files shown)


handlers/cancellation-sf-cases/src/main/scala/com/gu/cancellation/sf_cases/Handler.scala
```diff
@@ -47,54 +51,117 @@ object Handler extends Logging {
...
+      subscriptionName: SubscriptionName
+    ): ApiGatewayOp[CaseWithId] =
       for {
-        sfContactId <- lookup("Contact", "IdentityID__c", identityId)
-          .toApiGatewayOp("lookup SF contact from identityID")
-        sfSubscriptionIdContainer <- lookup("SF_Subscription__c", "Name", raiseCaseDetail.subscriptionName)
-          .toApiGatewayOp("lookup SF subscription ID")
-        raiseCaseResponse <- raiseCase(buildNewCaseForSalesforce(raiseCaseDetail, sfSubscriptionIdContainer, sfContactId))
-          .toApiGatewayOp("raise sf case")
+        sfContactId <- lookupByIdOp(SalesforceGenericIdLookupParams(
```
johnduffell: as discussed I'm not keen on the String String String thing even wrapped in a SalesforceGenericIdLookupParams, I'd rather have something where the Strings are only exposed at the edges of the program.  However this can be left for another phase of refactoring as it's not affecting the functionality it's more for maintainability.

handlers/cancellation-sf-cases/src/main/scala/com/gu/salesforce/cases/SalesforceCase.scala
```diff
@@ -26,25 +31,51 @@ object SalesforceCase extends Logging {
...
 
     def apply(sfRequests: Requests)(caseId: String, body: JsValue): ClientFailableOp[Unit] =
-      sfRequests.patch(body, s"${caseBaseUrl}/${caseId}")
+      sfRequests.patch(body, s"$caseSObjectsBaseUrl/$caseId")
 
   }
 
   object GetById {
 
     def apply[ResponseType](sfRequests: Requests)(caseId: String)(implicit ev1: Reads[ResponseType]): ClientFailableOp[ResponseType] =
```
johnduffell: not sure if you planned to change this back to the [ResponseType: Reads] instead of the implicit group.  Either way won't really cause any issue, just slightly more terse

handlers/cancellation-sf-cases/src/main/scala/com/gu/cancellation/sf_cases/Handler.scala
```diff
@@ -47,54 +51,117 @@ object Handler extends Logging {
...
+      raiseCaseOp: RaiseCase,
+      sfUpdateOp: (String, JsValue) => ClientFailableOp[Unit],
+      raiseCaseDetail: RaiseCaseDetail
+    )(
+      sfContactId: ResponseWithId,
+      sfSubscriptionIdContainer: ResponseWithId,
+      sfRecentCases: GetMostRecentCaseByContactId.RecentCases
+    ) = {
+      if (sfRecentCases.records.isEmpty) // no recent cases so create one
+        raiseCaseOp(buildNewCaseForSalesforce(raiseCaseDetail, sfSubscriptionIdContainer, sfContactId))
```
johnduffell: can we compose the raiseCaseOp and buildNewCaseForSalesforce before passing it into this function? so rather than passing in both we just pass in the composition of the two.
That way the sfSubscriptionIdContainer and sfContactId parameters can be partially applied into it beforehand as they are only used to pass through.
As a result this raiseOrResumeCase function would have fewer parameters and also would lose the direct dependency on buildNewCaseForSalesforce, meaning it's simpler.

handlers/cancellation-sf-cases/src/main/scala/com/gu/cancellation/sf_cases/Handler.scala
```diff
@@ -47,54 +51,117 @@ object Handler extends Logging {
...
+          .toApiGatewayOp("raise sf case")
+      else // recent case exists, so just update the reason and return the case
+        updateReasonOnRecentCase(sfUpdateOp)(raiseCaseDetail.reason, sfRecentCases.records.head)
```
johnduffell: again here we can probably compose updateReasonOnRecentCase and sfUpdateOp beforehand.  it wouldn't really simplify much compared with the above suggestion, but it might be worth removing the direct dependency

handlers/cancellation-sf-cases/src/main/scala/com/gu/cancellation/sf_cases/Handler.scala
```diff
@@ -113,14 +180,38 @@ object Handler extends Logging {
 
   object UpdateCase {
 
+    case class CaseWithContactId(ContactId: String)
+    implicit val caseReads: Reads[CaseWithContactId] = Json.reads[CaseWithContactId]
+
+    def verifyCaseBelongsToUser(
+      sfRequests: RestRequestMaker.Requests
```
johnduffell: if you pass in the
SalesforceGenericIdLookup(sfRequests)
and
SalesforceCase.GetById[CaseWithContactId](sfRequests)
rather than the whole sfRequests, although it is more parameters, they are more high level (and therefore less powerful)  This would make it simpler.  At the moment it is saying "verifyCaseBelongsToUser using any operation in SF" wheras it would actually say "verifyCaseBelongsToUser by only doing generic lookups and getting cases by id"

handlers/cancellation-sf-cases/src/main/scala/com/gu/cancellation/sf_cases/Handler.scala
```diff
@@ -113,14 +180,38 @@ object Handler extends Logging {
...
+      for {
+        sfContactIdFromIdentity <- SalesforceGenericIdLookup(sfRequests)(SalesforceGenericIdLookupParams(
+          sfObjectType = "Contact",
+          fieldName = "IdentityID__c",
+          lookupValue = identityId
+        )).toApiGatewayOp("lookup SF contact ID from identityID")
+        sfCaseDetail <- SalesforceCase.GetById[CaseWithContactId](sfRequests)(caseId)
+          .toApiGatewayOp("lookup SF Case details")
+        sfContactIdFromCase = sfCaseDetail.ContactId
+        _ <- if (sfContactIdFromIdentity.Id equals sfContactIdFromCase) ContinueProcessing(())
```
johnduffell: this is very clear, at my suggestion @pvighi was using a slightly nicer but less explicit looking function to convert boolean to a validation error which you may find clearer (or not!)
ie you could `import com.gu.util.reader.Types._` and use 
```scala
(sfContactIdFromIdentity.Id equals sfContactIdFromCase)
  .toApiGatewayContinueProcessing(ApiGatewayResponse.forbidden(
    s"Authenticated user (${sfContactIdFromIdentity.Id}) does not match ContactId ($sfContactIdFromCase) for Case ($caseId)"
  )
```

handlers/cancellation-sf-cases/src/main/scala/com/gu/salesforce/cases/SalesforceCase.scala
```diff
@@ -26,25 +31,51 @@ object SalesforceCase extends Logging {
...
+      val soqlQuery = s"SELECT Id " +
+          s"FROM Case " +
+          s"WHERE ContactId = '${params.contactId}' " +
+          s"AND Origin = '${params.caseOrigin}' " +
+          s"AND CreatedDate = LAST_N_DAYS:3 " +
+          s"AND SF_Subscription__c = '${params.subscriptionId}' " +
+          s"AND Subject = '${params.caseSubject}' " +
+          s"ORDER BY CreatedDate DESC " +
+          s"LIMIT ${params.limit}"
+      logger.info(soqlQuery)
```
johnduffell: better to prepend something to the log line, so we can understand what logged that line. e.g. 
```logger.info(s"using SF query $soqlQuery")`

File: lib/handler/src/main/scala/com/gu/util/apigateway/ApiGatewayResponse.scala [hunk 8]
```diff
@@ -54,36 +54,41 @@ object ApiGatewayResponse extends Logging {
     ApiResponse(statusCode, bodyTxt)
   }
 
-  val successfulExecution = ApiResponse(
+  val successfulExecution = messageResponse(
     "200",
-    toJsonBody(ResponseBody("Success"))
+    "Success"
```
johnduffell: thanks for simplifying that a bit!

handlers/cancellation-sf-cases/src/test/scala/com/gu/cancellation/sf_cases/EndToEndHandlerEffectsTest.scala
```diff
@@ -3,16 +3,16 @@ package com.gu.cancellation.sf_cases
 import java.io.{ByteArrayInputStream, ByteArrayOutputStream}
```
johnduffell: this EffectsTest looks like the only Test that's been changed as a result of this - if possible we should have some tests that can run locally whether running at the unit level or even end to end with mocked HTTP responses if you're feeling brave - perhaps this is why the code coverage has been dropping recently (it doesn't include effectsTests in the automated coverage)  I think this is on your tech debt list but I thought I'd mention just in case.

johnduffell:
looks good overall, I've mentioned a few things, but as a summary the main categories are
- Strings: very little is "only" a string, so it's not making good use of the type system.  Generally I'd only recommend working in Strings on the simplest lambda.
- internal dependencies: generally I am assessing the complexity of a codebase purely by looking at the wherher the function signatures feel right.  If the parameters look like they are too low level or there are some that are purely being piped through, there's an opportunity to apply them beforehand.  This makes it a lot easier to unit test as a side effect.  And also makes sure nothing low level ends up passed all the way through the code base.
- I didn't mention, but I think there may be a benefit from reducing the size of some of the files. Since everything is fairly loosely coupled it should be pretty much a case of doing a Refactor->Move on the objects inside, and pressing go.  This might make things easier to understand - up to your judgement.

I am aware this PR is including a lot of work already and I don't think anything is a stop-ship, it can either move to tech debt or bear in mind when you are doing the next work (or you can disagree!)  But some of the things you would probably want to deal with before shipping.
👍

handlers/cancellation-sf-cases/src/main/scala/com/gu/cancellation/sf_cases/Handler.scala
```diff
@@ -47,54 +51,117 @@ object Handler extends Logging {
...
         SF_Subscription__c = sfSubscriptionIdContainer.Id,
         Journey__c = "SV - At Risk - MB",
-        Enquiry_Type__c = raiseRequestBody.reason,
+        Enquiry_Type__c = raiseCaseDetail.reason.value,
         Status = "Closed",
-        Subject = "Online Cancellation Attempt"
+        Subject = STARTING_CASE_SUBJECT
       )
 
+    def updateReasonOnRecentCase(
```
pvighi: this is the most minor of comments but couldn't this method just be called updateCaseReason or setCaseReason ?
or is there something that makes it specific to recent cases ?

handlers/cancellation-sf-cases/src/main/scala/com/gu/salesforce/cases/SalesforceCase.scala
twrichards: ✅

lib/handler/src/main/scala/com/gu/util/apigateway/ApiGatewayResponse.scala
twrichards: 👍

handlers/cancellation-sf-cases/src/main/scala/com/gu/cancellation/sf_cases/Handler.scala
twrichards: ✅

handlers/cancellation-sf-cases/src/main/scala/com/gu/cancellation/sf_cases/Handler.scala
twrichards: ✅

handlers/cancellation-sf-cases/src/main/scala/com/gu/cancellation/sf_cases/Handler.scala
```diff
@@ -47,54 +51,117 @@ object Handler extends Logging {
...
+      sfUpdateOp: (String, JsValue) => Types.ClientFailableOp[Unit]
+    )(
+      reason: Reason,
+      recentCase: CaseWithId
+    ): ClientFailableOp[CaseWithId] = for {
+      _ <- sfUpdateOp(recentCase.Id, JsObject(Map("Enquiry_Type__c" -> JsString(reason.value))))
+    } yield recentCase
```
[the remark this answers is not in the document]
twrichards: ✅

handlers/cancellation-sf-cases/src/main/scala/com/gu/salesforce/cases/SalesforceCase.scala
twrichards: ✅

handlers/cancellation-sf-cases/src/main/scala/com/gu/cancellation/sf_cases/Handler.scala
twrichards: i've managed to do this with andThen but it first needs to be tupled - perhaps we can refine it tomorrow

handlers/cancellation-sf-cases/src/main/scala/com/gu/salesforce/cases/SalesforceCase.scala
```diff
@@ -11,70 +12,83 @@ object SalesforceCase extends Logging {
...
   implicit val caseWithIdReads = Json.reads[CaseWithId]
 
-  object Raise {
+  case class ContactId(value: String) extends AnyVal
+  implicit val formatContactId = Jsonx.formatInline[ContactId]
+
+  case class SubscriptionId(value: String) extends AnyVal
+  implicit val formatSubscriptionId = Jsonx.formatInline[SubscriptionId]
+
+  case class CaseOrigin(value: String = "Self Service") extends AnyVal
```
johnduffell: To look in terms of dependencies, we are passing this as a parameter but the type only ever has one inhabitant so conveys no information.  if we want to add another possibility for some reason in future, we can do so.
To look at it another way, I prefer to apply parameters (partially) as soon as we have them and only wait until later if we don't have that yet.  In this case we already know it at the wiring stage, so it would be in the first parameter group.  But unless we would need to substite it for testing, for trivial things we would just hard code the dependency directly.
So overall I'm personally strongly in favour of doing away with this case class completely and simply hard coding it in or around the query string.

handlers/cancellation-sf-cases/src/main/scala/com/gu/cancellation/sf_cases/Handler.scala
twrichards: good point - changed to updateCaseReason in next push

handlers/cancellation-sf-cases/src/main/scala/com/gu/salesforce/cases/SalesforceCase.scala
twrichards: ✅

handlers/cancellation-sf-cases/src/main/scala/com/gu/cancellation/sf_cases/Handler.scala
```diff
@@ -60,23 +60,27 @@ object Handler extends Logging {
     final case class SubscriptionName(value: String) extends AnyVal
     implicit val formatSubscriptionName = Jsonx.formatInline[SubscriptionName]
 
+    final case class ContactIdContainer(Id: String)
+    implicit val readsContactIdContainer = Json.reads[ContactIdContainer]
+    final case class SubscriptionIdContainer(Id: String)
```
johnduffell: if this has to be Id because it's part of the json structure I'd be tempted to use the word Wire rather than Container

johnduffell:
great it's been useful to discuss code structure so thanks but now let's get this out and deal with non effects tests and perhaps splitting into a few more files later! 👍 🚀

PR #161, opened 2018-07-26
Sqs send effect

pvighi:
This adds the sqs effects module to send messages to an sqs queue. 
This is just mostly code copied from support workers with the unnecessary bits taken out.
Also this is 99% @johnduffell's branch that I took over.

### DIFF
--- a/lib/effects-sqs/src/main/scala/com/gu/effects/sqs/AwsSQSSend.scala
+++ b/lib/effects-sqs/src/main/scala/com/gu/effects/sqs/AwsSQSSend.scala
@@ -1,0 +1,51 @@
+package com.gu.effects.sqs
+
+import com.amazonaws.auth._
+import com.amazonaws.auth.profile.ProfileCredentialsProvider
+import com.amazonaws.regions.Regions
+import com.amazonaws.services.sqs.AmazonSQSAsyncClientBuilder
+import com.amazonaws.services.sqs.model.SendMessageRequest
+import org.apache.log4j.Logger
+
+import scala.concurrent.Future
+
+object AwsSQSSend {
+
+  val logger = Logger.getLogger(getClass.getName)
+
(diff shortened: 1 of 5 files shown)


lib/effects-sqs/src/main/scala/com/gu/effects/sqs/AwsSQSSend.scala
```diff
@@ -0,0 +1,52 @@
...
+    logger.info(s"Sending message to SQS queue $queueUrl")
+    val messageResult = AwsAsync(sqsClient.sendMessageAsync, new SendMessageRequest(queueUrl, payload.value))
+    messageResult.recover {
+      case throwable =>
+        logger.error("Failed to send message due to $queueUrl due to:", throwable)
+        throw throwable
+    }.map { result =>
+      logger.info(s"Successfully sent message to $queueUrl: $result")
+      result
+    }
```
johnduffell: nit pick would be I'd be tempted to change to the recover/map into `transform` with `case Success(...)` `case Failure(...)` just to keep it all at the same level.  `transform` is kind of a `bimap` of the `Try` part of the future.

project/Dependencies.scala
```diff
@@ -9,6 +9,7 @@ object Dependencies {
   val scalatest = "org.scalatest" %% "scalatest" % "3.0.1" % "test"
   val jacksonDatabind = "com.fasterxml.jackson.core" % "jackson-databind" % "2.8.11.1"
   val awsS3 = "com.amazonaws" % "aws-java-sdk-s3" % "1.11.311"
+  val awsSQS = "com.amazonaws" % "aws-java-sdk-sqs" % "1.11.371"
```
johnduffell: should we be updating the libraries in general (not in this PR)

johnduffell:
👍 great

PR #162, opened 2018-07-30
refuse to merge contacts if there are conflicting identity ids

johnduffell:
This PR will receive an identity id from the input of the SF contact merging endpoint, and then check that no zuora account has a conflicting one.
If there is no specified identity id, it will make sure there are no identityids in any of the accounts.

The next stage is to remove and re-add the identity ids while doing the update, to avoid any sync issues.
@pvighi @paulbrown1982 @jacobwinch @abhichawla


johnduffell:
> handlers/sf-contact-merge/src/main/scala/com/gu/sf_contact_merge/EnsureNoAccountWithWrongIdentityId.scala

I have just pushed some coverage to this file as it was missed (also removed extra unneeded/complex parameters)

handlers/sf-contact-merge/src/main/scala/com/gu/sf_contact_merge/GetIdentityAndZuoraEmailsForAccounts.scala
```diff
@@ -0,0 +1,89 @@
+package com.gu.sf_contact_merge
```
johnduffell: this file is a rename and some changes from the file directly below, but github didn't realise, so it's harder to pick out the actual changes

handlers/sf-contact-merge/src/test/scala/com/gu/sf_contact_merge/EndToEndTest.scala
```diff
@@ -116,9 +127,9 @@ object EndToEndTest {
       |    ],
       |    "size": 2,
       |    "done": true
-      |}""".stripMargin
+      |}""".stripMargin.replaceAll("""\n""", "")
 
-  val updateAccountRequestBody = """{"crmId":"sfcont","sfContactId__c":"sfacc"}"""
+  val updateAccountRequestBody = """{"crmId":"sfacc","sfContactId__c":"sfcont"}"""
```
johnduffell: these were already blatantly the wrong way round, and I only spotted it because I further broke it to pass in the same ID to both and wondered why the test failed :facepalm:

handlers/sf-contact-merge/src/test/scala/com/gu/sf_contact_merge/GetEmailsTest.scala
```diff
@@ -12,17 +12,17 @@ class GetEmailsTest extends FlatSpec with Matchers {
...
+    val getContacts = GetEmails(querier)_
     val actual = getContacts(NonEmptyList(
       ContactId("cid1"),
       ContactId("cid2")
     ))
 
-    val expectedEmails = List(Some("peppa.pig@guardian.co.uk"), None)
-    actual.map(_.map(_.map(_.value))) should be(ClientSuccess(expectedEmails))
+    val expectedEmails = Map(ContactId("cid1") -> Some(EmailAddress("peppa.pig@guardian.co.uk")), ContactId("cid2") -> None)
+    actual should be(ClientSuccess(expectedEmails))
```
johnduffell: the map.map.map was a clue that getContacts was returning the wrong type

PR #163, opened 2018-07-30
add code to sf contact merge to remove and replace identity ids

johnduffell:
This pr is on the SF contact merge.  This is where we are trying to merge all SF accounts into one, where they are for a single person.

Since the most likely winning account is probably a subscription, but the most likely to have an identity account would be a online sub, we need to do the merge without messing up the identity id.

Since we have decided that the identity id and the sf contact represent one person, we should ensure we both merge the sf and the identity side.

This PR adds some logic to clean out all the identity ids before the merge (to prevent any unnecessary sync) and then add them all back once there is only a single SF account involved.  Th previous PR added the validation that it wouldn't lose anything in the process, this is actually doing the deed.

@pvighi @paulbrown1982 @jacobwinch

### DIFF
--- a/handlers/sf-contact-merge/src/main/scala/com/gu/sf_contact_merge/Handler.scala
+++ b/handlers/sf-contact-merge/src/main/scala/com/gu/sf_contact_merge/Handler.scala
@@ -4,22 +4,20 @@ import java.io.{InputStream, OutputStream}
 
 import com.amazonaws.services.lambda.runtime.Context
 import com.gu.effects.{GetFromS3, RawEffects}
-import com.gu.sf_contact_merge.GetContacts.{AccountId, IdentityId, SFContactId}
-import com.gu.sf_contact_merge.GetIdentityAndZuoraEmailsForAccounts._
-import com.gu.sf_contact_merge.TypeConvert._
-import com.gu.sf_contact_merge.UpdateAccountSFLinks.{CRMAccountId, SFPointer}
+import com.gu.sf_contact_merge.update.UpdateAccountSFLinks.{CRMAccountId, SFPointer}
+import com.gu.sf_contact_merge.update.{SetOrClearZuoraIdentityId, UpdateAccountSFLinks, UpdateSteps}
+import com.gu.sf_contact_merge.validation.GetContacts.{AccountId, IdentityId, SFContactId}
+import com.gu.sf_contact_merge.validation.{GetIdentityAndZuoraEmailsForAccounts, ValidationSteps}
 import com.gu.util.apigateway.ApiGatewayHandler.{LambdaIO, Operation}
 import com.gu.util.apigateway.{ApiGatewayHandler, ApiGatewayRequest, ApiGatewayResponse, ResponseModels}
 import com.gu.util.config.LoadConfigModule.StringFromS3
 import com.gu.util.config.{LoadConfigModule, Stage}
 import com.gu.util.reader.Types._
-import com.gu.util.resthttp.Types.ClientFailableOp
 import com.gu.util.zuora.SafeQueryBuilder.MaybeNonEmptyList
 import com.gu.util.zuora.{ZuoraQuery, ZuoraRestConfig, ZuoraRestRequestMaker}
 import okhttp3.{Request, Response}
 import play.api.libs.json.{Json, Reads}
 import scalaz.NonEmptyList
-import scalaz.syntax.traverse.ToTraverseOps
 
 object Handler {
 
@@ -44,8 +42,11 @@ object Handler {
       requests = ZuoraRestRequestMaker(getResponse, zuoraRestConfig)
       zuoraQuerier = ZuoraQuery(requests)
       wiredSteps = steps(
-        GetIdentityAndZuoraEmailsForAccounts(zuoraQuerier),
-        UpdateAccountSFLinks(requests)
+        ValidationSteps(GetIdentityAndZuoraEmailsForAccounts(zuoraQuerier)),
+        UpdateSteps(
+          SetOrClearZuoraIdentityId(requests),
+          UpdateAccountSFLinks(requests)
+        )
       ) _
       configuredOp = Operation.noHealthcheck(wiredSteps)
     } yield configuredOp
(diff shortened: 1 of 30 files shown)


handlers/sf-contact-merge/src/main/scala/com/gu/sf_contact_merge/Handler.scala
```diff
@@ -61,20 +62,16 @@ object Handler {
...
       accountIds <- MaybeNonEmptyList(mergeRequest.zuoraAccountIds)
         .toApiGatewayContinueProcessing(ApiGatewayResponse.badRequest("no account ids supplied"))
-      accountAndEmails <- getZuoraEmails(accountIds).toApiGatewayOp("get zuora emails")
-      _ <- AssertSameEmails(accountAndEmails.map(_.emailAddress))
-      _ <- EnsureNoAccountWithWrongIdentityId(accountAndEmails.map(_.account.identityId), mergeRequest.identityId)
-        .toApiGatewayReturnResponse(ApiGatewayResponse.notFound)
-      updateAccount = updateAccountSFLinks(mergeRequest.sFPointer)
-      _ <- accountIds.traverseU(updateAccount).toApiGatewayOp("updating all the accounts")
+      _ <- validation(accountIds, mergeRequest.identityId)
+      _ <- update(accountIds, mergeRequest.sFPointer, mergeRequest.identityId)
```
johnduffell: I have split the steps out into validation and update as separate steps,  this makes it more clear and theoretically testable.

handlers/sf-contact-merge/src/main/scala/com/gu/sf_contact_merge/Handler.scala
```diff
@@ -61,20 +64,17 @@ object Handler {
...
         .toApiGatewayContinueProcessing(ApiGatewayResponse.badRequest("no account ids supplied"))
-      accountAndEmails <- getZuoraEmails(accountIds).toApiGatewayOp("get zuora emails")
-      _ <- AssertSameEmails(accountAndEmails.map(_.emailAddress))
-      _ <- EnsureNoAccountWithWrongIdentityId(accountAndEmails.map(_.account.identityId), mergeRequest.identityId)
-        .toApiGatewayReturnResponse(ApiGatewayResponse.notFound)
-      updateAccount = updateAccountSFLinks(mergeRequest.sFPointer)
-      _ <- accountIds.traverseU(updateAccount).toApiGatewayOp("updating all the accounts")
+      _ <- validation(accountIds, mergeRequest.identityId)
+      _ <- update(mergeRequest.sFPointer)(accountIds, mergeRequest.identityId)
+        .toApiGatewayOp("update accounts with winning detils")
```
johnduffell: I have split the steps out into validationSteps and updateSteps as separate files below, this makes it more clear and theoretically testable.

handlers/sf-contact-merge/src/test/scala/com/gu/sf_contact_merge/EndToEndTest.scala
```diff
@@ -130,14 +130,20 @@ object EndToEndTest {
...
   val mock = new TestingRawEffects(postResponses = Map(
     POSTRequest("/action/query", accountQueryRequest) -> HTTPResponse(200, accountQueryResponse),
     POSTRequest("/action/query", contactQueryRequest) -> HTTPResponse(200, contactQueryResponse),
     POSTRequest("/accounts/2c92c0f9624bbc5f016253e573970b16", updateAccountRequestBody, "PUT") -> updateAccountResponse,
-    POSTRequest("/accounts/2c92c0f8644618e30164652a558c6e20", updateAccountRequestBody, "PUT") -> updateAccountResponse
+    POSTRequest("/accounts/2c92c0f8644618e30164652a558c6e20", updateAccountRequestBody, "PUT") -> updateAccountResponse,
+    POSTRequest("/accounts/2c92c0f9624bbc5f016253e573970b16", removeIdentityBody, "PUT") -> updateAccountResponse,
+    POSTRequest("/accounts/2c92c0f8644618e30164652a558c6e20", removeIdentityBody, "PUT") -> updateAccountResponse,
+    POSTRequest("/accounts/2c92c0f9624bbc5f016253e573970b16", addIdentityBody, "PUT") -> updateAccountResponse,
+    POSTRequest("/accounts/2c92c0f8644618e30164652a558c6e20", addIdentityBody, "PUT") -> updateAccountResponse
```
johnduffell: this test doesn't actually prove that any or all the PUTs are done (exactly once), it just proves that no other requests were made!  I could check the mock.requests var to make sure though.

handlers/sf-contact-merge/src/test/scala/com/gu/sf_contact_merge/update/identityid/GetIdentityIdForAccount.scala
```diff
@@ -0,0 +1,27 @@
+package com.gu.sf_contact_merge.update.identityid
```
johnduffell: this is in the test package - it's just to make sure the identity id was correctly set in the zuora effects test

handlers/sf-contact-merge/src/main/scala/com/gu/sf_contact_merge/Handler.scala
jacobwinch: minor: typo here

handlers/sf-contact-merge/src/main/scala/com/gu/sf_contact_merge/update/UpdateSteps.scala
```diff
@@ -0,0 +1,28 @@
...
+object UpdateSteps {
+  def apply(
+    setOrClearIdentityId: Option[IdentityId] => AccountId => Types.ClientFailableOp[Unit],
+    updateAccountSFLinks: AccountId => ClientFailableOp[Unit]
+  )(
+    accountIds: NonEmptyList[AccountId],
+    maybeUpdateIdentityId: Option[IdentityId]
+  ): ClientFailableOp[Unit] = {
+    for {
+      _ <- if (maybeUpdateIdentityId.isDefined) {
```
jacobwinch: minor: I find an if/else inside a for comprehension slightly hard to follow, so I would pull this into a small function

handlers/sf-contact-merge/src/test/scala/com/gu/sf_contact_merge/update/UpdateStepsTest.scala
```diff
@@ -0,0 +1,67 @@
...
+    val wired = UpdateSteps(setOrClearIdentityId, updateAccountSFLinks) _
+
+    val accountIds: NonEmptyList[GetContacts.AccountId] =
+      NonEmptyList(AccountId("account1"))
+
+    val maybeIdentityId: Option[GetContacts.IdentityId] = Some(IdentityId("newIdentityId"))
+
+    val actual = wired(accountIds, maybeIdentityId)
+
+    order.reverse should be(List("clear", "doLink", "addidentity"))
```
jacobwinch: This is a neat trick 👍

johnduffell: "trick" being the operative word 😆

johnduffell:
agree with your comment, I will be running this one at a time initially, but I would have to investigate every failure to make sure it hasn't messed anything up.  Having the "replace id" in a finally type block would make more sense.  If it's easy I'll do that.

PR #164, opened 2018-07-31
Email changes

pvighi:
send thank you email when CSRs create contribution subscriptions.
also some refactoring


handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/zuora/GetContacts.scala
```diff
@@ -0,0 +1,57 @@
...
+import com.gu.util.resthttp.Types.{ClientFailableOp}
+import play.api.libs.json.Json
+
+object GetContacts {
+
+  object WireModel {
+
+    case class ZuoraContact(firstName: String, lastName: String, workEmail: Option[String], country: Option[String])
+
+    case class ZuoraContacts(billToContact: ZuoraContact, soldToContact: ZuoraContact)
```
pvighi: The current version of the code only uses the billing contact. If we confirm that the soldTo contact is not needed  I will just remove the code that parses it into the response and just return a single contact

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/zuora/GetContacts.scala
```diff
@@ -0,0 +1,57 @@
...
+  import WireModel._
+
+  case class FirstName(value: String) extends AnyVal
+
+  case class LastName(value: String) extends AnyVal
+
+  case class Email(value: String) extends AnyVal
+
+  //TODO are the first or last names optional ?
+  case class Contact(firstName: FirstName, lastName: LastName, email: Option[Email], country: Option[Country])
```
pvighi: the Zuora UI won't let me save a contact without a first of last name. I'm not sure if I can assume all existing accounts will have them. I can run a query to confirm

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/zuora/GetContacts.scala
```diff
@@ -0,0 +1,57 @@
...
+  
+  //todo do we need both contacts ? also see what do to if one or both are missing
+  case class Contacts(billTo: Contact, soldTo: Contact)
```
pvighi: The current version of the code only uses the billing contact. If we confirm that the soldTo contact is not needed I will just remove the code that parses it into the response and just return a single contact

lib/effects-sqs/src/main/scala/com/gu/effects/sqs/AwsSQSSend.scala
```diff
@@ -18,7 +19,7 @@ object AwsSQSSend {
 
   case class Payload(value: String) extends AnyVal
 
-  def apply(queueName: QueueName)(payload: Payload)(implicit ex: ExecutionContext): Future[Unit] = {
```
pvighi: this class now imports the global execution context to avoid propagating it into the handler. I'm not sure if this is the best way to do it.
We now have the global execution context imported in two places: here and in the AsyncApiGatewayOp definition

lib/handler/src/main/scala/com/gu/util/reader/Types.scala
```diff
@@ -106,6 +106,12 @@ object Types extends Logging {
         case None => ReturnWithResponse(NoneResponse)
       }
 
+    def toApiGatewayReturnResponse(toApiResponse: A => ApiResponse): ApiGatewayOp[Unit] =
```
pvighi: I'm not sure why this shows as a difference, locally it tells my I didn't change this file

build.sbt
```diff
@@ -173,7 +173,7 @@ lazy val `identity-retention` = all(project in file("handlers/identity-retention
 
 lazy val `new-product-api` = all(project in file("handlers/new-product-api"))
   .enablePlugins(RiffRaffArtifact)
-  .dependsOn(zuora, handler, effectsDepIncludingTestFolder, testDep)
+  .dependsOn(zuora, handler,`effects-sqs`, effectsDepIncludingTestFolder, testDep)
```
johnduffell: tiny: space missing

handlers/new-product-api/cfn.yaml
```diff
@@ -56,6 +56,14 @@ Resources:
                             Resource:
                             - !Sub arn:aws:s3:::gu-reader-revenue-private/membership/support-service-lambdas/${Stage}/zuoraRest-${Stage}.*.json
                             - !Sub arn:aws:s3:::gu-reader-revenue-private/membership/support-service-lambdas/${Stage}/trustedApi-${Stage}.*.json
+                - PolicyName: SqsMessages
```
johnduffell: could call it SQSAddEmailRequest or something

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/Handler.scala
```diff
@@ -3,24 +3,32 @@ package com.gu.newproduct.api.addsubscription
...
+import com.gu.newproduct.api.addsubscription.zuora.GetPaymentMethod.DirectDebit
+import com.gu.newproduct.api.addsubscription.zuora.{CreateSubscription, GetAccount, GetContacts}
 import com.gu.util.Logging
 import com.gu.util.apigateway.ApiGatewayHandler.{LambdaIO, Operation}
 import com.gu.util.apigateway.ResponseModels.ApiResponse
 import com.gu.util.apigateway.{ApiGatewayHandler, ApiGatewayRequest, ApiGatewayResponse}
 import com.gu.util.config.LoadConfigModule.StringFromS3
 import com.gu.util.config.{LoadConfigModule, Stage}
+import com.gu.util.reader.AsyncTypes._
 import com.gu.util.reader.Types._
```
johnduffell: incidentally, "reader" monad was destroyed a long time ago from this code base. and "util" is not very specific. But we didn't get round to renaming anything!  This is part of the interface with API gateway handler (it's the name of the object that we return to the handler) so it should probably be in something like com.gu.apigateway package and maybe called something like AsyncTypeConvert ?

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/Handler.scala
```diff
@@ -34,12 +42,13 @@ object Handler extends Logging {
 object Steps {
 
   def addSubscriptionSteps(
-    prerequesiteCheck: AddSubscriptionRequest => ApiGatewayOp[Unit],
-    createMonthlyContribution: CreateReq => ClientFailableOp[SubscriptionName]
-  )(apiGatewayRequest: ApiGatewayRequest): ApiResponse = {
+    prerequisiteCheck: AddSubscriptionRequest => AsyncApiGatewayOp[ValidatedFields],
+    createMonthlyContribution: CreateReq => ClientFailableOp[SubscriptionName],
+    sendConfirmationEmail: (ZuoraAccountId, Currency, Option[DirectDebit], Int) => AsyncApiGatewayOp[Unit]
```
johnduffell: good apart from may be better to wrap amountMinorUnits in a nice class.
one handy way to get it over with early - in sf-contact-merge I treated the request that we do bodyAsCaseClass as a "wire" model and did a second step to convert to domain friendly objects.
https://github.com/guardian/zuora-auto-cancel/blob/db47d3c158f5fe806501a48e83e9920dcfc91bad/handlers/sf-contact-merge/src/main/scala/com/gu/sf_contact_merge/Handler.scala#L72
I didn't have any validation, but it could have been flatmapped if needed.

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/Handler.scala
```diff
@@ -34,12 +42,13 @@ object Handler extends Logging {
 object Steps {
 
   def addSubscriptionSteps(
-    prerequesiteCheck: AddSubscriptionRequest => ApiGatewayOp[Unit],
-    createMonthlyContribution: CreateReq => ClientFailableOp[SubscriptionName]
-  )(apiGatewayRequest: ApiGatewayRequest): ApiResponse = {
+    prerequisiteCheck: AddSubscriptionRequest => AsyncApiGatewayOp[ValidatedFields],
```
johnduffell: I guess this was what you meant about carrying on the refactoring - as nothing has changed in the behaviour of prerequisiteCheck but now it is returning a more powerful type.  Although in reality we would use an async http client in the end, I don't think it was necessary to switch it at this stage.
We should retain the ability to convert freely and in a tidy way "up" from the less powerful ApiGatewayOp to the AsyncApiGatewayOp, and really we should always use the least powerful one as possible.

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/Handler.scala
```diff
@@ -48,7 +57,12 @@ object Steps {
         request.acquisitionSource,
         request.createdByCSR
       )
-      subscriptionName <- createMonthlyContribution(req).toApiGatewayOp("create monthly contribution")
+      subscriptionName <- createMonthlyContribution(req).toAsyncApiGatewayOp("create monthly contribution")
+      directDebit = validatedFields.paymentMethod match {
```
johnduffell: `maybeDirectDebitInfo` ?

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/email/SendConfirmationEmail.scala
```diff
@@ -0,0 +1,60 @@
...
+
+object SendConfirmationEmail extends Logging {
+
+  def getPayload(
+    now: LocalDate,
+    contacts: Contacts,
+    currency: Currency,
+    directDebit: Option[DirectDebit],
+    amountMinorUnits: Int
+  ): AsyncApiGatewayOp[ETPayload] = {
```
johnduffell: I don't think this function is really get-payload,  it's actually just an call to `option.toAsyncApiGatewayContinueProcessing(ApiGatewayResponse.successfulExecution)`

we could get rid of this wrapper if we implement the above `toAsyncApiGatewayContinueProcessing` (however we want to name it)

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/email/SendConfirmationEmail.scala
```diff
@@ -0,0 +1,60 @@
...
+    directDebit: Option[DirectDebit],
+    amountMinorUnits: Int
+  ) = {
+
+    val response = for {
+      contacts <- getContacts(accountid).toAsyncApiGatewayOp("getting contacts from Zuora")
+      etPayload <- getPayload(now(), contacts, currency, directDebit, amountMinorUnits)
+      payloadString = Json.prettyPrint(Json.toJson(etPayload))
+      sendMessageResult <- sqsSend(Payload(payloadString)).toAsyncApiGatewayOp("sending sqs message")
+    } yield sendMessageResult
```
johnduffell: line 52 is just turning a domain object into a wire object, line 53 is turning a wire object into a String and line 54 is doing the op, so all very much digging down from scala into the low level.  however line 51 is at a higher level of as it's getting some information from zuora (which contains all the json/wire/domain stuff hidden)

I feel like these last 3 lines should at least be combined into a separate EmailContent => AyncApiGatewayOp[Unit] function (which could ether be called directly or passed in as a function in the parameters), and this for comprehension should be more like a 'steps' and call both of those.

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/email/SendConfirmationEmail.scala
```diff
@@ -0,0 +1,60 @@
...
+
+    response.replace(ContinueProcessing(()))
```
johnduffell: should we logger.error any kind of ReturnResponse already present at this point (not that we'd notice even if it did log an error) or write to a cloudwatch warnings metric?  just wondering how we'd know if this was failing to add things to the queue where the email address was present (assuming we are concerned about that)

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/zuora/GetContacts.scala
```diff
@@ -0,0 +1,57 @@
...
+
+    def fromWire(zuoraAccount: ZuoraContacts): Contacts = {
+      val zBillto = zuoraAccount.billToContact
+      val zSoldTo = zuoraAccount.soldToContact
+      Contacts(
+        billTo = Contact(
+          FirstName(zBillto.firstName),
+          LastName(zBillto.lastName),
+          zBillto.workEmail.map(Email),
+          zBillto.country.flatMap(CountryGroup.countryByNameOrCode(_))
```
johnduffell: we might as well also map this to the _.alpha2 at this stage because it looks like we don't use any other

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/zuora/GetPaymentMethod.scala
```diff
@@ -17,19 +25,65 @@ object GetPaymentMethod {
...
+      toStatus(paymentMethodWire.PaymentMethodStatus),
+      BankAccountName(accountName),
+      BankAccountNumberMask(accountNumberMask),
+      SortCode(sortCode),
+      MandateId(mandateId)
+    )
+
+    def toPaymentMethod: ClientFailableOp[PaymentMethod] = {
+      val methodType = stringToType.getOrElse(Type, Other)
+      if (methodType == BankTransfer) toDirectDebit(this)
```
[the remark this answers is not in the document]
johnduffell: this feels a bit like a pattern match, not sure if it would look neater this way?
```scala
def toPaymentMethod: ClientFailableOp[PaymentMethod] = 
  stringToType.getOrElse(Type, Other) match {
    case BankTransfer => toDirectDebit
    case other => ClientSuccess(NonDirectDebitMethod(toStatus(PaymentMethodStatus), methodType))
  }

File: handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/zuora/GetPaymentMethod.scala [hunk 17]
```diff
@@ -17,19 +25,65 @@ object GetPaymentMethod {
       "BankTransfer" -> BankTransfer
     )
 
-    def toPaymentMethod: PaymentMethod =
-      PaymentMethod(
-        status = if (PaymentMethodStatus == "Active") ActivePaymentMethod else NotActivePaymentMethod,
-        paymentMethodType = stringToType.getOrElse(Type, Other)
-      )
+    private def toStatus(statusString: String) = if (statusString == "Active") ActivePaymentMethod else NotActivePaymentMethod
```
johnduffell: toStatus gets a free dependency on `this` but it doesn't need it, so this weould be better outside the class (in  a companion object perhaps)

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/email/SendConfirmationEmail.scala
```diff
@@ -5,51 +5,52 @@ import java.time.LocalDate
...
+    amountMinorUnits: AmountMinorUnits
+  ): AsyncApiGatewayOp[ETPayload[ContributionFields]] = {
+    val maybeContributionFields = ContributionFields.fromData(amountMinorUnits, now, currency, directDebit, billTo)
+
+    maybeContributionFields.map { fields =>
+      val payload = ETPayload(fields.EmailAddress, fields)
+      AsyncApiGatewayOp(ContinueProcessing(payload))
+    }.getOrElse {
+      logger.info("Not enough data in zuora account to send contribution thank you email, skipping")
+      AsyncApiGatewayOp(ReturnWithResponse(ApiGatewayResponse.successfulExecution))
```
johnduffell: general thought:

yeah, I start to think more about that the discussion earlier that we shouldn't really be using any kind of "API gateway op" or "api gateway response" here (given that regardless of what we do, the result won't have any effect on any kind of api gateway response).

We are probably re-using the ApiGateway and ClientFailable ops slightly inappropriately in a fair amount of code.  e.g. I can imagine if we suddenly decided that we won't use API gateway at all, instead we will use some kind of incoming/outgoing queues system, we would expect to be able to reuse the steps and other code except the outer layer and wiring.  At least with the ClientFailable we are a bit more generic.  But if that contains a Future we won't want to overuse it.

handlers/new-product-api/src/test/scala/com/gu/newproduct/api/addsubscription/email/ETPayloadTest.scala
```diff
@@ -1,93 +1,28 @@
...
-      """.stripMargin
-    Json.toJson(testPayload) shouldBe Json.parse(expected)
+  case class TestPayload(data: String)
 
+  object TestPayload {
+    implicit val writes = Json.writes[TestPayload]
   }
 
-  it should "serialise  direct debit contributions Email to json" in {
+  it should "serialise payload to json" in {
```
johnduffell: good these tests are much more separated now 💯

lib/handler/src/main/scala/com/gu/util/reader/AsyncTypes.scala
```diff
@@ -58,4 +58,8 @@ object AsyncTypes extends Logging {
   implicit class SyncToAsync[A](apiGatewayOp: ApiGatewayOp[A]) {
     def toAsync: AsyncApiGatewayOp[A] = AsyncApiGatewayOp(Future.successful(apiGatewayOp))
   }
+
+  implicit class FSyncToAsync[A, B](f: B => ApiGatewayOp[A]) {
+    def toAsync: B => AsyncApiGatewayOp[A] = f andThen (_.toAsync)
```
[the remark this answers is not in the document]
pvighi: yeah I forgot to tidy up this. at the very least the function should go from A  to something that includes B. But renaming the params to something more useful is even better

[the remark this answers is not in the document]
pvighi: Also I like the idea than the only way someone would be looking at this code is by mistake haha

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/email/ContributionFields.scala
```diff
@@ -0,0 +1,69 @@
...
+        currency = currency.glyph,
+        edition = billTo.country.map(_.alpha2).getOrElse(""),
+        name = s"${billTo.firstName.value} ${billTo.lastName.value}",
+        product = "monthly-contribution",
+        `account name` = directDebit.map(_.accountName.value),
+        `account number` = directDebit.map(_.accountNumberMask.value),
+        `sort code` = directDebit.map(x => hyphenate(x.sortCode.value)),
+        `Mandate ID` = directDebit.map(_.mandateId.value),
+        `first payment date` = directDebit.map { _ =>
+          val firstPayment = created.plusDays(10) // todo the contract acceptance date should be at the same time as the first payment date
```
pvighi: I talked to Paul and this first payment day should match the contract acceptance date on the subscription we create. But we might not be setting that correctly ( I think we are just using the same date as the effective date at the moment)

lib/handler/src/main/scala/com/gu/util/reader/AsyncTypes.scala
```diff
@@ -0,0 +1,65 @@
...
+import scala.concurrent.Future
+
+object AsyncTypes extends Logging {
+
+  case class AsyncApiGatewayOp[A](underlying: Future[ApiGatewayOp[A]]) {
+
+    def map[B](f: A => B): AsyncApiGatewayOp[B] = AsyncApiGatewayOp {
+      underlying.map {
+        case ContinueProcessing(a) => ContinueProcessing(f(a))
+        case returnWithResponse: ReturnWithResponse => returnWithResponse
```
johnduffell: we might be able to implement this as `underlying.map(_.map(f))` as `ApiGatewayOp` already has the map implementation

lib/handler/src/main/scala/com/gu/util/reader/AsyncTypes.scala
```diff
@@ -0,0 +1,65 @@
...
+      }
+    }
+
+    def flatMap[B](f: A => AsyncApiGatewayOp[B]): AsyncApiGatewayOp[B] = AsyncApiGatewayOp {
+      underlying.flatMap {
+        case ContinueProcessing(a) => f(a).underlying
+        case returnWithResponse: ReturnWithResponse => Future.successful(returnWithResponse)
```
johnduffell: interestingly, `AsyncApiGatewayOp` is basically the same as an `EitherT[Future, ApiResponse, A]`.  That kind of makes it clear why we are not using the EitherT directly! 🙈 

from that it follows that we couldn't do the above map.map trick, but it would be possible (but probably a bad idea for clarity) to implement flatMap from the monad transformer: 
```scala
AsyncApiGatewayOp {
  EitherT(underlying.map(_.toDisjunction)).flatMap { f.andThen { apiAsync => 
      EitherT(apiAsync.underlying.map(_.toDisjunction))))
    }
  }.run.map(_.toApigatewayOp/*from disjunction*/)
}
```

lib/handler/src/main/scala/com/gu/util/reader/AsyncTypes.scala
```diff
@@ -0,0 +1,65 @@
...
+    def replace(replacement: ApiGatewayOp[A]): AsyncApiGatewayOp[A] = AsyncApiGatewayOp(underlying.map(_ => replacement))
+  }
+
+  implicit class AsyncApiResponseOps(apiGatewayOp: AsyncApiGatewayOp[ApiResponse]) {
+    def apiResponse: Future[ApiResponse] = apiGatewayOp.underlying.map(x => x.apiResponse)
+  }
+
+  object AsyncApiGatewayOp {
+    def apply[A](continue: ContinueProcessing[A]): AsyncApiGatewayOp[A] = AsyncApiGatewayOp(Future.successful(continue))
+    def apply[A](response: ReturnWithResponse): AsyncApiGatewayOp[A] = AsyncApiGatewayOp(Future.successful(response))
```
johnduffell: minor thought:
I'm not 100% keen on too much overloading of `apply` in the general case - just in case some refactoring changes one to the other by mistake and still compiles.
also, for a small reduction in the bracket count `(...)` we could have `AsyncApiGatewayOp.continueProcessing(a)` and `AsyncApiGatewayOp.returnWithResponse(apiResponse)` 
for an actual ApiGatewayOp we would use the neater? postfix `.toAsync` anyway

lib/handler/src/main/scala/com/gu/util/reader/AsyncTypes.scala
```diff
@@ -0,0 +1,65 @@
...
+    def apply[A](underlying: Future[ApiGatewayOp[A]]): AsyncApiGatewayOp[A] = {
+      val f = underlying recover {
```
johnduffell: `f` woudl be `successfulFutureFailTo500Error` or something

lib/handler/src/main/scala/com/gu/util/reader/AsyncTypes.scala
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  implicit class AsyncApiResponseOps(apiGatewayOp: AsyncApiGatewayOp[ApiResponse]) {
+    def apiResponse: Future[ApiResponse] = apiGatewayOp.underlying.map(x => x.apiResponse)
+  }
+
+  object AsyncApiGatewayOp {
+    def apply[A](continue: ContinueProcessing[A]): AsyncApiGatewayOp[A] = AsyncApiGatewayOp(Future.successful(continue))
+    def apply[A](response: ReturnWithResponse): AsyncApiGatewayOp[A] = AsyncApiGatewayOp(Future.successful(response))
+    def apply[A](underlying: Future[ApiGatewayOp[A]]): AsyncApiGatewayOp[A] = {
```
johnduffell: `underlying` could be `possblyFailedFuture`

johnduffell:
well, great job plugging all those futures together successfully and coming up with a nice result.  I think we have learnt a bit here, although having Futures leaking into tests that don't need them is a bit concerning.
I have done some work on whether it's possible to abstract away from the Future in tests, without having a "tagless final" or a "free monad" and made a little progress [here](https://github.com/johnduffell/sequence-hlist/blob/master/src/main/scala/profunctor/HttpMonoid.scala) however I haven't thought of a way of making it easy to understand, and the level of refactoring needed to separate out the code that's currently combined together and shouldn't really be would be large.

re PR size overall I'm glad we decided that refactoring everywhere in the code to use Futures was a bit big for one PR!  Normally we try to split the PRs are much as possible into either boring mass refactors or a blob of dense important code, which works very well.  As you mentioned earlier maybe this is a little on the tricky side due to the size.
This PR includes firstly the new ASync api gateway handler code and types, secondly the thing that wraps the SQS client into a thank you email client (plus the getContact thing), and finally also the bit that wires those into the steps including the type converters.  If those had been 3 or 4 separate PRs it would have helped no end.  But I can see that it was hard to see to the end without some prototyping.  Also I don't think this is a risky PR for shipping, as it isn't a big PR in terms of touching unrelated code.

I'll let you get through any comments you want to from the above and then I can take another quick look tomorow.

lib/handler/src/main/scala/com/gu/util/reader/AsyncTypes.scala
pvighi: aren't all futures possiblyFailed by default ?

johnduffell: they are when they're incoming, but by the time we wrap them into our thing, they must be a success (but may contain a ReturnWithResponse)
Just we use "underlying" for both situations, so I'd prefer to distinguish them.

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/Handler.scala
```diff
@@ -36,33 +36,43 @@ object Handler extends Logging {
...
-
 }
 
 object Steps {
 
   def addSubscriptionSteps(
     prerequisiteCheck: AddSubscriptionRequest => AsyncApiGatewayOp[ValidatedFields],
     createMonthlyContribution: CreateReq => ClientFailableOp[SubscriptionName],
-    sendConfirmationEmail: (ZuoraAccountId, Currency, Option[DirectDebit], AmountMinorUnits) => AsyncApiGatewayOp[Unit]
+    sendConfirmationEmail: ContributionsEmailData => AsyncApiGatewayOp[Unit],
```
johnduffell: good idea to wrap like this for neatness

lib/handler/src/main/scala/com/gu/util/reader/AsyncTypes.scala
pvighi: makes sense, I'll change it

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/email/SendConfirmationEmail.scala
johnduffell: already discussed this one in person

johnduffell:
now we have more comments than files changed in this pr!
👍 🚀 🐎 🏎 🏃 🚢

PR #165, opened 2018-08-02
only use first name for contribution email

pvighi:
It looks like when [signing up through the web](https://github.com/guardian/support-workers/blob/master/src/main/scala/com/gu/support/workers/lambdas/SendThankYouEmail.scala#L51) the "name" field sent to exact target is set to just the first name, not the full name. This field is not included in the email template used right now anyway so It should not make much of a difference.


johnduffell:
👍 🚀

PR #166, opened 2018-08-02
added empty product-catalog endpoint

pvighi:


handlers/new-product-api/cfn.yaml
```diff
@@ -55,7 +55,6 @@ Resources:
                             Action: s3:GetObject
                             Resource:
                             - !Sub arn:aws:s3:::gu-reader-revenue-private/membership/support-service-lambdas/${Stage}/zuoraRest-${Stage}.*.json
-                            - !Sub arn:aws:s3:::gu-reader-revenue-private/membership/support-service-lambdas/${Stage}/trustedApi-${Stage}.*.json
```
pvighi: a refactor some time ago removed the need for the trusted api token on lambdas that don't validate using that

johnduffell: thanks for spotting that

johnduffell:
👍 as discussed good idea to ship it so @david-pepper can start work, then we can separate the jars with some SBT wizardry so avoid pulling the SQS client etc into this endpoint

PR #167, opened 2018-08-02
added Catalog classes and some dummy catalog response

pvighi:


handlers/new-product-api/src/main/scala/com/gu/newproduct/api/productcatalog/Catalog.scala
```diff
@@ -0,0 +1,56 @@
...
+  id: String,
+  label: String,
+  startDateRules: Option[StartDateRules] = None
+)
+
+case class StartDateRules(
+  daysOfWeek: Option[List[DayOfWeek]] = None,
+  cutOffDay: Option[DayOfWeek] = None,
+  minDelay: Option[Int] = None,
+  maxDelay: Option[Int] = None,
```
johnduffell: looks good. just a couple of ideas for clarity:

minDelay could be `minDaysAfterNextCutOff`
maxDelay - the cutoffday itself is not directly relevant to the max, so we could either go relative to today `maxDaysAfterToday`, or to the earliest start `availablePeriodInDays`

also, what would it mean if minDelay is None as opposed to 0? do we need the Option?

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/productcatalog/Catalog.scala
```diff
@@ -0,0 +1,56 @@
...
+)
+
+
+case class Group(label: String, products: List[ProductInfo])
+
+case class Catalog(groups: List[Group])
+
+object DayOfWeek {
+  implicit val writes: Writes[DayOfWeek] = { (day: DayOfWeek) => JsString(day.toString) }
```
johnduffell: general comment:

it's very neat, but I'm not a fan of toString, because literally everything has one even if it's not what we expected, and the output is not generated anywhere obious so we can break the output by adding fields (day of week index?) etc.  In this case maybe no problem and it'll get picked up by the tests, but we could consider to add a `def dayName` to the trait, and get it to do a "toString" just to at least make it own its way of using toString?

johnduffell:
great! 👍 good work.
I'm sure there are better ways of checking the json structure (if we decide to stop pretty printing the json responses we don't want things to break) but overall it's nice clear result

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/productcatalog/Catalog.scala
pvighi: This is supposed to be the "wires" model so I added options all over the place so that if something is optional it just doesn't show up in the json. This is just a hard coded response so that David can start working, my plan is to have a nicer model that we use to actually perform the validation and that can also be converted into this to be serialised in the catalog response.
I was thinking in terms of the min and max delay working in the same way, as offsets from the cut off date. In this context a 0 is different from none because None would mean infinity 0  could be interpreted as you are only allowed to use the cut off date and nothing else.
I thought it might be simpler to make both limits but I can change it to be an amount of days from the minDelay if it better.

PR #170, opened 2018-08-03
add ses email send effect so we can make an anomalies email

johnduffell:
It would be nice to have an anomalies email that's a bit easier to add things to and ship than the old one.

This is the first stage.  It can send an email in DEV mode.  It's not connected to any lambda yet.

If anyone has ideas of things we can hit with a single REST call (or whatever) that would be actually useful, let me know.

### DIFF
--- a/lib/effects-ses/src/main/scala/com/gu/effects/ses/AWSAsyncHandler.scala
+++ b/lib/effects-ses/src/main/scala/com/gu/effects/ses/AWSAsyncHandler.scala
@@ -1,9 +1,9 @@
-package com.gu.effects.sqs
+package com.gu.effects.ses
 
 import java.util.concurrent.{Future => JFuture}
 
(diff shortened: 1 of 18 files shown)


lib/effects-ses/src/main/scala/com/gu/effects/ses/AWSAsyncHandler.scala
```diff
@@ -0,0 +1,43 @@
+package com.gu.effects.ses
```
johnduffell: this is just a copy from SQS package, we should really have an AWS async package to share both but with all the sbt boilerplate I couldn't quite face it for one class.

johnduffell:
the reduction in coverage is because I added SQS and SES to the build, and neither are unit tested (as they are specifically the side effects only)

PR #171, opened 2018-08-03
do the SF PATCH request against SF not zuora!

johnduffell:
@pvighi 
I accidentally used the zuoraRequests object rather than the sf one for the PATCH request.  The type checker didn't really save me from that error.  Perhaps the SFAuthenticate or the ZuoraRestRequestMaker should return a suitable type parameter so we can't do this!

I have also tweaked it all to use the non curried functions in all the wiring, see what you think.

### DIFF
--- a/handlers/sf-contact-merge/src/main/scala/com/gu/sf_contact_merge/getaccounts/GetIdentityAndZuoraEmailsForAccountsSteps.scala
+++ b/handlers/sf-contact-merge/src/main/scala/com/gu/sf_contact_merge/getaccounts/GetIdentityAndZuoraEmailsForAccountsSteps.scala
@@ -15,10 +15,10 @@ object GetIdentityAndZuoraEmailsForAccountsSteps {
     emailAddress: Option[EmailAddress]
   )
 
-  def apply(zuoraQuerier: ZuoraQuerier)(accountIds: NonEmptyList[AccountId]): ClientFailableOp[List[IdentityAndSFContactAndEmail]] = {
+  def apply(zuoraQuerier: ZuoraQuerier, accountIds: NonEmptyList[AccountId]): ClientFailableOp[List[IdentityAndSFContactAndEmail]] = {
 
-    val getEmails = GetEmails(zuoraQuerier)_
-    val getContacts = GetContacts(zuoraQuerier)_
+    val getEmails: NonEmptyList[GetEmails.ContactId] => ClientFailableOp[Map[GetEmails.ContactId, Option[EmailAddress]]] = GetEmails(zuoraQuerier, _)
+    val getContacts: NonEmptyList[AccountId] => ClientFailableOp[Map[GetEmails.ContactId, GetContacts.IdentityAndSFContact]] = GetContacts(zuoraQuerier, _)
 
     for {
       identityForBillingContact <- getContacts(accountIds)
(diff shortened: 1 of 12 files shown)


handlers/sf-contact-merge/src/main/scala/com/gu/sf_contact_merge/getaccounts/GetIdentityAndZuoraEmailsForAccountsSteps.scala
```diff
@@ -15,10 +15,10 @@ object GetIdentityAndZuoraEmailsForAccountsSteps {
     emailAddress: Option[EmailAddress]
   )
 
-  def apply(zuoraQuerier: ZuoraQuerier)(accountIds: NonEmptyList[AccountId]): ClientFailableOp[List[IdentityAndSFContactAndEmail]] = {
+  def apply(zuoraQuerier: ZuoraQuerier, accountIds: NonEmptyList[AccountId]): ClientFailableOp[List[IdentityAndSFContactAndEmail]] = {
 
-    val getEmails = GetEmails(zuoraQuerier)_
-    val getContacts = GetContacts(zuoraQuerier)_
+    val getEmails: NonEmptyList[GetEmails.ContactId] => ClientFailableOp[Map[GetEmails.ContactId, Option[EmailAddress]]] = GetEmails(zuoraQuerier, _)
+    val getContacts: NonEmptyList[AccountId] => ClientFailableOp[Map[GetEmails.ContactId, GetContacts.IdentityAndSFContact]] = GetContacts(zuoraQuerier, _)
```
johnduffell: it has to be able to work out the type from the context, otherwise you get a 
"missing parameter type for expanded function" and you have to either specify it after the _, or provide the context.

handlers/sf-contact-merge/src/test/scala/com/gu/sf_contact_merge/getaccounts/GetZuoraEmailsForAccountsEffectsTest.scala
```diff
@@ -22,7 +22,7 @@ class GetZuoraEmailsForAccountsEffectsTest extends FlatSpec with Matchers {
     val actual = for {
       zuoraRestConfig <- LoadConfigModule(Stage("DEV"), GetFromS3.fetchString)[ZuoraRestConfig].toApiGatewayOp("parse config")
       zuoraQuerier = ZuoraQuery(ZuoraRestRequestMaker(RawEffects.response, zuoraRestConfig))
-      getZuoraEmailsForAccounts = GetIdentityAndZuoraEmailsForAccountsSteps(zuoraQuerier) _
+      getZuoraEmailsForAccounts = GetIdentityAndZuoraEmailsForAccountsSteps(zuoraQuerier, _: NonEmptyList[AccountId])
```
pvighi: did you need to specify the type of the omitted param here ?

johnduffell: as discussed, yes because the type checker seems to want to work out the type of `_` before it can call the function.
```
error] /Users/jduffell/ws/zuora-auto-cancel/handlers/sf-contact-merge/src/test/scala/com/gu/sf_contact_merge/getaccounts/GetZuoraEmailsForAccountsEffectsTest.scala:25:91: missing parameter type for expanded function ((x$1: <error>) => GetIdentityAndZuoraEmailsForAccountsSteps(zuoraQuerier, x$1))
[error]       getZuoraEmailsForAccounts = GetIdentityAndZuoraEmailsForAccountsSteps(zuoraQuerier, _)
[error]                                                                                           ^
```

changing it to this makes it compile(!)
```scala
getZuoraEmailsForAccounts =
  (GetIdentityAndZuoraEmailsForAccountsSteps.apply _).curried(zuoraQuerier)
```
but I probably won't on functions with only the one param

PR #172, opened 2018-08-06
DO not merge! "Mill" build tool experiment.

johnduffell:
This is just open for comments.

I have observed that no one really knows how to make SBT do what it should.  Things that should be very easy are quite hard, due to the lack of concepts, even though SBT is kind of scala and has Intellij support.  Then I read http://www.lihaoyi.com/post/SowhatswrongwithSBT.html

Anyway I had a bit of of a play with Mill, which seems to have some reasonable momentum at the moment and from a dev I've heard of https://github.com/lihaoyi/mill

basically to use this you have to `brew install mill` and then you can `mill mill.scalalib.GenIdea/idea` in the root of the project.
Then you can ` mill handlers.sf-contact-merge.assembly` to make an out.jar.  And finally I have proved hopefully that it's possible to run our non effects tests only when running test target.  And I have managed to make a rudimentary `mill handlers.sf-contact-merge.riffRaffArtifact` which doesn't actually upload anything but it proves we can run the right code.

Really this PR is a mess, just a proof that the concept makes some sense and that we should be able to implement what we need.  But I have never seen someone say they really know whwat they're doing with SBT, so I don't think it really is helping us.  So it would be sensible to look at alternatives.  I have also seen cbt which seems to be less actively developed.  And pantsbuild which is non scala specific.

@jacobwinch @paulbrown1982 @pvighi @rupertbates @twrichards  thoughts would be useful!
@nicl I heard from Jacob that you've done some mill experiments too, so it would be good to catch up

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -1,0 +1,151 @@
+import java.io.{FileOutputStream, PrintWriter}
+
+import ammonite.ops._
+import mill._
+import mill.scalalib._
+import mill.util.Ctx
+import upickle.default.{macroRW, ReadWriter => RW}
+import $ivy.`com.amazonaws:aws-java-sdk-s3:1.11.311`
+import com.amazonaws.services.s3.{AmazonS3, AmazonS3Client, AmazonS3ClientBuilder}
+import com.amazonaws.auth.{AWSCredentialsProvider, AWSCredentialsProviderChain, DefaultAWSCredentialsProviderChain}
+import com.amazonaws.regions.Regions
+
+object handlers extends Module {
+
+  import lib._
+
+  object `sf-contact-merge` extends ScalaSettings {
+
+    /*
+    {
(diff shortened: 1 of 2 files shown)


johnduffell:
thanks @nicl !  Actually I just liberally copied and pasted bits of that plugin to prove that it was possible to do - I didn't copy enough to actually work, just enough to prove that I can do what I need (to myself)

Regarding dev experience, the idea generation command was fairly performant but it did confuse intellij a couple of times and I had to restart.

The mill cli you don't really use interactive mode, the startup time is supposed to be quick enough to just run it every time.

I think I wil carry on as a side project and see how it goes.

johnduffell:
this experiment was never continued, closing this, maybe one day it will come back in another form, or something else will!

PR #174, opened 2018-08-07
new grammar for separating http request creation from actually doing the http request

johnduffell:
This PR includes 2 parts (oops)

==

the main part is a new HttpOp which lets us put together things that have side effects (so far, a single HTTP call is supported) without having to use a var in the test.

==

the second part is sharing the Update identity id in salesforce code: after messing up being able to set the identity id in salesforce several times in the sf contact merge endpoint, I was about to write an effects test when I realised we already have the same operation together with tests in identity-backfill.

So rather than trying to copy it I decided to make a lib just for that operation, so we can pull it in exactly where it's needed.

==

@pvighi @jacobwinch @twrichards


lib/salesforce/src/main/scala/com/gu/salesforce/AnyVals.scala
```diff
@@ -0,0 +1,7 @@
+package com.gu.salesforce
+
+object AnyVals {
```
johnduffell: wasn't sure what to call this, I just wanted to put in any things that would be referenced from the below SfEffectsData.  Technically this and the file below are one other module, but we wouldn't use the DEV SF data without the dev SF client so we can get away with keeping them together.

lib/sf/updatesfidentity/src/main/scala/com/gu/identityBackfill/salesforce/UpdateSalesforceIdentityId.scala
```diff
@@ -0,0 +1,27 @@
...
+  case class WireRequest(IdentityID__c: String)
+  implicit val writes = Json.writes[WireRequest]
+
+  case class IdentityId(value: String)
+
+  def apply(patch: PatchRequest => ClientFailableOp[Unit]): (SFContactId, Option[IdentityId]) => ClientFailableOp[Unit] =
+    Function.untupled((toRequest _).tupled.andThen(patch))
+
+  def set(patch: PatchRequest => ClientFailableOp[Unit]): (SFContactId, IdentityId) => ClientFailableOp[Unit] =
+    Function.uncurried(Function.untupled((toRequest _).tupled.andThen(patch)).curried.andThen(_.compose[IdentityId](Some.apply)))
```
johnduffell: 😂 it's like when a magician does a trick and you have no idea what just happened

johnduffell: I just typed stuff until it compiled, and because of the types, there's no way it could be wrong

build.sbt
```diff
@@ -88,6 +88,16 @@ lazy val salesforce = all(project in file("lib/salesforce"))
     libraryDependencies ++= Seq(okhttp3, logging, scalaz, playJson, scalatest)
   )
 
+lazy val updatesfidentity = all(project in file("lib/sf/updatesfidentity"))
```
pvighi: is that supposed to be update sf identity ? because it kinda looks like a typo trying to write updatesIdentity

handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/Handler.scala
```diff
@@ -50,7 +52,10 @@ object Handler {
       val zuoraQuerier = ZuoraQuery(zuoraRequests)
       val getByEmail: EmailAddress => GetByEmail.ApiError \/ IdentityId = GetByEmail(response, identityConfig)
       val countZuoraAccounts: IdentityId => ClientFailableOp[Int] = CountZuoraAccountsForIdentityId(zuoraQuerier)
-      lazy val sfRequests: ApiGatewayOp[Requests] = SalesforceAuthenticate(response, sfConfig)
+
```
johnduffell: SalesforceAuthenticate(...) used to do the doAuth and the map to SalesforceRestRequestMaker.
Because I only want to do the auth part once, I've had to pull it out in order to add the map to the Http patch op below.

handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/salesforce/UpdateSalesforceIdentityId.scala
```diff
@@ -1,16 +0,0 @@
-package com.gu.identityBackfill.salesforce
```
johnduffell: this (and its tests) are now in the new shared lib.  I think it would be rare that sharing would be necessary, but these two lambdas really are doing exactly the same operation from a high level point of view.  From a tech POV this saves on effectstests which will improve their reliability.  Plus I kept shipping a broken operation even though there was a working one already in place.

build.sbt
johnduffell: changing to update-sf-identityid

handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/Handler.scala
```diff
@@ -92,22 +97,22 @@ object Handler {
...
+  def syncableSFToIdentity(sfRequests: ApiGatewayOp[Requests], stage: Stage)(sFContactId: SFContactId): ApiGatewayOp[Unit] =
     for {
       sfRequests <- sfRequests
       standardRecordType <- standardRecordTypeForStage(stage)
       syncable <- SyncableSFToIdentity(standardRecordType)(sfRequests)(sFContactId)
     } yield syncable
 
   def updateSalesforceIdentityId(
-    sfRequests: ApiGatewayOp[Requests]
+    sfRequests: ApiGatewayOp[HttpOp[PatchRequest]]
```
johnduffell: before we took something that could do arbitrary http requests (on SF) now we take a single http operation that we can input a patch request (json body and relative url path)

handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/Handler.scala
```diff
@@ -92,22 +97,22 @@ object Handler {
...
   )(
-    sFContactId: Types.SFContactId,
+    sFContactId: SFContactId,
     identityId: IdentityId
   ): ApiGatewayOp[Unit] =
     for {
       sfRequests <- sfRequests
-      _ <- UpdateSalesforceIdentityId(sfRequests)(sFContactId, identityId).toApiGatewayOp("zuora issue")
+      _ <- UpdateSalesforceIdentityId.set(sfRequests).run2(sFContactId, identityId).toApiGatewayOp("zuora issue")
```
johnduffell: because an HttpOp isn't actually a function, we have to ask it to turn itself into a function using a method I called `run`.  since the IN could be a tuple if there are multi parameters, I added `run2` (similar to Function1, Function2, Function3).  Alternatively we could scrap the run2 and change it from
`blah.run2(sFContactId, identityId)` to
`blah.run((sFContactId, identityId))`

lib/restHttp/src/main/scala/com/gu/util/resthttp/HttpOp.scala
```diff
@@ -0,0 +1,37 @@
+package com.gu.util.resthttp
```
johnduffell: this is the fundamental bit, basically it's just a function that you can prepend to, that can later do an http operation.  you can prepend pure functions to it in order to gradually move it from the original http level to the business logic level.

the run is needed to actually make it do something, but prepend is the only essential method.
In future we might make some way to chain calls together without running them, that would make the tests where we end up using `var`s become able to assert on the result rather than checking the vars (if we remember)

lib/restHttp/src/main/scala/com/gu/util/resthttp/HttpOp.scala
```diff
@@ -0,0 +1,37 @@
...
+case class HttpOp[IN](
+  inputToRequest: IN => Request,
+  effect: Request => Response,
+  responseToOutput: Response => ClientFailableOp[Unit]
+) {
+
+  def run(in: IN): ClientFailableOp[Unit] =
+    responseToOutput(effect(inputToRequest(in)))
+
+  def prepend[NEWIN](function: NEWIN => IN): HttpOp[NEWIN] =
```
jacobwinch: Would _forRequest_ be clearer to a novice? (Or has this novice misunderstood?!)

lib/salesforce/src/main/scala/com/gu/salesforce/AnyVals.scala
jacobwinch: Types?

lib/restHttp/src/main/scala/com/gu/util/resthttp/HttpOp.scala
```diff
@@ -0,0 +1,37 @@
+package com.gu.util.resthttp
+
+import com.gu.util.resthttp.RestRequestMaker.httpIsSuccessful
+import com.gu.util.resthttp.Types.ClientFailableOp
+import okhttp3.{Request, Response}
+
+case class HttpOp[IN](
```
jacobwinch: Is IN a request tied to a specific HTTP verb?

jacobwinch: (I don't feel like IN and NEWIN are helping me to understand this much!)

lib/restHttp/src/main/scala/com/gu/util/resthttp/HttpOp.scala
```diff
@@ -0,0 +1,37 @@
...
+}
+
+object HttpOp {
+
+  def apply(getResponse: Request => Response): HttpOp[Request] =
+    HttpOp[Request](identity, getResponse, httpIsSuccessful)
+
+  // convenience, untuples for you
+  implicit class HttpOpTuple2Ops[A1, A2](httpOpTuple2: HttpOp[(A1, A2)]) {
+    def run2(a1: A1, a2: A2): ClientFailableOp[Unit] = httpOpTuple2.run((a1, a2))
```
jacobwinch: Would _runRequest_ be slightly clearer?

lib/salesforce/src/main/scala/com/gu/salesforce/AnyVals.scala
johnduffell: I thought about that, but I want this to only cover markers for fields, rather than any composite data structures.  types in the type system are merely a way to connect together functions, and the only reason we need a shared concept of SFContactId is because we have some shared test data that we want to input into various tests.  So this is really more specific than AnyVals, it's actually test-data-type-markers .  but I'm not sure if that's a good name.

handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/Handler.scala
johnduffell: the other drawback I didn't mention is that if we don't use run2, when we want to pass in the function to a steps or similar, it's necessary to take is as 
```scala
(updateIdentity: ((SfContactId, IdentityId)) => ClientFailableOp[Unit])
```
raterht than

```scala
(updateIdentity: (SfContactId, IdentityId) => ClientFailableOp[Unit])
```

lib/restHttp/src/main/scala/com/gu/util/resthttp/HttpOp.scala
johnduffell: maybe beforeRequest?  I'm not sure what for means in this context.

lib/restHttp/src/main/scala/com/gu/util/resthttp/HttpOp.scala
```diff
@@ -0,0 +1,37 @@
...
+case class HttpOp[PARAM](
+  inputToRequest: PARAM => Request,
+  effect: Request => Response,
+  responseToOutput: Response => ClientFailableOp[Unit]
+) {
+
+  def runRequest(in: PARAM): ClientFailableOp[Unit] =
+    responseToOutput(effect(inputToRequest(in)))
+
+  def beforeRequest[UPDATEDPARAM](function: UPDATEDPARAM => PARAM): HttpOp[UPDATEDPARAM] =
```
paulbrown1982: I wonder if setupRequest is more semantic. Before makes me think I'm giving it another request to do before this one.

paulbrown1982:
👍 Looks great to me now that I understand things from our conversations. 

One thing that would improve it is a better title to the pull request.

handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/Handler.scala
```diff
@@ -92,22 +98,22 @@ object Handler {
...
+    sfRequests: ApiGatewayOp[HttpOp[PatchRequest]]
   )(
-    sFContactId: Types.SFContactId,
+    sFContactId: SFContactId,
     identityId: IdentityId
   ): ApiGatewayOp[Unit] =
     for {
       sfRequests <- sfRequests
-      _ <- UpdateSalesforceIdentityId(sfRequests)(sFContactId, identityId).toApiGatewayOp("zuora issue")
+      _ <- UpdateSalesforceIdentityId.set(sfRequests).runRequestUntupled(sFContactId, identityId).toApiGatewayOp("zuora issue")
```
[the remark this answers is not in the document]
johnduffell: didn't see your specific suggestion but I already changed to `runRequestMultiArg`.  I think "function" is not the right word because it "is" the function, it's not "with" the function.   We could add `With` if it would make it clearer though.

johnduffell:
thanks, I have re titled the PR to "new grammar for separating http request creation from actually doing the http request"

PR #175, opened 2018-08-07
Date validation from catalog

pvighi:
There's a lot of things in the PR that probably need to change but it works:

- Added server side representation of the catalog and made the existing one a "wire" catalog just for the json response of /product-catalog

- added a server side representation of the start date rules in the catalog that is actually used to validate requests


handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/validation/DateValidationRules.scala
```diff
@@ -0,0 +1,54 @@
+package com.gu.newproduct.api.addsubscription.validation
+import Validation._
+import java.time.{DayOfWeek, LocalDate}
+import java.time.temporal.TemporalAdjusters
+
+case class Days(value: Int) extends AnyVal
+
+trait DateRule {
```
pvighi: maybe this approach is to java-y (if that is even a word) but it works for both validating dates and serialising into the wires model. If we had just a loose date => ValidationResponse function it would be hard to find the original parameters to construct a wires representation

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/validation/PrerequisiteCheck.scala
```diff
@@ -13,12 +11,14 @@ import com.gu.newproduct.api.addsubscription.zuora.{GetAccount, GetAccountSubscr
...
 
+import scala.com.gu.newproduct.api.productcatalog.Catalog
+
 case class ValidatedFields(paymentMethod: PaymentMethod, currency: Currency)
 object PrerequisiteCheck {
   def apply(
     zuoraClient: RestRequestMaker.Requests,
     contributionRatePlanIds: List[ProductRatePlanId],
-    getCurrentDate: () => LocalDate
+    catalog: Catalog,
```
pvighi: everything is hard wired in this file anyway, but I could also just get the validating function here instead of the whole  catalog

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/validation/ValidateRequest.scala
```diff
@@ -1,21 +1,19 @@
...
-import java.time.LocalDate
-
 import com.gu.i18n.Currency
 import com.gu.newproduct.api.addsubscription.AddSubscriptionRequest
 import com.gu.newproduct.api.addsubscription.validation.Validation.BooleanValidation
 
 object ValidateRequest {
   def apply(
-    now: () => LocalDate,
+    maybeDateRule: Option[DateRule],
```
pvighi: I'll change this after lunch, this should probably just get the isValidFunction not the whole dateRule

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/validation/ValidationResult.scala
```diff
@@ -4,6 +4,7 @@ import com.gu.newproduct.api.addsubscription.validation.Conversion._
 import scalaz.{-\/, \/, \/-}
 
 sealed trait ValidationResult[+A] {
+  def isPassed: Boolean
```
pvighi: this isPassed boolean was only needed in an intermediate version of this code but I refactored and forgot to remove it. I'll delete it

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/productcatalog/WireModel.scala
```diff
@@ -0,0 +1,133 @@
...
+    }
+  }
+
+  object WireSelectableWindow {
+    implicit val writes = Json.writes[WireSelectableWindow]
+  }
+
+  object WireStartDateRules {
+    implicit val writes = Json.writes[WireStartDateRules]
+
```
pvighi: The complexity here is caused by the Catalog having date rules as a list of individual rules while the wire catalog just having one object with an optional val for each type of rule.  So we have to start with an empty WireStartDateRules object and keep creating new versions with additional rules.
Note that if there are two of the same type of rule only the last one in the list will be applied as the wires model does not support it.
Maybe we could do move these private defs somewhere else and have them implicitly passed as a param

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/productcatalog/Catalog.scala
```diff
@@ -1,61 +1,44 @@
-package com.gu.newproduct.api.productcatalog
```
pvighi: What was previously called Catalog is now the WireCatalog as it models the json response of the product-catalog endpoint

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/validation/DateValidationRules.scala
```diff
@@ -0,0 +1,54 @@
...
+    val maybeCutOffDate = cutOffDay.map(cutoffDayOfweek => today.`with`(TemporalAdjusters.next(cutoffDayOfweek)))
+    val baseDate = maybeCutOffDate.getOrElse(today)
+    val startOffset = startDelay.map(_.value).getOrElse(0)
+    val startDate = baseDate.plusDays(startOffset)
+    val isBeforeWindowStartDay = d.isBefore(startDate)
+    val maybeWindowEnd = size.map { windowSize => baseDate.plusDays(windowSize.value) }
+
+    def isOnWindowEndDayOrAfter = maybeWindowEnd.map(!d.isBefore(_)).getOrElse(false)
+
+    def errorMessage = s"$d is out of the selectable range: [$startDate - ${maybeWindowEnd.map(_.toString).getOrElse("")})"
```
pvighi: maybe the error messages could be tweaked as it now always displays something like
```2018-08-06 is out of the selectable range: [2018-08-13 -  2018-08-23)```
where '[' means inclusive of the start date and ')' means excluding the end date but probably it won't be clear to most people reading it.
I could subtract one to the end date  and show a range that includes both limits. also I could show special error messages for one day ranges and unlimited ranges (right now they are show with an empty string where the end date should be)

File: handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/Handler.scala [hunk 8]
```diff
@@ -84,12 +85,13 @@ object Steps {
...
-
+      getCurrentDate = () => RawEffects.now().toLocalDate
+      catalog = Catalog(getCurrentDate)
+      isValidStartDate = catalog.monthlyContribution.startDateRule.isValid
       getBillTo = GetBillToContact(zuoraClient.get[GetBillToResponse]) _
       createMonthlyContribution = CreateSubscription(zuoraIds.monthly, zuoraClient.post[WireCreateRequest, WireSubscription]) _
       contributionIds = List(zuoraIds.monthly.productRatePlanId, zuoraIds.annual.productRatePlanId)
-      getCurrentDate = () => RawEffects.now().toLocalDate
-      prerequisiteCheck = PrerequisiteCheck(zuoraClient, contributionIds, getCurrentDate) _
+      prerequisiteCheck = PrerequisiteCheck(zuoraClient, contributionIds, isValidStartDate) _
```
johnduffell: at some point we will have to rename all the existing prereq stuff to include the word `Contributions` to differentuate from Voucher stuff

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/TypeConvert.scala
```diff
@@ -41,4 +43,9 @@ object TypeConvert {
     }
   }
 
+  implicit class OptionalRuleConverter(maybeRule: Option[DateRule]) {
+    def alwaysValid(d: LocalDate) = Passed(())
+    def isValid: LocalDate => ValidationResult[Unit] = maybeRule.map(rule => rule.isValid _).getOrElse(alwaysValid _)
```
johnduffell: could be just `getOrElse(_ => Passed(()))` assuming the type checker is happy with that

johnduffell: or don't even use the d in the or else part
```scala
def isValid(d: LocalDate): ValidationResult[Unit] = maybeRule.map(_.isValid(d)).getOrElse(Passed(()))
```
or pattern match
```scala
def isValid(d: LocalDate): ValidationResult[Unit] = maybeRule match {
  case Some(rule) => rule.isValid(d)
  case None => Passed(())
```

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/validation/DateValidationRules.scala
johnduffell: as discussed perhaps the model is too general - so when we put it into the List we forget the information that there's zero or one of each, and which is which.
so changing to keep both the options (with some handy constructors if needed) will keep things  more consistent and get more compile time safety/refactoring support.

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/validation/DateValidationRules.scala
```diff
@@ -0,0 +1,54 @@
...
+
+trait DateRule {
+  def isValid(d: LocalDate): ValidationResult[Unit]
+}
+
+case class CompositeRule(rules: List[DateRule]) extends DateRule {
+  def isValid(d: LocalDate): ValidationResult[Unit] = {
+    val results = rules.map(_.isValid(d))
+    val failed = results.collect { case f: Failed => f }
+    if (failed.isEmpty) Passed(()) else failed.head
```
johnduffell: to be more type safe this line could be
```scala
failed match {
  case Nil => Passed(())
  case head :: _ => head
}
```

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/validation/DateValidationRules.scala
```diff
@@ -0,0 +1,70 @@
...
+) extends DateRule {
+  override def isValid(d: LocalDate): ValidationResult[Unit] = for {
+    _ <- daysOfWeekRule.isValid(d)
+    _ <- windowRule.isValid(d)
+  } yield (Passed(()))
+}
+
+case class DaysOfWeekRule(allowedDays: List[DayOfWeek]) extends DateRule {
+
+  override def isValid(d: LocalDate): ValidationResult[Unit] = allowedDays contains d.getDayOfWeek orFailWith errorMessage(d)
```
johnduffell: that's pretty cool, it is almost a sentence!
the only way it could be more like a sentence is to make `d` `requestedDate`
allowedDays contains requestedDate.getDayOfWeek orFailWith errorMessage(requestedDate)

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/validation/DateValidationRules.scala
```diff
@@ -0,0 +1,70 @@
...
+(
+  now: () => LocalDate,
+  cutOffDay: Option[DayOfWeek],
+  startDelay: Option[Days],
+  size: Option[Days],
+) extends DateRule {
+  override def isValid(d: LocalDate): ValidationResult[Unit] = {
+    val today = now()
+    val maybeCutOffDate = cutOffDay.map(cutoffDayOfweek => today.`with`(TemporalAdjusters.next(cutoffDayOfweek)))
+    val baseDate = maybeCutOffDate.getOrElse(today)
```
johnduffell:
```scala
cutOffDay match {
  case Some(cutoffDayOfweek) => today.`with`(TemporalAdjusters.next(cutoffDayOfweek)))
  case None => today
}
```

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/validation/DateValidationRules.scala
```diff
@@ -0,0 +1,70 @@
...
+    val allowedDaysStr = allowedDays.mkString(",")
+    val dateDayStr = date.getDayOfWeek.toString
+    s"invalid day of the week: $date is a $dateDayStr. Allowed days are $allowedDaysStr"
+  }
+}
+
+case class WindowRule
+(
+  now: () => LocalDate,
+  cutOffDay: Option[DayOfWeek],
```
johnduffell: should probabl call all these prefixed with `maybe`... to avoid confusion later in the function

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/validation/DateValidationRules.scala
```diff
@@ -0,0 +1,70 @@
...
+  startDelay: Option[Days],
+  size: Option[Days],
+) extends DateRule {
+  override def isValid(d: LocalDate): ValidationResult[Unit] = {
+    val today = now()
+    val maybeCutOffDate = cutOffDay.map(cutoffDayOfweek => today.`with`(TemporalAdjusters.next(cutoffDayOfweek)))
+    val baseDate = maybeCutOffDate.getOrElse(today)
+    val startOffset = startDelay.map(_.value).getOrElse(0)
```
johnduffell: I'd leave the .value until later as this is mixing the low level extraction with the high level default value - do `val startDelay = maybeStartDelay.getOrElse(Days(0))`
then `val startDate = baseDate.plusDays(startDelay.value)`

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/validation/DateValidationRules.scala
```diff
@@ -0,0 +1,70 @@
...
+    val startDate = baseDate.plusDays(startOffset)
+    val isBeforeWindowStartDay = d.isBefore(startDate)
+    val maybeWindowEnd = size.map { windowSize => baseDate.plusDays(windowSize.value) }
```
johnduffell: baseDate should be `startDate` as `size` it the window size rather than the distance from the fulfillent date (I think)

pvighi: basedate is just bad name for something that is next day of the week that matches the cut off day or today if none is defined.
start date is the basedate + delay that is when the window actually starts and the end is startDate + window size. Now I can't remember how we came up with this logic but I think that is how we agreed you could define a min number of days after the next fulfilment

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/validation/DateValidationRules.scala
```diff
@@ -0,0 +1,70 @@
...
+
+    def isOnWindowEndDayOrAfter = maybeWindowEnd.map(!d.isBefore(_)).getOrElse(false)
```
johnduffell: more tersely
```scala
def isOnOrAfterWindowEndDay = maybeWindowEnd.exists(!d.isBefore(_))
```
or even verbosely
```scala
val isWithinWindowEndDateExclusive = maybeWindowEnd exists {
  case Some(windowEnd) => d.isBefore(windowEnd)
  case None => true
}
```

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/validation/DateValidationRules.scala
johnduffell: this is actually a SF or lambda fix needed error , but the CSR could take action to get around it.  So it would probably be a 422 type error, and so would need to be understood by CSRs.  So in the long run we should make it nice.

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/validation/DateValidationRules.scala
```diff
@@ -0,0 +1,70 @@
...
+
+    def errorMessage = s"$d is out of the selectable range: [$startDate - ${maybeWindowEnd.map(_.toString).getOrElse("")})"
+
+    !isBeforeWindowStartDay && !isOnWindowEndDayOrAfter orFailWith (errorMessage)
```
johnduffell: given we had a negative above, I wouldn't want another negative here if avoidable.  how about
```scala
isAfterWindowStartDayInclusive && isBeforeWindowEndDateExclusive orFailWith (errorMessage)
```

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/validation/ValidationResult.scala
johnduffell: we can always add it back if we need it, simpler to delete for now

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/productcatalog/Catalog.scala
johnduffell: the package name starts with `scala` now?

johnduffell: image

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/productcatalog/Catalog.scala
```diff
@@ -1,61 +1,45 @@
...
-
-object Product {
-  implicit val writes = Json.writes[Product]
-}
-
 object Catalog {
-  implicit val writes = Json.writes[Catalog]
+  def apply(getCurrentDate: () => LocalDate): Catalog = {
+    val voucherWindowRule = WindowRule(
+      now = getCurrentDate,
```
johnduffell: I don't like that the rule gets the date, it should really only be passed in the to function when we want to validate a specific date relative to another

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/productcatalog/Catalog.scala
```diff
@@ -1,61 +1,45 @@
...
+      cutOffDay = Some(DayOfWeek.TUESDAY),
+      startDelay = Some(Days(20)),
+      size = Some(Days(28))
+    )
+    val weekendRule = DaysOfWeekRule(List(SATURDAY, SUNDAY))
```
johnduffell: can you choose either Sat or sun for a weekend pack, or only choose Sat? of course you can choose sun for a sunday only

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/productcatalog/WireModel.scala
```diff
@@ -0,0 +1,121 @@
...
+          WirePlanInfo.fromPlan(catalog.voucherEveryDay, "Every day")
+        )
+      )
+      val contributionProduct = WireProduct(
+        label = "Contribution",
+        plans = List(
+          WirePlanInfo.fromPlan(catalog.monthlyContribution, "Monthly")
+        )
+      )
+      WireCatalog(List(voucherProduct, contributionProduct))
```
johnduffell: just a general comment that it's a shame we have to know the exact structure of the cataloghere as well, so if we add something at the top level it won't automatically be picked up as needing to be here.  But unless we make the catalog some kind of tuple or HList it would be hard to make it complete.

lib/effects-sqs/src/test/scala/com/gu/effects/sqs/AWSSQSSendTest.scala
```diff
@@ -45,7 +45,6 @@ object SQSRead {
 
     logger.info(s"reading message to SQS queue $queueUrl")
 
-
     val request = new ReceiveMessageRequest(queueName.value)
```
johnduffell: glad it's not just me getting this change all the time, I keep reverting it.  hopefully once you merge it will all be happy again

johnduffell:
interesting that coveralls compares the PR coverage and the build coverage.  So in the comment it says "Overall coverage increased (+1.3%) to 70.576%" but the status says "Coverage decreased (-0.2%) to 69.867%"

anyway I've finished reviewing for now, I'm happy to take another look later as I think you have a few things you want to look at.  Looks great so far and sorry about the delay in looking.

As an overall comment we have the main Catalog which has mostly validation information, then we have a view onto that where we add descriptive names, which makes a wire catalog.  And then we have another view onto that where we add the current date (and key it by the plan id) which lets us run the validation code.
So we should keep in mind about how that affects the types, and make sure we don't end up passing too much information around in a super-Catalog object when functions don't actually need it all.

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/validation/DateValidationRules.scala
```diff
@@ -0,0 +1,86 @@
...
+      case Some(cutOffDayOfWeek) => now().minusDays(1) `with` TemporalAdjusters.next(cutOffDayOfWeek)
+      case None => now()
+    }
+    val startDelay = maybeStartDelay.getOrElse(Days(0))
+    val startDate = baseDate.plusDays(startDelay.value.toLong)
+    val maybeWindowEnd = maybeSize.map { windowSize => startDate.plusDays(windowSize.value.toLong) }
+    SelectableWindow(startDate, maybeWindowEnd)
+  }
+
+  override def isValid(dateToValidate: LocalDate): ValidationResult[Unit] = {
```
pvighi: now that I look at it after this last refactor this method is not unit tested even though it doesn't do much

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/validation/DateValidationRules.scala
```diff
@@ -42,29 +43,44 @@ case class DaysOfWeekRule(allowedDays: List[DayOfWeek]) extends DateRule {
...
-    val baseDate = maybeCutOffDate.getOrElse(today)
-    val startOffset = startDelay.map(_.value).getOrElse(0)
-    val startDate = baseDate.plusDays(startOffset)
-    val isBeforeWindowStartDay = d.isBefore(startDate)
-    val maybeWindowEnd = size.map { windowSize => baseDate.plusDays(windowSize.value) }
 
-    def isOnWindowEndDayOrAfter = maybeWindowEnd.map(!d.isBefore(_)).getOrElse(false)
+  val selectableWindow = {
+    val baseDate = maybeCutOffDay match {
+      case Some(cutOffDayOfWeek) => now().minusDays(1) `with` TemporalAdjusters.next(cutOffDayOfWeek)
```
johnduffell: are there likely to be any issues with close to the BST/GMT change?  I guess if we stick rigidly to LocalDate rather than including Time we should be OK?

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/validation/DateValidationRules.scala
```diff
@@ -0,0 +1,86 @@
...
+}
+
+case class WindowRule(
+  now: () => LocalDate,
+  maybeCutOffDay: Option[DayOfWeek],
+  maybeStartDelay: Option[Days],
+  maybeSize: Option[Days]
+) extends DateRule {
+
+  val selectableWindow = {
```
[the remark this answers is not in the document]
johnduffell: good point - the SelectableWindow is just the data structure produced from the rules and the current date, to be enforcable against a requested date.
so originally we have the catalog fields and the current date we combine them to get a Selectable window, and then we further combine with the requested date to get the yes/no is it valid.
That should probably match the structure of the code reasonably well.

handlers/new-product-api/src/test/scala/com/gu/newproduct/api/productcatalog/WindowRuleTest.scala
```diff
@@ -0,0 +1,81 @@
...
+
+  it should "define an unbounded window starting today if no params are defined" in {
+    val rule = WindowRule(
+      now = getWedAugust8,
+      maybeCutOffDay = None,
+      maybeStartDelay = None,
+      maybeSize = None
+    )
+
+    rule.selectableWindow shouldBe SelectableWindow(wednesdayAugust8, None)
```
johnduffell: I think you said something about this but I wasn't 100% sure - maybe I will work it out later. as I'd expect this to be just getting back the SelectableWindow directly rather than the rule and having to pick out from inside.

handlers/new-product-api/src/test/scala/com/gu/newproduct/api/productcatalog/WindowRuleTest.scala
```diff
@@ -0,0 +1,81 @@
...
+      maybeCutOffDay = Some(MONDAY),
+      maybeStartDelay = Some(Days(3)),
+      maybeSize = Some(Days(2))
+    )
+    val mondayNextWeek = LocalDate.of(2018, 8, 13)
+    val expectedWindowStart = mondayNextWeek.plusDays(3)
+    val expectedWindowEnd = expectedWindowStart.plusDays(2)
+    rule.selectableWindow shouldBe SelectableWindow(expectedWindowStart, Some(expectedWindowEnd))
+  }
+}
```
johnduffell: these tests are brilliant, it's so clear what's going on 🌟 very good work!

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/validation/DateValidationRules.scala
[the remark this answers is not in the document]
johnduffell: oh right I get it now!
so I think this has just shown we are modelling this a bit wrong.
isValid shouldn't be on the rule.

So if we are saying the rule is the basic data structure, that's ok.  But from the rule we should be able to _produce_ either the Wire model or the validation function. But the rule shouldn't really know about either itself (the wire model producer and the validation function producer would both depend on the rule)

In this case, we have already inadvertently made the validation function (selecablewindow) and the validation function producer (its the code in the braces just after these comments).  But they are tangled up in a way we can't easily remove.

Since the wire model producer is working OK, perhaps we could just structure the validation producer in a similar way.

handlers/new-product-api/src/test/scala/com/gu/newproduct/api/productcatalog/WindowRuleTest.scala
pvighi: What I'm trying to test with this is the code that generates a window from the cut off date, start delay and window size params. And right now that code is in the windowRule. 
I would like to make the window rule just take a SelectableWindow, and move the code that generates the window somewhere else (maybe a companion object of the SelectableWindow). But if we do that we would have to think about how we would generate the wires version of the rule to put in the product-catalog json response since the WindowRule would not know about the original parameters anymore.

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/validation/DateValidationRules.scala
pvighi: I hate time zone changes.
 I don't think it is very likely that it's going to be a problem here considering the CSRs won't be working when time zone changes happen.
But maybe we should investigate how LocalDate and LocalDateTime work. 
According to the [doc](https://docs.oracle.com/javase/8/docs/api/java/time/LocalDateTime.html), when you call now() they return the date in the Default timezone. If the default timezone is GMT (which sounds reasonable) we don't have a problem other than making sure any external client knows we always use GMT.

johnduffell:
👍 👍 👍 🚢

PR #177, opened 2018-08-07
capture the gaData in the SF cancellation case so we can see if the A/B test variants affected the users actions

twrichards:
minor change to forward on some extra GA data into the salesforce case for better reporting down the line

### DIFF
--- a/handlers/cancellation-sf-cases/src/main/scala/com/gu/cancellation/sf_cases/Handler.scala
+++ b/handlers/cancellation-sf-cases/src/main/scala/com/gu/cancellation/sf_cases/Handler.scala
@@ -68,7 +68,8 @@ object Handler extends Logging {
     case class RaiseCaseDetail(
       product: ProductName,
       reason: Reason,
-      subscriptionName: SubscriptionName
+      subscriptionName: SubscriptionName,
+      gaData: String
     )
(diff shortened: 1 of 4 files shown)


handlers/cancellation-sf-cases/src/main/scala/com/gu/cancellation/sf_cases/Handler.scala
```diff
@@ -68,7 +68,8 @@ object Handler extends Logging {
     case class RaiseCaseDetail(
       product: ProductName,
       reason: Reason,
-      subscriptionName: SubscriptionName
+      subscriptionName: SubscriptionName,
+      gaData: String
```
johnduffell: can this be a wrapper class rather than a string?
preumsblay the GA data is which bucket(s) they are in out of the running tests (but perhaps more generic than that)

It would be nice to have a unit test, then we could see what sort of thing would be included.

johnduffell:
regarding data protected, could there be enough detail in the GAData to make the original data we are putting in GA become personally identifying?  I'm guessing this would be more like bucket rather than specific browser id.

other than the above comments no problem 👍

johnduffell:
looks good, adding the unit test is useful 💯  👍

PR #178, opened 2018-08-09
add new planId field to be supplied in incoming request

johnduffell:
This PR adds a new field to the expected request called planId.
At present only monthly_contribution is supported as a value.
If anything else is specified, it will return an error explaining that.
If it is missing, this will be flagged as an error.

https://trello.com/c/fd8OIPgh/328-sl-accepts-plan-choice

@pvighi @david-pepper please take a look
David I can deploy to CODE if you like and if patricio isn't using it - this will make planId required, so it will break any client code that doesn't send that in the request.


handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/Handler.scala
```diff
@@ -42,8 +42,8 @@ object Steps {
 
   def addSubscriptionSteps(
     prerequisiteCheck: AddSubscriptionRequest => AsyncApiGatewayOp[ValidatedFields],
-    createMonthlyContribution: CreateReq => ClientFailableOp[SubscriptionName],
-    sendConfirmationEmail: ContributionsEmailData => AsyncApiGatewayOp[Unit],
+    createMonthlyContribution: ZuoraCreateSubRequest => ClientFailableOp[SubscriptionName],
```
johnduffell: I wanted to give this a better name because it was not clear between the incoming SL request and the outgoing request to the client that will hit zuora.  Not sure if this is OK.

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/validation/ValidatePlan.scala
```diff
@@ -0,0 +1,12 @@
+package com.gu.newproduct.api.addsubscription.validation
+
+import com.gu.newproduct.api.productcatalog.PlanId
+import com.gu.newproduct.api.addsubscription.validation.Validation._
+
+object ValidatePlan {
+  def apply(planId: PlanId): ValidationResult[Unit] = {
+    val requestingMonthlyContribution = planId.value == "monthly_contribution" // FIXME use the plan name from the catalog once available
```
johnduffell: this needs to be tied back into the catalog in the end, when the proper validation is implemented.

In the end it would need to find the right validation to do somehow, perhaps we just run all the validations in turn and if any pass it's ok, or more likely we choose which validation based on the plan id then make it deterministic. But either way this won't exist in the current form.

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/validation/ValidateRequest.scala
```diff
@@ -3,22 +3,28 @@ package com.gu.newproduct.api.addsubscription.validation
...
+    amountMinorUnits: AmountMinorUnits,
+    startDate: LocalDate
+  )
+
   def apply(
     now: () => LocalDate,
     limitsFor: Currency => AmountLimits
   )(
-    addSubscriptionRequest: AddSubscriptionRequest,
+    validatableFields: ValidatableFields,
```
johnduffell: since we only needed just 2 fields from the request, it was appropriate to take a minimal object here.  That makes the tests much simpler and clearer as we don't need to input anything that's unused

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/productcatalog/PlanId.scala
```diff
@@ -0,0 +1,4 @@
+package com.gu.newproduct.api.productcatalog
+
+// temporary until dateValidation branch is merged in
+case class PlanId(value: String) extends AnyVal
```
johnduffell: this is mirroring the PlanId with the same package but isn't in master yet, so this will be removed once that is merged or vice versa
https://github.com/guardian/zuora-auto-cancel/pull/175/files#diff-7f797707abfd3cb80e5d549dedd7d0e1R43

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/validation/ValidatePlan.scala
pvighi: Yeah I expect the next version of this to get get the correct planInformation for the provided Id and from there to customise the validation , email and subscription accordingly. If no matching plan is found then a validation error would be returned

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/validation/ValidateRequest.scala
pvighi: I'm not 100% sure we need a new object for 2 fields  but I'm against it either.

PR #179, opened 2018-08-09
separate rules model from validation

johnduffell:
This is just @pvighi  doing an experiment/prototype but I couldn't comment without making a PR.  This would merge into the main branch not into Master.


handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/validation/DateValidationRules.scala
```diff
@@ -60,27 +49,76 @@ case class WindowRule(
...
+    def orPass: LocalDate => ValidationResult[Unit] = maybeRule.getOrElse { anyDate: LocalDate => Passed(()) }
   }
+
 }
 
-case class SelectableWindow(start: LocalDate, maybeEndExclusive: Option[LocalDate]) {
-  def contains(date: LocalDate) = {
-    val isOnOrAfterWindowStart = date.isAfter(start) || date.isEqual(start)
+object DateValidator {
+  def validatorFor(now: () => LocalDate, dateRule: DateRule): LocalDate => ValidationResult[Unit] = dateRule match {
```
johnduffell: pattern match here makes more sense because DateRule is now a sealed trait - the compiler will warn us if we miss something.

just for background:
there are 3 main ways of doing polymorphism in scala -
1. is inheritance (which creates a dep from the classes to the polymorphic code, but separates the implementation for each class from each other)
2. is pattern matching (which reverses the dep - it's now from the blob of polymorphic code to the type, but the implementations are grouped together into one place)
3. is type classes (which creates a dep from each classes implemention to both the class and the definition, so separates the implementation for each class)

3 produces the most independent code, but it uses implicits, with all their confusing aspects!

Overall I think the way you have proposed is the best solution for now, but we can always revisit.

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/validation/DateValidationRules.scala
```diff
@@ -60,27 +49,76 @@ case class WindowRule(
...
+  ): ValidationResult[Unit] = {
+
+    def errorMessage = {
+      val windowEndStr = selectableWindow.maybeEndExclusive.map(_.toString).getOrElse("")
+      s"$dateToValidate is out of the selectable range: [${selectableWindow.start} - $windowEndStr)"
+    }
+
+    selectableWindow contains dateToValidate orFailWith errorMessage
+  }
+}
```
johnduffell: I think this way of splitting up is great, I feel like all the parts have a cohesive responsibliity now and should be easier to test.
I agree that moving stuff around will help a lot but I think in terms of dependencies between parts it's much better 🥇

johnduffell: just as a side thought now the dependencies are reduced, in terms of organising into packages, we should think whether to group by function (window, days of week) or type (validator, rule, wire catalog) or a combination.  I think it would depend on how the remaining connections go, and whether we are likely to make changes by function or by type.  The compiler will check it all anyway, so it is just for our convenience how we organise it.

PR #180, opened 2018-08-10
email pr 1: Contributions email refactor to make space for the voucher one

johnduffell:
This PR just moves all the contribution specific logic to one package.  Then we can put the voucher stuff in another one.
I also added a manual test, so we can run the email to ourselves.  It's working already but you have to pass your email address in as a parameter.
@pvighi @david-pepper


handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/Handler.scala
```diff
@@ -35,7 +36,7 @@ object Handler extends Logging {
   // Referenced in Cloudformation
   def apply(inputStream: InputStream, outputStream: OutputStream, context: Context): Unit =
     ApiGatewayHandler(LambdaIO(inputStream, outputStream, context)) {
-      Steps.operationForEffects(RawEffects.response, RawEffects.stage, GetFromS3.fetchString)
+      Steps.operationForEffects(RawEffects.response, RawEffects.stage, GetFromS3.fetchString, AwsSQSSend.apply)
```
johnduffell: I noticed we were picking up the effect elsewhere - although it seems we are not testing operationForEffects, we would still want to know it was isolated from AWS if we did.

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/email/ETPayload.scala
```diff
@@ -20,9 +20,12 @@ object ETPayload {
 
   implicit def writes[A: Writes] = Json.writes[ETPayload[A]]
 
-  def apply[A](email: String, fields: A): ETPayload[A] =
+  def apply[A](email: String, fields: A)(implicit name: DataExtensionName[A]): ETPayload[A] =
```
pvighi: why is the data extension name implicit here ?

johnduffell: that is the typeclass pattern, and it means that based on the type of the data extension it will know the id.  There should only be one per type.
we could just pass it in explicitly which would be clearer

johnduffell: I have updated to pass in and it seems better that way so I've pushed it

PR #182, opened 2018-08-13
addSubscription refactor

pvighi:
Some refactor in preparation for voucher validation being added.
The idea of this PR was to try to simplify the steps code in order to be able to add voucher validation.

 In an attempt to reduce the dependencies that need to be injected we compose zuora getters and validators in order to get a single function :  ZuoraId => ValidatedObject

All Zuora calls are grouped in a single getCustomer data call to simplify the wiring code. 
_ the planId parameter in the request is now parsed into a sealed trait and will return the same type as json format errors


handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/validation/Validation.scala
```diff
@@ -14,5 +18,17 @@ object Validation {
...
   }
 
+  implicit class ComposeValidation[ID, DATA](getter: ID => ClientFailableOp[DATA]) {
+    def andValidateWith[VALIDATED](validate: DATA => ValidationResult[VALIDATED], ifNotFoundReturn: Option[String] = None): ID => ApiGatewayOp[VALIDATED] =
+      (id: ID) =>
+        for {
+          data <- ifNotFoundReturn match {
+            case Some(notFoundError) => getter(id).toApiResponseCheckingNotFound("getting data", notFoundError)
+            case None => getter(id).toApiGatewayOp("getting data")
+          }
```
pvighi: IfNotFoundReturn is used to return a validation style error instead of a status 500 if  a zuora call returns 404. In this code we need to do that only on the initial call to inform the client the account id is invalid.
This could probably be further refactored to accept a ClientFailableOp to ApigatewayOp converter instead of this hard coded behaviour. But it is good enough for now

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/Handler.scala
```diff
@@ -40,39 +48,53 @@ object Handler extends Logging {
...
-      zuoraCreateSubRequest = ZuoraCreateSubRequest(
-        request.zuoraAccountId,
-        request.amountMinorUnits,
-        request.startDate,
-        acceptanceDate,
-        request.acquisitionCase,
-        request.acquisitionSource,
-        request.createdByCSR
-      )
+      request <- apiGatewayRequest.bodyAsCaseClass[AddSubscriptionRequest]().withLogging("parsed request with steps1!").toAsync
```
jacobwinch: minor: there's a typo in this logging

PR #183, opened 2018-08-14
Hard coded catalog values

pvighi:
Added all voucher plans and payment descriptions to the catalog response. The payment information is hardcoded for now to allow @david-pepper to make progress on the salesforce side. 
In the future we should probably use the catalog service (or zuora directly) to get at least the updated prices

### DIFF
--- a/handlers/new-product-api/src/main/scala/com/gu/newproduct/api/productcatalog/Catalog.scala
+++ b/handlers/new-product-api/src/main/scala/com/gu/newproduct/api/productcatalog/Catalog.scala
@@ -7,8 +7,17 @@ import com.gu.newproduct.api.productcatalog.PlanId._
 
 case class Catalog(
   voucherWeekend: Plan,
+  voucherSaturday: Plan,
+  voucherSunday:Plan,
   voucherEveryDay: Plan,
-  monthlyContribution: Plan
+  voucherSixDay: Plan,
+  voucherWeekendPlus: Plan,
+  voucherSaturdayPlus: Plan,
+  voucherSundayPlus:Plan,
+  voucherEveryDayPlus: Plan,
+  voucherSixDayPlus: Plan,
+  monthlyContribution: Plan,
+
 )
 
 object NewProductApi {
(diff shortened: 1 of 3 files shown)


handlers/new-product-api/src/main/scala/com/gu/newproduct/api/productcatalog/Catalog.scala
```diff
@@ -3,53 +3,90 @@ package com.gu.newproduct.api.productcatalog
...
+      voucherWeekend = Plan(VoucherWeekend, voucherSaturdayDateRules, monthlyPayment("20.76")),
+      voucherSixDay = Plan(VoucherSixDay, voucherMondayRules, monthlyPayment("41.12")),
+      voucherSixDayPlus = Plan(VoucherSixDayPlus, voucherMondayRules, monthlyPayment("47.62")),
+      voucherEveryDay = Plan(VoucherEveryDay, voucherMondayRules, monthlyPayment("47.62")),
+      voucherEveryDayPlus = Plan(VoucherEveryDayPlus, voucherMondayRules, monthlyPayment("51.96")),
+      voucherSaturday = Plan(VoucherSaturday, voucherSaturdayDateRules, monthlyPayment("10.36")),
+      voucherSaturdayPlus = Plan(VoucherSaturdayPlus, voucherSaturdayDateRules, monthlyPayment("21.62")),
+      voucherSunday = Plan(VoucherSunday, voucherSundayDateRules, monthlyPayment("10.79")),
+      voucherSundayPlus = Plan(VoucherSundayPlus, voucherSundayDateRules, monthlyPayment("22.06")),
+      monthlyContribution = Plan(MonthlyContribution, monthlyContributionRules)
```
pvighi: In the api response these plans are grouped by product, maybe as we add more functionality it would make sense to group them in the internal representation as well

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/productcatalog/Catalog.scala
```diff
@@ -3,53 +3,90 @@ package com.gu.newproduct.api.productcatalog
...
+
   val supported = List(MonthlyContribution)
+
   def fromName(name: String): Option[PlanId] = supported.find(_.name == name)
 }
 
-case class Plan(id: PlanId, startDateRules: StartDateRules = StartDateRules())
+case class Plan(id: PlanId, startDateRules: StartDateRules = StartDateRules(), paymentPlan: Option[PaymentPlan] = None)
+
+case class PaymentPlan(description: String)
```
pvighi: since the payment plan is not used for anything in this api at least for now, the most flexible thing to do is to just have a string describing it. 
One thing we are not tackling yet is multi currency products, as we currently only have vouchers which are gbp only and contributions that have no set price. In the future we might have to accept a currency query param on the /product-catalog endpoint or just return multiple values for all defined currencies

PR #184, opened 2018-08-15
voucher validation

pvighi:
if planId in the request is one of the voucher ids, it will perform the appropriate validation. If the validation passes a fake Id will be returned and no email will be sent.

We will probably have to refactor the handler at some point since the wiring code is getting more complicated and it is not tested at the moment.


handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/validation/Validation.scala
```diff
@@ -30,5 +30,14 @@ object Validation {
...
 
+  implicit class ComposeValidation[UNVALIDATED, VALIDATED](validate: UNVALIDATED => ValidationResult[VALIDATED]) {
+    def thenValidate[TWICEVALIDATED](validateAgain: VALIDATED => ValidationResult[TWICEVALIDATED]): UNVALIDATED => ValidationResult[TWICEVALIDATED] =
+      (unvalidated: UNVALIDATED) =>
+        for {
+          validated <- validate(unvalidated)
+          twiceValidated <- validateAgain(validated)
+        } yield twiceValidated
+  }
+
```
pvighi: Since there is a lot of overlap between the validation for different products I added this was of composing the output of one validation into the input of another. I use this to  define a generic validation that applies to all plans and then compose a custom validation for the specific plan.
An example of this being used is  [this](https://github.com/guardian/zuora-auto-cancel/pull/184/files#diff-a4dcedfa6bd7f3240003f68fea204cddR174)

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/validation/contribution/ContributionValidations.scala
```diff
@@ -0,0 +1,29 @@
...
+    isValidStartDate: LocalDate => ValidationResult[Unit],
+    limitsFor: Currency => AmountLimits
+  )(
+    validatableFields: ValidatableFields,
+    currency: Currency
+  ): ValidationResult[AmountMinorUnits] =
+    for {
+      amount <- validatableFields.amountMinorUnits getOrFailWith s"amount is missing"
+      _ <- isValidStartDate(validatableFields.startDate)
+      limits = limitsFor(currency)
```
pvighi: maybe we could split this in two different objects since right now this depends on the currency in order to be able to get the per currency amount limits.  Since the currency is in the zuora account we can only perform this validation after loading the data from zuora. 
We could validate the start date and the amount not being None before going to zuora. Also we could define fixed limits instead of having per currency limits like the web and this code would be much simpler

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/validation/Validation.scala
```diff
@@ -30,5 +30,14 @@ object Validation {
         } yield validatedData
   }
 
+  implicit class ComposeValidation[UNVALIDATED, VALIDATED](validate: UNVALIDATED => ValidationResult[VALIDATED]) {
+    def thenValidate[TWICEVALIDATED](validateAgain: VALIDATED => ValidationResult[TWICEVALIDATED]): UNVALIDATED => ValidationResult[TWICEVALIDATED] =
+      (unvalidated: UNVALIDATED) =>
+        for {
+          validated <- validate(unvalidated)
```
jacobwinch: Perhaps _initialValidation_ and _finalValidation_ would be more intuitive names here? The usage looks neat though 👍

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/validation/contribution/ContributionAccountValidation.scala
```diff
@@ -0,0 +1,14 @@
...
+
+import com.gu.newproduct.api.addsubscription.validation.{Failed, Passed, ValidatedAccount, ValidationResult}
+
+object ContributionAccountValidation {
+  def apply(
+    account: ValidatedAccount
+  ): ValidationResult[ValidatedAccount] = {
+    account.identityId match {
+      case Some(_) => Passed(account)
+      case None => Failed("Zuora account has no Identity Id")
```
jacobwinch: A thought for a future PR - we could consider creating guest identity accounts instead of failing the validation?

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/validation/voucher/ValidateContactsForVoucher.scala
```diff
@@ -0,0 +1,14 @@
...
+import com.gu.i18n.Country
+import com.gu.newproduct.api.addsubscription.validation.{Failed, Passed, ValidationResult}
+import com.gu.newproduct.api.addsubscription.zuora.GetContacts.Contacts
+
+object ValidateContactsForVoucher {
+  def apply(contacts: Contacts): ValidationResult[Contacts] = {
+    if (contacts.soldTo.country == Country.UK)
+      Passed(contacts)
+    else
+      Failed(s"Invalid country: ${contacts.soldTo.country.name}, only UK addresses are allowed")
```
jacobwinch: Nice clear message 👍

handlers/new-product-api/src/test/scala/com/gu/newproduct/api/addsubscription/StepsTest.scala
```diff
@@ -93,11 +115,23 @@ class StepsTest extends FlatSpec with Matchers {
...
       fakeSendEmails
+    ) _
+
+    def fakeValidateVoucherStart(id: PlanId, d: LocalDate) = Passed(())
+
+    val fakeAddVoucherSteps = Steps.addVoucherSteps(
+      fakeGetVoucherCustomerData,
+      fakeValidateVoucherStart
+    ) _
+    //TODO MAKE INDEPENDENT TESTS FOR CONTRIBUTIONS AND VOUCHERS
```
jacobwinch: 👍

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/Handler.scala
```diff
@@ -48,9 +51,9 @@ object Handler extends Logging {
 }
 
 object Steps {
-  def createZuoraSubRequest(request: AddSubscriptionRequest, acceptanceDate: LocalDate) = ZuoraCreateSubRequest(
+  def createZuoraSubRequest(request: AddSubscriptionRequest, acceptanceDate: LocalDate, amountMinorUnits: AmountMinorUnits) = ZuoraCreateSubRequest(
```
jacobwinch: I find amountMinorUnits a slightly confusing name - does this come from Zuora?

handlers/new-product-api/src/test/scala/com/gu/newproduct/api/addsubscription/validation/voucher/VoucherAccountValidationTest.scala
```diff
@@ -0,0 +1,24 @@
...
+    paymentMethodId = PaymentMethodId("id"),
+    autoPay = AutoPay(true),
+    accountBalanceMinorUnits = AccountBalanceMinorUnits(1233),
+    currency = GBP
+  )
+  it should "pass if currency in account is GBP " in {
+    VoucherAccountValidation(account) shouldBe Passed(account)
+  }
+
+  it should "pass if currency in account is not GBP " in {
```
jacobwinch: it should fail

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/Handler.scala
pvighi: It was suggested by David. Apparently it's the official currency agnostic term to mean cents / pence.  I was leaning towards priceInCents even though it is not technically accurate but in the end all I want is to make sure we know numbers here are multiplied by 100.
So if we come up with a better name we can change it

handlers/new-product-api/src/test/scala/com/gu/newproduct/api/addsubscription/validation/voucher/VoucherAccountValidationTest.scala
pvighi: 🤦‍♂️

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/Handler.scala
jacobwinch: Ah okay, I googled it and it makes more sense now... perhaps amountInMinorUnits would be slightly clearer if we are going to stick with that. Out of (more general) interest, why did we choose to work with  minor units?

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/validation/contribution/ContributionAccountValidation.scala
pvighi: we can talk about it. In the initial meeting for contribution that was decided as a way to prevent adding contributors without identity i don't remember if we decided against guest accounts or if it just a matter of making the smallest implementation that is useful.

Another thing to keep in mind is that this is used by a CSR that is most likely in contact with the customer. So we can show useful error messages and the CSR can fix things that even require user interaction. 
In this case I don't know if they have a quick and easy way to get the user to create an identity account though

jacobwinch: > In this case I don't know if they have a quick and easy way to get the user to create an identity account though

That was my main thought in this case.

> ...making the smallest implementation that is useful

👍for this approach.

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/Handler.scala
pvighi: I think it was to avoid dealing with rounding errors or anything like that, although I'm not sure it would be a problem here.
Regarding changing the name or whether we use cents or decimals I wouldn't change it in this PR because it would require a change on salesforce's side .

PR #185, opened 2018-08-17
create voucher subscriptions from new product API

pvighi:
Added the code to actually create voucher subscriptions from the /add-subscription endpoint.
There's probably some refactoring to do to extract the contribution and voucher specific code from the handler into their own files but I'm trying not to refactor too much in the same PR I add new functionality

### DIFF
--- a/handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/ZuoraIds.scala
+++ b/handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/ZuoraIds.scala
@@ -1,5 +1,7 @@
 package com.gu.newproduct.api.addsubscription
 
+import com.gu.newproduct.api.productcatalog.PlanId
+import com.gu.newproduct.api.productcatalog.PlanId._
 import com.gu.util.apigateway.ApiGatewayResponse
 import com.gu.util.config.Stage
 import com.gu.util.reader.Types.{ApiGatewayOp, _}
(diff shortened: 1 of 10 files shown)


handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/ZuoraIds.scala
```diff
@@ -1,47 +1,124 @@
 package com.gu.newproduct.api.addsubscription
 
+import com.gu.newproduct.api.productcatalog.PlanId
+import com.gu.newproduct.api.productcatalog.PlanId._
 import com.gu.util.apigateway.ApiGatewayResponse
 import com.gu.util.config.Stage
 import com.gu.util.reader.Types.{ApiGatewayOp, _}
 
 object ZuoraIds {
```
pvighi: when we start loading updated prices from the Zuora catalog we could also try to get these Ids instead of having them hard coded here, which is pretty hard to have a test to detect having the wrong ids in prod

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/ZuoraIds.scala
```diff
@@ -1,47 +1,124 @@
...
+          everyday = ProductRatePlanId("2c92c0f9555cf10501556e84a70440e2"),
+          sunday = ProductRatePlanId("2c92c0f95aff3b56015b1045fb9332d2"),
+          saturday = ProductRatePlanId("2c92c0f861f9c26d0161fc434bfe004c"),
+          weekend = ProductRatePlanId("2c92c0f8555ce5cf01556e7f01b81b94"),
+          sixDay = ProductRatePlanId("2c92c0f8555ce5cf01556e7f01771b8a"),
+          everydayPlus = ProductRatePlanId("2c92c0f95aff3b53015b10469bbf5f5f"),
+          sundayPlus = ProductRatePlanId("2c92c0f955a0b5bf0155b62623846fc8"),
+          saturdayPlus = ProductRatePlanId("2c92c0f961f9cf300161fc44f2661258"), //for some reason this is effective in the future in the dev catalog
+          weekendPlus = ProductRatePlanId("2c92c0f95aff3b54015b1047efaa2ac3"),
+          sixDayPlus = ProductRatePlanId("2c92c0f855c3b8190155c585a95e6f5a") //TODO for some reason at least in dev this sets all the charges for the days to quantity = 0 !
```
pvighi: I noticed by testing in dev that these two rateplans don't work there, I'll either fix them in zuora dev and remove the comments or remove the plans if they are not supposed to work before merging this

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/zuora/CreateSubscription.scala
```diff
@@ -56,21 +62,25 @@ object CreateSubscription {
...
 
+  case class ChargeOverride(
+    amountMinorUnits: AmountMinorUnits,
+    productRatePlanChargeId: ProductRatePlanChargeId
+  )
   case class ZuoraCreateSubRequest(
+    productRatePlanId: ProductRatePlanId,
     accountId: ZuoraAccountId,
-    amountMinorUnits: AmountMinorUnits,
+    maybeChargeOverride: Option[ChargeOverride],
```
johnduffell: regarding what we were discussing earlier about having this as a None for voucher all the time - that just means this is a zuora concept level model i.e. above json structure "wire model" level but below our product specific model.
So if we have logic that wants to work at the level of our products, we should have a suitable model for out products (which might be already in existence) and have a way of converting that into the slightly lower level zuora model.

johnduffell:
looks good! I notice there's some conflicts probably mostly just due to my package renaming in the PR I just merged, and also new code in the handler that probably should be covered in some way
https://coveralls.io/builds/18515141/source?filename=handlers%2Fnew-product-api%2Fsrc%2Fmain%2Fscala%2Fcom%2Fgu%2Fnewproduct%2Fapi%2Faddsubscription%2FHandler.scala
Other than that it's good to go 👍

PR #186, opened 2018-08-21
Get product prices from zuora catalog

pvighi:
Read the zuora catalog from the catalog-service file in s3 and fetch the prices to return in the new product api /product-catalog response.

This PR only loads the prices and relies on hard coded zuora Ids to find them.
 
I'm leaving loading Zuora ids from the catalog for later as it is of lower priority. For that we would need to add frontend_id to the voucher plans but we should figure something out to recognise the contribution charge for the amount override (unless we are happy assuming there will only be a single charge)


handlers/new-product-api/src/main/scala/com/gu/newproduct/api/productcatalog/Catalog.scala
```diff
@@ -59,16 +58,16 @@ object NewProductApi {
     val monthlyContributionRules = StartDateRules(windowRule = Some(monthlyContributionWindow))
 
     Catalog(
```
pvighi: I'm starting to think that maybe we don't need an internal representation of the catalog. We could keep the different parts separated. 
Leaving the price in service layer catalog representation meant that every part of the code that needed something from the catalog had to deal with loading the prices.

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/productcatalog/WireModel.scala
```diff
@@ -80,13 +82,13 @@ object WireModel {
     def toOptionalWireRules(startDateRules: StartDateRules): Option[WireStartDateRules] =
       if (startDateRules == StartDateRules()) None else Some(WireStartDateRules.fromStartDateRules(startDateRules))
 
-    def fromPlan(plan: Plan, label: String) = {
-
+    def fromPlan(plan: Plan, label: String, priceFor: PlanId => Option[AmountMinorUnits]) = {
+      val paymentDescription = priceFor(plan.id).map(amount => s"GBP ${amount.formatted} every month")
```
pvighi: All the plans we support right now are monthly GBP payments so the description is kinda hard coded

handlers/new-product-api/src/test/scala/com/gu/newproduct/api/productcatalog/CatalogWireTest.scala
```diff
@@ -39,7 +42,7 @@ class CatalogWireTest extends FlatSpec with Matchers {
         |              "sizeInDays": 28
         |            }
         |          },
-        |          "paymentPlan": "£47.62 every month"
+        |          "paymentPlan": "GBP 47.62 every month"
```
pvighi: david asked for the price format to be changed from £xx.xx to GBP xx.xx

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/productcatalog/PricesFromZuoraCatalog.scala
```diff
@@ -0,0 +1,109 @@
...
+      planId flatMap { planId =>
+        def toMinorUnits(amount: Double) = AmountMinorUnits((amount * 100).toInt)
+
+        val allPrices = for {
+          charge <- productRatePlanCharges
+          price <- charge.pricing
+        } yield price
+
+        val gbpAmounts = allPrices.collect { case Price(Some(amount), "GBP") => amount }
+        val totalPrice = if (gbpAmounts.isEmpty) None else Some(toMinorUnits(gbpAmounts.sum))
```
pvighi: I assumed that the price for a voucher plan was the sum of the prices of each of the charges for everyday and the numbers seem to add up. Is there an easier way to get this ?

handlers/new-product-api/src/test/scala/com/gu/newproduct/api/PricesFromZuoraCatalogTest.scala
```diff
@@ -0,0 +1,60 @@
...
+
+import scala.io.Source
+import scala.util.Try
+
+class PricesFromZuoraCatalogTest extends FlatSpec with Matchers {
+
+  val fakeGetStringFromS3: StringFromS3 = s3Location => {
+    s3Location shouldBe S3Location(bucket = "gu-zuora-catalog", key = "PROD/Zuora-DEV/catalog.json")
+    Try {
+      val source = Source.fromURL(getClass.getResource("/TestZuoraCatalog.json"))
```
pvighi: I'm using part of the dev catalog as test data which is being loaded from a json file. It was too big to hard code it here

handlers/new-product-api/src/test/scala/com/gu/newproduct/api/PricesFromZuoraCatalogTest.scala
```diff
@@ -0,0 +1,60 @@
...
+      source.mkString
+    }
+  }
+
+  it should "load catalog" in {
+
+    val rateplanToPlanId = Map(
+      ProductRatePlanId("VoucherEverydayId") -> VoucherEveryDay,
```
pvighi: I edited the dev catalog ids from the test data to make the test more clear, the downside of this is that it's not so easy to update the test catalog if we need to

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/TypeConvert.scala
```diff
@@ -48,4 +50,15 @@ object TypeConvert {
     }
   }
 
+  implicit class TryToClientFailableOp[A](tryValue: Try[A]) extends Logging {
+    def toClientFailable(action: String) = tryValue match {
+      case Success(response) => ClientSuccess(response)
+      case Failure(exception) => {
+        val errorMessage = s"exception thrown while trying to $action : ${exception.getMessage}"
```
johnduffell: use toString instead of getMessage so that you get the exception class name as well as the text

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/TypeConvert.scala
```diff
@@ -48,4 +50,15 @@ object TypeConvert {
...
+        logger.error(errorMessage, exception)
```
johnduffell: if we log at "error" level, we will get a log line even if it's an expected error and should fail later.  Not sure if that's OK, when it happens in the tests it looks like something failed.
I'd prefer to pass as much information up the chain in the GenericError instead if possible.

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/productcatalog/Handler.scala
```diff
@@ -3,24 +3,35 @@ package com.gu.newproduct.api.productcatalog
...
+  }
+
+  def runWithEffects(lambdaIO: LambdaIO, stage: Stage, fetchString: StringFromS3): ApiGatewayOp[Operation] = for {
+    zuoraIds <- ZuoraIds.zuoraIdsForStage(stage)
+    zuoraToPlanId = zuoraIds.voucherZuoraIds.zuoraIdToPlanid.get _
+    zuoraEnv = ZuoraEnvironment.EnvForStage(stage)
+    plansWithPrice <- PricesFromZuoraCatalog(zuoraEnv, fetchString, zuoraToPlanId).toApiGatewayOp("get prices from zuora catalog")
+    wireCatalog = WireCatalog.fromCatalog(NewProductApi.catalog, plansWithPrice.get)
+  } yield {
+    Operation.noHealthcheck {
```
johnduffell: nohealthcheck is a bit of a lie, when it's actually getting the wireCatalog before producting the operation 😆

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/productcatalog/PricesFromZuoraCatalog.scala
```diff
@@ -0,0 +1,109 @@
...
+import com.gu.util.config.LoadConfigModule.{S3Location, StringFromS3}
+import com.gu.util.config.ZuoraEnvironment
+import com.gu.util.resthttp.Types.ClientFailableOp
+import play.api.libs.json.Json
+import ZuoraCatalogWireModel._
+import com.gu.newproduct.api.addsubscription.AmountMinorUnits
+
+import scala.util.Try
+
+case class PlanWithPrice(planId: PlanId, maybepriceMinorUnits: AmountMinorUnits)
```
johnduffell: is it still maybe? `maybepriceMinorUnits`

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/productcatalog/PricesFromZuoraCatalog.scala
johnduffell: I seem to remember that's exactly how it works

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/productcatalog/PricesFromZuoraCatalog.scala
```diff
@@ -0,0 +1,109 @@
...
+
+      planId flatMap { planId =>
+        def toMinorUnits(amount: Double) = AmountMinorUnits((amount * 100).toInt)
+
+        val allPrices = for {
+          charge <- productRatePlanCharges
+          price <- charge.pricing
+        } yield price
+
+        val gbpAmounts = allPrices.collect { case Price(Some(amount), "GBP") => amount }
```
johnduffell: future note: we match with "GBP", which works well for voucher, but for GW we may have to group by currency

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/productcatalog/PricesFromZuoraCatalog.scala
```diff
@@ -0,0 +1,109 @@
...
+
+  object Rateplan {
+    implicit val reads = Json.reads[Rateplan]
+
+  }
+
+  case class Product(
+    id: String,
+    name: String,
+    ProductType__c: Option[String],
```
johnduffell: I don't think this is used (yet) and not sure about name, although maybe they are both useful for debugging.

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/TypeConvert.scala
pvighi: yeah I wasn't sure about that because this is the last time we can actually log the stack trace properly. but I guess we could pass more information into the GenericError

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/productcatalog/PricesFromZuoraCatalog.scala
```diff
@@ -0,0 +1,109 @@
...
+  case class Rateplan(
+    id: String,
+    name: String,
+    FrontendId__c: Option[String],
+    productRatePlanCharges: List[RateplanCharge]
+  ) {
+    def toParsedPlan(planIdFor: ProductRatePlanId => Option[PlanId]): Option[PlanWithPrice] = {
+      val planId = planIdFor(ProductRatePlanId(id))
+
+      planId flatMap { planId =>
```
johnduffell: I think everything after here only uses productRatePlanCharges, so could be a separate function if necessary.

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/productcatalog/WireModel.scala
```diff
@@ -98,26 +100,26 @@ object WireModel {
...
-          WirePlanInfo.fromPlan(catalog.voucherEveryDayPlus, "Every day+"),
-          WirePlanInfo.fromPlan(catalog.voucherSaturday, "Saturday"),
-          WirePlanInfo.fromPlan(catalog.voucherSaturdayPlus, "Saturday+"),
-          WirePlanInfo.fromPlan(catalog.voucherSixDay, "Six day"),
-          WirePlanInfo.fromPlan(catalog.voucherSixDayPlus, "Six day+"),
-          WirePlanInfo.fromPlan(catalog.voucherSunday, "Sunday"),
-          WirePlanInfo.fromPlan(catalog.voucherSundayPlus, "Sunday+"),
-          WirePlanInfo.fromPlan(catalog.voucherWeekend, "Weekend"),
-          WirePlanInfo.fromPlan(catalog.voucherWeekendPlus, "Weekend+")
+          WirePlanInfo.fromPlan(catalog.voucherEveryDay, "Every day", priceFor),
```
johnduffell: small style thing - can we put the priceFor in an earlier parameter group so we only have to apply it once, and then pass in the other two params differently each time?

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/productcatalog/PricesFromZuoraCatalog.scala
pvighi: Yes, we should also think if we will return all the currencies in the catalog response or if we will take a currency param in the get request

handlers/new-product-api/src/test/scala/com/gu/newproduct/api/PricesFromZuoraCatalogTest.scala
johnduffell: if it's too big, it will be hard to maintain. how can anyone looking at the test or the json know which part is interesting?
Would it help maintainability to write a minimal json for the purposes of testing that can be understood by a human easily, and have an effects test to grab the real catalog from s3 in order to test properly?

johnduffell:
looks good in general, great work on this, I don't think we should have any tests that are not easy to maintain (which also includes the data)  We should stick to using effects tests to do more complicated things as much as possible.  I agree that we want to know that the behaviour hasn't changed for the overall format, but we should look into having some kind of "recorder" functionality so that the data is automatically regenerated without intervention.
I have made a few comments but overall everything is going the right way so thanks  👍 I'll leave the comments in your hands!

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/productcatalog/PricesFromZuoraCatalog.scala
pvighi: What do you mean ?

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/TypeConvert.scala
johnduffell: see what you mean, I was not sure if it was ok, but you're right at this point we know it's definitely going to return an internal server error anyway, so in the lambdas it would always be logged at the right time.  it would only be a NotFound error that could be expected exception.  I noticed you already changed it but it's ok to log it to be honest!

PR #187, opened 2018-08-22
quick manual batch runner for sf contact merge

johnduffell:
This is just a quick way to run a batch of contacts into the merger without having to use post man for each one.

I considered using an SQS queue rather than api gateway, and generating the batches off a different lambda, but due to lack of experience in this I decided it would be too time consuming.

### DIFF
--- a/handlers/sf-contact-merge/src/test/scala/manualTest/RunBatch.scala
+++ b/handlers/sf-contact-merge/src/test/scala/manualTest/RunBatch.scala
@@ -34,13 +34,14 @@ object RunBatch {
 
     val result = for {
       args <- GetArgs(rawArgs)
-      postIt = HttpOp(Http.response).setupRequest(post(args.apiKey))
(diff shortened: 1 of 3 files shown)


handlers/sf-contact-merge/src/test/scala/manualTest/RunBatch.scala
```diff
@@ -0,0 +1,101 @@
...
+  edit the config and add to the Program Arguments.
+   */
+
+  def main(rawArgs: Array[String]): Unit = {
+
+    val result = for {
+      args <- GetArgs(rawArgs)
+      postIt = HttpOp(Http.response).setupRequest(post(args.apiKey))
+      jsons <- ReadFile(args.fileName)
+      updateIt = postIt.setupRequest[JsonString](a => BodyAsString(a.value))
```
pvighi: minor : maybe rename this from a to something a little more descriptive

PR #188, opened 2018-08-22
Maintain correct first name when merging SF contacts

johnduffell:
When we were merging the SF contacts, we decide the winner based on the reliability of the address.  This is more likely for print subscriptions.

However, where one of the source contacts was created via identity, that would have a first name, whereas the target SF one may not have.

This PR updates the validation to make sure that all zuora accounts have the same last name before going ahead with the merge.  In that case, if the losing account has just a full stop for the name, it will replace it with the one from the identity user.  If the first name is present in zuora already, it will write the winning one into the winning contact.  If there is no first name from either, it will just replace with a full stop.  This will ensure that the zuora sync will work.

I have also added a check that the winning SF contact id really is in the ones connected from zuora.  If not then it could indicate an issue.

I have also been doing a bit of tidyup of the function signatures.  Rather than explicitly specifying the function type, I have used a trait with a single abstract method.  This is experimental and may encourage excessive numbers of parameters, but see what you think.

I think there are a few codacy things to work through and I need to check the important lines have code coverage as I have a feeling I need to add some tests.  But have a look at the overall view in the mean time. @pvighi @paulbrown1982 @abhichawla


build.sbt
```diff
@@ -88,16 +88,6 @@ lazy val salesforce = all(project in file("lib/salesforce"))
     libraryDependencies ++= Seq(okhttp3, logging, scalaz, playJson, scalatest)
   )
 
-lazy val `update-sf-identityid` = all(project in file("lib/sf/update-sf-identityid"))
```
johnduffell: this was added as an attempt to share the update SF identity id code.  However now that we are updating first name as well, it is diverging a lot so I have given up and had two copies.

handlers/sf-contact-merge/src/main/scala/com/gu/sf_contact_merge/Handler.scala
```diff
@@ -48,48 +50,73 @@ object Handler {
...
-        UpdateSalesforceIdentityId(sfPatch).runRequestMultiArg,
-        UpdateAccountSFLinks(requests.put),
-        _,
-        _,
-        _
-      ),
-      _
-    ))
+    } yield Operation.noHealthcheck {
+      WireRequestToDomainObject {
```
johnduffell: previously we were trying to make a big for comprehension working from the api gateway request level all the way down to the domain level.  Then we were trying to artificially split between "validate" and "update" once at the domain level.
I decided it would make more sense to split by the level, so the `WireRequestToDomainObject` brings us down to the domain level and then steps just worries about that level

handlers/sf-contact-merge/src/main/scala/com/gu/sf_contact_merge/Handler.scala
```diff
@@ -48,48 +50,73 @@ object Handler {
...
       accountAndEmails <- getIdentityAndZuoraEmailsForAccounts(mergeRequest.zuoraAccountIds)
         .toApiGatewayOp("getIdentityAndZuoraEmailsForAccounts")
-      _ <- validation(mergeRequest.sFPointer.identityId, accountAndEmails)
+      _ <- validateEmails(accountAndEmails.map(_.emailAddress))
+      _ <- validateIdentityIds(accountAndEmails.map(_.identityId), mergeRequest.sFPointer.identityId)
+        .toApiGatewayReturnResponse(ApiGatewayResponse.notFound)
+      _ <- validateLastNames(accountAndEmails.map(_.lastName))
+      maybeIdentityFirstName = firstNameForIdentityAccount(accountAndEmails.map { info => (info.identityId, info.firstName) })
+      maybeOldFirstName = firstNameForSFContact(mergeRequest.sFPointer.sfContactId, accountAndEmails.map { info => (info.sfContactId, info.firstName) })
+      firstNameToUse <- firstNameIfNot(maybeOldFirstName, maybeIdentityFirstName)
```
johnduffell: should probably combine these three previous lines into one function - TODO

handlers/sf-contact-merge/src/main/scala/com/gu/sf_contact_merge/validate/AssertSame.scala
```diff
@@ -1,14 +1,17 @@
...
-import com.gu.sf_contact_merge.getaccounts.GetEmails.EmailAddress
 import com.gu.util.apigateway.ApiGatewayResponse
 import com.gu.util.reader.Types.ApiGatewayOp
 import com.gu.util.reader.Types.ApiGatewayOp.{ContinueProcessing, ReturnWithResponse}
 
-object AssertSameEmails {
+object AssertSame {
 
-  def apply(emailAddresses: List[Option[EmailAddress]]): ApiGatewayOp[Unit] =
+  def apply[EmailAddress]: AssertSame[EmailAddress] = (emailAddresses: List[EmailAddress]) =>
```
pvighi: This is now generic right? because all the names and the error message refer to email addresses

handlers/sf-contact-merge/src/main/scala/com/gu/sf_contact_merge/validate/EnsureNoAccountWithWrongIdentityId.scala
```diff
@@ -1,27 +1,38 @@
 package com.gu.sf_contact_merge.validate
 
-import com.gu.identityBackfill.salesforce.UpdateSalesforceIdentityId.IdentityId
+import com.gu.sf_contact_merge.update.UpdateSalesforceIdentityId.IdentityId
 
 object EnsureNoAccountWithWrongIdentityId { // make sure all accounts are either this identity id or none
+
+  val apply: EnsureNoAccountWithWrongIdentityId = //FIXME use the validation from new-product-api
```
pvighi: was this FIXME meant to be fixed in this PR or later ?

handlers/sf-contact-merge/src/test/scala/com/gu/sf_contact_merge/validate/AssertSameEmailsTest.scala
```diff
@@ -8,21 +8,21 @@ class AssertSameEmailsTest extends FlatSpec with Matchers {
   it should "be all ok with the same email addresses" in {
```
pvighi: minor: This should probably be called AssertSameTest now and maybe the test descriptions could not refer to emails

handlers/sf-contact-merge/src/main/scala/com/gu/sf_contact_merge/validate/AssertSame.scala
```diff
@@ -1,14 +1,17 @@
...
     if (emailAddresses.distinct.size == 1) ContinueProcessing(())
     else ReturnWithResponse(ApiGatewayResponse.notFound("those zuora accounts had differing emails"))
 
 }
+
+trait AssertSame[EmailAddress] {
+  def apply(emailAddresses: List[EmailAddress]): ApiGatewayOp[Unit]
```
pvighi: If we ever use the new-product-api validation for EnsureNoAccountWithWrongIdentityId we can probably use it for this as well

handlers/sf-contact-merge/src/test/scala/com/gu/sf_contact_merge/update/MoveIdentityIdTest.scala
```diff
@@ -0,0 +1,87 @@
...
+import com.gu.util.resthttp.Types.ClientSuccess
+import org.scalatest.{FlatSpec, Matchers}
+
+class MoveIdentityIdTest extends FlatSpec with Matchers {
+
+  it should "not edit identity id at all if we're not setting one" in {
+
+    var order = List[String]()
+
+    def setOrClearIdentityId(sfContactId: SFContactId, idid: SFContactUpdate): Types.ClientFailableOp[Unit] = {
```
pvighi: is idid the best name for this param ?

handlers/sf-contact-merge/src/test/scala/com/gu/sf_contact_merge/validate/AssertSameEmailsTest.scala
johnduffell: good spot, I'm actually testing assert same specifically in the case of Email where it is an Option.  So I think I will leave it the same for now, unless you prefer to have the class as AssertSameTest and just leave the descriptions mentioning it's specifically for the emails case.

johnduffell: I have renamed the actual AssertSame type parameter though, as that code is generic

handlers/sf-contact-merge/src/main/scala/com/gu/sf_contact_merge/Handler.scala
```diff
@@ -48,48 +50,87 @@ object Handler {
...
+    namesForContactIds.find(_.sfContactId == newSFContactId)
+      .toApiGatewayContinueProcessing(ApiGatewayResponse.notFound("winning contact id wasn't in any zuora account"))
+      .map(_.firstName)
+  }
+
+  def firstNameIfNot(maybeOld: Option[FirstName], maybeIdentity: Option[FirstName]): Option[FirstName] = {
+    (maybeOld, maybeIdentity) match {
+      case (Some(old), _) => Some(old)
+      case (None, Some(identityFirstName)) => Some(identityFirstName)
+      case (old, _) => old // will be none
```
pvighi: is this different than doing maybeOld orElse MaybeIdentity ?
if it is maybe we don't need this function at all

handlers/sf-contact-merge/src/test/scala/com/gu/sf_contact_merge/update/MoveIdentityIdTest.scala
```diff
@@ -11,76 +11,64 @@ import org.scalatest.{FlatSpec, Matchers}
...
 
-  it should "not edit identity id at all if we're not setting one" in {
+  final class Var() {
 
-    var order = List[String]()
+    var order = List[String]() // we want to check ordering of side effects...
 
-    def setOrClearIdentityId(sfContactId: SFContactId, idid: SFContactUpdate): Types.ClientFailableOp[Unit] = {
-      order = (idid match {
+    def setOrClearIdentityId(sfContactId: SFContactId, sfUpdateRequest: SFContactUpdate): Types.ClientFailableOp[Unit] = {
```
johnduffell: @pvighi  github had a funny and list your comment about idid being a funny name, but anyway I've done a bit of a rejig on this test now - it wasn't great and it still needs a var due to side effects.

handlers/sf-contact-merge/src/main/scala/com/gu/sf_contact_merge/Handler.scala
johnduffell: oh - good spot that's the plain english version too 😆

handlers/sf-contact-merge/src/main/scala/com/gu/sf_contact_merge/getaccounts/GetEmails.scala
```diff
@@ -12,22 +12,35 @@ object GetEmails {
   case class ContactId(value: String) extends AnyVal
 
   case class EmailAddress(value: String) extends AnyVal
+  case class FirstName(value: String) extends AnyVal
+  case class LastName(value: String) extends AnyVal
+  case class Record(emailAddress: Option[EmailAddress], firstName: Option[FirstName], lastName: LastName)
```
pvighi: The naming is a little confusing now. you still call GetEmails.apply buy it returns a map of ContactId -> Record
Also record seems like a really generic name, maybe we could use something more descriptive like personalDetails or ZuoraContactDetails

handlers/sf-contact-merge/src/test/scala/com/gu/sf_contact_merge/GetFirstNameToUseTest.scala
```diff
@@ -0,0 +1,25 @@
...
+    val contactIdsFromZuora = List(NameForContactId(SFContactId("wrong"), None))
+    val actual = GetFirstNameToUse.firstNameForSFContact(SFContactId("winningSfContact"), contactIdsFromZuora)
+    actual.toDisjunction.leftMap(_.statusCode) should be(-\/("404"))
+  }
+
+  it should "be happy if the winning contact is in a zuora account" in {
+    val someFirstName = Some(FirstName("hi"))
+    val contactIdsFromZuora = List(NameForContactId(SFContactId("winningSfContact"), someFirstName))
+    val actual = GetFirstNameToUse.firstNameForSFContact(SFContactId("winningSfContact"), contactIdsFromZuora)
+    actual should be(ContinueProcessing(someFirstName))
```
pvighi: really minor: 
 don't you think in cases like this we need to test that the correct value is picked by providing some other contacts and asserting that it picks the right one? or would you rather have the minimal test to keep it simple?

handlers/sf-contact-merge/src/test/scala/com/gu/sf_contact_merge/update/MoveIdentityIdTest.scala
```diff
@@ -0,0 +1,75 @@
...
+
+    val mock = new Var()
+
+    val maybeIdentityId: Option[IdentityId] = None
+    val sfPointer = LinksFromZuora(SFContactId("contnew"), CRMAccountId("crmcrm"), maybeIdentityId)
+    val maybeContactId = Some(OldSFContact(SFContactId("contold")))
+
+    val actual = MoveIdentityId(mock.setOrClearIdentityId)(sfPointer, maybeContactId, Some(FirstName("hello")))
+
+    mock.order.reverse should be(List("clear contold setname: DontChangeFirstName", "clear contnew setname: SetFirstName(FirstName(hello))"))
```
pvighi: really really tiny: maybe mock.order could return the reversed list already, otherwise it should be called reverseOrder

handlers/sf-contact-merge/src/main/scala/com/gu/sf_contact_merge/getaccounts/GetEmails.scala
johnduffell: good spot, thanks, I'll rename these all

handlers/sf-contact-merge/src/test/scala/com/gu/sf_contact_merge/GetFirstNameToUseTest.scala
johnduffell: I didn't do so because I wanted to test the boolean condition and its outcome rather than the built in `find` functionality, however I'll add one for completeness

handlers/sf-contact-merge/src/test/scala/com/gu/sf_contact_merge/update/MoveIdentityIdTest.scala
johnduffell: I'll just get rid of the reversedness in the first place, it won't be slow with only a couple of items in the list...

handlers/sf-contact-merge/src/test/scala/com/gu/sf_contact_merge/update/updateSFIdentityId/UpdateSalesforceIdentityIdTest.scala
```diff
@@ -0,0 +1,71 @@
...
+    val actual = UpdateSalesforceIdentityId.toRequest(
+      SFContactId("contactsf"),
+      SFContactUpdate(Some(IdentityId("identityid")), SetFirstName(FirstName("firstname")))
+    )
+    val expectedJson = JsObject(Seq(
+      "IdentityID__c" -> JsString("identityid"),
+      "FirstName" -> JsString("firstname")
+    ))
+    val expected = new PatchRequest(expectedJson, RelativePath("/services/data/v20.0/sobjects/Contact/contactsf"))
+    actual should be(expected)
```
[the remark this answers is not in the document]
johnduffell: if we decide to switch over, there's a lot of tech debt we can plough through to get off the old style request maker stuff.

PR #189, opened 2018-08-24
Voucher confirmation emails

pvighi:
Send thank you emails when adding voucher subscriptions using the new product api


handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/email/contributions/SendConfirmationEmailContributions.scala
```diff
@@ -22,21 +25,19 @@ object SendConfirmationEmailContributions extends Logging {
...
     val maybeContributionFields = toContributionFields(getCurrentDate(), data)
-    val response = for {
+    for {
       etPayload <- toPayload(maybeContributionFields)
-      sendMessageResult <- etSqsSend(etPayload).toAsyncApiGatewayOp("sending sqs message")
+      sendMessageResult <- etSqsSend(etPayload).toAsyncApiGatewayOp("sending contribution email sqs message")
     } yield sendMessageResult
 
-    response.replace(ContinueProcessing(()))
-
```
pvighi: making the email sending code always return success even if there was an error made testing harder.  I moved the error recovery outside to make testing straightforward. Now it is the callers responsibility to ignore and logs any errors returned by this

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/email/voucher/SendConfirmationEmailVoucher.scala
```diff
@@ -0,0 +1,33 @@
...
+    voucherEmailData.contacts.soldTo.email.map { email =>
+      val payload = ETPayload(email = email.value, fields = voucherEmailData, DataExtensionName("paper-voucher"))
+      ContinueProcessing(payload).toAsync
+    }.getOrElse {
+      val errorLogMessage = "No email address in zuora sold to contact, skipping voucher thank you email"
+      logger.warn(errorLogMessage)
+      val response = ApiGatewayResponse.messageResponse( "500","Internal server error")
+      ReturnWithResponse(response).toAsync
+    }
+}
```
pvighi: there is probably  a  chance here to  reduce duplication here, in particular because we will be adding more products. It would be nice if we could move of the code that is the same for all welcome emails to a shared location

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/email/voucher/VoucherEmailData.scala
```diff
@@ -0,0 +1,100 @@
...
+)
+
+object VoucherEmailData {
+  implicit val writes: Writes[VoucherEmailData] = (data: VoucherEmailData) => {
+    val fields: Map[String, String] = VoucherEmailFields(data)
+    Json.toJson(fields)
+  }
+}
+
+object VoucherEmailFields {
```
pvighi: we could not use the usual domain case class -> wire case class -> json strategy here as the play json serialiser macros only works up to a certain number of fields which we are exceeding here. Instead we do domain class -> Map[String,String] -> json

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/productcatalog/NewProductApi.scala
```diff
@@ -0,0 +1,55 @@
...
+import java.time.DayOfWeek
+import java.time.DayOfWeek.{MONDAY, SATURDAY, SUNDAY}
+import com.gu.newproduct.api.addsubscription.Formatters._
+import com.gu.newproduct.api.productcatalog.PlanId._
+
+object NewProductApi {
+  def catalog(priceFor: PlanId => Option[AmountMinorUnits]): Catalog = {
+
+    def paymentPlanFor = priceFor andThen { maybeAmount =>
+      maybeAmount.map(amount => PaymentPlan(s"GBP ${amount.formatted} every month"))
```
pvighi: now we share this description between the catalog response and the voucher email response. Since David wanted the catalog response to use GBP instead of "£". This had the side effect of making the confirmation email also use this format, which is different from the format used in the emails triggered from the front end.
This is a minor thing but maybe we should consider changing everything to use £ (or have different format for the email and the  catalog response)

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/productcatalog/Catalog.scala
```diff
@@ -35,44 +32,6 @@ case class Catalog(
   val planForId: Map[PlanId, Plan] = allPlans.map(x => x.id -> x).toMap
 }
 
-object NewProductApi {
```
pvighi: The catalog class definitions and the actual catalog instance for new product api were separated into different files to avoid the circular dependency errors. The object has the same name and is in the same package so not much should have changed for scala

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/Formatters.scala
```diff
@@ -1,7 +1,16 @@
...
+import com.gu.newproduct.api.productcatalog.AmountMinorUnits
+
 object Formatters {
+
   implicit class AmountOps(amountMinorUnits: AmountMinorUnits) {
     def formatted: String = (amountMinorUnits.value / BigDecimal(100)).bigDecimal.setScale(2).toPlainString
   }
+
+  implicit class SortCodeOps(sortCode: SortCode) {
+    def hyphenated = s"${sortCode.value.substring(0, 2)}-${sortCode.value.substring(2, 4)}-${sortCode.value.substring(4, 6)}"
```
johnduffell: it's existing code I realise, but a better model for sort code would probably be 3 separate strings or similar, rather than a single one in an AnyVal and then chopping it up later in the program, as there's no guarantee that the one in value would or wouldn't have the dashes in it.

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/Handler.scala
```diff
@@ -121,21 +122,33 @@ object Steps {
       zuoraCreateSubRequest = createZuoraSubRequest(request, acceptanceDate, Some(chargeOverride), contributionZuoraIds.productRatePlanId)
       subscriptionName <- createSubscription(zuoraCreateSubRequest).toAsyncApiGatewayOp("create monthly contribution")
       contributionEmailData = toContributionEmailData(request, account.currency, paymentMethod, acceptanceDate, contacts.billTo, amountMinorUnits)
-      _ <- sendConfirmationEmail(contributionEmailData)
+      _ <- sendConfirmationEmail(contributionEmailData).logAndContinue("send contribution confirmation email")
```
johnduffell: consider stronger wording to make it clearer it is turning left into right - recoverAndLog or logAndIgnoreError or something?

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/Handler.scala
```diff
@@ -168,12 +187,22 @@ object Steps {
...
+        catalog.planForId,
+        getZuoraIdForVoucherPlan,
+        getVoucherData,
+        isValidStartDateForPlan,
+        createSubscription,
+        sendVoucherEmail
+      ) _
 
       addSubSteps = handleRequest(
         addContribution = contributionSteps,
```
johnduffell: I think we discussed about the length of this for comprehension last week, based on that it might be worth splitting into the shared dependencies set up first, then two functions, once produces the wired up contributions stuff from the shared deps, one does the same for voucher, and those in turn get wired into the handleRequest.

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/Formatters.scala
pvighi: that is true, but I don't think the sort code is 3 different numbers, it's just a common way to format it.

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/Handler.scala
```diff
@@ -229,10 +258,10 @@ object Steps {
     )
   }
 
-  def emailQueueFor(stage: Stage) = stage match {
-    case Stage("PROD") => QueueName("contributions-thanks")
-    case Stage("CODE") => QueueName("contributions-thanks")
-    case _ => QueueName("contributions-thanks-dev")
+  case class EmailQueueNames(contributions: QueueName, voucher: QueueName)
+  def emailQueuesFor(stage: Stage) = stage match {
+    case Stage("PROD") | Stage("CODE") => EmailQueueNames(contributions = QueueName("contributions-thanks"), voucher = QueueName("subs-welcome-email"))
```
johnduffell: nice, didn't know you could do the OR thing

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/Handler.scala
pvighi: yes, but I would rather have a pure refactoring PR instead of including it in this one

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/email/contributions/SendConfirmationEmailContributions.scala
johnduffell: good idea, keep this logic more focussed

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/email/voucher/SendConfirmationEmailVoucher.scala
johnduffell: yep you are right this could even be business duplication - if cex say "all confirmation emails need a disclaimer at the bottom" or something, we would not want to edit them all separately

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/email/voucher/SendConfirmationEmailVoucher.scala
```diff
@@ -0,0 +1,33 @@
...
+import com.gu.util.reader.Types.ApiGatewayOp.{ContinueProcessing, ReturnWithResponse}
+import com.gu.util.reader.AsyncTypes._
+
+import scala.concurrent.Future
+
+object SendConfirmationEmailVoucher extends Logging {
+
+  def apply(
+    etSqsSend: ETPayload[VoucherEmailData] => Future[Unit],
+    getCurrentDate: () => LocalDate,
```
johnduffell: this file won't be reformatting due to the trailing comma

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/email/voucher/SendConfirmationEmailVoucher.scala
```diff
@@ -0,0 +1,33 @@
...
+  )(data: VoucherEmailData): AsyncApiGatewayOp[Unit] = for {
+    etPayload <- toPayload(data)
+    sendMessageResult <- etSqsSend(etPayload).toAsyncApiGatewayOp("sending voucher email sqs message")
+  } yield sendMessageResult
+
+  def toPayload(voucherEmailData: VoucherEmailData): AsyncApiGatewayOp[ETPayload[VoucherEmailData]] =
+    voucherEmailData.contacts.soldTo.email.map { email =>
+      val payload = ETPayload(email = email.value, fields = voucherEmailData, DataExtensionName("paper-voucher"))
+      ContinueProcessing(payload).toAsync
+    }.getOrElse {
```
johnduffell: tiny: .map then .getOrElse is fine, usually I use pattern match in this situation just to tie the whole block together

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/email/voucher/SendConfirmationEmailVoucher.scala
```diff
@@ -0,0 +1,33 @@
...
+      val errorLogMessage = "No email address in zuora sold to contact, skipping voucher thank you email"
+      logger.warn(errorLogMessage)
+      val response = ApiGatewayResponse.messageResponse( "500","Internal server error")
+      ReturnWithResponse(response).toAsync
```
johnduffell: can move the toAsync to line 19 and delete from here and 26

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/email/voucher/VoucherEmailData.scala
```diff
@@ -0,0 +1,100 @@
...
+      "subscription_rate" -> data.plan.paymentPlan.map(_.value).getOrElse("")
+    ) ++ paymentMethodFields(data.paymentMethod) ++ addressFields(data.contacts)
+
+  }
+
+  def toDescription(methodType: PaymentMethodType) = methodType match {
+    case CreditCardReferenceTransaction | CreditCard => "Credit/Debit Card"
+    case BankTransfer => "Direct Debit"
+    case PayPal => "PayPal"
+    case Other => "" //should not happen
```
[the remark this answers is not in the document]
johnduffell: that's a shame we have to duplicate that, we have lost from the types/code the fact that the email sending depends on the validation.  perhaps at least update the scala code comment to point that out.

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/zuora/GetContacts.scala
```diff
@@ -12,43 +12,98 @@ object GetContacts {
...
-    case class GetContactsResponse(billToContact: ZuoraBillTo, soldToContact: ZuoraSoldTo)
-
-    implicit val ZuoraBillToReads = Json.reads[ZuoraBillTo]
-    implicit val ZuoraSoldToReads = Json.reads[ZuoraSoldTo]
-    implicit val ZuoraContactsReads = Json.reads[GetContactsResponse]
+    def toField[FIELD](maybeString: Option[String], toField: String => FIELD): Option[FIELD] = maybeString match {
+      case None => None
+      case Some("") => None
+      case Some(string) => Some(toField(string))
+    }
```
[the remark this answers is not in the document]
johnduffell: it seems to do it sometimes but not others.  Another way is to do maybeString.filter(_ != "")

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/productcatalog/NewProductApi.scala
johnduffell: good spot, I wonder whether people would be confused by that, especially if it's a UK only product.  I would say err on the side of simplicity for now and keep the same representation for everything.

handlers/new-product-api/src/test/scala/com/gu/newproduct/api/addsubscription/email/contributions/SendConfirmationEmailContributionsTest.scala
```diff
@@ -82,34 +91,32 @@ class SendConfirmationEmailContributionsTest extends FlatSpec with Matchers {
 
     val send = SendConfirmationEmailContributions(sqsSend, today) _
 
-    val res = Await.result(send(testData).underlying, 3 seconds)
-
-    res.shouldBe(ContinueProcessing(()))
+    send(testData).underlying map {
+      res => res shouldBe ContinueProcessing(())
+    }
```
johnduffell: the aync tests are a lot nicer, but for some reason I just feel nervous about assertions being lost

handlers/new-product-api/src/test/scala/com/gu/newproduct/api/addsubscription/email/voucher/SendConfirmationEmailVoucherTest.scala
```diff
@@ -0,0 +1,68 @@
...
+import com.gu.newproduct.api.addsubscription.email.{DataExtensionName, ETPayload}
+import com.gu.newproduct.api.addsubscription.zuora.CreateSubscription.SubscriptionName
+import com.gu.newproduct.api.productcatalog.PlanId.VoucherSunday
+import com.gu.newproduct.api.productcatalog.{Plan, PlanDescription}
+import com.gu.util.apigateway.ApiGatewayResponse
+import com.gu.util.reader.Types.ApiGatewayOp.{ContinueProcessing, ReturnWithResponse}
+import org.scalatest.{AsyncFlatSpec, Matchers}
+
+import scala.concurrent.Future
+import scala.language.postfixOps
```
johnduffell: what does postfixOps do?

handlers/new-product-api/src/test/scala/com/gu/newproduct/api/addsubscription/email/voucher/SendConfirmationEmailVoucherTest.scala
```diff
@@ -0,0 +1,68 @@
...
+
+class SendConfirmationEmailVoucherTest extends AsyncFlatSpec with Matchers {
+  it should "send voucher confirmation email" in {
+    def sqsSend(payload: ETPayload[VoucherEmailData]): Future[Unit] = Future {
+      payload shouldBe ETPayload("soldToEmail@mail.com", testVoucherData, DataExtensionName("paper-voucher"))
```
johnduffell: I wonder if we can implement the SQS using the HttpOp a bit like we did with Patch so far, then we don't need to fake a future and then wait for it.

handlers/new-product-api/src/test/scala/com/gu/newproduct/api/addsubscription/email/voucher/VoucherEmailDataTest.scala
```diff
@@ -0,0 +1,145 @@
...
+      """.stripMargin
+
+    actualJson shouldBe Json.parse(expected)
+  }
+
+  it should "not include direct debit fields if payment method is not direct debit" in {
+
+    val cardVoucherData = directDebitVoucherData.copy(paymentMethod = NonDirectDebitMethod(ActivePaymentMethod, CreditCard))
+
+    val directDebitFields = List("bank_account_no", "bank_sort_code", "account_holder", "mandate_id")
```
johnduffell: could this be `directDebitFieldNames` ?

handlers/new-product-api/src/test/scala/com/gu/newproduct/api/addsubscription/email/voucher/VoucherEmailDataTest.scala
```diff
@@ -0,0 +1,145 @@
...
+
+  }
+
+  it should "IncludesDigipack should be true for plus plans " in {
+
+    val digipackPlans = List(VoucherEveryDayPlus, VoucherWeekendPlus, VoucherSixDayPlus, VoucherSaturdayPlus, VoucherSundayPlus)
+
+    val allDigipackPlanFields: List[Map[String, String]] = fieldsForPlanIds(digipackPlans)
+
+    allDigipackPlanFields.forall(_.get("IncludesDigipack").contains("true"))
```
johnduffell: nice, data driven testing!  🎉

johnduffell:
looks great, there's a lot there, but I think it's well broken down so the tests look great and the coverage is good overall.
Adding all the extra fields has made the testing harder to see the wood from the trees but that's hard to avoid.
I don't think any reason not to go for it, feel free to check the comments I made but most of them are just observations and not really suggestions
👍 💯 🚀

handlers/new-product-api/src/test/scala/com/gu/newproduct/api/addsubscription/email/voucher/SendConfirmationEmailVoucherTest.scala
pvighi: probably something that was needed in a previous version of the code but now IntelliJ seems to think it is needed, but it's not

PR #190, opened 2018-08-31
Small PR: ignore the case of the lastname when merging contacts

johnduffell:
When I am doing the name checking before merging contacts, I noticed that many of them have completely lower case surname on one of the contacts.
I don't think that's something to block merging, so I have added the code to ignore that.

@pvighi  @abhichawla


handlers/sf-contact-merge/src/main/scala/com/gu/sf_contact_merge/Handler.scala
```diff
@@ -54,8 +54,8 @@ object Handler {
       WireRequestToDomainObject {
         DomainSteps(
           GetIdentityAndZuoraEmailsForAccountsSteps(zuoraQuerier, _),
-          AssertSame[Option[EmailAddress]]("emails"),
-          AssertSame[LastName]("last names"),
+          AssertSame.emailAddress,
+          AssertSame.lastName,
```
johnduffell: I have pushed this out of the wiring to allow testing

PR #191, opened 2018-08-31
added one week to voucher start date validation rules

pvighi:
[the date validation in subs frontend](https://github.com/guardian/subscriptions-frontend/blob/e057a5b8dfd8ec94c48ebae4455f0e1fcb6aed5c/assets/javascripts/modules/checkout/voucherFields.jsx#L37) calculates voucher selectable dates adding 4 weeks to the first selectable date (or 3 weeks if it before wednesday 6 am). 

The validation here is simpler, the selectable window is 4 weeks long and starts on the next cut off day. The problem is when the cut off day is not a selectable day of the week for the selected plan. In these cases the window will be effectively smaller after applying day of the week rules.

The fix was to add an extra week to the voucher validation rules. 
I'm not sure if there's a corner case that would make this not match the frontend results. In any case, I don't think it would be a problem if CSRs can add vouchers that start a week later that the latest selectable date in the frontend

### DIFF
--- a/handlers/new-product-api/src/main/scala/com/gu/newproduct/api/productcatalog/NewProductApi.scala
+++ b/handlers/new-product-api/src/main/scala/com/gu/newproduct/api/productcatalog/NewProductApi.scala
@@ -15,7 +15,7 @@ object NewProductApi {
     val voucherWindowRule = WindowRule(
       maybeCutOffDay = Some(DayOfWeek.TUESDAY),
       maybeStartDelay = Some(DelayDays(20)),
-      maybeSize = Some(WindowSizeDays(28))
+      maybeSize = Some(WindowSizeDays(35))
     )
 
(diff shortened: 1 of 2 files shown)


handlers/new-product-api/src/main/scala/com/gu/newproduct/api/productcatalog/NewProductApi.scala
```diff
@@ -15,7 +15,7 @@ object NewProductApi {
     val voucherWindowRule = WindowRule(
       maybeCutOffDay = Some(DayOfWeek.TUESDAY),
       maybeStartDelay = Some(DelayDays(20)),
-      maybeSize = Some(WindowSizeDays(28))
+      maybeSize = Some(WindowSizeDays(35))
```
johnduffell: tiny idea: it's fine as it is but we could `WindowSizeDays.fromWeeks(5)` to make it slightly more self explanatory

johnduffell:
👍 👍 simple but effective

PR #192, opened 2018-08-31
Ignore provided identity id when merging contacts

johnduffell:
So I am linking zuora subs together with the winning SF contact, and also updating the identity id at the same time.
In the JSON provided by SF there was an identity id which I was checking to make sure it wasn't suggesting to delete an existing one.
However it turns out based on discussion this is not the right approach, the identity id provided by salesforce is not relevant and should be ignored, as it is the existing one on the winning contact.
This means that ones where the identity contact wasn't the winning one were being rejected by the pre checks.

This PR updates it to work out the correct identity id from zuora, and then write that to the winning contact and the zuora accounts.

Abhishek will remove the field from the provided JSON at some point if it's not too much trouble, as it's redundant and it will prevent the possibility of future confusion.

@pvighi @paulbrown1982 @abhichawla

### DIFF
--- a/handlers/cancellation-sf-cases/src/main/scala/com/gu/cancellation/sf_cases/TypeConvert.scala
+++ b/handlers/cancellation-sf-cases/src/main/scala/com/gu/cancellation/sf_cases/TypeConvert.scala
@@ -5,8 +5,8 @@ import com.gu.util.resthttp.Types.ClientFailableOp
 
 object TypeConvert {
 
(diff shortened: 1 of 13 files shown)


johnduffell:
since the builds are sick again, I will save this one for when I return.
Note to future self: I have a list of ones I need to reprocess in `/merge/loseidentity`  I just need to select the ones with the identity id error, and run them again once this is deployed.

PR #194, opened 2018-09-12
add the address fields to the zuora query

johnduffell:
We are merging together salesforce contacts where the contacts have the same email address.  This is good for GDPR because we have a coherent view of our connection with an individual and it's also good for self service as more of them will have an identity id afetrwards.

We always prefer an active subscription to be the winner when we are merging, but in some cases the cancelled one would be a recently cancelled paper sub, and the active could be a Friend.  The friend may not even have a billing address, so although the active should probably win, we should probably treat the adress we already have on the old contact to be the winner.

This PR is the first stage, just to read the billing addresses from zuora.  The next PRs will actually detect if the winner has no address by checking Address1, and if not, copy it over from one of the losing zuora billing contacts arbitrarily.

I need to find out that editing the address in zuora will actually reflect in salesforce to confirm that this technique is valid.  I didn't do this in salesforce because at present we don't look at the salesforce contacts at all. and we don't even have the losing contact ids without getting them from zuora.

@pvighi @paulbrown1982

### DIFF
--- a/handlers/sf-contact-merge/src/main/scala/com/gu/sf_contact_merge/getaccounts/GetZuoraContactDetails.scala
+++ b/handlers/sf-contact-merge/src/main/scala/com/gu/sf_contact_merge/getaccounts/GetZuoraContactDetails.scala
@@ -14,13 +14,36 @@ object GetZuoraContactDetails {
   case class EmailAddress(value: String) extends AnyVal
   case class FirstName(value: String) extends AnyVal
   case class LastName(value: String) extends AnyVal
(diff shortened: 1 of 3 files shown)


paulbrown1982:
👍 LGTM

johnduffell:
closing because this pr was trying to do it from zuora but I found that if we write to zuora it doesn't make its way into salesforce. I decided not to read from zuora and write to SF because of the complex rules of zuora to salesforce mapping https://github.com/guardian/membership-workflow/blob/master/app/model/SalesforceToZuoraMapper.scala

PR #195, opened 2018-09-14
make sure the winning contact has a billing address if there is one

johnduffell:
Replacing the previous PR https://github.com/guardian/zuora-auto-cancel/pull/194

## Background

We are merging together salesforce contacts where the contacts have the same email address. This is good for GDPR because we have a coherent view of our connection with an individual and it's also good for self service as more of them will have an identity id afetrwards.

We always prefer an active subscription to be the winner when we are merging, but in some cases the cancelled one would be a recently cancelled paper sub, and the active could be a Friend. The friend may not even have a billing address, so although the active should probably win, we should probably treat the adress we already have on the old contact to be the winner.

## this pr
This Pr will read the address from the winning contact in SF. If that doesn't have either the street address or the country, it will count the address as missing, and search for an address in the other contacts.  If another contact has an address it will just overwrite with that one.

The previous PR was trying to do it from zuora but I found that if we write to zuora it doesn't make its way into salesforce.  I decided not to read from zuora and write to SF because of the complex rules of zuora to salesforce mapping https://github.com/guardian/membership-workflow/blob/master/app/model/SalesforceToZuoraMapper.scala

Since the HttpOp has generally seemed to work OK for patch, I have added it for get as well and extended it with a flatMap method.  I have also followed the way of making traits/case classes to mark functions for specific purposes rather than just using their type.

@pvighi @paulbrown1982


handlers/sf-contact-merge/src/main/scala/com/gu/sf_contact_merge/Handler.scala
```diff
@@ -57,10 +60,11 @@ object Handler {
           AssertSame.emailAddress,
           AssertSame.lastName,
           EnsureNoAccountWithWrongIdentityId.apply,
-          MoveIdentityId(
-            UpdateSalesforceIdentityId(sfPatch).runRequestMultiArg
+          UpdateSFContacts(
+            UpdateSalesforceIdentityId(sfPatch)
           ),
```
pvighi: very minor: maybe this cold all be in the same line now ?

handlers/sf-contact-merge/src/main/scala/com/gu/sf_contact_merge/getsfcontacts/GetSfAddressOverride.scala
```diff
@@ -0,0 +1,39 @@
...
+      addressInWinner <- getSfAddress.apply(winningSFContactId)
+      winning <- addressInWinner match {
+        case Some(_) => ClientSuccess(None) // have a fine address already
+        case None =>
+          val potentialContactsWithAddresses = allSFContactIds.filter(_ != winningSFContactId).distinct.toStream
+          val potentialAddresses = potentialContactsWithAddresses.map(getSfAddress.apply)
+          val maybeSuitableAddress = potentialAddresses.find {
+            case fail: ClientFailure => true // give up, error
+            case ClientSuccess(Some(_)) => true // found a decent address
+            case ClientSuccess(None) => false
```
pvighi: you could group the two cases that return the same if you want. But you might want to keep it the way it is to improve readability of the comments
```scala
val maybeSuitableAddress = potentialAddresses.find {
   case ClientSuccess(Some(_) |  _: ClientFailure => true // return found address or error
   case ClientSuccess(None) => false // no suitable address
}
```

lib/salesforce/src/test/scala/com/gu/salesforce/dev/SFEffectsData.scala
```diff
@@ -6,7 +6,7 @@ object SFEffectsData {
   // this class represents a compile safe set of data that we (hope) is present in dev sf for the Effects tests.
 
   // this has the first name and last name, but no record type or country
-  val testContactHasName = SFContactId("0036E00000NLzPkQAL")
+  val testContactHasNamePhoneOtherAddress = SFContactId("0036E00000NLzPkQAL")
```
pvighi: hopefully we can add something in sf to make sure no one changes these contacts in a way that it breaks the tests

lib/restHttp/src/main/scala/com/gu/util/resthttp/HttpOp.scala
```diff
@@ -1,39 +1,56 @@
...
   // convenience, tuples for you
-  implicit class HttpOpOps[IN](httpOp: HttpOp[IN]) {
-    def setupRequestMultiArg[A1, A2](function2Arg: (A1, A2) => IN): HttpOp[(A1, A2)] =
+  implicit class HttpOpOps[IN, RESPONSE](httpOp: HttpOp[IN, RESPONSE]) {
+    def setupRequestMultiArg[A1, A2](function2Arg: (A1, A2) => IN): HttpOp[(A1, A2), RESPONSE] =
       httpOp.setupRequest(function2Arg.tupled)
   }
 
 }
+
```
pvighi: these changes look good, I still have to look at the definitions a little bit when looking at code that uses this but I guess I'll get used to this once I start to use it myself

lib/restHttp/src/main/scala/com/gu/util/resthttp/RestRequestMaker.scala
```diff
@@ -139,7 +141,11 @@ object RestRequestMaker extends Logging {
   def sendRequest(request: Request, getResponse: Request => Response): ClientFailableOp[String] = {
     logger.info(s"Attempting to do request")
     val response = getResponse(request)
-    httpIsSuccessful(response).map(_ => response.body.string)
+    filterIfSuccessful(response).map(_.body.string)
+  }
+
+  def filterIfSuccessful(response: Response): ClientFailableOp[Response] = {
```
pvighi: I find the naming here slightly confusing maybe this could be called toClientFailable. of course httpIsSuccessful could also be called to clientFailable.
So.. feel free to leave it as it is

lib/salesforce/src/main/scala/SalesforceAuthenticate.scala
```diff
@@ -45,12 +45,28 @@ object SalesforceAuthenticate extends Logging {
...
+    response: Request => Response,
+    sfAuth: SalesforceAuth
+  ): HttpOp[GetRequest, JsValue] =
+    HttpOp(response).setupRequest {
+      SalesforceRestRequestMaker.get(sfAuth)
+    }.flatMap {
+      filterIfSuccessful
+    }.map { response =>
+      Json.parse(response.body.string)
+    }
```
pvighi: these changes look good, I still have to look at the definitions a little bit when looking at code that uses this but I guess I'll get used to this once I start to use it myself

lib/restHttp/src/main/scala/com/gu/util/resthttp/HttpOp.scala
```diff
@@ -1,39 +1,56 @@
...
+  def runRequest(in: PARAM): ClientFailableOp[RESPONSE] =
     responseToOutput(effect(inputToRequest(in)))
 
   // this is effectively "contramap" so it is actually setting it up to run the given function before any previously setup functions
   // you can learn more about contravariant functors https://www.google.co.uk/search?q=contravariant+functor
-  def setupRequest[UPDATEDPARAM](function: UPDATEDPARAM => PARAM): HttpOp[UPDATEDPARAM] =
-    HttpOp(function.andThen(inputToRequest), effect, responseToOutput)
+  def setupRequest[UPDATEDPARAM](fromNewParam: UPDATEDPARAM => PARAM): HttpOp[UPDATEDPARAM, RESPONSE] =
+    HttpOp(fromNewParam.andThen(inputToRequest), effect, responseToOutput)
```
pvighi: these changes look good, I still have to look at the definitions a little bit when looking at code that uses this but I guess I'll get used to this once I start to use it myself

handlers/sf-contact-merge/src/main/scala/com/gu/sf_contact_merge/getsfcontacts/GetSfAddress.scala
```diff
@@ -0,0 +1,65 @@
...
+  )
+  implicit val wireResultReads = Json.reads[WireResult]
+
+  def apply(getOp: HttpOp[GetRequest, JsValue]): GetSfAddress =
+    new GetSfAddress(getOp.setupRequest(toRequest).parse[WireResult].map(toResponse).runRequest)
+
+  def toResponse(wireResult: WireResult): Option[SFAddress] = {
+    import wireResult._
+    for {
+      street <- OtherStreet.filter(_.length > 1) // could just be a comma or dot
```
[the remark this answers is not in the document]
pvighi: Oh you are right, I forgot there's no separate field for house number so even if someone lived in 34 X street this would have to be something like "34 X" at the bare minimum

handlers/sf-contact-merge/src/main/scala/com/gu/sf_contact_merge/getsfcontacts/GetSfAddressOverride.scala
pvighi: That could be better, I was trying to see if there was a way of finding the first matching value without having to explicitly return true or false but I could not come up with anything.
I have an irrational hatred for statements that resemble:
```scala
 if (something) then true else false
```
even though this is not exactly the case

lib/salesforce/src/test/scala/com/gu/salesforce/dev/SFEffectsData.scala
johnduffell: yes, I have just add a comment to the contact like we do in zuora.  I have made this more centralised in order to represent the relationship between the SF tests and external salesforce in one place. 
image

handlers/sf-contact-merge/src/main/scala/com/gu/sf_contact_merge/getsfcontacts/GetSfAddressOverride.scala
johnduffell: ok I have dealt with a related problem by turning two of the options into sealed traits with better names, I don't know if that completely addresses your comment though!

lib/restHttp/src/main/scala/com/gu/util/resthttp/RestRequestMaker.scala
johnduffell: the important fact is that this just checks the status code and fits into the flatMap pipeline after the basic HttpOp.  Perhaps HttpOp should have some kind of filter operation rather than using the over powerful flatMap to compose this. 
The whole code in this area of lower level rest and http stuff has long been a monolithic blob, and I hope the invention of HttpOp will allow a new land of nice separated composable functions (via flatMap and setupRequest).  However in the mean time some of the small functions are pulled out and some of the others are still mixed together.  I'm not sure of the best route to doing away with the Requests object completely.  Getting rid of it from SF would be relatively easy, but for zuora stuff it would be slower.
So I think basically you found a bigger problem and maybe we need better segregation between the old and new worlds and a strategy to move over.

johnduffell: PS after discussion in person I renamed the function as it is a sensible idea 💯

lib/restHttp/src/main/scala/com/gu/util/resthttp/HttpOp.scala
johnduffell: I can alwas rename `setupRequest` to `contramap` if it helps 😆

PR #196, opened 2018-09-18
ignore the case of email addresses when checking for discrepancies

johnduffell:
This is part of the work to merge salesforce contacts together.  We want to get all info about a single customer under a single contact and all connected to the same identity id.

I have noticed when doing the merge that a lot of the email failures, although a small percentage overall, are just due to case being incorrect.  Although the first part is technically case sensitive, I don't think anyone would legitimately use it that way.

This pr just ignores the case when validating that the merge looks sensible, before actually doing the merge.

@pvighi @paulbrown1982

### DIFF
--- a/handlers/sf-contact-merge/src/test/scala/manualTest/RunBatch.scala
+++ b/handlers/sf-contact-merge/src/test/scala/manualTest/RunBatch.scala
@@ -34,12 +34,12 @@ object RunBatch {
 
     val result = for {
       args <- GetArgs(rawArgs)
-      postString = HttpOp(Http.response).setupRequest(post(args.apiKey))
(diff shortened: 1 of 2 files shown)


handlers/sf-contact-merge/src/test/scala/manualTest/RunBatch.scala
```diff
@@ -34,12 +34,12 @@ object RunBatch {
 
     val result = for {
       args <- GetArgs(rawArgs)
-      postString = HttpOp(Http.response).setupRequest(post(args.apiKey))
+      postString = HttpOp(Http.response).setupRequest(post(args.apiKey)).flatMap(RestRequestMaker.toClientFailableOp)
```
johnduffell: This is just code that runs locally.  I did this , because the changes to HttpOp in the last pr to support (processing the get responses), no longer automatically detected errors.  Now that has to be flatMapped/composed directly.

PR #197, opened 2018-09-19
don't allow contact merge if a loser is a digital voucher subscriber

johnduffell:
We are merging contacts where the email address is the same, so that we can apply our retention policy properly and users can manage their subscription online and get emails related to their product.

We discovered that if we merge to a winning contact, where a losing contact had a digital voucher, that would mess up the digital voucher.

This will be a small proportion of the non linked users, however we really don't want to break them.  This means that it's safest to just treat them as a hard cohort and refuse to link them.

This PR adds validation based on the tick box that says they are or have been in the voucher trial.  If so, it will ensure that they are the winning contact.  If not, it will refuse to touch all the contacts.

I still have a few tests to add, I will check the coverage.  Also take a look at the Lazy thing I added.  This is to stop unnecessary calls to salesforce and has very wide applicability.  It is quite well coupled to ClientFailableOp but I didn't want to make it a monad with a transformer if I could help it.  Comments would be useful on how to make sure it is as clear as possible especially naming.

@paulbrown1982 @pvighi


handlers/sf-contact-merge/src/main/scala/com/gu/sf_contact_merge/Handler.scala
```diff
@@ -100,6 +107,18 @@ object DomainSteps {
     } yield ApiGatewayResponse.successfulExecution).apiResponse
 }
 
+object ValidateNoLosingDigitalVoucher {
+
+  def apply(losingContacts: List[LazyClientFailableOp[GetSfAddress.IsDigitalVoucherUser]]): ApiGatewayOp[Unit] =
+    losingContacts.toStream.map {
+      _.value.toApiGatewayOp("get SF address").flatMap {
+        case IsDigitalVoucherUser(true) => ReturnWithResponse(ApiGatewayResponse.notFound("failed validation due to a "))
```
johnduffell: TODO oops I need to finish typing that in and refernce the contact id if possible.

handlers/sf-contact-merge/src/main/scala/com/gu/sf_contact_merge/getsfcontacts/GetSfContacts.scala
```diff
@@ -0,0 +1,36 @@
...
+      val otherContacts = otherContactIds.map(getSfAddress.apply)
+      otherContacts
+    }
+    SFContactsForMerge(winnerContact, other)
+  }
+
+  case class SFContactsForMerge(
+    winner: LazyClientFailableOp[SFContact],
+    others: List[LazyClientFailableOp[SFContact]]
+  )
```
johnduffell: since this grouping of "winner" and "others" is so common, in SFContactId, LazyClientFailableOp and ClientFailableOp, I am not sure if it would be more sensible to either have 3 similar structures, or have one generic structure.  At the moment they are mostly floating around as two fields, except in the above case class.

handlers/sf-contact-merge/src/main/scala/com/gu/sf_contact_merge/Handler.scala
```diff
@@ -100,6 +107,18 @@ object DomainSteps {
     } yield ApiGatewayResponse.successfulExecution).apiResponse
 }
 
+object ValidateNoLosingDigitalVoucher {
```
johnduffell: this function needs a test

lib/restHttp/src/main/scala/com/gu/util/resthttp/HttpOp.scala
```diff
@@ -44,6 +47,21 @@ object HttpOp {
 
 }
 
+// this is used for non effectful/expensive operations to call them at the last minute, if at all.
+class LazyClientFailableOp[+VALUE](underlying: () => ClientFailableOp[VALUE]) {
```
johnduffell: I should add some specific test for this

handlers/sf-contact-merge/src/main/scala/com/gu/sf_contact_merge/GetFirstNameToUse.scala
```diff
@@ -0,0 +1,33 @@
...
+import com.gu.sf_contact_merge.getaccounts.GetZuoraContactDetails.FirstName
+import com.gu.sf_contact_merge.update.UpdateSalesforceIdentityId.IdentityId
+import com.gu.util.apigateway.ApiGatewayResponse
+import com.gu.util.reader.Types.{ApiGatewayOp, _}
+
+object GetFirstNameToUse {
+
+  case class NameForIdentityId(identityId: Option[IdentityId], firstName: Option[FirstName])
+  def firstNameForIdentityAccount(namesForIdentityIds: List[NameForIdentityId]): Option[FirstName] = {
+    namesForIdentityIds.find(_.identityId.isDefined).flatMap(_.firstName)
```
pvighi: Another way of doing this would be :
```scala
    namesForIdentityIds.collectFirst{ case NameForIdentityId(Some(_), Some(firstName)) => firstName } 
```

handlers/sf-contact-merge/src/main/scala/com/gu/sf_contact_merge/GetFirstNameToUse.scala
```diff
@@ -0,0 +1,35 @@
...
+import com.gu.util.apigateway.ApiGatewayResponse
+import com.gu.util.reader.Types.{ApiGatewayOp, _}
+
+object GetFirstNameToUse {
+
+  case class NameForIdentityId(identityId: Option[IdentityId], firstName: Option[FirstName])
+
+  def firstNameForIdentityAccount(namesForIdentityIds: List[NameForIdentityId]): Option[FirstName] =
+    namesForIdentityIds.collectFirst {
+      case NameForIdentityId(Some(_: IdentityId), Some(firstName: FirstName)) => firstName
```
pvighi: 🍾

lib/restHttp/src/main/scala/com/gu/util/resthttp/HttpOp.scala
```diff
@@ -44,6 +47,21 @@ object HttpOp {
...
+  lazy val value = underlying()
+
+  def flatMap[OUT](function: VALUE => LazyClientFailableOp[OUT]): LazyClientFailableOp[OUT] =
+    LazyClientFailableOp(underlying().flatMap(function(_).value))
+
+  def map[OUT](function: VALUE => OUT): LazyClientFailableOp[OUT] =
+    LazyClientFailableOp(underlying().map(function))
```
pvighi: We would have to make sure everyone understands that mapping a lazyClientFailableOp makes it execute because I'm not sure that is the expected behaviour.

lib/restHttp/src/main/scala/com/gu/util/resthttp/HttpOp.scala
pvighi: 👍 looks like a simple enough solution, after merging this we can go back and refactor the new product api to use it

handlers/sf-contact-merge/src/test/scala/com/gu/sf_contact_merge/getsfcontacts/GetSfAddressOverrideTest.scala
```diff
@@ -41,9 +43,9 @@ class GetSfAddressOverrideTest extends FlatSpec with Matchers {
 
     val invocationLog = new MockGetSfAddress()
```
pvighi: maybe rename this to mockgetSfAddress

handlers/sf-contact-merge/src/test/scala/com/gu/sf_contact_merge/getsfcontacts/GetSfContactsTest.scala
```diff
@@ -0,0 +1,87 @@
...
+      LazyClientFailableOp { () =>
+        invocationLog = invocationLog ++ List(s"get ${sfContactId.value}")
+        testAddress(sfContactId.value)
+      }
+
+  }
+
+  "getSfContacts" should "get the winner and the other correctly (lazy)" in {
+
+    val invocationLog = new MockGetSfAddress()
```
pvighi: here as well it would be clear if the mocks are not called logs

handlers/sf-contact-merge/src/test/scala/com/gu/sf_contact_merge/getsfcontacts/GetSfAddressOverrideTest.scala
johnduffell: I have completely rethought this test as it didn't really need to keep track of invocations except in one test, and that one it just needed to ensure that the last one wasn't unpacked.  So I turned it into a fail() call in that case only.

handlers/sf-contact-merge/src/test/scala/com/gu/sf_contact_merge/getsfcontacts/GetSfContactsTest.scala
johnduffell: I started that, and then eventually made a specific container object to hold winning and other data, and suddenly all the need for invocation logs went away and moved to the specific test for the lazy thing. 😁

PR #198, opened 2018-09-20
refactor most existing SF uses to use the HttpOp rather than Requests

johnduffell:
This is a small piece of tech debt.

In the past we had a monolithic object to do all http requests.  Later we invented an HttpOp object that lets you decompose all the parts of the http and test them separately.

Now we have decided it's actually rather nice, so we need to make progress in refactoring things to use it.  It seemed a sensible target to refactor all the SF stuff as there's relatively little of it, so I just did all the stuff in identity-backfill.  Only the SF cases lambda left to do, although since there are still 0% code coverage of that one (only effectsTests are there), I would like to have some coverage before doing a refactor.

@pvighi @paulbrown1982


handlers/cancellation-sf-cases/src/main/scala/com/gu/cancellation/sf_cases/Handler.scala
```diff
@@ -256,14 +256,14 @@ object Handler extends Logging {
...
+            sfAuth <- SalesforceAuthenticate.doAuth(response, config)
+          } yield SalesforceRestRequestMaker(sfAuth, response)
 
         val sfRequestsTest: LazySalesforceAuthenticatedReqMaker = () =>
           for {
             config <- loadTestSfConfig.toApiGatewayOp("load 'test' SF config")
-            sfRequests <- SalesforceAuthenticate(response, config)
-          } yield sfRequests
+            sfAuth <- SalesforceAuthenticate.doAuth(response, config)
+          } yield SalesforceRestRequestMaker(sfAuth, response)
```
johnduffell: this is just removing a function that does very little.  The `SalesforceRestRequestMaker` is the one that we need to get rid of in future.

handlers/identity-backfill/src/test/scala/com/gu/identityBackfill/salesforce/getContact/GetSFContactSyncCheckFieldsEffectsTest.scala
```diff
@@ -3,22 +3,28 @@ package com.gu.identityBackfill.salesforce.getContact
...
-      auth <- DevSFEffects(GetFromS3.fetchString, RawEffects.response)
-      result <- GetSFContactSyncCheckFields(auth)(SFEffectsData.testContactHasNamePhoneOtherAddress).toApiGatewayOp("failed")
+      sfConfig <- LoadConfigModule(Stage("DEV"), GetFromS3.fetchString)[SFAuthConfig].toApiGatewayOp("parse config")
+      sfAuth <- SalesforceAuthenticate.doAuth(RawEffects.response, sfConfig)
+      getOp = SalesforceAuthenticate.get(RawEffects.response, sfAuth)
+      result <- GetSFContactSyncCheckFields(getOp).apply(SFEffectsData.testContactHasNamePhoneOtherAddress).value.toApiGatewayOp("failed")
     } yield result
 
-    actual.toDisjunction should be(\/-(ContactSyncCheckFields(None, "One", "Day", None)))
+    actual.toDisjunction should be(\/-(ContactSyncCheckFields(None, "One", "Day", Some("Afghanistan"))))
```
johnduffell: this test was failing due to not matching the actual data, so I just fixed it up.

handlers/new-product-api/src/test/scala/com/gu/newproduct/api/addsubscription/zuora/GetAccountEffectsTest.scala
```diff
@@ -21,7 +21,7 @@ class GetAccountEffectsTest extends FlatSpec with Matchers {
     } yield res
     val expected = Account(
       identityId = Some(IdentityId("30000549")),
-      sfContactId = Some(SfContactId("sfContactId")),
+      sfContactId = Some(SfContactId("003g000001W5FCeAAN")),
```
johnduffell: this test was failing due to not having been run since the extra field was added, and it hadn't been checked against what was in zuora, so I fixed it up

handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/Handler.scala
```diff
@@ -98,11 +99,12 @@ object Handler {
     mappings.get(stage).toApiGatewayContinueProcessing(ApiGatewayResponse.internalServerError(s"missing standard record type for stage $stage"))
   }
 
-  def syncableSFToIdentity(sfRequests: ApiGatewayOp[Requests], stage: Stage)(sFContactId: SFContactId): ApiGatewayOp[Unit] =
+  def syncableSFToIdentity(sfRequests: ApiGatewayOp[HttpOp[GetRequest, JsValue]], stage: Stage)(sFContactId: SFContactId): ApiGatewayOp[Unit] =
     for {
       sfRequests <- sfRequests
       standardRecordType <- standardRecordTypeForStage(stage)
-      syncable <- SyncableSFToIdentity(standardRecordType)(sfRequests)(sFContactId)
+      fields <- GetSFContactSyncCheckFields(sfRequests).apply(sFContactId).value.toApiGatewayOp("zuora issue")
```
pvighi: this is very minor and I've seen it in other places but the param to toApiGatewayOp is supposed to be a description of the action:
https://github.com/guardian/zuora-auto-cancel/blob/master/lib/handler/src/main/scala/com/gu/util/reader/Types.scala#L162

in this case the error would be something like "error attempting to zuora issue" or something like that

johnduffell: the funny thing is, this is actually a call to salesforce...
I think these "action" are a bit like comments, not really kept up to date.  I wonder if we can make the compiler work them out somehow.

johnduffell: I think we could add a trace/action field to ClientFailableOp, HttpOp and ApiGatewayOp and then automatically pass them through.  Then at least we can get the basics right and the action would just add to the trace.  This would have to be a separate PR as it would be self contained.

pvighi: if you look at the definition of the toApiGatewayOp that I linked in my original comment the error side of the response is actually logged, the "action" param is only used to generate the internal server error response. 
The funny thing is that if you follow the internalServerError this param is not returned to the user ever so we could probably at least make it optional.

pvighi: for this PR I would at least make it no log zuora issue if there is a sf issue

johnduffell: > if you look at the definition of the toApiGatewayOp that I linked in my original comment the error side of the response is actually logged, the "action" param is only used to generate the internal server error response.

look at the line directly above that , it is logged there.  So we would actually get it logged once on the line above when it has the exception, and once with the action only when it constructs the internal server error!  Not ideal I agree.

PR #199, opened 2018-09-25
convert cancellation lambda to use httpop for SF operations

johnduffell:
As per the previous work
https://github.com/guardian/zuora-auto-cancel/pull/198
https://github.com/guardian/zuora-auto-cancel/pull/174
We have found a nice way to split out the preprocessing and postprocessing of HTTP operations in order to make the tests simpler and get rid of the monolithic http Requests object.  This Requests object is used by the zuora, sf and identity operations, however we don't call salesforce that much so it's reasonable to convert everything and delete the old way.

This PR is to convert most of the cancellation-sf-cases lambda to use this HttpOp.  There is still work to add a post operation so we can convert the Create salesforce case operation, but that can be a further PR.

This all passes the effectstest but there are no tests that run without credentials yet.

@pvighi @paulbrown1982 @twrichards have a look and see what you think!

### DIFF
--- a/handlers/cancellation-sf-cases/src/main/scala/com/gu/cancellation/sf_cases/Handler.scala
+++ b/handlers/cancellation-sf-cases/src/main/scala/com/gu/cancellation/sf_cases/Handler.scala
@@ -24,19 +24,24 @@ import com.gu.util.config.LoadConfigModule.StringFromS3
 import com.gu.util.config._
 import com.gu.util.reader.Types._
 import com.gu.util.resthttp.Types.ClientFailableOp
-import com.gu.util.resthttp.{RestRequestMaker, Types}
+import com.gu.util.resthttp.{HttpOp, RestRequestMaker, Types}
 import okhttp3.{Request, Response}
 import play.api.libs.json._
 
 object Handler extends Logging {
 
-  case class IdentityAndSfRequests(identityUser: IdentityUser, sfRequests: RestRequestMaker.Requests)
+  case class IdentityAndSfRequests(identityUser: IdentityUser, sfRequests: SfRequestMethods)
+  case class SfRequestMethods(
+    deprecatedTODO: RestRequestMaker.Requests,
+    patch: HttpOp[RestRequestMaker.PatchRequest, Unit],
+    get: HttpOp[RestRequestMaker.GetRequest, JsValue]
+  )
 
   type HeadersOption = Option[Map[String, String]]
   type IdentityAndSfRequestsApiGatewayOp = ApiGatewayOp[IdentityAndSfRequests]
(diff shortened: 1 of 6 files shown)


handlers/cancellation-sf-cases/src/main/scala/com/gu/cancellation/sf_cases/Handler.scala
```diff
@@ -24,19 +24,24 @@ import com.gu.util.config.LoadConfigModule.StringFromS3
...
+import com.gu.util.resthttp.{HttpOp, RestRequestMaker, Types}
 import okhttp3.{Request, Response}
 import play.api.libs.json._
 
 object Handler extends Logging {
 
-  case class IdentityAndSfRequests(identityUser: IdentityUser, sfRequests: RestRequestMaker.Requests)
+  case class IdentityAndSfRequests(identityUser: IdentityUser, sfRequests: SfRequestMethods)
+  case class SfRequestMethods(
+    deprecatedTODO: RestRequestMaker.Requests,
```
johnduffell: this is only used for the final remaining POST and I will just pop that in another PR straight after this

lib/restHttp/src/main/scala/com/gu/util/resthttp/HttpOp.scala
```diff
@@ -43,6 +43,13 @@ object HttpOp {
   implicit class HttpOpOps[IN, RESPONSE](httpOp: HttpOp[IN, RESPONSE]) {
     def setupRequestMultiArg[A1, A2](function2Arg: (A1, A2) => IN): HttpOp[(A1, A2), RESPONSE] =
       httpOp.setupRequest(function2Arg.tupled)
+    def setupRequestMultiArg[A1, A2, A3](function2Arg: (A1, A2, A3) => IN): HttpOp[(A1, A2, A3), RESPONSE] =
+      httpOp.setupRequest(function2Arg.tupled)
+  }
+
+  // convenience, untuples for you
+  implicit class HttpOpTuple3Ops[A1, A2, A3, RESPONSE](httpOp3Arg: HttpOp[(A1, A2, A3), RESPONSE]) {
+    def runRequestMultiArg: (A1, A2, A3) => ClientFailableOp[RESPONSE] = Function.untupled(httpOp3Arg.runRequest)
```
johnduffell: we didn't need a 3 arg version before, now that we do, I just added one.  I think scala 3 would get rid of this problem in the end, but if it's good enough for tuples it's good enough for httpop ;)

twrichards:
FYI - I deployed this branch to the DEV instance of the lambda then ran manage-frontend locally and went through a sample cancellation flow end-to-end... all worked nicely 👍

PR #201, opened 2018-09-25
use HttpOp for salesforce POST and remove old style Requests

johnduffell:
This PR finishes converting all the SF operations to the new HttpOp.  It also tidies up everything to make sense in terms of small functions.

I have noticed that it has added a bit to the wiring, this is inevitable because you no longer get all the operations in one big object.  On the other hand, you no longer have to decide in the operation itself which of the verbs to use.

There are some places, e.g. in cancellation sf cases lambda and the sf contact merge where the Lazy, we could wait longer before calling .value on it.  I think that Lazy should probably not be dependent on ClientFailableOp so that we can flatmap it all the way up to the ApiGatewayHandler.  But that would make the code more abstract, so it may be worth being less flexible for the time being.

@paulbrown1982 @pvighi @twrichards @mario-galic  comments/+1s please?


handlers/cancellation-sf-cases/src/main/scala/com/gu/cancellation/sf_cases/Handler.scala
```diff
@@ -261,22 +260,14 @@ object Handler extends Logging {
         val sfRequestsNormal: LazySalesforceAuthenticatedReqMaker = () =>
           for {
             config <- loadNormalSfConfig.toApiGatewayOp("load 'normal' SF config")
-            sfAuth <- SalesforceAuthenticate.doAuth(response, config)
-          } yield SfRequestMethods(
-            SalesforceRestRequestMaker(sfAuth, response),
-            SalesforceAuthenticate.patch(response, sfAuth),
-            SalesforceAuthenticate.get(response, sfAuth)
-          )
+            sfAuth <- SalesforceClient(response, config).value.toApiGatewayOp("Failed to authenticate with Salesforce")
```
johnduffell: when you call .value, it then makes the call and stores the result.  calling .value a second time uses the same value again.  This is much neater than having to pass around anonymous functions.

lib/restHttp/src/main/scala/com/gu/util/resthttp/RestRequestMaker.scala
```diff
@@ -141,11 +146,11 @@ object RestRequestMaker extends Logging {
...
     logger.info(s"Attempting to do request")
     val response = getResponse(request)
-    toClientFailableOp(response).map(_.body.string)
+    toClientFailableOp(response).map(_.value)
   }
 
-  def toClientFailableOp(response: Response): ClientFailableOp[Response] = {
-    httpIsSuccessful(response).map(_ => response)
+  def toClientFailableOp(response: Response): ClientFailableOp[BodyAsString] = {
+    httpIsSuccessful(response).map(_ => response).map(_.body.string).map(BodyAsString.apply)
```
johnduffell: this reduces the exposure of OkHttp slightly more than before.  I'm not keen on it being used by too much code.

lib/salesforce/src/main/scala/com/gu/salesforce/JsonHttp.scala
```diff
@@ -0,0 +1,45 @@
...
+
+import com.gu.salesforce.SalesforceClient.{GetMethod, StringHttpRequest, PatchMethod, PostMethod}
+import com.gu.util.resthttp.HttpOp.HttpWrapper
+import com.gu.util.resthttp.RestRequestMaker._
+import com.gu.util.resthttp.Types.{ClientFailableOp, ClientSuccess, GenericError}
+import play.api.libs.json.{JsValue, Json}
+
+import scala.util.Try
+
+object JsonHttp {
```
johnduffell: This represents the difference between a String based http request-response and a Json based one.  It will parse/serialise between strings at the lower level and json at the next level up.  Most of our rest calls would use this.

lib/salesforce/src/main/scala/com/gu/salesforce/SalesforceAuthenticate.scala
```diff
@@ -0,0 +1,60 @@
...
+import com.gu.util.Logging
+import com.gu.util.config.ConfigLocation
+import com.gu.util.reader.Types._
+import com.gu.util.resthttp.RestOp._
+import com.gu.util.resthttp.RestRequestMaker.toClientFailableOp
+import com.gu.util.resthttp.{HttpOp, LazyClientFailableOp}
+import okhttp3.{FormBody, Request, Response}
+import play.api.libs.json.{Json, Reads}
+
+object SalesforceAuthenticate extends Logging {
```
johnduffell: this is just an http op like any other that calls the SF authenticate endpoint and returns a (lazy) result.

lib/salesforce/src/main/scala/com/gu/salesforce/SalesforceClient.scala
```diff
@@ -0,0 +1,45 @@
+package com.gu.salesforce
+
+import com.gu.salesforce
+import com.gu.salesforce.SalesforceAuthenticate.{SFAuthConfig, SalesforceAuth}
+import com.gu.util.resthttp.RestRequestMaker.{BodyAsString, RelativePath, createBodyFromString, toClientFailableOp}
+import com.gu.util.resthttp.{HttpOp, LazyClientFailableOp}
+import okhttp3.{Request, Response}
+
+object SalesforceClient {
```
johnduffell: this is a kind of wiring steps class that will auth with SF and return a low level string based error checking http client.

johnduffell:
@mario-galic that's the quietest review I've ever had 😆 
Patricio is still looking as well and may have comments so I'll hold off for now

lib/salesforce/src/main/scala/com/gu/salesforce/SalesforceClient.scala
```diff
@@ -0,0 +1,45 @@
...
+
+  def apply(
+    response: Request => Response,
+    config: SFAuthConfig
+  ): LazyClientFailableOp[HttpOp[StringHttpRequest, BodyAsString]] =
+    salesforce.SalesforceAuthenticate(response)(config).map { sfAuth =>
```
pvighi: any reason in particular to specify the package name here?

lib/salesforce/src/main/scala/com/gu/salesforce/SalesforceClient.scala
```diff
@@ -0,0 +1,45 @@
...
+    override def builder: Request.Builder = new Request.Builder().post(createBodyFromString(body))
+  }
+  case class PatchMethod(body: BodyAsString) extends RequestMethod {
+    override def builder: Request.Builder = new Request.Builder().patch(createBodyFromString(body))
+  }
+  case object GetMethod extends RequestMethod {
+    override def builder: Request.Builder = new Request.Builder().get()
+  }
+
+  case class StringHttpRequest(relativePath: RelativePath, requestMethod: RequestMethod)
```
pvighi: Why is this called StringHttpRequest ?

lib/salesforce/src/main/scala/com/gu/salesforce/SalesforceClient.scala
johnduffell: must be intellij! I'll fix

lib/salesforce/src/main/scala/com/gu/salesforce/SalesforceClient.scala
johnduffell: it's an http request that uses a string as the body and is the lowest level that wraps the basic okhttp.Request (as opposed to a `JsValue` or a `ZuoraQuery` or a `List[FirstName]` which it might as we go further up the levels of abstraction)  Not sure if the naming is good though, as it should ideally be clear which level each thing sits at.

PR #202, opened 2018-09-27
Skip contact merge process if already done/no work to do

johnduffell:
Since we have run most of the scenarios already, we have a small number to redo.  Unfortunately it does a lot of calls and some preparation and tidyup of the SF contact, even if all the merging has already completed.

This PR adds a pre check that at least one of the zuora accounts will have its link changed.  If not it will skip it with a 200 and the body saying that there's no processing needed.

@pvighi @paulbrown1982 take a look

### DIFF
--- a/handlers/sf-contact-merge/src/test/scala/manualTest/RunBatch.scala
+++ b/handlers/sf-contact-merge/src/test/scala/manualTest/RunBatch.scala
@@ -39,7 +39,8 @@ object RunBatch {
       postJsonString = postString.setupRequest[JsonString](jsonString => BodyAsString(jsonString.value))
       requestWithResultAsTry = (jsonString: JsonString) =>
         postJsonString.runRequest(jsonString) match {
-          case ClientSuccess(_) => \/-(())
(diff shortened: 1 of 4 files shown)


handlers/sf-contact-merge/src/test/scala/manualTest/RunBatch.scala
```diff
@@ -39,7 +39,8 @@ object RunBatch {
       postJsonString = postString.setupRequest[JsonString](jsonString => BodyAsString(jsonString.value))
       requestWithResultAsTry = (jsonString: JsonString) =>
         postJsonString.runRequest(jsonString) match {
-          case ClientSuccess(_) => \/-(())
+          case ClientSuccess(body) if body.value.contains("Success") => \/-(())
+          case ClientSuccess(body) => \/-(body.value)
```
johnduffell: this is just so the batch runner will display in the summary how many times it has not had work to do, without cluttering up the main progress list with non-failures.  If I returned "nothing to do" as a 404, then it would draw attention away from the actual problems.

PR #204, opened 2018-09-27
add update email address facility to zuora

johnduffell:
This is also part of the merging +gnm email address support.  When we merge a load of accounts, some of which are +gnm, we need the ability to replace the email addresses in zuora.

This PR adds an optional email field to the zuora update which will cause the work email in the billing contact to be corrected.  This will only happen for +gnm addresses.

@paulbrown1982 @pvighi comment please

### DIFF
--- a/handlers/sf-contact-merge/src/main/scala/com/gu/sf_contact_merge/update/UpdateAccountSFLinks.scala
+++ b/handlers/sf-contact-merge/src/main/scala/com/gu/sf_contact_merge/update/UpdateAccountSFLinks.scala
@@ -6,7 +6,7 @@ import com.gu.sf_contact_merge.getaccounts.GetZuoraContactDetails.EmailAddress
 import com.gu.sf_contact_merge.update.UpdateSalesforceIdentityId.IdentityId
 import com.gu.util.resthttp.RestRequestMaker.{JsonResponse, PutRequest, RelativePath}
 import com.gu.util.resthttp.Types.ClientFailableOp
-import play.api.libs.json.{JsSuccess, Json, Reads}
+import play.api.libs.json.Json
 
 object UpdateAccountSFLinks {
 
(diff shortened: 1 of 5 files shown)


handlers/sf-contact-merge/src/main/scala/com/gu/sf_contact_merge/update/UpdateAccountSFLinks.scala
```diff
@@ -2,25 +2,32 @@ package com.gu.sf_contact_merge.update
...
+  case class WireZuoraAccount(
     crmId: String,
     sfContactId__c: String,
-    IdentityId__c: Option[String]
+    IdentityId__c: Option[String],
+    billToContact: Option[WireZuoraContact]
   )
-  implicit val writes = Json.writes[Request]
+  case class WireZuoraContact(
+    workEmail: String
```
johnduffell: I have realised that this turns on invoiceDeliveryPrefsEmail to true.  We need to make sure it's false if we are updating the email address. see https://www.zuora.com/developer/api-reference/#operation/PUT_Account

johnduffell: this is not an issue as there are no email configured.  it would only be a problem if we were clearing the email altogether, as zuora would validate it. https://github.com/guardian/membership-common/pull/528 has information about that situation.

handlers/sf-contact-merge/src/test/scala/com/gu/sf_contact_merge/update/identityid/GetIdentityIdForAccount.scala
```diff
@@ -15,15 +15,21 @@ object GetIdentityIdForAccount {
...
+    implicit val zaReadsZContact = Json.reads[ZContact]
     implicit val zaReadsBasicInfo = Json.reads[BasicInfo]
     implicit val zaReadsZuoraAccount = Json.reads[ZuoraAccount]
 
   }
 
-  def apply(requests: Requests)(accountId: AccountId): ClientFailableOp[BasicInfo] =
-    requests.get[ZuoraAccount](s"/accounts/${accountId.value}").map(_.basicInfo)
+  def apply(requests: Requests)(accountId: AccountId): ClientFailableOp[ZuoraAccount] =
+    requests.get[ZuoraAccount](s"/accounts/${accountId.value}")
```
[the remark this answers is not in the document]
johnduffell: good point, I've fixed locally and I'll push with that PR.

PR #205, opened 2018-09-28
make merge lambda happy to merge +gnm addresses

johnduffell:
As per my last couple of PRs this is the meat of the logic changes.  It needed structural changes in the flow of information so there are a lot of changes here.

+gnm addresses are where the call centre were trying to sign someone up who already had an email address in identity.  Therefore they signed them up with +gnm or +gnm1 before the @ in the email address.
These will be grouped by the SF duplicate contact generation and fed in the normal way to the lambda.  A +gnm could be the winning contact.  The lambda will have to deal with duplicates where some of the zuora accounts will have +gnm in the email address.

@pvighi @mario-galic @paulbrown1982 please take a look


handlers/sf-contact-merge/src/main/scala/com/gu/sf_contact_merge/Handler.scala
```diff
@@ -73,39 +81,108 @@ object Handler {
...
+        )
+        mergeRequest.zuoraAccountIds.traverseU(updateAccountSFLinks(linksFromZuora, _))
+      }.toApiGatewayOp("update accounts with winning details")
+      _ <- updateSFContacts(
+        mergeRequest.winningSFContact,
+        sfData.sfIdentityIdMoveData,
+        zuoraData.firstNameToUse,
+        sfData.maybeSFAddressOverride,
+        sfData.emailOverrides.needToOverwriteWith
+      ).toApiGatewayOp("update sf contact(s) to force a sync")
```
johnduffell: this high level structure is reasonable now (read from zuora, read from sf, update zuora, update sf) but the salesforce and zuora steps below are not tested and some of the conversions desparately need making into functions and adding tests.

handlers/sf-contact-merge/src/main/scala/com/gu/sf_contact_merge/getaccounts/GetZuoraContactDetails.scala
```diff
@@ -12,6 +12,9 @@ object GetZuoraContactDetails {
   case class ContactId(value: String) extends AnyVal
 
   case class EmailAddress(value: String) extends AnyVal
+  object EmailAddress {
+    def apply(value: String) = new EmailAddress(value.toLowerCase)
```
johnduffell: this is done at this stage because as soon as we read it in, we fundamentally don't care about the case, so we normalise at the edge.  However the +gnm we do care about in our code so we don't touch that until the business logic.

handlers/sf-contact-merge/src/main/scala/com/gu/sf_contact_merge/update/UpdateSFContacts.scala
```diff
@@ -29,25 +30,27 @@ object UpdateSFContacts {
...
           maybeOverwriteEmailAddress
         )
-        setOrClearIdentityId.apply(sfContactId, sFContactUpdate) // this causes the sync to identity and zuora
+        setOrClearIdentityId.apply(sfContactId.id, sFContactUpdate) // this causes the sync to identity and zuora
       }
     } yield ()
 
   case class OldSFContact(sfContactId: SFContactId)
 
+  case class IdentityIdToUse(value: IdentityId)
```
johnduffell: with so many strings around, wrapping them in value classes to get compiler help (e.g. a firstname and an accountid) is working really well.
However the same problem occurred at the next level- with so many SFContactIds around (.e.g oldsfcontact and winningsfContact), wrapping that again gets further help from the compiler and helps with writing the test.
The other advantage over named arguments is that the proof carries all the way through until you unwrap it, so you get a full chain of guarantee that you're making sense.

handlers/sf-contact-merge/src/test/scala/com/gu/sf_contact_merge/getaccounts/GetZuoraContactDetailsTest.scala
```diff
@@ -37,7 +37,7 @@ object GetZuoraContactDetailsTest {
     """{
       |    "records": [
       |        {
-      |            "WorkEmail": "peppa.pig@guardian.co.uk",
+      |            "WorkEmail": "peppa.pig@GUARDIAN.co.uk",
```
johnduffell: just to check that it really does a lowercase on it

handlers/sf-contact-merge/src/test/scala/com/gu/sf_contact_merge/EndToEndTest.scala
```diff
@@ -140,8 +140,9 @@ object EndToEndTest {
       |    "OtherCountry": "country1",
       |    "OtherAddress": null,
       |    "Phone": null,
-      |    "IdentityID__c": "12121",
-      |    "Digital_Voucher_User__c": true
+      |    "IdentityID__c": null,
+      |    "Digital_Voucher_User__c": true,
+      |    "Email": "hello+gnm1@email.com"
```
johnduffell: this end to end test is now testing the merging +gnm case, which is actually not very common.  Since this test is expensive to maintain, I don't want to add one for each case, but I'm not sure if I should stick with the most common case (non +gnm) still?

handlers/sf-contact-merge/src/test/scala/com/gu/sf_contact_merge/update/UpdateSFContactsTest.scala
```diff
@@ -33,42 +34,18 @@ class UpdateSFContactsTest extends FlatSpec with Matchers {
 
   }
 
-  it should "not edit identity id at all if we're not setting one" in {
```
johnduffell: this test was removed because it turned out that the only purpose of the first call was to clear the identity id, but there's no point in having a specific test that we update the contact that doesn't have an identity id anyway in order to remove it.
Now the code has been changed to only run that call in the first place if there was an old identity id.

handlers/sf-contact-merge/src/main/scala/com/gu/sf_contact_merge/DomainSteps.scala
```diff
@@ -0,0 +1,155 @@
...
+        )
+        mergeRequest.zuoraAccountIds.traverseU(updateAccountSFLinks(linksFromZuora, _))
+      }.toApiGatewayOp("update accounts with winning details")
+      _ <- updateSFContacts(
+        mergeRequest.winningSFContact,
+        sfData.sfIdentityIdMoveData,
+        zuoraData.firstNameToUse,
+        sfData.maybeSFAddressOverride,
+        sfData.maybeEmailOverride
+      ).toApiGatewayOp("update sf contact(s) to force a sync")
```
pvighi: Just a reminder of our chat, if nothing is changed no sync will be triggered

handlers/sf-contact-merge/src/test/scala/com/gu/sf_contact_merge/EndToEndTest.scala
pvighi: if you think this is the most complete case and we will only have one I would test the +gnm case

handlers/sf-contact-merge/src/main/scala/com/gu/sf_contact_merge/update/UpdateSFContacts.scala
pvighi: If it gets to complicated we could also have an alternative constructor directly from a string if we are converting something from a wire object or something like that.

handlers/sf-contact-merge/src/test/scala/com/gu/sf_contact_merge/EndToEndTest.scala
johnduffell: I don't think it's most complete, I think it's just different.  But I'm happy to leave it as it is.

handlers/sf-contact-merge/src/main/scala/com/gu/sf_contact_merge/update/UpdateSFContacts.scala
johnduffell: in the code itself I'm not sure we'd want to go through two levels at the same time as then the function would probably be doing too much. At the wire level we identify the named field as a string, at the next level we identify it as an Identity id, and then we identify it as the one to use.
I see your point if it were a test though, it would definitely tidy them up a bit in that case where we just want a literal.

PR #207, opened 2018-10-03
new product api: make effective date today when creating subs

pvighi:
Previously the effective date was set to the selected start date, matt mentioned that it should always be set to today.
After this PR a voucher sub created on 03/10/2018 with a selected start date of 19/11/2018 looks like this in zuora :
```
Contract effective date:03/10/2018 
Service activation date:03/10/2018 
Customer acceptance date:19/11/2018 
```
@paulbrown1982  do you agree with this ? I don't remember how we decided on what to use for the dates originally but we did spend some time deciding.

### DIFF
--- a/handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/Handler.scala
+++ b/handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/Handler.scala
@@ -1,7 +1,7 @@
 package com.gu.newproduct.api.addsubscription
 
 import java.io.{InputStream, OutputStream}
-import java.time.LocalDate
+import java.time.{LocalDate, LocalDateTime}
 
 import com.amazonaws.services.lambda.runtime.Context
 import com.gu.effects.sqs.AwsSQSSend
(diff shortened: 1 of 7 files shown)


handlers/new-product-api/src/test/scala/manualTest/AddSubscriptionManualTest.scala
```diff
@@ -9,8 +9,8 @@ import play.api.libs.json.{JsString, Json}
 object AddSubscriptionManualTest extends App {
   val requestBody =
     """{
-      |   "zuoraAccountId":"2c92c0f865244687016538e563b85fac",
-      |   "startDate":"2018-10-15",
+      |   "zuoraAccountId":"2c92c0f9661ff980016633b4200228af",
+      |   "startDate":"2018-11-12",
```
pvighi: i just pushed this by mistake but it doesn't matter, this is just a way of running the lambda call manually

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/Handler.scala
```diff
@@ -61,14 +61,13 @@ object Steps {
...
-    request.acquisitionCase,
-    request.acquisitionSource,
-    request.createdByCSR
+    productRatePlanId = productRatePlanId,
+    accountId = request.zuoraAccountId,
+    maybeChargeOverride = chargeOverride,
+    acceptanceDate = acceptanceDate,
+    acquisitionCase = request.acquisitionCase,
+    acquisitionSource = request.acquisitionSource,
+    createdByCSR = request.createdByCSR
```
pvighi: the real change is kinda hidden here because I added names to the params. But start date is not passed in anymore only acceptance date.  Maybe we should rename acceptance date to start date now that it is the only date in the request

PR #208, opened 2018-10-17
contact merge - Clear identity id from zuora if the winner doesn't have identity

johnduffell:
This is a bug found by Paul and Abhishek.  Previously when we were merging accounts, if any of the accounts had an identity id we would use that one for all of the zuora accounts, so there was never a need to clear an identity id from zuora.  If it was supposed to end up clear, it would have been clear in the first place.

Now that we are merging +gnm accounts, it is possible there is no identity id at all on the original non +gnm. This means that once we merged we should clean them all from zuora.

The bug was caused by the classic issue of "what does `None` mean?"  As such this PR fixes the bug and also uses a proper type with nice names.

@paulbrown1982 @pvighi


handlers/sf-contact-merge/src/test/scala/com/gu/sf_contact_merge/update/identityid/UpdateAccountSFLinksEffectsTest.scala
```diff
@@ -34,7 +33,7 @@ class UpdateAccountSFLinksEffectsTest extends FlatSpec with Matchers {
       updateAccount = update(ZuoraFieldUpdates(
         WinningSFContact(SFContactId(s"cont$unique")),
         CRMAccountId(s"acc$unique"),
-        Some(IdentityIdToUse(IdentityId(s"ident$unique"))),
+        ReplaceZuoraIdentityId(IdentityId(s"ident$unique")),
```
johnduffell: I updated this effects test to be the same and it passes.

I also temporarily changed it to use `ClearIdentityId` and make sure it worked right.  I decided not to leave that in as a separate effects test (was that the right decision?)

PR #210, opened 2018-10-30
when linking existing subs, create guest accounts if no identity

johnduffell:
So far we have always looked for an existing identity account when trying to find an identity account for an existing zuora/SF account.

This has made it hard to meet our targets for linking, as most people who need linking don't actually have an account.

This PR will make it create a guest (ghost) account for anyone with no existing identity account.  This means that as soon as they take control of their account (by setting the password) they will have access to manage it online.

There is a possible future PR possible to add their name to the identity account i.e. (privateFields-> (firstName, secondName)) [subs frontend reference](https://github.com/guardian/subscriptions-frontend/blob/e057a5b8dfd8ec94c48ebae4455f0e1fcb6aed5c/app/services/IdentityService.scala#L128) but this isn't done at present.  Anyway, it would be added by any identity sync from SF in future.

@paulbrown1982 see what you think
Also @pvighi @jacobwinch @twrichards might have suggestions especially naming/clarity

### DIFF
--- a/handlers/cancellation-sf-cases/src/main/scala/com/gu/cancellation/sf_cases/Handler.scala
+++ b/handlers/cancellation-sf-cases/src/main/scala/com/gu/cancellation/sf_cases/Handler.scala
@@ -160,11 +160,11 @@ object Handler extends Logging {
       (for {
         identityAndSfRequests <- sfBackendForIdentityCookieHeader(apiGatewayRequest.headers)
         raiseCaseDetail <- apiGatewayRequest.bodyAsCaseClass[RaiseCaseDetail]()
-        lookupByIdOp = SalesforceGenericIdLookup(identityAndSfRequests.sfClient.wrap(JsonHttp.get))
-        mostRecentCaseOp = SalesforceCase.GetMostRecentCaseByContactId(identityAndSfRequests.sfClient.wrap(JsonHttp.get))
-        createCaseOp = SalesforceCase.Create(identityAndSfRequests.sfClient.wrap(JsonHttp.post))
+        lookupByIdOp = SalesforceGenericIdLookup(identityAndSfRequests.sfClient.wrapWith(JsonHttp.get))
+        mostRecentCaseOp = SalesforceCase.GetMostRecentCaseByContactId(identityAndSfRequests.sfClient.wrapWith(JsonHttp.get))
+        createCaseOp = SalesforceCase.Create(identityAndSfRequests.sfClient.wrapWith(JsonHttp.post))
         wiredCreateCaseOp = buildWireNewCaseForSalesforce.tupled andThen createCaseOp
-        sfUpdateOp = SalesforceCase.Update(identityAndSfRequests.sfClient.wrap(JsonHttp.patch))
+        sfUpdateOp = SalesforceCase.Update(identityAndSfRequests.sfClient.wrapWith(JsonHttp.patch))
         updateReasonOnRecentCaseOp = updateCaseReason(sfUpdateOp)_
         newOrResumeCaseOp = newOrResumeCase(wiredCreateCaseOp, updateReasonOnRecentCaseOp, raiseCaseDetail)_
         wiredRaiseCase = raiseCase(lookupByIdOp, mostRecentCaseOp, newOrResumeCaseOp)_
@@ -214,11 +214,11 @@ object Handler extends Logging {
       (for {
         identityAndSfRequests <- sfBackendForIdentityCookieHeader(apiGatewayRequest.headers)
         pathParams <- apiGatewayRequest.pathParamsAsCaseClass[CasePathParams]()
-        lookupByIdOp = SalesforceGenericIdLookup(identityAndSfRequests.sfClient.wrap(JsonHttp.get))
-        getCaseByIdOp = SalesforceCase.GetById[CaseWithContactId](identityAndSfRequests.sfClient.wrap(JsonHttp.get))_
+        lookupByIdOp = SalesforceGenericIdLookup(identityAndSfRequests.sfClient.wrapWith(JsonHttp.get))
+        getCaseByIdOp = SalesforceCase.GetById[CaseWithContactId](identityAndSfRequests.sfClient.wrapWith(JsonHttp.get))_
         _ <- verifyCaseBelongsToUser(lookupByIdOp, getCaseByIdOp)(identityAndSfRequests.identityUser.id, pathParams.caseId)
         requestBody <- apiGatewayRequest.bodyAsCaseClass[JsValue]()
-        sfUpdateOp = SalesforceCase.Update(identityAndSfRequests.sfClient.wrap(JsonHttp.patch))
+        sfUpdateOp = SalesforceCase.Update(identityAndSfRequests.sfClient.wrapWith(JsonHttp.patch))
         _ <- sfUpdateOp(pathParams.caseId, requestBody).toApiGatewayOp("update case")
       } yield ApiGatewayResponse.successfulExecution).apiResponse
 
--- a/handlers/cancellation-sf-cases/src/main/scala/com/gu/cancellation/sf_cases/Handler.scala
+++ b/handlers/cancellation-sf-cases/src/main/scala/com/gu/cancellation/sf_cases/Handler.scala
@@ -9,13 +9,12 @@ import com.gu.effects.{GetFromS3, RawEffects}
 import com.gu.identity.IdentityCookieToIdentityUser.{CookieValuesToIdentityUser, IdentityId, IdentityUser}
 import com.gu.identity.{IdentityCookieToIdentityUser, IdentityTestUserConfig, IsIdentityTestUser}
 import com.gu.salesforce.SalesforceAuthenticate.{SFAuthConfig, SFAuthTestConfig}
-import com.gu.salesforce.SalesforceClient.StringHttpRequest
 import com.gu.salesforce.SalesforceGenericIdLookup.{FieldName, LookupValue, SfObjectType, TSalesforceGenericIdLookup}
 import com.gu.salesforce.cases.SalesforceCase
 import com.gu.salesforce.cases.SalesforceCase.Create.WireNewCase
 import com.gu.salesforce.cases.SalesforceCase.GetMostRecentCaseByContactId.TGetMostRecentCaseByContactId
 import com.gu.salesforce.cases.SalesforceCase.{CaseId, CaseSubject, CaseWithId, ContactId, SubscriptionId}
-import com.gu.salesforce.{JsonHttp, SalesforceClient, SalesforceGenericIdLookup}
+import com.gu.salesforce.{SalesforceClient, SalesforceGenericIdLookup}
 import com.gu.util.Logging
 import com.gu.util.apigateway.ApiGatewayHandler.{LambdaIO, Operation}
 import com.gu.util.apigateway.ResponseModels.ApiResponse
@@ -23,9 +22,10 @@ import com.gu.util.apigateway.{ApiGatewayHandler, ApiGatewayRequest, ApiGatewayR
 import com.gu.util.config.LoadConfigModule.StringFromS3
 import com.gu.util.config._
 import com.gu.util.reader.Types._
+import com.gu.util.resthttp.JsonHttp.StringHttpRequest
 import com.gu.util.resthttp.RestRequestMaker.BodyAsString
 import com.gu.util.resthttp.Types.ClientFailableOp
-import com.gu.util.resthttp.{HttpOp, Types}
+import com.gu.util.resthttp.{HttpOp, JsonHttp, Types}
 import okhttp3.{Request, Response}
 import play.api.libs.json._
 
--- a/handlers/identity-backfill/src/main/scala/com/gu/identity/CreateGuestAccount.scala
+++ b/handlers/identity-backfill/src/main/scala/com/gu/identity/CreateGuestAccount.scala
@@ -2,25 +2,17 @@ package com.gu.identity
 
 import com.gu.identityBackfill.Types.EmailAddress
 import com.gu.identityBackfill.salesforce.UpdateSalesforceIdentityId.IdentityId
-import com.gu.salesforce.SalesforceClient.StringHttpRequest
 import com.gu.util.resthttp.HttpOp.HttpOpWrapper
+import com.gu.util.resthttp.RestRequestMaker
 import com.gu.util.resthttp.RestRequestMaker._
-import com.gu.util.resthttp.{HttpOp, RestRequestMaker}
-import okhttp3.{HttpUrl, Request, Response}
 import play.api.libs.json.{JsValue, Json, Reads}
 
 object CreateGuestAccount {
 
-  case class WireGuestRegistrationResponse(
-    userId: String
-  )
+  case class WireGuestRegistrationResponse(userId: String)
   implicit val reads = Json.reads[WireGuestRegistrationResponse]
 
-  case class WireGuestRegistrationRequest(
-    primaryEmailAddress: String //,
-  //    privateFields: ,all optional
-  //    publicFields: req-displayName only
-  )
+  case class WireGuestRegistrationRequest(primaryEmailAddress: String)
   implicit val writes = Json.writes[WireGuestRegistrationRequest]
 
   case class WireIdentityResponse(status: String, guestRegistrationRequest: WireGuestRegistrationResponse)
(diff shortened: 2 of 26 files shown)


handlers/cancellation-sf-cases/src/main/scala/com/gu/cancellation/sf_cases/Handler.scala
```diff
@@ -160,11 +160,11 @@ object Handler extends Logging {
       (for {
         identityAndSfRequests <- sfBackendForIdentityCookieHeader(apiGatewayRequest.headers)
         raiseCaseDetail <- apiGatewayRequest.bodyAsCaseClass[RaiseCaseDetail]()
-        lookupByIdOp = SalesforceGenericIdLookup(identityAndSfRequests.sfClient.wrap(JsonHttp.get))
-        mostRecentCaseOp = SalesforceCase.GetMostRecentCaseByContactId(identityAndSfRequests.sfClient.wrap(JsonHttp.get))
-        createCaseOp = SalesforceCase.Create(identityAndSfRequests.sfClient.wrap(JsonHttp.post))
+        lookupByIdOp = SalesforceGenericIdLookup(identityAndSfRequests.sfClient.wrapWith(JsonHttp.get))
```
johnduffell: this is just a simple rename, not sure if it's more clear now?

handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/Handler.scala
```diff
@@ -53,25 +55,28 @@ object Handler {
...
 
       lazy val sfAuth: LazyClientFailableOp[HttpOp[StringHttpRequest, RestRequestMaker.BodyAsString]] = SalesforceClient(response, sfConfig)
-      lazy val sfPatch = sfAuth.map(_.wrap(JsonHttp.patch))
-      lazy val sfGet = sfAuth.map(_.wrap(JsonHttp.get))
+      lazy val sfPatch = sfAuth.map(_.wrapWith(JsonHttp.patch))
+      lazy val sfGet = sfAuth.map(_.wrapWith(JsonHttp.get))
 
       Operation(
-        steps = IdentityBackfillSteps(
+        steps = WireRequestToDomainObject(IdentityBackfillSteps(
```
johnduffell: I have added the WireRequestToDomainObject pattern just to split out the part that deals with the ApiGatewayRequest object, this makes the tests simpler for the Steps part.

handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/Handler.scala
```diff
@@ -129,15 +134,49 @@ object Handler {
...
   ): ApiResponse =
     (for {
-      identityId <- getByEmail(EmailAddress("john.duffell@guardian.co.uk"))
+      maybeIdentityId <- getByEmail.runRequest(EmailAddress("john.duffell@guardian.co.uk"))
         .toApiGatewayOp("problem with email").withLogging("healthcheck getByEmail")
+      identityId <- maybeIdentityId match {
+        case GetByEmail.ValidatedEmail(identityId) => ContinueProcessing(identityId)
+        case other =>
+          logger.error(s"failed healthcheck with $other")
+          ReturnWithResponse(ApiGatewayResponse.internalServerError("test identity id was not present"))
```
johnduffell: the test would only ever pass if the id was validated, however now validated/non validated is pulled out into a separate data type rather than falling into the Left of the Either

handlers/identity-backfill/src/main/scala/com/gu/identity/GetByEmail.scala
```diff
@@ -3,18 +3,17 @@ package com.gu.identity
...
 
 object GetByEmail {
 
-  sealed trait ApiError
-  case class OtherError(message: String) extends ApiError
-  case object NotFound extends ApiError
-  case object NotValidated extends ApiError
+  sealed trait MaybeValidatedEmail
+  case class ValidatedEmail(identityId: IdentityId) extends MaybeValidatedEmail
+  case object NotValidated extends MaybeValidatedEmail
```
johnduffell: this sealed trait is a bit weirdly named, it's kind of an Option hence the Maybe wording, and it's depedning if the email is validated, but it contains the identity id, rather than the email address.

handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/PreReqCheck.scala
```diff
@@ -1,41 +1,41 @@
...
-import TypeConvert._
-import com.gu.identityBackfill.salesforce.UpdateSalesforceIdentityId.IdentityId
-import com.gu.salesforce.TypesForSFEffectsData.SFContactId
 import com.gu.util.resthttp.LazyClientFailableOp
+import com.gu.util.resthttp.Types.{ClientFailableOp, ClientFailure, ClientSuccess, NotFound}
 
 object PreReqCheck {
 
-  case class PreReqResult(zuoraAccountId: Types.AccountId, sFContactId: SFContactId, requiredIdentityId: IdentityId)
+  case class PreReqResult(zuoraAccountId: Types.AccountId, sFContactId: SFContactId, existingIdentityId: Option[IdentityId])
```
johnduffell: not sure if better to have a sealed trait instead of an Option, generally I prefer to but this one I don't know if there's much in it

lib/salesforce/src/main/scala/com/gu/salesforce/SalesforceClient.scala
```diff
@@ -39,6 +39,6 @@ object SalesforceClient {
     override def builder: Request.Builder = new Request.Builder().get()
   }
 
-  case class StringHttpRequest(relativePath: RelativePath, requestMethod: RequestMethod)
+  case class StringHttpRequest(requestMethod: RequestMethod, relativePath: RelativePath, urlParams: UrlParams)
```
johnduffell: no idea why this is in the salesforce client, perhaps should be moved out

handlers/identity-backfill/src/main/scala/com/gu/identity/GetByEmail.scala
```diff
@@ -3,18 +3,17 @@ package com.gu.identity
...
+import com.gu.util.resthttp.Types.{ClientFailableOp, ClientSuccess, GenericError}
+import play.api.libs.json.{JsValue, Json, Reads}
 
 object GetByEmail {
 
-  sealed trait ApiError
-  case class OtherError(message: String) extends ApiError
-  case object NotFound extends ApiError
-  case object NotValidated extends ApiError
+  sealed trait MaybeValidatedEmail
```
jacobwinch: Not sure how much better this is, but we could just have something like:

```suggestion
  sealed trait IdentityEmailAddress
  case class ValidatedEmailAddress(identityId: IdentityId) extends IdentityEmailAddress
  case object UnvalidatedEmailAddress extends IdentityEmailAddress
```

handlers/identity-backfill/src/main/scala/com/gu/identity/GetByEmail.scala
```diff
@@ -29,43 +28,24 @@ object GetByEmail {
...
+      RestRequestMaker.toResult[UserResponse](_).flatMap(toResponse)
+    )
 
-  def userFromResponse(userResponse: UserResponse): ApiError \/ User =
+  def userFromResponse(userResponse: UserResponse): ClientFailableOp[User] =
     userResponse match {
-      case UserResponse("ok", user) => \/-(user)
-      case _ => -\/(OtherError("not an OK response from api"))
+      case UserResponse("ok", user) => ClientSuccess(user)
+      case _ => GenericError("not an OK response from api")
```
jacobwinch: If I was debugging this I'd want to know what the specific error was...

handlers/identity-backfill/src/main/scala/com/gu/identity/IdentityClient.scala
```diff
@@ -0,0 +1,41 @@
+package com.gu.identity
+
+import com.gu.salesforce.SalesforceClient.StringHttpRequest
+import com.gu.util.config.ConfigLocation
+import com.gu.util.resthttp.HttpOp
+import com.gu.util.resthttp.RestRequestMaker.{BodyAsString, toClientFailableOp}
+import okhttp3.{HttpUrl, Request, Response}
+import play.api.libs.json.{Json, Reads}
+
+object IdentityClient {
```
jacobwinch: 👍

handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/Handler.scala
```diff
@@ -53,25 +55,28 @@ object Handler {
...
+      lazy val sfPatch = sfAuth.map(_.wrapWith(JsonHttp.patch))
+      lazy val sfGet = sfAuth.map(_.wrapWith(JsonHttp.get))
 
       Operation(
-        steps = IdentityBackfillSteps(
+        steps = WireRequestToDomainObject(IdentityBackfillSteps(
           PreReqCheck(
-            getByEmail,
+            getByEmail.runRequest,
             GetZuoraAccountsForEmail(zuoraQuerier) _ andThen PreReqCheck.getSingleZuoraAccountForEmail,
```
jacobwinch: So if we have a user with two separate billing accounts and no identity account, we will not create a ghost account for these users?

lib/salesforce/src/main/scala/com/gu/salesforce/SalesforceClient.scala
johnduffell: ok I have moved this to rest http in the last commit

handlers/identity-backfill/src/main/scala/com/gu/identity/GetByEmail.scala
johnduffell: I think yours is similarly confusing, I think this is better:
```scala
  sealed trait IdentityAccount
  case class IdentityAccountWithValidatedEmail(identityId: IdentityId) extends IdentityAccount
  case object IdentityAccountWithUnvalidatedEmail extends IdentityAccount
```

handlers/identity-backfill/src/main/scala/com/gu/identity/GetByEmail.scala
johnduffell: yes agreed I'll add, although the logging in the Http effect is pretty comprehensive already
https://github.com/guardian/zuora-auto-cancel/blob/master/lib/effects/src/main/scala/com/gu/effects/Http.scala#L28

handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/Handler.scala
johnduffell: good point and there's no reason why that's needed from a data point of view, although I wouldn't like to add that change to this PR. 
 @paulbrown1982 this could be done as a separate change if necessary

handlers/identity-backfill/src/main/scala/com/gu/identity/CreateGuestAccount.scala
```diff
@@ -0,0 +1,31 @@
...
+
+object CreateGuestAccount {
+
+  case class WireGuestRegistrationResponse(userId: String)
+  implicit val reads = Json.reads[WireGuestRegistrationResponse]
+
+  case class WireGuestRegistrationRequest(primaryEmailAddress: String)
+  implicit val writes = Json.writes[WireGuestRegistrationRequest]
+
+  case class WireIdentityResponse(status: String, guestRegistrationRequest: WireGuestRegistrationResponse)
```
pvighi: guestRegistration**Request** is of type WireGuestRegistration**Response** ?

lib/restHttp/src/main/scala/com/gu/util/resthttp/RestRequestMaker.scala
```diff
@@ -29,8 +29,8 @@ object RestRequestMaker extends Logging {
       case success: JsSuccess[T] =>
         ClientSuccess(success.get)
       case error: JsError => {
-        logger.error(s"Failed to convert Zuora response to case case $error. Response body was: \n $bodyAsJson")
-        GenericError(s"Error when converting Zuora response to case class: $error")
+        logger.error(s"Failed to convert JSON response to case case $error. Response body was: \n $bodyAsJson")
```
twrichards: typo here, think it should be...
```suggestion
        logger.error(s"Failed to convert JSON response to case class $error. Response body was: \n $bodyAsJson")
```

handlers/identity-backfill/src/main/scala/com/gu/identity/GetByEmail.scala
```diff
@@ -29,43 +28,24 @@ object GetByEmail {
 
   }
 
-  def identityIdFromUser(user: User) =
-    IdentityId(user.id)
+  val wrapper: HttpOpWrapper[EmailAddress, GetRequestWithParams, JsValue, IdentityAccount] =
+    HttpOpWrapper[EmailAddress, GetRequestWithParams, JsValue, IdentityAccount](
+      emailAddress => GetRequestWithParams(RelativePath(s"/user"), UrlParams(Map("emailAddress" -> emailAddress.value))),
+      RestRequestMaker.toResult[UserResponse](_).flatMap(toResponse)
```
pvighi: maybe it's the fact that I have not really used this a lot but I think it would be more readable if we use the parameter names like this :
```scala
    HttpOpWrapper[EmailAddress, GetRequestWithParams, JsValue, IdentityAccount](
      fromNewParam = emailAddress => GetRequestWithParams(RelativePath(s"/user"), UrlParams(Map("emailAddress" -> emailAddress.value))),
      toNewResponse =  RestRequestMaker.toResult[UserResponse](_).flatMap(toResponse)
    )

```

and maybe given the input param a name instead of using underscore in the toNewResponse function might help

handlers/identity-backfill/src/main/scala/com/gu/identity/IdentityClient.scala
```diff
@@ -0,0 +1,41 @@
...
+import com.gu.util.resthttp.HttpOp
+import com.gu.util.resthttp.JsonHttp.StringHttpRequest
+import com.gu.util.resthttp.RestRequestMaker.{BodyAsString, toClientFailableOp}
+import okhttp3.{HttpUrl, Request, Response}
+import play.api.libs.json.{Json, Reads}
+
+object IdentityClient {
+
+  def apply(
+    response: Request => Response,
```
pvighi: minor: I always find it confusing when we call the function that takes a request and returns a response just "response" because it looks like its just a response and not the action of converting a request into a response

handlers/identity-backfill/src/main/scala/com/gu/identity/IdentityClient.scala
```diff
@@ -0,0 +1,41 @@
...
+    config: IdentityConfig
+  ): HttpOp[StringHttpRequest, BodyAsString] =
+    HttpOp(response).flatMap {
```
pvighi: minor: i always find it confusing when we call the function from request to response just "response" because it looks like it is just a response object

handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/Handler.scala
```diff
@@ -129,15 +134,49 @@ object Handler {
...
 object Healthcheck {
   def apply(
-    getByEmail: EmailAddress => \/[GetByEmail.ApiError, IdentityId],
+    getByEmail: HttpOp[EmailAddress, GetByEmail.IdentityAccount],
     countZuoraAccountsForIdentityId: IdentityId => ClientFailableOp[Int],
     sfAuth: LazyClientFailableOp[Any]
   ): ApiResponse =
     (for {
-      identityId <- getByEmail(EmailAddress("john.duffell@guardian.co.uk"))
+      maybeIdentityId <- getByEmail.runRequest(EmailAddress("john.duffell@guardian.co.uk"))
```
pvighi: now can't ever leave the guardian otherwise the health check will fail

handlers/identity-backfill/src/main/scala/com/gu/identity/CreateGuestAccount.scala
johnduffell: yeah that is totally bizarre, basically when you make a request identity will return the response under the key name of the request you made.  I wouldn't have got that right without the effectsTests pointing out which way round it should be!

handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/Handler.scala
johnduffell: that's not true, as long as I have an identity account it will be okay!

pvighi: but will you keep the guardian email account ?

handlers/identity-backfill/src/main/scala/com/gu/identity/GetByEmail.scala
johnduffell: better still, not to inline code into the structure, so I will split them out.  Then it will be this:
```scala
    HttpOpWrapper[EmailAddress, GetRequestWithParams, JsValue, IdentityAccount](
      fromNewParam = emailAddressToParams,
      toNewResponse = jsToWireModel.andThen(_.flatMap(wireToDomainModel))
    )
```

handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/Handler.scala
johnduffell: it's not the email, it's just the identity account, so as long as I don't forget my password I can stay logged in!  Of course I can't reset the password etc after that.
But you do have a good point, I just wanted an identity account that exists in CODE and PROD.  Using it in the effectsTests is one thing, but using in the health check is another.  Perhaps we should use an operation other than getByEmail to validate that we can get in touch with identity alright.  Now that we have the create guest account operation, I'm not sure we should create a guest account every 5 minutes though!

handlers/identity-backfill/src/main/scala/com/gu/identity/IdentityClient.scala
johnduffell: oh I just realised I saw your comment and fixed it in SalesforceClient instead 🙈 but yes I should have done it here, but probably not worth another PR at this point

PR #211, opened 2018-11-05
Send payment failure emails via SQS

QuarpT:
Work done:

Send payment failure emails via SQS to be sent by membership-workflow. We no longer use the ET trigger ID. Instead we use a campaign ID (DataExtensionName) recognised by membership-workflow.

Why are we doing this?

We are migrating email service providers. We don't want to add Braze as a third party dependency of the zuora-auto-cancel lambda. Instead we publish to SQS and rely on membership-workflow to send emails.

Before merge:

A membership workflow PR needs to created to support sending these 5 emails
Need to figure out how to get an sfContactId or identityId to send on the sqs queue


src/main/scala/com/gu/autoCancel/AutoCancelHandler.scala
```diff
@@ -51,4 +50,9 @@ object AutoCancelHandler extends App with Logging {
       operationForEffects(RawEffects.stage, GetFromS3.fetchString, RawEffects.response, RawEffects.now)
     }
 
+  def emailQueueFor(stage: Stage): QueueName = stage match {
+    case Stage("PROD") => QueueName("contributions-thanks")
```
QuarpT: Not sure whether to create a new queue for these emails... 
membership-workflow reads off three different queues for triggering emails, but behaves identically for each. I can't see the benefit of creating different queues for different emails.

pvighi: If we treat them all exactly the same I don't really know either

lib/effects-sqs/src/main/scala/com/gu/effects/sqs/AwsSQSSend.scala
```diff
@@ -38,6 +43,22 @@ object AwsSQSSend {
         Failure(throwable)
     }
   }
+
+  def sendSync(queueName: QueueName)(payload: Payload): Try[Unit] = {
```
pvighi: Why did we need a sync version of this ? 
couldn't this just use the async version internally and wait for it or is there something different ? (or just use the async version directly)

QuarpT: I need a sync version since there is no advantage lifting all the `ApiGatewayOp`s to `AsyncApiGatewayOp`s in the context of the payment failure and auto cancel lambdas, given the lambdas execute their tasks in sequentially. (The REST client is already blocking)

You're right, we could wait for the future version. I think this implementation is cleaner than wrapping `Await.result` around a scala Future.

johnduffell: I think in general it would be nice to move to async and this was a starting point, however I agree there is a lot of legacy sync code especially http calls.  There are a few places where it would be quicker if we could do some calls in parallel, but we never got as far as making an async http client.
changing the PF/AC lambdas to use async would make the later conversion easier, but it would have to be from someone who has time to go through it.  I think there is only one example of async lambda so far.

src/main/scala/com/gu/autoCancel/AutoCancelHandler.scala
```diff
@@ -28,15 +29,13 @@ object AutoCancelHandler extends App with Logging {
...
-      etConfig <- loadConfigModule[ETConfig].toApiGatewayOp("load et config")
     } yield {
       val zuoraRequests = ZuoraRestRequestMaker(response, zuoraRestConfig)
       AutoCancelSteps(
         AutoCancel.apply(zuoraRequests),
         AutoCancelDataCollectionFilter.apply(now().toLocalDate, ZuoraGetAccountSummary(zuoraRequests)),
-        etConfig.etSendIDs,
         ZuoraEmailSteps.sendEmailRegardingAccount(
-          EmailSendSteps(ETClient.sendEmail(response, etConfig), FilterEmail(stage)),
+          EmailSendSteps(AwsSQSSend.sendSync(emailQueueFor(stage))),
```
johnduffell: the convention is that any impure function must only be referenced at the main handler of the lambda rather than referenced directly elsewhere within the code.  So AwsSQSSend.sendSync would have to be passed in.  This just helps testability and means that if you call the top level with pure function then you know that the whole lambda is just a pure function.

src/main/scala/com/gu/autoCancel/AutoCancelHandler.scala
johnduffell: yes agreed I think we just didn't see the need to pay for a load of queues that were hardly used.  This was from the days where cost cutting was the focus  I believe.

src/main/scala/com/gu/autoCancel/AutoCancelSteps.scala
```diff
@@ -28,22 +28,21 @@ object AutoCancelSteps extends Logging {
   def apply(
     autoCancel: AutoCancelRequest => ApiGatewayOp[Unit],
     autoCancelFilter: AutoCancelCallout => ApiGatewayOp[AutoCancelRequest],
-    etSendIds: ETSendIds,
-    sendEmailRegardingAccount: (String, PaymentFailureInformation => EmailRequest) => ApiGatewayOp[Unit]
+    sendEmailRegardingAccount: (String, PaymentFailureInformation => EmailMessage) => ApiGatewayOp[Unit]
```
johnduffell: String is not very informative, this should have been wrapped in a value class as soon as it was known to have more meaning

johnduffell: oh, I just realised that is existing code 😆 
so only fix if you feel like it's easy enough to do

src/main/scala/com/gu/paymentFailure/Lambda.scala
```diff
@@ -35,27 +39,42 @@ object Lambda {
...
   // it's the only part you can't test of the handler
-  def handleRequest(inputStream: InputStream, outputStream: OutputStream, context: Context): Unit =
-    runForLegacyTestsSeeTestingMd(RawEffects.stage, GetFromS3.fetchString, RawEffects.response, LambdaIO(inputStream, outputStream, context))
+  def handleRequest(inputStream: InputStream, outputStream: OutputStream, context: Context): Unit = {
+    runForLegacyTestsSeeTestingMd(
+      RawEffects.stage,
+      GetFromS3.fetchString,
+      RawEffects.response,
+      LambdaIO(inputStream, outputStream, context),
+      AwsSQSSend.sendSync
```
johnduffell: oh yes this is pretty much what I meant for the other one, pass it in at this level

src/main/scala/com/gu/util/email/EmailSendSteps.scala
```diff
@@ -0,0 +1,106 @@
...
+    case _ => None
+  }
+  val cancelledId = EmailId("cancelled-payment-email")
+}
+
+case class EmailMessage(To: ToDef, DataExtensionName: String)
+
+trait EmailSqsSerialisation {
+  implicit val subscriberAttributesWrites = new Writes[SubscriberAttributesDef] {
+    override def writes(o: SubscriberAttributesDef): JsValue = {
```
johnduffell: perhaps this is just moved from elsewhere! but 
convention here is to create a low level case class with "wire" in the name, and just generate the default serialiser from that.  Then have a function to map from the high level domain model to the low level wire model.
This is just because custom serialisers tend to be harder to understand and test compared with a function between two case classes.

src/main/scala/com/gu/util/email/EmailSendSteps.scala
```diff
@@ -0,0 +1,106 @@
...
+      val allFields = fields ++ optionalFields
+      JsObject(allFields)
+    }
+  }
+  implicit val contactAttributesWrites = Json.writes[ContactAttributesDef]
+  implicit val toDefWrites = Json.writes[ToDef]
+  implicit val emailMessageWrites = Json.writes[EmailMessage]
+}
+
+object EmailSendSteps extends EmailSqsSerialisation {
```
johnduffell: the trait it just used to get the imports - why not just use an import instead?

src/main/scala/com/gu/util/email/EmailSendSteps.scala
```diff
@@ -0,0 +1,106 @@
...
+  def apply(sqsSend: Payload => Try[Unit])(emailRequest: EmailMessage): ApiGatewayOp[Unit] = {
+    sqsSend(Payload(Json.toJson(emailRequest).toString)) match {
+      case Success(_) =>
+        ContinueProcessing(())
+      case Failure(_) =>
+        logger.error(s"failed to send $emailRequest to sqs queue")
+        ReturnWithResponse(ApiResponse("500", "failure trigger email"))
```
johnduffell: failure to send to sqs in this case.

src/main/scala/com/gu/util/email/EmailSendSteps.scala
```diff
@@ -0,0 +1,106 @@
...
+  implicit val emailMessageWrites = Json.writes[EmailMessage]
+}
+
+object EmailSendSteps extends EmailSqsSerialisation {
+  def apply(sqsSend: Payload => Try[Unit])(emailRequest: EmailMessage): ApiGatewayOp[Unit] = {
+    sqsSend(Payload(Json.toJson(emailRequest).toString)) match {
+      case Success(_) =>
+        ContinueProcessing(())
+      case Failure(_) =>
+        logger.error(s"failed to send $emailRequest to sqs queue")
```
johnduffell: might be worth adding the `t` from the Failure as a second parameter to the logger

src/test/scala/com/gu/paymentFailure/EndToEndHandlerTest.scala
```diff
@@ -141,7 +148,7 @@ trait EndtoEndBaseData {
...
+  override val expectedEmailSend = """{"DataExtensionName":"first-failed-payment-email", "To":{"Address":"john.duffell@guardian.co.uk","SubscriberKey":"john.duffell@guardian.co.uk","ContactAttributes":{"SubscriberAttributes":{"subscriber_id":"A-S00071536","product":"Supporter","payment_method":"CreditCardReferenceTransaction","card_type":"Visa","card_expiry_date":"12/2019","first_name":"eSAFaBwm4WJZNg5xhIc","last_name":"eSAFaBwm4WJZNg5xhIc","paymentId":"2c92c0f95fc912eb015fcb2a481720e6","price":"$49.00","serviceStartDate":"17 November 2017","serviceEndDate":"16 November 2018", "billing_address1": "billingAddress1Value", "billing_address2": "billingAddress2Value", "billing_postcode": "billingPostcodeValue", "billing_city": "billingCityValue", "billing_state": "billingStateValue", "billing_country" : "billingCountryValue", "title" : "billingTitleValue"}}}}"""
```
johnduffell: 😂 if you fancied reformatting this and adding a stripMargin that would be a small improvement on the old version

src/test/scala/com/gu/paymentFailure/EndToEndHandlerTest.scala
```diff
@@ -203,7 +210,7 @@ object EndToEndDataWithBillingDetails extends EndtoEndBaseData {
 object EndToEndData extends EndtoEndBaseData {
```
johnduffell: not sure how this test ended up using inheritance.  with it testing everything apart from the side effects it is fundamentally a bit too complex, so I realise it's hard to make it clear what's going on!

src/test/scala/com/gu/paymentFailure/PaymentFailureHandlerTest.scala
```diff
@@ -75,29 +65,27 @@ class PaymentFailureHandlerTest extends FlatSpec with Matchers {
...
+            last_name = "User",
+            primaryKey = PaymentId("somePaymentId"),
+            price = "£49.00",
+            serviceStartDate = "21 November 2016",
+            serviceEndDate = "21 December 2016"
           )
         )
-      )
+      ),
+      "first-failed-payment-email"
```
johnduffell: could be clearer as a named parameter like the others?

src/test/scala/manualTest/EmailClientSystemTest.scala
```diff
@@ -1,70 +0,0 @@
-package manualTest
```
johnduffell: this is a big shame to delete - this test is very useful for sending the full range of emails for testing purposes.  Without it, we have to have faith that the actual external integration is working.  It could be part of membership workflow itself, but I think this project is the one that has full knowledge of the email formats.
Would it be possible to make this send to the queues with the right format?

johnduffell:
Thanks for this, just a few comments, some of which are more general ones.  Also I'm not sure what happened to the coverage, coveralls is saying it's gone down quite a lot.  This could just be because code that was tested well has been deleted (which is not a problem) and that there is new side effects stuff (which would only be testable with effectsTests) but it could be that some replacement stuff is missing tests.

lib/effects-sqs/src/main/scala/com/gu/effects/sqs/AwsSQSSend.scala
QuarpT: Refactoring to async could be seen as over optimisation in this context :trollface:

src/main/scala/com/gu/util/email/EmailSendSteps.scala
QuarpT: exception is already logged by sqs send

lib/effects-sqs/src/main/scala/com/gu/effects/sqs/AwsSQSSend.scala
johnduffell: true, scala programs are sequential by default, so we get the sync behaviour for free whether we want it or not.  If we had to model it in the types regardless, there wouldn't be that excuse!

src/main/scala/com/gu/util/email/EmailSendSteps.scala
QuarpT: This was moved. I agree that case class serialisers are preferable.

src/main/scala/com/gu/util/email/EmailSendSteps.scala
QuarpT: For this scenario isn't import vs trait just stylistic?

src/main/scala/com/gu/util/email/EmailSendSteps.scala
johnduffell: fair enough, if you have inclination update it, if not leave it as is!

src/main/scala/com/gu/util/email/EmailSendSteps.scala
johnduffell: didn't know that, should be fine to log as sqs send only.

src/test/scala/manualTest/EmailClientSystemTest.scala
QuarpT: yes, I will add this

johnduffell:
fyi coveralls will come back and automatically edit the original comment, once the travis build passes, so you can see the updated coverage.  the build status for coveralls only shows since the last build for some reason, wheras the comment at the top will show overall fo the PR

src/test/scala/com/gu/paymentFailure/EndToEndHandlerTest.scala
```diff
@@ -203,7 +240,28 @@ object EndToEndDataWithBillingDetails extends EndtoEndBaseData {
...
+      |        "first_name": "eSAFaBwm4WJZNg5xhIc",
+      |        "last_name": "eSAFaBwm4WJZNg5xhIc",
+      |        "paymentId": "2c92c0f95fc912eb015fcb2a481720e6",
+      |        "price": "$49.00",
+      |        "serviceStartDate": "17 November 2017",
+      |        "serviceEndDate": "16 November 2018"
+      |      }
+      |    }
+      |  }
+      |}"""
```
johnduffell: the build tests is failing because of this missing
```suggestion
      |}""".stripMargin
```

QuarpT:
@paulbrown1982 I'll quickly change this to sfContactId

johnduffell:
did you solve the issue with emails not getting sent through the test?
also you forgot to get a +1 or Approve for the changes before merging.
however, looks good to me 👍

PR #212, opened 2018-11-06
Direct Debit Mandate Syncing Process (GoCardless -> SalesForce)

twrichards:
NEW LAMBDA for polling GoCardless for Mandate Events and syncing into SalesForce

#### General Process
1. Fetch the last processed event from SalesForce (there might not be one)
2. Fetch `[batchSize]` worth of events from GoCardless events endpoint (filtering on Mandate Events), **which occur after** either
    1. the GoCardless Event ID returned from step one (i.e. SalesForce already has some GoCardless data)
    2. 25 Aug 2015 (since this is when the first event occurred in production GoCardless) - this would be the start of a back-fill
3. In a single request, lookup the IDs of all the Mandate records in SalesForce which are referenced by the current batch of events to process (so we can later link the new event to the existing mandate or create mandate too if it doesn't already exist)
4. In a single request fetch existing payment method & billing account records in SalesForce which are referenced by the current batch events to process 
5. Now for each mandate event in the batch 
    1. Create the `DD Mandate` record (if it doesn't already exist in SalesForce)
    2. Create the `DD Mandate Event` record (with the ID from the previous step, and the payment method & billing account references if they exist)
    3. Update the `DD Mandate record` to mark the new event as the latest event

#### Polling
Achieved by invoking this Lambda via a `AWS::Events::Rule` configured in [cfn.yml](cfn.yaml) (currently every 5 mins)

#### Batching (API limits)
To avoid eating up the SalesForce (or indeed GoCardless) API limit, each invocation of this lambda is limited to processing a limited number of events per invocation as defined in the config file (`goCardless-{STAGE}.json`) in the `gu-reader-revenue-private/membership/support-service-lambdas` S3 bucket. NOTE: this is capped at 200 in the code.

The number of actual API calls to Salesforce will be 3-5 times more than this batch limit (as a number of calls are required per mandate event, and this is variable as to whether the underlying mandate object needs to be created).

Perhaps in future we could be more sophisticated in how this limit works, since lots of the events come in bursts which therefore take longer to process if limited by batches, what we want is really a rolling 24 hour cap on API calls.

#### Back-fill & Sync Duality
This lambda is designed to both back-fill and keep SalesForce up to date (with no additional params required), essentially on each invocation of the Lambda it processes a certain number of GoCardless Mandate Events, using the _last processed_ event as a starting point. This means that if SalesForce doesn't have any events yet, this will just back-fill from the beginning of time, however if SalesForce has some events this will just continue from the last event.

#### SalesForce Objects
- DD Mandates (`DD_Mandate__c`)
- DD Mandate Events (`DD_Mandate_Event__c`)

#### Errors
There are a few log filters & alarms configured in [cfn.yml](cfn.yaml) 
- "GenericError Count" filer & alarm, is fired if "HTTP request was unsuccessful" is found in the logs. **Fairly important given this lambda is designed to stop processing events upon any error (to avoid missing data)**
- TODO add metric events processed filter and mandate records created filter (with alarm if this drops below certain threshold per 12 hours)
- TODO add 'StoppedRunning' alarm based on existing Invocations metric
- TODO add a 'SocketTimeoutException' alarm if occurs more than 3 times in a row


twrichards:
@pvighi @jacobwinch - since this is a pretty big PR, it might take a few passes, please could I get your thoughts on structure (appreciate its lacking any unit tests at this point, I will be adding those tomorrow, and the effects test currently aren't actually checking the results, just printing, those will get updated too)

handlers/sf-gocardless-sync/cfn.yaml
```diff
@@ -0,0 +1,95 @@
...
+                - PolicyName: LambdaPolicy
+                  PolicyDocument:
+                      Statement:
+                          - Effect: Allow
+                            Action:
+                            - logs:CreateLogGroup
+                            - logs:CreateLogStream
+                            - logs:PutLogEvents
+                            - lambda:InvokeFunction
+                            Resource: !Sub "arn:aws:logs:${AWS::Region}:${AWS::AccountId}:log-group:/aws/lambda/sf-gocardless-sync-${Stage}:log-stream:*"
```
jacobwinch: Reminder that this log group should have a 14 day retention policy

handlers/sf-gocardless-sync/cfn.yaml
```diff
@@ -0,0 +1,95 @@
...
+                - PolicyName: ReadPrivateCredentials
+                  PolicyDocument:
+                      Statement:
+                          - Effect: Allow
+                            Action: s3:GetObject
+                            Resource:
+                            - !Sub arn:aws:s3:::gu-reader-revenue-private/membership/support-service-lambdas/${Stage}/goCardless-${Stage}.*.json
+                            - !Sub arn:aws:s3:::gu-reader-revenue-private/membership/support-service-lambdas/${Stage}/trustedApi-${Stage}.*.json
```
jacobwinch: Can you remind me what this piece of config is for?

handlers/sf-gocardless-sync/cfn.yaml
```diff
@@ -0,0 +1,95 @@
...
+            Environment:
+                Variables:
+                  Stage: !Ref Stage
+            Role:
+                Fn::GetAtt:
+                - "GoCardlessSalesForceSyncRole"
+                - Arn
+            MemorySize: 1536 # TODO review this amount of memory is required
+            Runtime: java8
+            Timeout: 300
```
jacobwinch: This can now be longer running if necessary: https://aws.amazon.com/about-aws/whats-new/2018/10/aws-lambda-supports-functions-that-can-run-up-to-15-minutes/

jacobwinch: Oh I've just seen the cron expression, I doubt this is necessary...

handlers/sf-gocardless-sync/cfn.yaml
```diff
@@ -0,0 +1,95 @@
...
+        DependsOn:
+        - "GoCardlessSalesForceSyncRole"
+
+    GoCardlessSalesForceSyncScheduleRule:
+      Type: AWS::Events::Rule
+      Properties:
+        Description: "Trigger the GoCardlessSalesForceSync every 5 mins"
+        ScheduleExpression: "cron(0/5 * * * ? *)" # TODO consider more sophisticated cron schedule (to reduce api calls, e.g. overnight)
```
jacobwinch: Minor, personally I find rate(5 minutes) easier to understand: https://docs.aws.amazon.com/lambda/latest/dg/tutorial-scheduled-events-schedule-expressions.html

handlers/sf-gocardless-sync/cfn.yaml
```diff
@@ -0,0 +1,95 @@
...
+
+    InvokeGoCardlessSalesForceSyncLambdaPermission:
+      Type: AWS::Lambda::Permission
+      Properties:
+        FunctionName: !Ref GoCardlessSalesForceSyncLambda
+        Action: lambda:InvokeFunction
+        Principal: events.amazonaws.com
+        SourceArn: !GetAtt [GoCardlessSalesForceSyncScheduleRule, Arn]
+
+# TODO add some alarms
```
jacobwinch: 👍

handlers/sf-gocardless-sync/src/main/scala/com/gu/sf_gocardless_sync/Handler.scala
```diff
@@ -0,0 +1,148 @@
...
+
+  def forEachMandateUpdate(
+    goCardless: GcClient,
+    sf: SfClient,
+    existingSfMandates: ExistingMandateMutableMap,
+    relatedPaymentMethodAndBillingAccountIDs: Map[Reference, SfPaymentMethodDetail]
+  )(
+    gcMandateUpdateWithDetail: MandateUpdateWithMandateDetail
+  ): ClientFailableOp[Unit] = for {
+    sfMandate <- existingSfMandates.get(gcMandateUpdateWithDetail.event.links.mandate) match {
```
jacobwinch: I find for comprehensions like this tricky to follow, I would always prefer to pull something like a pattern match into a separate function...

handlers/sf-gocardless-sync/src/main/scala/com/gu/sf_gocardless_sync/Handler.scala
```diff
@@ -0,0 +1,148 @@
...
+      sfMandate.Id,
+      relatedPaymentMethodAndBillingAccountIDs.get(gcMandateUpdateWithDetail.mandate.reference)
+    )
+  } yield sfMandate match {
+    // if mandate had to be created in SF then always patch the 'Last Update'
+    case MandateWithSfId(_) => patchMandateOp
+    // if mandate already existed in SF then only patch the 'Last Update' if it's more recent
+    case MandateLookupDetail(_, _, _, lastUpdated) if lastUpdated.value < gcMandateUpdateWithDetail.event.created_at => //TODO might need proper datetime comparison
+      patchMandateOp
+    // otherwise
```
jacobwinch: Minor: don't need the comment here

handlers/sf-gocardless-sync/src/main/scala/com/gu/sf_gocardless_sync/Handler.scala
```diff
@@ -0,0 +1,148 @@
...
+    sf: SfClient,
+    updatesSinceLastProcessed: List[MandateUpdateWithMandateDetail] // try changing type to NonEmptyList
+  ) = for {
+    existingSfMandates <- SalesforceDDMandate.LookupAll(sf.client.wrapWith(JsonHttp.get))(updatesSinceLastProcessed.map(gcMandateUpdate => gcMandateUpdate.mandate.id)).toDisjunction
+    fetchRelatedPaymentMethodAndBillingAccountIDs = SalesforceDDMandate.GetAllPaymentMethodWithBillingAccountGivenGoCardlessReference(sf.client.wrapWith(JsonHttp.get))
+    relatedPaymentMethodAndBillingAccountIDs <- fetchRelatedPaymentMethodAndBillingAccountIDs(updatesSinceLastProcessed.map(gcMandateUpdate => gcMandateUpdate.mandate.reference)).toDisjunction
+  } yield updatesSinceLastProcessed.reverse.toStream.map(
+    forEachMandateUpdate(goCardless, sf, existingSfMandates, relatedPaymentMethodAndBillingAccountIDs)
+  ) collectFirst {
+      // stop processing stream on any ClientFailure
```
jacobwinch: Sorry if I'm being slow, but why is stopping on error beneficial in the context of this lambda?

handlers/sf-gocardless-sync/cfn.yaml
twrichards: is that configurable via CFN?

handlers/sf-gocardless-sync/cfn.yaml
twrichards: amazing thanks, doing this now

handlers/sf-gocardless-sync/cfn.yaml
pvighi: maybe something like this: 
https://github.com/guardian/subscriptions-frontend/blob/master/cloud-formation/subscriptions-app.cf.yaml#L264

handlers/sf-gocardless-sync/src/main/scala/com/gu/sf_gocardless_sync/Handler.scala
pvighi: Could it be that just one event in particular cannot be processed for some reason and that blocks everything else ?
Do we have a way to give up on one event after a certain amount of retries ?

handlers/sf-gocardless-sync/src/main/scala/com/gu/sf_gocardless_sync/Handler.scala
```diff
@@ -0,0 +1,148 @@
...
+        goCardless,
+        sf,
+        existingSfMandates
+      )(
+        gcMandateUpdateWithDetail,
+        relatedPaymentMethodAndBillingAccountIDs.get(gcMandateUpdateWithDetail.mandate.reference)
+      )
+      case Some(existingSfMandate) => ClientSuccess(existingSfMandate)
+    }
+    newMandateUpdateOp = SalesforceDDMandateUpdate.Create(sf.client.wrapWith(JsonHttp.post))
```
pvighi: this kind of thing could be an input param to this function so at this level you only care that you have access to a function that takes a WireNewMandateUpdate and returns a  ClientFailableOp[MandateUpdateWithSfId]

handlers/sf-gocardless-sync/src/main/scala/com/gu/sf_gocardless_sync/Handler.scala
```diff
@@ -0,0 +1,148 @@
...
+
+  type WiredClient = HttpOp[JsonHttp.StringHttpRequest, RestRequestMaker.BodyAsString]
+
+  case class SfClient(client: WiredClient) extends AnyVal
+  case class GcClient(client: WiredClient) extends AnyVal
+
+  def apply(inputStream: InputStream, outputStream: OutputStream, context: Context): Unit =
+    for {
+      goCardlessConfig <- LoadConfigModule(RawEffects.stage, GetFromS3.fetchString)[GoCardlessConfig]
+      goCardlessClient = GoCardlessClient(RawEffects.response, goCardlessConfig)
```
pvighi: once you start adding unit tests you will probably have to extract this kind of things in order to be able to test with fake stuff

handlers/sf-gocardless-sync/src/main/scala/com/gu/sf_gocardless_sync/salesforce/SalesforceDDMandate.scala
```diff
@@ -0,0 +1,132 @@
...
+      GoCardless_Mandate_ID__c: GoCardlessMandateID,
+      Last_Mandate_Update__c: MandateUpdateSfId,
+      Status_Changed_At__c: UpdateHappenedAt
+    ) extends WithMandateSfId
+    implicit val reads = Json.reads[MandateLookupDetail]
+
+    private case class MandateSearchQueryResponse(records: List[MandateLookupDetail])
+    private implicit val readsIds = Json.reads[MandateSearchQueryResponse]
+
+    type ExistingMandateMutableMap = scala.collection.mutable.Map[GoCardlessMandateID, MandateLookupDetail];
```
pvighi: mutable ? 😱

handlers/sf-gocardless-sync/src/main/scala/com/gu/sf_gocardless_sync/Handler.scala
```diff
@@ -0,0 +1,148 @@
...
+      relatedPaymentMethodAndBillingAccountIDs.get(gcMandateUpdateWithDetail.mandate.reference)
+    )
+  } yield sfMandate match {
+    // if mandate had to be created in SF then always patch the 'Last Update'
+    case MandateWithSfId(_) => patchMandateOp
+    // if mandate already existed in SF then only patch the 'Last Update' if it's more recent
+    case MandateLookupDetail(_, _, _, lastUpdated) if lastUpdated.value < gcMandateUpdateWithDetail.event.created_at => //TODO might need proper datetime comparison
+      patchMandateOp
+    // otherwise
+    case other => System.err.println(s"\n\njust added an out of order event \n\t${gcMandateUpdateWithDetail.event}\n\t$other\n\n")
```
pvighi: does this end up in the log properly ? maybe we should use the logger for this

handlers/sf-gocardless-sync/src/main/scala/com/gu/sf_gocardless_sync/Handler.scala
```diff
@@ -0,0 +1,148 @@
...
+  def createMandateInSf(
+    goCardless: GcClient,
+    sf: SfClient,
+    existingSfMandates: ExistingMandateMutableMap
+  )(
+    gcMandateUpdateWithDetail: MandateUpdateWithMandateDetail,
+    sfPaymentMethodDetailOption: Option[SfPaymentMethodDetail]
+  ): ClientFailableOp[MandateWithSfId] = for {
+    bankDetail <- GoCardlessDDMandateUpdate.GetBankDetail(goCardless.client.wrapWith(JsonHttp.get))(gcMandateUpdateWithDetail.mandate.links.customer_bank_account)
+    newSfMandate <- SalesforceDDMandate.Create(sf.client.wrapWith(JsonHttp.post))(WireNewMandate(
```
pvighi: Here is another case that might be easier to test if you take  prewired GoCardlessDDMandateUpdate.GetBankDetail and SalesforceDDMandate.Create functions as params instead of the goCardless and SfClients

handlers/sf-gocardless-sync/src/main/scala/com/gu/sf_gocardless_sync/Handler.scala
```diff
@@ -0,0 +1,148 @@
...
+    ))
+    _ = existingSfMandates.put(gcMandateUpdateWithDetail.mandate.id, MandateLookupDetail(
+      Id = newSfMandate.Id,
+      GoCardless_Mandate_ID__c = gcMandateUpdateWithDetail.mandate.id,
+      Last_Mandate_Update__c = MandateUpdateSfId(""),
+      Status_Changed_At__c = UpdateHappenedAt("")
+    ))
+  } yield newSfMandate
+
+  def patchLastMandateUpdateOnMandate(
```
pvighi: I find this name a little confusing

handlers/sf-gocardless-sync/src/main/scala/com/gu/sf_gocardless_sync/Handler.scala
twrichards: oops meant to change that to logger before commit

handlers/sf-gocardless-sync/cfn.yaml
twrichards: removed

handlers/sf-gocardless-sync/cfn.yaml
twrichards: kept the timeout, but I have adjusted and commented

handlers/sf-gocardless-sync/src/main/scala/com/gu/sf_gocardless_sync/Handler.scala
twrichards: done

handlers/sf-gocardless-sync/src/main/scala/com/gu/sf_gocardless_sync/Handler.scala
twrichards: talked about this in person, summary: I would like to wait and see if this ever happens, if it can't process and event I don't want it to be dropped/lost so stopping the whole process until it is investigated is advantageous - if this happens ever/enough to become a burden will consider implementing max retry logic

handlers/sf-gocardless-sync/cfn.yaml
twrichards: thanks for the link @pvighi - had to delete the stack (and the existing log group automatically created by the lambda) but upon recreating the stack all worked nicely 😄

handlers/sf-gocardless-sync/src/main/scala/com/gu/sf_gocardless_sync/Handler.scala
twrichards: done

handlers/sf-gocardless-sync/src/main/scala/com/gu/sf_gocardless_sync/salesforce/SalesforceDDMandate.scala
twrichards: i know - i refactored out in https://github.com/guardian/zuora-auto-cancel/pull/212/commits/3f00f96bfd50061f91d88de93cd114297e41eda0 using a tail recursive function which builds up an _(immutable)_ map over time since doing the composite call up front for all mandates (as we discussed) was cumbersome and I realised the map also needs to take into account the changes done in the patch (not just the create/post)

handlers/sf-gocardless-sync/src/main/scala/com/gu/sf_gocardless_sync/Handler.scala
twrichards: changed

handlers/sf-gocardless-sync/src/main/scala/com/gu/sf_gocardless_sync/Handler.scala
twrichards: done now

handlers/sf-gocardless-sync/README.md
```diff
@@ -0,0 +1,48 @@
...
+
+#### SalesForce Objects
+- DD Mandates (`DD_Mandate__c`)
+- DD Mandate Events (`DD_Mandate_Event__c`)
+
+#### Errors
+There are a few log filters & alarms configured in [cfn.yml](cfn.yaml) 
+- "GenericError Count" filer & alarm, is fired if "HTTP request was unsuccessful" is found in the logs. **Fairly important
+ given this lambda is designed to stop processing events upon any error (to avoid missing data)**
+- TODO add metric events processed filter and mandate records created filter (with alarm if this drops below certain threshold per 12 hours)
```
jacobwinch: Minor: I would create a card for this stuff rather than having TODOs in the README

PR #213, opened 2018-11-14
Improve zuora/salesforce identity id backfill

QuarpT:
Work done:

- If an identity user has an unvalidated email and they have no password then we can use this identity id in the backfill
- statusFields optional in identity response - not all users have status fields

Why are we doing this:

Increase the number of users backfilled with existing accounts.

### DIFF
--- a/handlers/identity-backfill/src/main/scala/com/gu/identity/GetByEmail.scala
+++ b/handlers/identity-backfill/src/main/scala/com/gu/identity/GetByEmail.scala
@@ -20,7 +20,10 @@ object GetByEmail {
     case class StatusFields(userEmailValidated: Boolean)
     implicit val statusFieldsReads: Reads[StatusFields] = Json.reads[StatusFields]
 
-    case class User(id: String, statusFields: StatusFields)
+    case class User(id: String, statusFields: Option[StatusFields]) {
+      val isUserEmailValidated: Boolean = statusFields.exists(_.userEmailValidated)
+    }
+
     implicit val userReads: Reads[User] = Json.reads[User]
 
     case class UserResponse(status: String, user: User)
@@ -42,7 +45,7 @@ object GetByEmail {
   def wireToDomainModel(userResponse: UserResponse): ClientFailableOp[IdentityAccount] = {
     for {
       user <- userFromResponse(userResponse)
-      identityId = if (user.statusFields.userEmailValidated) IdentityAccountWithValidatedEmail(IdentityId(user.id)) else IdentityAccountWithUnvalidatedEmail(IdentityId(user.id))
+      identityId = if (user.isUserEmailValidated) IdentityAccountWithValidatedEmail(IdentityId(user.id)) else IdentityAccountWithUnvalidatedEmail(IdentityId(user.id))
     } yield identityId
 
   }
--- a/handlers/identity-backfill/src/main/scala/com/gu/identity/GetByEmail.scala
+++ b/handlers/identity-backfill/src/main/scala/com/gu/identity/GetByEmail.scala
@@ -13,7 +13,7 @@ object GetByEmail {
 
   sealed trait IdentityAccount
   case class IdentityAccountWithValidatedEmail(identityId: IdentityId) extends IdentityAccount
-  case object IdentityAccountWithUnvalidatedEmail extends IdentityAccount
+  case class IdentityAccountWithUnvalidatedEmail(identityId: IdentityId) extends IdentityAccount
 
   object RawWireModel {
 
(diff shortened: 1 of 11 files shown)


handlers/identity-backfill/src/main/scala/com/gu/identity/GetByEmail.scala
```diff
@@ -13,14 +13,17 @@ object GetByEmail {
 
   sealed trait IdentityAccount
   case class IdentityAccountWithValidatedEmail(identityId: IdentityId) extends IdentityAccount
-  case object IdentityAccountWithUnvalidatedEmail extends IdentityAccount
+  case class IdentityAccountWithUnvalidatedEmail(identityId: IdentityId) extends IdentityAccount
```
johnduffell: if the identity id exists on all possibilities, just move it up a level and make both of these into a case object (effectively becomes a boolean with a better name)

handlers/identity-backfill/src/main/scala/com/gu/identity/GetByEmail.scala
```diff
@@ -13,14 +13,17 @@ object GetByEmail {
...
 
   object RawWireModel {
 
     case class StatusFields(userEmailValidated: Boolean)
     implicit val statusFieldsReads: Reads[StatusFields] = Json.reads[StatusFields]
 
-    case class User(id: String, statusFields: StatusFields)
+    case class User(id: String, statusFields: Option[StatusFields]) {
+      val isUserEmailValidated: Boolean = statusFields.exists(_.userEmailValidated)
```
johnduffell: not a big difference, but I tried to move towards keeping the wire and domain models clean, and put all the conversion code in the separate wireToDomainModel function.   This does apply to varying degress in the older lambdas though!

handlers/identity-backfill/src/test/scala/com/gu/identity/GetByIdentityIdTest.scala
```diff
@@ -0,0 +1,44 @@
...
+        |    "status": "ok",
+        |    "user": {
+        |        "id": "1234",
+        |        "hasPassword": false
+        |    }
+        |}
+      """.stripMargin
+    val actual = GetByIdentityId.wrapper.toNewResponse(Json.parse(identityResponse))
+    actual should be(ClientSuccess(IdentityUser(IdentityId("1234"), hasPassword = false)))
+  }
```
johnduffell: great, nice clear tests!

handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/PreReqCheck.scala
```diff
@@ -20,24 +21,43 @@ object PreReqCheck {
...
     syncableSFToIdentity: SFContactId => LazyClientFailableOp[ApiGatewayOp[Unit]]
   )(emailAddress: EmailAddress): ApiGatewayOp[PreReqResult] = {
     for {
-      maybeExistingIdentityId <- (getByEmail(emailAddress) match {
-        case ClientSuccess(GetByEmail.IdentityAccountWithValidatedEmail(identityId)) => ContinueProcessing(Some(identityId))
-        case NotFound(_) => ContinueProcessing(None)
-        case ClientSuccess(IdentityAccountWithUnvalidatedEmail) => ReturnWithResponse(ApiGatewayResponse.notFound("identity email not validated"))
-        case other: ClientFailure => ReturnWithResponse(ApiGatewayResponse.internalServerError(other.toString))
-      }).withLogging("GetByEmail")
+      maybeExistingIdentityId <- findExistingIdentityId(getByEmail, getByIdentityId, emailAddress)
```
johnduffell: suggestion to make it simpler/easier to test:

currently getByEmail and getByIdentityId are passed into PreReqCheck and then directly onto findExistingIdentityId. This gives PreReqCheck extra parameters it doesn't need so harder to test.
My suggestion would be to partially apply findExistingIdentityId in the wiring stage i.e. Handler on line 72, and then pass it into PreReqCheck as something like `EmailAddress => ApiGatewayOp[Option[IdentityId]]`
This means they can be tested separately which should help and it avoids PreReqCheck becoming all knowing.

johnduffell:
looks great, the main comments would be moving up the identity id field a level and simplifying the parameters of PreReqCheck and testing the new function separately.  These are important because they are keeping the structure of the program simple as the code changes.

handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/PreReqCheck.scala
QuarpT: can do

handlers/identity-backfill/src/main/scala/com/gu/identity/GetByEmail.scala
johnduffell: yeah you're right.... boolean makes more sense than keeping the specially named trait

handlers/identity-backfill/src/test/scala/com/gu/identityBackfill/PreReqCheckTest.scala
```diff
@@ -150,4 +131,45 @@ class PreReqCheckTest extends FlatSpec with Matchers {
     PreReqCheck.acceptableReaderType(ClientSuccess(readerTypes)).toDisjunction.leftMap(_.statusCode) should be(scalaz.-\/("404"))
   }
 
+  "findExistingIdentityId" should "continue processing with identity id for existing validated account" in {
```
johnduffell: yep much better with the tests split up, thanks

johnduffell: my comment about the new function is there's no need to bundle these new tests in the same file as the existing ones.  It would be more decoupled to put findExistingIdentityId in its own file and same goes for the tests.

PR #214, opened 2018-11-14
Salesforce export state machine

pvighi:
This state machine runs salesforce bulk api queries to generate csv files for data lake raw tables.
This PR is bigger than I would have liked because it started out as a proof of concept. I people find it too hard to review I could spend some time splitting it into smaller PRS

Still left to do:

1. This state machine writes the results into a temp bucket in the membership account, once we get the permissions sorted it should write into a bucket into the ophan account when run in PROD
1.  When the state machine runs it doesn't clean up the destination bucket. This means that if the state machine runs more than once in a single day the etl job would pick up all results from the day and the clean table will end up with duplicate data
1. When we are ready we need to add some rules to schedule the step function to run once per table every day


lib/restHttp/src/main/scala/com/gu/util/resthttp/RestRequestMaker.scala
```diff
@@ -180,7 +183,10 @@ object RestRequestMaker extends Logging {
     createBodyFromString(bodyAsString)
   }
 
-  def createBodyFromString(bodyAsString: BodyAsString): RequestBody = {
-    RequestBody.create(MediaType.parse("application/json"), bodyAsString.value)
+  case class ContentType(value: String) extends AnyVal
```
pvighi: I needed to set the content type header in my requests but for some reason OkHttp ignores it, to set the content type it has to be specified like this So I had to add content type as a param here. 
I left it as json by default to avoid refactoring everything but I can remove the default value and fix it everywhere

lib/restHttp/src/main/scala/com/gu/util/resthttp/JsonHttp.scala
```diff
@@ -23,7 +23,7 @@ object JsonHttp {
     override def builder: Request.Builder = new Request.Builder().get()
   }
 
-  case class StringHttpRequest(requestMethod: RequestMethod, relativePath: RelativePath, urlParams: UrlParams)
+  case class StringHttpRequest(requestMethod: RequestMethod, relativePath: RelativePath, urlParams: UrlParams, headers: List[Header] = List.empty)
```
pvighi: I needed to specify custom headers for my sf requests. 
now that I look at it, it looks weird that there is UrlParams object but the headers are defined as a List[Header].
I'm not sure there are any benefits but I can make a Headers class just for consistency.
Also I wanted to avoid making this PR even bigger by adding a new mandatory param so I made it empty by default. I could change that right now or in a future PR

lib/effects/src/main/scala/com/gu/effects/Http.scala
```diff
@@ -14,6 +14,7 @@ object Http extends Logging {
       .readTimeout(15, TimeUnit.SECONDS)
       .build()
 
+    ///TODO this seems to return (int, String) probably should return int
```
pvighi: @johnduffell do you have any idea why this doesn't just return an Int ? can be just change it ?

johnduffell: judging by the name, it originally did.  But it's often useful in the logs to see the content of the request, especially if it's a small one.
I think it would be better to rename to what it actually does do, and maybe use a case class instead of a tuple if you prefer.

handlers/sf-datalake-export/cfn.yaml
```diff
@@ -0,0 +1,336 @@
...
+  StartJob:
+    Type: AWS::Lambda::Function
+    Properties:
+      Description: start a sf export job
+      FunctionName:
+        !Sub sf-start-export-job-${Stage}
+      Code:
+        S3Bucket: zuora-auto-cancel-dist
+        S3Key: !Sub membership/${Stage}/sf-datalake-export/sf-datalake-export.jar
+      Handler: com.gu.sf_datalake_export.StartJob::apply
```
johnduffell: one jar for them all is OK if they all have the same code by definition.   but I I think at the moment we have the word Handler in the class name of most of the existing handlers, which makes it more obvious looking at the code which things shouldn't be freely renamed.

handlers/sf-datalake-export/cfn.yaml
```diff
@@ -0,0 +1,336 @@
...
+  BatchStateLambda:
+    Type: AWS::Lambda::Function
+    Properties:
+      Description: get state of batches related to an salesforce bulk api query
+      FunctionName:
+        !Sub sf-get-batch-state-${Stage}
+      Code:
+        S3Bucket: zuora-auto-cancel-dist
+        S3Key: !Sub membership/${Stage}/sf-datalake-export/sf-datalake-export.jar
+      Handler: com.gu.sf_datalake_export.GetBatches::apply
```
johnduffell: BatchStateLambda is scala class GetBatches but the state is called getBatchState
might be worth tidying up the naming in general to be consistent

handlers/sf-datalake-export/src/main/scala/com/gu/sf_datalake_export/StartJob.scala
```diff
@@ -0,0 +1,93 @@
...
+  case class WireResponse(
+    jobId: String,
+    jobName: String
+  )
+
+  object WireResponse {
+    implicit val writes = Json.writes[WireResponse]
+  }
+
+  def sfQueryInfoFor(objectName: String): Try[SfQueryInfo] = BulkApiParams.all.find(_.objectName.value == objectName) match {
```
johnduffell: would be slightly more type safe to lift objectName into a specific type and compare those rather than pass it in as a String.  Then at least intellij will warn if we start comparing two different types of strings

handlers/sf-datalake-export/src/main/scala/com/gu/sf_datalake_export/StartJob.scala
```diff
@@ -0,0 +1,93 @@
...
+    case Some(sfQueryInfo) => Success(sfQueryInfo)
+    case None => Failure(LambdaException(s"invalid object name $objectName"))
+  }
```
johnduffell: there might be an option `toSuccess` function where you just pass in the exception to use if None if you think it would be clearer.

handlers/sf-datalake-export/src/main/scala/com/gu/sf_datalake_export/StartJob.scala
johnduffell: also have you considered making `all` a map rather than a list?  or the object names could be a sealed trait so that it's more type safe

pvighi: The object name here represents the string value in the wire request so this is the step where I'm trying to match the string input with the known object types.

In my first version I had the classic sealed trait with objects for all the types but in the end I just decided to just directly find a corresponding SfQueryInfo that has all the data I need about that salesforce object

handlers/sf-datalake-export/src/main/scala/com/gu/sf_datalake_export/salesforce_bulk_api/GetJobBatches.scala
```diff
@@ -0,0 +1,95 @@
...
+    }
+  }
+
+  def toRequest(jobId: JobId): StringHttpRequest = {
+    val relativePath = RelativePath(s"/services/async/44.0/job/${jobId.value}/batch")
+    StringHttpRequest(GetMethod, relativePath, UrlParams.empty)
+  }
+
+  def toResponse(bodyAsString: BodyAsString): ClientFailableOp[Seq[BatchInfo]] = {
+    val xml: Elem = scala.xml.XML.loadString(bodyAsString.value)
```
johnduffell: what happens if this loadString finds it isn't valid html? exception?
it's only one line anyway, but if it gets complicated and is needed in several operations, this function could just take an Elem instead of bodyAsString.  Then the xml stuff could be handled by some lower level layer.  This prevents cluttering the operations with having to deal with xml as strings and makes them more testable.

handlers/sf-datalake-export/src/main/scala/com/gu/sf_datalake_export/salesforce_bulk_api/S3UploadFile.scala
```diff
@@ -0,0 +1,38 @@
...
+object S3UploadFile extends Logging {
+
+  case class FileContent(value: String) extends AnyVal
+  case class FileName(value: String) extends AnyVal
+  case class File(fileName: FileName, content: FileContent)
+  def apply(
+    stage: Stage,
+    s3Write: PutObjectRequest => Try[PutObjectResult]
+  )(
+    file: File
```
johnduffell: this naming could be slightly misleading that this is actually just the data, not the java.io.File object.

handlers/sf-datalake-export/src/main/scala/com/gu/sf_datalake_export/salesforce_bulk_api/SfQueries.scala
```diff
@@ -0,0 +1,622 @@
...
+      |Latest_Payment_Type__c,
+      |Id,
+      |Recovery_Method__c,
+      |SF_Subscription__c,
+      |Subscription_Name__c,
+      |SystemModstamp
+      |from
+      |Payment_Failure__c
+    """.stripMargin
+}
```
johnduffell: can this be one per file and then another to group them together? as 622 lines is a bit much

handlers/sf-datalake-export/src/main/scala/com/gu/sf_datalake_export/StartJob.scala
johnduffell: yes, in fact I notice you are using wire requests and responses for the State even in the logic, and not generally lifting into a domain specific model.  If it were a sealed trait in the domain model, it could be turned into a failure at the stage where we lift it, and then in the domain steps it would just do a pattern match on the trait or whatever, which would be compile safe. 
At the moment the only worry is there's no compile safety at the higher level, we just have to hope it finds.  Maybe it's not worth the boiler plate though, but it could be cleaner.

johnduffell:
looks really good overall, I think this will be very useful.  It is a shame the salesforce api is so weird, but I think your way of modelling it makes sense and still keeps the easy to test qualities of the other salesforce operations.

johnduffell:
👍 thanks for this, defintitely make the logging make sense as discussed, also coverage is good and even codacy is happy

lib/restHttp/src/main/scala/com/gu/util/resthttp/RestRequestMaker.scala
```diff
@@ -61,10 +61,16 @@ object RestRequestMaker extends Logging {
...
-  case class PostRequest(body: JsValue, path: RelativePath, headers: List[Header])
+  case class PostRequest(body: JsValue, path: RelativePath)
+
   object PostRequest {
-    //TODO REMOVE DEFAULT EMPTY HEADERS?
-    def apply[REQ: Writes](body: REQ, path: RelativePath, headers: List[Header] = List.empty): PostRequest = new PostRequest(Json.toJson(body), path, headers)
+    def apply[REQ: Writes](body: REQ, path: RelativePath): PostRequest = new PostRequest(Json.toJson(body), path)
+  }
+
+  case class PostRequestWithHeaders(body: JsValue, path: RelativePath, headers: List[Header])
```
johnduffell: I prefer to not have headers in the object when they're not needed so I like this change, but I agree the duplication of the class seems repetitive.

 maybe we could have a 
`case class WithHeaders[RequestType](basicRequest: RequestType, headers: List[Header])`
 and not have to repeat the whole object each time?

Depends on your views on type parameters a bit though - it would be a more "FP way"

PR #215, opened 2018-11-19
sf datalake export improvements

pvighi:
- added some missing fields to contacts query
- added gdpr filters to queries
- upload to the ophan account when running in prod
- added a boolean input parameter "uploadToDataLake".  This param is true by default in prod and false in other stages. In fact setting it to true and running the state machine in anything other than PROD will return an error. This will allow us to run the state machine in prod without uploading anything to the data lake.
- also passed the object name along the state machine so that the downloader can use it to generate the destination bucket name

things still to do: 
- as we come to final agreement on the fields we need in the clean table, remove all the unused fields from the raw export queries
- add some error handling that always closes the sf export job even if the state machine fails

### DIFF
--- a/handlers/sf-datalake-export/cfn.yaml
+++ b/handlers/sf-datalake-export/cfn.yaml
@@ -11,10 +11,11 @@ Parameters:
     Default: CODE
 Mappings:
   StageMap:
-    CODE: #we should write into a bucket in the ophan account really
-      destBucketArn: "arn:aws:s3:::gu-salesforce-export-test/*"
+    CODE:
+      destBuckets: "arn:aws:s3:::gu-salesforce-export-test/CODE/*"
     PROD:
-      destBucketArn: "arn:aws:s3:::gu-salesforce-export-test/*"
+      destBuckets: ["arn:aws:s3:::gu-salesforce-export-test/PROD/*", "arn:aws:s3:::ophan-clean-salesforce-customer-data-*"]
+
 Resources:
   StartJobRole:
     Type: AWS::IAM::Role
@@ -143,7 +144,7 @@ Resources:
           - Effect: Allow
             Action: s3:GetObject
             Resource: !Sub "arn:aws:s3:::gu-reader-revenue-private/membership/support-service-lambdas/${Stage}/sfExportAuth-${Stage}*.json"
-      - PolicyName: DestBucket
+      - PolicyName: DestBuckets
         PolicyDocument:
           Statement:
           - Effect: Allow
@@ -157,7 +158,7 @@ Resources:
             - s3:PutObject
             - s3:GetObjectVersion
             - s3:DeleteObjectVersion
-            Resource: !FindInMap [StageMap, !Ref Stage, destBucketArn]
+            Resource: !FindInMap [StageMap, !Ref Stage, destBuckets]
   DownloadBatch:
     Type: AWS::Lambda::Function
     Properties:
--- a/handlers/sf-datalake-export/src/main/scala/com/gu/sf_datalake_export/handlers/DownloadBatchHandler.scala
+++ b/handlers/sf-datalake-export/src/main/scala/com/gu/sf_datalake_export/handlers/DownloadBatchHandler.scala
@@ -7,7 +7,8 @@ import com.amazonaws.services.s3.model.{PutObjectRequest, PutObjectResult}
 import com.gu.effects.{GetFromS3, RawEffects}
 import com.gu.salesforce.SalesforceAuthenticate.{SFAuthConfig, SFExportAuthConfig}
 import com.gu.salesforce.SalesforceClient
-import com.gu.sf_datalake_export.salesforce_bulk_api.BulkApiParams.{ObjectName, SfObjectName}
+import com.gu.sf_datalake_export.handlers.StartJobHandler.UploadToDataLake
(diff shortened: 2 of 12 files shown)


handlers/sf-datalake-export/src/main/scala/com/gu/sf_datalake_export/handlers/DownloadBatchHandler.scala
```diff
@@ -87,6 +87,11 @@ object DownloadBatchHandler {
     } yield ()
   }
 
+  def uploadBasePath(stage: Stage) = stage match {
+    case Stage("PROD") => BasePath(s"gu-salesforce-export-test/PROD/raw")
+    case Stage(stageName) => BasePath(s"gu-salesforce-export-test/$stageName/raw")
+  }
```
pvighi: this seems kinda pointless now but the final version should upload to the correct raw bucket in the ophan account, this is just in preparation for that

handlers/sf-datalake-export/src/main/scala/com/gu/sf_datalake_export/salesforce_bulk_api/SfQueries.scala
```diff
@@ -185,10 +188,10 @@ object SfQueries {
       |Zuora_Subscription_Salesforce_ID__c,
       |Zuora_Subscription_Name__c
       |from SF_Subscription__c
+      |where Buyer__r.Account.GDPR_Deletion_Pending__c = false
```
pvighi: @david-pepper  @abhichawla  can you confirm the conditions added to all the queries in this file are the correct way to filter out data for GDPR ?

handlers/sf-datalake-export/src/main/scala/com/gu/sf_datalake_export/salesforce_bulk_api/SfQueries.scala
```diff
@@ -490,8 +499,10 @@ object SfQueries {
       |SystemModstamp
       |from
       |Fulfilment_Process_Information__c
+      |where Contact__r.Account.GDPR_Deletion_Pending__c =false
     """.stripMargin
 
+  //TODO do we need to do gdpr exclusions on this ? how?
   //the object Imovo_Contract__c doesn't seem to exist
   val imovoContract =
```
pvighi: @david-pepper  @abhichawla  I could not find a way to do the GDPR exclusions in this one, do we even need to for this query ?

handlers/sf-datalake-export/src/main/scala/com/gu/sf_datalake_export/salesforce_bulk_api/SfQueries.scala
```diff
@@ -551,6 +563,7 @@ object SfQueries {
       |Zuora_ReferenceId__c
       |from
       |Payment_Card__c
+      |where Zuora_BillingAccount__r.ProcessingAdvice__c != 'DoNotProcess'
```
pvighi: @david-pepper  @abhichawla I could not find a way to link this to the salesforce account so I did the GDPR exclusion based on the Zuora field, is this correct/the best way to do it ?

handlers/sf-datalake-export/src/main/scala/com/gu/sf_datalake_export/salesforce_bulk_api/SfQueries.scala
abhichawla: Its correct.

handlers/sf-datalake-export/src/main/scala/com/gu/sf_datalake_export/salesforce_bulk_api/SfQueries.scala
abhichawla: select count(Id) from Imovo_Contract__c where  Contact__r.Account.GDPR_Deletion_Pending__c = false

handlers/sf-datalake-export/src/main/scala/com/gu/sf_datalake_export/salesforce_bulk_api/SfQueries.scala
abhichawla: select count(Id)  from Payment_Card__c  where AccountID__r.GDPR_Deletion_Pending__c = false

handlers/sf-datalake-export/cfn.yaml
```diff
@@ -11,10 +11,11 @@ Parameters:
     Default: CODE
 Mappings:
   StageMap:
-    CODE: #we should write into a bucket in the ophan account really
-      destBucketArn: "arn:aws:s3:::gu-salesforce-export-test/*"
+    CODE:
+      destBuckets: "arn:aws:s3:::gu-salesforce-export-test/CODE/*"
     PROD:
-      destBucketArn: "arn:aws:s3:::gu-salesforce-export-test/*"
+      destBuckets: ["arn:aws:s3:::gu-salesforce-export-test/PROD/*", "arn:aws:s3:::ophan-clean-salesforce-customer-data-*"]
```
pvighi: This bucket bucket will occasionally contain prod data from test executions of the prod state machine. Maybe we should define it in this cloudformation. Right now it is configured manually but we probably want to ensure it is encrypted, not public and has some retention policy

pvighi: I will add the bucket configuration to the cloudformation in another PR

PR #216, opened 2018-11-21
Csr acquisition annual contribution

pvighi:
added Annual contributions to new product api.
~~We should wait until salesforce can handle the new available item in the catalog response before merging this~~ the salesforce side has already been updated to support this

### DIFF
--- a/handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/email/contributions/SendConfirmationEmailContributions.scala
+++ b/handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/email/contributions/SendConfirmationEmailContributions.scala
@@ -14,7 +14,8 @@ import com.gu.util.reader.AsyncTypes._
 import com.gu.util.reader.Types.ApiGatewayOp.{ContinueProcessing, ReturnWithResponse}
 import com.gu.newproduct.api.addsubscription.Formatters._
 import com.gu.newproduct.api.addsubscription.zuora.GetAccount.SfContactId
-import com.gu.newproduct.api.productcatalog.AmountMinorUnits
+import com.gu.newproduct.api.productcatalog.PlanId.{AnnualContribution, MonthlyContribution}
+import com.gu.newproduct.api.productcatalog.{AmountMinorUnits, PlanId}
 
 import scala.concurrent.Future
 
@@ -26,7 +27,8 @@ object SendConfirmationEmailContributions extends Logging {
     paymentMethod: PaymentMethod,
     amountMinorUnits: AmountMinorUnits,
     firstPaymentDate: LocalDate,
-    billTo: BillToContact
+    billTo: BillToContact,
+    planId: PlanId
   )
 
   def apply(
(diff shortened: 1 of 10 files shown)


handlers/new-product-api/src/main/scala/com/gu/newproduct/api/productcatalog/Catalog.scala
```diff
@@ -13,7 +13,8 @@ case class Catalog(
   voucherSundayPlus: Plan,
   voucherEveryDayPlus: Plan,
   voucherSixDayPlus: Plan,
-  monthlyContribution: Plan
+  monthlyContribution: Plan,
+  annualContribution: Plan
```
johnduffell: I just had a thought that a case class which contains about 10 of `Plan` is not easy for the compiler to check.  in future consider giving each Plan its own type if it would help type safety.  So rather than all being ` Plan` there would be 
```scala
monthlyContribution: MonthlyContributionPlan,
  annualContribution: AnnualContributionPlan
```
etc etc

pvighi: makes sense, but we would probably have to make them extends some common trait or something so that we can still treat all plans the same when we want to

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/productcatalog/ZuoraIds.scala
```diff
@@ -12,7 +12,12 @@ object ZuoraIds {
 
   case class PlanAndCharge(productRatePlanId: ProductRatePlanId, productRatePlanChargeId: ProductRatePlanChargeId)
 
-  case class ContributionsZuoraIds(monthly: PlanAndCharge, annual: PlanAndCharge)
+  case class ContributionsZuoraIds(monthly: PlanAndCharge, annual: PlanAndCharge) {
+    val byApiPlanId: Map[PlanId, PlanAndCharge] = Map(
+      MonthlyContribution -> monthly,
+      AnnualContribution -> annual
```
johnduffell: given that this should include all products, it seems a shame we get no compile safety that we remembered them both, and we have to deal with an Option when it should never happen at run time.
Would it be possible to either add an extra layer in the PlanId sealed abstract class and pattern match that here instead of a map, or have a byApiPlanId that works at the level above and pattern matches on all of the valid PlanIds in the sealed abstract class

johnduffell:
looks good, just a couple of comments about type safety 🗝 , but overall nice work 👍

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/email/contributions/SendConfirmationEmailContributions.scala
```diff
@@ -60,6 +62,11 @@ object SendConfirmationEmailContributions extends Logging {
 
   def toContributionFields(currentDate: LocalDate, data: ContributionsEmailData): Option[ContributionFields] = {
 
+    val productId = data.planId match {
+      case AnnualContribution => "annual-contribution"
+      case MonthlyContribution => "monthly-contribution"
+      case other => other.name
+    }
```
pvighi: it's super annoying to have to do this but plan.name uses underscores instead of dashes

johnduffell: ah yes as discussed this is just because the passed in type that we're matching on is too general for what we actually know it is, if we add the extra layer in the abstract classes at some point, this will go away

PR #219, opened 2018-11-22
remove unused fields from the contacts raw table

pvighi:
Removed every field from the contacts table that is not used anywhere. We might need to add some of them back as the requirements for the clean tables are refined.


svillafe:
Relevant PR in datalake: https://github.com/guardian/ophan-data-lake/pull/3079

PR #220, opened 2018-11-22
Add lambda to be used for batch emails

lmath:
We need a lambda to receive a call from salesforce with information about card expiry emails to be sent. 

This adds a lambda behind an API gateway that requires an API key (that will eventually serve that purpose). Currently it just replies status: 200 message: hello world. This is in the spirit of keeping PRs small when possible. 

To follow:
- Parse + validate the request body 
- Add items to the contributions-thanks queue
- Deal with if the whole batch fails
- Deal with if the partial batch fails

### DIFF
--- a/handlers/batch-email-sender/README.md
+++ b/handlers/batch-email-sender/README.md
@@ -1,0 +1,53 @@
+# batch-email-sender
+There is work done in Salesforce that, on a schedule, generates information about emails that need to be sent. 
+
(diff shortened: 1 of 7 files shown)


handlers/batch-email-sender/README.md
```diff
@@ -0,0 +1,53 @@
+# batch-email-sender
```
johnduffell: nice readme 👍

johnduffell:
looks good 👍 nice idea splitting things out as much as possible

PR #221, opened 2018-11-26
Backfill identity ids for users with multiple zuora accounts

QuarpT:
Backfill identity ids for users with multiple zuora accounts.

The lambda would previously fail validation if an email did not have 1 associated zuora account.
Now, if there are multiple zuora accounts we check if the CRM IDs are identical. If they are, we backfill all the associated SF contacts and zuora accounts.

Why are we doing this?

Improve the number of SalesForce and Zuora accounts with Identity IDs

### DIFF
--- a/handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/IdentityBackfillSteps.scala
+++ b/handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/IdentityBackfillSteps.scala
@@ -49,6 +49,7 @@ object IdentityBackfillSteps extends Logging {
   def updateSalesforceAccounts(updateSalesforceAccounts: (SFContactId, IdentityId) => ApiGatewayOp[Unit])(sfContactIds: Set[SFContactId], identityId: IdentityId): ApiGatewayOp[Unit] = {
 
     val failures = sfContactIds
+      .toSeq
       .map(updateSalesforceAccounts(_, identityId))
       .collect {
         case returnWithResponse: ReturnWithResponse => returnWithResponse
@@ -64,6 +65,7 @@ object IdentityBackfillSteps extends Logging {
   def updateZuoraAccounts(updateZuoraIdentityId: (AccountId, IdentityId) => ClientFailableOp[Unit])(accountIds: Set[AccountId], identityId: IdentityId): ApiGatewayOp[Unit] = {
 
     val failures = accountIds
+      .toSeq
       .map(updateZuoraIdentityId(_, identityId))
       .collect {
         case clientFailure: ClientFailure => clientFailure
@@ -72,6 +74,6 @@ object IdentityBackfillSteps extends Logging {
     if (failures.isEmpty)
       ContinueProcessing(())
     else
-      ReturnWithResponse(ApiGatewayResponse.badRequest("updateZuoraAccounts mutiple errors: " + failures.map(_.message).mkString(", ")))
+      ReturnWithResponse(ApiGatewayResponse.badRequest("updateZuoraAccounts multiple errors: " + failures.map(_.message).mkString(", ")))
   }
 }
--- a/handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/IdentityBackfillSteps.scala
+++ b/handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/IdentityBackfillSteps.scala
@@ -10,7 +10,7 @@ import com.gu.util.apigateway.ApiGatewayResponse
 import com.gu.util.apigateway.ResponseModels.ApiResponse
 import com.gu.util.reader.Types.ApiGatewayOp._
 import com.gu.util.reader.Types._
-import com.gu.util.resthttp.Types.{ClientFailableOp, ClientSuccess}
+import com.gu.util.resthttp.Types.{ClientFailableOp, ClientFailure, ClientSuccess}
 
 object IdentityBackfillSteps extends Logging {
 
@@ -22,8 +22,8 @@ object IdentityBackfillSteps extends Logging {
   def apply(
     preReqCheck: EmailAddress => ApiGatewayOp[PreReqResult],
     createGuestAccount: EmailAddress => ClientFailableOp[IdentityId],
-    updateZuoraIdentityId: (AccountId, IdentityId) => ClientFailableOp[Unit],
-    updateSalesforceIdentityId: (SFContactId, IdentityId) => ApiGatewayOp[Unit]
+    updateZuoraAccounts: (Set[AccountId], IdentityId) => ApiGatewayOp[Unit],
+    updateSalesforceAccounts: (Set[SFContactId], IdentityId) => ApiGatewayOp[Unit]
   )(request: DomainRequest): ApiResponse = {
 
     (for {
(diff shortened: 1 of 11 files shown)


handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/IdentityBackfillSteps.scala
```diff
@@ -46,4 +46,34 @@ object IdentityBackfillSteps extends Logging {
     else
       ContinueProcessing(())
 
+  def updateSalesforceAccounts(updateSalesforceAccounts: (SFContactId, IdentityId) => ApiGatewayOp[Unit])(sfContactIds: Set[SFContactId], identityId: IdentityId): ApiGatewayOp[Unit] = {
```
johnduffell: might be worth moving these into objects called UpdateSalesforceAccounts etc with an apply method, just to make it clearer that the above apply function is a self contained and this is not being called directly by it, it's only connected via the wiring in the handler.
I just mention this because line 37 looks like it's calling this directly but missing the first parameter group.

johnduffell: also it's weird that update SF account returns an ApiGatewayOp rather than ClientFailableOp.  Given that these two functions below are the same purpose it would be good if it's easy enough to tweak the type and just call the same function.

handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/PreReqCheck.scala
```diff
@@ -9,52 +9,72 @@ import com.gu.salesforce.TypesForSFEffectsData.SFContactId
...
-      _ <- syncableSFToIdentity(zuoraAccountForEmail.sfContactId).value.toApiGatewayOp("load SF contact").flatMap(identity)
-    } yield PreReqResult(zuoraAccountForEmail.accountId, zuoraAccountForEmail.sfContactId, maybeExistingIdentityId)
+      _ <- syncableSFToIdentity(zuoraAccounts.map(_.sfContactId))
+    } yield PreReqResult(zuoraAccounts.map(_.accountId).toSet, zuoraAccounts.map(_.sfContactId).toSet, maybeExistingIdentityId)
+  }
+
+  def checkSfContactsSyncable(syncableSFToIdentity: SFContactId => ApiGatewayOp[Unit])(sFContactIds: List[SFContactId]): ApiGatewayOp[Unit] = {
+    sFContactIds
+      .map(syncableSFToIdentity)
+      .collectFirst {
```
johnduffell: give that this is a List rather than a Stream, we already checked them all, so we might as well collect them all and report everything

handlers/identity-backfill/src/test/scala/com/gu/identityBackfill/IdentityBackfillStepsTest.scala
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  "updateSalesforceAccounts" should "update salesforce accounts successfully" in {
+    val result = IdentityBackfillSteps.updateSalesforceAccounts(
+      (_, _) => ContinueProcessing(())
+    )(
+        Set(SFContactId("sfContactId1"), SFContactId("sfContactId2")), IdentityId("123")
+      )
+
+    result shouldBe ContinueProcessing(())
```
johnduffell: checking the result is not so useful  in these cases since we are relying on the side effects not the result.  I've previously got around this deficiency by dropping the parameters to the side effecting function into a var history: List, and asserting it at the end.
I'll admit it complicates the structure of the tests a lot, but without using higher order data types e.g. HttpOp it's hard to get around

johnduffell:
looks great, and you've done a fair few good bits of cleanup along the way which is always very helpful! 🏅 
I've made a few comments, the testing side effects one I think is an important one to address if possible, the rest I'll leave up to you whether to use them or keep them for future reference
👍

handlers/identity-backfill/src/test/scala/com/gu/identityBackfill/IdentityBackfillStepsTest.scala
QuarpT: Is there a reason we're not using mockito for this sort of thing?

handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/IdentityBackfillSteps.scala
QuarpT: 👍  I've update this

handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/PreReqCheck.scala
```diff
@@ -9,52 +9,77 @@ import com.gu.salesforce.TypesForSFEffectsData.SFContactId
...
-      _ <- (zuoraAccountsForIdentityId == 0)
-        .toApiGatewayContinueProcessing(ApiGatewayResponse.notFound("already used that identity id"))
+      _ <- (zuoraAccountsForIdentityId == 0).toApiGatewayContinueProcessing(ApiGatewayResponse.notFound("already used that identity id"))
     } yield ()
   }
 
-  def getSingleZuoraAccountForEmail(
-    getZuoraAccountsForEmail: ClientFailableOp[List[ZuoraAccountIdentitySFContact]]
-  ): ApiGatewayOp[ZuoraAccountIdentitySFContact] = {
+  def validateZuoraAccountsFound(getZuoraAccountsForEmail: ClientFailableOp[List[ZuoraAccountIdentitySFContact]])(emailAddress: EmailAddress): ApiGatewayOp[List[ZuoraAccountIdentitySFContact]] = {
```
pvighi: I know this does not originate in this PR, but I'm a little confused as to why this clientFailableOp is called getZuoraAccountsForEmail.
Isn't it more a response from a previous call for getting the zuora accounts ?

QuarpT: you're right. I will rename it

handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/PreReqCheck.scala
```diff
@@ -9,52 +9,77 @@ import com.gu.salesforce.TypesForSFEffectsData.SFContactId
...
+
+    def validateNotEmpty(zuoraAccountsForEmail: List[ZuoraAccountIdentitySFContact]): ApiGatewayOp[Unit] = zuoraAccountsForEmail match {
+      case Nil => ReturnWithResponse(ApiGatewayResponse.badRequest(s"no zuora accounts found for $emailAddress"))
+      case _ => ContinueProcessing(())
+    }
+
+    def validateOneCrmId(zuoraAccountsForEmail: List[ZuoraAccountIdentitySFContact]) = {
+      if (zuoraAccountsForEmail.headOption.forall(head => zuoraAccountsForEmail.forall(_.crmId == head.crmId)))
```
pvighi: maybe there a more readable way to write this, maybe something similar to this: 
```scala
      zuoraAccountsForEmail.map(_.crmId).distinct.size == 1
````

File: handlers/identity-backfill/src/test/scala/com/gu/identityBackfill/IdentityBackfillStepsTest.scala [hunk 6]
```diff
@@ -0,0 +1,55 @@
...
+import com.gu.salesforce.TypesForSFEffectsData.SFContactId
+import com.gu.util.apigateway.ResponseModels.ApiResponse
+import com.gu.util.reader.Types.ApiGatewayOp.{ContinueProcessing, ReturnWithResponse}
+import com.gu.util.resthttp.Types.{ClientSuccess, GenericError}
+import org.scalatest.{FlatSpec, Matchers}
+
+class IdentityBackfillStepsTest extends FlatSpec with Matchers {
+
+  it should "backfill identity ids successfully" in {
+    val ApiResponse(statusCode, _, _) = IdentityBackfillSteps.apply(
```
pvighi: maybe in all these unit tests would be easier  to read if the parameters were passed by name so we could see what is returning what. Feel free to ignore if you disagree

PR #222, opened 2018-11-26
Batch Email Sender lambda

lmath:
We need a lambda to receive a call from salesforce with information about card expiry emails to be sent.

https://github.com/guardian/zuora-auto-cancel/pull/220 added a hello world lambda behind an API gateway

This PR implements batch email sends, currently specifically for Card Expiry emails. 

So, we would expect to be able to make a request like 
`POST {api gateway url}/CODE/email-batch`
header:
`x-api-key:{api key}`

(get url and api key from aws console)

sample body:
```
[  
   {  
      "payload":{  
      	 "record_id": "abcd",
         "to_address":"tester-test@test.com",
         "subscriber_id":"A-S00044748",
         "sf_contact_id":"0036E00000KtDaHQAV",
         "product":"Digital Pack",
         "next_charge_date":"2018-09-03",
         "last_name":"bla",
         "identity_id":"30002177",
         "first_name":"digital pack",
         "email_stage":"MBv1 - 1"
      },
      "object_name":"Card_Expiry__c"
   }
]
```
And this will cause an item to be added to contributions-thanks-dev
(https://github.com/guardian/membership-workflow/pull/154), in the format expected by the queue

And so, tester-test@test.com will receive a card expiry email 

#### Failure
- If the request body is not parsed correctly, the API returns a 400
- If an item reaches the queue, but fails to send, it will end up in the dead letters queue
- If the request is parsed, but any (or all items) are not added to the queue, then the API returns a 502 and a list of the ids that were not added to the queue (see the README for sample response)
(See the README for more details)


handlers/batch-email-sender/src/main/scala/com/gu/batchemailsender/api/batchemail/model/EmailBatch.scala
```diff
@@ -0,0 +1,100 @@
...
+      identity_id: Option[String],
+      first_name: String,
+      email_stage: String
+    )
+
+    implicit val emailBatchItemPayloadReads = Json.reads[WireEmailBatchItemPayload]
+    implicit val emailBatchItemReads = Json.reads[WireEmailBatchItem]
+    //    implicit val emailBatch = Json.reads[WireEmailBatch]
+
+    implicit val emailBatchReader = new Reads[WireEmailBatch] {
```
lmath: Note: David is going to check if it's possible to send me a request body like

```
{
	batch_items: [  
   		{  
   		   "payload":{  
   		   	 "record_id": "abcd",
   		      "to_address":"leigh-anne.mathieson@theguardian.com",
   		      "subscriber_id":"A-S00044748",
   		      "sf_contact_id":"0036E00000KtDaHQAV",
   		      "product":"Digital Pack",
   		      "next_charge_date":"2018-09-03",
   		      "last_name":"bla",
   		      "identity_id":"30002177",
   		      "first_name":"digital pack",
   		      "email_stage":"MBv1 - 1"
   		   },
   		   "object_name":"Card_Expiry__c"
   		}
	]
}
```

because then I do not need this emailBatchReader, but if it's not possible or cumbersome then this is the best thing I can think of if the request is an unnamed list 😬

twrichards:
👍 looks good!
Great explanatory description.

twrichards:
looks like coveralls is punishing you for a 0.005% drop in coverage, might be worth adding a tiny bit more just to make it happy

PR #223, opened 2018-11-28
Remove old results when re running salesforce export job

pvighi:
- Added a new lambda to delete files old csv files for the same date as the current execution. This allows the salesforce export job to be run multiple times on the same date without creating duplicated data.
- made the test buckets separate for the different environments and added the definition to the cloudformation to ensure it is encrypted and it has a deletion policy
- refactored the downloader lambda a little bit ( this was a side effect of the cleaning functionality starting out as part of the downloader lambda before moving it into its own state machine step)


handlers/sf-datalake-export/cfn.yaml
```diff
@@ -12,11 +12,29 @@ Parameters:
...
+      BucketEncryption:
+        ServerSideEncryptionConfiguration:
+          - ServerSideEncryptionByDefault:
+              SSEAlgorithm: AES256
+      LifecycleConfiguration:
+        Rules:
+          - TagFilters:
+              - Key: housekeeping
+                Value: delete
+            ExpirationInDays: 2
```
pvighi: The production version of this bucket will only be used when the state machine is run in test mode.
 But still,  just in case we make sure to clear everything in these buckets every couple of days

handlers/sf-datalake-export/src/main/scala/com/gu/sf_datalake_export/handlers/CleanBucketHandler.scala
```diff
@@ -0,0 +1,83 @@
...
+
+object CleanBucketHandler extends Logging {
+
+  def apply(inputStream: InputStream, outputStream: OutputStream, context: Context): Unit = {
+
+    def wiredOperation(state: WireState): Try[WireState] = {
+      val objectName = ObjectName(state.objectName)
+      val jobName = JobName(state.jobName)
+      val shouldUploadToDataLake = ShouldUploadToDataLake(state.uploadToDataLake)
+      val wiredS3PathFor = ExportS3Path(RawEffects.stage) _
```
johnduffell: I'd rather not mix the raw effects together with the wiring, as even though we don't usually test the wiring, it's at least referentially transparent when called with referentially transparent effects.

handlers/sf-datalake-export/src/main/scala/com/gu/sf_datalake_export/handlers/DownloadBatchHandler.scala
```diff
@@ -55,125 +51,125 @@ object DownloadBatchHandler {
...
+    } yield WireState.fromState(response)
+
   }
 
-  def apply(inputStream: InputStream, outputStream: OutputStream, context: Context): Unit = {
-    JsonHandler(
+  def apply(inputStream: InputStream, outputStream: OutputStream, context: Context): Unit = JsonHandler(
       lambdaIO = LambdaIO(inputStream, outputStream, context),
-      operation = wireOperation(RawEffects.stage, GetFromS3.fetchString, RawEffects.response, RawEffects.s3Write)
+      operation = wiredOperation
```
johnduffell: same here, I don't think side effects should ever appear other than the top level handler, I would prefer to revert this line

handlers/sf-datalake-export/src/main/scala/com/gu/sf_datalake_export/handlers/CleanBucketHandler.scala
pvighi: Ok, I'll change it. But I don't think it gives us much benefit for the added complexity of having to maintain the list of raw effect dependencies if we are not going to test it.
 Maybe if the dependencies had a type instead of having to write the full function input and outputs it would be less annoying to maintain.

johnduffell: feel free to give them a type!  I would say that's the better fix if that's the problem.
I think the benefit of referential transparency is very high hence the line in the sand.  Also if we start moving them down one level, why not another, and we're increasing the amount of code that we can't reason about as easily.  Just because we don't test at present doesn't mean we should give up as there are structural benefits too.  I'm still not sure how we can make it testable though without using implicits or something similar.

lib/effects/src/main/scala/com/gu/effects/AwsS3.scala
```diff
@@ -46,7 +47,40 @@ object UploadToS3 extends Logging {
   }
 
 }
+case class S3Path(bucketName: BucketName, key: Option[Key])
```
johnduffell: it would be possible in theory to have effects tests for these, if you think it is useful (I think I added for the SQS queue)

johnduffell:
looks great 👍 it's easy to get a feel for what's going on overall purely by looking at the types

src/test/scala/com/gu/effects/S3EffectsTest.scala
```diff
@@ -0,0 +1,65 @@
...
+import java.io.{ByteArrayInputStream, InputStream}
+
+import com.amazonaws.services.s3.model.{CannedAccessControlList, ObjectMetadata, PutObjectRequest}
+import com.amazonaws.util.IOUtils
+import com.gu.test.EffectsTest
+import org.scalatest.{AsyncFlatSpec, BeforeAndAfterAll, Ignore, Matchers}
+
+import scala.util.Success
+//Todo this test should be in the effects project but we need to refactor to be able to access the effectsTest tag from there
+
```
pvighi: @johnduffell  I ended up just moving my effects test to the main project. This way at least it runs when the effects test are run.

src/test/scala/com/gu/effects/S3EffectsTest.scala
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  def initialiseTestBucket: Unit = {
+    //put test data in
+    put("S3EffectsTest/test-prefix-file1", "this is file1")
+    put("S3EffectsTest/test-prefix-file2", "this is file2")
+    put("S3EffectsTest/ignored-prefix-file3", "this is file3")
+  }
+
+  it should "list bucket and delete" taggedAs EffectsTest in {
```
pvighi: This test wouldn't really work if other people run the same effects test at the same time. I wonder if this is the first test to introduce that limitation or if other effects tests have the same issue

johnduffell: no there are quite a few with that issue.  Just another small way that effects tests are more expensive than unit tests!  If it's easy, you could randomise the name it uses, and set the retention of the bucket low.
For the SF/zuora ones we could get the test to create a suitable subscription first and then work on that, but we would gain isolation but add more calls so it would be a trade off.

PR #224, opened 2018-11-28
Return error message on 500s

QuarpT:
When running scripts it's hard to tell what's happened on an internal server error. 

I can't see any secrets returned on 500s.  Is there a reason for hiding the message? Endusers don't hit any of this project's endpoints directly do they?

Edit: I've changed this to return 400s with error reasons on api client errors in Identity backfill.


pvighi:
Some endpoints are hit directly by end users, digital-subscription-expiry comes to mind but there might be others as well.
Whether there are secrets or not in the response is hard to tell since the error message depends on exceptions that could be generated  in many different places.  

I don't think this is the commonly expected behaviour and since this project is the default location for our future lambda based apis I would prefer not to do that.
But that's just my opinion, 
@johnduffell @paulbrown1982 @twrichards  : what do you think?

PR #225, opened 2018-11-29
PR INTO BRANCH: add wait steps to DSL and local runner

johnduffell:
This is a PR into our branch for step DSL not master!

This PR adds WaitStep to the DSL.  It works when you run locally (i.e. waits) but the CompiledSteps doesn't model it yet.

@pvighi


lib/scala-to-cfn/src/main/scala/com/gu/steps/InterpJson.scala
```diff
@@ -58,4 +58,17 @@ object InterpJson {
...
 
+  implicit def canRunWaitStepJson[REST <: StepsAlg](
+    implicit
+    canRunRest: InterpJson[REST]
+  ): InterpJson[WaitStep[REST]] =
+    new InterpJson[WaitStep[REST]] {
+      override type FINAL = canRunRest.FINAL
+
+      override def run(step: WaitStep[REST]): (List[JsValue => Option[JsValue]], JsValue => Option[FINAL]) = {
+        canRunRest.run(step.rest) // FIXME actually remember we need a delay?
```
johnduffell: we need to add a delay object into the `List` at this point - this is a TODO.  wihtout that the cloudformation won't include the delay.

PR #227, opened 2018-12-03
Step dsl gaia

alexduf:
I pasted the cfn generated hereafter.

It seems to behave correctly although still generates slight diff when trying to apply the cloudformation.
~I'll make a release of gaia 0.1.0 so that we can compile against it~
Gaia 0.1.0 has been released

```yaml
---
AWSTemplateFormatVersion: "2010-09-09"
Description: "Auto generated Lambda state machine"
Resources:
  TheRole:
    Type: "AWS::IAM::Role"
    Properties:
      AssumeRolePolicyDocument:
        Version: "2012-10-17"
        Statement:
        - Effect: "Allow"
          Principal:
            Service: "lambda.amazonaws.com"
          Action: "sts:AssumeRole"
  TheStatesRole:
    Type: "AWS::IAM::Role"
    Properties:
      Path: "/"
      Policies:
      - PolicyDocument:
          Version: "2012-10-17"
          Statement:
          - Effect: "Allow"
            Action:
            - "lambda:InvokeFunction"
            Resource: "*"
        PolicyName: "StatesExecutionPolicy"
      AssumeRolePolicyDocument:
        Version: "2012-10-17"
        Statement:
        - Effect: "Allow"
          Principal:
            Service: "states.eu-west-1.amazonaws.com"
          Action: "sts:AssumeRole"
  TheStateMachine:
    Type: "AWS::StepFunctions::StateMachine"
    Properties:
      DefinitionString:
        Fn::Sub:
        - "{\"StartAt\":\"GeneratedLambda0\",\"States\":{\"GeneratedLambda0\":{\"\
          Type\":\"Task\",\"Resource\":\"${GeneratedLambda0}\",\"Next\":\"GeneratedLambda1\"\
          },\"GeneratedLambda1\":{\"Type\":\"Task\",\"Resource\":\"${GeneratedLambda1}\"\
          ,\"End\":true}}}"
        - GeneratedLambda0: "GeneratedLambda0"
          GeneratedLambda1: "GeneratedLambda1"
      RoleArn:
        Fn::GetAtt:
        - "TheStatesRole"
        - "Arn"
  GeneratedLambda0:
    Type: "AWS::Lambda::Function"
    Properties:
      Environment:
        Variables:
          LAMBDA_ID: "0"
      Role: "TheRole"
      Handler: "com.gu.steps.Handler::apply"
      Code:
        S3Bucket: "johnd-dist"
        S3Key: "step.jar"
      MemorySize: 1536
      Runtime: "java8"
      Timeout: 300
  GeneratedLambda1:
    Type: "AWS::Lambda::Function"
    Properties:
      Environment:
        Variables:
          LAMBDA_ID: "1"
      Role: "TheRole"
      Handler: "com.gu.steps.Handler::apply"
      Code:
        S3Bucket: "johnd-dist"
        S3Key: "step.jar"
      MemorySize: 1536
      Runtime: "java8"
      Timeout: 300
```

### DIFF
--- a/lib/scala-to-cfn/src/main/scala/com/gu/steps/CompiledSteps.scala
+++ b/lib/scala-to-cfn/src/main/scala/com/gu/steps/CompiledSteps.scala
@@ -55,7 +55,10 @@ object CompiledSteps {
         handler = handler,
         code = Code().withS3Bucket("johnd-dist").withS3Key("step.jar"),
         runtime = "java8"
-      ).withEnvironment(Environment(Some(Map(envVar -> lambdaId.value.toString))))
+      )
+        .withEnvironment(Environment(Some(Map(envVar -> lambdaId.value.toString))))
+        .withMemorySize(1536)
+        .withTimeout(300)
     }
 
     val role = Role(roleName, WireCFN.assumeLambda)
--- a/lib/scala-to-cfn/src/main/scala/com/gu/steps/CompiledSteps.scala
+++ b/lib/scala-to-cfn/src/main/scala/com/gu/steps/CompiledSteps.scala
@@ -1,8 +1,13 @@
 package com.gu.steps
 
-import com.gu.steps.WireCFN.{WireCode, WireEnvironment, WireFnGetAtt, WireFnSub, WireLambdaFunctionProperties, WireProps, WireResource, WireRoleProps, WireStateMachineProps}
+import com.gu.gaia.Template
+import com.gu.gaia.aws.iam.role.{Policy, Role}
+import com.gu.gaia.aws.lambda.function.{Code, Environment, Function}
+import com.gu.gaia.aws.stepfunctions.statemachine.StateMachine
+import com.gu.gaia.primitives.{Expression, GetAtt, Sub}
 import com.gu.steps.WireStateMachine.{WireState, smW}
 import play.api.libs.json._
+import com.gu.gaia.syntax._
 
 import scala.util.Try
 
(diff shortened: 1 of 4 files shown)


lib/scala-to-cfn/src/main/scala/com/gu/steps/CompiledSteps.scala
```diff
@@ -38,48 +43,45 @@ object CompiledSteps {
...
+    val labelToLogicalName: Map[String, Expression[String]] = lambdas.map(aaa => aaa.resourceName -> Expression(aaa.resourceName)).toMap
+    val subStateMachine = Sub(Json.stringify(Json.toJsObject(wireStateMachine)(smW)), labelToLogicalName)
+    val stateMachine = StateMachine(stateMachineName, subStateMachine, GetAtt(statesRoleName, "Arn"))
+
+    Template()
+      .withDescription("Auto generated Lambda state machine")
+      .withResource(role)
+      .withResource(roleState)
+      .withResource(stateMachine)
+      .withResources(lambdas)
```
johnduffell: great, all this looks neater syntax wise than before

lib/scala-to-cfn/src/main/scala/com/gu/steps/CompiledSteps.scala
```diff
@@ -38,48 +43,45 @@ object CompiledSteps {
...
       )
-      StepData(s"GeneratedLambda${lambdaId.value}", fn)
+        .withEnvironment(Environment(Some(Map(envVar -> lambdaId.value.toString))))
+        .withMemorySize(1536)
+        .withTimeout(300)
     }
-    val role = (roleName, WireResource("AWS::IAM::Role", WireRoleProps(WireCFN.assumeLambda, None, None)))
-    val roleState = (statesRoleName, WireResource("AWS::IAM::Role", WireRoleProps(WireCFN.assumeStatesExecution, Some("/"), Some(WireCFN.policyStatesExecution))))
+
+    val role = Role(roleName, WireCFN.assumeLambda)
```
johnduffell: is the role not modelled by gaia?

johnduffell:
thanks alex that's fab, glad to see so much strcutre is auto generated.  I guess the problem with  roles is they're a bit free form, like the state machines, so we would not really be able to type check them from the AWS descriptions.  we can always tidy up our state machine model in future, and maybe generate from the AWS documentation if that's even a thing!
👍 
also @pvighi , @jacobwinch and maybe @mbarton  would be interested in progress

lib/scala-to-cfn/src/main/scala/com/gu/steps/CompiledSteps.scala
alexduf: No unfortunately AWS only gives us a clue about the type (they call it JSON) so anything can be put in there as long as it can be serialised.
I plan to add a type later as a "special case" during code generation to try to remain as typesafe as possible, but right now we have to deal with `AnyRef`

PR #240, opened 2018-12-31
Handle payment errors

jacobwinch:
We are currently serving a 500 if there is a payment failure when attempting to add a new product to an existing subscriber's account. This PR enables us to serve a more appropriate status code, which allows clients to handle these errors more gracefully. It should also stop us from getting so many false alarms.

Note that I can't find any documentation about the 53000060 error code from Zuora; but log inspection shows that this code is consistently returned on payment error.

### DIFF
--- a/lib/handler/src/main/scala/com/gu/util/apigateway/ApiGatewayResponse.scala
+++ b/lib/handler/src/main/scala/com/gu/util/apigateway/ApiGatewayResponse.scala
@@ -74,6 +74,11 @@ object ApiGatewayResponse extends Logging {
     "Credentials are missing or invalid"
   )
 
+  val paymentRequired = messageResponse(
+    "402",
+    "Payment was declined"
+  )
+
   def forbidden(message: String) = messageResponse(
     "403",
     message
--- a/lib/restHttp/src/main/scala/com/gu/util/resthttp/Types.scala
+++ b/lib/restHttp/src/main/scala/com/gu/util/resthttp/Types.scala
@@ -16,6 +16,8 @@ object Types {
 
   case class GenericError(message: String) extends ClientFailure
 
+  case class PaymentError(message: String) extends ClientFailure
+
   case class ClientSuccess[A](value: A) extends ClientFailableOp[A] {
     val isFailure = false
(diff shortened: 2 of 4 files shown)


pvighi:
about identifying the error code, 
have you looked at this ?
https://knowledgecenter.zuora.com/DC_Developers/C_REST_API/Responses_and_Errors#Error_Code_Format
they encode data in the different digits of the error code for some reason

lib/zuora/src/main/scala/com/gu/util/zuora/ZuoraRestRequestMaker.scala
```diff
@@ -30,9 +30,10 @@ object ZuoraRestRequestMaker extends Logging {
         logger.error(s"Zuora rejected our call $bodyAsJson")
         if (reasons.exists(_.code.toString.endsWith("40")))
           NotFound("Received a 'not found' response from Zuora")
+        else if (reasons.exists(_.code == 53000060))
+          PaymentError(s"Received a payment error from Zuora: $reasons")
```
pvighi: If we find ourselves having to add more specific errors in this general zuora handler we might have to rethink how this works

lib/handler/src/main/scala/com/gu/util/apigateway/ApiGatewayResponse.scala
```diff
@@ -74,6 +74,11 @@ object ApiGatewayResponse extends Logging {
     "Credentials are missing or invalid"
   )
 
+  val paymentRequired = messageResponse(
+    "402",
+    "Payment was declined"
```
johnduffell: this probably shouldn't be in the handler as it's specific to that lambda (although this is obviously not a new problem!)

lib/restHttp/src/main/scala/com/gu/util/resthttp/Types.scala
```diff
@@ -16,6 +16,8 @@ object Types {
 
   case class GenericError(message: String) extends ClientFailure
 
+  case class PaymentError(message: String) extends ClientFailure
```
johnduffell: this is a bit worrying that the general rest client knows about payments now.  I don't think that should be the case.

PR #243, opened 2019-01-09
extend Zuora Catalog timeout

johnduffell:
The default timeout is 15 seconds for all http calls.

However this afternoon things are timing out... 
image

which is causing a
```
java.net.SocketTimeoutException: timeout
```
which eventually caused an alarm
image

This PR extends the timeout to 2 minutes for that call.

The failing effects tests are due to changes in the SF contact that they use, there is now a health card in the SX trello to resolve this - thanks @twrichards !

I also noticed a few formatting issues which also got resolved by the scalaformat so they are in too.

@jacobwinch I will ship this tomorrow rather than rushing it out today.

### DIFF
--- a/handlers/batch-email-sender/src/test/scala/com/gu/EmailBatchTest.scala
+++ b/handlers/batch-email-sender/src/test/scala/com/gu/EmailBatchTest.scala
@@ -58,7 +58,7 @@ class EmailBatchTest extends FlatSpec {
   }
 
   "EmailBatch.fromWire" should "throw a jsresult exception when a required field is missing" in {
-     val sampleBatch =
+    val sampleBatch =
       """
         |{
         |     "batch_items":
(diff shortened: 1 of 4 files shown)


johnduffell:
the alarm also has an issue as 10 minutes isn't a  valid period, so this needs fixing too: 
```
The metric for this alarm has a non-standard period (10 minutes). We have changed it to the closest standard period. Please examine carefully the Period, Statistic, Threshold and Consecutive Periods settings and adjust before saving this alarm. Click "Cancel" to retain the "10 minutes" period setting.
```

johnduffell:
I have manually edited the alarm to be >3 rather than >1 to stop it going off constantly overnight.  This needs checking and reverting once this is shipped.

jacobwinch:
It looks like this settled down shortly after your change was made:

image

There's a PR here which should tidy these alerts up: https://github.com/guardian/zuora-auto-cancel/pull/244

> the alarm also has an issue as 10 minutes isn't a valid period

This is a weird one because it will allow you to Cloudform using a 10 minute period (https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-cw-alarm.html#cfn-cloudwatch-alarms-period). However, if you subsequently try to edit the alarm in the UI the warning that you've quoted is displayed 🙃

PR #245, opened 2019-01-10
Remove 4XX alarm

jacobwinch:
4XX includes payment errors (402s), which should not trigger alerts.


johnduffell:
👍 makes sense, as these should be dealt with at the SF level

PR #250, opened 2019-01-29
do not put payment failure emails in the queue without an email address

pvighi:
The contributions-thanks dead letters queue is getting filled with payment failure messages from accounts with no email address. 
With this change the payment failure callout from zuora will receive a bad request response if the json payload has no email.


pvighi:
Another option would be to change membership workflow to ignore messages without an email address instead of retrying and eventually putting them in the dead letters queue. 
One thing that might be an issue with this approach is that it would be a general change that would affect pretty much all emails since they all use the same queue  and queue consumer

pvighi:
I realised this was a problem with auto-cancel as well and in that case this solution would not work because the auto cancelling still has to work, just the email part has to be skipped

PR #251, opened 2019-01-31
Handle no email

pvighi:
We are getting a lot of items in the contributions-thanks dead letters queue. This is caused by payment failures / auto cancel email attempts on accounts that have no email address.

This PR will stop the auto-cancel api from putting messages in the queue with no destination email address:

- /payment-failure will return a "bad request" response with a message saying that the request must have an email address
- /auto-cancel will cancel the subscription and skip the email sending step if there is no email address. A warn messages will be logged but the call will return a successful response.

The payment failure changes have been tested in UAT, I'm currently in the process of testing the auto-cancel changes as it is not as simple.

### DIFF
--- a/src/main/scala/com/gu/autoCancel/AutoCancelCallout.scala
+++ b/src/main/scala/com/gu/autoCancel/AutoCancelCallout.scala
@@ -1,11 +1,11 @@
 package com.gu.autoCancel
 
-import play.api.libs.json.Json
+import play.api.libs.json.{JsResult, JsValue, Json, Reads}
 
 case class AutoCancelCallout(
   accountId: String,
   autoPay: String,
-  email: String,
+  email: Option[String],
   firstName: String,
   lastName: String,
   paymentMethodType: String,
@@ -21,5 +21,17 @@ case class AutoCancelCallout(
 }
 
 object AutoCancelCallout {
-  implicit val jf = Json.reads[AutoCancelCallout]
+
+  implicit val NoneForEmptyStringEMailReads: Reads[AutoCancelCallout] = new Reads[AutoCancelCallout] {
+
+    val standardReads = Json.reads[AutoCancelCallout]
+
+    override def reads(json: JsValue): JsResult[AutoCancelCallout] = standardReads.reads(json).map {
+      parsedCallout =>
+        parsedCallout.email match {
+          case Some("") => parsedCallout.copy(email = None)
+          case _ => parsedCallout
+        }
+    }
+  }
 }
--- a/src/main/scala/com/gu/autoCancel/AutoCancelSteps.scala
+++ b/src/main/scala/com/gu/autoCancel/AutoCancelSteps.scala
@@ -7,9 +7,9 @@ import com.gu.util.Logging
 import com.gu.util.apigateway.ApiGatewayHandler.Operation
 import com.gu.util.apigateway.{ApiGatewayRequest, ApiGatewayResponse}
 import com.gu.util.email.{EmailId, EmailMessage}
-import com.gu.util.reader.Types.ApiGatewayOp.ContinueProcessing
 import com.gu.util.reader.Types._
 import play.api.libs.json._
+import scalaz.\/
 
 object AutoCancelSteps extends Logging {
 
@@ -28,7 +28,7 @@ object AutoCancelSteps extends Logging {
   def apply(
     autoCancel: AutoCancelRequest => ApiGatewayOp[Unit],
     autoCancelFilter: AutoCancelCallout => ApiGatewayOp[AutoCancelRequest],
-    sendEmailRegardingAccount: (String, PaymentFailureInformation => EmailMessage) => ApiGatewayOp[Unit]
+    sendEmailRegardingAccount: (String, PaymentFailureInformation => String \/ EmailMessage) => ApiGatewayOp[Unit]
   ): Operation = Operation.noHealthcheck({ apiGatewayRequest: ApiGatewayRequest =>
     (for {
       autoCancelCallout <- apiGatewayRequest.bodyAsCaseClass[AutoCancelCallout]()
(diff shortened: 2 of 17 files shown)


src/main/scala/com/gu/paymentFailure/ToMessage.scala
```diff
@@ -3,67 +3,77 @@ package com.gu.paymentFailure
...
-          card_type = callout.creditCardType,
-          card_expiry_date = callout.creditCardExpirationMonth + "/" + callout.creditCardExpirationYear,
-          first_name = callout.firstName,
-          last_name = callout.lastName,
-          primaryKey = InvoiceId(callout.invoiceId),
-          price = price(paymentFailureInformation.amount, callout.currency),
-          serviceStartDate = serviceDateFormat(paymentFailureInformation.serviceStartDate),
-          serviceEndDate = serviceDateFormat(paymentFailureInformation.serviceEndDate)
+  def apply(callout: AutoCancelCallout, paymentFailureInformation: PaymentFailureInformation, sendId: EmailId): String \/ EmailMessage =
+    \/-(EmailMessage(
```
pvighi: this should return a left if there is no email address.. will fix it shortly

src/main/scala/com/gu/autoCancel/AutoCancelCallout.scala
```diff
@@ -21,5 +21,17 @@ case class AutoCancelCallout(
 }
 
 object AutoCancelCallout {
-  implicit val jf = Json.reads[AutoCancelCallout]
+
+  implicit val NoneForEmptyStringEMailReads: Reads[AutoCancelCallout] = new Reads[AutoCancelCallout] {
+
+    val standardReads = Json.reads[AutoCancelCallout]
+
+    override def reads(json: JsValue): JsResult[AutoCancelCallout] = standardReads.reads(json).map {
```
pvighi: I'm not sure if there is simpler or clearer way to do this but zuora sends an empty string to represent a missing email so I needed to make sure those cases are mapped to None instead of Some("")

src/main/scala/com/gu/autoCancel/AutoCancelSteps.scala
```diff
@@ -28,22 +31,26 @@ object AutoCancelSteps extends Logging {
...
     (for {
       autoCancelCallout <- apiGatewayRequest.bodyAsCaseClass[AutoCancelCallout]()
       urlParams <- apiGatewayRequest.queryParamsAsCaseClass[AutoCancelUrlParams]()
       _ <- AutoCancelInputFilter(autoCancelCallout, onlyCancelDirectDebit = urlParams.onlyCancelDirectDebit)
       acRequest <- autoCancelFilter(autoCancelCallout).withLogging(s"auto-cancellation filter for ${autoCancelCallout.accountId}")
       _ <- autoCancel(acRequest).withLogging(s"auto-cancellation for ${autoCancelCallout.accountId}")
-      request <- makeRequest(autoCancelCallout)
-      _ <- sendEmailRegardingAccount(autoCancelCallout.accountId, request)
+      request = makeRequest(autoCancelCallout) _
+      _ <- toSuccessApiResponse(sendEmailRegardingAccount(autoCancelCallout.accountId, request))
```
pvighi: It looks like the current code will return a failure if the email sending step fails even if the sub was successfully cancelled. It is probably not a big deal as the second time it will probably fail as the subscription is already cancelled

src/main/scala/com/gu/paymentFailure/GetPaymentData.scala
```diff
@@ -3,17 +3,15 @@ package com.gu.paymentFailure
...
 import com.gu.util.zuora.ZuoraGetInvoiceTransactions.{InvoiceItem, InvoiceTransactionSummary, ItemisedInvoice}
+import scalaz.{-\/, \/}
 
 object GetPaymentData extends Logging {
 
   case class PaymentFailureInformation(subscriptionName: String, product: String, amount: Double, serviceStartDate: LocalDate, serviceEndDate: LocalDate)
 
-  //todo for now just return an option here but the error handling has to be refactored a little bit
-  def apply(message: String)(invoiceTransactionSummary: InvoiceTransactionSummary): ApiGatewayOp[PaymentFailureInformation] = {
+  def apply(message: String)(invoiceTransactionSummary: InvoiceTransactionSummary): String \/ PaymentFailureInformation = {
```
pvighi: A lot of the refactoring in this PR is changing from Api responses to something more general. In general I find it easier to work with the responses if I only convert them to api response before returning them to the client

src/main/scala/com/gu/util/email/EmailSendSteps.scala
```diff
@@ -93,13 +96,13 @@ trait EmailSqsSerialisation {
 }
 
 object EmailSendSteps extends EmailSqsSerialisation {
-  def apply(sqsSend: Payload => Try[Unit])(emailRequest: EmailMessage): ApiGatewayOp[Unit] = {
+  def apply(sqsSend: Payload => Try[Unit])(emailRequest: EmailMessage): ClientFailableOp[Unit] = {
```
pvighi: since this is used in more than one lambda, it's better to not decide here what api responses it should return. The lambdas can decide on their own what kind of http response an email failure should be

pvighi:
[Investigate problems sending emails (membership-workflow alarms)](https://trello.com/c/NPVysngb/59-investigate-problems-sending-emails-membership-workflow-alarms)

src/main/scala/com/gu/autoCancel/AutoCancelSteps.scala
jacobwinch: Yes, I think you're right...

src/main/scala/com/gu/autoCancel/AutoCancelSteps.scala
```diff
@@ -28,22 +31,26 @@ object AutoCancelSteps extends Logging {
...
     } yield ApiGatewayResponse.successfulExecution).apiResponse
   })
 
-  def makeRequest(autoCancelCallout: AutoCancelCallout): ApiGatewayOp[PaymentFailureInformation => EmailMessage] = {
-    ContinueProcessing({ pFI: PaymentFailureInformation => ToMessage(autoCancelCallout, pFI, EmailId.cancelledId) })
+  def makeRequest(autoCancelCallout: AutoCancelCallout)(pFI: PaymentFailureInformation): String \/ EmailMessage =
```
jacobwinch: Pedantic: I prefer to be more verbose with naming (I'd favour paymentFailureInformation over pFI).

pvighi: yeah I just left it as it was before but I can take this chance to rename it

src/main/scala/com/gu/autoCancel/AutoCancelSteps.scala
```diff
@@ -28,22 +31,26 @@ object AutoCancelSteps extends Logging {
...
+    ToMessage(autoCancelCallout, pFI, EmailId.cancelledId)
 
+  def toSuccessApiResponse(clientFailableOp: ClientFailableOp[Unit]): ContinueProcessing[Unit] = clientFailableOp match {
```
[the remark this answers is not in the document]
pvighi: Yeah naming is the worst. The problem with those two names is that the op is not attempted in this function, this takes a clientFailableOp that is already a success or a failure and converts it to a ContinueProcessing apiGatewayOp.
I named it to successApiResponse to help readers understand that in the context of this code this means always return success.
I can name it something like **LogErrorsAndContinueProcessing** if you think it's better

PR #252, opened 2019-02-01
Separate contributions and voucher code in new product api

pvighi:
There is some tech debt in new product api, The add subscription endpoint handler has all the code for wiring and running the steps for adding vouchers and contributions.
There is some more refactoring to be made but as a first improvement this PR removes the voucher and contribution specific code into its own objects without modifying it too much.

### DIFF
--- a/handlers/new-product-api/src/main/scala/com/gu/newproduct/api/QueueNames.scala
+++ b/handlers/new-product-api/src/main/scala/com/gu/newproduct/api/QueueNames.scala
@@ -1,0 +1,13 @@
+package com.gu.newproduct.api
+
+import com.gu.effects.sqs.AwsSQSSend.QueueName
+import com.gu.util.config.Stage
+
+case class EmailQueueNames(contributions: QueueName, voucher: QueueName)
+
+object EmailQueueNames {
+  def emailQueuesFor(stage: Stage) = stage match {
+    case Stage("PROD") | Stage("CODE") => EmailQueueNames(contributions = QueueName("contributions-thanks"), voucher = QueueName("subs-welcome-email"))
+    case _ => EmailQueueNames(contributions = QueueName("contributions-thanks-dev"), voucher = QueueName("subs-welcome-email-dev"))
+  }
+}
--- a/handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/AddContribution.scala
+++ b/handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/AddContribution.scala
@@ -32,7 +32,7 @@ import com.gu.util.resthttp.RestRequestMaker.Requests
 import com.gu.util.resthttp.Types.ClientFailableOp
 
 import scala.concurrent.Future
-object Contributions {
+object AddContribution {
   def steps(
     getPlanAndCharge: PlanId => Option[PlanAndCharge],
     getCustomerData: ZuoraAccountId => ApiGatewayOp[ContributionCustomerData],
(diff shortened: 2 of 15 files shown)


handlers/new-product-api/src/main/scala/com/gu/newproduct/api/QueueNames.scala
```diff
@@ -0,0 +1,13 @@
+package com.gu.newproduct.api
+
+import com.gu.effects.sqs.AwsSQSSend.QueueName
+import com.gu.util.config.Stage
+
+case class EmailQueueNames(contributions: QueueName, voucher: QueueName)
+
+object EmailQueueNames {
+  def emailQueuesFor(stage: Stage) = stage match {
+    case Stage("PROD") | Stage("CODE") => EmailQueueNames(contributions = QueueName("contributions-thanks"), voucher = QueueName("subs-welcome-email"))
```
paulbrown1982: We really should consider consolidating all of these into one queue next Engineering Day!

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/Handler.scala
```diff
@@ -188,30 +80,28 @@ object Steps {
...
+        catalog,
+        zuoraIds,
+        zuoraClient,
         isValidStartDateForPlan,
         createSubscription,
-        sendVoucherEmail
-      ) _
+        awsSQSSend,
+        queueNames,
+        currentDate
```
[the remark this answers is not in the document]
pvighi: good catch.
It looks like this dependency was left in by mistake and passed around a little bit but never actually used. I'll remove it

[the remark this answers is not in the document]
pvighi: done

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/QueueNames.scala
```diff
@@ -0,0 +1,13 @@
+package com.gu.newproduct.api
+
+import com.gu.effects.sqs.AwsSQSSend.QueueName
+import com.gu.util.config.Stage
+
+case class EmailQueueNames(contributions: QueueName, voucher: QueueName)
+
+object EmailQueueNames {
+  def emailQueuesFor(stage: Stage) = stage match {
```
mario-galic: if you name it `def apply`, then you can use shorter syntax `EmailQueueNames(stage)` instead of `EmailQueueNames.emailQueuesFor(stage)`

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/AddContribution.scala
```diff
@@ -0,0 +1,127 @@
...
+import com.gu.newproduct.api.productcatalog.{AmountMinorUnits, PlanId}
+import com.gu.util.apigateway.ApiGatewayResponse.internalServerError
+import com.gu.util.reader.AsyncTypes.{AsyncApiGatewayOp, _}
+import com.gu.util.reader.Types.{ApiGatewayOp, OptionOps}
+import com.gu.util.resthttp.RestRequestMaker.Requests
+import com.gu.util.resthttp.Types.ClientFailableOp
+
+import scala.concurrent.Future
+object AddContribution {
+  def steps(
```
mario-galic: Could this be made `private`? I can see it being used only by tests. Can the test be modified such that it does not require utility method being made public?

pvighi: it could be made package private and have the test in the same package or something like that
Tests need to access this directly to be able to replace all the dependencies with fakes

PR #253, opened 2019-02-01
fix CSR acquisitions amount validation bug

pvighi:
When annual contributions were added to the new product api we (I) forgot to modify the amount min and max validation code. This means that the same min and max amounts are enforced for monthly or annual contributions.
In this pr we apply the same limits that are enforced in the support frontend.

### DIFF
--- a/handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/validation/AmountLimits.scala
+++ b/handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/validation/AmountLimits.scala
@@ -2,16 +2,56 @@ package com.gu.newproduct.api.addsubscription.validation
 
 import com.gu.i18n.Currency
 import com.gu.i18n.Currency._
+import com.gu.newproduct.api.productcatalog.PlanId
+import com.gu.newproduct.api.productcatalog.PlanId.AnnualContribution
+
+case class ContributionLimits(monthly: AmountLimits, annual: AmountLimits)
 
 case class AmountLimits(min: Int, max: Int)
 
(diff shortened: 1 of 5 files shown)


handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/validation/AmountLimits.scala
```diff
@@ -2,16 +2,56 @@ package com.gu.newproduct.api.addsubscription.validation
...
+  )
+
+  val eur = ContributionLimits(
+    monthly = AmountLimits(min = 200, max = 16600),
+    annual = AmountLimits(min = 1000, max = 200000)
+  )
+
+  def limitsFor(planId: PlanId, currency: Currency): AmountLimits = {
+    val contributionLimits = contributionLimitsfor(currency)
+    if (planId == AnnualContribution) contributionLimits.annual else contributionLimits.monthly
```
pvighi: At this point I already know this plan id is either monthly or annual contribution. 
The nicer way to do this would be to define a trait just for contribution plan ids and pattern match here, but I will do that later as I have another [PR](https://github.com/guardian/zuora-auto-cancel/pull/252) where I started refactoring this and I want to minimise the conflicts

pvighi:
[CSR Aquisitions bug: Recurring contribution amount validation uses the same limits for monthly and Annual. It should be around 12x higher for annual](https://trello.com/c/79ZVAxKr/157-csr-aquisitions-bug-recurring-contribution-amount-validation-uses-the-same-limits-for-monthly-and-annual-it-should-be-around-12x)

PR #254, opened 2019-02-05
Payment failure Lambda: do not return status 400 if account on payment failure has no email address

pvighi:
in my previous [PR](https://github.com/guardian/zuora-auto-cancel/pull/251) I made the payment failure lambda return 400 for requests that did not provide a destination email address.
The problem is that there's a bunch of zuora accounts with no email address and this triggers our 4xx error alarm.

In this PR we log a warning and return status 200 with an informational message in case the email address is not provided, but still return error messages if the email could not be sent for any other reason


pvighi:
[(1/1) avoid 4xx alarm on zuora auto cancel](https://trello.com/c/Euk6mVPw/165-avoid-4xx-alarm-on-zuora-auto-cancel)

PR #258, opened 2019-02-08
fixed failing effects tests

pvighi:
some effects tests were failing because data in salesforce / zuora changed.
This PR fixes them

### DIFF
--- a/handlers/digital-subscription-expiry/src/test/scala/com/gu/digitalSubscriptionExpiry/DigitalSubscriptionExpiryHandlerEffectsTest.scala
+++ b/handlers/digital-subscription-expiry/src/test/scala/com/gu/digitalSubscriptionExpiry/DigitalSubscriptionExpiryHandlerEffectsTest.scala
@@ -38,7 +38,7 @@ class DigitalSubscriptionExpiryHandlerEffectsTest extends FlatSpec with Matchers
 
     val expectedResponse =
       """
-        |{"statusCode":"200","headers":{"Content-Type":"application/json"},"body":"{\n  \"expiry\" : {\n    \"expiryDate\" : \"2018-11-30\",\n    \"expiryType\" : \"sub\",\n    \"content\" : \"SevenDay\"\n  }\n}"}
+        |{"statusCode":"200","headers":{"Content-Type":"application/json"},"body":"{\n  \"expiry\" : {\n    \"expiryDate\" : \"2019-11-30\",\n    \"expiryType\" : \"sub\",\n    \"content\" : \"SevenDay\"\n  }\n}"}
         |""".stripMargin
     responseString jsonMatches expectedResponse
   }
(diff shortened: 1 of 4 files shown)


lib/salesforce/src/test/scala/com/gu/salesforce/dev/SFEffectsData.scala
```diff
@@ -6,7 +6,7 @@ object SFEffectsData {
   // this class represents a compile safe set of data that we (hope) is present in dev sf for the Effects tests.
 
   // this has the first name and last name, but no record type or country
-  val testContactHasNamePhoneOtherAddress = SFContactId("0036E00000NLzPkQAL")
+  val testContactHasNamePhoneOtherAddress = SFContactId("0036E00000WtQpJ")
```
pvighi: the easiest way to fix this was to create a new contact to replace the test one that was being used before

handlers/identity-backfill/src/test/scala/com/gu/identityBackfill/salesforce/getContact/GetSFContactSyncCheckFieldsEffectsTest.scala
```diff
@@ -23,7 +23,7 @@ class GetSFContactSyncCheckFieldsEffectsTest extends FlatSpec with Matchers {
       result <- GetSFContactSyncCheckFields(getOp).apply(SFEffectsData.testContactHasNamePhoneOtherAddress).value.toDisjunction
     } yield result
 
-    actual should be(\/-(ContactSyncCheckFields(None, "One", "Day", Some("Afghanistan"))))
+    actual should be(\/-(ContactSyncCheckFields(Some("01220000000VB52AAG"), "testerson", "test", Some("Afghanistan"))))
```
pvighi: the old test contact had no recordTypeId but the new one is a standard customer. You cannot unset this value from the UI. Maybe I could unset it from the api but its easier to just change the assertion here

handlers/digital-subscription-expiry/src/test/scala/com/gu/digitalSubscriptionExpiry/DigitalSubscriptionExpiryHandlerEffectsTest.scala
```diff
@@ -38,7 +38,7 @@ class DigitalSubscriptionExpiryHandlerEffectsTest extends FlatSpec with Matchers
 
     val expectedResponse =
       """
-        |{"statusCode":"200","headers":{"Content-Type":"application/json"},"body":"{\n  \"expiry\" : {\n    \"expiryDate\" : \"2018-11-30\",\n    \"expiryType\" : \"sub\",\n    \"content\" : \"SevenDay\"\n  }\n}"}
+        |{"statusCode":"200","headers":{"Content-Type":"application/json"},"body":"{\n  \"expiry\" : {\n    \"expiryDate\" : \"2019-11-30\",\n    \"expiryType\" : \"sub\",\n    \"content\" : \"SevenDay\"\n  }\n}"}
```
johnduffell: good to fix, but I wonder if that check is a little too strict!  It would be nice to not compare the year too

pvighi: but then we would miss out on the fun of updating this test every year

handlers/identity-backfill/src/test/scala/com/gu/identityBackfill/salesforce/getContact/GetSFContactSyncCheckFieldsEffectsTest.scala
johnduffell: I've a feeling this was actually a test contact that @twrichards was using for something else

johnduffell:
thanks for fixing those up! 👍 🌟 
Reminds us that although the effects tests give us a lot of confidence that we're not breaking things, they also can be easily broken.

PR #259, opened 2019-02-11
Fix another effects test

pvighi:
An effects test was using the real current date instead the fake one defined for the test.


johnduffell:
good spot, nice 👍 🚀

PR #278, opened 2019-03-06
remove duplication 

pvighi:
On the new product API contribution confirmation emails were sent using custom code, this PR removes the duplication so that they are sent with the same code used for the rest of the products.

~~This PR should be merged only after we merge https://github.com/guardian/zuora-auto-cancel/pull/275 as the changes were made on top of the changes in that PR~~


handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/email/contributions/ContributionFields.scala
```diff
@@ -1,23 +1,53 @@
...
+  val firstPaymentDateFormat = DateTimeFormatter.ofPattern("EEEE, d MMMM yyyy")
+
+  def productId(plan: Plan) = plan.id match {
+    case AnnualContribution => "annual-contribution"
+    case MonthlyContribution => "monthly-contribution"
+    case other => other.name
+  }
+
+  def paymentMethodFields(paymentMethod: PaymentMethod, firstPaymentDate: LocalDate) = paymentMethod match {
+    case DirectDebit(status, accountName, accountNumberMask, sortCode, mandateId) => Map(
```
paulbrown1982: If we are creating a GoCardless direct debit payment method in Zuora, then we must also capture the customer's address too.

pvighi: I think this is just replicating what the frontend does :
https://github.com/guardian/support-frontend/blob/master/support-workers/src/main/scala/com/gu/emailservices/ContributionEmailFields.scala

PR #279, opened 2019-03-07
[RFC - Do not merge] Enable backfill for gift buyers

paulbrown1982:
Will update description tomorrow.

### DIFF
--- a/handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/IdentityBackfillSteps.scala
+++ b/handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/IdentityBackfillSteps.scala
@@ -46,13 +46,13 @@ object IdentityBackfillSteps extends Logging {
     else
       ContinueProcessing(())
 
-  def updateAccountWithIdentityId[A](
-    updateAccountWithIdentityId: (A, IdentityId) => ClientFailableOp[Unit]
-  )(ids: Option[A], identityId: IdentityId): ApiGatewayOp[Unit] = {
+  def updateBuyersIdentityId(
+    updateSalesforceContactIdentityId: (SFContactId, IdentityId) => ClientFailableOp[Unit]
+  )(ids: Option[SFContactId], identityId: IdentityId): ApiGatewayOp[Unit] = {
 
     val failures = ids
       .toSeq
-      .map(updateAccountWithIdentityId(_, identityId))
+      .map(updateSalesforceContactIdentityId(_, identityId))
       .zip(ids.toSeq)
       .collect {
         case (clientFailure: ClientFailure, id) => (id -> clientFailure.message).toString
@@ -61,13 +61,13 @@ object IdentityBackfillSteps extends Logging {
     if (failures.isEmpty)
       ContinueProcessing(())
     else
-      ReturnWithResponse(ApiGatewayResponse.badRequest("updateAccountsWithIdentityId multiple errors: " + failures.mkString(", ")))
+      ReturnWithResponse(ApiGatewayResponse.badRequest("updateBuyersIdentityId multiple errors: " + failures.mkString(", ")))
 
   }
 
-  def updateAccountsWithIdentityId[A](
-    updateAccountsWithIdentityId: (A, IdentityId) => ClientFailableOp[Unit]
-  )(ids: Set[A], identityId: IdentityId): ApiGatewayOp[Unit] = {
+  def updateZuoraBillingAccountsIdentityId[AccountId](
+    updateAccountsWithIdentityId: (AccountId, IdentityId) => ClientFailableOp[Unit]
+  )(ids: Set[AccountId], identityId: IdentityId): ApiGatewayOp[Unit] = {
 
     val failures = ids
       .toSeq
--- a/handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/IdentityBackfillSteps.scala
+++ b/handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/IdentityBackfillSteps.scala
@@ -34,7 +34,7 @@ object IdentityBackfillSteps extends Logging {
         case None => createGuestAccount(request.emailAddress)
       }).toApiGatewayOp("create guest identity account")
       _ <- updateZuoraAccounts(preReq.zuoraAccountIds, requiredIdentityId)
-      _ <- updateSalesforceAccount(preReq.maybeBuyerSFContactId, requiredIdentityId)
+      _ <- updateSalesforceAccount(preReq.maybeBuyer, requiredIdentityId)
       // need to remember which ones we updated?
     } yield ApiGatewayResponse.successfulExecution).apiResponse
 
--- a/handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/IdentityBackfillSteps.scala
+++ b/handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/IdentityBackfillSteps.scala
@@ -65,7 +65,7 @@ object IdentityBackfillSteps extends Logging {
 
   }
 
-  def updateZuoraBillingAccountsIdentityId[AccountId](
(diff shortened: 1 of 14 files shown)


handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/salesforce/GetSFContactSyncCheckFields.scala
```diff
@@ -54,14 +79,11 @@ object SyncableSFToIdentity {
...
-  )(
-    sFContactId: SFContactId
-  ) =
-    for {
-      _ <- if (ContactSyncCheck(standardRecordType)(fields))
-        ContinueProcessing(())
-      else
-        ReturnWithResponse(ApiGatewayResponse.notFound("this sf contact can't be synced back to zuora/identity"))
-    } yield ()
+    fields: ContactsByAccountIdQueryResponse
```
johnduffell: if we are following the philosophy of minimal coupling, if we are only using the "records" field we shouldn't pass in the whole query response object as the other fields are not necessary (this would add dependencies that weren't needed, which would harm testability)
in this case, better to pass in List[GetSFContactSyncCheckFields.ContactSyncCheckFields]
```suggestion
    fields: List[ContactSyncCheckFields]
```

handlers/identity-backfill/src/test/scala/com/gu/identityBackfill/EndToEndHandlerTest.scala
```diff
@@ -30,8 +30,9 @@ class EndToEndHandlerTest extends FlatSpec with Matchers {
         |""".stripMargin
 
     responseString jsonMatches expectedResponse
+
     requests should be(List(
-      BasicRequest("GET", "/services/data/v43.0/sobjects/Contact/00110000011AABBAAB", ""),
+      BasicRequest("GET", "/services/data/v43.0/query?q=SELECT Id, RecordTypeId, LastName, FirstName, OtherCountry, Email FROM Contact WHERE AccountId = %27crmId%27".replace(" ", "%20"), ""),
```
johnduffell: nice way to make it more readable

handlers/identity-backfill/src/test/scala/com/gu/identityBackfill/salesforce/getContact/ContactSyncCheckTest.scala
```diff
@@ -2,57 +2,81 @@ package com.gu.identityBackfill.salesforce.getContact
...
 import com.gu.identityBackfill.salesforce.ContactSyncCheck
 import com.gu.identityBackfill.salesforce.ContactSyncCheck.RecordTypeId
-import com.gu.identityBackfill.salesforce.GetSFContactSyncCheckFields.ContactSyncCheckFields
+import com.gu.identityBackfill.salesforce.GetSFContactSyncCheckFields.{ContactSyncCheckFields, ContactsByAccountIdQueryResponse}
+import com.gu.salesforce.TypesForSFEffectsData.SFContactId
 import org.scalatest.{FlatSpec, Matchers}
 
 class ContactSyncCheckTest extends FlatSpec with Matchers {
 
   it should "valid is true" in {
```
johnduffell: these test descriptions probably should be updated (although to be fair, given how clear the structure of the tests are, it is pretty obvious what they actually do!)

handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/IdentityBackfillSteps.scala
```diff
@@ -46,6 +46,25 @@ object IdentityBackfillSteps extends Logging {
...
+      }
+
+    if (failures.isEmpty)
+      ContinueProcessing(())
+    else
+      ReturnWithResponse(ApiGatewayResponse.badRequest("updateAccountsWithIdentityId multiple errors: " + failures.mkString(", ")))
+
+  }
+
   def updateAccountsWithIdentityId[A](
```
johnduffell: actualy, this code was only made generic and separated out because we were doing the same with SF and Zuora.  Now that it's one for SF and one for zuora, it could be less necessary.

handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/salesforce/GetSFContactSyncCheckFields.scala
```diff
@@ -54,14 +79,11 @@ object SyncableSFToIdentity {
   def apply(
```
johnduffell: since this isn't just checking if it's syncable, it's also returning the contact id, I'd suggest renaming this everywhere to something like getSFBillingContactIfSyncable ? and similar on ContactSyncCheck

handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/salesforce/GetSFContactSyncCheckFields.scala
```diff
@@ -37,15 +54,23 @@ object ContactSyncCheck {
...
-    val countryIsValid = country.trim != "" &&
-      CountryGroup.byOptimisticCountryNameOrCode(country).isDefined
-    correctRecordType && hasFirstName && hasLastName && countryIsValid
+    queryResponse: GetSFContactSyncCheckFields.ContactsByAccountIdQueryResponse
+  ): Option[SFContactId] = {
+
+    def correctRecordType(fields: ContactSyncCheckFields) = fields.RecordTypeId.contains(standardRecordType.value)
+
+    queryResponse.records.filter(correctRecordType) match {
+      case List(contactSyncCheckFields) => {
```
johnduffell:
```suggestion
      case onlyBillingContactFields :: Nil => {
```

handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/salesforce/GetSFContactSyncCheckFields.scala
```diff
@@ -37,15 +54,23 @@ object ContactSyncCheck {
...
+        val hasFirstName = contactSyncCheckFields.FirstName.trim != ""
+        val hasLastName = contactSyncCheckFields.LastName.trim != ""
+        val email = contactSyncCheckFields.Email.getOrElse("").trim
+        val emailIsValid = email.length > 3 && email.contains("@")
+        val country = contactSyncCheckFields.OtherCountry.getOrElse("")
+        val countryIsValid = country.trim != "" && CountryGroup.byOptimisticCountryNameOrCode(country).isDefined
+        if (hasFirstName && hasLastName && emailIsValid && countryIsValid) Some(SFContactId(contactSyncCheckFields.Id)) else None
+      }
+      case _ => None
```
johnduffell: I realise it didn't have that before so this would be an improvement, but experience tells me that it might be useful (maybe in another PR) to know what the problem was and a readout of the contact in question?  So make this an Either rather than an Option.

johnduffell:
looks great, I've made a few small comments but it's good that you managed to get your head round it all.  It's interesting looking back on this codebase as I can see there are pros and cons.  I'm not sure how much of it was unexpected for you?  We did start using Single Abstract Method pattern later on rather than lots of anonymous functions, to make intellij more likely to be able to find the actual implementation from the call site, but I'm sure there are other areas of confusion.
coverage looks great, the lines that aren't covered don't look important.
Overall looks good to me! just try it one at a time and check the results carefully 👍

handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/salesforce/GetSFContactSyncCheckFields.scala
paulbrown1982: Agreed, that's a good idea. Will do that either in a new PR or once all other code review changes are complete.

paulbrown1982:
Also see: https://github.com/guardian/zuora-auto-cancel/pull/280

PR #280, opened 2019-03-08
Improved error messages

paulbrown1982:
- Improved error messages when an Identity account is not in a safe-enough state to assign to the CRM.
- Improved error messages which explain why a Salesforce Contact is not updatable - usually it's because we know it won't sync properly between Zuora and Identity.

### DIFF
--- a/handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/salesforce/GetSFContactSyncCheckFields.scala
+++ b/handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/salesforce/GetSFContactSyncCheckFields.scala
@@ -63,15 +63,13 @@ object ContactSyncCheck {
 
     contactSyncCheckFields.filter(isStandardContact) match {
       case fields :: Nil => {
-        val hasFirstName = fields.FirstName.trim != ""
-        val hasLastName = fields.LastName.trim != ""
         val email = fields.Email.getOrElse("")
         val emailIsValid = email.length > 3 && email.contains("@")
         val country = fields.OtherCountry.getOrElse("").trim
-        val countryIsValid = country != "" && CountryGroup.byOptimisticCountryNameOrCode(country).isDefined
-        if (!hasFirstName) {
+        val countryIsValid = country.nonEmpty && CountryGroup.byOptimisticCountryNameOrCode(country).isDefined
+        if (fields.FirstName.trim.isEmpty) {
           \/.left(s"Contact ${fields.Id} is not syncable - does not have a first name")
-        } else if (!hasLastName) {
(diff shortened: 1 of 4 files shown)


handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/salesforce/GetSFContactSyncCheckFields.scala
```diff
@@ -56,19 +57,36 @@ object ContactSyncCheck {
...
+      case fields :: Nil => {
+        val email = fields.Email.getOrElse("")
         val emailIsValid = email.length > 3 && email.contains("@")
-        val country = onlyBillingContactFields.OtherCountry.getOrElse("")
-        val countryIsValid = country.trim != "" && CountryGroup.byOptimisticCountryNameOrCode(country).isDefined
-        if (hasFirstName && hasLastName && emailIsValid && countryIsValid) Some(SFContactId(onlyBillingContactFields.Id)) else None
+        val country = fields.OtherCountry.getOrElse("").trim
+        val countryIsValid = country.nonEmpty && CountryGroup.byOptimisticCountryNameOrCode(country).isDefined
+        if (fields.FirstName.trim.isEmpty) {
+          \/.left(s"Contact ${fields.Id} is not syncable - does not have a first name")
```
johnduffell: great these messages will be really helpful

handlers/identity-backfill/src/test/scala/com/gu/identityBackfill/FindExistingIdentityIdTest.scala
```diff
@@ -35,7 +35,7 @@ class FindExistingIdentityIdTest extends FlatSpec with Matchers {
     FindExistingIdentityId(
       _ => ClientSuccess(IdentityAccount(IdentityId("100"), isUserEmailValidated = false)),
       _ => ClientSuccess(IdentityUser(IdentityId("100"), hasPassword = true))
-    )(EmailAddress("email@email.email")) should be(ReturnWithResponse(ApiGatewayResponse.notFound(s"identity email not validated but password is set IdentityId(100)")))
+    )(EmailAddress("email@email.email")) should be(ReturnWithResponse(ApiGatewayResponse.notFound(s"Identity account not validated but password is set: 100")))
```
johnduffell: It's not something you just introduced, but I would say these are "over testing"
normally I would ignore the message, as having a check for the message itself seems to have very little effect on the accuracy of the message.  Also it discourages updating the messages to be more helpful, as you have to fix up loads of tests!
What I normally do is map the ApiGatewayResponse to an either of response code, and assert just that.

However as I say it's not something you added, so feel free to keep it in mind for future.

PR #284, opened 2019-03-25
Add braze to salesforce file upload

mario-galic:
## How does it work?

1. Braze [writes](https://www.braze.com/docs/developer_guide/rest_api/export/#users-by-segment-endpoint) a zipped CSV file to S3 bucket `braze-to-salesforce-file-upload` in `membership` AWS account
1. Lambda `braze-to-salesforce-file-upload` is triggered by an event message which contains the filenames
1. Unzip to get raw CSV file
1. Salesforce REST API uploads CSV to Salesforce Documents
1. Delete the file from S3 bucket after successful upload to Salesforce

## How do we know when it fails?

Cloudwatch Alert email is sent to SX mailing list.

## How to retry?

After successful upload to Salesforce, files are deleted from S3 bucket. Therefore, to retry upload:

1. check what files are still in the S3 bucket,
1. Download these files locally
1. Re-upload them to S3 bucket

### DIFF
--- a/handlers/braze-to-salesforce-file-upload/cfn.yaml
+++ b/handlers/braze-to-salesforce-file-upload/cfn.yaml
@@ -1,0 +1,81 @@
+Description: When braze uploads file to S3 bucket, trigger a Lambda event which uploads file to Salesforce
+
+Parameters:
+  BucketName:
+    Description: S3 Bucket name
+    Type: String
+
+  Stage:
+    Description: Stage name
+    Type: String
+    AllowedValues:
+      - CODE
+      - PROD
+    Default: CODE
+
+Resources:
+  Bucket:
+    Type: AWS::S3::Bucket
+    DependsOn: BucketPermission
+    Properties:
+      BucketName: !Ref BucketName
+      NotificationConfiguration:
+        LambdaConfigurations:
+          - Event: 's3:ObjectCreated:*'
+            Function: !GetAtt BucketWatcher.Arn
+
+  BucketPermission:
+    Type: AWS::Lambda::Permission
+    Properties:
+      Action: 'lambda:InvokeFunction'
+      FunctionName: !Ref BucketWatcher
+      Principal: s3.amazonaws.com
+      SourceAccount: !Ref "AWS::AccountId"
+      SourceArn: !Sub "arn:aws:s3:::${BucketName}"
+
+  BucketWatcher:
+    Type: AWS::Lambda::Function
+    Properties:
+      Description: Upload file to Salesforce
+      FunctionName: !Sub braze-to-salesforce-file-upload-${Stage}
+      Code:
+        S3Bucket: zuora-auto-cancel-dist
+        S3Key: !Sub membership/${Stage}/braze-to-salesforce-file-upload/braze-to-salesforce-file-upload.jar
+      Handler: com.gu.salesforce.braze.upload.UploadFileToSalesforceLambda::handle
+      Environment:
+        Variables:
+          Stage: !Ref Stage
+      Role: !GetAtt LambdaExecutionRole.Arn
+      Runtime: java8
+      MemorySize: 1536
+      Timeout: 60
+
+  LambdaExecutionRole:
+    Type: AWS::IAM::Role
+    Properties:
+      AssumeRolePolicyDocument:
+        Version: '2012-10-17'
+        Statement:
+          - Effect: Allow
+            Principal: {Service: [lambda.amazonaws.com]}
+            Action: ['sts:AssumeRole']
+      Path: /
+      ManagedPolicyArns:
+        - "arn:aws:iam::aws:policy/service-role/AWSLambdaBasicExecutionRole"
+      Policies:
+        - PolicyName: S3Policy
+          PolicyDocument:
+            Version: '2012-10-17'
+            Statement:
+              - Effect: Allow
+                Action:
+                  - s3:GetObject
+                Resource: !Sub "arn:aws:s3:::${BucketName}/*"
+
+        - PolicyName: ReadPrivateCredentials
+          PolicyDocument:
+            Statement:
+              - Effect: Allow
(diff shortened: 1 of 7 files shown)


mario-galic:
[*** GOAL - Spike: decide on price rise letter selections technical approach ***](https://trello.com/c/vayZEanm/284-goal-spike-decide-on-price-rise-letter-selections-technical-approach)

mario-galic:
[*** GOAL Clean all Objects in RAW ***](https://trello.com/c/zSKEj01X/286-goal-clean-all-objects-in-raw)

handlers/braze-to-salesforce-file-upload/cfn.yaml
```diff
@@ -0,0 +1,121 @@
...
+    Condition: IsProd
+    Properties:
+      AlarmName: braze-to-salesforce-file-upload
+      AlarmDescription: >
+        Failed to upload file from braze to Salesforce.
+        Latcham will not be able to send physical letters.
+        Refer to https://github.com/guardian/zuora-auto-cancel/blob/master/handlers/braze-to-salesforce-file-upload/README.md
+        on how to debug and retry.
+      AlarmActions:
+        - !Sub arn:aws:sns:${AWS::Region}:${AWS::AccountId}:MarioTest
```
jacobwinch: This should be changed before merging

handlers/braze-to-salesforce-file-upload/src/main/scala/com/gu/salesforce/braze/upload/UploadFileToSalesforceLambda.scala
```diff
@@ -0,0 +1,214 @@
...
+object Postconditions {
+  def apply(event: Event) = {
+    assert(S3BucketIsEmpty(), "Bucket should be empty after upload to Salesforce")
+    SuccessResponse(event)
+  }
+}
+
+object Preconditions {
+  def apply(event: Event): Any = {
+    assert(event.Records.forall(_.eventName == "ObjectCreated:Put"), "Only creation events should be handled")
```
jacobwinch: Do we want to add any basic checks on the file itself (e.g. correct format, not less than x number of rows etc.)? Of course this could be added in a separate PR if you think it's worth doing...

handlers/braze-to-salesforce-file-upload/cfn.yaml
mario-galic: I intended to change it after we see it working in prod for a while, so I do not spam.

handlers/braze-to-salesforce-file-upload/src/main/scala/com/gu/salesforce/braze/upload/UploadFileToSalesforceLambda.scala
```diff
@@ -0,0 +1,214 @@
...
+object S3BucketIsEmpty {
+  def apply(): Boolean =
+    AmazonS3Client.builder.build().listObjects(BucketName()).getObjectSummaries.size() == 0
+}
+
+object SalesforceDocumentFolderId {
+  def apply(): String = {
+    System.getenv("Stage") match {
+      case "CODE" => "00l25000000FITF" // marioTest
+      case "PROD" => "00l0J000002DA3K" // FIXME: currently marioTest, so use `Guardian Weekly Price Rise Letters 2019` b00l0J000002DA22
```
jacobwinch: Again, I guess we can update this before merging?

handlers/braze-to-salesforce-file-upload/src/main/scala/com/gu/salesforce/braze/upload/UploadFileToSalesforceLambda.scala
```diff
@@ -0,0 +1,214 @@
...
+
+object S3BucketIsEmpty {
+  def apply(): Boolean =
+    AmazonS3Client.builder.build().listObjects(BucketName()).getObjectSummaries.size() == 0
+}
+
+object SalesforceDocumentFolderId {
+  def apply(): String = {
+    System.getenv("Stage") match {
+      case "CODE" => "00l25000000FITF" // marioTest
```
jacobwinch: I'd create a folder called 'Guardian Weekly Price Rise Letters 2019' to make it obvious for other developers who want to test / debug this in CODE

handlers/braze-to-salesforce-file-upload/src/main/scala/com/gu/salesforce/braze/upload/UploadFileToSalesforceLambda.scala
mario-galic: There already exists folder [Guardian Weekly Price Rise Letters 2019](https://eu7.salesforce.com/015?fcf=00l0J000002DA22) but was thinking of migrating to it in separate PR after testing in production.

handlers/braze-to-salesforce-file-upload/cfn.yaml
jacobwinch: Ah, fair enough 👍for that approach

handlers/braze-to-salesforce-file-upload/src/main/scala/com/gu/salesforce/braze/upload/UploadFileToSalesforceLambda.scala
mario-galic: We should indeed do such checks. In fact, we should assume no trust in file because how do we know the new braze segmentation process is correct. However, since there are multiple different files with multiple different formats, and with multiple mixed concerns, then preconditions explode in size. When faced with such complexity, my own rule of thumb is, what is the single most critical path to check? In this case, it seems to be, the consent's column `dm_consent`, that is, not violating GDPR. It seems this column is always at index 12 (counting from 0) in all files, so I will check consent is populated for all rows.

PR #286, opened 2019-04-01
Contact merge: Make last name check a little bit more permissive

paulbrown1982:
A fair few contacts are not merging because their last names have whitespace or dots and commas 
on one contact and not the other. This makes the check a little bit more permissive, whilst still ensuring a sensible check is performed.

cc @abhichawla


johnduffell:
looks good, good spot updating the test as well 👍

PR #287, opened 2019-04-15
Remove "Zuora account balance is not zero" validation step

twrichards:
Whilst Home Delivery acquisitions are disabled in `support-frontend` it's increasingly likely CSRs will need to create new Home Delivery subs in SalesForce (i.e. via `new-product-api` in this repo). However currently there is a validation requirement for the account balance to be zero, which again is increasingly likely not to be the case for the time being, so removing this validation.

The corresponding changes to SalesForce (to WARN the CSR if non-zero) look like this...
#### Zero Account Balance
image

#### NON-Zero Account Balance
image

### DIFF
--- a/handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/validation/ValidateAccount.scala
+++ b/handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/validation/ValidateAccount.scala
@@ -17,7 +17,6 @@ object ValidateAccount {
   def apply(account: Account): ValidationResult[ValidatedAccount] = {
     for {
       _ <- account.autoPay.value orFailWith "Zuora account has autopay disabled"
-      _ <- (account.accountBalanceMinorUnits.value == 0) orFailWith "Zuora account balance is not zero"
       paymentMethodId <- account.paymentMethodId getOrFailWith "Zuora account has no default payment method"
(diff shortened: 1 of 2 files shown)


handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/validation/ValidateAccount.scala
```diff
@@ -17,7 +17,6 @@ object ValidateAccount {
   def apply(account: Account): ValidationResult[ValidatedAccount] = {
     for {
       _ <- account.autoPay.value orFailWith "Zuora account has autopay disabled"
-      _ <- (account.accountBalanceMinorUnits.value == 0) orFailWith "Zuora account balance is not zero"
```
pvighi: I think this will disable this validation for all products, not just home delivery. I'm not sure if that is a problem but I though I would mention it just in case

twrichards: good point, although I was aware - i will confirm with Matt that is what he wants before deploying

twrichards: confirmed with Matt, he wants it off for all products

PR #288, opened 2019-05-10
RENAME from 'zuora-auto-cancel' to 'support-service-lambdas'

twrichards:
A very long overdue task of renaming this repo (and everything else) from `zuora-auto-cancel` to `support-service-lambdas` to reflect the broader nature of this mono-repo.

I've created a new S3 bucket entitled `support-service-lambdas-dist` (which much of the config in this PR now references) and `sync`ed the content from the existing `zuora-auto-cancel-dist` (which I will delete at a later date).

Also updated the config in TeamCity as best I could.


twrichards:
@johnduffell I might have been a bit trigger happy with the find & replace, the thing that is actually the original _'zuora auto cancel'_ process perhaps shouldn't have been renamed - would really appreciate your input...

cloudformation.yaml
```diff
@@ -75,10 +75,10 @@ Resources:
         Properties:
             Description: Handles auto-cancellations for membership and subscriptions
             FunctionName:
-                !Sub zuora-auto-cancel-${Stage}
+                !Sub support-service-lambdas-${Stage}
```
pvighi: this is the actual auto cancel lambda isn't it ? it should probably keep its name

cloudformation.yaml
```diff
@@ -263,7 +263,7 @@ Resources:
     ZuoraAutoCancelAPIDeployment:
         Type: AWS::ApiGateway::Deployment
         Properties:
-            Description: Deploys zuora-auto-cancel-api into an environment/stage
+            Description: Deploys support-service-lambdas-api into an environment/stage
```
pvighi: I think this should keep the old name

cloudformation.yaml
```diff
@@ -254,7 +254,7 @@ Resources:
     ZuoraAutoCancelAPIStage:
         Type: AWS::ApiGateway::Stage
         Properties:
-            Description: Stage for zuora-auto-cancel-api
+            Description: Stage for support-service-lambdas-api
```
pvighi: I think this should keep the old description

pvighi:
🍾 I thought I would not to see this refactor in my lifetime.

I don't know if this would change the api gateway urls but if it does make sure to update Zuora, Salesforce, Fastly or whatever is using them

johnduffell:
yes the original auto cancel lambda should probably keep its name

cloudformation.yaml
johnduffell: yes, there's actually a separate tech debt task to get rid of the root level cfn/lambdas and put them all in the separated format.  However it's complicated by the fact that this lambda is in an api gateway with 2 other non cancellation lambdas.  I agree keep the name for the function, but renaming the Jar and the API gateway is sensible enough, with the caveats for care on the URL changing.  These URLs don't (yet) have a domain, so it's in zuora or SF or Stripe as a long unfriendly URL.

johnduffell:
I think patricio has covered it all well, so I concur with him! Thanks for doing this, hope the build stuff is as easy as search and replace too! 👍 ship it!

PR #289, opened 2019-05-10
Create initial version of Guardian Weekly holiday-stop-processor lambda

kelvin-chappell:
This is a lambda that covers the simplest possible case and just outputs the response from the subscription update if it's successful, or an error message if there's a failure.

I still have to get some environment variables from S3.

I've included a [standalone app](handlers/zuora-gw-holiday-stop/src/main/scala/com/gu/zuoragwholidaystop/StandaloneApp.scala) to make it easy to test end to end. 

The [lambda](handlers/zuora-gw-holiday-stop/src/main/scala/com/gu/zuoragwholidaystop/Handler.scala) and the standalone app are thin wrappers around the [essential process](handlers/zuora-gw-holiday-stop/src/main/scala/com/gu/zuoragwholidaystop/HolidayStopProcess.scala).

/cc @paulbrown1982 @mario-galic @twrichards


handlers/zuora-gw-holiday-stop/src/main/scala/com/gu/zuoragwholidaystop/Subscription.scala
```diff
@@ -0,0 +1,23 @@
...
+    ratePlanCharges: Seq[RatePlanCharge]
+)
+
+case class RatePlanCharge(price: Double, billingPeriod: String) {
+
+  val weekCount: Int = {
+    // inexact values
+    billingPeriod match {
+      case "Month"   => 4
+      case "Quarter" => 12
```
paulbrown1982: A quarter is mostly 13 weeks (hence that x 4 = 52).

handlers/zuora-gw-holiday-stop/src/main/scala/com/gu/zuoragwholidaystop/SubscriptionUpdate.scala
```diff
@@ -0,0 +1,54 @@
...
+      stoppedPublicationDate: LocalDate
+  ): SubscriptionUpdate = {
+    val credit =
+      autoRenewingHolidayCredit(subscription)
+    SubscriptionUpdate(
+      Seq(
+        Add(
+          productRatePlanId,
+          contractEffectiveDate = stoppedPublicationDate,
+          customerAcceptanceDate = stoppedPublicationDate,
```
paulbrown1982: This would need to align to an invoicing date, else it will generate a dangling negative invoice (which does have a background job to clear it up, but that's an awkward project to maintain as it has to use the SOAP API to create credit balance adjustments).

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/Config.scala
```diff
@@ -0,0 +1,74 @@
...
+  def apply(): Either[String, Config] = {
+    val stage = Option(System.getenv("Stage")).getOrElse("DEV")
+    val s3Config = configFromS3(stage)
+    stage match {
+      case "PROD" =>
+        s3Config map { secretConfig =>
+          Config(
+            secretConfig,
+            holidayCreditProductRatePlanId = "2c92a0fc5b42d2c9015b6259f7f40040",
+            // todo
```
twrichards: what is there TODO?

kelvin-chappell: Well spotted!  That's a placeholder because I had difficulty creating the new one-off charge.  I'll fill in the value before I merge.

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/Config.scala
```diff
@@ -0,0 +1,73 @@
...
+  zuoraAccess: ZuoraAccess,
+  holidayCreditProductRatePlanId: String,
+  holidayCreditProductRatePlanChargeId: String
+)
+
+case class ZuoraAccess(baseUrl: String, username: String, password: String)
+
+object Config {
+
+  private def configFromS3(stage: String): Either[String, ZuoraAccess] = {
```
mario-galic: Config errors are probably unrecoverable, so personally, I don't model them and just throw fast.

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/Config.scala
```diff
@@ -0,0 +1,73 @@
...
+    val profileName = "membership"
+    val bucketName = "gu-reader-revenue-private"
+    val key =
+      if (stage == "DEV")
```
mario-galic: Can DEV config be renamed to match PROD such that instead of switching on stage we write 

```
val stage = getStage
val key = s"membership/support-service-lambdas/$stage/zuoraRest-$stage.v1.json"
```

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/Credit.scala
```diff
@@ -0,0 +1,16 @@
+package com.gu.holidaystopprocessor
+
+import scala.math.BigDecimal.RoundingMode
+
+object Credit {
```
mario-galic: Instead of `Credit.autoRenewingHolidayAmount(sub)` we could write `AutoRenewingHolidayAmount(sub)` if we provide the following definition

```
object AutoRenewingHolidayAmount { def apply(sub) ... }
```

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/SubscriptionUpdate.scala
```diff
@@ -0,0 +1,59 @@
...
+  customerAcceptanceDate: LocalDate,
+  serviceActivationDate: LocalDate,
+  chargeOverrides: Seq[ChargeOverride]
+)
+
+case class ChargeOverride(
+  productRatePlanChargeId: String,
+  HolidayStart__c: LocalDate,
+  HolidayEnd__c: LocalDate,
+  price: Double
```
mario-galic: From my understanding `price` should be negative so could we add `require(price <= 0)` in constructor.

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/Credit.scala
```diff
@@ -0,0 +1,16 @@
...
+
+  def autoRenewingHolidayAmount(subscription: Subscription): Double = {
+    def roundUp(d: Double): Double =
+      BigDecimal(d).setScale(2, RoundingMode.UP).toDouble
+    subscription.originalRatePlanCharge map { charge =>
+      val recurringPayment = charge.price
+      val numPublicationsInPeriod = charge.weekCount
+      -roundUp(recurringPayment / numPublicationsInPeriod)
```
mario-galic: Negative result seems to be an invariant here, se would could assert on this. See comment on `ChargeOverride` construction.

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/HolidayStopProcess.scala
```diff
@@ -0,0 +1,25 @@
+package com.gu.holidaystopprocessor
+
+object HolidayStopProcess {
```
mario-galic: How about rename to `ProcessHolidayStop` as it is an action.

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/HolidayStopProcess.scala
```diff
@@ -0,0 +1,25 @@
...
+
+    val subscriptionDetails =
+      Zuora.subscriptionGetResponse(config.zuoraAccess) _
+
+    val updatedSubscription =
+      Zuora.subscriptionUpdateResponse(config.zuoraAccess) _
+
+    val holidayCreditToAdd = SubscriptionUpdate.holidayCreditToAdd(config) _
+
+    subscriptionDetails(stop.subscriptionName).right flatMap { subscription =>
```
mario-galic: Why `.right`? Eithers are already right-biased in Scala 2.12.

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/StandaloneApp.scala
```diff
@@ -0,0 +1,20 @@
+package com.gu.holidaystopprocessor
+
+import java.time.LocalDate
+
+object StandaloneApp extends App {
```
mario-galic: Maybe document this is for testing purposes only.

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/Subscription.scala
```diff
@@ -0,0 +1,36 @@
...
+  ratePlanCharges: Seq[RatePlanCharge]
+)
+
+case class RatePlanCharge(
+  price: Double,
+  billingPeriod: Option[String],
+  effectiveEndDate: LocalDate
+) {
+  val weekCount: Int = {
+    // inexact values
```
mario-galic: Maybe move the comment in the name itself like `weekCountApprox`. It is ugly but it probably should be to provide signal to users.

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/SubscriptionUpdate.scala
```diff
@@ -0,0 +1,59 @@
...
+
+  def holidayCreditToAdd(config: Config)(
+    subscription: Subscription,
+    stoppedPublicationDate: LocalDate
+  ): SubscriptionUpdate = {
+    val credit = Credit.autoRenewingHolidayAmount(subscription)
+    val effectiveDate = subscription.originalRatePlanCharge map {
+      _.effectiveEndDate.plusDays(1)
+    } getOrElse stoppedPublicationDate
+    val currentTerm =
```
mario-galic: Factor this out into `expandTerm` function and document the need as it is confusing institutional knowledge.

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/SubscriptionUpdate.scala
```diff
@@ -0,0 +1,59 @@
...
+
+case class SubscriptionUpdate(currentTerm: Option[Int], add: Seq[Add])
+
+object SubscriptionUpdate {
+
+  def holidayCreditToAdd(config: Config)(
+    subscription: Subscription,
+    stoppedPublicationDate: LocalDate
+  ): SubscriptionUpdate = {
+    val credit = Credit.autoRenewingHolidayAmount(subscription)
```
mario-galic: Inline this

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/SubscriptionUpdate.scala
```diff
@@ -0,0 +1,59 @@
...
+          )
+        )
+      )
+    )
+  }
+}
+
+case class Add(
+  productRatePlanId: String,
+  contractEffectiveDate: LocalDate,
```
mario-galic: `contractEffectiveDate`, `customerAcceptanceDate`, and `serviceActivationDate` should always be the same so we could require it at construction time.

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/Zuora.scala
```diff
@@ -0,0 +1,46 @@
...
+
+object Zuora {
+
+  implicit val backend: SttpBackend[Id, Nothing] = HttpURLConnectionBackend()
+
+  private def normalised[E <: io.circe.Error, R](
+    body: String, result: String => Either[E, R]
+  ): Either[String, R] = result(body).left.map(e => s"Failed to decode '$body': ${e.toString}")
+
+  def subscriptionGetResponse(
```
mario-galic: Maybe rename to `getSubscription`?

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/SubscriptionUpdate.scala
```diff
@@ -0,0 +1,59 @@
...
+        subscription.currentTermPeriodType match {
+          case "Month" => Some(24)
+          case _ => None
+        }
+      } else None
+    SubscriptionUpdate(
+      currentTerm,
+      Seq(
+        Add(
+          productRatePlanId = config.holidayCreditProductRatePlanId,
```
paulbrown1982: This will eventually be product-specific.

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/Credit.scala
```diff
@@ -0,0 +1,16 @@
+package com.gu.holidaystopprocessor
+
+import scala.math.BigDecimal.RoundingMode
+
+object Credit {
+
+  def autoRenewingHolidayAmount(subscription: Subscription): Double = {
```
mario-galic: Could we assert price is not wildly out-of-bounds, like say negative million.

handlers/holiday-stop-processor/src/test/scala/com/gu/holidaystopprocessor/CreditTest.scala
```diff
@@ -0,0 +1,41 @@
...
+      termEndDate = LocalDate.now,
+      price = 37.5,
+      billingPeriod = "Quarter",
+      effectiveEndDate = LocalDate.of(2019, 5, 15)
+    )
+    val credit = Credit.autoRenewingHolidayAmount(subscription)
+    credit shouldBe -2.89
+  }
+
+  "autoRenewingHolidayCredit" should "be correct for an annual billing period" in {
```
mario-galic: `it` removes having to repeat `autoRenewingHolidayCredit` on every test like so

```
it should "be correct for another quarterly billing period"
```

instead of 

```
"autoRenewingHolidayCredit" should "be correct for another quarterly billing period" 
```

handlers/holiday-stop-processor/src/test/scala/com/gu/holidaystopprocessor/CreditTest.scala
```diff
@@ -0,0 +1,41 @@
+package com.gu.holidaystopprocessor
+
+import java.time.LocalDate
+
+import org.scalatest.{FlatSpec, Matchers}
+
+class CreditTest extends FlatSpec with Matchers {
```
mario-galic: Add test `"autoRenewingHolidayAmount" should "return negative price"`

handlers/holiday-stop-processor/src/test/scala/com/gu/holidaystopprocessor/SubscriptionUpdateTest.scala
```diff
@@ -0,0 +1,77 @@
...
+      subscription = Fixtures.mkSubscription(
+        termEndDate = LocalDate.of(2019, 12, 1),
+        price = 42.1,
+        billingPeriod = "Quarter",
+        effectiveEndDate = LocalDate.of(2020, 5, 3)
+      ),
+      stoppedPublicationDate = LocalDate.of(2019, 5, 18)
+    )
+    update shouldBe SubscriptionUpdate(
+      currentTerm = Some(24),
```
mario-galic: @paulbrown1982 Can you confirm it is safe to extend `termEndDate` at will?

handlers/holiday-stop-processor/cfn.yaml
```diff
@@ -0,0 +1,64 @@
...
+      Code:
+        S3Bucket: support-service-lambdas-dist
+        S3Key: !Sub membership/${Stage}/holiday-stop-processor/holiday-stop-processor.jar
+      Handler: com.gu.holidaystopprocessor.Handler::handle
+      Environment:
+        Variables:
+          Stage: !Ref Stage
+      Role:
+        !GetAtt HolidayStopProcessorRole.Arn
+      MemorySize: 256
```
mario-galic: Put a gig?

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/Config.scala
```diff
@@ -0,0 +1,73 @@
+package com.gu.holidaystopprocessor
```
mario-galic: Maybe rename to `com.go.holiday`

handlers/holiday-stop-processor/cfn.yaml
```diff
@@ -0,0 +1,64 @@
+AWSTemplateFormatVersion: "2010-09-09"
+Description: Updates subscriptions with outstanding holiday stops.
```
mario-galic: Add alarm on failed lambda

handlers/holiday-stop-processor/src/test/scala/com/gu/holidaystopprocessor/SubscriptionUpdateTest.scala
paulbrown1982: On an auto-renewing sub, yup no problem.

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/Config.scala
kelvin-chappell: I've remodelled this slightly in an upcoming PR but will leave this as is here.

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/Config.scala
kelvin-chappell: Yes, that would be sensible.  I don't understand why the dev file has a different naming scheme to the other stages in S3.  Shall I just change it?

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/Credit.scala
kelvin-chappell: Yes, I'll do that.  I was thinking it could be used as a home for all types of credit calculation but that kind of imagining the future is to be avoided.

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/SubscriptionUpdate.scala
kelvin-chappell: Should we be using assertions in production code?  I'd rather set up a scalacheck test of the credit calculation code to ensure it never returns a positive value.

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/Credit.scala
kelvin-chappell: I'll set up a scalacheck test with this as an invariant.

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/HolidayStopProcess.scala
kelvin-chappell: I didn't see that.  Thanks.  Removed now.

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/StandaloneApp.scala
kelvin-chappell: Done.

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/Subscription.scala
kelvin-chappell: Done.

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/SubscriptionUpdate.scala
kelvin-chappell: Done.

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/SubscriptionUpdate.scala
kelvin-chappell: Done.

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/Zuora.scala
kelvin-chappell: As discussed, think in a functional language methods should be the name of the thing they return rather than the name of a process as in an OO language.  Although this isn't consistent because the way Scala is used is not entirely one or the other.

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/SubscriptionUpdate.scala
kelvin-chappell: At the moment this code is all very GW specific.  I expect to be parameterised more as its scope expands to more products.

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/Credit.scala
kelvin-chappell: I'll add this to a scalacheck test.

handlers/holiday-stop-processor/src/test/scala/com/gu/holidaystopprocessor/CreditTest.scala
kelvin-chappell: Done.

handlers/holiday-stop-processor/cfn.yaml
kelvin-chappell: Shall we try this and see what happens?  In Code it was using less than 200 MB but I'm sure this will increase when the SF bit is integrated.  So I'll expand as needed.

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/Config.scala
kelvin-chappell: I was worried that there will be multiple components doing holiday stop work.  An IDE will insert the package name when you create a new class so there should be a small likelihood of a typo.

handlers/holiday-stop-processor/cfn.yaml
kelvin-chappell: Will do this in a separate PR. https://trello.com/c/DkXszGaJ

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/Subscription.scala
```diff
@@ -0,0 +1,36 @@
...
+
+case class Subscription(
+  termEndDate: LocalDate,
+  currentTerm: Int,
+  currentTermPeriodType: String,
+  autoRenew: Boolean,
+  ratePlans: Seq[RatePlan]
+) {
+  val originalRatePlanCharge: Option[RatePlanCharge] = for {
+    ratePlan <- ratePlans.lastOption
```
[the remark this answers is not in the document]
kelvin-chappell: Now sorting by charge effective start date before taking the rate plan.

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/HolidayStopProcess.scala
kelvin-chappell: Same reasoning as below.  ie. functional naming rather than OO naming.

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/SubscriptionUpdate.scala
kelvin-chappell: I think setting them all to the same variable should be enough?

handlers/holiday-stop-processor/src/test/scala/com/gu/holidaystopprocessor/CreditTest.scala
kelvin-chappell: Done.

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/Credit.scala
kelvin-chappell: Done.

kelvin-chappell:
@mario-galic:

> * readme documenting architecture and how it relates to Salesforce client    

Card for this: https://trello.com/c/07hEeLaM

> * assurances price is not being increased

There are now some tests for this 

> * cloudwatch alarm via email to SX on failure

Card for this: https://trello.com/c/DkXszGaJ

PR #294, opened 2019-05-17
Holiday Stops : shared library (mainly SalesForce integration)

twrichards:
Added new shared library `holiday-stops` (in `lib`) for use by handlers: `holiday-stops-api` (_insert PR reference here_) and `holiday-stops-processor` (https://github.com/guardian/support-service-lambdas/pull/289).

Largely for interacting with new SalesForce objects `Holiday Stop Request` and `Holiday Stop Request Actioned Zuora Ref` (which have a _master-detail_ relationship)

Also provides simple `ActionCalculator` for calculating the dates of issues impacted by holiday stop, this can be used both to present client-side and also compare the count against actioned count so far to determine what still needs to be processed.

PR #300, opened 2019-05-29
FIX recent bad refactor which was string concatenating a RelativePath

twrichards:
Introduced in commit https://github.com/guardian/support-service-lambdas/pull/294/commits/5761a77d8d79e282495be630c24a9e5fd9b28413 as part of https://github.com/guardian/support-service-lambdas/pull/294

Caused the cancellation cases lambda and the GoCardless sync lambda to break for a short period, so old master was deployed to stabilise, but this is proper FIX.


twrichards:
I missed this because I misinterpreted an _effects test_ failure.

PR #304, opened 2019-05-31
 Read holiday stop requests from Salesforce

kelvin-chappell:
This is the first step in integrating the holiday-stop processing lambda with Salesforce.  Salesforce holds a table of holiday-stop requests.  This PR reads those requests and sends them to Zuora to be processed.

There will be a separate PR to write the amendment details made by Zuora back to Salesforce.

### DIFF
--- a/handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/Handler.scala
+++ b/handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/Handler.scala
@@ -1,5 +1,6 @@
 package com.gu.holidaystopprocessor
 
+import cats.implicits._
 import com.amazonaws.services.lambda.runtime.Context
 import com.gu.util.Logging
 import io.circe.generic.auto._
@@ -25,10 +26,7 @@ object Handler
           case Right(response) => logger.info(response)
         }
 
-        responses collectFirst {
-          case Left(msg) => Left(new RuntimeException(msg))
-        } getOrElse
-          Right(responses collect { case Right(response) => response })
+        responses.toList.sequence.leftMap(new RuntimeException(_))
     }
   }
 }
--- a/handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/Handler.scala
+++ b/handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/Handler.scala
@@ -1,23 +1,34 @@
 package com.gu.holidaystopprocessor
 
 import com.amazonaws.services.lambda.runtime.Context
+import com.gu.util.Logging
 import io.circe.generic.auto._
 import io.github.mkotsur.aws.handler.Lambda
 import io.github.mkotsur.aws.handler.Lambda._
 
-object Handler extends Lambda[HolidayStop, HolidayStopResponse] {
+object Handler
+  extends Lambda[HolidayStop, Seq[HolidayStopResponse]]
+  with Logging {
 
   override protected def handle(
     holidayStop: HolidayStop,
     context: Context
-  ): Either[Throwable, HolidayStopResponse] = {
+  ): Either[Throwable, Seq[HolidayStopResponse]] = {
+
     Config() match {
       case Left(msg) => Left(new RuntimeException(s"Config failure: $msg"))
       case Right(config) =>
-        HolidayStopProcess(config, holidayStop) match {
-          case Left(msg) => Left(new RuntimeException(msg))
-          case Right(response) => Right(response)
+        val responses = HolidayStopProcess(config)
+
+        responses foreach {
+          case Left(msg) => logger.error(msg)
+          case Right(response) => logger.info(response)
         }
+
+        responses collectFirst {
+          case Left(msg) => Left(new RuntimeException(msg))
+        } getOrElse
+          Right(responses collect { case Right(response) => response })
     }
   }
 }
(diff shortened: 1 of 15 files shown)


handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/HolidayStopProcess.scala
```diff
@@ -2,27 +2,26 @@ package com.gu.holidaystopprocessor
...
-  def process(
+  def processHolidayStop(
     config: Config,
     getSubscription: String => Either[String, Subscription],
     updateSubscription: (Subscription, SubscriptionUpdate) => Either[String, Unit],
-    getLastAmendment: Subscription => Either[String, Amendment],
-    stop: HolidayStop
-  ): Either[String, HolidayStopResponse] = {
+    getLastAmendment: Subscription => Either[String, Amendment]
+  )(stop: HolidayStop): Either[String, HolidayStopResponse] = {
```
mario-galic: Out-of-scope for this PR, but could `processHolidayStop` be rewritten to use for-yield like so:

```
    for {
      subscription <- getSubscription(stop.subscriptionName)
      _ <- if (subscription.autoRenew) Right(()) else Left("Cannot currently process non-auto-renewing subscription")
      update <- Right(SubscriptionUpdate.holidayCreditToAdd(config, subscription, stop.stoppedPublicationDate))
      _ <- updateSubscription(subscription, update)
      amendment <- getLastAmendment(subscription)
    } yield {
      HolidayStopResponse(amendment.code, update.price)
    }
```

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/Handler.scala
```diff
@@ -1,23 +1,34 @@
...
-          case Left(msg) => Left(new RuntimeException(msg))
-          case Right(response) => Right(response)
+        val responses = HolidayStopProcess(config)
+
+        responses foreach {
+          case Left(msg) => logger.error(msg)
+          case Right(response) => logger.info(response)
         }
+
+        responses collectFirst {
```
mario-galic: This can be shortened using cat's [`Traverse.sequence`](https://github.com/typelevel/cats/blob/f10484b67d545fd677747d300ea54580a44050a2/core/src/main/scala/cats/Traverse.scala#L83) like so

```
import cats.implicits._
responses.toList.sequence.leftMap(new RuntimeException(_))
```

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/Salesforce.scala
```diff
@@ -0,0 +1,27 @@
...
+
+object Salesforce {
+
+  private def thresholdDate: LocalDate = LocalDate.now.plusDays(14)
+
+  def holidayStopRequests(
+    config: Config,
+    productNamePrefix: String
+  ): Either[String, List[HolidayStopRequest]] =
+    SalesforceClient(RawEffects.response, config.sfCredentials).value.flatMap { sfAuth =>
```
mario-galic: Personally, I would use a standard HTTP client, similar to what has been done for Zuora requests.

handlers/holiday-stop-processor/src/test/scala/com/gu/holidaystopprocessor/HolidayStopProcessTest.scala
```diff
@@ -37,60 +33,55 @@ class HolidayStopProcessTest extends FlatSpec with Matchers with EitherValues {
   }
 
   "HolidayStopProcess" should "give correct amendment" in {
-    val response = HolidayStopProcess.process(
+    val response = HolidayStopProcess.processHolidayStop(
       config,
       getSubscription(Right(subscription)),
       updateSubscription(Right(Unit)),
```
mario-galic: This has not been introduced in this PR, however `Right(Unit)` should be replaced with `Right(())`, because `Unit` is the _type_ while `()` is the _value_ inhabiting `Unit`, similar to how `Int` is the type while `7` is the value inhabiting `Int`.

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/HolidayStopProcess.scala
kelvin-chappell: Yes but I'm not sure it would be any clearer

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/Handler.scala
kelvin-chappell: I wasn't aware of that.  Thanks for the tip.  I'll do it.

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/Salesforce.scala
kelvin-chappell: I can return to this when I'm more familiar with the Salesforce API.  It's isolated so can be easily changed.

handlers/holiday-stop-processor/src/test/scala/com/gu/holidaystopprocessor/HolidayStopProcessTest.scala
kelvin-chappell: Yes it's strange how that compiles.  Changed it now.  Thanks.

lib/holiday-stops/src/main/scala/com/gu/holiday_stops/ActionCalculator.scala
```diff
@@ -14,21 +12,23 @@ object ActionCalculator {
...
+      val dateRange = 0 to Days.daysBetween(from, to).getDays
+      dateRange.foldLeft(List.empty[LocalDate]) { (acc, i) =>
+        val d = from.plusDays(i)
+        if (p(d)) acc :+ d
+        else acc
+      }
+    }
+
+    val from = hsr.Start_Date__c.value
+    val to = hsr.End_Date__c.value
```
paulbrown1982: Just checking whether this is inclusive or exclusive, maybe a comment may help the next person who comes along.

kelvin-chappell: I've done some renaming now to make this clearer.

PR #306, opened 2019-05-31
Update Zuora rate plan IDs for holiday credits

kelvin-chappell:
I have created entirely new rate plans in each Zuora stage to avoid new charges appearing in existing holiday credits.  These are the IDs of the new plans and their charges.


kelvin-chappell:
Not much value in peer review of this change.

PR #312, opened 2019-06-04
Export holiday-stop results to Salesforce

kelvin-chappell:
This change writes the details of amendments to subscriptions made by Zuora back to Salesforce.

Implementation details:
* There's a [new method](https://github.com/guardian/support-service-lambdas/compare/kc-sf-write?expand=1#diff-c5ebc37fd50df1a8bbff179733bf2ffcR29) to write holiday-stop responses back to Salesforce
* `HolidayStops` and `HolidayStopResponses` now keep the original request ID from Salesforce so that it can be passed back after the update

That's about it.

### DIFF
--- a/handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/HolidayStop.scala
+++ b/handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/HolidayStop.scala
@@ -3,22 +3,27 @@ package com.gu.holidaystopprocessor
 import java.time.LocalDate
 
 import com.gu.holiday_stops.ActionCalculator
-import com.gu.salesforce.holiday_stops.SalesforceHolidayStopRequest.HolidayStopRequest
+import com.gu.salesforce.holiday_stops.SalesforceHolidayStopRequest.{HolidayStopRequest, HolidayStopRequestId}
 
 case class HolidayStop(
+  requestId: HolidayStopRequestId,
   subscriptionName: String,
   stoppedPublicationDate: LocalDate
 )
 
 object HolidayStop {
 
-  def holidayStopsToApply(config: Config): Either[String, List[HolidayStop]] =
-    Salesforce.holidayStopRequests(config, "Guardian Weekly") map {
+  def holidayStopsToApply(getRequests: String => Either[String, Seq[HolidayStopRequest]]): Either[String, Seq[HolidayStop]] =
+    getRequests("Guardian Weekly") map {
       _ flatMap toHolidayStops
     }
 
-  private def toHolidayStops(request: HolidayStopRequest): List[HolidayStop] =
+  private def toHolidayStops(request: HolidayStopRequest): Seq[HolidayStop] =
     ActionCalculator.publicationDatesToBeStopped(request) map { date =>
-      HolidayStop(request.Subscription_Name__c.value, Time.toJavaDate(date))
+      HolidayStop(
+        requestId = request.Id,
+        subscriptionName = request.Subscription_Name__c.value,
+        stoppedPublicationDate = Time.toJavaDate(date)
+      )
     }
 }
--- a/handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/HolidayStop.scala
+++ b/handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/HolidayStop.scala
@@ -13,7 +13,7 @@ case class HolidayStop(
 
 object HolidayStop {
 
-  def holidayStopsToApply(getRequests: String => Either[String, Seq[HolidayStopRequest]]): Either[String, Seq[HolidayStop]] =
+  def holidayStopsToApply(getRequests: String => Either[OverallFailure, Seq[HolidayStopRequest]]): Either[OverallFailure, Seq[HolidayStop]] =
     getRequests("Guardian Weekly") map {
       _ flatMap toHolidayStops
     }
--- a/handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/HolidayStopProcess.scala
+++ b/handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/HolidayStopProcess.scala
@@ -5,7 +5,7 @@ import com.gu.salesforce.holiday_stops.SalesforceHolidayStopRequestActionedZuora
 
 object HolidayStopProcess {
 
-  def apply(config: Config): Seq[Either[String, HolidayStopResponse]] = {
+  def apply(config: Config): ProcessResult = {
     val sfCredentials = config.sfCredentials
     val zuoraCredentials = config.zuoraCredentials
     processHolidayStops(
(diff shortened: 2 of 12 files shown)


handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/HolidayStopProcess.scala
```diff
@@ -1,18 +1,46 @@
...
+      updateSubscription,
+      getLastAmendment
     ) _
-    HolidayStop.holidayStopsToApply(config) match {
+    HolidayStop.holidayStopsToApply(getRequests) match {
       case Left(msg) => Seq(Left(msg))
-      case Right(holidayStops) => holidayStops map response
+      case Right(holidayStops) =>
+        val responses = holidayStops.map(response)
+        val exportResult = exportAmendments(responses.collect { case Right(successes) => successes })
```
mario-galic: We seem to discard errors in `responses` which are results of applying holiday stop in Zuora.

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/HolidayStopProcess.scala
```diff
@@ -1,18 +1,46 @@
...
+  }
+
+  def processHolidayStops(
+    config: Config,
+    getRequests: String => Either[String, Seq[HolidayStopRequest]],
+    getSubscription: String => Either[String, Subscription],
+    updateSubscription: (Subscription, SubscriptionUpdate) => Either[String, Unit],
+    getLastAmendment: Subscription => Either[String, Amendment],
+    exportAmendments: Seq[HolidayStopResponse] => Either[String, Unit]
+  ): Seq[Either[String, HolidayStopResponse]] = {
```
mario-galic: Consider changing the return type from 

```
Seq[Either[String, HolidayStopResponse]]
``` 
to 

```
Either[String, Seq[HolidayStopResponse]]
```

Top level handler has return type `Either[Throwable, Seq[HolidayStopResponse]]` so we should match that. We should be able to write simple for-yield without nested pattern-matches and collects, perhaps something like so

```
  for {
    holidayStops <- HolidayStop.holidayStopsToApply(getRequests)
    responses <- holidayStops.map(processHolidayStop(config, getSubscription, updateSubscription, getLastAmendment)).toList.sequence
    exportResult <- exportAmendments(responses)
  } yield { responses }
```

arausuy:
A bit more general but I'm a bit confused by the mixture of postfix and dot notation

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/Time.scala
```diff
@@ -1,10 +1,14 @@
 package com.gu.holidaystopprocessor
 
 import java.time.LocalDate
+
 import org.joda.time.{LocalDate => JodaLocalDate}
 
 object Time {
```
arausuy: Can't we achieve this with enrichment rather than explicit calls?

handlers/holiday-stop-processor/src/test/scala/com/gu/holidaystopprocessor/Fixtures.scala
```diff
@@ -27,6 +28,15 @@ object Fixtures {
       )
     )
 
+  def mkHolidayStopRequest(id: String) = HolidayStopRequest(
+    Id = HolidayStopRequestId(id),
+    Start_Date__c = HolidayStopRequestStartDate(Time.toJodaDate(LocalDate.of(2019, 1, 1))),
+    End_Date__c = HolidayStopRequestEndDate(Time.toJodaDate(LocalDate.of(2019, 2, 15))),
```
arausuy: Can't we just make a JodaDate instance directly?

Why do we need to use both LocalDate packages?

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/HolidayStop.scala
```diff
@@ -3,22 +3,27 @@ package com.gu.holidaystopprocessor
...
 
-  def holidayStopsToApply(config: Config): Either[String, List[HolidayStop]] =
-    Salesforce.holidayStopRequests(config, "Guardian Weekly") map {
+  def holidayStopsToApply(getRequests: String => Either[String, Seq[HolidayStopRequest]]): Either[String, Seq[HolidayStop]] =
+    getRequests("Guardian Weekly") map {
       _ flatMap toHolidayStops
     }
 
-  private def toHolidayStops(request: HolidayStopRequest): List[HolidayStop] =
+  private def toHolidayStops(request: HolidayStopRequest): Seq[HolidayStop] =
```
arausuy: Can we be more specific than Seq? i.e Set?

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/Time.scala
kelvin-chappell: The Joda time library is deprecated in favour of java time now. (See https://www.joda.org/joda-time/).  But in most cases Joda time is being used in this repo.  Rather than change it everywhere I thought by having an explicit converter it would be clear where one library was being used rather than the other.  As they have very similar names for time concepts, such as LocalDate, I thought it might be more confusing if I did this conversion implicitly.

handlers/holiday-stop-processor/src/test/scala/com/gu/holidaystopprocessor/Fixtures.scala
kelvin-chappell: See above.

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/HolidayStop.scala
kelvin-chappell: We're not doing anything specifically setlike with them so it seemed best to keep the type as general as possible, without it becoming meaningless.

kelvin-chappell:
@arausuy I tend to avoid dots where I think it makes the code clearer.  So sometimes there's a lack of consistency.  Are there any cases where you think the lack of dots is confusing or could have unforeseen effects?

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/HolidayStopProcess.scala
kelvin-chappell: I have a new return type now, which separates out individual exceptions in holiday stops from general processing exceptions.  So there's no loss of information in what's returned.

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/HolidayStopProcess.scala
kelvin-chappell: I wanted to return all the individual results to the handler so that it could do what it likes with them - log them most likely.  I've now got a new return type that holds an optional overall exception and individual exceptions per holiday stop.

arausuy:
> @arausuy I tend to avoid dots where I think it makes the code clearer. So sometimes there's a lack of consistency. Are there any cases where you think the lack of dots is confusing or could have unforeseen effects?

It was more just consistent style across multiple developers - I think its more of a general code style question rather than a dot notation vs postfix.

PR #315, opened 2019-06-06
Add more logging to holiday-stop processing lambda

kelvin-chappell:


handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/Handler.scala
```diff
@@ -2,28 +2,30 @@ package com.gu.holidaystopprocessor
...
       case Right(config) =>
         val processResult = HolidayStopProcess(config)
 
+        logger.log(s"${processResult.holidayStopsToApply.size} holiday stops to apply:")
+        processResult.holidayStopsToApply.foreach(stop => logger.log(stop.toString))
+
         processResult.holidayStopResults foreach {
-          case Left(failure) => logger.error(failure.reason)
-          case Right(response) => logger.info(response)
+          case Left(failure) => logger.log(failure.reason)
```
mario-galic: Are we losing log levels here? Why not extend `LazyLogging` from `scala-logging`?

kelvin-chappell: Reverted to how it was now, but with console logger configured.

kelvin-chappell:
I'll have a go at the for comprehension approach.

kelvin-chappell:
@mario-galic I think the difficulty in the for-comprehension is that I want to export any amendments that succeed and log failures.  But that's difficult to do neatly in a for-comprehension.  Can pair on this later if you can see a solution.  But will do in a separate PR.

PR #319, opened 2019-06-10
Use correct effective date for GW holiday stops

kelvin-chappell:
This change ensures that the negative charge is applied to the next invoice by using the charge-through date (the first day of the next service period) as the effective date.

### DIFF
--- a/handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/SubscriptionUpdate.scala
+++ b/handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/SubscriptionUpdate.scala
@@ -2,7 +2,7 @@ package com.gu.holidaystopprocessor
 
 import java.time.LocalDate
 
-case class SubscriptionUpdate(currentTerm: Option[Int], add: Seq[Add]) {
+case class SubscriptionUpdate(add: Seq[Add]) {
 
   val price: Double = {
     val optPrice = for {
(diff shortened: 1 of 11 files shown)


handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/SubscriptionUpdate.scala
```diff
@@ -15,42 +15,32 @@ case class SubscriptionUpdate(currentTerm: Option[Int], add: Seq[Add]) {
...
-          customerAcceptanceDate = effectiveDate,
-          serviceActivationDate = effectiveDate,
-          chargeOverrides = Seq(
-            ChargeOverride(
-              productRatePlanChargeId = config.holidayCreditProductRatePlanChargeId,
-              HolidayStart__c = stoppedPublicationDate,
-              HolidayEnd__c = stoppedPublicationDate,
-              price = HolidayCredit(subscription)
+  ): Either[HolidayStopFailure, SubscriptionUpdate] = {
+    subscription.originalRatePlanCharge.flatMap(_.chargedThroughDate)
```
paulbrown1982: Does chargedThroughDate also need `.plusDays(1)`?

kelvin-chappell: No, the chargedThroughDate is the first date of the next billing period.  I've checked the invoice previews with @mario-galic and it does the right thing, ie. holiday stops taking place in this billing period appear in the next invoice.  There's a separate card for considering what to do if holiday stop goes over threshold into the next billing period.

mario-galic: Note the difference between `Service period` shown on an invoice, and corresponding API dates `[effectiveStartDate, effectiveEndDate)`:
 
* Service Period: 20.10.2018-19.01.2019
* `[effectiveStartDate, effectiveEndDate)`: 20.10.2018-20.01.2019

On an invoice the end date is _inclusive_ whist on an API response it is _exclusive_

kelvin-chappell: I think it's consistent.  The `chargedThroughDate` is the first day in the next service period if you're paying in advance.

PR #321, opened 2019-06-12
Don't allow Zuora to apply duplicate holiday-stop credits

kelvin-chappell:
There are some scenarios where we ask Zuora to apply the same holiday-stop twice.  This could be where sending the result of the amendment back to Salesforce fails for example.  In that case on the next run of the lambda the same amendment would be applied again and there would be nothing to stop it appearing as two or more credits for the same holiday stop on the next invoice.

This PR ensures that the same holiday stop cannot be applied twice.  It works on the assumption that a 'Holiday Credit' charge on a 'Discounts' plan for the same stopped publication date as the requested holiday stop means that the holiday stop has already been applied.

It's been easier to identify duplicates by inspecting the `Subscription` record, rather than by looking for relevant `Amendments`.  So I've changed the model a bit to avoid referring to `Amendments` at all.  This means that the lambda is sending `RatePlanCharge` IDs back to Salesforce rather than `Amendment` IDs.  Both uniquely identify the holiday-stop amendment to the original subscription.  So there'll be some follow-up work to change references in the service layer and the Salesforce UI from 'amendment' to 'charge'.

This is how the results look in Salesforce with the old amendment IDs and the new RatePlanCharge IDs:

image

It will be marginally more difficult to look these up in Zuora as RatePlanCharges don't exist as top-level objects.  They will have to be looked up by Subscription name instead to find the relevant charge in the subscription details.

I've tested this in Code.

### DIFF
--- a/handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/Subscription.scala
+++ b/handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/Subscription.scala
@@ -20,6 +20,24 @@ case class Subscription(
       charge <- ratePlan.ratePlanCharges.headOption
     } yield charge
   }
+
+  def ratePlanCharge(stop: HolidayStop): Option[RatePlanCharge] = {
+
+    def isMatchingPlan(plan: RatePlan): Boolean = plan.productName == "Discounts"
+
+    def isMatchingCharge(charge: RatePlanCharge): Boolean =
+      charge.name == "Holiday Credit" &&
+        charge.HolidayStart__c.contains(stop.stoppedPublicationDate) &&
+        charge.HolidayEnd__c.contains(stop.stoppedPublicationDate)
+
+    val charges = for {
+      plan <- ratePlans if isMatchingPlan(plan)
+      charge <- plan.ratePlanCharges.find(isMatchingCharge)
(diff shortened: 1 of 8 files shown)


handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/Subscription.scala
```diff
@@ -20,6 +20,24 @@ case class Subscription(
       charge <- ratePlan.ratePlanCharges.headOption
     } yield charge
   }
+
+  def ratePlanCharge(stop: HolidayStop): Option[RatePlanCharge] = {
+
+    def isMatchingPlan(plan: RatePlan): Boolean = plan.productName == "Discounts"
+
+    def isMatchingCharge(charge: RatePlanCharge): Boolean =
```
paulbrown1982: This is only compatible for single-day-range stops - I get that is how we're setting them going forward, I just wonder about what will happen when we encounter an old-world holiday? I think simply testing the range should cover this scenario: e.g. `charge.HolidayStart__c.exists(_.isSameOrBefore(stop.stoppedPublicationDate)) && charge.HolidyEnd__c.exists(_.isSameOrAfter(stop.stoppedPublicationDate))`

PR #331, opened 2019-06-20
WIP: add metric push api for alerting client side errors

johnduffell:
We have trouble at the moment that when things break on the client side, we don't know about it.  We investigated using sentry but this was not possible.  Now we have decided to have a fake img on the failure page that fetches this URL and we can cloudwatch alert on that.

Changes:
The most scary change is actually right at the bottom, I have updated the AWS lib version for everything!

TODO
DONE - deploy and test in CODE
DONE - check caching
DONE - check response content type
DONE - actually add alarm
- add runscope test
- add alarm if the metric pushed by runscope has no data


johnduffell:
thanks tom, if it's slow then it's not a problem as it won't be blocking anything on the page it's basicallyfire and forget.  I did/am considering API gateway->queue and then have a lambda batching off the queue, but I (hope) it won't be high enough volume for that to be an issue.  IF we're getting that many errors then we have bigger concerns than too many executions.

I was originally planning sentry, but unfortunately sentry is per person rather than setting up "urgent 9-5" style team alerts.  I think having consistency with the other alerts would be a plus point.  Our sentry is not clean and I'm not sure someone has the motivation to be an advocate for it/keeping it clear and cared about.  If someone will that's great but we just feel this will give more benefit for the effort with the current situation.

johnduffell:
On that note though it would be good if we can have a similar setup for TS lambdas as we have here for scala ones, so that it's easy(er) to pop one in and share libraries where necessary.  Either that or cross compile all these ones to ScalaJs 😜

handlers/metric-push-api/src/main/scala/com/gu/metric_push_api/Handler.scala
```diff
@@ -0,0 +1,94 @@
...
+  implicit val wireReads = Json.reads[WireUrlParams]
+
+  def toUrlParams(urlParamsWire: WireUrlParams): Either[String, MetricPushRequest] =
+    ValidMetricName.fromString(urlParamsWire.metricName).toRight("invalid metric name").map(MetricPushRequest.apply)
+
+  def apply(
+    steps: MetricPushRequest => Try[Unit]
+  ): ApiGatewayRequest => ResponseModels.ApiResponse = req =>
+    (for {
+      wireUrlParams <- req.queryParamsAsCaseClass[WireUrlParams]()
```
jfsoul: I was thinking we could also pull out the headers and validate that the `Referer` header has an expected domain `support.theguardian.com`.  The consequences of unexpected requests to this lambda seem fairly innocuous, especially if we limit the concurrent executions, but if the referrer check is simple then it wouldn't hurt.

lib/handler/src/main/scala/com/gu/util/apigateway/ApiGatewayResponse.scala
```diff
@@ -5,22 +5,34 @@ import java.io.{OutputStream, OutputStreamWriter}
...
-  case class ApiResponse(statusCode: String, body: String, headers: Headers = new Headers)
+  sealed abstract class CacheHeader(val keyValues: List[(String, JsValueWrapper)])
+  case object CacheDefault extends CacheHeader(List.empty)
+  case object CacheNoCache extends CacheHeader(List(
+    // this forces revalidation, but doesn't prevent storing. Use Private & NoStore
+    "Cache-control" -> "no-store"
+  ))
+  case class Headers(contentType: Option[String] = Some("application/json"), cache: CacheHeader = CacheDefault)
+
+  case class ApiResponse(statusCode: String, body: Option[String], headers: Headers)
```
jfsoul: I'm missing why we want to model the `body` as an `Option` now

handlers/metric-push-api/cfn.yaml
```diff
@@ -0,0 +1,190 @@
...
+        Default: CODE
+
+Conditions:
+  CreateProdMonitoring: !Equals [ !Ref Stage, PROD ]
+
+Mappings:
+    StageMap:
+        CODE:
+            ApiName: metric-push-api-api-CODE
+            DomainName: metric-push-api-code.membership.guardianapis.com
```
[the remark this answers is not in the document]
johnduffell: this is because of the route 53 zone we are using for these, if there's a `code.dev-guardianapis.com` one then maybe we can put it in there.  Or we can add one.  I'll have a look.

handlers/metric-push-api/cfn.yaml
```diff
@@ -0,0 +1,190 @@
...
+            Code:
+                S3Bucket: support-service-lambdas-dist
+                S3Key: !Sub membership/${Stage}/metric-push-api/metric-push-api.jar
+            Handler: com.gu.metric_push_api.Handler::apply
+            Environment:
+                Variables:
+                  Stage: !Ref Stage
+            Role:
+                !GetAtt MetricPushRole.Arn
+            MemorySize: 1536
```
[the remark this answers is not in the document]
johnduffell: it's just a "plenty" value to make sure it executes quickly, we've never worried too much about the cost, but I can try lower values if you think it's excessive.

handlers/metric-push-api/cfn.yaml
```diff
@@ -0,0 +1,190 @@
...
+    MetricPushMethod:
+        Type: AWS::ApiGateway::Method
+        Properties:
+            AuthorizationType: NONE
+            ApiKeyRequired: false
+            RestApiId: !Ref MetricPushAPI
+            ResourceId: !Ref MetricPushProxyResource
+            HttpMethod: GET
+            RequestParameters:
+              method.request.querystring.identityId: true
```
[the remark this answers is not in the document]
johnduffell: no, that's probably because I copied and pasted the sf-contact-merge one, good spot

lib/handler/src/main/scala/com/gu/util/apigateway/ApiGatewayResponse.scala
johnduffell: in the response, if it's a 204 then there is no response body
https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html
> The 204 response MUST NOT include a message-body, and thus is always terminated by the first empty line after the header fields.

johnduffell:
@jfsoul with these settings and no lambda backing the API gateway:
image
we can get these logs, which seem sufficient to understand any browser issue, and we get per method count alerting.
image
If we want to add additional information we could add it into the request query string which is logged above.
The only down side is this is logged as a separate line from the headers, so if we use that functionality it will be hard to correlate with the browser headers just be searching.

jfsoul:
Sounds good to me @johnduffell - if it's simpler, I say ship it!

johnduffell:
ok, I (think) i've got the basic cloudformation working, I haven't deleted all the scala changes yet or fixed up the other comments, but I'm going to do some testing in CODE.

handlers/metric-push-api/cfn.yaml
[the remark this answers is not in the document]
johnduffell: looking at https://console.aws.amazon.com/route53/home?region=eu-west-1#resource-record-sets:Z1E4V12LQGXFEC it does seem to be the convention.  Whether it should be or not is another matter!  it might even have been me that started the convention by mistake.

There is a *.support.guardianapis.com certificate and record set so I will at least change "membership" to "support".
To go further, it would have to be a separate cert (and record set) for *.code.dev-support.guardianapis.com, do you think it's worth doing that?

lib/handler/src/test/scala/com/gu/util/apigateway/ApiGatewayResponseTest.scala
```diff
@@ -0,0 +1,21 @@
+package com.gu.util.apigateway
```
johnduffell: I wrote two tests, one for the existing and one for my chances.  But I thought even after deleting my changes it was worth keeping the other test.  Does seem a bit out of place in this PR though.

project/Dependencies.scala
```diff
@@ -1,6 +1,8 @@
 import sbt._
 
 object Dependencies {
+  
+  val awsVersion = "1.11.574"
```
johnduffell: I thought it was worth keeping this AWS update since I've noticed there is a new one.  Not all the effectsTests pass, because of changes to dev salesforce, but I the s3 stuff is working, the SQS effectsTest still passes, and the SES stuff doesn't seem to be used (any more?)

handlers/metric-push-api/src/main/scala/com/gu/metric_push_api/Handler.scala
johnduffell: I'm not checking this in the API gateway only version.  I'm sure it's possible to check based on my reading the docs a lot, but I am reluctant to try! 😴

johnduffell:
I just realised that we can't add a runscope test as we only have one endpoint/metric now.  I think with the reduced complexity maybe this isn't so necessary.

PR #333, opened 2019-06-26
 Enable fetching holiday-stop details from Salesforce

kelvin-chappell:
This will allow us to see what has already been written to Salesforce so that by comparing it with what's in Zuora we can see what needs to be backfilled.

The fetch method takes start and end threshold dates so that we can backfill in batches and then run on a schedule to pick up any holiday stops that have been manually applied in Zuora since a given date.

The data required for the comparison comes from across the holiday-stop table and the holiday-stop-zuora-action-refs table.  We need the subscription name and its child results.


mario-galic:
This backfill should be one-time thing, and not part of the main codebase. Agent's should be trained to go via new holiday stop service instead.

kelvin-chappell:
Following discussion with Jess and Mario I'm now going to write this slightly differently as part of a script that I can run from my machine.

PR #334, opened 2019-06-28
Backfill Salesforce with pre-existing holiday stops from Zuora

kelvin-chappell:
This code is for a one-off process to find pre-existing holiday stops in Zuora that don't already exist in Salesforce.  It then writes them to Salesforce, to the holiday stop requests table and to the child holiday stop zuora refs table.  This is to ensure that there's a single place where CSRs can look to find the complete history of a subscriber's holiday stops.

The script takes a date range to find stops within so that it's scalable and can potentially be re-run in the future.

I'm aware there aren't any unit tests as it's a one-off process.  It does have a dry-run capability so that it's possible to see in advance what it's going to write to Salesforce without actually writing it.


handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopbackfill/Config.scala
```diff
@@ -0,0 +1,64 @@
...
+import scala.io.Source
+
+case class Config(
+  zuoraConfig: ZuoraConfig,
+  sfConfig: SFAuthConfig
+)
+
+case class ZuoraConfig(
+  baseUrl: String,
+  holidayStopProcessor: HolidayStopProcessor
```
mario-galic: Why not  inline oauth like so 

```
case class ZuoraConfig(
  baseUrl: String,
  oauth: Oauth
)
```

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopbackfill/Config.scala
```diff
@@ -0,0 +1,64 @@
...
+      else AmazonS3Client.builder
+    val inputStream =
+      builder.build().getObject(bucketName, key).getObjectContent
+    val rawJson = Source.fromInputStream(inputStream).mkString
+    decode[T](rawJson).left map { e =>
+      ConfigFailure(s"Could not read secret config file from S3://$bucketName/$key: ${e.toString}")
+    }
+  }
+
+  def build(): Either[ConfigFailure, Config] = {
```
mario-galic: We could just construct with `Config()` instead of `Config.build()` if renamed to `apply`.

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopbackfill/HolidayStopDetail.scala
```diff
@@ -0,0 +1,133 @@
...
+import com.gu.util.Time
+import com.softwaremill.sttp.Response
+import io.circe.generic.auto._
+import io.circe.parser.decode
+
+case class ZuoraHolidayStop(subscriptionName: String, chargeNumber: String, startDate: LocalDate, endDate: LocalDate, creditPrice: Double)
+
+object ZuoraHolidayStop {
+
+  def holidayStopsAlreadyInZuora(queryResponse: String => Response[String])(startThreshold: LocalDate, endThreshold: Option[LocalDate]): Either[ZuoraFetchFailure, Seq[ZuoraHolidayStop]] = {
```
mario-galic: Maybe just `start/end` instead of `startThreshold/endThreshold`

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopbackfill/HolidayStopDetail.scala
```diff
@@ -0,0 +1,133 @@
...
+    for {
+      body <- response.body.left.map(ZuoraFetchFailure)
+      stop <- decodeMultiline(body)
+    } yield stop
+  }
+}
+
+object SalesforceHolidayStop {
+
+  def holidayStopsAlreadyInSalesforce(sfCredentials: SFAuthConfig)(startThreshold: LocalDate, endThreshold: Option[LocalDate]): Either[SalesforceFetchFailure, Seq[HolidayStopRequestDetails]] = {
```
mario-galic: Word `threshold` seems superfluous.

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopbackfill/Backfiller.scala
```diff
@@ -0,0 +1,30 @@
...
+
+object Backfiller {
+
+  /*
+   * This makes two passes to update Salesforce to ensure it's failsafe.
+   * If any call fails it should leave Salesforce in a consistent state.
+   * First, the holiday request table is updated, and then the zuora refs child table.
+   */
+  def backfill(startThreshold: LocalDate, endThreshold: Option[LocalDate], dryRun: Boolean): Either[BackfillFailure, Unit] = {
+    for {
```
mario-galic: Why is return type `Unit`? Can we return more information so we can log exact stops backfilled as part of `StandaloneApp` success.

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopbackfill/StandaloneApp.scala
```diff
@@ -0,0 +1,11 @@
+package com.gu.holidaystopbackfill
+
+import java.time.LocalDate
+
+// This is backfill app to be run from a dev machine.
+object StandaloneApp extends App {
+  Backfiller.backfill(LocalDate.of(2017, 4, 1), Some(LocalDate.of(2021, 6, 1)), dryRun = false) match {
+    case Left(failure) => println(s"Failed: $failure")
+    case Right(_) => println("Success!")
```
mario-galic: Can we log more info about specific stops.

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopbackfill/Queries.scala
```diff
@@ -0,0 +1,20 @@
+package com.gu.holidaystopbackfill
+
+import java.time.LocalDate
+
+object Queries {
```
mario-galic: Maybe move this into `ZuoraHolidayStop` as it is the only place where it is used.

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopbackfill/HolidayStopDetail.scala
```diff
@@ -0,0 +1,133 @@
...
+        NewHolidayStopRequest(
+          HolidayStopRequestStartDate(Time.toJodaDate(zuoraStop.startDate)),
+          HolidayStopRequestEndDate(Time.toJodaDate(zuoraStop.endDate)),
+          SubscriptionNameLookup(SubscriptionName(zuoraStop.subscriptionName))
+        )
+      }
+      .distinct
+  }
+
+  def zuoraRefsToBeBackfilled(inZuora: Seq[ZuoraHolidayStop], inSalesforce: Seq[HolidayStopRequestDetails]): Seq[HolidayStopRequestActionedZuoraRef] = {
```
mario-galic: There are dates calculations here. How do you know it works? Did you check off-by-ones?

kelvin-chappell:
Thanks @mario-galic.  I don't know enough about Zuora360 but in general I don't think it will ever be feasible to sync data automatically between two disparate system except very coursely.  So this kind of manual syncing is likely to be necessary occasionally.  As it's a one-off script it won't need to be maintained in principle.  The more fundamental problem is why we hold data in two separate systems that needs to be synced.  And that needs a lot more thought than this PR.

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopbackfill/Config.scala
kelvin-chappell: As discussed, this wouldn't work as this is to decode some Json in the form 
`... { holidayStopProcessor: ... { oauth : ...`.  
ie. it has a three-level nested structure.

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopbackfill/HolidayStopDetail.scala
kelvin-chappell: This was attempt to make it clear that these are the start and end dates of a search, rather than of a particular holiday stop.  But probably not helping so I'll shorten.

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopbackfill/Backfiller.scala
kelvin-chappell: Yes that's true.

PR #336, opened 2019-07-01
Tweak support client side alarm thresholds to reduce noise

jfsoul:
This PR should be just matching what we have running, where we have changed threshold handling to reduce noise.

### DIFF
--- a/handlers/metric-push-api/cfn.yaml
+++ b/handlers/metric-push-api/cfn.yaml
@@ -173,5 +173,6 @@ Resources:
         Period: 60
         EvaluationPeriods: 5
         Statistic: Sum
+        TreatMissingData: notBreaching
       DependsOn:
         - MetricPushAPI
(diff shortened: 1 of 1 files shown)


handlers/metric-push-api/cfn.yaml
```diff
@@ -168,9 +168,11 @@ Resources:
...
             Value: !FindInMap [StageMap, !Ref Stage, ApiName]
         ComparisonOperator: GreaterThanOrEqualToThreshold
-        Threshold: 3
+        Threshold: 2
+        DatapointsToAlarm: 3
         Period: 60
-        EvaluationPeriods: 1
+        EvaluationPeriods: 5
         Statistic: Sum
+        TreatMissingData: notBreaching
```
jlieb10: What does this key `TreatMissingData` mean, out of curiosity?

jfsoul: It defines what to do in the case that no events of the specified type have been recorded for the evaluation period.  E.g. for error alarms we generally treat missing data as good (not breaching), whereas for request count we might treat missing data as bad (breaching) because it is effectively the same as zero in these cases.

jlieb10: Thanks

PR #339, opened 2019-07-03
Kougios/417 add i movo contract salesforce data to the lake

kostas-kougios-guardian:
PR to download the imovo contract csv from salesforce.

The important changes are in cfn.yaml where I changed the statemachine to trigger the imovo-contract csv export too.

The other changes include EndToEndManualTest which is similar to other manual tests and does the whole salesforce-job --> wait ---> export csv to s3. I used that to get data.

I am not sure for the whole end to end wiring of this reg. bucket names.

### DIFF
--- a/handlers/sf-datalake-export/cfn.yaml
+++ b/handlers/sf-datalake-export/cfn.yaml
@@ -485,13 +485,6 @@ Resources:
              "objectName": "CardExpiry"
             }
           RoleArn: !GetAtt [ TriggerRole, Arn ]
-        - Arn: !Ref StateMachine
-          Id: !Sub trigger_sf_export-ImovoContract-${Stage}
-          Input: |
-            {
-             "objectName": "ImovoContract"
-            }
-          RoleArn: !GetAtt [ TriggerRole, Arn ]
   ScheduleRule2:
     Type: "AWS::Events::Rule"
     Condition: IsProd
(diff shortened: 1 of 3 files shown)


handlers/sf-datalake-export/src/test/scala/com/manual_test/EndToEndManualTest.scala
```diff
@@ -0,0 +1,96 @@
...
+    val request = Json.prettyPrint(Json.toJson(batches))
+
+    msg(request)
+
+    val testInputStream = new ByteArrayInputStream(request.getBytes)
+    val testOutput = new ByteArrayOutputStream()
+    DownloadBatchHandler(testInputStream, testOutput, null)
+  }
+
+  def msg(s: Any): Unit = {
```
wojtek-tg: Why not use something more sophisticated like a logger?

kostas-kougios-guardian: Oh these classes are just for devs to run locally (those are not really tests or anything, please see the other similar classes in that package). They use println() to dump useful info

wojtek-tg: Ok

handlers/sf-datalake-export/cfn.yaml
```diff
@@ -485,6 +485,13 @@ Resources:
              "objectName": "CardExpiry"
             }
           RoleArn: !GetAtt [ TriggerRole, Arn ]
+        - Arn: !Ref StateMachine
```
mario-galic: This is likely going to fail because there is a limit of 5 targets per `AWS::Events::Rule`. Probably, it should be put under new `ScheduleRule3` resource.

kostas-kougios-guardian: ok thanks, fixed

mario-galic:
Add the corresponding S3 bucket in ophan. See example https://github.com/guardian/ophan-data-lake/pull/3957/

mario-galic:
I think the bucket needs to be in production before this can be merged:
https://eu-west-1.console.aws.amazon.com/states/home?region=eu-west-1#/executions/details/arn:aws:states:eu-west-1:865473395570:execution:salesforce-export-PROD:3ad3b1ab-f68b-72ea-cb92-bde26e991b25_764c360f-1728-b101-a18d-a44995f43579

PR #341, opened 2019-07-05
Only export unique holiday-stop Zuora refs to Salesforce

kelvin-chappell:
Currently if the holiday-stop processor is run twice the same result set from Zuora will be exported to Salesforce so there will be a duplicate set of Zuora refs for every stop request in Salesforce.

This change should stop that from happening by fetching the pre-existing Zuora refs with each holiday request from Salesforce.  This way the processor can inspect and dedup the Zuora refs before exporting them to Salesforce.

### DIFF
--- a/lib/holiday-stops/src/main/scala/com/gu/salesforce/holiday_stops/SalesforceHolidayStopRequestActionedZuoraRef.scala
+++ b/lib/holiday-stops/src/main/scala/com/gu/salesforce/holiday_stops/SalesforceHolidayStopRequestActionedZuoraRef.scala
@@ -4,7 +4,7 @@ import java.time.LocalDate
 
 import ai.x.play.json.Jsonx
 import com.gu.salesforce.SalesforceConstants._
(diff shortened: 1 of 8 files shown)


lib/holiday-stops/src/main/scala/com/gu/salesforce/holiday_stops/SalesforceHolidayStopRequestActionedZuoraRef.scala
```diff
@@ -68,7 +88,9 @@ object SalesforceHolidayStopRequestActionedZuoraRef extends Logging {
     def toRequest(productNamePrefix: ProductName, startThreshold: LocalDate, endThreshold: LocalDate): GetRequestWithParams = {
       val soqlQuery =
         s"""
-          |select Holiday_Stop_Request__r.Id, Holiday_Stop_Request__r.Subscription_Name__c,
+          |select Holiday_Stop_Request__r.Id, Holiday_Stop_Request__r.Start_Date__c,
```
twrichards: could the common (with the new `LookupByProductNamePrefixAndDate`) parts of this query be combined into a `val` ??

PR #342, opened 2019-07-10
Fetch Holiday-stop requests from Salesforce with outer join

kelvin-chappell:
This changes the model that we fetch from Salesforce so that it now looks like:
```
  case class HolidayStopRequestDetails(
    request: HolidayStopRequest,
    zuoraRefs: Option[Seq[ZuoraRef]]
  )
```

This fixes a bug I introduced in #341 where a holiday stop that had no existing Zuora refs would not be fetched from Salesforce to be processed.


twrichards:
Makes sense to fix the bug, but will be redundant when we move to pre-populating that refs table for every publication date of the holiday stop, that we decided in the meeting yesterday. Not sure if it's worth us just biting the bullet and making that change now.

kelvin-chappell:
@twrichards I'd like to fix the bug anyway so that the code's always in some kind of working state.  But I agree that should go on to make the change in functionality.

PR #344, opened 2019-07-12
Log log4j messages through logback

kelvin-chappell:
This makes the logback configuration work correctly and avoids the warning message:
```
log4j:WARN No appenders could be found for logger (com.amazonaws.AmazonWebServiceClient).
log4j:WARN Please initialize the log4j system properly.
log4j:WARN See http://logging.apache.org/log4j/1.2/faq.html#noconfig for more info.
```

[Difference between slf4j-log4j12 and log4j-over-slf4j:](https://stackoverflow.com/a/31044844/5205022)

> **log4j-over-slf4j**
> 
> Use this if your code or some libraries you are using uses Log4j directly, but you want to use a different SLF4J binding than Log4j. It will route the Log4j API calls to SLF4J to the binding you choose. You need to remove the Log4j library from your classpath and replace it with this dependency.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -74,7 +74,7 @@ lazy val zuora = all(project in file("lib/zuora"))
     effects % "test->test"
   )
   .settings(
-    libraryDependencies ++= Seq(okhttp3, logging, scalaz, playJson, scalatest, jacksonDatabind)
+    libraryDependencies ++= Seq(okhttp3, scalaz, playJson, scalatest, jacksonDatabind) ++ logging
   )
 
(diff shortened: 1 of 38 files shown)


arausuy:
I'm assuming the introduction of logback classic fixed the tc builds?

kelvin-chappell:
@arausuy Yes but unfortunately found more problems with AWS's lambda logger, which goes through Log4J and wasn't working properly in cloudwatch logs. 
This is now superseded by #344 which doesn't use the lambda logger at all.

PR #345, opened 2019-07-16
Replace log4j with logback

mario-galic:
[ Log4J has been deprecated by Logback.](https://stackoverflow.com/a/2160175/5205022)

### DIFF
--- a/lib/effects-ses/src/main/scala/com/gu/effects/ses/AwsSESSend.scala
+++ b/lib/effects-ses/src/main/scala/com/gu/effects/ses/AwsSESSend.scala
@@ -5,15 +5,13 @@ import com.amazonaws.auth.profile.ProfileCredentialsProvider
 import com.amazonaws.regions.Regions
 import com.amazonaws.services.simpleemail.AmazonSimpleEmailServiceAsyncClientBuilder
 import com.amazonaws.services.simpleemail.model._
-import org.apache.log4j.Logger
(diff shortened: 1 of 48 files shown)


lib/effects-ses/src/main/scala/com/gu/effects/ses/AwsSESSend.scala
```diff
@@ -5,15 +5,13 @@ import com.amazonaws.auth.profile.ProfileCredentialsProvider
 import com.amazonaws.regions.Regions
 import com.amazonaws.services.simpleemail.AmazonSimpleEmailServiceAsyncClientBuilder
 import com.amazonaws.services.simpleemail.model._
-import org.apache.log4j.Logger
+import com.typesafe.scalalogging.{LazyLogging, Logger}
```
arausuy: Do we need to import logger & lazy logging here?

johnduffell:
nice work updating this.  Just came to steal the LogImplicit class and noticed it was all updated! 👍

lib/restHttp/src/main/scala/com/gu/util/resthttp/Logging.scala
```diff
@@ -1,9 +1,2 @@
 package com.gu.util.resthttp
```
johnduffell: probably could have deleted this whole file as it's now empty

PR #349, opened 2019-07-22
Restructure holiday-stop detail table

kelvin-chappell:
https://trello.com/c/7enQ31Y8/197-update-model-of-holiday-stop-request-details-in-sf

I've made enough changes to get the holiday-stop processor working.
There's a little more work to do on the backfiller but will do that in another PR.

Also I've temporarily added the manual creation of a holiday-stop detail to the `SalesforceHolidayStopRequestEndToEndEffectsTest` to ensure it passes.  
(As the triggering mechanism for holiday stop details isn't complete yet.)

All the work to change the model was done by @twrichards...
> Previously model in Salesforce was a parent/master object `Holiday_Stop_Request__c` with child/detail object `Holiday_Stop_Request_Actioned_Zuora_Ref__c` which ONLY contained publications/issues which had been _actioned_ (credited in Zuora and therefore prevented fulfilment).

> **NOW** this child/detail object has been renamed to `Holiday_Stop_Requests_Detail__c` and restructured to contain ALL publications/issues for the parent holiday stop request (both _actioned_ and _pending_), changes specifically...
> - object definition (names, description, record name, auto-number display format)
> - `Charge_Code__c` is now nullable - since it's only for those in actioned state
> - new formula field `Is_Actioned__c` which is false if `Charge_Code__c` is blank
> - renamed field `Price__c` to `Actual_Price__c` since that's written by the `holiday-stop-processor` when the credit is actually applied
> - added field `Estimate_Price__c` as a placeholder for when we do a preview amendment at creation/edit time
> - new formula fields for pulling details from parent (`Subscription_Name__c` and `Product_Name__c`)
> - various layout tweaks

> PLUS the parent/master object `Holiday_Stop_Request__c` required some accompanying changes...
> - relationship name to child now `Holiday_Stop_Request_Detail__r` (to reflect rename)
> - changed roll-up summary field `Actioned_Count__c` to count only those children which have `Is_Actioned__c` = true
> - new roll-up summary field `Pending_Count__c ` which counts the children which have `Is_Actioned__c` = false
> - new  roll-up summary field `Total_Issues_Publications_Impacted_Count__c` which counts ALL the children
> - various layout tweaks

PR #351, opened 2019-07-23
Implement holiday-stop backfill with new detail model

kelvin-chappell:
This fixes some problems introduced to the backfiller by the changes to the holiday-stop detail model in #349.
The main changes here are:
* Fetching the holiday-stop request with its child details rather than the detail records with their parent
* There's a new  `BackfillActionedSalesforceHolidayStopRequestsDetail` object specifically for the backfiller to use to create a new actioned holiday-stop detail
* Removed the date ranges because that was getting quite confusing and there's only about 2000 legacy holiday stops to add (will need to consider a batching operation when backfilling other products).

### DIFF
--- a/handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopbackfill/HolidayStopDetail.scala
+++ b/handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopbackfill/HolidayStopDetail.scala
@@ -6,7 +6,7 @@ import java.time.{DayOfWeek, LocalDate}
 
 import com.gu.salesforce.SalesforceAuthenticate.SFAuthConfig
 import com.gu.salesforce.holiday_stops.SalesforceHolidayStopRequest.{HolidayStopRequest, HolidayStopRequestEndDate, HolidayStopRequestStartDate, NewHolidayStopRequest, SubscriptionNameLookup}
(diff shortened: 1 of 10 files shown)


handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopbackfill/HolidayStopDetail.scala
```diff
@@ -115,16 +108,35 @@ object SalesforceHolidayStop {
      * There should be a request ID available for each subscription and stopped publication date
      * as in the first pass the parent holiday requests will have been populated.
      */
-    def correspondingRequestId(zStop: ZuoraHolidayStop): Option[HolidayStopRequestId] = None
+    def correspondingRequest(zStop: ZuoraHolidayStop): Option[HolidayStopRequest] = {
+      inSalesforce find { sfStop =>
+        val startDate = Time.toJavaDate(sfStop.Start_Date__c.value)
```
mario-galic: Remind me again why are we using multiple date libraries?

kelvin-chappell: Because joda time is now deprecated.
See https://www.joda.org/joda-time
> Note that from Java SE 8 onwards, users are asked to migrate to java.time (JSR-310) - a core part of the JDK which replaces this project.

PR #352, opened 2019-07-24
Enable holiday-stop amendments to be applied to annual subscriptions

kelvin-chappell:
This fixes a bug where amendments could not be applied to annual subscriptions because their term-end date falls before the first day of their next billing period, which is the effective date of the amendment.  Zuora will not allow an amendment to be effective after the term of the subscription ends.

In these cases, the subscription is now updated to extend the term by the number of days needed to fit the effective date into the current term of the subscription.

This is an example of the amendment that's applied:

```
{
	"currentTerm": "376",
	"currentTermPeriodType": "Day",
        "add": [
        {
            "productRatePlanId": "{{productRatePlanId}}",
            "contractEffectiveDate": "2020-08-02",
            "customerAcceptanceDate": "2020-08-02",
            "serviceActivationDate": "2020-08-02",
            "chargeOverrides": [
                {
                    "productRatePlanChargeId": "{{productRatePlanChargeId}}",
                    "HolidayStart__c": "2019-08-06",
                    "HolidayEnd__c": "2019-08-06",
                    "price": "-12.34"
                }
            ]
        }
    ]
}
```

### DIFF
--- a/handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/SubscriptionUpdate.scala
+++ b/handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/SubscriptionUpdate.scala
@@ -2,7 +2,11 @@ package com.gu.holidaystopprocessor
 
 import java.time.LocalDate
 
-case class SubscriptionUpdate(add: Seq[Add]) {
+case class SubscriptionUpdate(
+  currentTerm: Option[Int],
+  currentTermPeriodType: Option[String],
+  add: Seq[Add]
+) {
 
   val price: Double = {
     val optPrice = for {
@@ -20,9 +24,19 @@ object SubscriptionUpdate {
     subscription: Subscription,
     stoppedPublicationDate: LocalDate
   ): Either[HolidayStopFailure, SubscriptionUpdate] = {
+
+    case class ExtendedTerm(length: Int, unit: String)
+
     subscription.originalRatePlanCharge.flatMap(_.chargedThroughDate)
       .toRight(HolidayStopFailure("Original rate plan charge has no charged through date.  A bill run is needed to fix this.")).map { effectiveDate =>
+
+        val extendedTerm: Option[ExtendedTerm] =
+          if (effectiveDate.isAfter(subscription.termEndDate)) Some(ExtendedTerm(366, "Day"))
+          else None
+
         SubscriptionUpdate(
+          currentTerm = extendedTerm.map(_.length),
+          currentTermPeriodType = extendedTerm.map(_.unit),
           Seq(
             Add(
               productRatePlanId = config.holidayCreditProductRatePlanId,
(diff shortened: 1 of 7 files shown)


kelvin-chappell:
On reflection needs to extend term by (effective date - term start date) days rather than one day.

mario-galic:
Check invoice makes sense after term extension by doing a bill run with target date in the next annual billing period.

kelvin-chappell:
This is an example of the amendment that's applied:
```
{
	"currentTerm": "376",
	"currentTermPeriodType": "Day",
        "add": [
        {
            "productRatePlanId": "{{productRatePlanId}}",
            "contractEffectiveDate": "2020-08-02",
            "customerAcceptanceDate": "2020-08-02",
            "serviceActivationDate": "2020-08-02",
            "chargeOverrides": [
                {
                    "productRatePlanChargeId": "{{productRatePlanChargeId}}",
                    "HolidayStart__c": "2019-08-06",
                    "HolidayEnd__c": "2019-08-06",
                    "price": "-12.34"
                }
            ]
        }
    ]
}
```

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/SubscriptionUpdate.scala
```diff
@@ -20,9 +25,23 @@ object SubscriptionUpdate {
     subscription: Subscription,
     stoppedPublicationDate: LocalDate
   ): Either[HolidayStopFailure, SubscriptionUpdate] = {
+
+    case class ExtendedTerm(length: Int, unit: String)
+
     subscription.originalRatePlanCharge.flatMap(_.chargedThroughDate)
       .toRight(HolidayStopFailure("Original rate plan charge has no charged through date.  A bill run is needed to fix this.")).map { effectiveDate =>
+
+        val extendedTerm: Option[ExtendedTerm] =
```
mario-galic: Could we factor term extension concept into its own file perhaps like so

```
case class ExtendedTerm(length: Int, unit: String)
object ExtendedTerm {
  def apply(effectiveDate: LocalDate, subscription: Subscription): Option[ExtendedTerm] = {
      if (effectiveDate.isAfter(subscription.termEndDate)) {
        Some(ExtendedTerm(
          length = DAYS.between(subscription.termStartDate, effectiveDate).toInt,
          unit = "Day"
        ))
      } else None
  }
}
```

and document it why it is necessary by pasting in the exact api error we get without the extension. 

Separately, we should probably have a short meeting to discuss if there is a more general solution for term extension problem, beside coding it every time we add new feature.

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/SubscriptionUpdate.scala
```diff
@@ -20,9 +25,23 @@ object SubscriptionUpdate {
...
+          if (effectiveDate.isAfter(subscription.termEndDate)) {
```
mario-galic: Predicate should probably be `isAfter or equal` because `chargedThroughDate` is **exclusive**, that is, `chargedThroughDate` is actually the first day of the next invoiced period.

PR #353, opened 2019-07-25
Make names more specific for Holiday Credit RatePlanCharge update

mario-galic:
* `SubscriptionUpdate` -> `HolidayCreditUpdate`
* Remove some dead code


kelvin-chappell:
Perhaps leave `HolidayCredit` as is, because a credit is an amount.

PR #302, opened 2019-05-30
holiday-stop-api

twrichards:
## new handler : `holiday-stop-api`

This PR is mainly just the CloudFormation to present the various routes and wire them into the functions in the `holiday-stops` shared library (see https://github.com/guardian/support-service-lambdas/pull/294), but includes some necessary refactors along the way.

All endpoints require...

- an `x-api-key` header (generated by the `HolidayStopApiKey` in the CloudFormation)
- an `x-identity-id` header which specifies the identityID of the user to request data for

| Method | Endpoint | Description |
| --- | --- | --- | 
| GET | `/{STAGE}/potential?startDate={yyyy-MM-dd}&endDate={yyyy-MM-dd}`  (with `x-product-name-prefix` header set)  | returns an array of dates for each issue impacted between the dates for the given product |
| GET | `/{STAGE}/hsr` | returns all holiday stops (past & present) for the user |
| GET | `/{STAGE}/hsr` (with `x-product-name-prefix` header set) | returns all holiday stops (past & present) but including a calculated 'first available date' based on the type of product |
| GET | `/{STAGE}/hsr /{SUBSCRIPTION_NAME}` (with `x-product-name-prefix` header set) | returns all holiday stops (past & present) for the user filtered on the specified subscription also including a calculated 'first available date' based on the type of product |
| POST | `/{STAGE}/hsr` | creates a new all holiday stop, example body `{ "start": "2023-06-10", "end": "2024-06-14", "subscriptionName": "A-S00071783" }`|
| DELETE | `/{STAGE}/hsr /{SUBSCRIPTION_NAME} /{SF_ID}` | deletes the holiday stop request from SalesForce (with Id matching `{SF_ID}`) |

PR #358, opened 2019-07-26
replace identity-cookie with identity-auth-core

Mullefa:
When authenticating a user, we want to check that their session hasn't been invalidated. To do this, we need to callback to the identity API. This PR introduces the library identity-auth-core to do this.

Raised as a draft PR since there are some questions (inlined as PR comments) which I'd appreciate answers to.

### DIFF
--- a/handlers/cancellation-sf-cases/cfn.yaml
+++ b/handlers/cancellation-sf-cases/cfn.yaml
@@ -10,6 +10,7 @@ Parameters:
             - CODE
             - DEV
         Default: CODE
+    IdentityApiServerAccessToken:
 
 Conditions:
   CreateProdMonitoring: !Equals [ !Ref Stage, PROD ]
@@ -87,6 +88,7 @@ Resources:
             Environment:
                 Variables:
                   Stage: !Ref Stage
+                  IdentityApiServerAccessToken: !Ref IdentityApiServerAccessToken
             Role:
                 Fn::GetAtt:
                 - "CancellationSFCasesRole"
--- a/handlers/cancellation-sf-cases/cfn.yaml
+++ b/handlers/cancellation-sf-cases/cfn.yaml
@@ -11,6 +11,8 @@ Parameters:
             - DEV
         Default: CODE
     IdentityApiServerAccessToken:
+        Description: server access token for identity API
+        Type: String
 
 Conditions:
   CreateProdMonitoring: !Equals [ !Ref Stage, PROD ]
--- a/handlers/cancellation-sf-cases/src/main/scala/com/gu/identity/IdentityCookieToIdentityUser.scala
+++ b/handlers/cancellation-sf-cases/src/main/scala/com/gu/identity/IdentityCookieToIdentityUser.scala
@@ -1,13 +1,14 @@
 package com.gu.identity
 
-import com.gu.identity.cookie.{IdentityCookieDecoder, PreProductionKeys, ProductionKeys}
+import com.gu.identity.auth.{IdentityAuthService, UserCredentials}
 import com.gu.util.Logging
 import com.gu.util.apigateway.ApiGatewayResponse._
 import com.gu.util.reader.Types._
+import org.http4s.Uri
 
 object IdentityCookieToIdentityUser extends Logging {
 
-  type CookieValuesToIdentityUser = (String, String) => Option[IdentityUser]
+  type CookieValuesToIdentityUser = String => Option[IdentityUser]
 
   final case class IdentityId(value: String) extends AnyVal
 
@@ -22,8 +23,7 @@ object IdentityCookieToIdentityUser extends Logging {
       headers <- headersOption.toApiGatewayContinueProcessing(badRequest("no headers"))
       cookieHeader <- headers.get("Cookie").toApiGatewayContinueProcessing(badRequest("no cookie"))
       scGuU <- extractCookieHeaderValue(cookieHeader, "SC_GU_U")
-      guU <- extractCookieHeaderValue(cookieHeader, "GU_U")
-      identityUser <- cookiesToIdentityUser(scGuU, guU).toApiGatewayContinueProcessing(unauthorized)
+      identityUser <- cookiesToIdentityUser(scGuU).toApiGatewayContinueProcessing(unauthorized)
     } yield identityUser
 
   private def extractCookieHeaderValue(cookieHeader: String, specificCookieName: String): ApiGatewayOp[String] = {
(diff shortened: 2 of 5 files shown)


handlers/cancellation-sf-cases/cfn.yaml
```diff
@@ -10,6 +10,9 @@ Parameters:
             - CODE
             - DEV
         Default: CODE
+    IdentityApiServerAccessToken:
```
Mullefa: To make a server access token for the identity API available to the lambda, I'm introducing it as a parameter in the cloud formation and then using this parameter to set a lambda environment variable. Is this the preferred way to expose this server access token, or should I do something else?

handlers/cancellation-sf-cases/src/main/scala/com/gu/identity/IdentityCookieToIdentityUser.scala
```diff
@@ -35,14 +35,34 @@ object IdentityCookieToIdentityUser extends Logging {
...
-      userFromGuU <- cookieDecoder.getUserDataForGuU(guU)
-      displayName = if (userFromScGuU.id equals userFromGuU.getUser.id) userFromGuU.getUser.publicFields.displayName else None
-    } yield IdentityUser(IdentityId(userFromScGuU.id), displayName)
+  private def getIdentityApiUri(isProd: Boolean): Uri = {
+    val raw = if (isProd) "https://idapi.theguardian.com" else "https://idap.code.dev-theguardian.com"
+    Uri.unsafeFromString(raw)
   }
 
+  private def serverAccessToken: String =
+    Option(System.getenv("IdentityApiServerAccessToken"))
```
Mullefa: If the server access token isn't available I'm throwing an exception; is this ok?

handlers/cancellation-sf-cases/cfn.yaml
johnduffell: if it's something needed in DEV/CODE/PROD, normally we would add a relevant config file in s3 and read that in as part of startup.  This would make it easier to cloudform as it's all in one place.
Since configmanager? parameter store? is the latest thing, you could put it there and add a reader for config from there if you fancy pioneering it.  As all the config is self contained that wouldn't require refactoring the other config at this stage.

handlers/cancellation-sf-cases/src/main/scala/com/gu/identity/IdentityCookieToIdentityUser.scala
```diff
@@ -1,13 +1,14 @@
...
+import com.gu.identity.auth.{IdentityAuthService, UserCredentials}
 import com.gu.util.Logging
 import com.gu.util.apigateway.ApiGatewayResponse._
 import com.gu.util.reader.Types._
+import org.http4s.Uri
 
 object IdentityCookieToIdentityUser extends Logging {
 
-  type CookieValuesToIdentityUser = (String, String) => Option[IdentityUser]
+  type CookieValuesToIdentityUser = String => Option[IdentityUser]
```
johnduffell: this is good to simplify because a (String, String) is easy to get backwards! 💯

twrichards:
Sorry only just seen this PR - I need to get round to refactoring out the dependency on identity all together from this repo. Since it's proxied by `manage-frontend` anyway... `manage` can pick the right env. instance of this lambda based on info it already knows and authenticate via API key (like in holiday-stops-api - see https://github.com/guardian/support-service-lambdas/pull/302 and https://github.com/guardian/manage-frontend/pull/220).

I will get round to this in the next couple of weeks and it will significantly simplify this lambda 🎉 -- so it's up to you if you need to merge this change in the meantime or your schedules allow for just closing this PR in the knowledge it will go away anyway soon??

handlers/cancellation-sf-cases/src/main/scala/com/gu/identity/IdentityCookieToIdentityUser.scala
johnduffell: @twrichards knows more about the setup of this particular lambda, personally I'd not use them, in the ones I set up and the libraries I avoided exceptions because they don't appear in the type signature.  However in this case as long as the exception causes it to fail the health check it would be ok.

handlers/cancellation-sf-cases/src/main/scala/com/gu/identity/IdentityCookieToIdentityUser.scala
```diff
@@ -35,14 +35,34 @@ object IdentityCookieToIdentityUser extends Logging {
...
+
+  def defaultCookiesToIdentityUser(isProd: Boolean)(scGuU: String): Option[IdentityUser] = {
+    val uri = getIdentityApiUri(isProd)
+
+    val identityAuthService = {
+      import scala.concurrent.ExecutionContext.Implicits.global
+      IdentityAuthService.unsafeInit(uri, serverAccessToken)
+    }
+
+    identityAuthService.getUserFromCredentials(UserCredentials.SCGUUCookie(scGuU))
```
johnduffell: I think the biggest departure from my point of view is that the previous function was a pure function and any side effect was in the handler.  This is to make it possible to unit/component test without accidentally calling out to any external services.  The strategy I generally used in the lambdas was to pass in the minimal side effect function from the top, e.g. `IdentityCookie => Option[IdentityUser]`
This just lets us make more assuptions about what setup is needed for writing tests (i.e. the compiler will check everything) and less manual setup needed e.g. env vars or mocking things that won't be caught.

johnduffell:
sorry didn't spot Tom's comment, I think that makes mine a bit more reduandant!

Mullefa:
@twrichards - thanks for the update. On refactor, how will the lambda authenticate a user?

Mullefa:
Closing this PR as at some point in the next couple of weeks it will be refactored to handle user authentication in the same way as the rest of the lambdas in this repo: check their status using the user status endpoint in identity API and if they're signed in call the respective lambda. 

@twrichards is very busy atm, so if the work hasn't been done within a couple of weeks we can consider re-opening this PR, or doing the re-factor for him.

twrichards:
> @twrichards - thanks for the update. On refactor, how will the lambda authenticate a user?

Promised I would reply to clarify this. 

So `manage-frontend` (the only consumer of this `cancellation-sf-cases` lambda, which is the only thing with the `identity-auth` dependency) proxies calls to this lambda/API-Gateway, and all calls to `manage-frontend` are first checked via the `/redirect` endpoint in `IDAPI` (and hence authenticated - if not signed in in the last hour they're redirected to signin/reauthenticate).
The refactor I will be doing soon, is to instead of forwarding the cookies to the lambda so it can authenticate, manage-frontend simply communicates with lambda with API key and passes identity id in a header, after manage has used IDAPI /redirect to authenticate the client.
This removes the need for anything in this repo to need identity-auth.

Mullefa:
thanks for clarifying @twrichards 👌

twrichards:
the refactor that eliminates the identity dependencies from this lambda is finally being worked on - see https://github.com/guardian/support-service-lambdas/pull/540

twrichards:
Finally released https://github.com/guardian/support-service-lambdas/pull/540 as promised (7 months ago 😉)

PR #363, opened 2019-07-30
Document IntegrationHttpMethod needs to be POST

twrichards:
The `IntegrationHttpMethod` (between API Gateway with Lambda) needs to always be `POST` so this change makes that explicit in all the places it's used. 

**Because https://stackoverflow.com/a/55457240/5205022**

_Pulling out some of the changes in https://github.com/guardian/support-service-lambdas/pull/302 to simplify it_

### DIFF
--- a/cloudformation.yaml
+++ b/cloudformation.yaml
@@ -160,7 +160,7 @@ Resources:
               method.request.querystring.apiToken: true
             Integration:
               Type: AWS_PROXY
-              IntegrationHttpMethod: POST
+              IntegrationHttpMethod: POST # this for the interaction between API Gateway and Lambda and MUST be POST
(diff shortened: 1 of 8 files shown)


cloudformation.yaml
```diff
@@ -160,7 +160,7 @@ Resources:
               method.request.querystring.apiToken: true
             Integration:
               Type: AWS_PROXY
-              IntegrationHttpMethod: POST
+              IntegrationHttpMethod: POST # this for the interaction between API Gateway and Lambda and MUST be POST
```
mario-galic: Add a link to where this is documented, for example.

```
Must be POST because https://stackoverflow.com/a/55457240/5205022
```

twrichards: done in the `HOWTO-create-lambda.md` README - to avoid placing it everywhere

PR #368, opened 2019-07-31
Replace Seq with List to unlock cats

mario-galic:
No typeclasses for Seq #1222: https://github.com/typelevel/cats/issues/1222, which forces unnecessary `toList`, for example, `s.toList.sequence`,


johnduffell:
sounds like a good idea, I never know the difference between List and Seq apart from the ability to pattern match!  And it would be interesting to have cats rather than scalaz in some parts of the repo, I've never seen them coexist in this way in a repo! 👍

PR #371, opened 2019-08-06
Calculate credit amount as a function of current guardian weekly subscription

mario-galic:
Main change is in `HolidayCredit.scala`.

Calculating holiday credit amount depends on picking the correct guardian weekly rate plan. Since subscription can have arbitrary number of arbitrary rate plans we cannot simply depend on sorting by date and picking the earliest.

This PR introduces the concept of `CurrentGuardianWeeklySubscription` which is entity that satisfies the following predicates

```
sealed trait CurrentGuardianWeeklyRatePlanCondition
case object RatePlanIsGuardianWeekly extends CurrentGuardianWeeklyRatePlanCondition
// case object TodayHasBeenInvoiced extends CurrentGuardianWeeklyRatePlanCondition // FIXME: This is a stronger check than RatePlanHasBeenInvoiced but requires significant refactoring of tests
case object RatePlanHasBeenInvoiced extends CurrentGuardianWeeklyRatePlanCondition
case object RatePlanHasACharge extends CurrentGuardianWeeklyRatePlanCondition
case object RatePlanHasOnlyOneCharge extends CurrentGuardianWeeklyRatePlanCondition
case object ChargeIsQuarterlyOrAnnual extends CurrentGuardianWeeklyRatePlanCondition
```

Most important is `RatePlanIsGuardianWeekly` which is checked against hardcoded list of `productRatePlanIds`.

This complication is a result of decision to model things like discounts as products which makes it much harder determining what actual product the users has.

Note there are few FIXMEs here. In particular the following **must** be addressed in separate PR:

```
// FIXME: This should be function of CurrentGuardianWeeklySubscription
object NextBillingPeriodStartDate
```

### DIFF
--- a/handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/CurrentGuardianWeeklySubscription.scala
+++ b/handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/CurrentGuardianWeeklySubscription.scala
@@ -10,8 +10,7 @@ sealed trait CurrentGuardianWeeklyRatePlanCondition
 case object RatePlanIsGuardianWeekly extends CurrentGuardianWeeklyRatePlanCondition
 // case object TodayHasBeenInvoiced extends CurrentGuardianWeeklyRatePlanCondition // FIXME: This is a stronger check than RatePlanHasBeenInvoiced but requires significant refactoring of tests
 case object RatePlanHasBeenInvoiced extends CurrentGuardianWeeklyRatePlanCondition
-case object RatePlanHasACharge extends CurrentGuardianWeeklyRatePlanCondition
-case object RatePlanHasOnlyOneCharge extends CurrentGuardianWeeklyRatePlanCondition
+case object RatePlanHasExactlyOneCharge extends CurrentGuardianWeeklyRatePlanCondition
 case object ChargeIsQuarterlyOrAnnual extends CurrentGuardianWeeklyRatePlanCondition
 
 /**
@@ -75,8 +74,8 @@ object CurrentGuardianWeeklySubscription {
       .find { ratePlan =>
         List[(CurrentGuardianWeeklyRatePlanCondition, Boolean)](
           RatePlanIsGuardianWeekly -> guardianWeeklyProductRatePlanIds.contains(ratePlan.productRatePlanId),
-          RatePlanHasACharge -> ratePlan.ratePlanCharges.nonEmpty,
-          RatePlanHasOnlyOneCharge -> (ratePlan.ratePlanCharges.size == 1),
+          //          RatePlanHasACharge -> ratePlan.ratePlanCharges.nonEmpty,
+          RatePlanHasExactlyOneCharge -> (ratePlan.ratePlanCharges.size == 1),
           // FIXME: Enable this after test refactoring - the problem is LocalDate.now() call
           //          TodayHasBeenInvoiced ->
           //            Try {
--- a/handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/CurrentGuardianWeeklySubscription.scala
+++ b/handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/CurrentGuardianWeeklySubscription.scala
@@ -7,11 +7,38 @@ import scala.util.Try
  * Conditions defining what Guardian Weekly subscription the customer has today.
  */
 sealed trait CurrentGuardianWeeklyRatePlanCondition
-case object RatePlanIsGuardianWeekly extends CurrentGuardianWeeklyRatePlanCondition
-// case object TodayHasBeenInvoiced extends CurrentGuardianWeeklyRatePlanCondition // FIXME: This is a stronger check than RatePlanHasBeenInvoiced but requires significant refactoring of tests
-case object RatePlanHasBeenInvoiced extends CurrentGuardianWeeklyRatePlanCondition
-case object RatePlanHasExactlyOneCharge extends CurrentGuardianWeeklyRatePlanCondition
-case object ChargeIsQuarterlyOrAnnual extends CurrentGuardianWeeklyRatePlanCondition
+case object RatePlanIsGuardianWeekly extends CurrentGuardianWeeklyRatePlanCondition {
+  def apply(ratePlan: RatePlan, guardianWeeklyProductRatePlanIds: List[String]): Boolean =
+    guardianWeeklyProductRatePlanIds.contains(ratePlan.productRatePlanId)
+}
+case object RatePlanHasBeenInvoiced extends CurrentGuardianWeeklyRatePlanCondition {
+  def apply(ratePlan: RatePlan): Boolean = {
+    Try {
+      val fromInclusive = ratePlan.ratePlanCharges.head.processedThroughDate.get
+      val toExclusive = ratePlan.ratePlanCharges.head.chargedThroughDate.get
+      toExclusive.isAfter(fromInclusive)
+    }.getOrElse(false)
+  }
+}
+case object RatePlanHasExactlyOneCharge extends CurrentGuardianWeeklyRatePlanCondition {
+  def apply(ratePlan: RatePlan): Boolean = (ratePlan.ratePlanCharges.size == 1)
+}
+case object ChargeIsQuarterlyOrAnnual extends CurrentGuardianWeeklyRatePlanCondition {
+  def apply(ratePlan: RatePlan): Boolean =
+    Try {
+      List("Annual", "Quarter").contains(ratePlan.ratePlanCharges.head.billingPeriod.get)
+    }.getOrElse(false)
+}
+//case object TodayHasBeenInvoiced extends CurrentGuardianWeeklyRatePlanCondition { // FIXME: This is a stronger check than RatePlanHasBeenInvoiced but requires significant refactoring of tests
+//  def apply(ratePlan: RatePlan): Boolean =
+//    Try {
+//      PeriodContainsDate(
+//        startPeriodInclusive = ratePlan.ratePlanCharges.head.processedThroughDate.get,
+//        endPeriodExcluding = ratePlan.ratePlanCharges.head.chargedThroughDate.get,
+//        date = LocalDate.now()
+//      )
+//    }.getOrElse(false)
+//}
 
 /**
  * Model representing 'What Guardian Weekly does the customer have today?'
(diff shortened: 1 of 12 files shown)


handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/HolidayCredit.scala
```diff
@@ -2,18 +2,28 @@ package com.gu.holidaystopprocessor
...
   }
 }
+
+// FIXME: Is this assumption safe?
+object BillingPeriodToApproxWeekCount {
+  def apply(billingPeriod: String): Int =
+    billingPeriod match {
+      case "Quarter" => 13
+      case "Annual" => 52
+      case _ => 52
```
twrichards: This default case scares me 😱 ... at the very least we should have a case for `"Monthly"` with a value of `52/12` (not a rounded number, since it will be used in a division anyway) - not too relevant for GW because we don't off monthly GW (but actually we probably have some in practice 😢 )
But more generally I'd perhaps rather not have a default case and let it `MatchError` so it at least it fails/alerts if we have unexpected subscriptions

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/CurrentGuardianWeeklySubscription.scala
```diff
@@ -0,0 +1,110 @@
...
+ * Is date between two dates where including the start while excluding the end?
+ */
+object PeriodContainsDate extends ((LocalDate, LocalDate, LocalDate) => Boolean) {
+  def apply(
+    startPeriodInclusive: LocalDate,
+    endPeriodExcluding: LocalDate,
+    date: LocalDate
+  ): Boolean =
+    (date.isEqual(startPeriodInclusive) || date.isAfter(startPeriodInclusive)) && date.isBefore(endPeriodExcluding)
+}
```
kelvin-chappell: This function looks like it's at a different level of abstraction, as it has no reference to Guardian Weekly so could perhaps go elsewhere in a `Dates` object of some sort.

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/CurrentGuardianWeeklySubscription.scala
```diff
@@ -0,0 +1,110 @@
...
+
+/**
+ * Invoiced period defined by [startDateIncluding, endDateExcluding) specifies the current period for which
+ * the customer has been billed. Today must be within this period.
+ *
+ * @param startDateIncluding service active on startDateIncluding; corresponds to processedThroughDate
+ * @param endDateExcluding service stops one endDateExcluding; corresponds to chargedThroughDate
```
kelvin-chappell:
```suggestion
 * @param endDateExcluding service ends on endDateExcluding; corresponds to chargedThroughDate
```

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/CurrentGuardianWeeklySubscription.scala
```diff
@@ -0,0 +1,110 @@
...
+  endDateExcluding: LocalDate
+) {
+  private val todayIsWithinCurrentInvoicedPeriod: Boolean = PeriodContainsDate(
+    startPeriodInclusive = startDateIncluding,
+    endPeriodExcluding = endDateExcluding,
+    date = LocalDate.now()
+  )
+  // FIXME: Enable this after test refactoring
+  //  require(todayIsWithinCurrentInvoicedPeriod, "Today should be within [startDateIncluding, endDateExcluding)")
+}
```
kelvin-chappell: Surely if invoiced period is current it will always include today?

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/CurrentGuardianWeeklySubscription.scala
```diff
@@ -0,0 +1,110 @@
...
+ */
+object CurrentGuardianWeeklySubscription {
+  def apply(subscription: Subscription, guardianWeeklyProductRatePlanIds: List[String]): CurrentGuardianWeeklySubscription =
+    subscription
+      .ratePlans
+      .find { ratePlan =>
+        List[(CurrentGuardianWeeklyRatePlanCondition, Boolean)](
+          RatePlanIsGuardianWeekly -> guardianWeeklyProductRatePlanIds.contains(ratePlan.productRatePlanId),
+          RatePlanHasACharge -> ratePlan.ratePlanCharges.nonEmpty,
+          RatePlanHasOnlyOneCharge -> (ratePlan.ratePlanCharges.size == 1),
```
kelvin-chappell: This is more strict than previous condition.  Does that mean previous condition `RatePlanHasACharge` is redundant?

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/HolidayCredit.scala
```diff
@@ -2,18 +2,28 @@ package com.gu.holidaystopprocessor
 
 import scala.math.BigDecimal.RoundingMode
 
+/**
+ * Replicating manual calculation found in
+ * https://docs.google.com/document/d/1N671Ei_nbln4ObZOWKvQZHnNkTgIx8EvdVJg1dHo_ak
+ *
+ * FIXME: Discounts should be taken into account?
```
kelvin-chappell: Not if initially replicating manual calculation.

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/Subscription.scala
```diff
@@ -58,16 +60,8 @@ case class RatePlanCharge(
   effectiveStartDate: LocalDate,
   chargedThroughDate: Option[LocalDate],
   HolidayStart__c: Option[LocalDate],
-  HolidayEnd__c: Option[LocalDate]
-) {
+  HolidayEnd__c: Option[LocalDate],
+  processedThroughDate: Option[LocalDate],
```
kelvin-chappell:
```suggestion
  processedThroughDate: Option[LocalDate]
```

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/CurrentGuardianWeeklySubscription.scala
```diff
@@ -0,0 +1,110 @@
...
+        ).forall(_._2)
+      }
+      .map { currentGuardianWeeklyRatePlan => // these ugly gets are safe due to above conditions
+        new CurrentGuardianWeeklySubscription(
+          subscriptionNumber = subscription.subscriptionNumber,
+          billingPeriod = currentGuardianWeeklyRatePlan.ratePlanCharges.head.billingPeriod.get,
+          price = currentGuardianWeeklyRatePlan.ratePlanCharges.head.price,
+          invoicedPeriod = CurrentInvoicedPeriod(
+            startDateIncluding = currentGuardianWeeklyRatePlan.ratePlanCharges.head.processedThroughDate.get,
+            endDateExcluding = currentGuardianWeeklyRatePlan.ratePlanCharges.head.chargedThroughDate.get
```
kelvin-chappell: Some duplication of `currentGuardianWeeklyRatePlan.ratePlanCharges.head` here

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/CurrentGuardianWeeklySubscription.scala
```diff
@@ -0,0 +1,110 @@
...
+          ),
+          ratePlanId = currentGuardianWeeklyRatePlan.id,
+          productRatePlanId = currentGuardianWeeklyRatePlan.productRatePlanId
+        )
+      }.getOrElse(throw new RuntimeException(s"Subscription does not have a current Guardian Weekly rate plan: ${subscription}"))
```
arausuy: Can we get stats out to cloudwatch here?

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/CurrentGuardianWeeklySubscription.scala
```diff
@@ -0,0 +1,110 @@
...
+          // FIXME: Enable this after test refactoring - the problem is LocalDate.now() call
+          //          TodayHasBeenInvoiced ->
+          //            Try {
+          //              PeriodContainsDate(
+          //                startPeriodInclusive = ratePlan.ratePlanCharges.head.processedThroughDate.get,
+          //                endPeriodExcluding = ratePlan.ratePlanCharges.head.chargedThroughDate.get,
+          //                date = LocalDate.now()
+          //              )
+          //            }.getOrElse(false),
+          RatePlanHasBeenInvoiced -> Try {
```
arausuy: I think refactoring this into something a bit more obvious would help

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/HolidayCredit.scala
mario-galic: We should throw. There should not be any monthly GWs whether on old GW plans or new GW plans. If there exists such a sub, then it was manually entered, extremely rare, and represents an error which needs to be fixed in Zuora by migrating it to appropriate plan.

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/CurrentGuardianWeeklySubscription.scala
mario-galic: It is not used because of the FIXME regarding `TodayHasBeenInvoiced`. Tests require non-trivial modification to enable it. Alternatively we could enable the condition only in PROD, but then we need to pass in config object to get at the stage.

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/CurrentGuardianWeeklySubscription.scala
mario-galic: I've seen cases of `processedThroughDate == chargedThroughDate` which means we are looking at wrong rate plan, so this is just double checking. `chargedThroughDate` is excluded to today must not be `chargedThroughDate`.

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/CurrentGuardianWeeklySubscription.scala
mario-galic: Yes there is alarm card in backlog: https://trello.com/c/DkXszGaJ/19-holiday-stop-processor-has-cloudwatch-alarm

PR #377, opened 2019-08-06
add success marker file to salesforce export state machine

pvighi:
Currently the salesforce data lake ETL jobs are triggered by time. We would like to trigger the jobs on the presence of the raw files instead. 

The current job generates returns raw data for a table in multiple files. The amount and name of the files is not known beforehand.

This PR adds a dummy file we can upload to S3 once we are sure all the raw CSV files have been uploaded.
The ETL job can be configured to trigger on the presence this dummy file in the raw bucket.

Tested in `CODE`, of course this doesn't test uploading to the actual raw directory in the Ophan account. I copied the S3 write permissions from the downloader lambda so hopefully it works in PROD as well

### DIFF
--- a/handlers/sf-datalake-export/src/main/scala/com/gu/sf_datalake_export/handlers/EndJobHandler.scala
+++ b/handlers/sf-datalake-export/src/main/scala/com/gu/sf_datalake_export/handlers/EndJobHandler.scala
@@ -3,7 +3,6 @@ package com.gu.sf_datalake_export.handlers
 import java.io.{InputStream, OutputStream}
 
 import com.amazonaws.services.lambda.runtime.Context
-import com.amazonaws.services.s3.model.{PutObjectRequest, PutObjectResult}
 import com.gu.effects.{GetFromS3, RawEffects, S3Path}
 import com.gu.salesforce.SalesforceAuthenticate.{SFAuthConfig, SFExportAuthConfig}
 import com.gu.salesforce.SalesforceClient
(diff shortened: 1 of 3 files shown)


handlers/sf-datalake-export/src/main/scala/com/gu/sf_datalake_export/handlers/EndJobHandler.scala
```diff
@@ -35,22 +45,35 @@ object EndJobHandler {
...
 
     val loadConfig = LoadConfigModule(stage, fetchString)
     val jobId = JobId(request.jobId)
+    val shouldUploadToDataLake = ShouldUploadToDataLake(request.uploadToDataLake)
+    val jobName = JobName(request.jobName)
+    val objectName = ObjectName(request.objectName)
+
+    val uploadPath: S3Path = ExportS3Path(stage)(objectName, shouldUploadToDataLake)
+
+    val successFile = {
```
mario-galic: Can you factor this out into utility function with uglier name and comment explaining the oddity, for example,

```
/**
 * Currently the salesforce data lake ETL jobs are triggered by time...
 */
private def uploadDummySuccessFileToTriggerOphanJob(request)...
```

handlers/sf-datalake-export/src/main/scala/com/gu/sf_datalake_export/handlers/EndJobHandler.scala
```diff
@@ -35,22 +44,44 @@ object EndJobHandler {
...
     fetchString: StringFromS3,
+    uploadDummySuccessFileToTriggerOphanJob: WireRequest => Try[_],
     getResponse: Request => Response
   )(request: WireRequest): Try[WireResponse] = {
 
     val loadConfig = LoadConfigModule(stage, fetchString)
     val jobId = JobId(request.jobId)
 
     for {
+      _ <- uploadDummySuccessFileToTriggerOphanJob(request)
```
pvighi: @mario-galic like this ?

PR #378, opened 2019-08-07
Fix scalariform warning

kelvin-chappell:
This stops Travis giving formatting warnings like this:
```
[warn] Scalariform parser error for /home/travis/build/guardian/support-service-lambdas/lib/holiday-stops/src/test/scala/com/gu/holiday_stops/ActionCalculatorTest.scala: expected start of definition, but was Token(VAL,val,876,val)
```
See:
* https://docs.scala-lang.org/sips/completed/trailing-commas.html
and
* https://github.com/sbt/sbt-scalariform/releases/tag/1.8.3

### DIFF
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -1,7 +1,7 @@
 addSbtPlugin("com.gu" % "sbt-riffraff-artifact" % "1.1.4")
 
 resolvers += Resolver.typesafeRepo("releases")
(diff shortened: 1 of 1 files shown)


kelvin-chappell:
This actually makes the problem worse, by converting the warning into an error.  Will wait for a fix for sbt/sbt-scalariform/issues/70

PR #382, opened 2019-08-08
Simplify next available date calculation

mario-galic:
The algorithm simplifies to

```
/**
 * If there are less than 5 days between today and the day after next publication day,
 * then Saturday after next (i.e., next-next Saturday),
 * otherwise next Saturday
 */
```

### DIFF
--- a/lib/holiday-stops/src/test/scala/com/gu/holiday_stops/ActionCalculatorTest.scala
+++ b/lib/holiday-stops/src/test/scala/com/gu/holiday_stops/ActionCalculatorTest.scala
@@ -30,15 +30,15 @@ class ActionCalculatorTest extends FlatSpec with Matchers {
     type Today = LocalDate
     type FirstAvailableDate = LocalDate
     val gwTodayToFirstAvailableDate = ListMap[Today, FirstAvailableDate](
-      LocalDate.of(2019, 6,  1) -> LocalDate.of(2019, 6,  8), // first available on Sun
-      LocalDate.of(2019, 6,  2) -> LocalDate.of(2019, 6,  8), // first available on Sun
-      LocalDate.of(2019, 6,  3) -> LocalDate.of(2019, 6,  8), // first available on Sun
-      LocalDate.of(2019, 6,  4) -> LocalDate.of(2019, 6, 15), // jump on Tue, a day before processor run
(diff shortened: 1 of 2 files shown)


lib/holiday-stops/src/test/scala/com/gu/holiday_stops/ActionCalculatorTest.scala
```diff
@@ -54,41 +54,26 @@ class ActionCalculatorTest extends FlatSpec with Matchers {
...
-      LocalDate.of(2019, 7,  1) -> LocalDate.of(2019, 7,  6), // first available on Sun
-      LocalDate.of(2019, 7,  2) -> LocalDate.of(2019, 7, 13) // jump on Tue, a day before processor run
+      LocalDate.of(2019, 6, 25) -> LocalDate.of(2019, 7, 6), // jump on Tue, a day before processor run
+      LocalDate.of(2019, 6, 26) -> LocalDate.of(2019, 7, 6), // first available on Sun
+      LocalDate.of(2019, 6, 27) -> LocalDate.of(2019, 7, 6), // first available on Sun
+      LocalDate.of(2019, 6, 28) -> LocalDate.of(2019, 7, 6), // first available on Sun
+      LocalDate.of(2019, 6, 29) -> LocalDate.of(2019, 7, 6), // first available on Sun
+      LocalDate.of(2019, 6, 30) -> LocalDate.of(2019, 7, 6), // first available on Sun
+      LocalDate.of(2019, 7, 1) -> LocalDate.of(2019, 7, 6), // first available on Sun
+      LocalDate.of(2019, 7, 2) -> LocalDate.of(2019, 7, 13) // jump on Tue, a day before processor run
```
kelvin-chappell: I think given the rule these tests could be done more clearly as a scalacheck-style property test.
Eg. we know with GW that if today is a Sunday the next available date will be this Saturday, and if today is a Thursday the next available date will be the following Saturday, etc.

mario-galic: I would prefer to keep it explicit so that one can pull up a calendar app and visually cross-reference the test cases with the calendar.

twrichards: as per discussion, explicit approach we have now is quicker/easier to understand

PR #383, opened 2019-08-09
Schedule processing of holiday stops

kelvin-chappell:


kelvin-chappell:
@twrichards As it stands, yes.  I can alter the alarm condition so it only fires once in a 24 hour period.

mario-galic:
@kelvin-chappell No, let it alarm every time it runs.

kelvin-chappell:
Actually, according to https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/AlarmThatSendsEmail.html:
```
After an alarm invokes an action due to a change in state, its subsequent behavior depends on the type of action that you have associated with the alarm. ... For Amazon SNS notifications, no additional actions are invoked. 
```
so it will only send an email out once until service is restored.

PR #386, opened 2019-08-13
Restrict number of holiday stops fetched from Salesforce

kelvin-chappell:
Doing an unrestricted fetch brought back an incomplete dataset, which meant that some Zuora data wasn't matching Salesforce records.  This meant that we had several hundred backfilled holiday-stop requests in Salesforce without any actual stopped publication dates.

By restricting to a certain date range we avoid having an incomplete dataset.  I don't quite understand what causes the incompleteness at the moment.  There appears to be a limit presumably caused by the join in some way.  But this limit varies between ~300 and ~600 records depending on the query.  Using a limit clause in the query doesn't help.


handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopbackfill/Backfiller.scala
```diff
@@ -16,10 +16,12 @@ object Backfiller {
     for {
       config <- Config(stage)
       stopsInZuora <- Right(holidayStopsAlreadyInZuora(src))
-      requestsInSf1 <- holidayStopRequestsAlreadyInSalesforce(config)
+      minStartDate = stopsInZuora.map(_.startDate).minBy(_.toEpochDay)
+      maxEndDate = stopsInZuora.map(_.endDate).maxBy(_.toEpochDay)
+      requestsInSf1 <- holidayStopRequestsAlreadyInSalesforce(config)(minStartDate, maxEndDate)
```
twrichards: what do these `1` and `2` mean, are they before/after? 
could they have a more meaningful name?

lib/holiday-stops/src/main/scala/com/gu/salesforce/holiday_stops/SalesforceHolidayStopRequest.scala
```diff
@@ -87,17 +87,21 @@ object SalesforceHolidayStopRequest extends Logging {
 
   }
 
-  object LookupByProductNamePrefix {
+  object LookupByDateRangeAndProductNamePrefix {
 
-    def apply(sfGet: HttpOp[RestRequestMaker.GetRequestWithParams, JsValue]): ProductName => ClientFailableOp[List[HolidayStopRequest]] =
+    def apply(sfGet: HttpOp[RestRequestMaker.GetRequestWithParams, JsValue]): (LocalDate, LocalDate, ProductName) => ClientFailableOp[List[HolidayStopRequest]] =
```
twrichards: could we use the value class wrappers here `HolidayStopRequestStartDate` and `HolidayStopRequestEndDate` for some extra type safety (to avoid any chance of mixing up the params).

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopbackfill/Backfiller.scala
kelvin-chappell: Yes I'll make it a bit clearer

lib/holiday-stops/src/main/scala/com/gu/salesforce/holiday_stops/SalesforceHolidayStopRequest.scala
kelvin-chappell: Yes I'll do it

PR #389, opened 2019-08-23
CloudWatch alarms previously pointing to subscriptions_dev now point to fulfilment-dev

twrichards:
**CloudWatch alarms** previously pointing to `subscriptions_dev` now point to `fulfilment-dev` 
(or `reader-revenue-dev` in the case of `catalog-service` since that's shared)

### DIFF
--- a/handlers/catalog-service/cfn.yaml
+++ b/handlers/catalog-service/cfn.yaml
@@ -210,7 +210,7 @@ Resources:
       Condition: CreateProdOnlyResources
       Properties:
         AlarmActions:
-        - !Sub arn:aws:sns:${AWS::Region}:${AWS::AccountId}:fulfilment-dev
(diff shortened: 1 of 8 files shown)


handlers/catalog-service/cfn.yaml
```diff
@@ -210,7 +210,7 @@ Resources:
       Condition: CreateProdOnlyResources
       Properties:
         AlarmActions:
-        - !Sub arn:aws:sns:${AWS::Region}:${AWS::AccountId}:subscriptions_dev
+        - !Sub arn:aws:sns:${AWS::Region}:${AWS::AccountId}:fulfilment-dev
```
johnduffell: I feel like catalog service is kind of everyone's concern as it would cause an issue for selling subscriptions

twrichards: changed to `reader-revenue-dev` like the metric-push lambda

PR #392, opened 2019-08-28
Move holiday-stop credit calculation

kelvin-chappell:
This moves the credit calculation to the `holiday-stops` lib where it's accessible to both the `holiday-stop-processor` lambda for actual credits, and the `holiday-stop-api` lambda for estimated credits.

Next step will be to integrate it into the `holiday-stop-api`.

### DIFF
--- a/lib/holiday-stops/src/main/scala/com/gu/holiday_stops/Subscription.scala
+++ b/lib/holiday-stops/src/main/scala/com/gu/holiday_stops/Subscription.scala
@@ -1,4 +1,4 @@
-package com.gu.holidaystopprocessor
+package com.gu.holiday_stops
 
 import java.time.LocalDate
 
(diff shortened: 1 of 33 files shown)


lib/holiday-stops/src/main/scala/com/gu/holiday_stops/Subscription.scala
```diff
@@ -12,32 +12,22 @@ case class Subscription(
...
+        start.isEqual(stop.stoppedPublicationDate) || start.isBefore(stop.stoppedPublicationDate)
+      } &&
+      charge.HolidayEnd__c.exists { end =>
+        end.isEqual(stop.stoppedPublicationDate) || end.isAfter(stop.stoppedPublicationDate)
+      }
 
     val charges = for {
       plan <- ratePlans if isMatchingPlan(plan)
-      charge <- plan.ratePlanCharges.find(isMatchingCharge)
+        charge <- plan.ratePlanCharges.find(isMatchingCharge)
```
arausuy: Formatting change?

kelvin-chappell: Scalariform is supposed to sort that out.  I think it's skipping this file because it doesn't like trailing commas in lists.  I'll fix it.

PR #395, opened 2019-08-30
Only run holiday-stop processor on Wednesdays

kelvin-chappell:
The processor will have no effect on other days, until we use it for other products.


twrichards:
is this not just creating one more thing to change in future when we support other products (which is going to be soon hopefully)?

kelvin-chappell:
It's a good point. Possibly not worth the small temporary efficiency gain.

PR #396, opened 2019-08-30
Add path api gateway request

frj:
This PR add a "path" field to the ApiGatewayRequest case class in order to support routing requests based on the full path.

Previously code could only match on path parameters making it difficult for a single lambdas to distinguish between request for different resources.

Making use of the path parameters strongly couples the code to the api gateway config/request routing, in the future it might be simpler just to use patten matching on the path?

Also included:
- Refactoring of the holiday stop api lambda routing which wasn't matching on the 'resource' name but using coincidental details of the query string to route requests
- Added some addition testing for the holiday stops api lambda (this will be improved on in future PRs)


handlers/holiday-stop-api/src/main/scala/com/gu/holiday_stops/Handler.scala
```diff
@@ -57,16 +60,14 @@ object Handler extends Logging {
...
-      request => (request.httpMethod match { // TODO will need to match against path params too to support edit endpoint
-        case Some("GET") => request.queryStringParameters match {
-          case Some(_) => stepsForPotentialHolidayStop _
-          case None => stepsToListExisting _
-        }
-        case Some("POST") => stepsToCreate _
-        case Some("DELETE") => stepsToDelete _
+      request => ((request.httpMethod,
+                   request.path) match { // TODO will need to match against path params too to support edit endpoint
+        case (Some("GET"), Some(POTENTIAL_PROXY_RESOURCE_PATH)) => stepsForPotentialHolidayStop _
```
twrichards: @johnduffell and I did have a play around this morning to see if we could make something a bit more like a router - didn't quite get to anything just right, and might be worth seeing if there is a lightweight dependency for this purpose rather than re-inventing the wheel. We could really cut down the CloudFormation by having a catch all there and doing all the path parameter definition here - would like to explore more, given more time in future

handlers/holiday-stop-api/src/main/scala/com/gu/holiday_stops/Handler.scala
```diff
@@ -116,8 +117,8 @@ object Handler extends Logging {
 
   def stepsToCreate(req: ApiGatewayRequest, sfClient: SfClient): ApiResponse = {
 
-    val verifyContactOwnsSubOp = SalesforceSFSubscription.SubscriptionForSubscriptionNameAndContact(sfClient.wrapWith(JsonHttp.getWithParams))
-    val createOp = SalesforceHolidayStopRequest.CreateHolidayStopRequestWithDetail(sfClient.wrapWith(JsonHttp.post))
+    val verifyContactOwnsSubOp: (SubscriptionName, Contact) => Types.ClientFailableOp[Option[MatchingSubscription]] = SalesforceSFSubscription.SubscriptionForSubscriptionNameAndContact(sfClient.wrapWith(JsonHttp.getWithParams))
```
twrichards: are these explicit types necessary?

johnduffell:
> Making use of the path parameters strongly couples the code to the api gateway config/request routing

what does that mean?  I know you can specify parameters to validate in the cfn, but I think they get through regardless.  Or did you mean something else?

handlers/holiday-stop-api/src/main/scala/com/gu/holiday_stops/Handler.scala
```diff
@@ -57,16 +60,14 @@ object Handler extends Logging {
...
-          case None => stepsToListExisting _
-        }
-        case Some("POST") => stepsToCreate _
-        case Some("DELETE") => stepsToDelete _
+      request => ((request.httpMethod,
+                   request.path) match { // TODO will need to match against path params too to support edit endpoint
+        case (Some("GET"), Some(POTENTIAL_PROXY_RESOURCE_PATH)) => stepsForPotentialHolidayStop _
+        case (Some("GET"), Some(GET_ALL_AND_CREATE_PROXY_RESOURCE_REGEX())) => stepsToListExisting _
+        case (Some("POST"), Some(GET_ALL_AND_CREATE_PROXY_RESOURCE_REGEX())) => stepsToCreate _
+        case (Some("DELETE"), Some(GET_ALL_AND_CREATE_PROXY_RESOURCE_REGEX())) => stepsToDelete _
```
johnduffell: is request.httpMethod really optional? it's not in real life, so if it's ended up in an option outside of our control, can we squash None into an error before it gets into this logic?  same goes for request.path

johnduffell: would it be worth pattern matching the resource first, and if it's `GET_ALL_AND_CREATE_PROXY_RESOURCE_REGEX` then match the method?

handlers/holiday-stop-api/src/main/scala/com/gu/holiday_stops/Handler.scala
johnduffell: scala is not great for converting between equivalent things transparently, but if the type is a bit wtf, the equivalent to consider would be to use `VerifyContact` and then define either a type alias, or for proper type checking at the expense of boilerplate, a trait similar to this 
```scala
trait VerifyContact {
  def apply(subName: SubscriptionName, contact: Contact): Types.ClientFailableOp[Option[MatchingSubscription]]
}
```

handlers/holiday-stop-api/src/main/scala/com/gu/holiday_stops/Handler.scala
twrichards: this would be a lot cleaner if possible

johnduffell:
looks good to me but I've chucked in a few random thoughts - use as you see fit!! 😆

handlers/holiday-stop-api/src/main/scala/com/gu/holiday_stops/Handler.scala
mario-galic: Custom router will fit nicely next to the custom http client.

frj:
@johnduffell RE:

> Making use of the path parameters strongly couples the code to the api gateway config/request routing

All I was saying was that we currently use the pathParams map in the ApiGatewayRequest object to get hold of path parameter. 

As I understand it the keys for that map are derived from the cloud formation config. 

I seems a bit of a long winded process to get hold of path parameters when they could just pop out of the path pattern matching.

handlers/holiday-stop-api/src/main/scala/com/gu/holiday_stops/Handler.scala
frj: I just expanded them to work out what was going on, ill remove them

johnduffell:
I don't think we directly derive anything like that from cfn, but it's possible there is some hard coded expected values from when the code wasn't fully refactored in the past, but that's probably because fixing that up is still in the tech debt doc.

I guess the main point I'm making is that we should design the URL structure to make sense, and then we should work out how to implement it.  Not based on how the code looks now as changing the code safely is much easier than changing the URL structure later.

handlers/holiday-stop-api/src/main/scala/com/gu/holiday_stops/Handler.scala
```diff
@@ -45,28 +45,65 @@ object Handler extends Logging {
...
-  def operationForEffects(
-    response: Request => Response,
-    stage: Stage,
-    fetchString: StringFromS3,
-  ): ApiGatewayOp[Operation] = {
+  val POTENTIAL_PROXY_RESOURCE_PATH = "/potential"
+  val GET_ALL_AND_CREATE_PROXY_RESOURCE_REGEX = """/hsr.*""".r
+
+  def operationForEffects(response: Request => Response, stage: Stage,
+                          fetchString: StringFromS3): ApiGatewayOp[Operation] = {
```
johnduffell: not sure if this was intentiional, but generally I would stick with consistent indent for both parameters and code (i.e. 2 space in most of this code)
if it wasn't intentional, I think there's a setting in intellij to turn off to stop it doing this randomly!

johnduffell: we managed to find the settings in intellij to stop that happening randomly - first found by Adam F:

image

PR #397, opened 2019-08-30
README for holiday-stop-api lambda

twrichards:
mainly collating the PR write-ups from https://github.com/guardian/support-service-lambdas/pull/302 and https://github.com/guardian/support-service-lambdas/pull/388, but with some extra about how the multiple/cross environments work


twrichards:
> Looks good. Maybe include the expected query parameters in the table as well?

They're already there.

kelvin-chappell:
You're quite right.  They are already there.

johnduffell:
nice work, docs 📈 with upward trend! 👍

PR #400, opened 2019-09-02
Apply GDPR DoNotProcess for Zuora Account on the ophan lake side

mario-galic:
https://trello.com/c/5QPjfBvi/331-understand-the-reasons-for-the-differences-between-rawzu-cleanzuora

Corresponding lake PR: https://github.com/guardian/ophan-data-lake/pull/4423

Fix for why

> thousands are in clean and not raw:*
> ```
> SELECT approx_distinct(name) FROM clean.zuora_subscription WHERE name NOT IN (
>     SELECT DISTINCT subscription_name FROM raw.zu_subs
> ) 
> 

Having mutable fields in filter does not work with Zuora stateful export API so we have to add corresponding fields to export and filter them on the ophan-data-lake side.

### DIFF
--- a/handlers/zuora-datalake-export/src/main/scala/com/gu/zuora/datalake/export/ExportLambda.scala
+++ b/handlers/zuora-datalake-export/src/main/scala/com/gu/zuora/datalake/export/ExportLambda.scala
@@ -135,7 +135,7 @@ object Query extends Enum[Query] {
 
   case object Account extends Query(
     "Account",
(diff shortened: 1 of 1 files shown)


mario-galic:
[Understand the reasons for the differences between raw.zu & clean.zuora](https://trello.com/c/5QPjfBvi/331-understand-the-reasons-for-the-differences-between-rawzu-cleanzuora)

PR #404, opened 2019-09-04
Filter out cancelled holiday stops

kelvin-chappell:
This is to ensure the holiday-stop processor ignores cancelled subscriptions.

It depends on guardian/salesforce#40.

### DIFF
--- a/lib/holiday-stops/src/main/scala/com/gu/salesforce/holiday_stops/SalesforceHolidayStopRequestsDetail.scala
+++ b/lib/holiday-stops/src/main/scala/com/gu/salesforce/holiday_stops/SalesforceHolidayStopRequestsDetail.scala
@@ -87,7 +87,10 @@ object SalesforceHolidayStopRequestsDetail extends Logging {
           | FROM $holidayStopRequestsDetailSfObjectRef
           | WHERE Product_Name__c LIKE '${productNamePrefix.value}%'
           | AND Stopped_Publication_Date__c = ${date.toString}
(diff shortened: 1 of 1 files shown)


lib/holiday-stops/src/main/scala/com/gu/salesforce/holiday_stops/SalesforceHolidayStopRequestsDetail.scala
```diff
@@ -87,6 +87,7 @@ object SalesforceHolidayStopRequestsDetail extends Logging {
           | FROM $holidayStopRequestsDetailSfObjectRef
           | WHERE Product_Name__c LIKE '${productNamePrefix.value}%'
           | AND Stopped_Publication_Date__c = ${date.toString}
+          | AND Subscription_Status__c != 'Cancelled'
```
mario-galic: Are you sure it is `Cancelled` with two Ls because Zuora sometimes uses single L?

kelvin-chappell: Yes.  Running this query in Prod gives me 10 rows:
```
select name, status__c from SF_Subscription__c 
where status__c = 'Cancelled'
limit 10
```

kelvin-chappell:
Now using cancellation effective date but I'm not sure what should happen when the stop date coincides with the cancellation effective date.  So now if cancellation becomes effective before the stop date the subscription will be ignored.

PR #407, opened 2019-09-06
Holiday stop within Guardian Weekly N-for-N intro plan

mario-galic:
- N-for-N is modelled as an extra optional field `introNforNMode` on  `CurrentGuardianWeeklySubscription`
- Unfortunately, now we have forks in both next billing period calculation, as well as credit calculation.
- I think if N-for-N has been modelled as discount in Zuora, then this PR would not be necessary.
- I think N-for-N despite having `Guardian Weekly` in the name does not behave as regular Guardian Weekly in Zuora, hence future planning session need to consider it a separate product.

Note there is a FIXME that needs to be addressed by a separate PR:

```
/* FIXME: estimation for N-for-N will be wrong */
```


mario-galic:
[Holiday stop within GW N-for-N period](https://trello.com/c/Wa0maywo/381-holiday-stop-within-gw-n-for-n-period)

lib/holiday-stops/src/main/scala/com/gu/holiday_stops/HolidayCredit.scala
```diff
@@ -9,12 +11,17 @@ import scala.math.BigDecimal.RoundingMode
...
  */
 object HolidayCredit {
-  def apply(currentGuardianWeeklySubscription: CurrentGuardianWeeklySubscription): Double = {
-    val recurringPrice = currentGuardianWeeklySubscription.price
-    val numPublicationsInPeriod = BillingPeriodToApproxWeekCount(currentGuardianWeeklySubscription)
-    def roundUp(d: Double): Double = BigDecimal(d).setScale(2, RoundingMode.UP).toDouble
-    -roundUp(recurringPrice / numPublicationsInPeriod)
+  def apply(currentGuardianWeeklySubscription: CurrentGuardianWeeklySubscription, stoppedPublicationDate: LocalDate): Double = {
+    if (currentGuardianWeeklySubscription.introNforNMode.isDefined && stoppedPublicationDate.isBefore(currentGuardianWeeklySubscription.invoicedPeriod.startDateIncluding)) {
+      -roundUp(1) // hardcoded because N-for-N means 1 currency unit per issue
```
kelvin-chappell: Should perhaps give the magic number a name rather than a comment

lib/holiday-stops/src/main/scala/com/gu/holiday_stops/CurrentGuardianWeeklySubscription.scala
```diff
@@ -234,7 +247,16 @@ object CurrentGuardianWeeklySubscription {
           invoicedPeriod = predictedInvoicePeriod,
           ratePlanId = currentGuardianWeeklyWithoutInvoice.id,
           productRatePlanId = currentGuardianWeeklyWithoutInvoice.productRatePlanId,
-          introNforNMode = true
+          introNforNMode = Some(IntroNForNMode(
+            billingPeriod = gwNForNwRatePlan.ratePlanCharges.head.billingPeriod.get,
+            price = gwNForNwRatePlan.ratePlanCharges.head.price,
+            invoicedPeriod = CurrentInvoicedPeriod(
+              startDateIncluding = gwNForNwRatePlan.ratePlanCharges.head.processedThroughDate.get,
+              endDateExcluding = gwNForNwRatePlan.ratePlanCharges.head.chargedThroughDate.get
```
kelvin-chappell: We might be bitten by these heads and gets in future.  I know there's a safety check earlier but still seems like it could be dangerous if it gets refactored.

PR #410, opened 2019-09-16
Add mapping for holiday-stop confirmation emails to batch-email sender

kelvin-chappell:
See also https://github.com/guardian/membership-workflow/pull/202.


handlers/batch-email-sender/src/main/scala/com/gu/batchemailsender/api/batchemail/model/EmailToSend.scala
```diff
@@ -55,7 +55,7 @@ object EmailToSend {
       case ("Payment_Failure__c", "DD_PF2") => "SV_DDpaymentfailure2"
       case ("Payment_Failure__c", "DD_PF3") => "SV_DDpaymentfailure3"
       case ("Payment_Failure__c", "DD_PF4") => "SV_DDpaymentfailure4"
+      case ("Holiday_Stop_Request__c", _) => "SV_HolidayStopConfirmation"
```
mario-galic: Is there no `email_stage` parameter? Will there be only one type of holiday stop email?

kelvin-chappell: Yes I suppose this should be "confirm".  I'll change it.

PR #411, opened 2019-09-16
Include extra fields for Braze holiday-stop confirmation template

kelvin-chappell:
There are some very specific fields used in the holiday-stop Braze template that haven't been included for any other template so far.  It should be quite obvious what the new fields mean, I hope.

This follows on from #410

### DIFF
--- a/handlers/batch-email-sender/src/main/scala/com/gu/batchemailsender/api/batchemail/model/EmailBatch.scala
+++ b/handlers/batch-email-sender/src/main/scala/com/gu/batchemailsender/api/batchemail/model/EmailBatch.scala
@@ -22,13 +22,17 @@ object EmailBatch {
       identity_id: Option[String],
       first_name: String,
       email_stage: String,
-      holiday_start_date: Option[String],
-      holiday_end_date: Option[String],
-      stopped_credit_sum: Option[String],
-      currency_symbol: Option[String],
-      stopped_issue_count: Option[String]
+      holiday_stop_request: Option[WireHolidayStopRequest]
+    )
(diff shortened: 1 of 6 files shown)


handlers/batch-email-sender/src/main/scala/com/gu/batchemailsender/api/batchemail/model/EmailBatch.scala
```diff
@@ -21,7 +21,12 @@ object EmailBatch {
       last_name: String,
       identity_id: Option[String],
       first_name: String,
-      email_stage: String
+      email_stage: String,
```
mario-galic: Seems strange to have all these holiday-specific fields at the top level. Is the following not possible

```
case class HolidayStop(
  holiday_start_date: Option[String],
  ...
)
```

and then have it as an optional field

```
case class WireEmailBatchItemPayload(
      ...
      email_stage: String,
      holidayStop: Option[HolidayStop]
    )
```

kelvin-chappell: Good point.  I wasn't sure how much flexibility there was in building the payload.  But json is json so this must be possible and would certainly be an improvement.  It would also reflect the fact that these fields are all required for a holiday stop.  I'll do it!

kelvin-chappell:
@mario-galic From experimenting and discussing with Krystian it looks like nested properties aren't supported in Braze templates.  So the best I can do is nest the structure that Salesforce sends to the service layer.

PR #412, opened 2019-09-16
Fetch first availabile date for sundays

frj:
This modifies the get existing holiday stops endpoint to support returning 'product specifics' (firstAvailableDate, issueDayOfWeek, annualIssueLimit) based on the productType/ratePlanName combination rather than the product name prefix header.

This lookup currently only supports Newspaper Voucher/Sunday only and Guardian Weekly subscriptions.

The endpoint still supports looking up specifics using the x-product-name-prefix header for Guardian Weekly only.

Notes:
- Temporally removed MutabilityFlags from holiday stops returned in the result as these fields are not in use. Calculating their values requires the additional complication of fetching productType and ratePlanName for each holiday stop from Salesforce
-  Changed lots of Play json 'Reads' to 'Formats' to support serialisation/deserialisation in tests

### DIFF
--- a/lib/holiday-stops/src/main/scala/com/gu/holiday_stops/ActionCalculator.scala
+++ b/lib/holiday-stops/src/main/scala/com/gu/holiday_stops/ActionCalculator.scala
@@ -6,10 +6,20 @@ import java.time.{DayOfWeek, LocalDate}
 
 import com.gu.salesforce.holiday_stops.SalesforceHolidayStopRequestsDetail.{ProductName, ProductRatePlanKey, ProductRatePlanName, ProductType}
 
-case class ProductSpecifics(
+case class LegacyProductSpecifics(
   firstAvailableDate: LocalDate,
   issueDayOfWeek: Int,
   annualIssueLimit: Int
+)
+
+case class ProductSpecifics(
+  annualIssueLimit: Int,
+  issueSpecifics: List[IssueSpecifics]
+)
+
+case class IssueSpecifics(
+  firstAvailableDate: LocalDate,
+  issueDayOfWeek: Int,
 )
 
 /**
@@ -31,24 +41,33 @@ case class ProductSpecifics(
  */
 object ActionCalculator {
 
+
+  /**
+   * @param annualIssueLimit         Maximum number of days holidays that can be taken annually for a subscription
+   * @param issueConstants           Constants specific to the issues covered by the type of subscription
+   */
+  case class SuspensionConstants(
+    annualIssueLimit: Int,
+    issueConstants: List[IssueSuspensionConstants]
+  )
+
   /**
    * @param issueDayOfWeek           Weekday corresponding to publication issue date printed on the paper, for example, Friday for GW
    * @param processorRunLeadTimeDays Number of days (including one safety-net day) before publication issue date when the holiday processor runs.
    *                                 One safety-day before fulfilment day. Safety day gives us an opportunity to fix issues before fulfilment runs.
-   * @param annualIssueLimit
    */
-  sealed abstract class ProductSuspensionConstants(
+  sealed abstract class IssueSuspensionConstants(
     val issueDayOfWeek: DayOfWeek,
     val processorRunLeadTimeDays: Int,
-    val annualIssueLimit: Int
   ) {
     def firstAvailableDate(today: LocalDate): LocalDate
   }
 
-  case object GuardianWeeklySuspensionConstants extends ProductSuspensionConstants(
+  val GuardianWeeklySuspensionConstants = SuspensionConstants(6, List(GuardianWeeklyIssueSuspensionConstants))
+
+  case object GuardianWeeklyIssueSuspensionConstants extends IssueSuspensionConstants(
     issueDayOfWeek = DayOfWeek.FRIDAY,
     processorRunLeadTimeDays = 8 + (1 /* safety-day */ ), //one (safety) day before the Thursday of the week before the Friday issue day
-    annualIssueLimit = 6
   ) {
     val minDaysBetweenTodayAndFirstAvailableDate = 5
     val maxDaysBetweenTodayAndFirstAvailableDate = 11
@@ -82,10 +101,11 @@ object ActionCalculator {
     }
   }
 
-  case object SundayVoucherSuspensionConstants extends ProductSuspensionConstants(
+  val SundayVoucherSuspensionConstants = SuspensionConstants(6, List(SundayVoucherIssueSuspensionConstants))
+
+  case object SundayVoucherIssueSuspensionConstants extends IssueSuspensionConstants(
     issueDayOfWeek = DayOfWeek.SUNDAY,
     processorRunLeadTimeDays = 1,
-    annualIssueLimit = 6
   ) {
     def firstAvailableDate(today: LocalDate): LocalDate = {
       today.plus(processorRunLeadTimeDays.toLong, ChronoUnit.DAYS).`with`(TemporalAdjusters.next(DayOfWeek.SUNDAY))
@@ -93,7 +113,7 @@ object ActionCalculator {
   }
 
   // TODO this will likely need to change to return an array of days of week (when we support more than just GW)
-  def suspensionConstantsByProduct(productNamePrefix: ProductName): ProductSuspensionConstants =
+  def suspensionConstantsByProduct(productNamePrefix: ProductName): SuspensionConstants =
     productNamePrefix.value match {
       case s if s.startsWith("Guardian Weekly") => GuardianWeeklySuspensionConstants
       case _ => throw new RuntimeException(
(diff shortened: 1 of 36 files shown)


lib/holiday-stops/src/main/scala/com/gu/holiday_stops/ActionCalculator.scala
```diff
@@ -85,10 +82,34 @@ object ActionCalculator {
     }
   }
 
+  case object SundayVoucherSuspensionConstants extends ProductSuspensionConstants(
+    issueDayOfWeek = DayOfWeek.SUNDAY,
+    processorRunLeadTimeDays = 1,
+    annualIssueLimit = 6
+  ) {
+    def firstAvailableDate(today: LocalDate): LocalDate = {
```
mario-galic: Should first available date be simply next day as there is nothing to fulfil? Reader simply does not take the voucher to the agent, that is, there is no delivery?

lib/holiday-stops/src/main/scala/com/gu/holiday_stops/ActionCalculator.scala
```diff
@@ -85,10 +82,34 @@ object ActionCalculator {
...
+    productNamePrefix.value match {
+      case s if s.startsWith("Guardian Weekly") => GuardianWeeklySuspensionConstants
+      case _ => throw new RuntimeException(
+        s"Failed to determine ProductSuspensionConstants because of unexpected productName: $productNamePrefix "
+      )
+    }
+
+  def suspensionConstantsByProductRatePlanKey(
+    productKey: ProductRatePlanKey
+  ): Either[ActionCalculatorError, List[ProductSuspensionConstants]] = productKey match {
```
mario-galic: Why `Lists`? Why `Eithers`? If product key is unknown this is a system configuration issue so there is no recovery mechanism and we do not wish to proceed with processing anything.

lib/holiday-stops/src/test/scala/com/gu/holiday_stops/ActionCalculatorTest.scala
```diff
@@ -81,24 +94,45 @@ class ActionCalculatorTest extends FlatSpec with Matchers {
...
-
+      LocalDate.of(2019, 5, 24),
+      LocalDate.of(2019, 5, 31),
+      LocalDate.of(2019, 6, 7),
+      LocalDate.of(2019, 6, 14),
+      LocalDate.of(2019, 6, 21)
+    )
+  }
+  it should "calculate first available date for Sunday Voucher" in {
+    val gwTodayToFirstAvailableDate = ListMap[Today, FirstAvailableDate](
```
mario-galic: Naming typo, perhaps `sundayVoucherFirstAvailableDate`

lib/holiday-stops/src/test/scala/com/gu/holiday_stops/ActionCalculatorTest.scala
```diff
@@ -81,24 +94,45 @@ class ActionCalculatorTest extends FlatSpec with Matchers {
...
+      LocalDate.of(2019, 9, 9) -> LocalDate.of(2019, 9, 15), // first available on Sun
```
mario-galic: As mentioned above, are we sure of these dates?

lib/holiday-stops/src/test/scala/com/gu/salesforce/holiday_stops/SalesForceHolidayStopsEffects.scala
```diff
@@ -0,0 +1,31 @@
...
+import play.api.libs.json.Json
+
+object SalesForceHolidayStopsEffects {
+
+  def listHolidayStopsResponseBody =
+    """{
+      | "records": []
+      |}""".stripMargin
+
+  def listHolidayRequestQuery(contactId: String, subscriptionName: String) =
```
mario-galic: Is this on purpose with all the handmade encodings?

lib/holiday-stops/src/main/scala/com/gu/holiday_stops/ActionCalculator.scala
twrichards: yes you should remove the `'with'(TemporalAdjusters.next(DayOfWeek.SUNDAY))` part of this

lib/holiday-stops/src/main/scala/com/gu/salesforce/holiday_stops/SalesforceHolidayStopRequestsDetail.scala
```diff
@@ -28,6 +28,11 @@ object SalesforceHolidayStopRequestsDetail extends Logging {
   case class ProductName(value: String) extends AnyVal
   implicit val formatProductName = Jsonx.formatInline[ProductName]
 
+  case class ProductRatePlanName(value: String) extends AnyVal
+  case class ProductType(value: String) extends AnyVal
+
+  case class ProductRatePlanKey(productType: ProductType, ratePlanName: ProductRatePlanName)
```
mario-galic: Consider documenting this is how we "uniquely identify a product".

I was kind of hoping we would solve the problem of "uniquely identifying product" in a way that we could simply speak of a `Product` and `Product ID` as opposed to zuora model N-tuple (product, product rate plan, ..), but alas.

lib/holiday-stops/src/main/scala/com/gu/holiday_stops/ActionCalculator.scala
frj: The List is just preparing for the future where we will be supporting subscriptions with more than one issue.

The either thing is just a preference, if something is going to fail its a good thing to represent that in the type signature rather than just throwing an exception.

Youre right that we control the systems that call the api so we are probably not going to hit the error condition, but in theory we might want to return a response code relevant to the specific error, eg a 40x in this case.

lib/holiday-stops/src/test/scala/com/gu/holiday_stops/ActionCalculatorTest.scala
frj: 👍

lib/holiday-stops/src/main/scala/com/gu/holiday_stops/ActionCalculator.scala
frj: 👍

lib/holiday-stops/src/main/scala/com/gu/holiday_stops/ActionCalculator.scala
mario-galic: If `Left`s are acted on then `Either` makes sense, otherwise it is there just for the sake of it. For example, in the processor we encode errors as `Left`s so we do not interrupt processing of other holiday requests in the batch. But I am fan of `Either` so keep it if you want. However now someone has to look at logs instead of team being automatically alerted on lambda failure wtr unknown product. This scenario looks to me more 500 than bad request if you are planning to feed response of it.

mario-galic: I still do not understand the need for lists. I thought we would be handling each product individually. So `Sunday` would be handled the same wether part of `Weekend` or `Everyday`. Anyways we do not need them right now so it is a case of premature abstraction.

lib/holiday-stops/src/test/scala/com/gu/holiday_stops/ActionCalculatorTest.scala
frj: 👍

lib/holiday-stops/src/test/scala/com/gu/salesforce/holiday_stops/SalesForceHolidayStopsEffects.scala
frj: I realise it doesnt look that good but the http mocking "framework" needs the url to be an exact match. 

I had tried to build the string from an unescaped value but I was getting caught out by intellij removing spaces from the multi line string.

I think the answer to this type of problem is to either improve the http mocking framework so that you can be a bit more intelligent about matching request, or move things over to use sttp which has a mocking framework that does this better.

Ill maybe change this back to an unescaped string but make it non multi line as you can be explicit about the spaces and line endings

kelvin-chappell: 👍 for moving everything over to sttp

lib/holiday-stops/src/main/scala/com/gu/salesforce/holiday_stops/SalesforceHolidayStopRequestsDetail.scala
frj: 👍

lib/holiday-stops/src/main/scala/com/gu/holiday_stops/ActionCalculator.scala
frj: Ive pushed some changes that came out of a conversation with Tom, they could make things more or less confusing. 

The idea is that when the user wants to book a holiday for a subscription, an api call is made identifying the type of subscription using the Product Type/Rate Plan combination.

We then return all of the  information relevant for that subscription, something along the lines of:
{
 ...
  issueSpecifics: [
    {
      firstAvailableDate: "...", 
      issueDayOfWeek: saturday
     },
    {
      firstAvailableDate: "...", 
      issueDayOfWeek: sunday
     }
  ],
  annualIssueLimit: 10
}

So this function provides the mapping between Product Type/Rate Plan and information from each of the ProductRatePlanCharges contained by that rate plan.

You are right it is strictly premature abstraction but it came out of the analysis we had done around the identification of the product, it seemed like the right way to go.

PR #413, opened 2019-09-16
save estimated credits during create AND factor out V1 of /potential endpoint

twrichards:
### MUST be released after https://github.com/guardian/manage-frontend/pull/274

- Eliminated V1 of /potential endpoint (and cleaned up naming and README accordingly) - following https://github.com/guardian/manage-frontend/pull/274
- Now calculates credit on actual creation call (rather than just on /potential) and then posts to SalesForce in the 'composite tree' call. Did some refactors along the way such as...
  - introduction of type alias `PartiallyWiredCreditCalculator` and created this up-front in the `operationForEffects` function within `Handler` rather than passing the `config` and `backend` instances around everywhere

### DIFF
--- a/lib/holiday-stops/src/main/scala/com/gu/holiday_stops/CreditCalculator.scala
+++ b/lib/holiday-stops/src/main/scala/com/gu/holiday_stops/CreditCalculator.scala
@@ -6,10 +6,14 @@ import com.gu.salesforce.holiday_stops.SalesforceHolidayStopRequestsDetail.Subsc
 import com.softwaremill.sttp.{Id, SttpBackend}
 
 object CreditCalculator {
+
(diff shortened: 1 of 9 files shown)


lib/holiday-stops/src/main/scala/com/gu/holiday_stops/CreditCalculator.scala
```diff
@@ -6,10 +6,14 @@ import com.gu.salesforce.holiday_stops.SalesforceHolidayStopRequestsDetail.Subsc
 import com.softwaremill.sttp.{Id, SttpBackend}
 
 object CreditCalculator {
+
+  type PartiallyWiredCreditCalculator = (SubscriptionName, LocalDate) => Either[HolidayError, Double]
```
mario-galic: Everything in this repo is "partially wired".

PR #414, opened 2019-09-17
holiday-stops: log credit calculation errors

twrichards:
In https://github.com/guardian/support-service-lambdas/pull/413 I left behind a TODO.

Anytime the system cannot determine the 'credit' we should log accordingly - this PR does exactly that.

Uses `<|` from [mouse](https://github.com/typelevel/mouse) - which can be replaced for native implementation of 'tap' once we upgrade to Scala 2.13 (as per [comment](https://github.com/guardian/support-service-lambdas/pull/414#pullrequestreview-289481017) from @mario-galic).

### DIFF
--- a/lib/holiday-stops/src/main/scala/com/gu/holiday_stops/CreditCalculator.scala
+++ b/lib/holiday-stops/src/main/scala/com/gu/holiday_stops/CreditCalculator.scala
@@ -3,11 +3,21 @@ package com.gu.holiday_stops
 import java.time.LocalDate
 
 import com.gu.salesforce.holiday_stops.SalesforceHolidayStopRequestsDetail.SubscriptionName
+import com.gu.util.Logging
 import com.softwaremill.sttp.{Id, SttpBackend}
 
 object CreditCalculator {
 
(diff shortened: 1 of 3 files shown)


lib/holiday-stops/src/main/scala/com/gu/holiday_stops/CreditCalculator.scala
```diff
@@ -3,12 +3,22 @@ package com.gu.holiday_stops
...
 
 import com.gu.salesforce.holiday_stops.SalesforceHolidayStopRequestsDetail.SubscriptionName
+import com.gu.util.Logging
 import com.softwaremill.sttp.{Id, SttpBackend}
 
 object CreditCalculator {
 
   type PartiallyWiredCreditCalculator = (SubscriptionName, LocalDate) => Either[HolidayError, Double]
 
+  implicit class HolidayErrorImplicits[T](theEither: Either[HolidayError, T]) extends Logging {
```
mario-galic: Just extend `LazyLogging` on top-level object.

twrichards:
someone will have to merge this while I'm away

PR #415, opened 2019-09-18
Fetch impacted publications for sundays

frj:
This modifies /potential holiday stops endpoint to support returning impacted dates for a holiday stop date range based on the productType/ratePlanName combination rather than the product name prefix header.

Using the estimate credit flag is not supported (using it will fail silently returning a response with no estimate)

This way of using the endpoint supports Newspaper Voucher/Sunday and Guardian Weekly subscriptions.

The endpoint still supports using the x-product-name-prefix header for Guardian Weekly only.


handlers/holiday-stop-api/src/main/scala/com/gu/holiday_stops/Handler.scala
```diff
@@ -141,26 +142,57 @@ object Handler extends Logging {
   }).toApiGatewayOp(s"either '$HEADER_IDENTITY_ID' header OR '$HEADER_SALESFORCE_CONTACT_ID' (one is required)")
 
   case class PotentialHolidayStopsPathParams(subscriptionName: SubscriptionName)
-  case class PotentialHolidayStopsQueryParams(startDate: LocalDate, endDate: LocalDate, estimateCredit: Option[String])
+
+  case class PotentialHolidayStopsQueryParams(
+    startDate: LocalDate,
+    endDate: LocalDate,
+    estimateCredit: Option[String],
```
kelvin-chappell: Shouldn't this be boolean?

lib/holiday-stops/src/test/scala/com/gu/holiday_stops/ActionCalculatorTest.scala
```diff
@@ -124,4 +124,41 @@ class ActionCalculatorTest extends FlatSpec with Matchers with EitherValues {
...
+    ) shouldEqual Right(List(
+      LocalDate.of(2019, 5, 26),
+      LocalDate.of(2019, 6, 2),
+      LocalDate.of(2019, 6, 9),
+      LocalDate.of(2019, 6, 16),
+      LocalDate.of(2019, 6, 23)
+    ))
+  }
+  it should "return an error for an unsupported product rate plan" in {
+    val unsuportedProductRatePlanKey =
```
kelvin-chappell:
```suggestion
    val unsupportedProductRatePlanKey =
```

handlers/holiday-stop-api/src/main/scala/com/gu/holiday_stops/Handler.scala
frj: Its a bit fiddly but this case class gets parsed from the queryStringParameters Json that gets passed through from the api gateway. 

In the json all the query values are strings so you would need to do a custom mapping for the case class to make this field a boolean.

lib/holiday-stops/src/test/scala/com/gu/holiday_stops/ActionCalculatorTest.scala
frj: 👍

handlers/holiday-stop-api/src/main/scala/com/gu/holiday_stops/Handler.scala
kelvin-chappell: 👍

frj:
[Fetch Sunday Voucher impacted publications from holiday-stop-api](https://trello.com/c/JjzKISfq/409-fetch-sunday-voucher-impacted-publications-from-holiday-stop-api)

PR #416, opened 2019-09-18
sf-gocardless-sync: alarm only if there are sustained problems with the sync

twrichards:
changed alarm condition to be only if there are sustained problems with the sync to avoid unnecessary investigation into single failures (since this lambda just picks up where it left off - every 5 mins currently).

Also added a description with 'impact' and link to the alarms doc (with further explanation).


twrichards:
someone will have to merge this while I'm away

PR #421, opened 2019-09-20
Calculate sunday voucher credit amount

frj:
This PR implements the credit calculation for sunday vouchers


frj:
[Calculate Sunday Voucher holiday credit amount](https://trello.com/c/M41N8BqD/408-calculate-sunday-voucher-holiday-credit-amount)

lib/holiday-stops/src/main/scala/com/gu/holiday_stops/CreditCalculator.scala
```diff
@@ -21,14 +23,51 @@ object CreditCalculator extends LazyLogging {
...
+        stoppedPublicationDate
+      ) _,
+      sundayVoucherCredit(
+        sundayVoucherRatePlanId,
+        stoppedPublicationDate
+      ) _
+    )
+
+    //Returns the result of the first function that returns a right
+    creditCalculatorFunctions
```
mario-galic: This is excellent idea! I think it will be key to performance as we handle more products. Could we go with lazy collection though, instead of foldRight, to make laziness explicit like so:

```
lazy val f1 = guardianWeeklyCredit(guardianWeeklyProductRatePlanIds, gwNforNProductRatePlanIds, stoppedPublicationDate)(subscription)
lazy val f2 = sundayVoucherCredit(sundayVoucherRatePlanId)(subscription)
(f1 #:: f2 #:: Stream.empty) collectFirst { case Right(x) => x } toRight(ZuoraHolidayWriteError(s"Failed to calculate credit for ${subscription.subscriptionNumber}"))
```

lib/holiday-stops/src/main/scala/com/gu/holiday_stops/CreditCalculator.scala
```diff
@@ -21,14 +23,51 @@ object CreditCalculator extends LazyLogging {
...
         stoppedPublicationDate
       )(subscription)
     } yield credit) <| (logger.error("Failed to calculate holiday stop credits", _))
 
+  def calculateCredit(
+    guardianWeeklyProductRatePlanIds: List[String],
+    gwNforNProductRatePlanIds: List[String],
+    sundayVoucherRatePlanId: String,
+    stoppedPublicationDate: LocalDate
+  )(subscription: Subscription) = {
```
mario-galic: Public interfaces should have return type annotation.

lib/holiday-stops/src/main/scala/com/gu/holiday_stops/BillingPeriodToApproxWeekCount.scala
```diff
@@ -0,0 +1,15 @@
+package com.gu.holiday_stops
+
+// FIXME: Is this assumption safe?
+object BillingPeriodToApproxWeekCount {
+  def apply(subscriptionNumber: String, billingPeriod: String): Int =
+    billingPeriod match {
+      case "Quarter" => 13
+      case "Annual" => 52
+      case "Month" => 4
+      case _ => throw new RuntimeException(
```
mario-galic: Actually, lets go with `Either` here since processor should continue processing other holiday stop requests in the batch. Maybe separate PR if it is out-of-scope.

lib/holiday-stops/src/test/scala/com/gu/holiday_stops/CreditCalculatorSpec.scala
```diff
@@ -0,0 +1,45 @@
...
+
+import java.time.LocalDate
+
+import io.circe.generic.auto._
+import io.circe.parser.decode
+import org.scalatest._
+
+import scala.io.Source
+
+class CreditCalculatorSpec extends FlatSpec with Matchers with EitherValues {
```
mario-galic: This is a kind of unit test I would prefer over protocol wiring tests if we had to make a choice between the two, because here we are testing key piece of business logic which would be silently failing as logic error as opposed to wiring testing which would very likely surface on the first UAT if not even at compile time.

lib/holiday-stops/src/main/scala/com/gu/holiday_stops/CreditCalculator.scala
mario-galic: https://stackoverflow.com/questions/58035836/lazily-traverse-collection-of-functions-stopping-on-first-success

mario-galic: Actually, `orElse` already provides these semantics simply like so

```
f1(in) orElse f2(in) orElse f3(in) orElse Left(ZuoraHolidayWriteError(s"Failed to calculate credit for ${subscription.subscriptionNumber}"))
```

frj: 👍

PR #424, opened 2019-09-25
stop holiday stops api making multiple unnecessary zuora calls to get subscription

frj:
This PR solves the issue that the creditCalculator function looks up the subscription from Zuora every time it is called.

A 'factory' function is introduced that does the subscription lookup and returns the 'calculator' function that just takes a local date which is used to estimate the credits for each holiday stop detail.


mario-galic:
Just take out ` Zuora.subscriptionGetResponse` request from calculator and put it in `stepsForPotentialHolidayStop` for-comprehension, and the pass resulting subscription to calculator.

frj:
[Stop holiday stops /potentials endpoint from making multiple calls to fetch the same subscription](https://trello.com/c/bmSIdzoB/494-stop-holiday-stops-potentials-endpoint-from-making-multiple-calls-to-fetch-the-same-subscription)

PR #432, opened 2019-09-27
Remove config signature pollution from Holiday stops methods

mario-galic:
Pass the whole config object around instead of bits and pieces of it. For example,

 ```
def writeHolidayStopToZuora(
    holidayCreditProduct: HolidayCreditProduct,
    guardianWeeklyProductRatePlanIds: List[String],	
    gwNforNProductRatePlanIds: List[String],
    ...
```

becomes

```
def writeHolidayStopToZuora(
    config: Config,
    ...
```

### DIFF
--- a/handlers/holiday-stop-api/src/main/scala/com/gu/holiday_stops/Handler.scala
+++ b/handlers/holiday-stop-api/src/main/scala/com/gu/holiday_stops/Handler.scala
@@ -63,8 +63,7 @@ object Handler extends Logging {
       validateRequestAndCreateSteps(
         request,
         CreditCalculator.calculateCredit(
-          config.guardianWeeklyConfig.productRatePlanIds,
-          config.guardianWeeklyConfig.nForNProductRatePlanIds,
+          config,
           config.sundayVoucherConfig.productRatePlanChargeId
         ),
         getSubscriptionFromZuora(config, backend)
--- a/handlers/holiday-stop-api/src/main/scala/com/gu/holiday_stops/Handler.scala
+++ b/handlers/holiday-stop-api/src/main/scala/com/gu/holiday_stops/Handler.scala
@@ -62,10 +62,7 @@ object Handler extends Logging {
     } yield Operation.noHealthcheck(request => // checking connectivity to SF is sufficient healthcheck so no special steps required
       validateRequestAndCreateSteps(
         request,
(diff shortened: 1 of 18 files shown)

### CODE REVIEW

File: handlers/holiday-stop-api/src/main/scala/com/gu/holiday_stops/Handler.scala [hunk 1]
```diff
@@ -62,11 +62,7 @@ object Handler extends Logging {
     } yield Operation.noHealthcheck(request => // checking connectivity to SF is sufficient healthcheck so no special steps required
       validateRequestAndCreateSteps(
         request,
-        CreditCalculator.calculateCredit(
-          config.guardianWeeklyConfig.productRatePlanIds,
-          config.guardianWeeklyConfig.nForNProductRatePlanIds,
-          config.sundayVoucherConfig.productRatePlanChargeId
-        ),
+        CreditCalculator.calculateCredit(config),
```
johnduffell: it may look simpler, but it's actually more complicated - calculateCredit now needs a Zuora and SF config.
which is more complex:
`def f: (Boolean, Int, Char) => String`
or
`def f: TouchpointBackend => String`

Yes, you have to trade off between passing in a ready made big object that's easy to read and making custom case classes, but otherwise you end up with all knowing objects that are needed everywhere, and very difficult to mock or separate into a separate code base.

johnduffell:
I would also argue that product ids etc are not secrets, and could just be put in a Map in the source code, meaning they would be checked at compile time.  That does mean we would have to recompile if we wanted to add a new environment, but that's a lot more rare than other code changes in my experience.

kelvin-chappell:
Perhaps have a concrete implementation of `Config` for each product-type and stage?

mario-galic:
- @johnduffell @frj Yes I agree with you regarding decoupling and have merged PRs before that refactored passing in big config object to passing only what is necessary, for example, see https://github.com/guardian/support-service-lambdas/pull/367/files#diff-0f02e76143f7d0a3b46272902ff9325bR19 however I am reversing this decision because we will try abstracting away processing of holiday stops for all products in the same way, and differing signatures due to config is interfering with that. 

- Product IDs are not secrets and we already hardcode them by env.

- Unit tests should in no way depend on Touchpoint systems. We should unit test business logic, not http requests or config plumbing, so it should be safe to null touchpoint config.

PR #434, opened 2019-09-27
Add invoice date to response to potential holiday-stop request

kelvin-chappell:
This remodels the `PotentialHolidayStop` class slightly so that it includes the expected dates of the invoices on which the holiday credit will be deducted.

I've done this in such a way that the json response will still be usable without modifying calling code.  The new invoice date field will be unused initially so I've given it a dummy value of `1970-01-01` until I move the code that works out the actual invoice date.

### DIFF
--- a/lib/holiday-stops/src/main/scala/com/gu/holiday_stops/subscription/HolidayStopCredit.scala
+++ b/lib/holiday-stops/src/main/scala/com/gu/holiday_stops/subscription/HolidayStopCredit.scala
@@ -1,0 +1,16 @@
+package com.gu.holiday_stops.subscription
+
+import java.time.LocalDate
+
+import play.api.libs.json.{Format, Json}
+
+case class HolidayStopCredit(amount: Double, invoiceDate: LocalDate)
+
+object HolidayStopCredit {
+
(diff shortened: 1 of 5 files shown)


lib/holiday-stops/src/main/scala/com/gu/holiday_stops/subscription/HolidayStopCredit.scala
```diff
@@ -0,0 +1,16 @@
...
+
+import java.time.LocalDate
+
+import play.api.libs.json.{Format, Json}
+
+case class HolidayStopCredit(amount: Double, invoiceDate: LocalDate)
+
+object HolidayStopCredit {
+
+  implicit val format: Format[HolidayStopCredit] = Json.format[HolidayStopCredit]
```
frj: Is this actually used, should we ever be converting this to json other than with a HolidayStop

kelvin-chappell: You're quite right; it's not used.  I put it in before I changed the json generation.

PR #441, opened 2019-10-02
Stopped publication date should be after current invoice start date

mario-galic:
We need to be able to handle boundary scenario where `chargedThroughDate` is before `stoppedPublicationDate`. For example consider Guardian Weekly with

```
"processedThroughDate": "2019-07-05",
"chargedThroughDate": "2019-10-05",
```

and holiday scheduled for `2019-10-11` which is after `chargedThroughDate`.

Here processor would run on `2019-10-02` because it runs `processorLeadTime` days in advance, that is 9 days in advance, however at that point next invoice has not yet been generated (it is generated on `2019-10-05`).

So we relax conditions and require just for stopped publication date to not be in the past.


lib/holiday-stops/src/main/scala/com/gu/holiday_stops/subscription/GuardianWeeklySubscription.scala
```diff
@@ -103,14 +117,15 @@ object GuardianWeeklySubscription {
...
     lazy val maybeGwWithIntro =
       (for {
         nForN <- subscription.ratePlans.find(ratePlan => productIsGuardianWeekly(ratePlan) && ratePlan.ratePlanCharges.exists(_.billingPeriod.contains("Specific_Weeks")))
         regular <- subscription.ratePlans.find(ratePlan => productIsGuardianWeekly(ratePlan) && !ratePlan.ratePlanCharges.exists(_.billingPeriod.contains("Specific_Weeks")))
         predictedInvoicedPeriod <- PredictedInvoicedPeriod(regular, nForN)
         regularRpc <- regular.ratePlanCharges.headOption
-        stoppedPublicationDateFallsWithinPredictedInvoicedPeriod = PeriodContainsDate(predictedInvoicedPeriod.startDateIncluding, predictedInvoicedPeriod.endDateExcluding, stoppedPublicationDate.value)
-        _ <- if (stoppedPublicationDateFallsWithinPredictedInvoicedPeriod) Some({}) else None
+        stoppedPublicationDateIsAfterInvoicedPeriodStartDate = stoppedPublicationDate.value.isEqual(predictedInvoicedPeriod.startDateIncluding) || stoppedPublicationDate.value.isAfter(predictedInvoicedPeriod.startDateIncluding)
+        _ <- if (stoppedPublicationDateIsAfterInvoicedPeriodStartDate) Some({}) else None
```
kelvin-chappell: Is there a better way to convert a boolean to an option, or is this to make logic deliberately explicit?

lib/holiday-stops/src/test/resources/StoppedPublicationDateOutsideInvoice.json
```diff
@@ -0,0 +1,130 @@
+{
+  "success": true,
+  "id": "2c92a00b6d4dcd86016d65babf1a5820",
+  "accountId": "2c92a00869b2e5a40169ba617a910b5c",
+  "accountNumber": "A00801499",
+  "accountName": "0010J000023YMVPQA4",
+  "invoiceOwnerAccountId": "2c92a00869b2e5a40169ba617a910b5c",
+  "invoiceOwnerAccountNumber": "A00801499",
+  "invoiceOwnerAccountName": "0010J000023YMVPQA4",
+  "subscriptionNumber": "A-S00831863",
```
kelvin-chappell: Should some of this be anonymised?  Would need to be rebased if so.

lib/holiday-stops/src/main/scala/com/gu/holiday_stops/subscription/GuardianWeeklySubscription.scala
kelvin-chappell: Actually, forget that.  It's fine.

mario-galic: mouse provides `option` extension method.

PR #445, opened 2019-10-03
Zuora does not allow amending cancelled subscriptions with holiday stop

mario-galic:
Consider [`A-S00836178`](https://www.zuora.com/apps/Subscription.do?method=view&id=2c92a0086d4dcd5a016d67fed9ad7d0a). Guardian Weekly subscriber scheduled a holiday stop for `2019-10-11` however subscriptions will be cancelled on `2019-12-07`. Two issues arise

- Zuora prevents amendments on cancelled subscriptions, so issue will still be delivered.
- We cannot apply holiday credit to the next invoice as there is no next invoice. (We can handle this with manual refund).


mario-galic:
[Applying holiday for cancelled subscription which was scheduled for before cancellation date](https://trello.com/c/zarRtOjb/518-applying-holiday-for-cancelled-subscription-which-was-scheduled-for-before-cancellation-date)

PR #447, opened 2019-10-03
Find invoice dates for holiday stops beyond the current billing period

kelvin-chappell:
This determines the first day of each billing period from the current period onwards.  It uses the date of the first day of the billing period following the period in which the holiday stop falls as the effective date of the holiday credit amendment.

It will be interesting to see what happens when a subscription becomes effective on 29 February of a leap year.  If I can find a subscription like that I'll use it as a test case.

There's also some refactoring of unit tests, following discussion [here](https://stackoverflow.com/questions/56900131/why-does-scalatest-by-default-not-enforce-types-when-matching).

### DIFF
--- a/lib/holiday-stops/src/main/scala/com/gu/holiday_stops/subscription/StoppedProduct.scala
+++ b/lib/holiday-stops/src/main/scala/com/gu/holiday_stops/subscription/StoppedProduct.scala
@@ -1,5 +1,6 @@
 package com.gu.holiday_stops.subscription
 
+import java.time.LocalDate.now
 import java.time.{LocalDate, Period}
 
 import acyclic.skipped
@@ -71,7 +72,8 @@ abstract class StoppedProduct(
    */
   private def nextBillingPeriodStartDate: LocalDate = {
 
-    logger.info(s"Calculating nextBillingPeriodStartDate for $this")
+    if (stoppedPublicationDate.isAfter(now.plusYears(5)))
+      logger.info(s"Calculating nextBillingPeriodStartDate for curious case $this")
 
     if (billingPeriod == "Specific_Weeks") invoicedPeriod.endDateExcluding
     else {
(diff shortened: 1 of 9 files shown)


lib/holiday-stops/src/main/scala/com/gu/holiday_stops/subscription/StoppedProduct.scala
```diff
@@ -65,7 +63,43 @@ abstract class StoppedProduct(
...
+      }
+
+      val latestPossibleInvoiceDate = LocalDate.now.plusYears(3)
+
+      @tailrec
+      def go(invoicePeriod: CurrentInvoicedPeriod): LocalDate = {
+        if (invoicePeriod.containsDate(stoppedPublicationDate))
+          invoicePeriod.endDateExcluding
+        else {
+          if (invoicePeriod.endDateExcluding.isBefore(latestPossibleInvoiceDate))
```
mario-galic: Do you need to check for `latestPossibleInvoiceDate`? It would make recursion much nicer without it. Seems impossible case and anyways lambda would timeout.

lib/holiday-stops/src/main/scala/com/gu/holiday_stops/subscription/StoppedProduct.scala
```diff
@@ -53,19 +52,53 @@ abstract class StoppedProduct(
...
+   *
+   * @return Date of the first day of the billing period
+   *         following this <code>stoppedPublicationDate</code>.
+   *         [[com.gu.holiday_stops.subscription.StoppedProductTest]]
+   *         shows examples of the expected outcome.
    */
-  private def nextBillingPeriodStartDate: LocalDate = invoicedPeriod.endDateExcluding
+  private def nextBillingPeriodStartDate: LocalDate = {
+
+    logger.info(s"Calculating nextBillingPeriodStartDate for $this")
```
mario-galic: Maybe `if (stoppedPublicationDate after crazyDistantFuture) logger.warn("blablabl")` to avoid noise.

PR #450, opened 2019-10-04
Remove coveralls coverage support

kelvin-chappell:
Does anyone care about code coverage?
The Coveralls check fails on almost every build but none of us think it's important.
The effect is that we become indifferent to whether our builds are green or red before we deploy them.  This can be dangerous because then we miss things like teamcity failures that we do care about.

I get the impression we're only using Travis to get code coverage.
The Codacy check looks like it's driven directly from Github by a webhook.

I used #160 as a model for what to do.


kelvin-chappell:
Thanks @johnduffell.  Unless there are any objections from anyone I'll merge this on Tuesday morning.

arausuy:
I've been meaning to bring this up for awhile - It makes sense to me to remove it.

PR #451, opened 2019-10-10
enhancing batch email sender to send holiday stop details

frj:
This PR contains changes to allow holiday stops emails to be sent that include a break down of how refunds are going to be applied to the users bills:
- Added stopped credit detail list to EmailBatchItemPayload adding support for new fields to the /email-batch endpoint 
- Replaced hand generation of JValues for EmailPayloadSubscriberAttributes with normal case class based generation, as hand coding more complicated json structures seemed unnecessary. If we want to support other types of email in the future we can abstract EmailToSend on the type of SubscriberAttributes


handlers/batch-email-sender/src/main/scala/com/gu/batchemailsender/api/batchemail/model/EmailBatch.scala
```diff
@@ -80,7 +83,14 @@ object EmailBatch {
...
           stopped_issue_count = emailBatchPayload.holiday_stop_request.map(stop =>
-            StoppedIssueCount(stop.stopped_issue_count))
+            StoppedIssueCount(stop.stopped_issue_count)),
+          stopped_credit_details = emailBatchPayload.holiday_stop_request
+            .flatMap(stop => stop.stopped_credit_details)
+            .map { detailList =>
+              detailList.map { detail =>
+                StoppedCreditDetail(StoppedCreditDetailAmount(detail.amount), StoppedCreditDetailDate(detail.date))
+              }
+            }
```
kelvin-chappell: Could this be done in a for-comprehension?

frj: 👍

PR #452, opened 2019-10-10
Filter out expired rate plans before processing holiday stops

kelvin-chappell:
If a Guardian Weekly subscription had an old rate plan that was migrated to a new plan, it was indeterminate which one would be chosen to calculate holiday credit from. 
This change should ensure that only unexpired rate plans are considered for credit calculations.


lib/holiday-stops/src/main/scala/com/gu/holiday_stops/subscription/GuardianWeeklySubscription.scala
```diff
@@ -1,27 +1,47 @@
...
+      List(
+        "Guardian Weekly - Domestic",
+        "Guardian Weekly - ROW",
+        "Guardian Weekly Zone A",
+        "Guardian Weekly Zone B",
+        "Guardian Weekly Zone C"
+      ).contains(ratePlan.productName)
+
+    lazy val isExpired =
+      ratePlan.ratePlanCharges.exists(_.chargedThroughDate.exists(_.isBefore(MutableCalendar.today)))
```
kelvin-chappell: @mario-galic Is it possible for a current rate plan to have an expired rate plan charge, but have another rate plan charge that is current?

mario-galic: I do not think so. Newspaper Voucher has multiple RatePlanCharges and from my observation so far they all seem to have the same invoice dates.

PR #455, opened 2019-10-14
Snyk fixes

twrichards:
Bit more involved than anticipated... in trying to bump **`jackson-databind`** to fix a load of vulnerabilities, an interesting `assembly` problem occurred during the 'merge' phase, which required creating a new reusable `assemblyMergeStrategy` to 'discard' the duplicate `module-info.class`, which is referenced in all the handlers' `built.sbt` files. (see https://stackoverflow.com/questions/54834125/sbt-assembly-deduplicate-module-info-class)

Also bumped **Scala** and **SBT** versions (and also **`sbt-dependency-graph`** to ensure `snyk monitor` continues to work).


project/Dependencies.scala
```diff
@@ -29,4 +32,14 @@ object Dependencies {
...
   val enumeratum = "com.beachape" %% "enumeratum" % "1.5.13"
+
+  // to resolve merge clash of 'module-info.class'
+  // see https://stackoverflow.com/questions/54834125/sbt-assembly-deduplicate-module-info-class
+  val assemblyMergeStrategyDiscard = assemblyMergeStrategy in assembly := {
+    case PathList("module-info.class") => MergeStrategy.discard
+    case x =>
+      val oldStrategy = (assemblyMergeStrategy in assembly).value
+      oldStrategy(x)
+  }
```
kelvin-chappell: As this only discards duplicates of `module-info` perhaps call it `mergeStrategyDiscardModuleInfo` or something like that so it doesn't give impression all duplicates being discarded?

twrichards: good idea, done in https://github.com/guardian/support-service-lambdas/pull/455/commits/c2eea7af1566f6f84c9e52d256e209ace04aba7d

PR #458, opened 2019-10-17
Adding RevenueEventItemInvoiceItem Query to zuora-datalake-export

wojtek-tg:
## Why
Data Delivery wants to test exporting largest possible dataset using the current methods to determine if we should extend the current process / evaluate some PaaS solution or back the possible direct S3 export.

## Does it work
We've checked the query in Zuora web UI (with limit) and it worked as expected. 

## Related PR:
https://github.com/guardian/ophan-data-lake/pull/4607

### DIFF
--- a/handlers/zuora-datalake-export/src/main/scala/com/gu/zuora/datalake/export/ExportLambda.scala
+++ b/handlers/zuora-datalake-export/src/main/scala/com/gu/zuora/datalake/export/ExportLambda.scala
@@ -214,6 +214,13 @@ object Query extends Enum[Query] {
     "ophan-raw-zuora-increment-invoiceitem",
     "InvoiceItem.csv"
   )
+
+  case object RevenueEventItemInvoiceItem extends Query(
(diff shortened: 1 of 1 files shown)


wojtek-tg:
> Looks good to me, you just need to rebase with master.

Thanks, rebased.

mario-galic:
If `RevenueEventItemInvoiceItem` is larger than `InvoiceItem` then full export will likely not complete within 8 hours at which point Zuora terminates the job. See https://github.com/guardian/support-service-lambdas/blob/master/handlers/zuora-datalake-export/README.md#how-to-extract-a-full-load-of-invoiceitem

wojtek-tg:
I'm making a note that there are problems with Zuora data export and we opened a support case - that's why this isn't merged. I'll update this branch and notify you when I'll be merging this.

mario-galic:
Should this be close?

wojtek-tg:
@mario-galic is right we most probably don't need it now.

PR #462, opened 2019-10-21
write Expected_Invoice_Date__c to Salesforce in the create call

twrichards:
### holiday-stop-api
Refactor to to make the credit calculation call more directly in the create call **(so the invoiceDate is accessible to send to Salesforce as `Expected_Invoice_Date__c`)** which meant that the `CreditCalculation` type alias could be eliminated and needn't be passed around various places in the `Handler`.

### DIFF
--- a/lib/holiday-stops/src/main/scala/com/gu/salesforce/holiday_stops/SalesforceHolidayStopRequest.scala
+++ b/lib/holiday-stops/src/main/scala/com/gu/salesforce/holiday_stops/SalesforceHolidayStopRequest.scala
@@ -5,7 +5,7 @@ import java.time.format.DateTimeFormatter
 import java.util.UUID
 
 import ai.x.play.json.Jsonx
-import com.gu.holiday_stops.{CreditCalculation, ZuoraHolidayError}
+import com.gu.holiday_stops.ZuoraHolidayError
 import com.gu.holiday_stops.subscription.{HolidayStopCredit, StoppedProduct, Subscription}
 import com.gu.salesforce.RecordsWrapperCaseClass
 import com.gu.salesforce.SalesforceConstants._
@@ -173,8 +173,6 @@ object SalesforceHolidayStopRequest extends Logging {
         .runRequest
 
     def buildBody(
-      creditCalculator: CreditCalculation
-    )(
       start: LocalDate,
       end: LocalDate,
       publicationDatesToBeStopped: List[LocalDate],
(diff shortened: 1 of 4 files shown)


lib/holiday-stops/src/main/scala/com/gu/salesforce/holiday_stops/SalesforceHolidayStopRequest.scala
```diff
@@ -187,9 +186,13 @@ object SalesforceHolidayStopRequest extends Logging {
           SF_Subscription__c = sfSubscription.Id,
           Holiday_Stop_Request_Detail__r = RecordsWrapperCaseClass(
             publicationDatesToBeStopped.map { stoppedPublicationDate =>
+              val expectedCredit: Either[ZuoraHolidayError, HolidayStopCredit] =
```
mario-galic: Maybe 

```
            publicationDatesToBeStopped.map { stoppedPublicationDate =>
              StoppedProduct(zuoraSubscription, StoppedPublicationDate(stoppedPublicationDate)).map(_.credit) match {
                case Right(HolidayStopCredit(amount, invoiceDate)) =>
                  CompositeTreeHolidayStopRequestsDetail(
                    stoppedPublicationDate,
                    Some(HolidayStopRequestsDetailChargePrice(amount)),
                    Some(HolidayStopRequestsDetailExpectedInvoiceDate(invoiceDate))
                  )
                case Left(error) =>
                  logger.error(...)
                  CompositeTreeHolidayStopRequestsDetail(stoppedPublicationDate, None, None)
              }
           }
```

twrichards: I did have something a bit like this first but I feel like what I have now is more readable and avoids the repetition of the `CompositeTreeHolidayStopRequestsDetail()`

twrichards: I did have something a bit like this first but I feel like what I have now is more readable and avoids the repetition of the `CompositeTreeHolidayStopRequestsDetail()`

mario-galic: Then maybe add `logger.error` via `tap`

PR #463, opened 2019-10-22
holiday-stop-api: introduce withdrawn timestamp to replace actual deletion

twrichards:
## MUST be released AFTER  https://github.com/guardian/salesforce/pull/93

- re-purpose DELETE endpoint to write a timestamp to `Withdrawn_Time__c`
- retrieve `Withdrawn_Time__c` and `Is_Withdrawn__c` from SF and expose `withdrawnTime` on the 'list existing' endpoint
- re-introduced `mutabilityFlags` so its easy to know on the client whether the holiday stop 
  - `isFullyEditable` (can be withdrawn or start/end dates moved) or 
  - `isEndDateEditable` (where only the end date of the holiday stop can be changed)

### DIFF
--- a/handlers/holiday-stop-api/src/main/scala/com/gu/holiday_stops/WireHolidayStopRequest.scala
+++ b/handlers/holiday-stop-api/src/main/scala/com/gu/holiday_stops/WireHolidayStopRequest.scala
@@ -1,6 +1,6 @@
 package com.gu.holiday_stops
 
-import java.time.LocalDate
+import java.time.{LocalDate, LocalDateTime}
 
 import cats.implicits._
 import com.gu.holiday_stops.subscription.Subscription
@@ -20,7 +20,8 @@ object WireHolidayStopRequest {
       estimatedPrice = detail.Estimated_Price__c.map(_.value),
       actualPrice = detail.Actual_Price__c.map(_.value),
       invoiceDate = detail.Expected_Invoice_Date__c.map(_.value)
-    ))).getOrElse(List())
+    ))).getOrElse(List()),
+    withdrawnTime = sfHolidayStopRequest.Withdrawn_Time__c.map(_.value)
   )
 
   def calculateMutabilityFlags(firstAvailableDate: LocalDate, actionedCount: Int, endDate: LocalDate): MutabilityFlags = {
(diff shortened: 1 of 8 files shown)


handlers/holiday-stop-api/src/main/scala/com/gu/holiday_stops/WireHolidayStopRequest.scala
```diff
@@ -20,25 +22,39 @@ object WireHolidayStopRequest {
...
+    actionedCount: Int,
+    firstPublicationDate: LocalDate,
+    lastPublicationDate: LocalDate
+  ): MutabilityFlags = {
+    if (actionedCount == 0 && firstAvailableDate.isAfter(firstPublicationDate.plusDays(2))) {
       // TODO log warning (with CloudWatch alert) as indicates processing of holiday stop is well overdue
     }
-    // TODO perhaps also check that actioned count is expected value and alert if not
+    val firstPublicationDateIsGreaterOrEqualToFirstAvailableDate =
+      firstPublicationDate.isEqual(firstAvailableDate) || firstPublicationDate.isAfter(firstAvailableDate)
```
kelvin-chappell: I think we do this quite a lot so might be worth having a date method to do it. ie `isEqualToOrAfter` then could be done easier inline.

lib/holiday-stops/src/test/scala/com/gu/salesforce/holiday_stops/SalesForceHolidayStopsEffects.scala
```diff
@@ -16,7 +16,8 @@ object SalesForceHolidayStopsEffects {
 
   def listHolidayQuery(contactId: String, subscriptionName: String) =
     "\n SELECT Id, Start_Date__c, End_Date__c, Subscription_Name__c, Product_Name__c," +
-      "\n Actioned_Count__c, Pending_Count__c, Total_Issues_Publications_Impacted_Count__c, (\n" +
+      "\n Actioned_Count__c, Pending_Count__c, Total_Issues_Publications_Impacted_Count__c," +
+      "\n Withdrawn_Time__c, Is_Withdrawn__c, (\n" +
```
kelvin-chappell: Why not use the `"""|""".stripMargin` format here?

twrichards: no idea i didn't add this test, merely updated it for the new fields

PR #465, opened 2019-10-25
Fix credit calculation for non invoiced subs

frj:
This PR fixes an issue where holiday stops credit/refund dates cannot be calculated for subscriptions that have not been invoiced.

The core issue is:
- The current approach used to map [subscription+holiday stop date] -> [ratePlanCharge+next invoice date] uses chargedThroughDate/processedThrough date and some complex logic to work around 6for6 subscriptions.
- chargedThroughDate/processedThrough are not populated until an invoice has been generated. These values are also updated by zuora on subsequent billing runs, adding to the confusion

These changes replace the above logic with the following:
-  RatePlanCharge.effectiveStartDate and RatePlanCharge.billingPeriod (and other fields) are used to build the expected billing schedule for a RatePlanCharge including when that schedule terminates (eg when the 6for6 period terminates). 
- The schedule is then used to select which ratePlanChange is in effect for a specific stop date.
- The price from ratePlanCharge is used to work out credit amount
- The schedule is used to calculate the next billing date

Future enhancements. 
- The approach to working out the next bill date is over simplified, ie end of the billing period for the stop date +1 day. This works for the current use cases but could be improved to select the next billing period, the complication being that it could be in another ratePlanCharges schedule, eg 6for6 -> 'Normal Weekly' scenario  
- The billing schedule could be made 'day of week' aware(, currently the VoucherSubscription still uses its own logic to select the correct RatePlanCharge based of the day of the week). This would allow us to use common code to create the credit details for both Guardian Weekly and Vouchers


lib/holiday-stops/src/main/scala/com/gu/holiday_stops/subscription/RatePlanChargeInfo.scala
```diff
@@ -0,0 +1,21 @@
+package com.gu.holiday_stops.subscription
+
+import com.gu.holiday_stops.ZuoraHolidayError
+import acyclic.skipped
+
+case class RatePlanChargeInfo(
+  ratePlan: RatePlanCharge,
```
kelvin-chappell: Probably better if called `ratePlanCharge` to avoid confusion

lib/holiday-stops/src/main/scala/com/gu/holiday_stops/subscription/RatePlanChargeBillingSchedule.scala
```diff
@@ -0,0 +1,182 @@
...
+trait RatePlanChargeBillingSchedule {
+  def billingPeriodForDate(date: LocalDate): Either[ZuoraHolidayError, BillingPeriod]
+  def isDateCoveredBySchedule(date: LocalDate): Boolean
+}
+
+case class BillingPeriod(startDate: LocalDate, endDate: LocalDate) {
+  def containsDate(date: LocalDate): Boolean =
+    (date.isEqual(startDate) || date.isAfter(startDate)) &&
+      date.isBefore(endDate)
+}
```
kelvin-chappell: Perhaps needs comment that the `endDate` of the `BillingPeriod` is exclusive.

lib/holiday-stops/src/main/scala/com/gu/holiday_stops/subscription/RatePlanChargeBillingSchedule.scala
```diff
@@ -0,0 +1,182 @@
...
+    optionalBillingPeriodId: Option[String],
+    optionalSpecificBillingPeriod: Option[Int],
+    effectiveStartDate: LocalDate,
+    optionalEndDateCondition: Option[String],
+    upToPeriodsType: Option[String],
+    upToPeriods: Option[Int]
+  ): Either[ZuoraHolidayError, RatePlanChargeBillingSchedule] = {
+    for {
+      endDateCondition <- optionalEndDateCondition.toRight(ZuoraHolidayError("RatePlanCharge.endDateCondition is required"))
+      billingPeriodId <- optionalBillingPeriodId.toRight(ZuoraHolidayError("RatePlanCharge.billingPeriod is required"))
```
kelvin-chappell: If these are both required, shouldn't they be checked before going into this method?  And then types not optional.

lib/holiday-stops/src/main/scala/com/gu/holiday_stops/subscription/RatePlanChargeBillingSchedule.scala
```diff
@@ -0,0 +1,182 @@
...
+          }
+        }
+
+        findBillingPeriodForDate(date, 0)
+      }
+    }
+  }
+
+  private def billingPeriodForZuoraId(zuoraBillingPeriodId: String, optionalSpecificBillingPeriod: Option[Int]): Either[ZuoraHolidayError, Period] = {
+    zuoraBillingPeriodId match {
```
kelvin-chappell: Perhaps `billingPeriodId` should be called `billingPeriodName` throughout to make clear this isn't an ID in the sense of an ID of an object of some sort.

lib/holiday-stops/src/main/scala/com/gu/holiday_stops/subscription/RatePlanChargeBillingSchedule.scala
```diff
@@ -0,0 +1,182 @@
...
+ *   This field/value combination indicates the rate plan end is based on a fixed number of Billing Periods defined by
+ *   the RatePlanCharge.upToPeriods field
+ *
+ *
+ * No other values in these fields are supported attempting to create a billing schedule for those RatePlanCharges will
+ * return an error.
+ *
+ * For more information see:
+ * https://knowledgecenter.zuora.com/DC_Developers/G_SOAP_API/E1_SOAP_API_Object_Reference/ProductRatePlanCharge
+ */
```
kelvin-chappell: Nice commentary on how billing periods work.  I wasn't aware of the different varieties.

lib/holiday-stops/src/main/scala/com/gu/holiday_stops/subscription/RatePlanChargeBillingSchedule.scala
frj: Ah, well spotted the dates are actually inclusive, the containsDate function isnt broken but not 
currently used, for the moment ill remove it.

lib/holiday-stops/src/main/scala/com/gu/holiday_stops/subscription/RatePlanChargeBillingSchedule.scala
frj: The main entry point to RatePlanChargeBillingSchedule through the forRatePlanCharge, these fields are optional on RatePlanCharge. 

Ill maybe just inline this method to make things clearer

lib/holiday-stops/src/main/scala/com/gu/holiday_stops/subscription/RatePlanChargeBillingSchedule.scala
frj: 👍

mario-galic:
Is this related to https://github.com/guardian/support-service-lambdas/pull/551?

PR #466, opened 2019-10-28
holiday-stop-api: surface salesforce error messages

twrichards:
After introducing 'withdrawal' of holiday stops (see https://github.com/guardian/support-service-lambdas/pull/463) which has custom validation (see https://github.com/guardian/salesforce/pull/96) the custom message back from Salesforce needed exposing. This PR does that for 'withdrawal' but also addresses a similar TODO for the 'create' endpoint (the validation there relates to overlapping holiday stops).

This PR also includes re-categorising missing identity/sf_contact header as `400 Bad Request` rather than generic 500 error.

**In order to surface those messages from Salesforce all the way to the surface required some changes deep into the wiring of how REST requests are abstracted in this repo - see https://github.com/guardian/support-service-lambdas/pull/466/commits/98afb0c3ce189a6b5711092c8cd9c56e16b4d611 for only these changes.**

### DIFF
--- a/handlers/holiday-stop-api/src/main/scala/com/gu/holiday_stops/Handler.scala
+++ b/handlers/holiday-stop-api/src/main/scala/com/gu/holiday_stops/Handler.scala
@@ -3,7 +3,6 @@ package com.gu.holiday_stops
 import java.io.{InputStream, OutputStream}
 import java.time.LocalDate
 
-import cats.syntax.either._
 import com.amazonaws.services.lambda.runtime.Context
 import com.gu.effects.{GetFromS3, RawEffects}
 import com.gu.holiday_stops.subscription.{StoppedProduct, Subscription}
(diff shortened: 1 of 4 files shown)


handlers/holiday-stop-api/src/main/scala/com/gu/holiday_stops/Handler.scala
```diff
@@ -18,9 +18,10 @@ import com.gu.util.apigateway.ResponseModels.ApiResponse
 import com.gu.util.apigateway.{ApiGatewayHandler, ApiGatewayRequest, ApiGatewayResponse}
 import com.gu.util.config.LoadConfigModule.StringFromS3
 import com.gu.util.config._
-import com.gu.util.reader.Types._
+import com.gu.util.reader.Types.{logger, _}
```
kelvin-chappell: Do you need `logger` and `_`?

twrichards: good spot, thanks, darn IntelliJ - fixed

lib/salesforce/src/main/scala/com/gu/salesforce/SalesforceClient.scala
```diff
@@ -37,4 +43,23 @@ object SalesforceClient {
...
+
+  private case class SalesforceErrorResponseBody(message: String, errorCode: String)
+  private implicit val readsSalesforceErrorResponseBody = Json.reads[SalesforceErrorResponseBody]
+
+  def parseSalesforceErrorResponseAsCustomError(errorBody: String): ClientFailure = try {
+    Json.parse(errorBody).as[List[SalesforceErrorResponseBody]] match {
+      case singleSfError :: Nil =>  CustomError(s"${singleSfError.errorCode} : ${singleSfError.message}")
+      case multipleSfErrors => CustomError(
+        multipleSfErrors.groupBy(_.errorCode).mapValues(_.map(_.message)).mkString.take(500)
+      )
```
twrichards: FYI this is required to support the failure response from batch calls where there might be multiple errors - not actually used in the current use case, but since this `SalesforceClient` is widely used it should handle this type of response as `shouldExposeSalesforceErrorMessageInClientFailure` might be set to true in future use cases (were batching is being used).

PR #469, opened 2019-10-31
SalesforceClient can now take an optional alternate per-request 'access_token'

twrichards:
… to use as Bearer token for calls to Salesforce (either passed directly or found from optional header `X-Ephemeral-Salesforce-Access-Token`)

This provides us a mechanism to pass a CSR's session ID for example (or other generated access token) in the `X-Ephemeral-Salesforce-Access-Token` header to an API Gateway fronted lambda in this repo so that all Salesforce REST calls out from the lambda will use the provided access token.
As a result any creations/modifications will be correctly attributed to them rather than the API, so rather than...
image
...we'll get something like...
image

**NOTE:** this PR doesn't include utilising this for any of the existing handlers, that will follow in subsequent PR(s).

### DIFF
--- a/lib/salesforce/src/main/scala/com/gu/salesforce/SalesforceClient.scala
+++ b/lib/salesforce/src/main/scala/com/gu/salesforce/SalesforceClient.scala
@@ -22,16 +22,18 @@ object SalesforceClient extends LazyLogging {
           shouldExposeSalesforceErrorMessageInClientFailure.toOption(parseSalesforceErrorResponseAsCustomError _)
         )
       }.setupRequest[StringHttpRequest] {
-        withAuth(sfAuth)
+        withAuthAndBaseUrl(sfAuth)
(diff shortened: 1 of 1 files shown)


lib/salesforce/src/main/scala/com/gu/salesforce/SalesforceClient.scala
```diff
@@ -22,16 +22,18 @@ object SalesforceClient extends LazyLogging {
           shouldExposeSalesforceErrorMessageInClientFailure.toOption(parseSalesforceErrorResponseAsCustomError _)
         )
       }.setupRequest[StringHttpRequest] {
-        withAuth(sfAuth)
+        withAuthAndBaseUrl(sfAuth)
```
twrichards: no change in behaviour just making the method name more representative

PR #471, opened 2019-10-31
Holiday stop cancelation endpoint

frj:
This PR adds the holiday stops cancelation endpoint.

This endpoint applies the logic to update existing holiday stops in the event that the subscription is canceled.

The CSR will have the option to refund the user for holiday stops that have already been booked as the will not be credited during the normal operation of the holiday stops processor, because updates can no longer be made to zuora for that sub and the user may not have another bill for credits to be applied to.

In this case this endpoint will update unprocessed holiday stops details that are for stops before the cancelation date. 

Setting the charge code to 'ManualRefund_Cancellation' and the actual price to the estimated price. 

This supports reporting on the amount of money refunded in the case of holiday stops cancelation and in the event of the sub being uncanceled will prevent the processor issuing  double money back.

If the CSR does not issue a refund then this endpoint should not be called, in this case the processor will ignore the holiday stops due to the subscription being canceled and no credit will be applied.


frj:
[Holiday suspensions when a user cancels: Create new holiday-stop-cancelation-endpoint](https://trello.com/c/J8Y4ztXw/611-holiday-suspensions-when-a-user-cancels-create-new-holiday-stop-cancelation-endpoint)

mario-galic:
> in the event of the sub being uncanceled

What does this mean?

PR #472, opened 2019-10-31
Adding endpoint to get holiday stops affected by cancellation.

frj:
This PR adds support to the /hsr/<sub>/cancel for GET requests.

This will generate a 'preview' of the holiday stops that will get updated during a POST to the same endpoint. For more details see:

https://github.com/guardian/support-service-lambdas/pull/471

This endpoint is intended to be used by the salesforce UI to display information to the CSR about how much to manually refund to the customer in compensation for holiday stops that have been created but will not be honoured because the holiday stop processor will not process cancelled subscriptions.

### DIFF
--- a/handlers/holiday-stop-api/src/test/scala/com/gu/holiday_stops/HandlerTest.scala
+++ b/handlers/holiday-stop-api/src/test/scala/com/gu/holiday_stops/HandlerTest.scala
@@ -280,6 +280,7 @@ class HandlerTest extends FlatSpec with Matchers {
         operation
           .steps(
             cancelHolidayStops(
+              "POST",
               subscriptionName,
               contactId,
               effectiveCancellationDate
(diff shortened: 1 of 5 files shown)


frj:
[Holiday suspensions when a user cancels: Date filtering in Get /holiday-stops endpoint](https://trello.com/c/NeRlnrUX/610-holiday-suspensions-when-a-user-cancels-date-filtering-in-get-holiday-stops-endpoint)

handlers/holiday-stop-api/src/test/scala/com/gu/holiday_stops/HandlerTest.scala
```diff
@@ -348,9 +413,9 @@ class HandlerTest extends FlatSpec with Matchers {
     )
   }
 
-  private def cancelHolidayStops(subscriptionName: String, sfContactId: String, effectiveCancellationDate: LocalDate) = {
+  private def cancelHolidayStops(method: String, subscriptionName: String, sfContactId: String, effectiveCancellationDate: LocalDate) = {
```
twrichards: This might be nicer to be a Boolean entitled `dryRun` and that the `GET` is mapped to `true` and `POST` to `false` **outside** the function - then the function needn't care about http verbs

kelvin-chappell:
Has something gone wrong with the rebase here?  It looks like some commits that are already in master have got included.

frj:
@kelvin-chappell I had two PRs one was merging into the other and it all went a bit weird its fixed now.

PR #487, opened 2019-11-13
Remove header logging from HTTP client

arausuy:
In this PR I'm removing a log line that prints out the headers of a http request - Logging out the plain text headers is not always the best idea :)


twrichards:
IMO it's a shame to lose all headers from the logs, can't we do a `filterKeys` on just the offending header(s)?

arausuy:
Can always pop the headers on debug rather than Info

PR #504, opened 2019-11-28
Add fulfilment-date-calculator bucket path structure

mario-galic:
``` 
s3://bucketName/productName/date_productName.json
```

### DIFF
--- a/handlers/fulfilment-date-calculator/src/main/scala/com/gu/supporter/fulfilment/FulfilmentDateCalculator.scala
+++ b/handlers/fulfilment-date-calculator/src/main/scala/com/gu/supporter/fulfilment/FulfilmentDateCalculator.scala
@@ -10,11 +10,12 @@ import io.github.mkotsur.aws.handler.Lambda._
 
 class FulfilmentDateCalculator extends Lambda[String, String] with LazyLogging {
   override def handle(todayOverride: String, context: Context) = {
-    writeToBucket("changeme.json", """{ "foo": 42 }""")
+    writeToBucket("guardianWeekly", "2020-01-01", """{ "foo": 42 }""")
     Right(todayOverride)
   }
 
-  private def writeToBucket(filename: String, content: String) = {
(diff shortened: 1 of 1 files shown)


mario-galic:
[Write file to fulfilment-date-calculator S3 bucket](https://trello.com/c/30Tj3JC1/726-write-file-to-fulfilment-date-calculator-s3-bucket)

handlers/fulfilment-date-calculator/src/main/scala/com/gu/supporter/fulfilment/FulfilmentDateCalculator.scala
```diff
@@ -10,11 +10,12 @@ import io.github.mkotsur.aws.handler.Lambda._
...
-    writeToBucket("changeme.json", """{ "foo": 42 }""")
+    writeToBucket("guardianWeekly", "2020-01-01", """{ "foo": 42 }""")
     Right(todayOverride)
   }
 
-  private def writeToBucket(filename: String, content: String) = {
+  private def writeToBucket(product: String, date: String, content: String) = {
+    val filename = s"${product}/${date}_${product}.json"
     val s3Client = AmazonS3Client.builder.build
     val stage = System.getenv("Stage").toLowerCase
```
twrichards: is there a reason we're lower-casing the `Stage`, the pattern I've observed in other buckets is to have directories within the bucket for the stage (typically in CAPS)?
If there's not a technical reason why I'll make a PR for this? _I will also add the `Stage` into the filename too (which is handy if they're downloaded to know quickly if you're dealing with DEV/CODE/PROD)._

mario-galic: Separates concerns. CODE cloudformation stack can be deleted without affecting PROD stack. `fulfilment-export-*`. buckets are like this.

PR #505, opened 2019-11-28
Delivery api

frj:
This is an implementation of the delivery records api.

The api contains the following endpoint:

GET /delivery-records/`<subscription name>`

Eg:
```bash
 curl -H "x-salesforce-contact-id: 0033E00001BRLNWQA5" -H "x-api-key: xxx"  https://3hbmr496t2.execute-api.eu-west-1.amazonaws.com/DEV/delivery-records/A-S00052409
```
Returns:
```json
{
  "results": [
    {
      "deliveryDate": "2020-01-03",
      "deliveryInstruction": null,
      "deliveryAddress": null,
      "hasHolidayStop": true
    },
    {
      "deliveryDate": "2020-01-10",
      "deliveryInstruction": "Leave by the gnome",
      "deliveryAddress": "90 York Way, London N1 9AG",
      "hasHolidayStop": true
    }
  ]
} 
```

Some refactoring has been done to the project:

- Code to fetch config from s3 factored out of effects project
- Salesforce case classes and constants factored out to salesforce-core project

The api lambda uses a new stack, details as follows:

- The app uses http4s with a binding for seamless integration with api gateway/lambda requests. See:   
  - https://http4s.org/
  - https://github.com/howardjohn/scala-server-lambda
- The api gateway config uses a 'catchall' proxy resource so no need for addition config for new endpoints 
- Calls to the salesforce api use a new sttp based client, there is also a sales force sttp based salesforce stub using the sttp stubbing framework
- The code uses the tagless final approach for flexibility and the cats effects type classes where appropriate


handlers/delivery-records-api/src/main/scala/com/gu/delivery_records_api/Handler.scala
```diff
@@ -0,0 +1,11 @@
+package com.gu.delivery_records_api
+
+import io.github.howardjohn.lambda.http4s.Http4sLambdaHandler
+import cats.syntax.either._
+
+object Handler extends Http4sLambdaHandler(
+  DeliveryRecordsApiApp()
+    .value
+    .unsafeRunSync()
+    .valueOr(error => throw new RuntimeException(error.toString))
```
kelvin-chappell:
```suggestion
    .valueOr(error => throw new RuntimeException(error))
```

lib/salesforce/core/src/main/scala/com/gu/salesforce/SalesforceQueryConstants.scala
```diff
@@ -0,0 +1,46 @@
...
+    optionalStartDate: Option[LocalDate],
+    optionalEndDate: Option[LocalDate]
+  ) =
+    s"SELECT (" +
+      s"   SELECT Delivery_Date__c, Delivery_Address__c, Delivery_Instructions__c, Has_Holiday_Stop__c" +
+      s"   FROM Delivery_Records__r " +
+      s"   ${deliveryDateFilter(optionalStartDate, optionalEndDate)}" +
+      s") " +
+      s"FROM SF_Subscription__c WHERE Name = '${escapeString(subscriptionNumber)}' " +
+      s"                          AND ${contactToWhereClausePart(contact)}"
```
kelvin-chappell: Would be easier to read as a single text block """..."""

handlers/delivery-records-api/build.sbt
```diff
@@ -0,0 +1,24 @@
...
+
+assemblyJarName := "delivery-records-api.jar"
+riffRaffPackageType := assembly.value
+riffRaffUploadArtifactBucket := Option("riffraff-artifact")
+riffRaffUploadManifestBucket := Option("riffraff-builds")
+riffRaffManifestProjectName := "MemSub::Subscriptions::Lambdas::Delivery Record API"
+riffRaffArtifactResources += (file("handlers/delivery-records-api/cfn.yaml"), "cfn/cfn.yaml")
+
+libraryDependencies ++= Seq(
+)
```
kelvin-chappell: Can this go?

handlers/delivery-records-api/src/test/scala/com/gu/delivery_records_api/DeliveryRecordsApiTest.scala
```diff
@@ -0,0 +1,200 @@
...
+        .stubQuery(auth, deliveryRecordsQuery(IdentityId(identityId), subscriptionNumber, None, None), validSalesforceResponseBody)
+
+    val app = createApp(salesforceBackendStub)
+    val response = app.run(
+      Request(
+        method = Method.GET,
+        Uri(path = s"/delivery-records/${subscriptionNumber}"),
+        headers = Headers.of(Header("x-identity-id", identityId))
+      )
+    ).value.unsafeRunSync().get
```
kelvin-chappell: Do you need to run effect for test?

handlers/delivery-records-api/src/main/scala/com/gu/delivery_records_api/Handler.scala
frj: In this case the 'error' is a case class ie the Left of an either so it needs to be toStringed

handlers/delivery-records-api/src/test/scala/com/gu/delivery_records_api/DeliveryRecordsApiTest.scala
frj: Yes Im afraid so, nothing happens otherwise.

mario-galic:
- Is this going to be followed up with refactoring of all projects to the new approach? If not, then at the point this PR is merged, the codebase becomes more complicated than before as now we have to deal with two different approaches.
- Currently [http4s-lambda](https://mvnrepository.com/artifact/io.github.howardjohn/http4s-lambda) does not have Scala 2.13 build which would prevent us migrating to 2.13. There is already one other dependency like that (which I introduced, aws-lambda-scala).
- Lambdas minimise concurrency and error handling concerns, so I do not see the need for much more beyond basic http client and stdlib Scala. For example, `zuora-datalake-export` and `braze-to-salesforce-file-upload` do not use any of the fancy stuff and it seems to work fine. Also different sub-projects in `support-service-lambdas` are not really related, for example, `batch-email-sender` is a separate concern than `holiday-stop-processor`, etc. so if they are separate concern why force `/lib` on the project.

frj:
@mario-galic 
1. Agreed. Lets try and migrate the other projects to use this approach
2. Good spot, I just took a quick look at the project and adding a cross build for 2.13 should be no problem, ill raise a PR when Ive got a moment.
3. Agreed, I guess the main goal of using these libraries was to improve readability. This is just a first stab at it, we can work on the approach to simplify it further.
4. Not 100% sure I understand the point, I dont think I introduced any additional dependencies on code in libs, just made the libs more fine grained?

mario-galic:
> Good spot, I just took a quick look at the project and adding a cross build for 2.13 should be no problem, ill raise a PR when Ive got a moment.
@guardian/supporter-experience  Upgrade to 2.13 has now been blocked for 4 months. @frj Either remove this dependency, or fix the problem, and do not introduce one-man-show dependencies into our codebase again, especially when it is flagged during code review.

frj:
@mario-galic I raised a PR to add the 2.13 cross build which was merged, so I was hopeful that the project was active. As you're pointing out thats not the case.

Its not a big deal, the part of the dependency we use is really just a few lines of code. 

Ill take a background task of reproducing the the functionality in the code base.

mario-galic:
> Ill take a background task of reproducing the the functionality in the code base.

@guardian/supporter-experience 
 Are you going it do this task, after you migrate the rest of the codebase to http4s, or before? Just for the record, in the past four months how many lines of code have you migrated to your new approach? And how much will your contracting fee be to repay this tech debt?

project/Dependencies.scala
```diff
@@ -28,15 +31,26 @@ object Dependencies {
   val supportInternationalisation = "com.gu" %% "support-internationalisation" % "0.9"
   val circe = "io.circe" %% "circe-generic" % circeVersion
   val circeParser = "io.circe" %% "circe-parser" % circeVersion
+  val circeJava8 = "io.circe" %% "circe-java8" % circeVersion
```
mario-galic, line 34: Another one blocking migration to 2.13.

PR #513, opened 2019-12-04
Provide date override for fulfilment-dates-calculator for testing 

mario-galic:
PROD inputs `null`, we can override with `"yyyy-MM-dd"`

### DIFF
--- a/handlers/fulfilment-date-calculator/src/main/scala/com/gu/supporter/fulfilment/FulfilmentDateCalculator.scala
+++ b/handlers/fulfilment-date-calculator/src/main/scala/com/gu/supporter/fulfilment/FulfilmentDateCalculator.scala
@@ -22,17 +22,19 @@ case class FulfilmentDates(
   nextAffectablePublicationDateOnFrontCover: LocalDate
 )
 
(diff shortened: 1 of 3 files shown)


handlers/fulfilment-date-calculator/src/main/scala/com/gu/supporter/fulfilment/FulfilmentDateCalculator.scala
```diff
@@ -24,15 +24,20 @@ case class FulfilmentDates(
 
 class FulfilmentDateCalculator extends Lambda[String, String] with LazyLogging {
   override def handle(todayOverride: String, context: Context) = {
-    val today = LocalDate.now()
-    val todayAsString = today.format(DateTimeFormatter.ofPattern("yyyy-MM-dd"))
-    val fulfilmentDates = FulfilmentDates(today, today, today, today, today, today)
-    writeToBucket("WEEKLY", todayAsString, fulfilmentDates.asJson.spaces2)
-    Right(todayOverride)
+    val today = inputToDate(todayOverride)
```
twrichards: to we need it to default to today if nothing is provided?

PR #516, opened 2019-12-06
fulfilment-dates-calculator: fetch & use bank holidays from gov.uk in isWorkingDay

mario-galic:
Lazily fetch `https://www.gov.uk/bank-holidays.json`...

- if valid, store it to the S3 bucket (for use as a future fallback) and return it as the case class `GovUkBankHolidays`
- if gov.uk is down or response is invalid, return the one from the bucket as fallback and return it as the case class `GovUkBankHolidays`

This list is used in the `isWorkingDay` function quite neatly (see https://github.com/guardian/support-service-lambdas/pull/516/files#diff-34905b31c269a802cee4d6bb38ff1e20R19)


twrichards:
we could have it write the response to S3 each time, as a fallback in case the service is unreachable

handlers/fulfilment-date-calculator/src/main/scala/com/gu/supporter/fulfilment/BankHoliday.scala
```diff
@@ -0,0 +1,27 @@
...
+  `england-and-wales`: BankHolidays,
+  scotland: BankHolidays,
+  `northern-ireland`: BankHolidays
+)
+
+case class Event(
+  title: String,
+  date: String,
+  notes: String,
+  bunting: Boolean
```
kelvin-chappell: Will you be using the bunting?

PR #522, opened 2019-12-11
using zuora "product type" to generate paths for fulfilment date files

frj:
This changes the path generation of fulfilment date files so they are written to paths based on the  zuora product type. 

eg:
```
s3://fulfilment-date-calculator-dev/Newspaper - Home Delivery/2019-12-11_Newspaper - Home Delivery.json
```
This avoid clients which have access to the zuora subscription from needing to store mappings from the product type to the identifiers used in these paths.


twrichards:
I like this in principal - but not so sure about `+`s for the space character - could we do a replace with `_` perhaps??

PR #533, opened 2019-12-17
Use fulfilment dates file for holiday stops

frj:
This PR reworks the firstAvailableDate calculation in the hoilday stops api get exsiting endpoint to use the holidayStopFirstAvailableDate date from the fulfilment dates file.

The details are:
- Added productType to SubscriptionData, this could be obtained from Zuora however its unfortunately not part of the json returned from the subscription endpoint, hence its been hard coded in the SupportedProducts class
- exposed editionDaysOfWeek on SubscritionData to support generating the correct 'IssueSpecifics' for a subscription
- subscriptionAnnualIssueLimit is calculated using a hard coded issue limit per edition, multiplied by the number of editionDaysOfWeek, this works for EchoLegacy as well as the other product types
- Wiring up FulfilmentDatesFetcher into stepsToListExisting and generating the IssueSpecifics from the dates and the SubscriptionData fields


handlers/holiday-stop-api/src/main/scala/com/gu/holiday_stops/Handler.scala
```diff
@@ -209,9 +216,16 @@ object Handler extends Logging {
         .toApiGatewayOp(s"lookup Holiday Stop Requests for contact $contact")
       subscription <- getSubscription(subName)
         .toApiGatewayOp(s"get subscription $subName")
+      subscriptionData <- SubscriptionData(subscription)
+        .toApiGatewayOp(s"extract subscription data from subscription")
+      fulfilmentDates <- fulfilmentDatesFetcher.getFulfilmentDates(
+        subscriptionData.productType,
+        MutableCalendar.today
```
twrichards: is this the best way to get `today` - we have used `now` on `LocalDate` everywhere else - the use of anything with `Mutable` in the name should be avoided

frj: This is 'mutable' so you can set its value in tests

frj: BTW there is a java.time.Clock which fulfils a similar purpose, you would need to pass it into the function your testing, ie no mutable/singleton state which I agree we 
would be better without.

twrichards: I see why you wanted this, for testing the endpoint where it would be nasty to pass it a date... but ironically the only time you fake it you're giving it `now` anyway - `MutableCalendar.setFakeToday(Some(LocalDate.now()))` https://github.com/guardian/support-service-lambdas/pull/533/files#diff-2c79809f0da6bd8e91b0e514396c9951R168 so it can probably be eliminated for now - can cross that bridge if/when one ever wants to test end to end with an artificial 'today'.

frj: As discussed. MutableCalandar is something we introduced to get round issues testing subscriptions that would expire at some point and test would start failing randomly.

For consistency we should probably get the time in the same way all every where in the code.

The test where I set MutableCalandar to now is just overriding the value set in other tests.

We could migrate to passing Clock around the code which would avoid having a magic mutable value to set the time or we could have a look at refactoring the tests so they all work with the current time. 

Thats probably non trivial so ill leave it as is for the moment.

lib/holiday-stops/src/main/scala/com/gu/holiday_stops/ActionCalculator.scala
```diff
@@ -1,205 +1,13 @@
...
-            )
-          }
-        )
-      }
-  }
-
-  def latestOf(head: LocalDate, tail: LocalDate*) = (head :: tail.toList).max[LocalDate](_ compareTo _)
-}
-
-case class ActionCalculatorError(message: String)
```
kelvin-chappell: 🔥

PR #536, opened 2019-12-23
Use fulfilment dates file in holiday stops processor

frj:
This PR contains the changes to use the fulfilment dates file to decide what date holiday stops should be process for by the holiday stops processor.

This eliminates any code differences between the processing for different types subscription ie processing for Weekly/Voucher/Delivery only varies in how far ahead of time stops should be processed.

This also enables processing of holiday stops for home delivery subscriptions.

Details as follows:
- Fulfilment date file is fetched from s3 for a particular product type and date, any days that have a holidayStopProcessorTargetDate are processed for the specified dates. This could be multiple dates eg on a thursday where holiday stops need to be processed for Sat/Sun/Mon issues in time for the fulfilment file to be generated on the Friday. This allows the processor to be run once per product type rather than multiple times for each product variant.
- SOQL to get holiday stops updated to filter by product type and multiple dates.
- Deleted redundant product variant specific processor.


lib/holiday-stops/src/main/scala/com/gu/salesforce/holiday_stops/SalesforceHolidayStopRequestsDetail.scala
```diff
@@ -81,57 +80,38 @@ object SalesforceHolidayStopRequestsDetail extends Logging {
       | Estimated_Price__c, Charge_Code__c, Actual_Price__c, Expected_Invoice_Date__c
       |""".stripMargin
 
-  private def soqlFilterClause(stoppedPublicationDate: LocalDate) = s"""
-      | Stopped_Publication_Date__c = ${stoppedPublicationDate.toString}
+  private def soqlFilterClause(stoppedPublicationDates: List[LocalDate]) = s"""
+      | Stopped_Publication_Date__c IN (${stoppedPublicationDates.map(_.toString).mkString(", ")})
```
kelvin-chappell: Might want to explicitly map the dates to strings in a function to make their format clear.
Eg. `def toYyyyMmDdFormat(d: LocalDate) = d.toString` or something like that

lib/holiday-stops/src/main/scala/com/gu/salesforce/holiday_stops/SalesforceHolidayStopRequestsDetail.scala
```diff
@@ -81,57 +80,38 @@ object SalesforceHolidayStopRequestsDetail extends Logging {
...
       logger.info(s"using SF query : $soqlQuery")
       RestRequestMaker.GetRequestWithParams(RelativePath(soqlQueryBaseUrl), UrlParams(Map("q" -> soqlQuery)))
     }
+
+    def createSoql(dates: List[LocalDate], productType: ZuoraProductType) = {
+      s"""
+         | $SOQL_SELECT_CLAUSE
+         | FROM $holidayStopRequestsDetailSfObjectRef
+         | WHERE Holiday_Stop_Request__r.SF_Subscription__r.Product_Type__c = '${productType.name}'
+         | AND${soqlFilterClause(dates)}
```
kelvin-chappell:
```suggestion
         | AND ${soqlFilterClause(dates)}
```

lib/holiday-stops/src/main/scala/com/gu/salesforce/holiday_stops/SalesforceHolidayStopRequestsDetail.scala
```diff
@@ -81,57 +80,38 @@ object SalesforceHolidayStopRequestsDetail extends Logging {
...
+      val soqlQuery = createSoql(dates, productType)
       logger.info(s"using SF query : $soqlQuery")
       RestRequestMaker.GetRequestWithParams(RelativePath(soqlQueryBaseUrl), UrlParams(Map("q" -> soqlQuery)))
     }
+
+    def createSoql(dates: List[LocalDate], productType: ZuoraProductType) = {
+      s"""
+         | $SOQL_SELECT_CLAUSE
+         | FROM $holidayStopRequestsDetailSfObjectRef
+         | WHERE Holiday_Stop_Request__r.SF_Subscription__r.Product_Type__c = '${productType.name}'
```
kelvin-chappell: Do all the product type names definitely correspond with their SF counterparts?

frj: So I believe that the product types are derived from the rate plan charges of the zuora subscription during the process of syncing the subscriptions to salesforce.

PR #543, opened 2020-01-08
Zioify configuration of holiday-stop code

kelvin-chappell:
This is a small first step in integrating ZIO into the holiday-stop code.  It just wraps the config in a ZIO wrapper.  The real benefits will come when other services are zioified.  But this is to keep the change small, easier to review and less risky to deploy.

Immediately some parameters whose only purpose is to facilitate testing become redundant, but I haven't removed these redundant parameters yet.

There should be no difference in the effect of the code.


handlers/holiday-stop-api/src/main/scala/com/gu/holiday_stops/Handler.scala
```diff
@@ -29,38 +29,56 @@ import com.softwaremill.sttp.{HttpURLConnectionBackend, Id, SttpBackend}
...
-        operationForEffects(
-          RawEffects.response,
-          RawEffects.stage,
-          GetFromS3.fetchString,
-          HttpURLConnectionBackend(),
-          UUID.randomUUID().toString
-        )
-      )
+
+    val runtime = new DefaultRuntime {}
```
regiskuckaertz: I suggest creating the runtime as a static value outside of `apply`, that will significantly reduce latency.

lib/holiday-stops/src/main/scala/com/gu/holiday_stops/ConfigLive.scala
```diff
@@ -0,0 +1,11 @@
+package com.gu.holiday_stops
+
+import com.gu.effects.GetFromS3
+import zio.IO
+
+trait ConfigLive extends Configuration {
+  val configuration: Configuration.Service = new Configuration.Service {
+    val config: IO[OverallFailure, Config] =
+      IO.fromEither[OverallFailure, Config](Config.fromS3(GetFromS3.fetchString))
```
regiskuckaertz: It looks like this will perform a synchronous request to S3? If yes, I suggest wrapping each call to S3 in `IO.effect`, i.e.:
```scala
IO.effect(GetFromS3.fetchString).flatMap(Config.fromS3(_)).absolve
```

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/Handler.scala
```diff
@@ -5,35 +5,48 @@ import java.time.LocalDate
...
-
-      case Right(config) =>
-        val results = Processor.processAllProducts(config, processDateOverride, HttpURLConnectionBackend(), GetFromS3.fetchString)
-        results.foreach(result => ProcessResult.log(result))
-        results.flatMap(_.overallFailure.toList) match {
+
+    val main: ZIO[Console with Configuration, io.Serializable, List[ZuoraHolidayWriteResult]] =
+      for {
+        config <- Configuration.factory.config.tapError(e => putStrLn(s"Config failure: ${e.reason}"))
+        results <- ZIO.effect(Processor.processAllProducts(config, processDateOverride, HttpURLConnectionBackend(), GetFromS3.fetchString))
```
regiskuckaertz: It looks as thought the costly operations are inside that `Processor` thing? Anything this can be "zioified" too?

kelvin-chappell: Yes there's plenty in there to zioify but I'm just taking a top-down approach and wrapping the whole thing up.  My plan is to take on one effectful service at a time and simplify the code as I go.

lib/holiday-stops/src/main/scala/com/gu/holiday_stops/ConfigLive.scala
kelvin-chappell: `GetFromS3.fetchString` is a single-arg function so I'm not sure this would work.  Although I could try `IO.effect(Config.fromS3(GetFromS3.fetchString)).absolve`.  The failure type then becomes a `Serializable` because `OverallFailure` isn't an exception at the moment.  It should make a single call to S3.

handlers/holiday-stop-api/src/test/scala/com/gu/holiday_stops/ConfigTest.scala
```diff
@@ -0,0 +1,13 @@
+package com.gu.holiday_stops
+
+import java.io.Serializable
+
+import com.gu.effects.FakeFetchString
+import zio.IO
+
+trait ConfigTest extends Configuration {
+  val configuration: Configuration.Service = new Configuration.Service {
+    val config: IO[Serializable, Config] =
```
mario-galic: Why we have to specify `Serializable`?

kelvin-chappell: I'm now converting `ApiFailure`s into `RuntimeException`s so that there no longer needs to be a `Serializable` in the return type, as the exceptional case is always a `Throwable`.

PR #544, opened 2020-01-09
[delivery-records-api] expose delivery problem cases

twrichards:
On each delivery record in the `results` list there is now a `problemCaseId` if that delivery record has a problem case associated.

There is also a new top level key `deliveryProblemMap`, which holds extra detail about the cases with ID's specified above, like...

- `subject`
- `description`
- `problemType` (from the 'Case Subcategory` in Salesforce)

### DIFF
--- a/handlers/delivery-records-api/src/main/scala/com/gu/delivery_records_api/DeliveryRecordsService.scala
+++ b/handlers/delivery-records-api/src/main/scala/com/gu/delivery_records_api/DeliveryRecordsService.scala
@@ -4,11 +4,11 @@ import java.time.LocalDate
 
 import cats.Monad
 import cats.data.EitherT
-import com.gu.salesforce.{Contact, RecordsWrapperCaseClass, SFApiDeliveryRecord}
-import com.gu.salesforce.SalesforceQueryConstants.deliveryRecordsQuery
+import com.gu.salesforce.{Contact, RecordsWrapperCaseClass}
 import com.gu.salesforce.sttp.SalesforceClient
 import io.circe.generic.auto._
 import cats.implicits._
+import com.gu.salesforce.SalesforceQueryConstants.{contactToWhereClausePart, escapeString}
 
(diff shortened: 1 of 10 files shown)


handlers/delivery-records-api/src/main/scala/com/gu/delivery_records_api/DeliveryRecordsService.scala
```diff
@@ -107,5 +125,33 @@ object DeliveryRecordsService {
...
+  def deliveryRecordsQuery(
+    contact: Contact,
+    subscriptionNumber: String,
+    optionalStartDate: Option[LocalDate],
+    optionalEndDate: Option[LocalDate]
+  ) =
+    s"""SELECT (
+       |    SELECT Delivery_Date__c, Delivery_Address__c, Delivery_Instructions__c, Has_Holiday_Stop__c, Address_Line_1__c,
+       |           Address_Line_2__c, Address_Line_3__c, Address_Town__c, Address_Country__c, Address_Postcode__c,
+       |           Case__c, Case__r.Id, Case__r.Subject, Case__r.Description, Case__r.Case_Closure_Reason__c
```
twrichards: this is the only new line, the rest was moved from shared lib since it's delivery record specific

lib/salesforce/sttp-client/src/main/scala/com/gu/salesforce/sttp/SalesforceClient.scala
```diff
@@ -70,7 +72,11 @@ object SalesforceClient extends LazyLogging {
             "Authorization" -> s"Bearer ${auth.access_token}",
             "X-SFDC-Session" -> auth.access_token,
           )
-          .response(asJson[QueryRecordsWrapperCaseClass[A]])
+          .mapResponse(responseBodyString => {
+            logger.info(responseBodyString)
```
twrichards: also added logging of raw response

PR #548, opened 2020-01-16
delivery-records-api : detect changes in delivery address and delivery instructions

twrichards:
...exposed via `isChangedAddress` and `isChangedDeliveryInstruction` respectively.

In the UI we highlight changes in delivery address and/or delivery instructions (so users perhaps can realise why they're having a delivery problem), detecting these was attempted in the client (see https://github.com/guardian/manage-frontend/pull/353#discussion_r366918664) but is better in the API... hence this PR.

### DIFF
--- a/handlers/delivery-records-api/src/main/scala/com/gu/delivery_records_api/DeliveryRecordsService.scala
+++ b/handlers/delivery-records-api/src/main/scala/com/gu/delivery_records_api/DeliveryRecordsService.scala
@@ -10,6 +10,8 @@ import io.circe.generic.auto._
 import cats.implicits._
 import com.gu.salesforce.SalesforceQueryConstants.{contactToWhereClausePart, escapeString}
 
+import scala.annotation.tailrec
+
 final case class DeliveryRecord(
   deliveryDate: Option[LocalDate],
   deliveryInstruction: Option[String],
@@ -21,7 +23,9 @@ final case class DeliveryRecord(
   addressCountry: Option[String],
   addressPostcode: Option[String],
   hasHolidayStop: Option[Boolean],
-  problemCaseId: Option[String]
+  problemCaseId: Option[String],
+  isChangedAddress: Option[Boolean],
+  isChangedDeliveryInstruction: Option[Boolean]
 )
 
 final case class DeliveryProblemCase(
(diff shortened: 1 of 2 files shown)


handlers/delivery-records-api/src/main/scala/com/gu/delivery_records_api/DeliveryRecordsService.scala
```diff
@@ -52,6 +54,36 @@ object DeliveryRecordsService {
...
+  private def detectChangeSkippingNones[T](
+    allPrevious: List[DeliveryRecord],
+    fieldExtractor: DeliveryRecord => Option[T]
+  )(
+    value: T
+  ): Boolean = allPrevious match {
+    case head :: tail => fieldExtractor(head) match {
+      case Some(previousValue) => value != previousValue // this detects the change
+      case None => detectChangeSkippingNones(tail, fieldExtractor)(value) // this skips over a None at the head
+    }
```
kelvin-chappell: This could be done in a single set of cases:
```
case Some(previousValue) :: tail => ...
case None :: tail => ...
```

twrichards: not really (I did set out to do that) - if you notice the inner match is based on **`fieldExtractor(head)`**  - if I was to do it with a single pattern match, I'd have to do something like 
```case DeliveryRecord(_, _, _, _, _, _, _, _, _, Some(previousValue),_, _, _) :: tail```
 which I figure was uglier and harder to maintain.

What I've done is more functional since the `fieldExtractor` of type `T` is passed in, making it easier if we wanted to detect changes in other fields.

Comment (kelvin-chappell) [reply handlers/delivery-records-api/src/main/scala/com/gu/delivery_records_api/DeliveryRecordsService.scala]: I see what you mean!

File: handlers/delivery-records-api/src/main/scala/com/gu/delivery_records_api/DeliveryRecordsService.scala [hunk 2]
```diff
@@ -52,6 +54,36 @@ object DeliveryRecordsService {
     Delivery_Records__r: Option[RecordsWrapperCaseClass[SFApiDeliveryRecord]]
   )
 
+  private def detectChangeSkippingNones[T](
```
kelvin-chappell:
```suggestion
  private def detectChangeSkippingNoneAtHead[T](
```

handlers/delivery-records-api/src/main/scala/com/gu/delivery_records_api/DeliveryRecordsService.scala
```diff
@@ -52,6 +54,36 @@ object DeliveryRecordsService {
...
+    value: T
+  ): Boolean = allPrevious match {
+    case head :: tail => fieldExtractor(head) match {
+      case Some(previousValue) => value != previousValue // this detects the change
+      case None => detectChangeSkippingNones(tail, fieldExtractor)(value) // this skips over a None at the head
+    }
+    case Nil => false // if we reach the beginning of the list and we don't have an address to compare to, don't mark this as changed
+  }
+
+  private def transformSfApiDeliveryRecords(accumulator: List[DeliveryRecord], sfRecord: SFApiDeliveryRecord): List[DeliveryRecord] =
```
kelvin-chappell:
```suggestion
  private def transformSfApiDeliveryRecords(accumulator: List[DeliveryRecord], sfRecord: SfApiDeliveryRecord): List[DeliveryRecord] =
```

twrichards: i'm following the naming convention already present in the repo

handlers/delivery-records-api/src/main/scala/com/gu/delivery_records_api/DeliveryRecordsService.scala
twrichards: done

PR #549, opened 2020-01-20
Move credit-processor lib into its own project

kelvin-chappell:
This is the final step in making the credit processor available for different types of credit.
So we can now use it to process holiday-stop credits, and very soon delivery-problem credits as well. 

The project dependency hierarchy now looks like this:
```
holiday-stop-processor -> holiday-stops -> zuora-core
holiday-stop-processor -> credit-processor -> zuora-core
```

See also #547

Tested holiday-stop processor in Code.


lib/credit-processor/src/main/scala/com/gu/creditprocessor/Processor.scala
```diff
@@ -60,29 +62,22 @@ object Processor {
...
-        subscription,
-        request.publicationDate.value,
-        maybeExtendedTerm,
-        HolidayStopCredit(issueData.credit, issueData.nextBillingPeriodStartDate)
-      )
-      _ <- if (subscription.hasCreditAmendment(request)) Right(()) else updateSubscription(subscription, subscriptionUpdate)
+      subscriptionUpdate <- updateToApply(creditProduct, subscription, request.publicationDate)
+      _ <- if (subscription.hasCreditAmendment(request)) Right(()) else {
+        updateSubscription(subscription, subscriptionUpdate)
+      }
```
mario-galic: Indentation looks a bit strange

PR #550, opened 2020-01-21
Fix credit date calculation by recalculating effective start date

frj:
This includes some fixes to the invoice date calculation used when refunding credit to the user.

Refunds were broken because credits were being applied to subscriptions with dates that did not align with the invoice dates for the main rate plan charges for that subscription. This would result in the refunds being applied in a separate "negative" invoice rather than as a reduced amount on the normal invoice. Zuora does not automatically convert negative invoices to credit on an account, hence the customer would not receive the refund.

Additional logic was implemented in order to get the correct invoice date for specific types of subscription:
* Some subscriptions invoice dates are not derived from RatePlanCharge.effectiveStartDate. These were detected checking the output of the billing schedule calculation against the processedThroughDate (which is effectively the date of the last invoice for the rate plan charge).
* For subs where the effectiveStartDate was not correct the following fields were used to get the start date of the subscription
  * RatePlanCharge.tirggerEvent. This indicates what date to use for the start of the rate plan charge. This can be:
    * CustomerAcceptanceDate - from subscripiton
    * ContracttEffectiveDate - from subscription
    * SpecificDate - supplied in RatePlanCharge.tirggerDate
  * BillingDay indicates which day the invoice is generated.
    * ChargeTriggerDay - most rate plan charges will generate an invoice on the start date of the charge
    * DefaultFromCustomer - some Echo legacy subscriptions adjust the invoice date to the bill cycle day defined in the customers account


lib/zuora-core/src/main/scala/com/gu/zuora/subscription/RatePlanChargeBillingSchedule.scala
```diff
@@ -121,39 +222,105 @@ object RatePlanChargeBillingSchedule {
...
+    val dateWithCorrectMonth = if (date.getDayOfMonth < billCycleDay) {
+      date.plusMonths(1)
+    } else {
+      date
+    }
+
+    val lastDateOfMonth = dateWithCorrectMonth `with` TemporalAdjusters.lastDayOfMonth()
+
+    dateWithCorrectMonth.withDayOfMonth(Math.min(lastDateOfMonth.getDayOfMonth, billCycleDay))
+  }
```
kelvin-chappell: It seems strange how the Java time lib does date operations:
```
Welcome to the Ammonite Repl 1.8.1
(Scala 2.12.10 Java 1.8.0_232)
If you like Ammonite, please support our development at www.patreon.com/lihaoyi
@ import java.time._
import java.time._

@ LocalDate.of(2020,1,31).plusMonths(1)
res1: LocalDate = 2020-02-29
```
I thought it should be `2020-03-02`

frj: That is the 'magic' of plusMonths, it give you the equivalent day in however many months time.

frj:
I agree its getting more complicated.

To be honest using the api is not such a bad idea as it eliminates the complexity.

I didn't do it as it would be a fairly fiddly change and there would inevitably be a bunch of edge cases we would need to work through to get to the point where the current issue was fixed.

This felt like the way to go to fix the current issue but it might be worth looking into moving to use the api in the future.

mario-galic:
Here is implementation based on out-of-the-box billing preview https://github.com/guardian/support-service-lambdas/tree/switch-over-to-invoicing-api-preview-which-uses-zoura-billing-preview

PR #554, opened 2020-02-03
Implement delivery-problem credit processing lambda

kelvin-chappell:
This adds a new project to hold the lambda, and makes a few changes elsewhere: adding a patch method to the STTP salesforce client, and moving a few credit-processing files around.

Currently the delivery date is held in Zuora in the `HolidayStart__c` and `HolidayEnd__c` fields, which seems a bit odd but avoids creating new custom fields.  It would be difficult not to store this anywhere.

There are [new rate plans](https://github.com/guardian/support-service-lambdas/blob/6d9e62b0f4f26bff61967696dcdfbbbd0a48eabc/handlers/delivery-problem-credit-processor/src/main/scala/com/gu/deliveryproblemcreditprocessor/DeliveryCreditProduct.scala) in each stage to record these delivery-problem credits in Zuora, so that they're distinguishable from holiday-stop credits.

Also the account that writes to Zuora is the holiday-stop-processor user, which might be confusing but can be changed later.


handlers/delivery-problem-credit-processor/src/main/scala/com/gu/deliveryproblemcreditprocessor/DeliveryCreditProcessor.scala
```diff
@@ -0,0 +1,227 @@
...
+
+  private lazy val zuoraConfig: Task[ZuoraConfig] =
+    config(
+      LoadConfigModule(stage, GetFromS3.fetchString)
+        .apply[ZuoraConfig](ConfigLocation("zuoraRest", 1), ZuoraConfig.reads)
+        .toEither
+    )
+
+  private def config[A](a: Either[ConfigFailure, A]): Task[A] =
+    Task.effect(a).absolve.mapError {
```
frj: This is more of a general comment rather than something we should do now, but Im thinking we could use the cats effect type classes rather than a specific effect implementation in the code eg:
```
Sync[F].delay(a)....
```
And just commit to an effect implementation as late as possible.  

That way if we want to extract code to a shared lib it will the least amount of work to make it compatible with as much other code as possible.

handlers/delivery-problem-credit-processor/cfn.yaml
```diff
@@ -0,0 +1,134 @@
...
+
+Conditions:
+  IsProd: !Equals [!Ref "Stage", "PROD"]
+
+Mappings:
+  StageMap:
+    CODE:
+      ScheduleName: delivery-problem-credit-processor-schedule-code
+    PROD:
+      ScheduleName: delivery-problem-credit-processor-schedule
```
mario-galic: Other cloudformations do not seem to follow this convention of dropping PROD postfix for PROD names.

handlers/delivery-problem-credit-processor/src/main/scala/com/gu/deliveryproblemcreditprocessor/DeliveryCreditProcessor.scala
```diff
@@ -0,0 +1,227 @@
...
+    for {
+      sfAuthConfig <- sfConfig
+      zConfig <- zuoraConfig
+      zAccessToken <- zuoraAccessToken(zConfig)
+      processResults <- Task.foreach(productTypes)(processProduct(sfAuthConfig, zConfig, zAccessToken))
+      creditResults <- Task.foreach(processResults)(dealWithProcessResult)
+    } yield creditResults.flatten
+  }
+
+  def dealWithProcessResult(processResult: ProcessResult[DeliveryCreditResult]): Task[List[DeliveryCreditResult]] =
```
mario-galic: By `dealWith` you mean `handle`? When the name lacks meaning it signals for possible inlining of functionality.

handlers/delivery-problem-credit-processor/src/main/scala/com/gu/deliveryproblemcreditprocessor/DeliveryCreditProcessor.scala
```diff
@@ -0,0 +1,227 @@
...
+          LocalDateTime.now
+        )
+        salesforceClient.patch(
+          deliveryObject,
+          objectId = result.deliveryId,
+          body = actioned
+        ).leftMap { e =>
+          SalesforceApiFailure(e.message)
+        }
+      }.sequence.map { _ => () }
```
mario-galic: Why is the right result discarded with `.map( _ => ())`? Maybe instead we could use existential type `SalesforceApiResponse[_]` to indicate actual value does not matter.

handlers/delivery-problem-credit-processor/src/test/scala/com/gu/deliveryproblemcreditprocessor/DeliveryCreditRequestTest.scala
```diff
@@ -0,0 +1,38 @@
...
+import org.scalactic.TypeCheckedTripleEquals
+import org.scalatest.{EitherValues, FlatSpec, Matchers}
+
+import scala.io.Source
+
+class DeliveryCreditRequestTest
+  extends FlatSpec
+  with Matchers
+  with EitherValues
+  with TypeCheckedTripleEquals {
```
mario-galic: It might we worth factoring out `extends FlatSpec with Matchers with EitherValues with TypeCheckedTripleEquals` into our own trait as we use it in most of our tests.

handlers/delivery-problem-credit-processor/src/test/scala/com/gu/deliveryproblemcreditprocessor/DeliveryCreditRequestTest.scala
```diff
@@ -0,0 +1,38 @@
...
+
+  "Json decode" should "decode SF response correctly" in {
+    val json = Source.fromResource("sf-credit-request.json").mkString
+    decode[RecordsWrapperCaseClass[DeliveryCreditRequest]](json).right.value should ===(
```
mario-galic: Maybe we should start using [diffx-scalatest](https://stackoverflow.com/a/60025871/5205022) for case class pretty diffs.

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/HolidayCreditProduct.scala
```diff
@@ -10,19 +10,25 @@ import com.gu.zuora.subscription.CreditProduct
...
  */
 object HolidayCreditProduct {
+
+  val productRatePlanChargeName = "Holiday Credit"
+
   val Prod = CreditProduct(
     productRatePlanId = "2c92a0076ae9189c016b080c930a6186",
-    productRatePlanChargeId = "2c92a0086ae928d7016b080f638477a6"
+    productRatePlanChargeId = "2c92a0086ae928d7016b080f638477a6",
+    productRatePlanChargeName
```
mario-galic: Maybe add trailing comma.

project/Dependencies.scala
```diff
@@ -44,6 +44,8 @@ object Dependencies {
   val http4sServer = "org.http4s" %% "http4s-server" % http4sVersion
   val catsCore = "org.typelevel" %% "cats-core" % catsVersion
   val catsEffect = "org.typelevel" %% "cats-effect" % catsEffectVersion
+  val scalaLambda = "io.github.mkotsur" %% "aws-lambda-scala" % "0.1.1"
```
mario-galic: aws-lambda-scala 0.2.0 now supports Scala 2.13

handlers/delivery-problem-credit-processor/cfn.yaml
kelvin-chappell: Now following the convention.

handlers/delivery-problem-credit-processor/src/main/scala/com/gu/deliveryproblemcreditprocessor/DeliveryCreditProcessor.scala
kelvin-chappell: I've renamed it and inlined its child methods.

handlers/delivery-problem-credit-processor/src/main/scala/com/gu/deliveryproblemcreditprocessor/DeliveryCreditProcessor.scala
kelvin-chappell: Done now.

handlers/delivery-problem-credit-processor/src/test/scala/com/gu/deliveryproblemcreditprocessor/DeliveryCreditRequestTest.scala
kelvin-chappell: Difficult to know where to put this trait so I'll leave it for now.  Currently there's only one test class in this subproject.

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/HolidayCreditProduct.scala
kelvin-chappell: This isn't really a list that's likely to grow so will leave out the trailing comma.

handlers/delivery-problem-credit-processor/src/test/scala/com/gu/deliveryproblemcreditprocessor/DeliveryCreditRequestTest.scala
kelvin-chappell: Not much benefit in this case now that the test is passing but it's a very useful lib.

project/Dependencies.scala
kelvin-chappell: Bumped it.

handlers/delivery-problem-credit-processor/src/main/scala/com/gu/deliveryproblemcreditprocessor/DeliveryCreditProcessor.scala
kelvin-chappell: I'll do some more investigation after this PR.  Thanks for the explanation @frj.

PR #555, opened 2020-02-04
delivery-records-api : delivery problem reporting

twrichards:
- Exposed `id` for each delivery record (so it can be used in the new `POST` to link delivery records to the case).
- Added new `POST` handler  (to avoid repetition see [updated README](https://github.com/guardian/support-service-lambdas/blob/2bee7c09/handlers/delivery-records-api/README.md#post-stagedelivery-recordssubscription_name)) which uses [Salesforce Composite Requests](https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/resources_composite_composite.htm) to 
  - create the Salesforce `Case`
  - link the delivery records to the newly created case
  - optionally update any contact phone numbers

  ... all **as a single transaction** (if any part fails, the other parts are rolled back).
- Expose existing credit information for each delivery record and to receive/store the quoted credit information from the client, in the POST. 
IMPORTANTLY: the date quoted to the user is actually what's used by the `delivery-problem-credit-processor` (see https://github.com/guardian/support-service-lambdas/pull/563), this is to avoid any nasty edge cases where the invoice date is quoted (say just before midnight) then when the processor runs _later_ it calculates a different invoice date. The credit amount however is recalculated by the processor (and updates the delivery record) - to avoid misuse of this API.
- Supports marking the case as a 'repeat delivery problem'


.jvmopts
```diff
@@ -1,2 +1,3 @@
 -Xmx2G
--XX:MaxMetaspaceSize=2G
\ No newline at end of file
+-XX:MaxMetaspaceSize=2G
+-Xss128m
```
twrichards: The Circe auto-derivation was causing sporadic `StackOverflowError`, by increasing the stack space this prevented.

Unfortunately the same change had to be made in the configuration of the build step in TeamCity (lost hours to this discrepancy 😢)

handlers/delivery-records-api/src/main/scala/com/gu/delivery_records_api/DeliveryRecordApiRoutes.scala
```diff
@@ -45,6 +47,14 @@ object DeliveryRecordApiRoutes {
         }
     }
 
+    def createDeliveryProblem(
+      subscriptionNumber: String,
+      contact: Contact,
+      detail: CreateDeliveryProblem
+    ): EitherT[F, F[Response[F]], SFApiCompositeResponse] =
+      deliveryRecordsService.createDeliveryProblemForSubscription(subscriptionNumber, contact, detail)
+        .leftMap(error => InternalServerError.apply(error))
```
kelvin-chappell:
```suggestion
        .leftMap(InternalServerError(_))
```

.jvmopts
kelvin-chappell: Has this been reported as an issue in the Circe repo?

twrichards: Yep its number 3️⃣on https://circe.github.io/circe/codecs/known-issues.html

handlers/delivery-records-api/src/main/scala/com/gu/delivery_records_api/DeliveryRecordApiRoutes.scala
twrichards: Done.

PR #560, opened 2020-02-06
delivery-records-api: expose problem types (from SF picklist)

twrichards:
For self-service 'Delivery Problem Reporting' the user needs to select a problem type, so to avoid hardcoding in the client, we now use the `describe` endpoint of the `Case` sobject (SF REST API) to fetch the problem types and expose these as an String array with top level key of `problemTypes` on the main `GET` endpoint.

The mechanism used by SF to represent the interdependency between picklists is very unusual/arcane (base64 encoding of a byte array, where each bits represents the validity of the picklist option based on the selection in the parent/controlling field - see https://salesforce.stackexchange.com/questions/201775/picklists-validfor-attribute). This PR is mainly the implementation and tests to handle this.

This PR also includes:

- exposing the salesforce ID of each delivery record so they can be used for linking to cases (during delivery problem reporting - see https://github.com/guardian/support-service-lambdas/pull/555)
- slight restructuring of the recent changes to SalesforceClient (e.g. naming of type params and removing some slight duplication)


twrichards:
@guardian/supporter-experience after some deliberation, I'm going to **close** this PR rather than merge, since there are business requirements which make the required list options in the self-service client distinct enough from SF picklist (designed for CSRs), such that it's better to maintain the self-service list in the client.

Despite the work I put into it implementing the mechanism to understand the unusual/arcane picklist item's `validFor` field on the `describe` endpoint - it would've added significant complexity so it's no bad thing not to be merging.

PR #565, opened 2020-02-10
Apply discounts to credits

mario-galic:
* Calculate credit by taking into account potential discounts, otherwise return original credit
* It should be able to handle any percentage-based future discounts
* This PR is using [`<|` ]( https://stackoverflow.com/a/57981000/5205022)and `thrush`, which correspond to Scala 2.13 `tap` and `pipe`, but alas, due to a single dependency, we are unable to move to Scala 2.13.
* **FIXME**: We need to make sure credit calculation goes through a single code path. Right now onus is on the user to make sure discounts are applied to credit.


mario-galic:
[GOAL: Holiday suspension and delivery problem reporting credit functionality should take into account discounts](https://trello.com/c/g4ghRV3T/759-goal-holiday-suspension-and-delivery-problem-reporting-credit-functionality-should-take-into-account-discounts)

lib/zuora-core/src/main/scala/com/gu/zuora/subscription/RatePlanChargeData.scala
```diff
@@ -55,17 +56,19 @@ object RatePlanChargeData {
...
       approximateBillingPeriodWeeks <- approximateBillingPeriodWeeksForName(billingPeriodName, ratePlanCharge.specificBillingPeriod)
-      price = -roundUp(ratePlanCharge.price / approximateBillingPeriodWeeks)
-    } yield price
+      roundedCredit = -round2Places(ratePlanCharge.price / approximateBillingPeriodWeeks)
+    } yield {
+      roundedCredit
+    }
   }
 
+  def round2Places(d: Double): Double = BigDecimal(d).setScale(2, RoundingMode.UP).toDouble
```
kelvin-chappell: Perhaps this method should be in another object as it has nothing specifically to do with RatePlanCharges.

lib/zuora-core/src/main/scala/com/gu/zuora/subscription/SubscriptionData.scala
```diff
@@ -32,7 +36,7 @@ trait SubscriptionData {
   def subscriptionAnnualIssueLimit: Int
   def editionDaysOfWeek: List[DayOfWeek]
 }
-object SubscriptionData {
+object SubscriptionData  {
```
kelvin-chappell: Strange spacing

lib/zuora-core/src/main/scala/com/gu/zuora/subscription/SubscriptionData.scala
```diff
@@ -50,33 +54,66 @@ object SubscriptionData {
...
           ratePlanChargeData <- ratePlanChargeDataForDate(nonZeroRatePlanChargeDatas, issueDate)
           billingPeriod <- ratePlanChargeData.billingSchedule.billDatesCoveringDate(issueDate)
-        } yield IssueData(issueDate, billingPeriod, ratePlanChargeData.issueCreditAmount)
+        } yield {
+          discountedIssueData(IssueData(issueDate, billingPeriod, ratePlanChargeData.issueCreditAmount))
+        }
+      }
+
+      // Calculate credit by taking into account potential discounts, otherwise return original credit
+      def discountedIssueData(issueData: IssueData): IssueData = {
```
kelvin-chappell: Perhaps should be called `withAnyDiscount` or something like that to show that there isn't necessarily any discount.

lib/zuora-core/src/main/scala/com/gu/zuora/subscription/SubscriptionData.scala
```diff
@@ -50,33 +54,66 @@ object SubscriptionData {
...
+        val discounts: List[Double] =
+          subscription
+            .ratePlans
+            .iterator
+            .filter(_.productName == "Discounts")
+            .flatMap(_.ratePlanCharges.map(rpc => (rpc.discountPercentage, rpc.effectiveEndDate)))
+            .collect { case (percentage, effectiveEndDate) if percentage.isDefined && effectiveEndDate.isAfter(issueData.issueDate) => percentage }
+            .flatten
+            .map(_ / 100)
+            .toList
```
kelvin-chappell: Very nice!

lib/zuora-core/src/main/scala/com/gu/zuora/subscription/SubscriptionData.scala
```diff
@@ -83,13 +83,18 @@ object SubscriptionData {
...
+
         val discounts: List[Double] =
           subscription
             .ratePlans
             .iterator
             .filter(_.productName == "Discounts")
-            .flatMap(_.ratePlanCharges.map(rpc => (rpc.discountPercentage, rpc.effectiveEndDate)))
-            .collect { case (percentage, effectiveEndDate) if percentage.isDefined && effectiveEndDate.isAfter(issueData.issueDate) => percentage }
+            .flatMap(_.ratePlanCharges.map(rpc => (rpc.discountPercentage, rpc.effectiveStartDate, rpc.effectiveEndDate)))
+            .collect { case (percent, start, end) if percent.isDefined && isActiveDiscount(start, end) => percent }
```
kelvin-chappell: Perhaps there should be tests for within and outside date boundaries to see what happens, as it won't be immediately clear when we return to the code.

PR #568, opened 2020-02-11
holiday-stop-api : historical cut-off on the 'list existing' endpoint

twrichards:
The upcoming backfill (of all Home Delivery Holiday Stops created using the old platform), will produce lots of historical holiday stop requests. 

To avoid too much visual noise for self-service users, we need to add a cut off on the 'list existing' endpoint of the `holiday-stop-api`.

This cut-off, will exclude holiday stop requests where the `Max_Expected_Invoice_Date__c`* is older than 6 months, however for some records that field will be null (from before we calculated this value), so if it's null we exclude based on `End_Date__c` being older than 6 months.

ALSO
- in addition documenting the above in the README, restructured it away from tabular, which makes it easier to read.
- cleaned up tests and removed some old redundant functions.

---

*`Max_Expected_Invoice_Date__c` is a new 'rollup' field created on the `Holiday_Stop_Request__c` object [to facilitate this PR] - it finds the maximum `Expected_Invoice_Date__c` among the child `Holiday_Stop_Requests_Detail__c` records.

### DIFF
--- a/handlers/holiday-stop-api/README.md
+++ b/handlers/holiday-stop-api/README.md
@@ -6,20 +6,40 @@ All endpoints require...
   - `x-identity-id` header which specifies the identityID of the user to request data for _(sent in the `manage-frontend` use-case)_
   - `x-salesforce-contact-id` header which specifies the Salesforce contact ID of the user to request data for _(sent in the CSR UI (in Salesforce) use-case)_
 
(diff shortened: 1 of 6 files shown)


twrichards:
- TESTED in DEV/CODE
- `Max_Expected_Invoice_Date__c` created in DEV, UAT & PROD Salesforce

PR #569, opened 2020-02-11
Digital voucher api: Fetching configuration from property store

frj:
This PR adds the fetching of configuration for the digital voucher api from the aws parameter store. 

For more details see:
* Aws Parameter Store docs - https://docs.aws.amazon.com/systems-manager/latest/userguide/systems-manager-parameter-store.html
* Simple config, the gu library integrating the parameter store with typesafe config - https://github.com/guardian/simple-configuration
* Circe config for binding typesafe config to case classes - https://github.com/circe/circe-config
* Typesafe config - https://github.com/lightbend/config

Note: there are a couple of minor changes to the digital voucher stubbing bundled up with this pr

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -29,6 +29,7 @@ val scalaSettings = Seq(
   },
 
   autoCompilerPlugins := true,
+  resolvers += "Guardian Platform Bintray" at "https://dl.bintray.com/guardian/platforms"
 )
 
 // fixme this whole file needs splitting down appropriately
@@ -373,7 +374,7 @@ lazy val `digital-voucher-api` = all(project in file("handlers/digital-voucher-a
   .dependsOn(`effects-s3`, `config-core`)
   .settings(
     libraryDependencies ++=
-      Seq(http4sLambda, http4sDsl, http4sCirce, http4sServer, circe, sttpAsycHttpClientBackendCats, scalatest)
+      Seq(http4sLambda, http4sDsl, http4sCirce, http4sServer, circe, circeConfig, sttpAsycHttpClientBackendCats, scalatest, simpleConfig)
         ++ logging
   )
   .enablePlugins(RiffRaffArtifact)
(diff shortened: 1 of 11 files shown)


handlers/digital-voucher-api/src/test/resources/digital-voucher-api-secret-dev.conf
```diff
@@ -0,0 +1,2 @@
+imovoApiKey: xxxxx
```
kelvin-chappell: Could this be done by an env overload?  ie. something like `{?IMOVO_API_KEY}` (not sure of exact format) and then feed in from environment variable in dev.  I think you can do this in most config libraries.

frj: This file is just there to get the tests working.

I hadn't put too much thought into running the app locally. 

Possibly the 'simple-config' way of doing things would be to set the APP, STACK and STAGE environment variables, set up your local aws keys and the app will fetch its config from the property store.

You can just reference environment variables in a typesafe config file with ${var} syntax, so maybe having a dev config file on the class path makes sense.

Ill have a think about it.

build.sbt
```diff
@@ -375,7 +376,7 @@ lazy val `digital-voucher-api` = all(project in file("handlers/digital-voucher-a
   .dependsOn(`effects-s3`, `config-core`)
   .settings(
     libraryDependencies ++=
-      Seq(http4sLambda, http4sDsl, http4sCirce, http4sServer, circe, sttpAsycHttpClientBackendCats, scalatest)
+      Seq(http4sLambda, http4sDsl, http4sCirce, http4sServer, circe, circeConfig, sttpAsycHttpClientBackendCats, scalatest, simpleConfig)
```
mario-galic: So just to read config we need further two dependencies?

mario-galic:
When will you migrate the rest of the codebase to this new approach?

PR #576, opened 2020-02-18
Reduce Scala lambda package size by excluding scala-compiler.jar

mario-galic:
scala-compiler.jar is 10 MB and seems to be used only for compile-time macro expansion:
https://stackoverflow.com/questions/60280689/safe-to-exclude-scala-compiler-jar-dependency-from-runtime-classpath


johnduffell:
nice work! very impressive

PR #583, opened 2020-02-21
Trap runtime exceptions thrown by HTTP client

kelvin-chappell:
Such as connection timeouts, etc.

### DIFF
--- a/handlers/digital-voucher-api/src/main/scala/com/gu/digital_voucher_api/imovo/ImovoClient.scala
+++ b/handlers/digital-voucher-api/src/main/scala/com/gu/digital_voucher_api/imovo/ImovoClient.scala
@@ -41,13 +41,6 @@ object ImovoClient extends LazyLogging {
       body: Option[B]
     ): EitherT[F, ImovoClientException, A] = {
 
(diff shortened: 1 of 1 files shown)


handlers/digital-voucher-api/src/main/scala/com/gu/digital_voucher_api/imovo/ImovoClient.scala
```diff
@@ -49,7 +50,9 @@ object ImovoClient extends LazyLogging {
       val request = body.fold(requestWithoutBody)(b => requestWithoutBody.body(b))
 
       for {
-        response <- EitherT.right[ImovoClientException](request.send())
+        response <- EitherT(request.send().map(_.asRight[ImovoClientException]).recover {
+          case e => Left(ImovoClientException(e.toString))
+        })
```
frj: I found something even better:
```suggestion
        response <- request.send().attemptT.leftMap(e => ImovoClientException(e.toString))
```

kelvin-chappell: Nice solution! Thanks.  I don't think I was likely to come up with this in a hurry.

PR #584, opened 2020-02-24
Batch email sender: digital voucher support

frj:
This contains the changes to the batch email sender api to add support for sending digital voucher emails. This includes
- Adding support for additional digital_voucher.card_url field to the email batch item payload, which is forwarded on to the membership_workflow app via sqs.
- Adding mappings for the following object name/email stage combinations
  -  Digital_Voucher__c/create -> SV_VO_NewCard (email containing voucher barcode link to support immediate redemption) 
  - Digital_Voucher__c/replace -> SV_VO_ReplacementCard (email containing new voucher barcode link to support replacement of digital voucher due to loss misuse of card/fulfilment details)

### DIFF
--- a/handlers/batch-email-sender/src/main/scala/com/gu/batchemailsender/api/batchemail/model/EmailBatch.scala
+++ b/handlers/batch-email-sender/src/main/scala/com/gu/batchemailsender/api/batchemail/model/EmailBatch.scala
@@ -28,7 +28,8 @@ object EmailBatch {
       first_name: String,
       email_stage: String,
       modified_by_customer: Option[Boolean],
-      holiday_stop_request: Option[WireHolidayStopRequest]
+      holiday_stop_request: Option[WireHolidayStopRequest],
+      digital_voucher: Option[WireDigitalVoucher]
     )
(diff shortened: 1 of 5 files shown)


handlers/batch-email-sender/src/main/scala/com/gu/batchemailsender/api/batchemail/model/EmailBatch.scala
```diff
@@ -39,9 +40,11 @@ object EmailBatch {
       stopped_credit_summaries: Option[List[WireHolidayStopCreditSummary]]
     )
     case class WireHolidayStopCreditSummary(credit_amount: Double, credit_date: String)
+    case class WireDigitalVoucher(card_url: String)
```
kelvin-chappell:
```suggestion
    case class WireDigitalVoucher(barcode_url: String)
```

frj: well spotted 👍

PR #586, opened 2020-02-27
Make digital voucher API methods closer to rest spec

kelvin-chappell:
This is splitting hairs but seems to be more consistent with how methods are meant to be used, and express lifecycle better.  There's a small window now when we can change these before it becomes more difficult.

### DIFF
--- a/handlers/digital-voucher-api/src/main/scala/com/gu/digital_voucher_api/DigitalVoucherApiRoutes.scala
+++ b/handlers/digital-voucher-api/src/main/scala/com/gu/digital_voucher_api/DigitalVoucherApiRoutes.scala
@@ -102,13 +102,13 @@ object DigitalVoucherApiRoutes {
     }
 
     HttpRoutes.of[F] {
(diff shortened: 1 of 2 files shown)


frj:
Genuinely not wanting to get into a REST pedant competition but... 

The create endpoint is arguable it contains an id so is a member resource so PUT is fine, it may not be idempotent  in it current state but that could be fixed. I think POST makes it more ambiguous as to the meaning of the endpoint.

 The replace endpoint doesnt have an id hence is a collection resource so using a PUT suggest you are going to replace all of the 'replace's with this call, what ever that would mean. POST suggests you would be adding a single 'replace' to the collection, which seems more in line which what the endpoint does.

Likewise the calling DELETE on the /cancel endpoint suggests you are going to delete all the cancelations, using POST suggests you are adding a cancelation to the collection of cancelations.

All of this is going to change as these endpoints are going to become subscription id based, so maybe we should agree on the api when we make those changes?

kelvin-chappell:
I take the point about confusing semantics between collections and members.  Thanks for clarifying.  Will revisit when endpoints change.

PR #590, opened 2020-03-04
AWS CDK code to generate sf-move-subscriptions-api infrastructure cfn.yml file

mkuzdowicz:
## Overview
AWS CDK code to generate sf-move-subscriptions-api infrastructure cfn.yml file
AND
Boilerplate for dummy lambda jar to test if it works in AWS

## What was done
- AWS CDK code to generate cloud-formation for lambda, role and api gateway
- dummy SCALA code to have dummy jar for lambda handler

## More context

this code is implementation in AWS CDK to create initial infrastructure for `API for moving subscriptions in Salesforce` functionality 

This PR introduce creation of infrastructure cloud-formation template by using AWS CDK
it is very easy, described in https://github.com/guardian/support-service-lambdas/blob/dd58491dbb52b2a881317b5cc764c27380e37a37/handlers/sf-move-subscriptions-api/README.md#how-to-generateupdate-cfnyml-by-aws-cdk

## Tests
- [x] in code


mkuzdowicz:
> It looks good but there seems to be a lot of boilerplate to avoid writing a yaml config file, and might be difficult to maintain for developers who expect to edit yaml. Will need to discuss what the benefits of CDK are.

the benefits: 
- it is more modern tool then yml
- one of the biggest may come in the future as this is more modern tool and you can deploy with CDK and avoid yml
- other benefit you can write tests to your template

mkuzdowicz:
> Thanks for the explanation. It all makes sense. 👍

Thanks 👍

PR #591, opened 2020-03-05
DO NOT MERGE: removing redundant digtal voucher api endpoints 

frj:
This PR removes digital voucher api endpoints that have been superseded by the subscription based endpoints.

This PR should not be merged until the work has been done to salesforce to migrate the code to use it. (For status see:https://trello.com/c/RT3382yw/957-imovo-api-updates-update-salesforce-to-use-subscription-based-endpoints)

### DIFF
--- a/handlers/digital-voucher-api/src/main/scala/com/gu/digital_voucher_api/Model.scala
+++ b/handlers/digital-voucher-api/src/main/scala/com/gu/digital_voucher_api/Model.scala
@@ -4,8 +4,6 @@ case class SfSubscriptionId(value: String) extends AnyVal
 
 case class RatePlanName(value: String) extends AnyVal
 
-case class CampaignCode(value: String) extends AnyVal
-
-case class CampaignCodeSet(card: CampaignCode, letter: CampaignCode)
-
 case class SchemeName(value: String) extends AnyVal
+
+case class SubscriptionVouchers(cardCode: String, letterCode: String)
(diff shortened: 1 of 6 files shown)


handlers/digital-voucher-api/src/main/scala/com/gu/digital_voucher_api/Model.scala
```diff
@@ -4,8 +4,6 @@ case class SfSubscriptionId(value: String) extends AnyVal
 
 case class RatePlanName(value: String) extends AnyVal
 
-case class CampaignCode(value: String) extends AnyVal
-
-case class CampaignCodeSet(card: CampaignCode, letter: CampaignCode)
-
 case class SchemeName(value: String) extends AnyVal
```
kelvin-chappell: Is 'scheme name' the term that Imovo use, or should it be `CampaignName`?

frj: scheme name is the imovo nomenclature:
https://github.com/guardian/digital-voucher-api-documentation/blob/master/api/swagger/swagger.yaml#L275

It has a slightly different meaning than campaign, ie the scheme has multiple campaign codes.

kelvin-chappell: Oh yes, it's at least 2 to 1.

PR #594, opened 2020-03-06
refactoring digital voucher http4s controller to avoid nested Fs

frj:
The http4s helper methods unnecessarily wrap responses in F eg:
```scala
val foo: IO[Response] = Ok("foo")
```

This seems only to benefit the simple case ie where the controller doesnt
perform any side effects.

eg:

```scala
HttpRoutes.of[IO] {
   case GET -> Root / "foo"=> Ok("Blah")  
}
```
If the controller does create an IO eg in making an api call to another service
the IO[Response] created by the helper methods needs to be flatMapped into
the response from the api.

eg:
```scala
def effect() :  IO[String]

HttpRoutes.of[IO] {
  case GET -> Root / "foo"=> 
    for {
      stuff <- effect()
      result <- OK(stuff) 
    } yield result
} 
```

This app makes use of Either/EitherT which makes it a bit more awkward to flatten the IO[Response] type at the point where they were created. 

I think this is why I didnt do it to start with. If you look at the types in the Routes codes they tend to be of the form:

```scala
EitherT[|IO, IO[Response], IO[Response]]
```
 
Ideally they would be:
```scala
EitherT[|IO, Response, Response]
```

In reality its not a big deal as it only requires an extra 'flatten' at the end of each route method.

However I had a go at doing it the 'right' way, and I ended up using some fairly obscure EitherT/cats methods to get it to work.

What does everyone think, is this better or should we leave the code as is.

### DIFF
--- a/handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/HolidayStopCreditProcessor.scala
+++ b/handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/HolidayStopCreditProcessor.scala
@@ -10,7 +10,7 @@ import com.gu.holiday_stops.Config
 import com.gu.salesforce.holiday_stops.SalesforceHolidayStopRequestsDetail.HolidayStopRequestsDetail
 import com.gu.util.config.Stage
 import com.gu.zuora.Zuora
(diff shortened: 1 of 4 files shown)


kelvin-chappell (commented):
I'm inclined to leave as is.  I think this makes the code more difficult to understand.

mario-galic:
Note sure where the code went after force pushing...

PR #596, opened 2020-03-10
Making zuora config generic in [zuora-core] module

mkuzdowicz:
Making zuora config generic in `zuora-core` module

 `zuora-core` module is used in
- holiday-stop-processor
- delivery-problem-credit-processor

and will be used in new

- sf-move-subscriptions-api

## what was the problem
`Zuora.accessTokenGetResponse` function which should use generic config was hardcoded  to holidayStopProcessor config

## What changed
this PR introduces
-  `ZuoraRestOauthConfig` which is generic
- backward compatibility by introducing:
 ```
trait ZuoraConfig
case class ZuoraRestOauthConfig extend ZuoraConfig
case class HolidayStopProcessorZuoraConfig extends ZuoraConfig
```  
Model
and introducing `accessTokenGetResponseV2` and `accessTokenGetResponse` functions

 backward compatibility was introduced to keep this PR relatively small

## Tests
- [x] in CODE


mkuzdowicz:
> That's much better! I didn't realize Zuora-core still had a dependency on holiday stops.

Thank you 👍

PR #603, opened 2020-03-13
[sf-move-subscriptions-api] adding riffraff build config and CI/CD for generating cfn templates from AWS CDK

mkuzdowicz:
## overview
[sf-move-subscriptions-api] adding riffraff build config and enabling generation of `cdk-cfn.yaml` file on teamcity build time

## what was done
- adding riffraff build config
- moving cdk directory to root to be able to build cfn templates from cdk on build time in CI/CD
- creating `generate_cfn_templates.sh` script which will build cfn templates from cdk on build time in CI/CD
- README instructions

## tests
- [x] CODE


handlers/sf-move-subscriptions-api/cfn.yaml
```diff
@@ -0,0 +1,538 @@
+Parameters:
+  stage:
```
mkuzdowicz: please ignore the cfn.yml file as it is autogenerated by cdk in handlers/sf-move-subscriptions-api/cdk/lib/sf-move-subscriptions-stack.ts file

twrichards: please can we make it generate at build time then? so it's not in version control (in addition to the CDK code) but is still available for riff-raff purposes

mkuzdowicz: i don't think that riffraff is supporting that at the moment

mkuzdowicz: aa you mean in team city

mkuzdowicz: that would be use-full

mkuzdowicz: more less done

PR #605, opened 2020-03-16
Home Delivery Problem Reporting (case tweaks)

twrichards:
https://trello.com/c/1rXMTCPN/1007-hd-home-delivery-problem-reporting-case-assignment
image

The `Case Owner` changes are done in Salesforce - see https://github.com/guardian/salesforce/pull/184.


handlers/delivery-records-api/src/test/scala/com/gu/delivery_records_api/SFApiCompositeCreateDeliveryProblemTest.scala
```diff
@@ -59,6 +59,7 @@ class SFApiCompositeCreateDeliveryProblemTest extends FlatSpec with Matchers {
          |        },
          |        "Origin" : "Self Service",
          |        "Status" : "New",
+         |        "Priority" : null,
          |        "Subject" : "[Self Service] Delivery Problem : No delivery (Guardian Weekly - A-S123456)",
```
mkuzdowicz: `Priority` should that be None or null ?
why we set that to null fo we are using Option type ?

handlers/delivery-records-api/src/main/scala/com/gu/delivery_records_api/DeliveryRecordsService.scala
```diff
@@ -57,7 +57,19 @@ case class CreateDeliveryProblem(
...
+  val isNotAutoCredit: Boolean = deliveryRecords.exists(_.invoiceDate.isEmpty)
+
+  val status: String = if (
+    isHomeDelivery ||
+    repeatDeliveryProblem.contains(true) ||
+    isNotAutoCredit
+  ) "New" else "Closed"
+
+  val priority: Option[String] = Some("High").filter(_ => isNotAutoCredit && isHomeDelivery)
+}
```
mkuzdowicz: so `High` and `None` are the only options ?

twrichards: see my other reply about this for more detail, but `High` is the only value we want to set it to explicitly, `None` is essentially 'not setting it' (as it was before)

handlers/delivery-records-api/src/test/scala/com/gu/delivery_records_api/SFApiCompositeCreateDeliveryProblemTest.scala
mkuzdowicz: ok

handlers/delivery-records-api/src/main/scala/com/gu/delivery_records_api/DeliveryRecordsService.scala
mkuzdowicz: ok

PR #606, opened 2020-03-16
[sf-move-subscriptions-api] adding dry-run path to make smoke testing in PROD easy

mkuzdowicz:
## Overview

[sf-move-subscriptions-api] adding dry-run path to make smoke testing in PROD easy

## What was done

- adding `subscription/move/dry-run` path to be able to test if connections/calls to all downstream dependencies works, but the api call will not make any changes

## Why

- that enable smoke testing
- that makes testing in PROD much easier

## Tests
- [x] CODE

### DIFF
--- a/handlers/sf-move-subscriptions-api/src/main/scala/com/gu/sf/move/subscriptions/api/SFMoveSubscriptionsService.scala
+++ b/handlers/sf-move-subscriptions-api/src/main/scala/com/gu/sf/move/subscriptions/api/SFMoveSubscriptionsService.scala
@@ -34,7 +34,18 @@ class SFMoveSubscriptionsService[F[_]: Monad](
     )
   )
 
-  def moveSubscription(moveSubscriptionData: MoveSubscriptionReqBody, dryRun: Boolean): EitherT[F, MoveSubscriptionServiceError, MoveSubscriptionServiceSuccess] = {
+  def moveSubscription(moveSubscriptionData: MoveSubscriptionReqBody): EitherT[F, MoveSubscriptionServiceError, MoveSubscriptionServiceSuccess] = {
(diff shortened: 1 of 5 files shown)


handlers/sf-move-subscriptions-api/src/main/scala/com/gu/sf/move/subscriptions/api/SFMoveSubscriptionsService.scala
```diff
@@ -34,6 +35,17 @@ class SFMoveSubscriptionsService[F[_]: Monad](
...
+    moveSubscriptionInternal(moveSubscriptionData, updateAccountByMovingSubscriptionRun)
+  }
+
+  def moveSubscriptionDryRun(moveSubscriptionData: MoveSubscriptionReqBody): EitherT[F, MoveSubscriptionServiceError, MoveSubscriptionServiceSuccess] = {
+    moveSubscriptionInternal(moveSubscriptionData, updateAccountByMovingSubscriptionDryRun)
+  }
+
+  private def moveSubscriptionInternal(
+    moveSubscriptionData: MoveSubscriptionReqBody,
+    updateAccountByMovingSubscription: (AccessToken, SttpBackend[Id, Nothing]) => (Subscription, ZuoraAccountMoveSubscriptionCommand) => ZuoraApiResponse[MoveSubscriptionAtZuoraAccountResponse]
```
frj: Just my opinion but Im not a huge fan of passing functions around like this, it makes the code quite difficult to follow, ie you can just click on the function definition to find out what happens next. You could just have dryRun flag and an if statement. If its more complicated that that defining a trait allows you to find the different implementations more easily.

In this case its simple enough to work out whats going on, but there are examples eg hoilday-stop-processor where its got out of hand and its difficult to follow the code.

PR #612, opened 2020-03-17
Zuora SAR processing lambda

johnduffell:
**Sorry John not sure how to switch the author back over to myself!

This PR adds 2 new lambdas to handle Subject Access Requests to Zuora. These two new lambdas are the ZuoraSarHandler and the PerformZuoraSarHandler.

1. PerformZuoraSarHandler
This lambda deals with making calls to Zuora to retrieve all information associated with a users email address (including their account summary, account object and invoices). As the results are retrieved, it writes to a /pending path in s3. Once the SAR is finished, it copies all results in /pending to /completed. If the calls to Zuora fail for some reason, it writes a file to a /failed path in s3.

2. ZuoraSarHandler
This lambda is triggered by [Baton](https://github.com/guardian/baton), our new GDPR tool used by Data Privacy. The lambda can be called to perform two process, either initiating the PerformZuoraSarHandler or checking the status of the work performed by the PerformZuoraSarHandler. It checks status by first checking if a /failed path exists for a given initiation id (sent by Baton), if not then it checks for a /completed path. It reports the status (Completed, Failed or Pending) back to Baton based on these paths.

I've also added a new effects-lambda library, to handle AWS lambda invocations, and extending the functionality of effects-s3 to handle copying files.


handlers/zuora-sar/src/local/com.gu.zuora.sar/ZuoraSarLocalRun.scala
```diff
@@ -0,0 +1,44 @@
+package com.gu.zuora.sar
```
johnduffell: the package name looks ok, but for some reason the directory name has dots instead of slashes
image

handlers/zuora-sar/src/main/scala/com.gu.zuora.sar/CirceCodecs.scala
```diff
@@ -0,0 +1,69 @@
...
+  import io.circe.generic.auto._
+
+  implicit val sarRequestDecoder: Decoder[SarRequest] =
+    Decoder.instance[SarRequest] { cursor =>
+      cursor.downField("dataProvider").as[String].flatMap {
+        case "zuora" => cursor.as[String]
+        case unrecognisedProvider: String =>
+          throw new ParsingFailure(
+            s"invalid data provider : $unrecognisedProvider",
+            null
```
johnduffell: is this a standard pattern for circe? for scala normally would return some kind of Decoder.failed or similar rather than exception.

handlers/zuora-sar/src/main/scala/com.gu.zuora.sar/ZuoraHelper.scala
```diff
@@ -0,0 +1,89 @@
...
+case class InvoiceSuccess(fullInvoice: JsValue) extends ZuoraSarSuccess
+
+trait ZuoraSarError
+case class ZuoraClientError(message: String) extends ZuoraSarError
+case class JsonDeserialisationError(message: String) extends ZuoraSarError
+case class S3WriteError(message: String) extends ZuoraSarError
+
+case class ZuoraHelperError(message: String)
+
+case class ZuoraHelper(zuoraClient: Requests, zuoraDownloadClient: Requests, zuoraQuerier: ZuoraQuerier) extends LazyLogging {
```
johnduffell: I would avoid calling it "helper" or similar, as that can be a magnet for it growing over time.  If you want to group them together, something like ZuoraSarRequest would be more tightly scoped.

laurenemms: You're right. How about ZuoraSarService?

johnduffell: sounds good!

handlers/zuora-sar/src/main/scala/com.gu.zuora.sar/ZuoraHelper.scala
```diff
@@ -0,0 +1,89 @@
...
+      contactQuery <- zoql"SELECT AccountId FROM Contact where WorkEmail=${emailAddress}"
+      queryResult <- zuoraQuerier[ZuoraContact](contactQuery).map(_.records)
+    } yield queryResult
+  }
+
+  private def accountSummary(accountId: String): Either[ClientFailure, JsValue] =
+    zuoraClient.get[JsValue](s"accounts/$accountId/summary").toDisjunction
+
+  private def accountObj(accountId: String): Either[ClientFailure, JsValue] = {
+    zuoraClient.get[JsValue](s"object/account/$accountId", WithoutCheck).toDisjunction
```
johnduffell: are you wanting to get errors back as a JsValue, or did the check fail when it shouldn't have?
If you want the errors as json, then that's fine, if not, there could be an issue with the reader that needs fixing https://github.com/guardian/support-service-lambdas/blob/master/lib/zuora/src/main/scala/com/gu/util/zuora/ZuoraObjects.scala#L20

johnduffell:
looks good to me! as requested I have mostly looked at the overall structure rather than the scala/file itself, and I think you've got the right level of reuse and self containment, so we are not ending up pulling anything anywhere it shouldn't go.
 I haven't scrutinised the code in as much detail but I still had a quick look over and it looks good.  This is a relatively simple lambda (in a good way!) so I think you've written it at a good level of clarity.  Of course let me know if you have any particular concerns.
Definitely worth getting a review from someone who knows about the subject area too, but from my point of view it's all good.

handlers/zuora-sar/src/main/scala/com.gu.zuora.sar/ZuoraHelper.scala
laurenemms: Ah I see. I've just looked into this and it seems that the `jsonIsSuccessful` check that the ZuoraRestRequestMaker uses to validate the result looks for a `success` or `Success` field that isn't actually returned by this endpoint... https://github.com/guardian/support-service-lambdas/blob/master/lib/zuora/src/main/scala/com/gu/util/zuora/ZuoraObjects.scala#L23

I can validate the response with some custom check in this method rather than hacking the existing Reader to also look for another field to consider it a successful response... What do you think?

handlers/zuora-sar/src/local/com.gu.zuora.sar/ZuoraSarLocalRun.scala
laurenemms: My mistake - fixed now.

handlers/zuora-sar/src/main/scala/com.gu.zuora.sar/ZuoraHelper.scala
johnduffell: yeah, this is the joy of zuora, nothing is consistent
However, looking at the api docs https://www.zuora.com/developer/api-reference/#operation/Object_GETAccount
it looks like the http status code is enough (some of the other endpoints return 200 with "success" or "Success" = false, hence the need for the check)
So don't worry about it  unless you feel the need for additional checks.

laurenemms: I've added a little [check](https://github.com/guardian/support-service-lambdas/pull/612/files#diff-06c58eec7f1d02d9e032a4cfa5098196R53) just in case. Thanks for pointing this out

laurenemms:
> looks good to me! as requested I have mostly looked at the overall structure rather than the scala/file itself, and I think you've got the right level of reuse and self containment, so we are not ending up pulling anything anywhere it shouldn't go.
> I haven't scrutinised the code in as much detail but I still had a quick look over and it looks good. This is a relatively simple lambda (in a good way!) so I think you've written it at a good level of clarity. Of course let me know if you have any particular concerns.
> Definitely worth getting a review from someone who knows about the subject area too, but from my point of view it's all good.

Thanks so much for looking at this. I'll send it to my team for a review as well. There was one last question I had. You'll see in the cloudformation that I have some security groups commented out. In other lambdas, we've included the vpc ids and subnets in the CF but they've all been private repos. Wanted to check if you wanted me to use environment variables or something to pass this into the CF template?

handlers/zuora-sar/src/main/scala/com.gu.zuora.sar/ZuoraHelper.scala
tylr-mk: Success = false is just about my favourite thing

handlers/zuora-sar/src/main/scala/com.gu.zuora.sar/S3Helper.scala
```diff
@@ -105,6 +106,7 @@ object S3Helper extends S3Service with LazyLogging {
       zuoraError: ZuoraSarError,
       config: ZuoraSarConfig): Either[S3Error, S3WriteSuccess] = {
     val resultsPath = s"${config.resultsPath}/$initiationId/failed/$randomUUID"
+    logger.info("Uploading file to failed path in S3.")
```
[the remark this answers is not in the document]
laurenemms: The `UploadToS3.putStringWithAcl` includes logging

laurenemms:
> I've made a sprinkling of comments which are largely none functional, and in some parts resolved by pursuant commits. None of which should be in the way of this PR being approved.
> 
> As summary for record, I'm only really able to observe and check if there's craziness in the code while my Scala and GU baton architecture knowledge improves. And genuinely this looks like great work on a big thing.
> 
> Notable things which I'd encourage over time are:
> 
> * Add tests into the commits with the functional changes.
> * Try not to commit commented code.
> * And the biggie: Break PRs up
> 
> This huge thing is great, but piecemeal**ing** the process or the solution would help with sharing the workload throughout the whole process -- though I know I'm preaching a bit to the choir and I'm fully aware that isn't a reality or practicality atm.
> 
> What I'm saying at length is. Considering everything, this looks solid. 🚀

These are all really good points and I've made a note of the things to work on. Thank you!

johnduffell:
sorry I missed this, regarding the vpc IDs I'm not sure really, looks like we put a default in support frontend https://github.com/guardian/support-frontend/blob/d11c734f34f100515206d7adad87fb33e30f8ace/support-frontend/cloud-formation/cfn.yaml#L7
It might be a question for the Security channel on gchat!

PR #613, opened 2020-03-18
Digital voucher cancellation processor get subs to cancel from salesforce

frj:
This PR contains the changed to the digital voucher cancellation processor to fetch the subscriptions that should be cancelled from salesforce.

The subscriptions are just logged, future PRs will add the calls to imovo to stop redemption and to 'tag' the subscriptions as having been processed so they are not re-processed.

A basic test is included that just stubs out the calls to salesforce but does not actually test anything.

Details:
- The sttp backend uses the HttpUrlConnection backend in an 'IO' wrapper from sttp-cats library which avoid the async-http/netty dependencies. For details see: https://github.com/guardian/support-service-lambdas/compare/dv-processor-get-subs-to-cancel?expand=1#diff-61664fab2e23cfef1c374b2aba226d24R40


handlers/digital-voucher-cancellation-processor/src/main/scala/com/gu/digital_voucher_cancellation_processor/DigitalVoucherCancellationProcessor.scala
```diff
@@ -0,0 +1,40 @@
...
+import com.gu.salesforce.SalesforceQueryConstants.formatDate
+import com.typesafe.scalalogging.LazyLogging
+import io.circe.generic.auto._
+
+case class DigitalVoucherCancellationProcessorError(message: String)
+
+object DigitalVoucherCancellationProcessor extends LazyLogging {
+
+  case class DigitalVoucherQueryRecord(Id: String, SF_Subscription__r: SubscriptionQueryRecord)
+  case class SubscriptionQueryRecord(Id: String)
```
mkuzdowicz: what is the better naming convention `SubscriptionQueryRecord` or `SubscriptionQueryResult` ?

handlers/digital-voucher-cancellation-processor/src/main/scala/com/gu/digital_voucher_cancellation_processor/DigitalVoucherCancellationProcessor.scala
```diff
@@ -0,0 +1,40 @@
+package com.gu.digital_voucher_cancellation_processor
```
mkuzdowicz: maybe `DigitalVoucherCancellationProcessor` -> `DigitalVoucherCancellationProcessorService` ?

we have similar convention
Handler -> App -> Service

handlers/digital-voucher-cancellation-processor/src/main/scala/com/gu/digital_voucher_cancellation_processor/Handler.scala
```diff
@@ -17,4 +18,8 @@ object Handler extends LazyLogging {
       .valueOr(error => throw new RuntimeException(error.toString))
     logger.info(s"Processor ran successfully")
   }
+
+  def main(args: Array[String]): Unit = {
+    handle(new StringInputStream(""), new ByteArrayOutputStream())
+  }
```
mkuzdowicz: why you are using `com.amazonaws.util.StringInputStream` over `ByteArrayInputStream` ?

PR #616, opened 2020-03-19
Moving root module `support-service-lambdas` to [zuora-callout-apis] module

mkuzdowicz:
## Overview

- this PR move code from `./src` directory root directory to `/handlers/zuora-callout-apis` dir
- and rearrange main `build.sbt` a bit because of that

- it does not have any business logic changes

## Reason

any work for example on `zuora-auto-cancel` api is a bit difficult now especially in terms of testing that in the cloud because you will need to compile the whole project to regenerate the artefact

- that change enable quick compilation of that single module by running `sbt zuora-callout-apis/assembly`

## Tests
- [x] CODE

testing was deploying by riff raff and checking if the build will still work


mario-galic:
What is `zuora-callout-apis` suppose to mean? Everything is calling out to zuora. Can you give it a specific meaningful name?

handlers/zuora-callout-apis/README.md
```diff
@@ -1,6 +1,8 @@
-This refers to the monolithic code in the main src folders in the root of the project.  These will be moved to this subdir in future.
+This refers to the monolithic code in the `handlers/zuora-callout-apis`
 
-## root
+These will be moved to this subdir in future.
```
johnduffell: probably don't need this line! As they are already moved now.

johnduffell: although you could write that there are actually 3 lambdas here which are mostly separate so should be in 3 separate handlers/jars, rather than all in the one.  Or at least the zuora and stripe ones separated.

johnduffell:
only just saw I was up for review on this one, but it is good work well overdue so thanks for doing it.
Agree with mario on the naming, although that's a smaller issue than it being in `root`, it could be better named and/or further split down in future.

PR #627, opened 2020-03-25
enable digipack access for paper customers (voucher & HD)

twrichards:
...because Coronavirus

Much like https://github.com/guardian/members-data-api/pull/431 (which was for identity users), this does the same for those who use `Surname`/`PostCode` AND `Subscription ID` to authenticate.

image

### DIFF
--- a/handlers/digital-subscription-expiry/src/main/scala/com/gu/digitalSubscriptionExpiry/zuora/GetSubscriptionExpiry.scala
+++ b/handlers/digital-subscription-expiry/src/main/scala/com/gu/digitalSubscriptionExpiry/zuora/GetSubscriptionExpiry.scala
@@ -46,8 +46,8 @@ object GetSubscriptionExpiry {
     def isActiveDigipack(charge: RatePlanCharge) = isDigipackName(charge.name) && chargeSpansDate(charge)
     val hasActiveDigipackCharges = subscription.ratePlans.flatMap(_.ratePlanCharges).exists(isActiveDigipack)
 
-    // FIXME should only exist during Coronavirus measures (where digipack access is expanded to cover paper customers)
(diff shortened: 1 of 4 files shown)


handlers/digital-subscription-expiry/src/main/scala/com/gu/digitalSubscriptionExpiry/zuora/GetSubscriptionExpiry.scala
```diff
@@ -39,11 +39,18 @@ object GetSubscriptionExpiry {
...
+      !charge.effectiveEndDate.isBefore(dateToCheck) &&
+        !charge.effectiveStartDate.isAfter(dateToCheck)
 
-    val hasActiveDigipackCharges = subscription.ratePlans.map(_.ratePlanCharges).flatten.exists(isActiveDigipack)
+    def isActiveDigipack(charge: RatePlanCharge) = isDigipackName(charge.name) && chargeSpansDate(charge)
+    val hasActiveDigipackCharges = subscription.ratePlans.flatMap(_.ratePlanCharges).exists(isActiveDigipack)
 
-    if (hasActiveDigipackCharges) {
+    // FIXME should only exist during Coronavirus measures (where digipack access is expanded to cover paper customers)
+    @deprecated def isNewspaper(ratePlan: RatePlan) = ratePlan.productName.toUpperCase.startsWith("NEWSPAPER")
```
mario-galic: You can put the message in the annotation `@deprecated("should only exist during Coronavirus measures")`

twrichards: good suggestion, thanks, done https://github.com/guardian/support-service-lambdas/pull/627/commits/83a01425ba02a6f237b96bd14127a2e31f9d363f

PR #630, opened 2020-03-26
Replace https4 lambda

frj:
The [scala-server-lambda](https://github.com/howardjohn/scala-server-lambda) project which gives us http4s-lambda has lapsed into inactivity.

Specifically there has been no release of my PR that add the cross build for 2.13.

There are a few options but probably the most straightforward one is to just implement the binding ourselves. 

This is relatively simple ie parse the aws api gateway request, convert it to a http4s and then the reverse with the response from the http4s response.

Note: this PR initially migrates the digital voucher api, the other projects dependent on scala-server-lamda will be migrated in future PRs.


handlers/digital-voucher-api/src/main/scala/com/gu/digital_voucher_api/Handler.scala
```diff
@@ -1,8 +1,8 @@
 package com.gu.digital_voucher_api
 
-import io.github.howardjohn.lambda.http4s.Http4sLambdaHandler
```
kelvin-chappell, line 3: Where was this being brought in from?  Is there another dependency somewhere to remove?

frj: Ive removed the dependency from the digital-voucher-api project, its still being used in the delivery-record-api, ill do that in another PR, just to er on the side of caution.

kelvin-chappell: 👍

PR #631, opened 2020-03-27
Remove intermediary email message model

mario-galic:
All this lambda does is takes salesforce payload and converts it to braze SQS message for membership-workflow to process.

The main change here is to remove intermediary model:

**Before**

```
WireEmailBatchItemPayload -> EmailBatchItemPayload -> EmailPayloadSubscriberAttributes
```

**After**

```
SalesforceBatchItem -> BrazeSqsMessage
```

Does not fix existing issue with silent PROD errors.


handlers/batch-email-sender/src/main/scala/com/gu/batchemailsender/api/batchemail/SalesforceToBrazeTransformations.scala
```diff
@@ -0,0 +1,36 @@
...
+      case sfStreetPattern(line1, _) if line1.nonEmpty => Some(line1)
+      case _ => None
+    }
+
+  def sfStreetToLine2(in: String): Option[String] =
+    in match {
+      case sfStreetPattern(_, line2) if line2.nonEmpty => Some(line2)
+      case _ => None
+    }
+}
```
kelvin-chappell: I'm sure we're doing this same thing in lots of different places.  Probably in slightly different ways.  Could it be shared?  But not for this PR.

mario-galic:
Tested in PROD. At least `SV_HolidayStopConfirmation` sent by Braze PROD.

PR #639, opened 2020-04-09
Process fewer holiday credits more often

kelvin-chappell:
This is to avoid the holiday-stop processing lambda timing out.

Following this change, it will have the same capacity but should be far less likely to time out.
If Zuora slows down or we add more products we'll have to look at it again.

### DIFF
--- a/handlers/holiday-stop-processor/cfn.yaml
+++ b/handlers/holiday-stop-processor/cfn.yaml
@@ -121,7 +121,7 @@ Resources:
     Properties:
       Description: Trigger processing of holiday stops every hour (to give 24 attempts a day)
       Name: !FindInMap [StageMap, !Ref Stage, ScheduleName]
-      ScheduleExpression: "cron(0 * ? * * *)"
+      ScheduleExpression: "cron(0/20 * ? * * *)"
       State: ENABLED
       Targets:
         - Arn: !GetAtt HolidayStopProcessor.Arn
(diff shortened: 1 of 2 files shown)


handlers/holiday-stop-processor/cfn.yaml
```diff
@@ -121,7 +121,7 @@ Resources:
     Properties:
       Description: Trigger processing of holiday stops every hour (to give 24 attempts a day)
```
twrichards, line 122: Might want to update this line

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/Salesforce.scala
```diff
@@ -21,7 +21,7 @@ object Salesforce extends LazyLogging {
    * Beware! This isn't a scalable solution.  If all the products between them
    * have > 7200 credit requests to be processed in a 24 hour window some will be missed.
    */
-  private val batchSize = 300
+  private val batchSize = 100
```
twrichards, line 24: Could this not be say 150? Still shouldn't time out in the 15 mins, but we end up doing 450 an hour instead of 300

kelvin-chappell: With 3 products that would give possible max of 450 which would still be pushing the 15 min window.  Could try 120?

handlers/holiday-stop-processor/cfn.yaml
kelvin-chappell: Good point

PR #644, opened 2020-04-20
new product api: adding guardian weekly to catalog endpoint

frj:
This PR adds Guardian Weekly Domestic and ROW product meta data to the new product api catalog endpoint.

Details:
- Start date rules for GW have a cutoff day of friday, 14 day delay before fulfilment can start and a window of 28 days. This gives the same start date selection behaviour as the subscriptions frontend.
- The GW products have an additional field 'enabledForDeliveryCountries' which will enable salesforce to select the Domestic or ROW product based on the customers delivery address. 
- Additional cloudformation config to support deploying the api in DEV, also pointing pre-prod envs at the output of the CODE zuora catalog export in s3

### DIFF
--- a/handlers/new-product-api/cfn.yaml
+++ b/handlers/new-product-api/cfn.yaml
@@ -20,7 +20,7 @@ Mappings:
             ApiName: new-product-api-DEV
             DomainName: new-product-api-dev.membership.guardianapis.com
             ApiGatewayTargetDomainName: d-ecyddyj7nk.execute-api.eu-west-1.amazonaws.com
-            ZuoraCatalogLocation: arn:aws:s3:::gu-zuora-catalog/DEV/Zuora-UAT/catalog.json
+            ZuoraCatalogLocation: arn:aws:s3:::gu-zuora-catalog/CODE/Zuora-DEV/catalog.json
             ContributionThanksEmailQueue: contributions-thanks-dev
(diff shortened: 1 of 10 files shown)


handlers/new-product-api/cfn.yaml
```diff
@@ -60,7 +68,7 @@ Resources:
                           - Effect: Allow
                             Action: s3:GetObject
                             Resource:
-                            - !Sub arn:aws:s3:::gu-reader-revenue-private/membership/support-service-lambdas/${Stage}/zuoraRest-${Stage}.*.json
+                            - !Sub arn:aws:s3:::gu-reader-revenue-private/membership/support-service-lambdas/${Stage}/zuoraRest-${Stage}*.json
```
kelvin-chappell, line 71: If this is prod only, does it need stage variables?

frj: When you say PROD only?

The api is in all envs now, this change was just to get dev working, ie the file is zuoraRest-DEV.json in dev but zuoraRest-CODE.v1.json in code. The regex wasnt working for the dev version.

kelvin-chappell: That's what comes of looking at the partial file! I thought this was part of the Prod stage map.  Ignore.

PR #645, opened 2020-04-20
Add endpoint to create holiday stop with a bulk suspension reason

kelvin-chappell:
This adds a new endpoint that populates the `BulkSuspensionReason__c` picklist field in a holiday-stop request.


handlers/holiday-stop-api/cfn.yaml
```diff
@@ -329,18 +354,20 @@ Resources:
...
         - HolidayStopApiCreateMethod
+        - BulkHolidayStopApiCreateMethod
         - HolidayStopApiDeleteMethod
         - HolidayStopApiEditMethod
         - HolidayStopApiCancelPostMethod
         - HolidayStopApiCancelGetMethod
 
-    HolidayStopApiDeployment:
+    # Bizarrely you have to update the deployment to include new endpoints in each stage
+    HolidayStopApiDeployment2:
```
frj, line 370: I think you can go into the aws console and 'manually' deploy using Resources -> Actions -> Deploy Api from the api page.

kelvin-chappell: Thanks for the tip.  I find the hierarchy of the API gateway config very strange.

PR #656, opened 2020-05-14
[holiday-stop-api] fix MutabilityFlags

twrichards:
Whether a customer can amend their holiday stop (either self-serve or via CSR) is determined by the `mutabilityFlags` returned from the API...
- `isFullyMutable` - for when both the start and end date can be changed
- `isEndDateEditable` - for when only the end date can be changed (UIs ensure this end date can't be moved before the `firstAvailableDate`

However there was a bug in the `isEndDateEditable` calculation, because it didn't consider _actioned_ publications (i.e. already locked into Zuora). So in the case of backfilled holiday stops where they're all already _actioned_ (because they were extracted from Zuora - as the old system wrote straight to Zuora, unlike our new just-in-time model). 

This bug meant customers could change the end date on future holiday stops that they had originally booked in the old system. Performing these amends  only updated the data in SF rather than Zuora (which is where it matters for fulfilment/credit purposes). 

This affected 15 holidays stops, which are being addressed separately.

### PLUS
Added additional server-side validation on the endpoints which actually perform the mutation...

- block on withdrawing holiday stops which have actioned publications (this is actually already blocked by SF trigger, but this doesn't hurt)
- on the amend call, fails before deleting any holiday stops publications which are actioned

### DIFF
--- a/handlers/holiday-stop-api/src/main/scala/com/gu/holiday_stops/Handler.scala
+++ b/handlers/holiday-stop-api/src/main/scala/com/gu/holiday_stops/Handler.scala
@@ -465,7 +465,8 @@ object Handler extends Logging {
       contact <- extractContactFromHeaders(req.headers)
       pathParams <- req.pathParamsAsCaseClass[SpecificHolidayStopRequestPathParams]()
       existingForUser <- lookupOp(contact, None, None).toDisjunction.toApiGatewayOp(s"lookup Holiday Stop Requests for contact $contact")
-      _ = existingForUser.exists(_.Id == pathParams.holidayStopRequestId).toApiGatewayContinueProcessing(ApiGatewayResponse.forbidden("not your holiday stop"))
+      hsr <- existingForUser.find(_.Id == pathParams.holidayStopRequestId).toApiGatewayContinueProcessing(ApiGatewayResponse.forbidden("not your holiday stop"))
+      _ = hsr.Holiday_Stop_Request_Detail__r.exists(_.records.exists(_.Is_Actioned__c)).toApiGatewayContinueProcessing(ApiGatewayResponse.forbidden("can't withdraw holiday stop with any actioned items"))
       _ <- withdrawOp(pathParams.holidayStopRequestId).toDisjunction.toApiGatewayOp(
         exposeSfErrorMessageIn500ApiResponse(s"withdraw Holiday Stop Request for subscription ${pathParams.subscriptionName.value} of contact $contact")
       )
--- a/handlers/holiday-stop-api/src/main/scala/com/gu/holiday_stops/WireHolidayStopRequest.scala
+++ b/handlers/holiday-stop-api/src/main/scala/com/gu/holiday_stops/WireHolidayStopRequest.scala
@@ -11,49 +11,48 @@ import play.api.libs.json._
 
 object WireHolidayStopRequest {
 
-  def apply(firstAvailableDate: LocalDate)(sfHolidayStopRequest: HolidayStopRequest): HolidayStopRequestFull = HolidayStopRequestFull(
-    id = sfHolidayStopRequest.Id.value,
-    startDate = sfHolidayStopRequest.Start_Date__c.value,
-    endDate = sfHolidayStopRequest.End_Date__c.value,
-    subscriptionName = sfHolidayStopRequest.Subscription_Name__c,
-    publicationsImpacted = sfHolidayStopRequest
-      .Holiday_Stop_Request_Detail__r
-      .map(_.records.map(toHolidayStopRequestDetail)).getOrElse(List()),
-    withdrawnTime = sfHolidayStopRequest.Withdrawn_Time__c.map(_.value),
-    bulkSuspensionReason = sfHolidayStopRequest.Bulk_Suspension_Reason__c,
-    mutabilityFlags = calculateMutabilityFlags(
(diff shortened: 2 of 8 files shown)


handlers/holiday-stop-api/src/main/scala/com/gu/holiday_stops/WireHolidayStopRequest.scala
```diff
@@ -11,49 +11,48 @@ import play.api.libs.json._
...
-    }
-    val firstPublicationDateIsGreaterOrEqualToFirstAvailableDate =
-      firstPublicationDate.isEqual(firstAvailableDate) || firstPublicationDate.isAfter(firstAvailableDate)
+    val publicationsOnOrAfterFirstAvailableDate = details.filterNot(_.publicationDate isBefore firstAvailableDate)
     MutabilityFlags(
-      isFullyMutable = !isWithdrawn && actionedCount == 0 && firstPublicationDateIsGreaterOrEqualToFirstAvailableDate,
-      isEndDateEditable = !isWithdrawn && firstAvailableDate.isBefore(lastPublicationDate)
+      isFullyMutable = isNotWithdrawn && actionedCount == 0 && publicationsOnOrAfterFirstAvailableDate.length == details.length,
+      isEndDateEditable =
+        isNotWithdrawn && publicationsOnOrAfterFirstAvailableDate.nonEmpty && !publicationsOnOrAfterFirstAvailableDate.exists(_.isActioned)
```
mario-galic, line 55: Is it maybe correct to write

```
isEndDateEditable = isNotWithdrawn && publicationsOnOrAfterFirstAvailableDate.nonEmpty && actionedCount == 0
```

instead of 

```
isEndDateEditable = isNotWithdrawn && publicationsOnOrAfterFirstAvailableDate.nonEmpty && !publicationsOnOrAfterFirstAvailableDate.exists(_.isActioned)
```

handlers/holiday-stop-api/src/main/scala/com/gu/holiday_stops/WireHolidayStopRequest.scala
```diff
@@ -11,49 +11,48 @@ import play.api.libs.json._
...
       actualPrice = detail.Actual_Price__c.map(_.value),
-      invoiceDate = detail.Expected_Invoice_Date__c.map(_.value)
+      invoiceDate = detail.Expected_Invoice_Date__c.map(_.value),
+      isActioned = detail.Is_Actioned__c
     )
   }
 
   def calculateMutabilityFlags(
-    isWithdrawn: Boolean,
+    isNotWithdrawn: Boolean,
```
mario-galic, line 46: I always worry about refactoring booleans, but it seems you updated everything accordingly.

handlers/holiday-stop-api/src/main/scala/com/gu/holiday_stops/WireHolidayStopRequest.scala
twrichards: Nope because `actionedCount` refers to ALL of the individual publications, but the purpose of this `isEndDateEditable` is for when the holiday stop is partially actioned but not yet entirely actioned (i.e. end date can still be changed)

handlers/holiday-stop-api/src/main/scala/com/gu/holiday_stops/WireHolidayStopRequest.scala
twrichards: Yeah I was in two minds, I just felt it was more readable than `!isWithdrawn` in *two* places

twrichards:
## MANUAL TEST
Take this test holiday stop where all items are 'actioned' (akin to a backfilled holiday stop)...
image
In the MMA UI (which hangs off these `mutabilityFlags`)...
#### BEFORE 😢
image
#### AFTER
image

twrichards:
Manually tested https://github.com/guardian/support-service-lambdas/pull/656/commits/139d5d7a32a1f632f600b12056e6fc9e10525714 and https://github.com/guardian/support-service-lambdas/pull/656/commits/f41a9407445bdede73fadfebebbc1fdfe16b8e3a ✅

handlers/holiday-stop-api/src/main/scala/com/gu/holiday_stops/Handler.scala
```diff
@@ -465,7 +465,8 @@ object Handler extends Logging {
       contact <- extractContactFromHeaders(req.headers)
       pathParams <- req.pathParamsAsCaseClass[SpecificHolidayStopRequestPathParams]()
       existingForUser <- lookupOp(contact, None, None).toDisjunction.toApiGatewayOp(s"lookup Holiday Stop Requests for contact $contact")
-      _ = existingForUser.exists(_.Id == pathParams.holidayStopRequestId).toApiGatewayContinueProcessing(ApiGatewayResponse.forbidden("not your holiday stop"))
+      hsr <- existingForUser.find(_.Id == pathParams.holidayStopRequestId).toApiGatewayContinueProcessing(ApiGatewayResponse.forbidden("not your holiday stop"))
+      _ = hsr.Holiday_Stop_Request_Detail__r.exists(_.records.exists(_.Is_Actioned__c)).toApiGatewayContinueProcessing(ApiGatewayResponse.forbidden("can't withdraw holiday stop with any actioned items"))
```
johnduffell, line 469: should have been `<-` not `=`

PR #636, opened 2020-03-31
Fix silent Salesforce parsing email errors

mario-galic:
Related PR: https://github.com/guardian/support-service-lambdas/pull/635

This fixes one of the silent errors - returning success on parsing errors. The remaining problems are at least

- no alarms on any error
- what does Salesforce do with non-200 reponse?

```
FIXME: batch-email-sender seems to be implemented on partial success principle in which case what should the response code be?
```

The issue was there probably since inception of batch-email-sender.

PR #657, opened 2020-05-14
Alarm on failed emails

mario-galic:
Related PRs: 

- https://github.com/guardian/support-service-lambdas/pull/635
- https://github.com/guardian/support-service-lambdas/pull/636

Fixes no alarms on any error.

### DIFF
--- a/handlers/batch-email-sender/cfn.yaml
+++ b/handlers/batch-email-sender/cfn.yaml
@@ -61,8 +61,7 @@ Resources:
     Condition: IsProd
     Properties:
       AlarmName: "URGENT 9-5 - PROD: Failed to send email triggered by Salesforce"
-      AlarmDescription: Email message sent from Salesforce to the Service Layer could not be processed. Logs at /aws/lambda/batch-email-sender-PROD
-  repo at https://github.com/guardian/support-service-lambdas/blob/master/handlers/batch-email-sender/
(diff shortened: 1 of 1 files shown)


mario-galic:
[TASK: Fix silent failures in email pipeline](https://trello.com/c/Y9bZ97Mk/1108-task-fix-silent-failures-in-email-pipeline)

handlers/batch-email-sender/cfn.yaml
```diff
@@ -52,6 +56,29 @@ Resources:
                   - "arn:aws:sqs:${AWS::Region}:${AWS::AccountId}:${QueueName}"
                   - { QueueName: !FindInMap [ StageMap, !Ref Stage, EmailSendingQueueName ]}
 
+  FailedRefundAlarm:
```
twrichards, line 59: Probably not the logical name you want

twrichards:
This lambda has some fairly thorough documentation https://github.com/guardian/support-service-lambdas/blob/e9888f89adee375f85bfc10fec69f1ef36ea0916/handlers/batch-email-sender/README.md#responses , that specifically talks about the various response codes.

mario-galic:
>  that specifically talks about the various response codes.

Those response code were never returned. All errors were previously swallowed and 200 was returned.

mario-galic:
> I think there's deeper improvements that could be made here.

In #636 I introduced the following fixme:

```
FIXME: batch-email-sender seems to be implemented on partial success principle in which case what should the response code be?
```

Partial success is much harder to model. It means Salesforce would have to parse the response for both failures and success and have specific logic for those types of email that have to set the corresponding flag.

PR #658, opened 2020-05-21
Add new Baton account invoke permissions

laurenemms:
This adds permissions for the Zuora SAR lambda to be invoked from the new Baton AWS account.

The identity invoke role will be deleted once the migration is completed.

https://trello.com/c/oFgWKRhL/858-add-role-to-all-lambdas-for-new-baton-aws-account

### DIFF
--- a/handlers/zuora-sar/cfn.yaml
+++ b/handlers/zuora-sar/cfn.yaml
@@ -19,9 +19,12 @@ Parameters:
     VpcSubnets:
         Description: Subnets to use in VPC
         Type: CommaDelimitedList
+    BatonAccountId:
+      Description: Account Id for Baton AWS account
+      Type: String
 
 Resources:
-    IdentityInvokeRole:
+    IdentityInvokeRole: #TODO delete once Baton is migrated to new account
         Type: AWS::IAM::Role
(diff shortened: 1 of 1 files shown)


laurenemms:
> Same here LGTM. But _root_.
> 
> Please do add some more detail to the PR comment or in the conversation answer why root and what it means. And add a link to somewhere else in codebase where this is. And ensure to add Trello cards for your TODOs. 👍
> 
> But this is super and technical in a way that I am yet to learn.

Root basically just means giving the permissions to the Baton AWS account itself. This allows cross-account invocation. Root user ARNs are often used in the Principal elements of policies.

The Principal element's specify who can assume the role (more info here).

Here are a couple of other example from ophan-data-lake where we are using root user arns to permission cross-account functionality:

https://github.com/guardian/ophan-data-lake/blob/6fabb528e8c701233d545b378408008c20bc5094/cloudformation/cross-account-athena-access.yaml#L30
https://github.com/guardian/ophan-data-lake/blob/877497e35bf2ebae84e707f05d052db32aea4ed8/baton-lambda/cfn.yaml#L79

I've added cards for the TODOs.

PR #660, opened 2020-05-22
Revert 500 email response on partial success and parsing errors 

mario-galic:
Fixes regression introduced at https://github.com/guardian/support-service-lambdas/commit/7bbddd89af7ade2225c3669f66dcea457d8f3006?diff=split#diff-3552780353f082a1e387abaf81274d57R32 which changed the response from 502 

```
{
  "message": "There were items that were not added to the queue.",
  "code": 502,
  "failed_item_ids": [
    "12345"
  ]
}
```

to 500 

```
{
  "message": "Internal server error"
}
```

Note this also reverts handling of parsing errors to previous behaviour where invalid batch items  were just silently dropped, so the question remains

```
FIXME: What should be the behaviour with some/all items having parsing errors?
```

Here are examples of what salesforce expects

- https://github.com/guardian/salesforce/blob/215073420ece6b8cc83d5310f9860b6130748696/src/classes/batch_sendHolidayStopsToServiceLayer.cls#L72
- https://github.com/guardian/salesforce/blob/215073420ece6b8cc83d5310f9860b6130748696/src/classes/batch_PaymentFailuresToServiceLayer.cls#L88

### DIFF
--- a/handlers/batch-email-sender/build.sbt
+++ b/handlers/batch-email-sender/build.sbt
@@ -20,3 +20,10 @@ libraryDependencies ++= Seq(
 )
 
 assemblyMergeStrategyDiscardModuleInfo
+
+lazy val deployAwsLambda = taskKey[Unit]("Directly update AWS lambda code from DEV instead of via RiffRaff for faster feedback loop")
+deployAwsLambda := {
+  import scala.sys.process._
+  assembly.value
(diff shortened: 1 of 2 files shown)


mario-galic:
Note the difference between batch items with parsing errors and errors due to failure to put (correctly deserialised) items on SQS. The ones with parsing errors were always silently dropped. Even in holiday stops example Salesforce was not aware of them. The 502 message format makes Salesforce aware only of items that failed to be sent **after** being successfully deserialised.

Indeed I will do first name changes separately.

PR #663, opened 2020-05-27
Save metadata file with record count in Datalake raw bucket

mario-galic:
For example, save `InvoiceItem.metadata` to `s3://ophan-raw-zuora-increment-invoiceitem/`
```
{
	"jobId": "2c92c0f9725046a2017255f7d84e5a48",
	"fileId": "2c92c09472503749017255fc3c9379c8",
	"batchId": "2c92c0f9725046a2017255f7d8925a55",
	"status": "completed",
	"name": "InvoiceItem",
	"recordCount": 48650
}
```

Datalake job can then have a pre-condition to check the row count in `.csv` against `.metadata`.

PR #664, opened 2020-05-28
Revert "Save metadata file with record count in Datalake raw bucket"

mario-galic:
Reverts guardian/support-service-lambdas#663

Ophan job failed as it just tries to read everything in the raw bucket as csv.

### DIFF
--- a/handlers/zuora-datalake-export/README.md
+++ b/handlers/zuora-datalake-export/README.md
@@ -94,19 +94,7 @@ If extracting via postman make sure to use the **exact same** `partner` and `pro
 
 ## How to perform full export?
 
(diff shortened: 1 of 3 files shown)


mario-galic:
does it currently read the whole directory in case there's more than one file ?
or can we just tell it to read just one specific file ?
maybe this line ?
https://github.com/guardian/ophan-data-lake/blob/master/etl/subscriptions/src/main/scala/com/gu/datalake/etl/zuora/ZuoraAccountLoader.scala#L24
(I'm not guessing without running any code of course)
if you can get away with just reading one file you can try to add the file name to the line I linked above and test running the job again using runspark and that probably would fix it
if you need to read multiple files you need to provide a wildcard I think

PR #666, opened 2020-05-29
Revert "Revert "Save metadata file with record count in Datalake raw bucket""

mario-galic:
Reverts guardian/support-service-lambdas#664

Save under 

```
bucket/metadata/Refund.metadata
```

instead of 

```
bucket/Refund.metadata
```

because the lake job fails otherwise as it tries to read it as CSV


handlers/zuora-datalake-export/src/main/scala/com/gu/zuora/datalake/export/ExportLambda.scala
```diff
@@ -436,7 +436,7 @@ object SaveCsvToBucket extends LazyLogging {
         val bucket = Query.withName(batch.name).s3Bucket
         val metadata: String = Printer.spaces2.pretty(Metadata(job, batch).asJson)
         val csvRequestWithAcl = putRequestWithAcl(bucket, key = Query.withName(batch.name).s3Key, csvContent)
-        val metadataRequestWithAcl = putRequestWithAcl(bucket, key = s"${batch.name}.metadata", metadata)
+        val metadataRequestWithAcl = putRequestWithAcl(bucket, key = s"metadata/${batch.name}.metadata", metadata)
```
twrichards, line 439: could this extension be `.json` (or `.metadata.json`) now that it's in a `metadata` directory? Just so it describes its content better (and more stuff will be open it by default, including the AWS Console opening it in the browser rather than forcing a download)

PR #667, opened 2020-06-01
Saturday has no whitespace in ratePlanName

mario-galic:
image

Attempt to fix

> unable to book delivery issues at all.

which throws

```
ZuoraApiFailure(Could not derive product type as there are no supported rateplan charges)
/aws/lambda/holiday-stop-api-PROD 
```

@frj I am unable to figure out how to test this using your abstracted unit tests, so this PR is a guess. Tests should be readable scripts, understandable as a unit. Also why did you change the codebase to rely on user inputted text fields? The code used to work on precise IDs, which computers prefer. There is no need for human "understandable" abstractions on config level.

### DIFF
--- a/lib/zuora-core/src/main/scala/com/gu/zuora/subscription/SupportedProduct.scala
+++ b/lib/zuora-core/src/main/scala/com/gu/zuora/subscription/SupportedProduct.scala
@@ -93,6 +93,7 @@ object SupportedProduct {
         SupportedRatePlan("Fiveday", fiveDayCharges),
         SupportedRatePlan("Multi-day", everyDayCharges),
         SupportedRatePlan("Saturday ", saturdayCharges),
+        SupportedRatePlan("Saturday", saturdayCharges),  // Some do not have whitespace
         SupportedRatePlan("Saturday+", saturdayCharges),
         SupportedRatePlan("Sixday", sixDayCharges),
         SupportedRatePlan("Sixday+", sixDayCharges),


kelvin-chappell:
Is the `"Saturday "` value even used? It looks like you could just replace L95 with L96

mario-galic:
I've tested it with Redmond and it seems it fixed the issue. So it seems `"Saturday "` is indeed used. I am unable to follow the flow of code here so could not figure out where to put `trim` safely. Maybe we can merge this, and then smarter people can clean it up in separate PR.

frj:
I think this fix is fine, its only that specific productRatePlan that has the extra space in it.

Its a bit of a mystery of why the space isnt there in the ratePlan, but we can assume that extra spaces will not get randomly added in for other ratePlans. 

So adding the trim for every case is not necessary.

mario-galic:
> I think this fix is fine, its only that specific productRatePlan that has the extra space in it.

How do you know this? Did you ran a query to extract all the names in product catalogue and checked each one?

frj:
The catalog is exported to s3:

https://s3.console.aws.amazon.com/s3/buckets/gu-zuora-catalog/PROD/Zuora-PROD/?region=eu-west-1&tab=overview

I looked in there.

mario-galic:
Next check every single record on every object that references `ratePlanName` as it held at different points in time a different value.

mario-galic:
Also do not forget to check the same thing in Salesforce as time is out of phase with Salesforce.

PR #670, opened 2020-06-29
Add redemption history endpoint

arausuy:
## What does this change?
This PR has a first pass at returning 20 redemption attempts for a Digital Voucher.

We will look to expand this out to have pagination on the next go round.

The API returns a successful response when hit.


lib/imovo/imovo-sttp-client/src/main/scala/com/gu/imovo/ImovoClient.scala
```diff
@@ -175,6 +197,20 @@ object ImovoClient extends LazyLogging {
           None
         )
       }
+
+      override def getRedemptionHistory(subscriptionId: SfSubscriptionId): EitherT[F, ImovoClientException, ImovoRedemptionHistoryResponse] = {
```
kelvin-chappell, line 201: Perhaps add a link to documentation of the call here

PR #671, opened 2020-07-02
Add CorporateAccountId__c and RedemptionCode__c to Zuora Subscription export 

johnduffell:
This PR adds the new corporate subs fields to the DL export query.

https://trello.com/c/GfucWmsP/3068-get-redemption-code-corporate-code-into-dl-for-the-redemption-reporting-subs-table

We need these to report back to the corporate client the uptake of their subscriptions in aggregate, so they know if they are getting value for money from the deal.  They also might want to encourage their customers to sign up, if the uptake is not going well.

Once we have this data we can also link to engagement via the identity ids etc.

The code itself is not essential if we have the corporate ID, but it will be useful for diagnosing issues with particular codes without having to query zuora directly.

@mario-galic I gather you are the expert on zuora exports as this will presumably need a full export from subs table to be done - perhaps you can help there please!

### DIFF
--- a/handlers/zuora-datalake-export/src/main/scala/com/gu/zuora/datalake/export/ExportLambda.scala
+++ b/handlers/zuora-datalake-export/src/main/scala/com/gu/zuora/datalake/export/ExportLambda.scala
@@ -188,7 +188,7 @@ object Query extends Enum[Query] {
   )
   case object Subscription extends Query(
     "Subscription",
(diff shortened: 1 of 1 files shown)


twrichards:
FYI @guardian/supporter-experience 

because Zuora decides to wang in additional columns that you haven't asked for in your query, you will need to carefully create a sister PR on the Ophan side to update the schema used for ingestion - being mindful of the order

https://github.com/guardian/zuora-full-export will also need to be updated (it's the awesome tool that @mario-galic made to make full exports safer and easier)

johnduffell:
ah, perfect comment @twrichards - so it's the order that matters!! I was just staring at the DL code wondering how it maps the field names onto the query! 😂

mario-galic:
Adding a field results in full export of just that object. In the best case scenario the amount of work is about one day:

- Add fields to Subscription object
- Execute incremental export lambda
- Zuora Stateful AQuA API triggers a full export of just Subscription, whilst other objects continue as is
- If we are lucky Zuora will not terminate the query within 8 hours, then
  - wait for the incremental export lambda to fail (because of 15 min AWS Lambda limit)
  - once Zuora batch jobs are completed manually copy the fully exported Subscription.csv to corresponding raw bucket in datalake
  - run spark job to ingest it
  - re-run the export lambda by setting `{"exportFromDate": "YYYY-MM-DD"}` where `YYYY-MM-DD` is few days ago, which should bring just incremental changes for all objects, and thus succeed within 15 min limit
  - re-run all the spark jobs via airflow by just clearing Accounts (which should be the parent of all tables) in airflow

If Zuora decides `Subscription` is too big for 8 hour limit, then 
- `zuora-full-export` app has to be used to extract `Subscription`
- manually copy it to corresponding raw datalake bucket
- ingest by running the job
- place permanent `WHERE` filter on Subscription object to retrieve from date after the full export

mario-galic:
@johnduffell @rupertbates Please add these fields to PROD Zuora as currently it errors

```
FROM Subscription, Error = There is no field named CorporateAccountId__c
```

In DEV it looks like this

image

Please fill in description with enough information to be able to follow up the meaning of this field.

johnduffell:
was a typo, "account" had 3 c's, now fixed

PR #675, opened 2020-07-22
Add digital voucher export query

arausuy:
## What does this change?

We need to export the card and letter values to the lake. This PR contains the query to make that happen. 

## How to test
You can use the test class `StartJobManualTest.scala` to trigger a run of it in DEV. Alternatively you can trigger the lambda via the AWS console.

The appearance of a file in s3 containing current Digital Voucher data.

We need to ensure that we have enabled object access in Salesforce otherwise the job will fail.


mario-galic:
I think you need to add schedule rule for new object to cloudformation.

arausuy:
Thanks @mario-galic I was just about to ask how it is triggered - I've pushed changes in the cfn.yaml now

PR #680, opened 2020-08-13
Enable credit products to vary with subscriptions

kelvin-chappell:
## What does this change?
This refactor enables us to have a range of credit products (rate plans) that are specific to particular product rate plans.
Eg. a credit product for taxable vouchers and another one for non-taxable home delivery.

After this, the only changes to support different credit products will be in `HolidayCreditProduct` and `DeliveryCreditProduct`.

Nothing will change.  The holiday and delivery-credit processors will continue to work.

### DIFF
--- a/handlers/delivery-problem-credit-processor/src/main/scala/com/gu/deliveryproblemcreditprocessor/DeliveryCreditProduct.scala
+++ b/handlers/delivery-problem-credit-processor/src/main/scala/com/gu/deliveryproblemcreditprocessor/DeliveryCreditProduct.scala
@@ -1,5 +1,6 @@
 package com.gu.deliveryproblemcreditprocessor
 
+import com.gu.creditprocessor.Processor.CreditProductForSubscription
 import com.gu.util.config.Stage
 import com.gu.zuora.subscription.CreditProduct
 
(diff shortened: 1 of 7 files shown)


handlers/delivery-problem-credit-processor/src/main/scala/com/gu/deliveryproblemcreditprocessor/DeliveryCreditProduct.scala
```diff
@@ -25,9 +26,10 @@ object DeliveryCreditProduct {
...
   )
 
-  def forStage(stage: Stage): CreditProduct = stage match {
-    case Stage("PROD") => DeliveryCreditProduct.Prod
-    case Stage("CODE") => DeliveryCreditProduct.Code
-    case _ => DeliveryCreditProduct.Dev
-  }
+  def forStage(stage: Stage): CreditProductForSubscription =
+    stage match {
+      case Stage("PROD") => _ => DeliveryCreditProduct.Prod
```
arausuy, line 31: What is the need for the new `_ =>`?

arausuy (approved):

PR #681, opened 2020-08-18
Replace dead dependency with out-of-the-box functionality to unblock migration to Scala 2.13

mario-galic:
## What does this change?

https://github.com/guardian/support-service-lambdas/pull/505 has pulled into our codebase 9 new dependencies out of which at least couple are dead.

circe-java8 is dead since circe 0.12.0-M3 and there is no Scala 2.13 build so we are blocked from migrating. We cannot just upgrade circe because it contains breaking changes https://github.com/circe/circe/releases/tag/v0.12.0-M3, so replace with regular circe-core codec.

Eventually being able to migrate to Scala 2.13.

No idea why this snippet is necessary. Seems to be used only by test named `SalesforceClientTest`.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -99,7 +99,7 @@ lazy val `salesforce-sttp-client` = all(project in file("lib/salesforce/sttp-cli
   )
   .settings(
     libraryDependencies ++=
-      Seq(sttp, sttpCirce, sttpCats % Test, scalatest, catsCore, catsEffect, circe, circeJava8) ++ logging
+      Seq(sttp, sttpCirce, sttpCats % Test, scalatest, catsCore, catsEffect, circe) ++ logging
   )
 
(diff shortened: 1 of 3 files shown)


mario-galic:
This codec is for `Instant` type which seems to be used only at `DigitalVoucherUpdate` https://github.com/guardian/support-service-lambdas/blob/4f89a44c99a7d0c738fc5a46c571c3c64a78985a/handlers/digital-voucher-cancellation-processor/src/main/scala/com/gu/digital_voucher_cancellation_processor/DigitalVoucherCancellationProcessorService.scala#L21

however at that point in code it is using automatic codecs `import io.circe.generic.auto._` and NOT this custom one. This is the only place where `Instant` is used whilst throughout the rest of the handlers `LocalDateTime` or `LocalDate` is used. I wonder why this deviation from standard was necessary? Was there a reason and we have to change everything to `Instant` now, or was `Instant` unnecessary?

PR #682, opened 2020-08-19
Retry Zuora lock contention on subscription update

mario-galic:
## What does this change?

- https://trello.com/c/9cvEbVoa/1543-fix-zuora-locking-contention-alarm-staleobjectstateexception-s
- Attempt to address `StaleObjectStateException` with retry as per Zuora recommendation.
- Retry is implemented using vanilla Scala and attempted right away. If it does not help, we can look into more sophisticated implementation with backoff or library. @kelvin-chappell implemented one using ZIO backend in https://github.com/guardian/price-migration-engine/pull/126
- This should work for holiday and delivery credits

----

As a side note, I have improved a bit `deployAwsLambda` so now from root project

```
 deploy holiday-stop-processor CODE
```

however this is still just a quick fix. I can make it better via sbt plugin later.

## How to test

Look out for log message

```
WARN - Retrying A-S00080884 due to locking contention 53500050... Follow up if all is OK.
```

and make sure to check PROD for double writes.

No more related alarm in our inbox.

This is a retry on a write (PUT) to Zuora which carries risk of double-crediting. However retry was already effectively happening on subsequent run of processor and so far no one reported over-crediting.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -424,3 +424,17 @@ initialize := {
   assert(sys.props("java.specification.version") == "1.8",
     "Java 8 is required for this project.")
 }
+
+lazy val deployAwsLambda = inputKey[Unit]("Directly update AWS lambda code from DEV instead of via RiffRaff for faster feedback loop")
+deployAwsLambda := {
+  import scala.sys.process._
+  import complete.DefaultParsers._
+  val Seq(name, stage) = spaceDelimited("<arg>").parsed
+  s"aws lambda update-function-code --function-name $name-$stage --zip-file fileb://handlers/$name/target/scala-2.12/$name.jar --profile membership --region eu-west-1".!
+}
+
+// run from root project: deploy holiday-stop-processor CODE
+commands += Command.args("deploy", "<name stage>") { (state, args) =>
+  val Seq(name, stage) = args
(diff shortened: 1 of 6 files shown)


build.sbt
```diff
@@ -424,3 +424,17 @@ initialize := {
   assert(sys.props("java.specification.version") == "1.8",
     "Java 8 is required for this project.")
 }
+
+lazy val deployAwsLambda = inputKey[Unit]("Directly update AWS lambda code from DEV instead of via RiffRaff for faster feedback loop")
```
arausuy, line 428: This is great

lib/credit-processor/src/main/scala/com/gu/creditprocessor/Processor.scala
```diff
@@ -37,7 +38,9 @@ object Processor {
       subscription: Subscription,
       update: SubscriptionUpdate
     ): ZuoraApiResponse[Unit] =
-      Zuora.subscriptionUpdateResponse(config, zuoraAccessToken, sttpBackend)(subscription, update)
+      retryLockingContention(2, subscription.subscriptionNumber) {
```
arausuy, line 41: What two here? A name would be good here

PR #685, opened 2020-09-03
Rename references from master to main

kelvin-chappell:
## What does this change?
All references to default branch now refer to 'main' rather than 'master'.

## How to test
Check whether it works in Teamcity and in Riffraff.

All subprojects deployed successfully.

Yes.

## Images
N/A


paulbrown1982 (approved):
👍

PR #687, opened 2020-09-07
Constant names should be in upper camel case

mario-galic:
https://docs.scala-lang.org/style/naming-conventions.html#constants-values-variable-and-methods


kelvin-chappell (approved):
👍

PR #688, opened 2020-09-08
Configure HolidayCreditProduct for Code stage

kelvin-chappell:
This is the same as #686 but for the Code stage rather than the Dev stage.

We believe that tax arrangements can be adequately covered by credit products for GW, HD and Voucher.  There might be no need for separate 'Plus' products so I haven't created them in the Zuora UAT environment and they're not catered for in the codebase.

### DIFF
--- a/handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/HolidayCreditProduct.scala
+++ b/handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/HolidayCreditProduct.scala
@@ -2,7 +2,7 @@ package com.gu.holidaystopprocessor
 
 import com.gu.creditprocessor.Processor.CreditProductForSubscription
 import com.gu.util.config.Stage
-import com.gu.zuora.subscription.{CreditProduct, RatePlan, RatePlanCharge}
+import com.gu.zuora.subscription.{CreditProduct, RatePlan}
 
 /**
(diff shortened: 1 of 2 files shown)


mario-galic (approved):
I wonder if it is worth replacing if-else chain with something like

```
subscription =>
  subscription
    .ratePlans
    .flatMap(planToCreditProduct)
    .headOption
    .getOrElse(throw new IllegalArgumentException(s"No holiday credit product available for subscription ${subscription.subscriptionNumber}"))
```

where

```
def planToCreditProduct(plan: RatePlan): Option[CreditProduct] =
  plan.productName match {
    case v if v.startsWith("Guardian Weekly") => Some(HolidayCreditProduct.Code.GuardianWeekly)
    case "Newspaper Delivery" => Some(HolidayCreditProduct.Code.HomeDelivery)
    case "Newspaper Voucher" | "Newspaper Digital Voucher" => Some(HolidayCreditProduct.Code.Voucher)
    case _ => None
  }
```

In Scala 2.13 we could use interpolated string pattern

```
  plan.productName match {
    case s"Guardian Weekly$_" => Some(HolidayCreditProduct.Code.GuardianWeekly)
    ...
  }
```

kelvin-chappell:
@mario-galic I tried playing around with this code but I think probably the way it is is as clear as anything else.  So I'm going to merge it as-is.  I like the new case match though so might revisit once all the stages are complete.

PR #689, opened 2020-09-08
Make invoice date for delivery credit requests optional

kelvin-chappell:
## What does this change?
The delivery credit processor used to rely on credit requests having their next invoice date pre-populated by a call to the holiday-stop API.  With this change it becomes possible for the credit processor to determine and return the invoice date if it hasn't already been filled in.  This makes it easier to do bulk requests if necessary.

## How to test
It's been tested it in the Code environment.

A delivery-credit request with no invoice date populated results in a successful credit applied and the invoice date is written back to Salesforce.

Yes.

## Images
N/A


mario-galic:
How does it know when to apply the credit if the date is not provided?

kelvin-chappell:
The processor works it out [here](https://github.com/guardian/support-service-lambdas/blob/main/lib/zuora-core/src/main/scala/com/gu/zuora/subscription/SubscriptionUpdate.scala#L29) if it's not provided.  Holiday-stop requests don't provide invoice dates.

PR #693, opened 2020-09-10
Enable DEV stage for holiday-stop-processor

mario-galic:


kelvin-chappell (approved):
I couldn't decide whether to take the Dev code out of the HolidayCreditProduct logic or create a Dev stage.  But this way it is at least consistent with all our other projects. 👍

PR #696, opened 2020-09-11
Add home-delivery delivery-credit product in Dev

kelvin-chappell:
This is the equivalent to #686 but for delivery credits, as opposed to holiday credits, in the Dev stage.

There's also a refactoring of the HolidayStopCredit code.

And an update of the description of the delivery credit schedule.


lib/config-core/src/main/scala/com/gu/util/config/Stage.scala
```diff
@@ -5,5 +5,10 @@ case class Stage(value: String) extends AnyVal {
 }
 
 object Stage {
-  def apply(): Stage = Option(System.getenv("Stage")).fold(Stage("DEV"))(stage => Stage(stage))
+
+  val Dev: Stage = Stage("DEV")
+  val Code: Stage = Stage("CODE")
+  val Prod: Stage = Stage("PROD")
+
+  def apply(): Stage = Option(System.getenv("Stage")).fold(Dev)(stage => Stage(stage))
```
mario-galic, line 13: Another alternative is 

```
sys.env.get("Stage").fold(Dev)(stage => Stage(stage))
```

kelvin-chappell: Good tip.  So many different ways to do Java in Scala!

handlers/delivery-problem-credit-processor/src/test/scala/com/gu/deliveryproblemcreditprocessor/Fixtures.scala
```diff
@@ -0,0 +1,15 @@
+package com.gu.deliveryproblemcreditprocessor
+
+import com.gu.zuora.subscription.Subscription
+import io.circe.generic.auto._
+import io.circe.parser.decode
+import org.scalatest.Assertions
+
+import scala.io.Source
+
+object Fixtures extends Assertions {
```
mario-galic, line 10: The name `Fixture` might be a bit confusing because ScalaTest uses the name for dedicated facilities for fixtures  https://www.scalatest.org/user_guide/sharing_fixtures. But I know we use the same patter elsewhere.

mario-galic (approved):

PR #697, opened 2020-09-14
Add delivery-problem credit products for Prod stage

kelvin-chappell:
This configures different credit products for home-delivery delivery problems and Guardian Weekly delivery problems, so that they can each have a different tax configuration.

Follows on from #696 

**NB.  This cannot be merged until the new tax configuration is ready and confirmed.**


mario-galic (approved):
I did a quick search for hardcoded `Delivery-problem credit` https://github.com/search?q=org%3Aguardian+Delivery-problem+credit&type=code

It seems it is used only in the sense of `RatePlanCharge.Name` (as opposed to `RatePlan.Name`) which did not change.

arausuy (approved):
With the tax change my understanding is we now only need these two products for delivery credits!

I'm going to trust the mappings!

kelvin-chappell:
@arausuy That's how I understand it.  It should be easy to tweak now if we end up needing more products.

kelvin-chappell:
@mario-galic I changed the name of the default product a few weeks ago so hopefully if anything is relying on it it will already have broken.

PR #699, opened 2020-09-15
Salesforce gdpr billing account removal

david-pepper:
# Description
This feature enables the automated removal of Zuora records in Salesforce for GDPR Retention, by removing the crmId in Zuora of Billing Accounts marked for removal in Salesforce. 

When the crmId is removed from a Billing Account in Zuora, z360 will delete the Billing Account, and all of its related Zuora records (Zuora__Subscription__c, zqu__ProductRatePlan__c, Zuora__PaymentMethod__c, Zuora__Refund__c etc) in Salesforce.

This lambda works alongside logic in Salesforce to surface Billing Accounts for removal, as shown in the image below:

image

## Lambda Actions
The lambda performs the following actions:

1. Performs a GET HttpRequest to Salesforce, to obtain a list of all Billing Accounts where the parent Account record is marked as ‘GDPR_Billing_Accounts_Ready_for_Removal__c’ 

2. For each Billing Account retrieved, perform an individual callout to Zuora to remove the CrmId and set status=’Canceled’

3. Any failures that occur (for example if the Billing Account is already canceled, or where there are active subscriptions, or something else), will be collected and two subsequent PATCH HttpRequests will be made to Salesforce:

- Update Billing Account field GDPR_Removal_Attempts__c by incrementing by 1
- Insert Apex_Error__c record that contains the details of the error

### Identifying Billing Accounts for removal
Billing Accounts will be marked for removal when the parent Account, and its related records fall outside the currently agreed time window for retaining the records, which is currently set to be 30 months since the last 'touchpoint'

The last **touchpoint** is defined as the most recent interaction with the customer, which could be the creation date or last modified of a case, the cancellation date of a subscription, or the date of a response to a customer survey, amongst others. 

The specific logic for the last touchpoint for a given set of customer records is contained within the AccountRetentionBatch job in Salesforce, and is not the focus of this PR.

### Lamdba maximum runtime is 15 minutes
During testing, it was observed that processing a set of 200 records in a response from Salesforce, from the executing initial query, to writing back to Salesforce any errors that occurred, took approximately 2.5 minutes, which is well within the 15 minute time window for operations to take place. 

### Salesforce API limits
Salesforce API callouts are minimised to 2 or 4 depending on whether there are any failures when calling out to Zuora:

 - When querying for records from Salesforce, one query is used to get a maximum of 200 records in one call

- The lambda then splits this record list and performs individual callouts to Zuora to do the necessary update.

- Any failures are then collated into a list and parsed into a json object that uses the Salesforce REST API Composite request to update all relevant Billing Accounts, and another json object that performs another composite request to insert the relevant Apex Error records.

## Scheduling
When the process is activated for the first time, there will be a much higher number of records (>40,000)  to process than there would be for the usual daily run (estimated to be less than 200 per day). 

In order to process higher numbers of records, using recursion was briefly implemented and tested, but it was subsequently decided that the added complexity of using recursion could be eliminated by simply scheduling the lambda to execute with a higher frequency (once every 20 minutes between 10:00 and 15:00 each day) to clear the backlog of records, and once complete, to reschedule the lambda to run daily to process the daily cohort of records which have fallen outside the retention window.

### Risk
An attempt to remove the crmId in Zuora for a Billing Account fails 

### How it is handled
This will result in an update to GDPR_Removal_Attempts__c field on Salesforce Billing Account record, as well as the creation of corresponding Apex Error records where the error message and error code from the failure response from Zuora are logged. 

Each time the lambda runs, it will extract Billing Accounts where GDPR_Removal_Attempts__c <= 5, therefore for each Billing Account that encounters an error when updating in Zuora, there will be 4 subsequent retries before it is omitted from the resultset returned to the Lambda for processing.

### Risk
5 or more consecutive attempts to remove the crmId in Zuora for a Billing Account fail

### How it is handled
It is intended that a report will be scheduled in Salesforce, that will identify any Billing Accounts where GDPR_Removal_Attempts__c >=5. This report will be run weekly, and the contents of which will be emailed to the relevant team to investigate the errors.

### Risk
Calling back to Salesforce to increment Removal Attempts on Billing Account or creating Error records fails

### How it is handled
In this scenario, the data stored in Salesforce will not accurately represent the state of the Billing Account’s journey for removal, in that its value for  GDPR_Removal_Attempts__c will remain as 0, and/or there will be no Apex Errors to provide further context to the issue. 

This is handled by the addition of a new datetime field on the Salesforce Account object (parent to Billing Account). The field will contain the date time that the Account and its related records (which include Billing Account) were initially marked as ‘ready for gdpr removal’. This field will let us know if any records have been ready for removal for an unintended amount of time, for example 1 week or more. A report should be scheduled to identify any Accounts where the date they are marked for removal is more than one week in the past at the time the report is run.

## Testing
A Runner collection was executed in Postman to generate the following test data sets:

1. 200 Billing Accounts that should successfully be removed by the execution of the lambda

2. 200 Billing Accounts that should fail when execution of the lambda takes place. This was achieved by ensuring that each of the Billing Accounts had 1 active subscription

3. A combination of 100 Billing Accounts that should be successfully removed and 100 Billing Accounts that should fail (as in 2 above)

Each of the test scenarios yielded the expected results, namely that successes resulted in the Billing Accounts being removed from Salesforce, and failures resulted in the incrementing of GDPR_Removal_Attempts__c by 1, and an Apex Error records being generated for each failure.

See [BillingAccountRemover Test Data](https://docs.google.com/spreadsheets/d/1QX_uP9Yxmm8xQHh2BZePOPXtNq4u-ZH8w2qYFXYnC78/edit?usp=sharing)

**BillingAccounts.GDPR_Removal_Attempts__c**
image

**New Apex Error on failure**
image

## Release
It is intended to deploy this solution to production but not to schedule it to run until a future date. There are still some decisions to be made as to whether the scope of the actions that the lambda performs should be expanded, for example to delete the billing account entirely in Zuora, or to set values in some other fields.

### DIFF
--- a/handlers/sf-billing-account-remover/build.sbt
+++ b/handlers/sf-billing-account-remover/build.sbt
@@ -1,0 +1,13 @@
+version := "0.1"
+
+scalaVersion := "2.13.3"
+
+libraryDependencies += "org.scalaj" %% "scalaj-http" % "2.4.2"
+libraryDependencies += "com.lihaoyi" %% "upickle" % "0.9.5" // SBT
+val circeVersion = "0.12.3"
+
+libraryDependencies ++= Seq(
+  "io.circe" %% "circe-core",
+  "io.circe" %% "circe-generic",
+  "io.circe" %% "circe-parser"
+).map(_ % circeVersion)
--- a/handlers/sf-billing-account-remover/build.sbt
+++ b/handlers/sf-billing-account-remover/build.sbt
@@ -3,7 +3,6 @@ version := "0.1"
 scalaVersion := "2.13.3"
 
 libraryDependencies += "org.scalaj" %% "scalaj-http" % "2.4.2"
-libraryDependencies += "com.lihaoyi" %% "upickle" % "0.9.5" // SBT
 val circeVersion = "0.12.3"
 
 libraryDependencies ++= Seq(
--- a/handlers/sf-billing-account-remover/src/main/scala/com/gu/sf_billing_account_remover/BillingAccountRemover.scala
+++ b/handlers/sf-billing-account-remover/src/main/scala/com/gu/sf_billing_account_remover/BillingAccountRemover.scala
@@ -7,14 +7,15 @@ import io.circe.syntax._
 import scalaj.http._
 
 object BillingAccountRemover extends App {
+  //Salesforce
   case class SfAuthDetails(access_token: String, instance_url: String)
 
-  //Salesforce
   case class SfGetBillingAccsResponse(
     done: Boolean,
     records: Seq[BillingAccountsRecords.Records],
     nextRecordsUrl: Option[String] = None
   )
+
   case class SfGetCustomSettingResponse(
     done: Boolean,
     records: Seq[CustomSettingRecords.Records]
@@ -27,6 +28,7 @@ object BillingAccountRemover extends App {
                                       attributes: Attributes = Attributes(
                                         `type` = "Zuora__CustomerAccount__c"
                                       ))
+
   case class SfBillingAccountsToUpdate(allOrNone: Boolean,
                                        records: Seq[SfBillingAccountToUpdate])
 
@@ -36,6 +38,7 @@ object BillingAccountRemover extends App {
                                    attributes: Attributes = Attributes(
                                      `type` = "Apex_Error__c"
                                    ))
+
   case class SfErrorRecordsToCreate(allOrNone: Boolean,
                                     records: Seq[SfErrorRecordToCreate])
 
@@ -73,12 +76,13 @@ object BillingAccountRemover extends App {
     )
 
   var counter: Int = 0;
-  processPage()
 
-  def processPage(): Unit = {
+  processBillingAccounts()
+
+  def processBillingAccounts(): Unit = {
     val t1 = System.nanoTime
 
-    val iteration = for {
+    for {
       config <- optConfig map (c => Right(c)) getOrElse Left(
         new RuntimeException("Missing config value")
       )
@@ -95,7 +99,6 @@ object BillingAccountRemover extends App {
 
     val duration = (System.nanoTime - t1) / 1e9d
     println("Elapsed time: " + duration + "s")
-    println("maxAttempts: " + iteration)
 
   }
 
--- a/handlers/sf-billing-account-remover/src/main/scala/com/gu/sf_billing_account_remover/BillingAccountRemover.scala
+++ b/handlers/sf-billing-account-remover/src/main/scala/com/gu/sf_billing_account_remover/BillingAccountRemover.scala
@@ -84,7 +84,7 @@ object BillingAccountRemover extends App {
       config <- optConfig map (c => Right(c)) getOrElse Left(
         new RuntimeException("Missing config value")
       )
-      sfAuthDetails <- SfLogin(config.salesforceConfig)
+      sfAuthDetails <- decode[SfAuthDetails](auth(config.salesforceConfig))
       getCustomSettingResponse <- getSfCustomSetting(sfAuthDetails)
       maxAttempts = getCustomSettingResponse.records(0).Property_Value__c
       getBillingAccountsResponse <- getSfBillingAccounts(
@@ -124,7 +124,7 @@ object BillingAccountRemover extends App {
 
     val customSettingFromSf =
       getMaxAttemptsCustomSetting(sfAuthentication)
-
+    println("customSettingFromSf:" + customSettingFromSf)
     val decodedCustomSetting = decode[SfGetCustomSettingResponse](
       customSettingFromSf
     ) map { customSettingObject =>
@@ -300,12 +300,6 @@ object BillingAccountRemover extends App {
       .body
 
   }
-
-  def SfLogin(salesforceConfig: SalesforceConfig) =
-    decode[SfAuthDetails](auth(salesforceConfig)) match {
-      case Right(responseObject) => Right(responseObject)
-      case Left(ex)              => Left(ex)
-    }
 
   object SfUpdateBillingAccounts {
     def apply(
--- a/handlers/sf-billing-account-remover/src/main/scala/com/gu/sf_billing_account_remover/BillingAccountRemover.scala
+++ b/handlers/sf-billing-account-remover/src/main/scala/com/gu/sf_billing_account_remover/BillingAccountRemover.scala
@@ -6,6 +6,8 @@ import io.circe.parser._
 import io.circe.syntax._
 import scalaj.http._
 
+import scala.util.Try
+
 object BillingAccountRemover extends App {
   //Salesforce
   case class SfAuthDetails(access_token: String, instance_url: String)
@@ -197,22 +199,16 @@ object BillingAccountRemover extends App {
   def insertSfErrorRecs(sfAuthDetails: SfAuthDetails,
                         jsonBody: String): Either[Throwable, String] = {
 
-    val response =
-      try {
-        Right(
-          Http(
-            s"${sfAuthDetails.instance_url}/services/data/v45.0/composite/sobjects"
-          ).header("Authorization", s"Bearer ${sfAuthDetails.access_token}")
-            .header("Content-Type", "application/json")
-            .put(jsonBody)
-            .method("POST")
-            .asString
-            .body
-        )
-      } catch {
-        case ex: Throwable => Left(ex)
-      }
-    response
+    Try {
+      Http(
+        s"${sfAuthDetails.instance_url}/services/data/v45.0/composite/sobjects"
+      ).header("Authorization", s"Bearer ${sfAuthDetails.access_token}")
+        .header("Content-Type", "application/json")
+        .put(jsonBody)
+        .method("POST")
+        .asString
+        .body
+    }.toEither
   }
 
   def updateSfBillingAccs(sfAuthDetails: SfAuthDetails,
@@ -259,7 +255,7 @@ object BillingAccountRemover extends App {
 
   def getBillingAccountsFromSf(maxAttempts: Int,
                                sfAuthDetails: SfAuthDetails): String = {
-    val limit = 1;
+    val limit = 5;
     val devQuery =
       s"Select Id, Zuora__Account__c, GDPR_Removal_Attempts__c, Zuora__External_Id__c from Zuora__CustomerAccount__c where Name like '%Bill-Acc-Remover-ctc%' and createddate>=2020-09-17T09:17:45.000+0000 order by Name limit $limit"
 
--- a/handlers/sf-billing-account-remover/src/main/scala/com/gu/sf_billing_account_remover/BillingAccountRemover.scala
+++ b/handlers/sf-billing-account-remover/src/main/scala/com/gu/sf_billing_account_remover/BillingAccountRemover.scala
@@ -113,15 +113,10 @@ object BillingAccountRemover extends App {
     sfAuthentication: SfAuthDetails
   ): Either[Error, SfGetBillingAccsResponse] = {
 
-    val billingAccountListFromSf =
-      getBillingAccountsFromSf(maxAttempts: Int, sfAuthentication)
-    val decodedBillingAccounts = decode[SfGetBillingAccsResponse](
-      billingAccountListFromSf
-    ) map { billingAccountsObject =>
-      billingAccountsObject
-    }
-
-    decodedBillingAccounts
+    //can we bring this into a one liner and remove the method?
+    decode[SfGetBillingAccsResponse](
+      getBillingAccountsFromSf(maxAttempts, sfAuthentication)
+    )
   }
   def getSfCustomSetting(
     sfAuthentication: SfAuthDetails
--- a/handlers/sf-billing-account-remover/src/main/scala/com/gu/sf_billing_account_remover/BillingAccountRemover.scala
+++ b/handlers/sf-billing-account-remover/src/main/scala/com/gu/sf_billing_account_remover/BillingAccountRemover.scala
@@ -10,15 +10,16 @@ object BillingAccountRemover extends App {
   case class SfAuthDetails(access_token: String, instance_url: String)
 
   //Salesforce
-  case class SfGetBillingAccsResponse(done: Boolean,
-                                      records: Seq[Records],
-                                      nextRecordsUrl: Option[String] = None)
-  case class Records(Id: String,
-                     Zuora__Account__c: String,
-                     Zuora__External_Id__c: String,
-                     GDPR_Removal_Attempts__c: Int,
-                     ErrorMessage: Option[String] = None,
-                     ErrorCode: Option[String] = None)
+  case class SfGetBillingAccsResponse(
+    done: Boolean,
+    records: Seq[BillingAccountsRecords.Records],
+    nextRecordsUrl: Option[String] = None
+  )
+  case class SfGetCustomSettingResponse(
+    done: Boolean,
+    records: Seq[CustomSettingRecords.Records]
+  )
+
   case class Attributes(`type`: String)
 
   case class SfBillingAccountToUpdate(id: String,
@@ -71,41 +72,40 @@ object BillingAccountRemover extends App {
       )
     )
 
-  val limit = 200;
-  val query =
-    s"Select Id, Zuora__Account__c, GDPR_Removal_Attempts__c, Zuora__External_Id__c from Zuora__CustomerAccount__c where Active_Subs__c > 0 and Zuora__External_Id__c != null limit $limit"
   var counter: Int = 0;
-  processPage(query)
+  processPage()
 
-  def processPage(query: String): Unit = {
+  def processPage(): Unit = {
     val t1 = System.nanoTime
-
-    println("process records with query:" + query)
 
     val iteration = for {
       config <- optConfig map (c => Right(c)) getOrElse Left(
         new RuntimeException("Missing config value")
       )
       token <- authToken(config.salesforceConfig)
-      getBillingAccountsResponse <- getSfBillingAccounts(query, token)
+      getCustomSettingResponse <- getSfCustomSetting(token)
+      maxAttempts = getCustomSettingResponse.records(0).Property_Value__c
+      getBillingAccountsResponse <- getSfBillingAccounts(maxAttempts, token)
       sfRecords = getBillingAccountsResponse.records
 
       failedUpdates = updateRecordsInZuora(sfRecords)
       _ <- updateBillingAccountsInSf(token, failedUpdates)
       _ <- insertErrorRecordsInSf(token, failedUpdates)
-    } yield getBillingAccountsResponse
+    } yield maxAttempts
 
     val duration = (System.nanoTime - t1) / 1e9d
     println("Elapsed time: " + duration + "s")
+    println("maxAttempts: " + iteration)
 
   }
 
   def getSfBillingAccounts(
-    query: String,
+    maxAttempts: Int,
     bearerToken: String
   ): Either[Error, SfGetBillingAccsResponse] = {
 
-    val billingAccountListFromSf = getBillingAccountsFromSf(query, bearerToken)
+    val billingAccountListFromSf =
+      getBillingAccountsFromSf(maxAttempts: Int, bearerToken)
     println("billingAccountListFromSf: " + billingAccountListFromSf)
     val decodedBillingAccounts = decode[SfGetBillingAccsResponse](
       billingAccountListFromSf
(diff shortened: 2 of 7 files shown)


handlers/sf-billing-account-remover/src/main/scala/com/gu/sf_billing_account_remover/BillingAccountRemover.scala
```diff
@@ -0,0 +1,351 @@
...
+  }
+
+  def insertSfErrorRecs(bearerToken: String,
+                        jsonBody: String): Either[Throwable, String] = {
+
+    val response =
+      try {
+        Right(
+          Http(
+            "https://gnmtouchpoint--dev.my.salesforce.com/services/data/v45.0/composite/sobjects"
```
kelvin-chappell, line 206: References to the SF domain will need to be in env vars as it will be stage-specific.

kelvin-chappell (approved):
Could do with a Readme, which at this stage would just describe what the lambda is for.  And there are some comments outstanding from https://github.com/david-pepper/gdpr-record-removal/pull/1 but otherwise 👍

david-pepper:
[GOAL: Finish GDPR Cancel Billing Accounts in Zuora (AWS)](https://trello.com/c/CW7mVbhJ/1630-goal-finish-gdpr-cancel-billing-accounts-in-zuora-aws)

handlers/sf-billing-account-remover/src/main/scala/com/gu/sf_billing_account_remover/BillingAccountRemover.scala
```diff
@@ -0,0 +1,350 @@
...
+      }
+    }
+
+  }
+
+  def insertSfErrorRecs(sfAuthDetails: SfAuthDetails,
+                        jsonBody: String): Either[Throwable, String] = {
+
+    val response =
+      try {
```
mario-galic, line 201: This can be simplified to

```
  def insertSfErrorRecs(
    sfAuthDetails: SfAuthDetails,
    jsonBody: String
  ): Either[Throwable, String] = {
    Try {
      Http(s"${sfAuthDetails.instance_url}/services/data/v45.0/composite/sobjects")
        .header("Authorization", s"Bearer ${sfAuthDetails.access_token}")
        .header("Content-Type", "application/json")
        .put(jsonBody)
        .method("POST")
        .asString
        .body
    }.toEither
  }
```

Note the `Try {...}.toEither`. In fact `Either[Throwable, String]` is semantically equivalent to `Try`, so the signature could be 

```
def insertSfErrorRecs(
    sfAuthDetails: SfAuthDetails,
    jsonBody: String
  ): Try[String]
```

Then `.toEither` call is not necessary. `Try` works just like `Either`, but instread of `Left` and `Right` it has `Success` and `Failure`.

handlers/sf-billing-account-remover/src/main/scala/com/gu/sf_billing_account_remover/BillingAccountRemover.scala
```diff
@@ -0,0 +1,350 @@
...
+          "password" -> s"${salesforceConfig.password}${salesforceConfig.token}"
+        )
+      )
+      .asString
+      .body
+
+  }
+
+  def SfLogin(salesforceConfig: SalesforceConfig) =
+    decode[SfAuthDetails](auth(salesforceConfig)) match {
```
mario-galic, line 309: This can be simplified to just `decode[SfAuthDetails](auth(salesforceConfig))`. Note how in pattern matching the values are not actually being transformed, that is, it is effectively `x => x`.

handlers/sf-billing-account-remover/src/main/scala/com/gu/sf_billing_account_remover/BillingAccountRemover.scala
```diff
@@ -0,0 +1,350 @@
...
+      ),
+      ZuoraConfig(
+        apiAccessKeyId = zuoraApiAccessKeyId,
+        apiSecretAccessKey = zuoraApiSecretAccessKey
+      )
+    )
+
+  processBillingAccounts()
+
+  def processBillingAccounts(): Unit = {
```
mario-galic, line 80: Possible simplification

```
def processBillingAccounts(): Unit = {
  for {
    config <- optConfig.toRight(new RuntimeException("Missing config value"))
    sfAuthDetails <- SfLogin(config.salesforceConfig)
    getCustomSettingResponse <- getSfCustomSetting(sfAuthDetails)
    maxAttempts = getCustomSettingResponse.records.headOption.getOrElse(throw new RuntimeException("There should be at least one record")).Property_Value__c
    getBillingAccountsResponse <- getSfBillingAccounts(maxAttempts, sfAuthDetails)
  } {
    val sfRecords = getBillingAccountsResponse.records
    val allUpdates = updateRecordsInZuora(sfRecords)
    val failedUpdates = allUpdates.filter(_.ErrorCode.isDefined)
    if (failedUpdates.nonEmpty) {
      updateBillingAccountsInSf(sfAuthDetails, failedUpdates)
      insertErrorRecordsInSf(sfAuthDetails, failedUpdates)
    }
  }
}
```

Note the usage of `toRight` and `headOption`. Also note the difference between `for {...} {...}` `for {...} yield {...}. When your return type is `Unit` you can just use former because it is just sugar for `foreach` (whilst for-yield is sugar for `flatMap/map`.

handlers/sf-billing-account-remover/build.sbt
```diff
@@ -0,0 +1,13 @@
+version := "0.1"
+
+scalaVersion := "2.13.3"
+
+libraryDependencies += "org.scalaj" %% "scalaj-http" % "2.4.2"
+libraryDependencies += "com.lihaoyi" %% "upickle" % "0.9.5" // SBT
```
mario-galic, line 6: Note that upickle and circe are both serving the same purpose - they are JSON libraries. I do not think you are using `upickle` here.

handlers/sf-billing-account-remover/src/main/scala/com/gu/sf_billing_account_remover/BillingAccountRemover.scala
```diff
@@ -0,0 +1,350 @@
...
+
+    val billingAccountListFromSf =
+      getBillingAccountsFromSf(maxAttempts: Int, sfAuthentication)
+    val decodedBillingAccounts = decode[SfGetBillingAccsResponse](
+      billingAccountListFromSf
+    ) map { billingAccountsObject =>
+      billingAccountsObject
+    }
+
+    decodedBillingAccounts
```
kelvin-chappell, line 117:
```suggestion
    decode[SfGetBillingAccsResponse](
      getBillingAccountsFromSf(maxAttempts: Int, sfAuthentication)
    )
```

handlers/sf-billing-account-remover/src/main/scala/com/gu/sf_billing_account_remover/BillingAccountRemover.scala
```diff
@@ -0,0 +1,350 @@
...
+    val customSettingFromSf =
+      getMaxAttemptsCustomSetting(sfAuthentication)
+
+    val decodedCustomSetting = decode[SfGetCustomSettingResponse](
+      customSettingFromSf
+    ) map { customSettingObject =>
+      customSettingObject
+    }
+
+    decodedCustomSetting
```
kelvin-chappell, line 132:
```suggestion
    decode[SfGetCustomSettingResponse](
      getMaxAttemptsCustomSetting(sfAuthentication)
    )
```

handlers/sf-billing-account-remover/src/main/scala/com/gu/sf_billing_account_remover/BillingAccountRemover.scala
```diff
@@ -0,0 +1,350 @@
...
+        case ex: Throwable => { Left(ex) }
+      }
+    response
+  }
+
+  def updateZuoraBillingAcc(billingAccountForRemovalAsJson: String,
+                            zuoraBillingAccountId: String) = {
+    Http(
+      s"https://rest.apisandbox.zuora.com/v1/object/account/$zuoraBillingAccountId"
+    ).header("apiAccessKeyId", System.getenv("apiAccessKeyId"))
```
kelvin-chappell, line 243: Better to put `apiAccessKeyId` in `Config` and pass `Config` through to method.  That way the program will fail fast if it doesn't have the value in its environment.

handlers/sf-billing-account-remover/src/main/scala/com/gu/sf_billing_account_remover/BillingAccountRemover.scala
```diff
@@ -0,0 +1,350 @@
...
+      } catch {
+        case ex: Throwable => { Left(ex) }
+      }
+    response
+  }
+
+  def updateZuoraBillingAcc(billingAccountForRemovalAsJson: String,
+                            zuoraBillingAccountId: String) = {
+    Http(
+      s"https://rest.apisandbox.zuora.com/v1/object/account/$zuoraBillingAccountId"
```
kelvin-chappell, line 242: The URL is stage-specific so best if in an env var and passed through in `Config`.

handlers/sf-billing-account-remover/src/main/scala/com/gu/sf_billing_account_remover/BillingAccountRemover.scala
```diff
@@ -0,0 +1,350 @@
...
+    ).header("apiAccessKeyId", System.getenv("apiAccessKeyId"))
+      .option(HttpOptions.readTimeout(30000))
+      .header("apiSecretAccessKey", System.getenv("apiSecretAccessKey"))
```
kelvin-chappell, line 245: Better to put env var in `Config` and pass `Config` through to method. That way the program will fail fast if it doesn't have the value in its environment.

handlers/sf-billing-account-remover/src/main/scala/com/gu/sf_billing_account_remover/BillingAccountRemover.scala
```diff
@@ -0,0 +1,350 @@
...
+      s"Select Id, Zuora__Account__c, GDPR_Removal_Attempts__c, Zuora__External_Id__c from Zuora__CustomerAccount__c where Name like '%Bill-Acc-Remover-ctc%' and createddate>=2020-09-17T09:17:45.000+0000 order by Name limit $limit"
+
+    //val prodQuery =
+    s"Select Id, Zuora__Account__c, GDPR_Removal_Attempts__c, Zuora__External_Id__c from Zuora__CustomerAccount__c where Zuora__External_Id__c != null AND Account.GDPR_Billing_Accounts_Ready_for_Removal__c = true AND GDPR_Removal_Attempts__c < $maxAttempts limit $limit"
+
+    val query = devQuery
+
+    Http(s"${sfAuthDetails.instance_url}/services/data/v20.0/query/")
+      .param("q", query)
+      .option(HttpOptions.readTimeout(30000))
```
kelvin-chappell, line 273: The read timeout is repeated around the program so might be better to have a method that encapsulates the call to `Get` that just takes the URL and query string as arguments.

handlers/sf-billing-account-remover/src/main/scala/com/gu/sf_billing_account_remover/BillingAccountRemover.scala
david-pepper: The first suggestion works nicely. Then second suggestion of using Try[String] as the return type and removing the .toEither  call gives this compilation error:

image

david-pepper: Is the Try a scala.util method or should it be from some other library?

mario-galic: You would have to replace all method signatures from `Either[A, B]` to `Try[B]`. Indeed `Try` is standard Scala so `import scala.util.Try`.

handlers/sf-billing-account-remover/src/main/scala/com/gu/sf_billing_account_remover/BillingAccountRemover.scala
david-pepper: Done

handlers/sf-billing-account-remover/src/main/scala/com/gu/sf_billing_account_remover/BillingAccountRemover.scala
david-pepper: Done

handlers/sf-billing-account-remover/src/main/scala/com/gu/sf_billing_account_remover/BillingAccountRemover.scala
david-pepper: Can I further improve this by removing the getSfBillingAccounts() method altogether and moving the decode statement up  to processBillingAccounts() as a one liner:

```
    for {
      config <- optConfig.toRight(new RuntimeException("Missing config value"))
      sfAuthDetails <- decode[SfAuthDetails](auth(config.salesforceConfig))
      getCustomSettingResponse <- getSfCustomSetting(sfAuthDetails)
      maxAttempts = getCustomSettingResponse.records.headOption
        .getOrElse(
          throw new RuntimeException(
            "There should be at least one record returned for MaxAttempts"
          )
        )
        .Property_Value__c
      getBillingAccountsResponse <- decode[SfGetBillingAccsResponse](
        getBillingAccountsFromSf(maxAttempts, sfAuthDetails)
      )
    }
```

This does work, but do we lose the ability to throw an error if one is encountered?

david-pepper: I've simulated an error in the query, and the process exits quietly rather than throw an error. Is there a way to decorate the decode call to provide more information in this scenario?

handlers/sf-billing-account-remover/src/main/scala/com/gu/sf_billing_account_remover/BillingAccountRemover.scala
david-pepper: This is done but curious to know if we can remove this method altogether in favour of a one liner that will still maintain error throwing capabilities

handlers/sf-billing-account-remover/src/main/scala/com/gu/sf_billing_account_remover/BillingAccountRemover.scala
david-pepper: Good shout. This is done.

handlers/sf-billing-account-remover/src/main/scala/com/gu/sf_billing_account_remover/BillingAccountRemover.scala
david-pepper: Done

handlers/sf-billing-account-remover/src/main/scala/com/gu/sf_billing_account_remover/BillingAccountRemover.scala
david-pepper: Agreed, I've updated the code to use only one method for SF GET to take query as a parameter

handlers/sf-billing-account-remover/build.sbt
david-pepper: well spotted. I've removed this

handlers/sf-billing-account-remover/src/main/scala/com/gu/sf_billing_account_remover/BillingAccountRemover.scala
david-pepper: Yep I was going to ask about the best way to do this so thanks for preempting that. I've updated the code to add a zuoraInstanceUrl property to ZuoraConfig

handlers/sf-billing-account-remover/src/main/scala/com/gu/sf_billing_account_remover/BillingAccountRemover.scala
mario-galic: Maybe something like 

```
def writeFailuresToSf(failedUpdates: Seq[BillingAccountsRecords.Records]): Either[Throwable, String] = {
  for {
    config <- optConfig.toRight(new RuntimeException("Missing config value"))
    sfAuthDetails <- SfLogin(config.salesforceConfig)
    _ <- updateBillingAccountsInSf(sfAuthDetails, failedUpdates)
    result <- insertErrorRecordsInSf(sfAuthDetails, failedUpdates)
  } yield result
}

def processBillingAccounts(): Either[Exception, Seq[BillingAccountsRecords.Records]] = {
  (for {
    config <- optConfig.toRight(new RuntimeException("Missing config value"))
    sfAuthDetails <- SfLogin(config.salesforceConfig)
    getCustomSettingResponse <- getSfCustomSetting(sfAuthDetails)
    maxAttempts = getCustomSettingResponse.records.headOption.getOrElse(throw new RuntimeException("There should be at least one record")).Property_Value__c
    getBillingAccountsResponse <- decode[SfGetBillingAccsResponse](getBillingAccountsFromSf(maxAttempts, sfAuthDetails))
  } yield {
    updateRecordsInZuora(getBillingAccountsResponse.records)
  })
}

processBillingAccounts() match {
  case Left(error) => throw error
  case Right(allUpdates) =>
    val failedUpdates = allUpdates.filter(_.ErrorCode.isDefined)
    if (failedUpdates.nonEmpty) {
      writeFailuresToSf(failedUpdates).getOrElse(throw new RuntimeException("Failed to notify Salesforce of failed updates to Zuora"))
      throw new RuntimeException(s"Failed to write updates to zuora: $failedUpdates")
    } else {
      "all ok"
    }
}
```

The key is to understand that `Left` is just a regular value (like say `42`) so if the whole program returns `Left` and we did not log an error or converted left to a thrown exception, then we have what I call *silent error*. Consider the difference between 

```
object A extends App {
  val e: Either[Throwable, Int] = Left(new RuntimeException("boom"))
  e
}
```

and 

```
object B extends App {
  throw new RuntimeException("boom")
}
```
Try to run both programs and figure out why A completes as if all is ok.

handlers/sf-billing-account-remover/src/main/scala/com/gu/sf_billing_account_remover/BillingAccountRemover.scala
david-pepper: I've tried this, but it's still giving issues:

image

david-pepper: > You would have to replace all method signatures from `Either[A, B]` to `Try[B]`. Indeed `Try` is standard Scala so `import scala.util.Try`.

Why do we have to replace all method signatures for only one method signature to work? Can the class not have method A with a Try[A] return type and method B with a Either [B, A] return type? Why are the return types of 2 or more methods interferring with the compilability of the program?

mario-galic: Because `processBillingAccounts` (which seems to me main business logic of the program) is using a for-comprehension, and the types of **all** the steps in the for-comprehension chain have to "align". For example, the following will not work

```
val aOpt: Option[Int] = Some(41)
val bEither: Either[String, Int] = Right(1)

for {
  a <- aOpt
  b <- bEither  // nope, it needs to be Option :(
} yield a + b
```

The type of the overall for-comprehension is determined by the **first** step in the chain, which is in above case `a <- aOpt`. 

Maybe just proceed with `Either` in this PR, you could always re-factor it in subsequent PRs if you want.

david-pepper: Ok will keep the Either return types

handlers/sf-billing-account-remover/src/main/scala/com/gu/sf_billing_account_remover/BillingAccountRemover.scala
```diff
@@ -0,0 +1,312 @@
...
+      .asString
+      .body
+  }
+
+  def writeErrorsBackToSf(
+    sfAuthDetails: SfAuthDetails,
+    failedUpdates: Seq[BillingAccountsRecords.Records]
+  ): Unit = {
+
+    updateBillingAccountsInSf(sfAuthDetails, failedUpdates)
```
mario-galic, line 252: Quick way to avoid silent error would be

```
(for {
  _ <- updateBillingAccountsInSf(sfAuthDetails, failedUpdates)
  _ <- insertErrorRecordsInSf(sfAuthDetails, failedUpdates)
} yield ()).left.foreach(e => throw e)
```

handlers/sf-billing-account-remover/src/main/scala/com/gu/sf_billing_account_remover/BillingAccountRemover.scala
```diff
@@ -0,0 +1,312 @@
...
+        apiAccessKeyId = zuoraApiAccessKeyId,
+        apiSecretAccessKey = zuoraApiSecretAccessKey,
+        zuoraInstanceUrl = zuoraInstanceUrl
+      )
+    )
+
+  processBillingAccounts()
+
+  def processBillingAccounts(): Unit = {
+    for {
```
mario-galic, line 85: Quick way to avoid silent error 

```
    (for {
      config <- optConfig.toRight(new RuntimeException("Missing config value"))
      ...
    } yield {
      ...
      if (failedUpdates.nonEmpty) {
        writeErrorsBackToSf(sfAuthDetails, failedUpdates)
      }
    }).left.foreach(e => throw e)
  }
```

handlers/sf-billing-account-remover/src/main/scala/com/gu/sf_billing_account_remover/BillingAccountRemover.scala
david-pepper: Great, this is done

mario-galic (approved):
Well done! This is rather impressive for first Scala PR. Also excelent PR description! 

Note currently this is stand alone app. I guess in subsequent PRs it will become a sub-project lambda with dedicated AWS CloudFormation.

PR #700, opened 2020-09-15
DEV env cleaner to prevent SF dev filling up

johnduffell:
TODO: add CD to riffraff for main branch

We only have 200MB storage in SF DEV, which is filled up very quickly by the Z360 sync in conjunction with any tests we run against DEV zuora.

Now that we aim to run the IT tests every hour on a schedule, this is filling up extremely quickly.

The best way to clean things up is to cancel both the subscription and the account in Zuora.

This PR adds a cleanup lambda which run on a daily basis to do the above steps on DEV zuora.

There will also be an alarm so that we know that it has stopped working.

Comments: I noticed the zuora reports library was containing a fair bit of lambda specific stuff, which I had to work around in places.  Ideally the functionality would be separated a bit better, but I managed to put in dummy values where necessary.  I could have split it out a bit more in this PR, at the risk of breaking things in the existing zuora reports step functions.


handlers/dev-env-cleaner/build.sbt
```diff
@@ -0,0 +1,17 @@
+import Dependencies._
```
kelvin-chappell, line 1: Is this being used?

handlers/dev-env-cleaner/src/main/resources/logback.xml
```diff
@@ -0,0 +1,11 @@
...
+    <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
+        <encoder>
+            <pattern>%date %level - %message%n%xException</pattern>
+        </encoder>
+    </appender>
+
+    <root level="INFO">
+        <appender-ref ref="STDOUT" />
+    </root>
+</configuration>
```
kelvin-chappell, line 11: Is logback being used anywhere?

handlers/dev-env-cleaner/src/main/scala/com/gu/cleaner/Handler.scala
```diff
@@ -0,0 +1,231 @@
...
+import com.gu.util.zuora.{ZuoraRestConfig, ZuoraRestRequestMaker}
+import com.gu.zuora.reports._
+import com.gu.zuora.reports.aqua.{AquaJobResponse, AquaQuery, AquaQueryRequest, ZuoraAquaRequestMaker}
+import com.gu.zuora.reports.dataModel.Batch
+import play.api.libs.json.{JsSuccess, Json, Reads}
+
+import scala.annotation.tailrec
+import scala.io.Source
+
+object Handler {
```
kelvin-chappell, line 25: Doesn't the handler have to implement `RequestHandler`?

kelvin-chappell (approved):
Good idea 👍

handlers/dev-env-cleaner/src/main/scala/com/gu/cleaner/Handler.scala
johnduffell: no that's just a convenience to make sure the signature is correct, actually it is triggered by name/reflection ( a bit like the main method)
I can add it if it's the convention though?

It all does work fine in CODE.

handlers/dev-env-cleaner/src/main/resources/logback.xml
johnduffell: not sure, I just copied this from another lambda!  I will delete.
I always get confused about the amount of logging libraries in Java+Scala
image

handlers/dev-env-cleaner/build.sbt
johnduffell: good point I will remove

johnduffell: actually it's used for `assemblyMergeStrategyDiscardModuleInfo` and `awsVersion`

handlers/dev-env-cleaner/src/main/scala/com/gu/cleaner/Handler.scala
kelvin-chappell: It might be useful - I can just imagine someone removing one of the parameters because it isn't used and not finding out it's necessary until they deploy.

johnduffell: fair point, I've added it (seemed to be actually `RequestStreamHandler`)

PR #702, opened 2020-09-21
allow 15 minutes for dev env cleaner to run

johnduffell:
following #700 it turns out that now we have changed the it tests to run hourly, it takes more than 5 minutes for the dev env cleaner to clean all the subscriptions.

`2020-09-20T16:26:05.189Z 7543b93e-3d93-4752-97b4-ad47d96d67ea Task timed out after 300.10 seconds`

We could modify it to do some of them in parallel, but thatmight overload zuora, so I decided to try extending the timeout for now.


tomrf1 (approved):

johnduffell:
this is working again now taking 591127.12 ms i.e. just under 10 minutes.  It should be quicker normally, as it's had to catch up a few days of data.
image

PR #703, opened 2020-09-21
Initial contact-us-api PR

ripecosta:
## What does this change?
This is the first PR for the contact-us-api (Trello card [here](https://trello.com/c/WbynGBtC/1501-contact-form-create-new-lambda-for-contact-form-server-side)). The goal of lambda is to receive requests from the Contact Us form, validate them, and create Cases in Salesforce.
For now it's an app that reads the JSON of a contact-us form request and creates a case for it in our Salesforce dev instance. Further PRs will convert it a deployment-ready lambda.

This is the JSON it accepts:
```
{
    "topic": string,
    "subtopic"?: string, // Optional because not all topics have subtopics
    "subsubtopic"?: string, // Optional because not all subtopics have subsubtopics
    "name": string,
    "email": string,
    "subject": string,
    "message": string,
    "attachment"?: { // Attachments are optional
        "name": string,
        "contents": base64 encoded string, // File contents
    }
}
```

And this is the JSON it outputs:
```
{
    "success": boolean,
    "error"?: string, // Error message only present if success is false
}
```

## Further Work
- Turn it into a lambda
- Add CloudFormation
- Read config from AWS
- Adapt build steps to allow for PROD and DEV deployment
- Add riffraff build step
- Expand testing
- Add logging and error handling
- Add request validation (user input and captcha)


handlers/contact-us-api/src/main/scala/com/gu/contact_us_api/SalesforceConnector.scala
```diff
@@ -0,0 +1,69 @@
...
+
+import com.gu.contact_us_api.models.{ContactUsRequest, SFAuthFailure, SFAuthSuccess, SFCompositeRequest, SFCompositeResponse, SFErrorDetails}
+import io.circe.parser._
+import io.circe.generic.auto._
+import io.circe.syntax._
+import scalaj.http.Http
+
+object SalesforceConnector {
+  private val authEndpoint = "https://test.salesforce.com/services/oauth2/token"
+  private val reqEndpoint = "https://gnmtouchpoint--DEV.my.salesforce.com/services/data/v43.0/composite/"
```
kelvin-chappell, line 11: Stage-specific so should come through as env vars.

handlers/contact-us-api/src/main/scala/com/gu/contact_us_api/SalesforceConnector.scala
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  def sendReq(token: String, request: SFCompositeRequest): Either[Throwable, Unit] = {
+    val response = Http(reqEndpoint)
+      .header("Content-Type", "application/json")
+      .header("Authorization", s"Bearer $token")
+      .postData(request.asJson.toString())
+      .asString
+
+    println(response.body)
```
kelvin-chappell, line 50: Should be some kind of log statement.

handlers/contact-us-api/src/main/scala/com/gu/contact_us_api/SalesforceConnector.scala
```diff
@@ -0,0 +1,69 @@
...
+    if (response.isSuccess) {
+      decode[SFCompositeResponse](response.body) match {
+        case Left(value) => Left(value)
+        case Right(compositeResponse) => {
+          if(compositeResponse.isSuccess) Right(())
+          else Left(new Throwable(s"Could not complete composite request. Status code: ${response.code}. ${compositeResponse.errorsAsString.getOrElse("")}"))
+        }
+      }
+    } else {
```
kelvin-chappell, line 59:
```suggestion
      decode[SFCompositeResponse](response.body) flatMap {
        compositeResponse => {
          if(compositeResponse.isSuccess) Right(())
          else Left(new Throwable(s"Could not complete composite request. Status code: ${response.code}. ${compositeResponse.errorsAsString.getOrElse("")}"))
        }
      }
    } else {
```

handlers/contact-us-api/src/main/scala/com/gu/contact_us_api/SalesforceConnector.scala
```diff
@@ -0,0 +1,69 @@
...
+      decode[List[SFErrorDetails]](response.body) match {
+        case Left(value) => Left(value)
+        case Right(errors) => {
+          val errorDetails = errors.map(error => error.asString).mkString(", ")
+          Left(new Throwable(s"Could not complete request. Status code: ${response.code}. ${errorDetails}"))
+        }
+      }
```
kelvin-chappell, line 66:
```suggestion
      decode[List[SFErrorDetails]](response.body) flatMap {
        errors => 
          val errorDetails = errors.map(error => error.asString).mkString(", ")
          Left(new Throwable(s"Could not complete request. Status code: ${response.code}. ${errorDetails}"))
      }
```

handlers/contact-us-api/src/main/scala/com/gu/contact_us_api/contactUsApp.scala
```diff
@@ -0,0 +1,26 @@
...
+        "attachment": {
+          "name": "printscreen.jpeg",
+          "contents": "junkfornow"
+        }
+      }
+    """
+
+    println(ContactUs.processReq(json).asJson)
+  }
+}
```
kelvin-chappell, line 26: Should be in a test?

handlers/contact-us-api/src/main/scala/com/gu/contact_us_api/models/AttachmentDetails.scala
```diff
@@ -0,0 +1,5 @@
+package com.gu.contact_us_api.models
+
+case class AttachmentDetails(name: String, contents: String) {
+  val asSFRequestItem: SFRequestItem = SFAttachmentRequest(name, contents)
```
kelvin-chappell, line 4: Should be a method of the `SFRequestItem` companion object?

handlers/contact-us-api/src/main/scala/com/gu/contact_us_api/models/ContactUsRequest.scala
```diff
@@ -0,0 +1,13 @@
...
+case class ContactUsRequest(topic: String, subtopic: Option[String], subsubtopic: Option[String], name: String, email: String, subject: String, message: String, attachment: Option[AttachmentDetails]) {
+  val isValid: Boolean = true
+
+  val asSFCompositeRequest: SFCompositeRequest = {
+    val requestList = List(
+      SFCaseRequest(topic, subtopic, subsubtopic, name, email, subject, message)
+    ) ++ attachment.map(i => i.asSFRequestItem)
+
+    SFCompositeRequest(requestList)
+  }
```
kelvin-chappell, line 12: Should be in the `SFCompositeRequest` companion object?

handlers/contact-us-api/build.sbt
```diff
@@ -0,0 +1,19 @@
+import Dependencies._
+
+ThisBuild / scalaVersion     := "2.13.2"
+ThisBuild / version          := "0.1.0-SNAPSHOT"
+ThisBuild / organization     := "com.gu"
+ThisBuild / organizationName := "gu"
```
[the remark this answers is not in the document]
ripecosta: I managed to get this to work using the way the repo builds things now so this will be gone in next commit.

handlers/contact-us-api/src/main/scala/com/gu/contact_us_api/SalesforceConnector.scala
ripecosta: Left over from debugging. Removed.

handlers/contact-us-api/src/main/scala/com/gu/contact_us_api/SalesforceConnector.scala
ripecosta: I knew there's a way to simplify that. Thanks!

handlers/contact-us-api/src/main/scala/com/gu/contact_us_api/SalesforceConnector.scala
ripecosta: Will move to the config object for now and will make it an env variable in a following PR.

handlers/contact-us-api/src/main/scala/com/gu/contact_us_api/ContactUsConfig.scala
```diff
@@ -0,0 +1,10 @@
+package com.gu.contact_us_api
+
+object ContactUsConfig {
+  // TODO: Obtain these from AWS instead
```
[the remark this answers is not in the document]
ripecosta: 👍

handlers/contact-us-api/src/main/scala/com/gu/contact_us_api/ContactUsConfig.scala
```diff
@@ -0,0 +1,10 @@
...
+  val clientID: String = System.getenv("CUClientID")
+  val clientSecret: String = System.getenv("CUClientSecret")
+  val username: String = System.getenv("CUUsername")
+  val password: String = System.getenv("CUPassword")
+  val token: String = System.getenv("CUToken")
```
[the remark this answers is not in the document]
ripecosta: 👍 Will make this change in a separate PR when I'm focusing on env variables.

handlers/contact-us-api/src/main/scala/com/gu/contact_us_api/contactUsApp.scala
ripecosta: This is just a placeholder that also made it easier to debug. Will be replaced completely gone in the next PR when this is turned into a proper lambda.

handlers/contact-us-api/src/main/scala/com/gu/contact_us_api/models/AttachmentDetails.scala
ripecosta: How so?

handlers/contact-us-api/src/main/scala/com/gu/contact_us_api/models/SFRequestItem.scala
```diff
@@ -0,0 +1,70 @@
...
+}
+
+object SFAttachmentRequest {
+  val method = "POST"
+  val url = "/services/data/v20.0/sobjects/Attachment/"
+  val referenceId = "newAttachment"
+
+  implicit val encodeSFAttachmentRequest: Encoder[SFAttachmentRequest] = new Encoder[SFAttachmentRequest] {
+    final def apply(a: SFAttachmentRequest): Json = Json.obj(
+      ("method", Json.fromString(method)),
```
mario-galic, line 60: Circe provides automatic encoding, there is no need for hand-carving JSON.

handlers/contact-us-api/src/main/scala/com/gu/contact_us_api/models/SFResponse.scala
```diff
@@ -0,0 +1,30 @@
...
+
+case class SFSuccessResponse(httpStatusCode: Int, referenceId: String) extends SFResponse {
+  override def isSuccess: Boolean = true
+  override def errorsAsString: Option[String] = None
+}
+
+object SFResponse {
+  implicit val decodeSFResponse: Decoder[SFResponse] = {
+    List[Decoder[SFResponse]](
+      Decoder[SFErrorResponse].asInstanceOf[Decoder[SFResponse]],
```
mario-galic, line 26: `asInstanceOf` should not be necessary in Scala code. It switches off compiler protections. I think circe has `widen` for ADT decoding.

handlers/contact-us-api/src/test/scala/com/gu/contact_us_api/models/ContactUsResponseTests.scala
```diff
@@ -0,0 +1,20 @@
...
+import org.scalatest.flatspec.AnyFlatSpec
+import org.scalatest.matchers.should
+
+class ContactUsResponseTests extends AnyFlatSpec with should.Matchers {
+  val errorMsg = "error"
+  val successJson = Json.obj(("success", Json.fromBoolean(true)))
+  val failureJson = Json.obj(("success", Json.fromBoolean(false)), ("error", Json.fromString(errorMsg)))
+
+  "ContactUsSuccessfulResponse" should "encode into expected json object" in {
+    ContactUsSuccessfulResponse().asInstanceOf[ContactUsResponse].asJson shouldBe successJson
```
mario-galic, line 14: `asInstanceOf` is not necessary.

handlers/contact-us-api/src/main/scala/com/gu/contact_us_api/contactUsApp.scala
```diff
@@ -0,0 +1,31 @@
...
+        "attachment": {
+          "name": "printscreen.jpeg",
+          "contents": "junkfornow"
+        }
+      }
+    """
+
+    val contactUsReqHandler = new ContactUs(new SalesforceConnector())
+
+    println(contactUsReqHandler.processReq(json).asJson)
```
mario-galic, line 29: Make sure there are no silent errors. Sometimes this happens when errors are being modelled as pure values instead of exceptions.

handlers/contact-us-api/src/main/scala/com/gu/contact_us_api/models/ContactUsResponse.scala
```diff
@@ -0,0 +1,49 @@
...
+
+case class ContactUsSuccessfulResponse() extends ContactUsResponse {
+  override val success: Boolean = true
+}
+
+case class ContactUsFailureResponse(message: String) extends ContactUsResponse {
+  override val success: Boolean = false
+}
+
+// What's the relationship between this val's name and circe knowing it's the decoder?
```
mario-galic, line 17: The pattern is called **type class**. Scala implements them using `implicit` mechanisms. It is an alternative to object-oriented design techniques which tend to use subtype polymorphism. Confusingly, Scala 2 uses implicits to implement three different things 

- type classes
- extension methods
- implicit conversions (discouraged)

In Scala 3, this is made clearer where type classes and extension methods get dedicated different syntax.

handlers/contact-us-api/src/main/scala/com/gu/contact_us_api/models/ContactUsResponse.scala
```diff
@@ -0,0 +1,49 @@
...
+// Why does intelliJ want to simplify this by removing the apply?
```
mario-galic, line 18: When a trait has a **single** undefined method, then definition can be simplified a bit.

handlers/contact-us-api/src/main/scala/com/gu/contact_us_api/models/ContactUsResponse.scala
```diff
@@ -0,0 +1,49 @@
+package com.gu.contact_us_api.models
+
+import io.circe.{Encoder, Json}
+
+trait ContactUsResponse {
```
mario-galic, line 5: If the intention is to create ADT (algebraic data type) then this should be `sealed` trait.

handlers/contact-us-api/src/main/scala/com/gu/contact_us_api/models/ContactUsResponse.scala
```diff
@@ -0,0 +1,49 @@
...
+  val success: Boolean
+}
+
+case class ContactUsSuccessfulResponse() extends ContactUsResponse {
```
mario-galic, line 9: Empty case classes tend to be discouraged. Scala has `case object` for that purpose.

handlers/contact-us-api/src/main/scala/com/gu/contact_us_api/models/ContactUsResponse.scala
```diff
@@ -0,0 +1,49 @@
...
+  override val success: Boolean = false
+}
+
+// What's the relationship between this val's name and circe knowing it's the decoder?
+// Why does intelliJ want to simplify this by removing the apply?
+object ContactUsResponse {
+  implicit val encodeContactUsResponse: Encoder[ContactUsResponse] = new Encoder[ContactUsResponse] {
+    final def apply(a: ContactUsResponse): Json = {
+      a match {
+        case a: ContactUsSuccessfulResponse => ContactUsSuccessfulResponse.encodeContactUsSuccessfulResponse(a)
```
mario-galic, line 23: I think encoding ATDs in circe is something like

```
case a: ContactUsSuccessfulResponse => a.asJson
```

handlers/contact-us-api/src/main/scala/com/gu/contact_us_api/models/SFRequestItem.scala
ripecosta: Yes, but in this case the json I need to output has 1) different field names, and 2) added fields (such as method and url). Is there a way for circe to automatically encode that?

handlers/contact-us-api/src/main/scala/com/gu/contact_us_api/models/ContactUsResponse.scala
ripecosta: 👍

handlers/contact-us-api/src/main/scala/com/gu/contact_us_api/models/ContactUsResponse.scala
ripecosta: 👍

handlers/contact-us-api/src/main/scala/com/gu/contact_us_api/models/ContactUsResponse.scala
ripecosta: 👍

handlers/contact-us-api/src/main/scala/com/gu/contact_us_api/models/ContactUsResponse.scala
ripecosta: 👍

handlers/contact-us-api/src/main/scala/com/gu/contact_us_api/models/SFResponse.scala
ripecosta: They use Functor's widen in their example. I'm using that instead now.

handlers/contact-us-api/src/test/scala/com/gu/contact_us_api/models/ContactUsResponseTests.scala
ripecosta: 👍

handlers/contact-us-api/src/main/scala/com/gu/contact_us_api/models/ContactUsResponse.scala
ripecosta: What about the `success` val that it overrides? And wouldn't an object lose the ability to be encoded using circe?

handlers/contact-us-api/src/main/scala/com/gu/contact_us_api/contactUsApp.scala
ripecosta: I'm going to handle that when I introduce logging and error handling. If there is an error I still need to send a response to the client to report it so I can't just throw them. My plan is to handle the error just before I convert it into a Success/Failure response in `ContactUs.processReq`, which is why I'm returning `Either` up to that point.

handlers/contact-us-api/src/main/scala/com/gu/contact_us_api/models/SFRequestItem.scala
mario-galic: If you do not want to keep one-to-one mapping between case classes and json, then option is to have different case classes, where the target case class does have one-to-one mapping, and transform between them at "higher" level before json encoding takes place.

handlers/contact-us-api/src/main/scala/com/gu/contact_us_api/models/AttachmentDetails.scala
kelvin-chappell: Just that generally constructors should be in the companion object of the class they're constructing.  But that's really a matter of personal choice

handlers/contact-us-api/src/main/scala/com/gu/contact_us_api/models/ContactUsResponse.scala
```diff
@@ -0,0 +1,48 @@
...
+import io.circe.{Encoder, Json}
+import io.circe.syntax._
+
+sealed trait ContactUsResponse {
+  val success: Boolean
+}
+
+case class ContactUsSuccessfulResponse() extends ContactUsResponse {
+  override val success: Boolean = true
+}
```
kelvin-chappell, line 12:
```suggestion
case class ContactUsSuccessfulResponse(success: Boolean) extends ContactUsResponse
```

handlers/contact-us-api/src/main/scala/com/gu/contact_us_api/models/ContactUsResponse.scala
```diff
@@ -0,0 +1,48 @@
...
+
+case class ContactUsFailureResponse(message: String) extends ContactUsResponse {
+  override val success: Boolean = false
+}
```
kelvin-chappell, line 16:
```suggestion
case class ContactUsFailureResponse(success:Boolean, message: String) extends ContactUsResponse
```

handlers/contact-us-api/src/main/scala/com/gu/contact_us_api/models/ContactUsResponse.scala
```diff
@@ -0,0 +1,48 @@
...
+
+object ContactUsResponse {
+  implicit val encodeContactUsResponse: Encoder[ContactUsResponse] = new Encoder[ContactUsResponse] {
+    final def apply(a: ContactUsResponse): Json = {
+      a match {
+        case a: ContactUsSuccessfulResponse => a.asJson
+        case a: ContactUsFailureResponse => a.asJson
+      }
+    }
```
kelvin-chappell, line 25: If `ContactUsResponse` was instead an `Either[A, B]` of some type you could fold over it like:
```
response.fold(_.asJson, _.asJson)
```

handlers/contact-us-api/src/main/scala/com/gu/contact_us_api/models/ContactUsResponse.scala
ripecosta: This would allow the instancing of `ContactUsSuccessfulResponse` where `success` equals false, which something I don't want to do. Why is the original solution not ok?

handlers/contact-us-api/src/main/scala/com/gu/contact_us_api/models/ContactUsResponse.scala
ripecosta: This would allow the instancing of `ContactUsFailureResponse ` where `success` equals true, which something I don't want to do. Why is the original solution not ok?

handlers/contact-us-api/src/main/scala/com/gu/contact_us_api/models/ContactUsResponse.scala
kelvin-chappell: In that case would be better as a case object or just convert the `ContactUsResponse` trait into a case class and hold the success value there.

handlers/contact-us-api/src/main/scala/com/gu/contact_us_api/models/ContactUsResponse.scala
kelvin-chappell: In that case would be better as a case object or just convert the ContactUsResponse trait into a case class and hold the success value there.

handlers/contact-us-api/src/main/scala/com/gu/contact_us_api/models/ContactUsResponse.scala
kelvin-chappell: A bit petty.  Ignore.

kelvin-chappell (approved):
Great work @ripecosta!
You can merge this now and then convert it into a lambda in another PR.

PR #705, opened 2020-09-22
better error message for dev env cleaner failures

johnduffell:
The dev env cleaner tries to clean all the subscriptions one by one first, and stores the success/fail status in a list.
Then it checks the list for any failures, throwing an exception if there are any.

This means that if there is a failure, you might have to scroll up quite a bit to find it, and it was not clear from the message.

This PR makes the message a bit clearer so we are more likely to find the problem without having to debug.


i-hardy (approved):
👍

PR #708, opened 2020-09-29
Create subproject for sf billing account remover and add cloudformation and riffraff integration

david-pepper:
Change Description

- convert the standalone project in support-service-lambdas to a subproject
- add cloudformation and riffraff integration


handlers/sf-billing-account-remover/build.sbt
```diff
@@ -1,6 +1,16 @@
-version := "0.1"
+//import Dependencies._
```
mario-galic, line 1: You can re-enabled this and then replace bellow your `libraryDependencies` with something like

```
libraryDependencies ++= Seq(
  circe,
  circeParsers,
  scalajHttp
)
```

where we use variables from `Dependencies.scala`

handlers/sf-billing-account-remover/build.sbt
```diff
@@ -1,6 +1,16 @@
...
+description := "Export salesforce data to the data lake"
+
+assemblyJarName := "sf-billing-account-remover.jar"
+riffRaffPackageType := assembly.value
+riffRaffUploadArtifactBucket := Option("riffraff-artifact")
+riffRaffUploadManifestBucket := Option("riffraff-builds")
+riffRaffManifestProjectName := "MemSub::Membership Admin::SF Billing Account Remover"
+riffRaffArtifactResources += (file(
+  "handlers/sf-billing-account-remover/cfn.yaml"
+), "cfn/cfn.yaml")
```
mario-galic, line 14: Don't forget to add `assemblyMergeStrategyDiscardModuleInfo` to your `build.sbt`

handlers/sf-billing-account-remover/build.sbt
david-pepper: This is done and is compiling as expected

handlers/sf-billing-account-remover/build.sbt
david-pepper: Done

mario-galic (approved):

handlers/sf-billing-account-remover/cfn.yaml
```diff
@@ -0,0 +1,88 @@
...
+      FunctionName:
+        !Sub sf-billing-account-remover-${Stage}
+      Description: Remove Billing Accounts and related records from Salesforce (via Zuora)
+      Code:
+        S3Bucket: support-service-lambdas-dist
+        S3Key: !Sub membership/${Stage}/sf-billing-account-remover/sf-billing-account-remover.jar
+      Handler: com.gu.sf_billing_account_remover.BillingAccountRemover::lambda
+      Environment:
+        Variables:
+          Stage: !Ref Stage
```
mario-galic, line 81: You will have to link the `Parameters` specified at the top of CF to here under`Variables`, for example, `System.getenv("zuoraInstanceUrl")` would correspond to

```
Variables:
  Stage: !Ref Stage
  zuoraInstanceUrl: !Ref zuoraInstanceUrl
  ...
```

Also this can be simplified (maybe in some future PR) to replace environmental variables with reading secrets from a file in S3. For example, here is a snippet that could do that:

```
  def readConfig(): Config = {
    val stage = System.getenv("Stage")
    val bucketName = "gu-reader-revenue-private"
    val key = s"membership/support-service-lambdas/$stage/sfAuth-$stage.v1.json"
    val inputStream = AmazonS3Client.builder.build().getObject(bucketName, key).getObjectContent
    val rawJson = Source.fromInputStream(inputStream).mkString
    decode[Config](rawJson)
      .getOrElse(throw new RuntimeException(s"Could not read secret config file from S3://$bucketName/$key"))
  }
```

which would correspond to security policy 

```
Statement:
  - Effect: Allow
    Action: s3:GetObject
    Resource: !Sub arn:aws:s3:::gu-reader-revenue-private/membership/support-service- lambdas/${Stage}/sfAuth-${Stage}*.json
```

mario-galic (approved):

PR #711, opened 2020-09-30
clean more dev test data, also increase frequency

johnduffell:
From inspecting the subs created in DEV zuora after an IT test run, I found there are also test@gu.com being created.

This PR adds that address to the cleaner, so it will clean more thoroughly.

I have also upped it from once a day to twice.  This is becasue the alarm period is 1 day, and we want it to at least have two goes, to avoid false alarms.


kelvin-chappell (approved):
👍

PR #712, opened 2020-09-30
Bump zuora export Lambda memory

mario-galic:
## What does this change?

Exporting 10 days worth of data OOMed.

## How to test

Just run the lambda with 10 days ago `yyyy-mm-dd`


kelvin-chappell (approved):
Curious number!

mario-galic:
That is what they specify as a limit here https://docs.aws.amazon.com/lambda/latest/dg/gettingstarted-limits.html

PR #713, opened 2020-09-30
Fetch digital voucher suspensions from Salesforce

kelvin-chappell:
This creates a new subproject for sending digital voucher suspensions from Salesforce to the Digital Voucher API.  The project has a lambda that will be run on an hourly schedule and will fetch any future suspensions that have already been processed in Zuora.  The effect is to ensure the vouchers can't be used and won't be charged for during the suspension period. 

This PR just does the fetching from Salesforce.  The sending to the Digital Voucher API will be coming next.


handlers/digital-voucher-suspension-processor/src/main/scala/com/gu/digitalvouchersuspensionprocessor/Handler.scala
```diff
@@ -0,0 +1,37 @@
...
+import cats.arrow.FunctionK
+import cats.data.EitherT
+import cats.effect.IO
+import cats.implicits._
+import com.amazonaws.services.lambda.runtime.{Context, RequestHandler}
+import com.softwaremill.sttp.impl.cats.implicits._
+import com.softwaremill.sttp.{HttpURLConnectionBackend, Id, SttpBackend}
+import com.typesafe.scalalogging.LazyLogging
+
+object Handler extends RequestHandler[Map[String, String], String] with LazyLogging {
```
mario-galic, line 12: Do you need `RequestHandler`? Is the lambda going to actually receive any input?

handlers/digital-voucher-suspension-processor/src/main/scala/com/gu/digitalvouchersuspensionprocessor/Salesforce.scala
```diff
@@ -0,0 +1,55 @@
...
+object Salesforce {
+
+  case class Suspension(
+    subscriptionNumber: String,
+    sfSubscriptionId: String,
+    suspendedPublicationDate: LocalDate
+  )
+
+  object Suspension {
+    implicit val decoder: Decoder[Suspension] = cursor => {
```
mario-galic, line 21: Why custom codec?

kelvin-chappell: It's quite a simple decoding job, it doesn't seem to be more code than building an artificial hierarchy of case classes and it leaves me with a case class with valid Scala field names.

handlers/digital-voucher-suspension-processor/src/main/scala/com/gu/digitalvouchersuspensionprocessor/Config.scala
```diff
@@ -0,0 +1,31 @@
+package com.gu.digitalvouchersuspensionprocessor
+
+import com.gu.salesforce.SFAuthConfig
+
+case class Config(salesforce: SFAuthConfig)
+
+object Config {
+
+  def fromEnv(): Either[ConfigFailure, Config] = {
```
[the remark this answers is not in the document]
kelvin-chappell: So I can run it in Intellij without getting config from AWS.

handlers/digital-voucher-suspension-processor/src/main/scala/com/gu/digitalvouchersuspensionprocessor/Handler.scala
kelvin-chappell: No.  I think however I do it I have to have a method that expects some sort of input.  Have you got an example where I can avoid an input argument?

mario-galic: If no input is required then all that is necessary is

```
object Handler {
  def handleRequest() = ???
}
```

handlers/digital-voucher-suspension-processor/src/main/scala/com/gu/digitalvouchersuspensionprocessor/Salesforce.scala
mario-galic: Note sure what you mean. It is doubling the amount of code.

mario-galic (approved):

handlers/digital-voucher-suspension-processor/src/main/scala/com/gu/digitalvouchersuspensionprocessor/Config.scala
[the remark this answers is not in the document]
mario-galic: Do you mean from param store?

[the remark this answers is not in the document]
mario-galic: For running from IJ you could just dump the whole json from secrets file into a single env var and then add it to the Run Configuration, something like so

```
decode[Config](getenv("Config"))
```

where in RC 

```
![image](https://user-images.githubusercontent.com/13835317/94702495-bd5beb80-0335-11eb-9767-4c5cca50513e.png)

```

handlers/digital-voucher-suspension-processor/src/main/scala/com/gu/digitalvouchersuspensionprocessor/Salesforce.scala
kelvin-chappell: We really need a generic Salesforce decoder/encoder that does snakecase and then does something with the trailing '__c'.
Something like:
`SF_Subscription__c` <-> `CustomSFSubscription`
although the recognition of custom fields is a bit clumsy.

handlers/digital-voucher-suspension-processor/src/main/scala/com/gu/digitalvouchersuspensionprocessor/Config.scala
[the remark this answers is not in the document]
kelvin-chappell: It will fetch from secrets manager into the cloudformation.  Like [this](https://github.com/guardian/price-migration-engine/blob/main/lambda/cfn.yaml#L400-L414).

handlers/digital-voucher-suspension-processor/src/main/scala/com/gu/digitalvouchersuspensionprocessor/Handler.scala
kelvin-chappell: 👍

handlers/digital-voucher-suspension-processor/src/main/scala/com/gu/digitalvouchersuspensionprocessor/Salesforce.scala
kelvin-chappell: Will leave for now and possibly revisit.

PR #707, opened 2020-09-29
Convert Contact Us App into a Lambda

ripecosta:
## What does this change?
Converts the Contact Us API (PR #703) into a lambda ready for manual deployment. Adds Riff Raff config.

This is the JSON it expects:
```
{
    "topic": string,
    "subtopic"?: string, // Optional because not all topics have subtopics
    "subsubtopic"?: string, // Optional because not all subtopics have subsubtopics
    "name": string,
    "email": string,
    "subject": string,
    "message": string,
    "attachment"?: { // Attachments are optional
        "name": string,
        "contents": base64 encoded string, // File contents
    }
}
```

And this is the JSON it returns:
```
{
    "success": boolean,
    "error"?: string, // Error message only present if success is false
}
```

## Further Work
- Add CloudFormation
- Read config from AWS
- Expand testing
- Add request validation
- Final refactor for that sleek and polished look

PR #714, opened 2020-09-30
Refactor contact-us-api

ripecosta:
## What does this change?
Refactors the contact-us-api, continuing on the work of PRs #703 and #707.
- Improves error handling by considering more edge cases
- Improves logging clarity by adding better log messages
- Improves function purity by bubbling errors to the higher level before logging them
- Improves the responses of the lambda by not including error details in the response and by using more accurate status codes
- Improves how it reads environment variables by not assuming that they exist
- Simplifies parts of the code
- Small style and formatting improvements

## Further Work
- Add CloudFormation
- Expand testing


handlers/contact-us-api/src/main/scala/com/gu/contact_us_api/ContactUs.scala
```diff
@@ -1,27 +1,20 @@
...
+import com.gu.contact_us_api.ParserUtils._
 import io.circe.generic.auto._
 
-class ContactUs(SFConnector: SalesforceConnector) extends Logging {
-  def processReq(json: String): Either[ContactUsFailureResponse, ContactUsSuccessfulResponse] = {
-    val result = for {
-      req <- decode[ContactUsRequest](json)
-      resp <- SFConnector.handle(req.asSFCompositeRequest)
-    } yield resp
+class ContactUs(SFConnector: SalesforceConnector) {
```
kelvin-chappell, line 7: If you're going to use this pattern, it might be better if `SalesforceConnector` were a trait so that the code is decoupled from the implementation and more testable.

handlers/contact-us-api/src/main/scala/com/gu/contact_us_api/ContactUsConfig.scala
```diff
@@ -1,13 +1,35 @@
...
-  val clientID: String = System.getenv("CUClientID")
-  val clientSecret: String = System.getenv("CUClientSecret")
-  val username: String = System.getenv("CUUsername")
-  val password: String = System.getenv("CUPassword")
-  val token: String = System.getenv("CUToken")
-
-  val authEndpoint = "https://test.salesforce.com/services/oauth2/token"
-  val reqEndpoint = "https://gnmtouchpoint--DEV.my.salesforce.com/services/data/v43.0/composite/"
+
+  val env: Either[Throwable, ContactUsEnvConfig] = {
```
kelvin-chappell, line 7: Better to define your own `Failure` type for the Left branch or use a `Try[ContactUsEnvConfig]` instead.

handlers/contact-us-api/src/main/scala/com/gu/contact_us_api/ContactUsConfig.scala
```diff
@@ -1,13 +1,35 @@
...
+          username.getOrElse(""),
+          password.getOrElse(""),
+          token.getOrElse(""),
+          s"https://${authDomain.getOrElse("")}/services/oauth2/token",
+          s"https://${reqDomain.getOrElse("")}/services/data/v43.0/composite/",
+        )
+      )
+    } else {
+      Left(new Throwable("Unable to obtain environment variables."))
+    }
```
kelvin-chappell, line 32:
```suggestion
    for {
        clientId <- sys.env.get("clientID").toRight("...")
        clientSecret <- sys.env.get("clientSecret").toRight("...")
        username <- sys.env.get("username").toRight("...")
        password <- sys.env.get("password").toRight("...")
        token <- sys.env.get("token").toRight("...")
        authDomain <- sys.env.get("authDomain").toRight("...")
        reqDomain <- sys.env.get("reqDomain").toRight("...")
    } yield 
        ContactUsEnvConfig(
          clientId,
          clientSecret,
          username,
          password,
          token,
          s"https://$authDomain/services/oauth2/token",
          s"https://$reqDomain/services/data/v43.0/composite/",
        )
      )
```

handlers/contact-us-api/src/main/scala/com/gu/contact_us_api/Handler.scala
```diff
@@ -2,34 +2,51 @@ package com.gu.contact_us_api
...
 
-      case Left(fail)=>
-        new APIGatewayProxyResponseEvent()
-          .withStatusCode(500)
-          .withBody(fail.asJson.toString())
+      case Left(fail) =>
+        logger.error(fail.errorDetails)
+
+        val statusCode = if (fail.errorType == "Input") 400 else 500
+        val message = if (fail.errorType == "Input") "Invalid input" else "Internal server error"
```
kelvin-chappell, line 40:
```suggestion
        val (statusCode, message) = if (fail.errorType == "Input") (400, "Invalid input") else (500, "Internal server error")
```

handlers/contact-us-api/src/main/scala/com/gu/contact_us_api/models/ContactUsEnvConfig.scala
```diff
@@ -0,0 +1,3 @@
+package com.gu.contact_us_api.models
+
+case class ContactUsEnvConfig(clientID: String, clientSecret: String, username: String, password: String, token: String, authEndpoint: String, reqEndpoint: String)
```
kelvin-chappell, line 3: There's an implementation detail leaking here.  If you changed from using environment vars to some other implementation it wouldn't affect this config type.  So perhaps:
```suggestion
case class ContactUsConfig(clientID: String, clientSecret: String, username: String, password: String, token: String, authEndpoint: String, reqEndpoint: String)
```

handlers/contact-us-api/src/main/scala/com/gu/contact_us_api/ContactUs.scala
```diff
@@ -1,27 +1,20 @@
...
-
-  def buildResponse(result: Either[Throwable, Unit]): Either[ContactUsFailureResponse, ContactUsSuccessfulResponse] = {
-    result match {
-      case Left(error) => {
-        logger.error(error.getMessage)
-        Left(ContactUsFailureResponse(error.getMessage))
-      }
-      case Right(_) => Right(ContactUsSuccessfulResponse())
+  def processReq(json: String): Either[ContactUsError, Unit] = {
+    try {
```
mario-galic, line 10: `try-catch` expressions are usually replaced with more idiomatic `Try` monad. However `Either` inside a `Try` is a strange type.

handlers/contact-us-api/src/main/scala/com/gu/contact_us_api/ContactUs.scala
```diff
@@ -1,27 +1,20 @@
...
+      for {
+        req <- decode[ContactUsRequest](json, Some("ContactUsRequest"), "Input")
+        resp <- SFConnector.handle(req.asSFCompositeRequest)
+      } yield resp
+    } catch {
+      case e: Throwable => Left(ContactUsError("Fatal", s"Something crashed: $e"))
```
mario-galic, line 16: `Something crashed` adds no information

handlers/contact-us-api/src/main/scala/com/gu/contact_us_api/ContactUsConfig.scala
```diff
@@ -1,13 +1,35 @@
...
+    val clientSecret: Option[String] = sys.env.get("clientSecret")
+    val username: Option[String] = sys.env.get("username")
+    val password: Option[String] = sys.env.get("password")
+    val token: Option[String] = sys.env.get("token")
+    val authDomain: Option[String] = sys.env.get("authDomain")
+    val reqDomain: Option[String] = sys.env.get("reqDomain")
+
+    val envList = List(clientID, clientSecret, username, password, token, authDomain, reqDomain)
+
+    if (envList.size == envList.flatten.size) {
```
mario-galic, line 18: This type of checks are usually replaced with for comprehension. See how Kelvin does it https://github.com/guardian/support-service-lambdas/pull/713/files#diff-39717759f565d3c1a275f050f77aeb52R13

Note your are going from `null` to `None` to `Left` to `Throwable`, which begs the question why not just got directly to `Throwable`, since if say `password` does not exist we might as well crash as nothing can proceed.

handlers/contact-us-api/src/main/scala/com/gu/contact_us_api/ContactUsConfig.scala
```diff
@@ -1,13 +1,35 @@
...
+      Right(
+        ContactUsEnvConfig(
+          clientID.getOrElse(""),
+          clientSecret.getOrElse(""),
+          username.getOrElse(""),
+          password.getOrElse(""),
```
mario-galic, line 24: What is gain by empty string as default for password etc?

handlers/contact-us-api/src/main/scala/com/gu/contact_us_api/ParserUtils.scala
```diff
@@ -0,0 +1,16 @@
+package com.gu.contact_us_api
+
+import com.gu.contact_us_api.models.ContactUsError
+import io.circe.Decoder
+import io.circe.parser
+
+object ParserUtils {
+
+  def decode[T: Decoder](input: String, decodeTargetDesc: Option[String] = None, errorType: String = "Decode"): Either[ContactUsError, T] = {
```
mario-galic, line 9: What does this do?

ripecosta: It decodes the input json string into class T using circe. If decoding fails it replaces the Throwable with a ContactUsError containing more detailed and human readable information from decodeTargetDesc and errorType.
I'll add a comment to the code.

handlers/contact-us-api/src/main/scala/com/gu/contact_us_api/ContactUs.scala
ripecosta: Agree. Will be changed in upcoming commit.

handlers/contact-us-api/src/main/scala/com/gu/contact_us_api/ContactUs.scala
ripecosta: Addressed in an upcoming commit.

handlers/contact-us-api/src/main/scala/com/gu/contact_us_api/ContactUsConfig.scala
ripecosta: Addressed in upcoming commit.

handlers/contact-us-api/src/main/scala/com/gu/contact_us_api/ContactUsConfig.scala
ripecosta: This is the product of a bad approach. The empty string would never happen due to a check that made before it. The approach is updated in an upcoming commit.

handlers/contact-us-api/src/main/scala/com/gu/contact_us_api/ContactUs.scala
```diff
@@ -1,27 +1,16 @@
...
+import com.gu.contact_us_api.ParserUtils._
 import io.circe.generic.auto._
 
-class ContactUs(SFConnector: SalesforceConnector) extends Logging {
-  def processReq(json: String): Either[ContactUsFailureResponse, ContactUsSuccessfulResponse] = {
-    val result = for {
-      req <- decode[ContactUsRequest](json)
+class ContactUs(SFConnector: SFConnector) {
+
+  def processReq(json: String): Either[ContactUsError, Unit] = {
```
kelvin-chappell, line 9: It's a small point, more for future reference than anything, but as the `SFConnector` only has a single method, you could actually treat it as a function instead of a trait.  (I didn't realise that it only had one method).

Then the signature of this method could be `def processReq(json: String, handle: SFCompositeRequest => Either[ContactUsError, Unit])`
and the class could be an `Object`.

kelvin-chappell (approved):
Looking much clearer now! 👍

handlers/contact-us-api/src/main/scala/com/gu/contact_us_api/ContactUs.scala
ripecosta: If I understand correctly this change would mean I could convert all my classes to objects(`Handler`, `ContactUs` and `SalesforceConnector`. Is that right?

PR #716, opened 2020-10-05
Allow JDK 11

mario-galic:
JDK 11 should be fully supported by Scala and SBT:
* https://www.scala-sbt.org/1.x/docs/Setup.html
* https://docs.scala-lang.org/overviews/jdk-compatibility/overview.html

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -455,8 +455,8 @@ lazy val `http4s-lambda-handler` = all(project in file("lib/http4s-lambda-handle
 
 initialize := {
   val _ = initialize.value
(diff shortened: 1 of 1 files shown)


david-pepper (approved):
Should remove a few headaches

kelvin-chappell:
🎉 
Does the version need to be specified at all? ie can we leave it out?

mario-galic:
Currently AWS Lambda officially [supports](https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html) just 8 and 11:

image

Also sbt recommends the same. However Scala itself should work across the board.

PR #719, opened 2020-10-06
also clean test accounts as well as subs

johnduffell:
In the previous PR https://github.com/guardian/support-service-lambdas/pull/711 I added some code to clean test subscriptions, but I forgot to clean the accounts as well.  This actually meant there was more subscriptions data in salesforce as it stored the cancellation as well as the original sub.
Cancelling the account as well will wipe the lot.

I am running it locally and the space usage in SF is decreasing, so this should work on the daily cleaner as well.

cc @david-pepper @rupertbates @ripecosta


ripecosta (approved):
👍 Thanks for looking into this

PR #720, opened 2020-10-06
Fix contact-us-api RiffRaff config

ripecosta:
## What does this change?
Fixes the RiffRaff config, which was not working as intended.


kelvin-chappell (approved):
👍

PR #721, opened 2020-10-06
Add logging to sf billing account remover

david-pepper:
Change Description

Add logging to the lambda and enable update to new SF field on Billing Account to capture the datestamp of any failed attempts to update in Zuora

### DIFF
--- a/handlers/sf-billing-account-remover/src/main/scala/com/gu/sf_billing_account_remover/BillingAccountRemover.scala
+++ b/handlers/sf-billing-account-remover/src/main/scala/com/gu/sf_billing_account_remover/BillingAccountRemover.scala
@@ -1,5 +1,7 @@
 package com.gu.sf_billing_account_remover
 
+import java.time.LocalDateTime
+import com.typesafe.scalalogging.LazyLogging
 import io.circe._
 import io.circe.generic.auto._
 import io.circe.parser._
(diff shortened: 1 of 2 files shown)


mario-galic:
This [absurd](https://stackoverflow.com/questions/26880677/intellij-scala-plugins-case-class-indentation-is-absurd) indentation is likely caused by IntelliJ which can be disabled `Align when multiline`

image

handlers/sf-billing-account-remover/src/main/scala/com/gu/sf_billing_account_remover/BillingAccountRemover.scala
```diff
@@ -214,6 +213,9 @@ object BillingAccountRemover extends App {
     zuoraConfig: ZuoraConfig,
     accountToDelete: BillingAccountsRecords.Records
   ): Option[BillingAccountsRecords.Records] = {
+    logger.info(
+      s"Updating Billing Account in Zuora ($accountToDelete.Zuora__External_Id__c)"
```
mario-galic, line 217: Note there is a typo here. Curly braces are necessary when using dot notation inside string interpolation like so

```
logger.info(s"Updating Billing Account in Zuora ${accountToDelete.Zuora__External_Id__c}" )
```

david-pepper: Nice catch. This has been done

mario-galic (approved):

PR #722, opened 2020-10-06
Fix wrong confirmation email for Subscription Card (Digital Voucher)

mario-galic:
## What does this change?

Fix error introduced by https://github.com/guardian/support-service-lambdas/pull/679 which caused sending paper voucher confirmation email instead of Subscription Card one.

## How to test

Try CSR acquisition of Subscription Card in UAT and see if correct confirmation email is sent.

SV_SC_WelcomeDay0 campaign should pick up.

No major risk apparent given correct campaign mapping.


arausuy (approved):
Good find!

PR #724, opened 2020-10-07
Contact Us API: Secrets are now versioned

ripecosta:
## What does this change?
Uses the version when fetching the key-values from secret managers to make it easier to update credentials.

This is necessary because redeploying the stack in cloud formation after updating the secrets did not update the lambda's environment variables. This is because it considered the stack to be exactly the same and it wouldn't update any of the resources.

### DIFF
--- a/handlers/contact-us-api/cfn.yaml
+++ b/handlers/contact-us-api/cfn.yaml
@@ -9,6 +9,13 @@ Parameters:
       - CODE
     Default: CODE
 
+Mappings:
+  StageMap:
+    CODE:
+      SecretsVersion: "9a7414c9-6711-44fa-a683-b1d11643c8c7"
+    PROD:
+      SecretsVersion: "0cee634d-a5f4-454d-a3a2-9d57ee90e154"
+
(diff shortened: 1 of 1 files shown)


kelvin-chappell (approved):
👍 

For future reference of other reviewers.  We're going to document this.
The problem seems to be in the way that Cloudformation updates resources.
It's described [here](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/dynamic-references.html):
> Updating a secret in Secrets Manager does not automatically update the secret in CloudFormation. In order for CloudFormation to update a secretsmanager dynamic reference, you must perform a stack update that updates the resource containing the dynamic reference, either by updating the resource property that contains the secretsmanager dynamic reference, or updating another of the resource's properties.

So we have to use versioned secrets.

And the way to get the version number is by fetching with the CLI and using the UUID of the version ID, as described [here](https://github.com/guardian/price-migration-engine/blob/main/lambda/README.md#config).  The AWS console doesn't show the version number.

PR #725, opened 2020-10-08
Add deployment configuration for digital-voucher suspension lambda

kelvin-chappell:
This adds Cloudformation and Riffraff configuration to the project.

Riffraff has successfully [deployed](https://riffraff.gutools.co.uk/deployment/view/9c93355f-c2fe-4c3a-944c-68f66123d60e) the lambda to Code.

### DIFF
--- a/handlers/digital-voucher-suspension-processor/cfn.yaml
+++ b/handlers/digital-voucher-suspension-processor/cfn.yaml
@@ -16,6 +16,9 @@ Resources:
     Type: AWS::Serverless::Function
     Properties:
       FunctionName: !Sub digital-voucher-suspension-processor-${Stage}
+      Description: >
+        Suspends fulfilment of digital voucher subscriptions.
+        Source: https://github.com/guardian/support-service-lambdas/tree/main/handlers/digital-voucher-suspension-processor
       Runtime: java8
       MemorySize: 1536
       Timeout: 900
--- a/handlers/digital-voucher-suspension-processor/cfn.yaml
+++ b/handlers/digital-voucher-suspension-processor/cfn.yaml
@@ -1,0 +1,35 @@
+Transform: AWS::Serverless-2016-10-31
+
+Parameters:
+
+  Stage:
+    Description: Stage name
+    Type: String
+    AllowedValues:
+      - PROD
+      - CODE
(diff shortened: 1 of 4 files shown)


handlers/digital-voucher-suspension-processor/cfn.yaml
```diff
@@ -0,0 +1,35 @@
...
+    Description: Stage name
+    Type: String
+    AllowedValues:
+      - PROD
+      - CODE
+    Default: CODE
+
+Resources:
+
+  SuspensionLambda:
```
mario-galic, line 15: Where is description and role?

kelvin-chappell: This is a `Transform: AWS::Serverless-2016-10-31` template so the role is created implicitly.  
But I'll add a description.

mario-galic (approved):

handlers/digital-voucher-suspension-processor/cfn.yaml
```diff
@@ -0,0 +1,38 @@
...
+      Runtime: java8
+      MemorySize: 1536
+      Timeout: 900
+      CodeUri:
+        Bucket: support-service-lambdas-dist
+        Key: !Sub membership/${Stage}/digital-voucher-suspension-processor/digital-voucher-suspension-processor.jar
+      Handler: com.gu.digitalvouchersuspensionprocessor.Handler::handleRequest
+      Environment:
+        Variables:
+          salesforceUrl: !Sub '{{resolve:secretsmanager:digital-voucher-suspension-processor-${Stage}:SecretString:salesforceUrl}}'
```
ripecosta, line 31: Perhaps it's best to use versions here?

kelvin-chappell: After the recent discussion I think we can come up with a better way to manage versions.
I can't imagine any reason you would ever want to use anything but the latest version of a secret.

ripecosta: Agree that we probably always want the latest version of a secret. And it's very likely we'll come up with that better way to manage version before we need to change this lambda's credentials. So yeah 👍

PR #727, opened 2020-10-08
Update Salesforce query for finding subscriptions to suspend so that it works in every stage

kelvin-chappell:
The product type is different in every deployment stage but the product name appears to be the same everywhere.
So this change uses the product name to find subscriptions to suspend instead.

It's been tested in the Code lambda.


mario-galic:
Why is it different in every environement?

mario-galic (approved):

kelvin-chappell:
> Why is it different in every environment?

I don't think there's any good reason.  It does make you wonder what the benefit of a UAT environment is.

PR #732, opened 2020-10-09
Update sbt 1.4.0

mario-galic:
## What does this change?

https://github.com/sbt/sbt/releases/tag/v1.4.0

## How to test

Try to build.

Local build, TC build, and snyk monitor happy.

No major risk apparent.


johnduffell (approved):
good solution 👏

PR #735, opened 2020-10-12
Process digital voucher suspensions a day in advance of their effective date

kelvin-chappell:
Currently we process suspensions on the day that they are effective.  This means that the subscription card is usable for at least the start of the suspension day and potentially for several hours during the day if there is any problem with either the holiday-stop processor or the digital voucher suspension processor.

This change pushes the threshold for holiday stops back a day so that the suspension can be processed the day before it becomes effective.  This gives us some confidence that the suspension will be processed correctly in advance of when it becomes effective.

To test:
* Should be blocked from creating a holiday stop for tomorrow in frontend and in Salesforce.  But should be able to create a stop for the day after tomorrow.
* Holiday stops should be being processed for the next day.


arausuy:
My understanding is that this behaviour is correct. i-movo can only block redemption if the request is received before midnight of the intended day. i.e if my holiday is on the 02/06/2021 i-movo must receive the request on 01/06/2021 at the very latest

arausuy (approved):

kelvin-chappell:
@arausuy That's right so in fact we have to do it this way anyway.

PR #736, opened 2020-10-13
Enable product override in holiday-stop processor

kelvin-chappell:
Currently, the holiday-stop processing lambda can take an optional date argument.  If provided, only holiday credit requests for that publication date will be processed.  The problem is that all product types will be processed for that date.  If we want to test the effect of, say, processing digital voucher suspensions on a certain date we can't do it without also processing every other type of subscription.

This change alters the type of argument that the lambda expects so that it's now a combination of product type and stopped publication date, like so:
```
{
  "productType": {
    "name": "Guardian Weekly"
  },
  "stopDate": "2020-10-13"
}
```

This has been tested in Code, with and without the argument.  It works as expected.

The change also includes some refactoring, removing the standalone app and adding an endpoint to the handler for standalone use.


handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/HolidayStopCreditProcessor.scala
```diff
@@ -18,25 +18,45 @@ import scala.util.Try
...
     config: Config,
-    processDateOverride: Option[LocalDate],
+    productTypeAndStopDateOverride: Option[ProductTypeAndStopDate],
     backend: SttpBackend[Id, Nothing],
     fetchFromS3: S3Location => Try[String]
-  ): List[ProcessResult[ZuoraHolidayCreditAddResult]] =
+  ): List[ProcessResult[ZuoraHolidayCreditAddResult]] = {
+
+    val productTypesToProcess =
+      productTypeAndStopDateOverride.map(productTypeAndStopDate =>
```
mario-galic, line 42: Would pattern match be clearer?

```
    productTypeAndStopDateOverride match {
      case Some(ProductTypeAndStopDate(productType, _)) => 
        List(productType)
        
      case None =>
        List(
          NewspaperHomeDelivery,
          NewspaperVoucherBook,
          NewspaperDigitalVoucher,
          GuardianWeekly,
        )
    }
```

mario-galic (approved):
Does maybe some readme need updating?

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/HolidayStopCreditProcessor.scala
kelvin-chappell: 👍

PR #739, opened 2020-10-14
Use new secrets in Contact Us API

kelvin-chappell:
This is the same as #730 but for the Contact Us API stacks.

Secrets are now rearranged in a form that makes it clear which credentials each stack is using and makes it easier to rotate them in one place.


ripecosta (approved):
👍

PR #740, opened 2020-10-14
Remove subproject build.sbt files

mario-galic:
## What does this change?

* Remove  all subproject `handler/${name}/build.sbt` files and configure all subprojects via single root `build.sbt` as per official sbt [recommendation](https://www.scala-sbt.org/1.x/docs/Multi-Project.html#Appendix%3A+Subproject+build+definition+files)

  > We recommend putting all project declarations and settings in the root build.sbt file in order to keep all build definition under a single file
* There is now a single way to configure basic AWS Scala lambda project via [`lambdaProject`](https://github.com/guardian/support-service-lambdas/pull/740/files?diff=split&w=1#diff-5634c415cd8c8504fdb973a3ed092300b43c4b8fc1e184f7249eb29a55511f91R245) method.
* **Rename all riff-raff projects to format `support-service-lambdas::$projectName`**. For example

  old name

  ```
  MemSub::Membership Admin::Metric Push API
  ```

  new name

  ```
  support-service-lambdas::metric-push-api
  ```

## How to test

1. `sbt test`
1. `sbt projects`
1. Try deploying via riff-raff using the new name

- Easier to find project in riff-raff as there is now one-to-one mapping between riff-raff and repo
- Less chance of forking approaches if `lambdaProject` is used consistently
- Maybe less red false positive squiggly lines in IJ

* There should be no changes in business logic as this is just sbt reconfiguration.  
* There will be confusion over old riff-raff names and new names as for a while both will be showing up in riff-raff deploy form UI https://riffraff.gutools.co.uk/deployment/request, so we need to remove the [old riff-raff artifacts](https://s3.console.aws.amazon.com/s3/buckets/riffraff-artifact/?region=eu-west-1&tab=overview&prefixSearch=MemSub::Membership%20Admin) from dev-tools riff-raff bucket which we can do with a script.


handlers/zuora-callout-apis/build.sbt
```diff
@@ -1,3 +1,4 @@
+// FIXME: This sub-project builds.sbt should be removed
```
johnduffell, line 1: oh yeah, actually this is the "old" mono lambda build, which is actually 3 lambdas into one jar.  Once the new/standard approach was decided on, no one ever had the impetus to go back and rejig these.
Would be a great thing to make them into the new format though, even if they are still bundled together.

build.sbt
```diff
@@ -222,238 +222,272 @@ lazy val `zuora-core` = all(project in file("lib/zuora-core"))
...
-        http4sServer,
-        sttpAsyncHttpClientBackendCats,
-        scalatest,
-        diffx,
-        scalaMock,
-      )
-      ++ logging
-  )
+// FIXME: This seems to be non-standard
+// FIXME: Why is the name in sub-project build.sbt support-service-lambda
```
johnduffell, line 266: because originally it was the only lambda in the repo (it was called zuora-auto-cancel in those days though). Then it had a couple of other lambdas tacked on the side.  About a year ago it was moved from the root of the project into zuora-callout-apis subfolder to get it out of the way.  But noone has ever gone in again and modernised it into the way that is now standard.

build.sbt
```diff
@@ -222,238 +222,272 @@ lazy val `zuora-core` = all(project in file("lib/zuora-core"))
...
+// FIXME: Why is riff-raff not refering to CF?
```
johnduffell, line 267: the cfn is manually applied(!) I remember adding auto cfn for a new lambda one time, as it wasn't in any existing ones.

johnduffell:
it is great to freshen things up, but I won't lie, that is a monster root build.sbt.  For everything else we write, we divide it appropriately into source files to make it easier to maintain.  I have seen the recommendation from sbt, but I have no idea why they recommend that or whether their assumptions apply to us.  Do you have any additional information?

On the other side, I think it will help snyk monitor as it won't find all the other `build.sbt` any more.  But we could just rename them `somethingElse.sbt`

johnduffell:
we use `support:` (one colon) rather than `support-` (one dash) in the support-frontend repo
https://github.com/guardian/support-frontend/blob/89da4af850a30620697903cec3de0a78fa4eec64/support-workers/build.sbt#L47

mario-galic:
> that is a monster root build.sbt

image

mario-galic:
> I have seen the recommendation from sbt, but I have no idea why they recommend that or whether their assumptions apply to us

According to sbt contributor [dwijnand](https://gitter.im/sbt/sbt?at=5cf4bf755f26b34a8b797ad9) 

> There's a few reasons. For starters it's just a less common setup. Secondly it can lead to more mistakes: I've seen too many instances of build users being confused about their build because they were looking at <root>/build.sbt and forgetting that some settings are in foo/build.sbt. It's easier to review and edit a single build.sbt, breaking down complex aspects of the build by topic, rather than by project (by moving the code into a file under project/*.scala, either as raw code or as a local sbt plugin). Finally (and probably the most important one) having the build settings in foo/build.sbt gives the illusion that you can start the build in that directory and have that project only (like Maven): that isn't true for sbt because it risks using a different sbt version and it doesn't access the plugins and custom code defined in ../project.

Furthermore major scala projects all use single `build.sbt` approach, such as scala proper, cats, metals, etc.

project/Dependencies.scala
```diff
@@ -64,6 +64,7 @@ object Dependencies {
   val enumeratum = "com.beachape" %% "enumeratum" % "1.5.13"
   val scalaXml = "org.scala-lang.modules" %% "scala-xml" % "1.3.0"
   val stripe = "com.stripe" % "stripe-java" % "5.28.0"
+  val betterFiles = "com.github.pathikrit" %% "better-files" % "3.9.1"
```
kelvin-chappell, line 67: If this is a refactor, why is this introduced?

kelvin-chappell (approved):
If Riffraff deploys are still working, it's a great improvement 👍 

There seemed to be some inconsistencies in how people treated properties like dependencies - in some sub-projects they were in root and in others they were local.  This makes it much clearer where to look.

project/Dependencies.scala
mario-galic: It is not introduced, it was simply moved to dependencies. Anyways it is used by `braze-to-salesforce-file-upload` which is actually dead project.

build.sbt
```diff
@@ -222,238 +222,272 @@ lazy val `zuora-core` = all(project in file("lib/zuora-core"))
...
-  )
-
-lazy val `delivery-problem-credit-processor` =
-  all(project in file("handlers/delivery-problem-credit-processor"))
-    .dependsOn(
-      `credit-processor`,
-      `salesforce-sttp-client`,
-      effects
-    )
+def lambdaProject(projectName: String, projectDescription: String, dependencies: Seq[sbt.ModuleID] = Nil, isCdk: Boolean = false) = {
```
arausuy, line 245: I really like this approach, nice and clear

arausuy (approved):

PR #743, opened 2020-10-21
Replace custom next invoice date calculation with billing-preview in credit processor

mario-galic:
## What does this change?

* https://trello.com/c/i0a10FyJ/1542-replace-billing-date-calculation-in-support-service-lambdas-with-calls-to-billing-preview-endpoint
* Start replacing https://github.com/guardian/support-service-lambdas/pull/550 with out-of-the-box functionality to predict when the credit should be applied.
* Start wiring in https://github.com/guardian/invoicing-api/pull/20
* Note for now the new functionality will run alongside the old one so we can validate by comparing the two calculations. Once validated I will address the `FIXME`s.

## How to test

* Let it run in PROD for a while and see if `testInProdNextInvoiceDate` ever logs the following error

```
ERROR - testInProdNextInvoiceDate failed because Add(2c92c0f96b03800b016b081fc04f1ba2,2020-12-11,2020-12-11,2020-12-11,List(ChargeOverride(2c92c0f96b03800b016b081fc0f41bb4,2020-11-06,2020-11-06,-1.0))) =/= 2020-10-08
```

After few days if no errors are logged we can begin to switch over to new functionality.

The risks have been mitigated by deploying in steps, that is, not immediately switching over to the new calculation and yet exercising the new paths at the same time.

### DIFF
--- a/lib/credit-processor/src/main/scala/com/gu/creditprocessor/NextInvoiceDate.scala
+++ b/lib/credit-processor/src/main/scala/com/gu/creditprocessor/NextInvoiceDate.scala
@@ -6,9 +6,8 @@ import com.softwaremill.sttp._
 import com.softwaremill.sttp.circe._
 import io.circe.generic.auto._
 
-case class NextInvoiceDate(nextInvoiceDate: LocalDate)
-
 object NextInvoiceDate {
+  private case class NextInvoiceDate(nextInvoiceDate: LocalDate)
   private implicit val sttpBackend: SttpBackend[Id, Nothing] = HttpURLConnectionBackend()
   private val xApiKey = sys.env.getOrElse("InvoicingApiKey", throw new RuntimeException("Missing x-api-key for invoicing-api"))
   private val invoicingApiUrl = sys.env.getOrElse("InvoicingApiUrl", throw new RuntimeException("Missing invoicing-api url"))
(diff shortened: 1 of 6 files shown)


lib/credit-processor/src/main/scala/com/gu/creditprocessor/NextInvoiceDate.scala
```diff
@@ -0,0 +1,26 @@
+package com.gu.creditprocessor
+
+import java.time.LocalDate
+import com.gu.zuora.subscription.ZuoraApiFailure
+import com.softwaremill.sttp._
+import com.softwaremill.sttp.circe._
+import io.circe.generic.auto._
+
+case class NextInvoiceDate(nextInvoiceDate: LocalDate)
```
kelvin-chappell, line 9: Perhaps add a comment to say that this is for json deserialisation, as not immediately obvious.

kelvin-chappell (approved):
👍 

The `Processor` could do with some refactoring at some point.  It's become very difficult to follow.  Perhaps we can put all the function arguments it takes together into a trait.  Anyway, one for a rainy day.

lib/credit-processor/src/main/scala/com/gu/creditprocessor/NextInvoiceDate.scala
mario-galic: I moved it inside the object and made it private instead.

mario-galic:
Here is a branch preforming full switchover that can be used as a guide to replace the custom functionality https://github.com/guardian/support-service-lambdas/tree/switch-over-to-invoicing-api-preview-which-uses-zoura-billing-preview

PR #745, opened 2020-10-23
Disable lambda retry on zuora-datalake-export

mario-galic:
## What does this change?

When lambda is invoked [asynchronously](https://aws.amazon.com/blogs/architecture/understanding-the-different-ways-to-invoke-lambda-functions/) such as via `AWS::Events::Rule` then AWS automatically retries it two more times in case of a failure. If the lambda is not idempotent then this represents a significant bug, which is indeed the case for `zuora-datalake-export` when triggered to fetch changes from last increment. 

For example, due to [zuora outage](https://trust.zuora.com/incidents/t759bhp7pw6y ), although batch jobs finished successfully, downloading the CSV from Zuora failed, which caused three invocations of lambda

image

and three batch jobs which all succeeded

image

However the last batch jobs had no data because nothing changed in such a short span of time

image

which means empty CSVs **overwrote** files in raw data lake buckets, and data lake did not alarm as a fresh files was copied over after all. This effectively introduced a **silent error** into the system.

Example async invocation 
```
aws lambda invoke --function-name zuora-datalake-export-CODE --invocation-type Event --payload '{"exportFromDate": "2020-10-21"}' response.json --profile membership --debug --cli-binary-format raw-in-base64-out
```

## How to test

Inject failure into lambda, and then execute it asynchronously and see if it is automatically retried by looking at the `Recent Invocations` section under `Monitoring` tab in AWS Console. 
https://docs.aws.amazon.com/lambda/latest/dg/invocation-async.html

N/A

No major risk.

### DIFF
--- a/handlers/zuora-datalake-export/cfn.yaml
+++ b/handlers/zuora-datalake-export/cfn.yaml
@@ -118,3 +118,11 @@ Resources:
         Threshold: 1
         TreatMissingData: notBreaching
 
+    DisableAutomaticLambdaRetry:
+      Type: AWS::Lambda::EventInvokeConfig
+      Properties:
+        FunctionName: !Sub zuora-datalake-export-${Stage}
+        MaximumRetryAttempts: 0
+        Qualifier: '$LATEST'
+      DependsOn: ExportLambda
+


kelvin-chappell (approved):
👍

PR #746, opened 2020-10-30
Update aws sdk to 1.11.890

tomrf1:
flagged by snyk for jackson vulnerabilities


project/Dependencies.scala
```diff
@@ -4,7 +4,7 @@ import sbtassembly.AssemblyPlugin.autoImport.{MergeStrategy, assemblyMergeStrate
 import sbtassembly.PathList
 
 object Dependencies {
-  val awsVersion = "1.11.879"
+  val awsVersion = "1.11.890"
```
johnduffell, line 7: is that showing that it fixed it for you?
https://mvnrepository.com/artifact/com.amazonaws/aws-java-sdk-core/1.11.889
image

https://mvnrepository.com/artifact/com.amazonaws/aws-java-sdk-core/1.11.890
image

johnduffell:
I think the issue is actually related to @mario-galic 's refactor, he was fixing it a couple of weeks ago so I'm not sure whether that is still in progress

project/Dependencies.scala
tomrf1: oh good spot. So it's not actually resolved by the latest aws sdk

tomrf1: and I see we do in fact have a recent enough override set in `Dependencies.scala` for `jackson-databind`, but we're not using it in enough projects

johnduffell: no it never is, AWS don't seem to have a process for increasing the version numbers as new versions arrive

johnduffell (approved):
thanks, looks good 👍

mario-galic:
Resolving this issue would require migration to AWS Java SDK v2. I do no think forcing old software to run with newer dependencies against which it was not built nor tested is a good idea (although in this case it will likely not matter as we do not seem to use jackson for json).

kelvin-chappell:
Starting the migration to v2 in #747

johnduffell:
migrating to SDK V2 would be a great thing, especially in this project which is entirely lambdas! given the improvements in startup time etc. 👏

kelvin-chappell:
@tomrf1 Do you mind if I close this now that the repo's using SDK v2?

PR #747, opened 2020-11-02
Update AWS cloudwatch SDK to v2

kelvin-chappell:
This is the first step in upgrading AWS SDK v1 to v2.

Benefits:
* Version 2 uses Jackson 2.10.x which should resolve [this Snyk vulnerability](https://app.snyk.io/vuln/SNYK-JAVA-COMFASTERXMLJACKSONCORE-559094)
* Much improved performance
* Fewer dependencies
* Improved consistency 
* Easier to use
* Strongly enforced immutability
* https://aws.amazon.com/blogs/developer/the-aws-sdk-for-java-will-no-longer-support-java-6/

(I've taken some of these benefits from [our Trello card](https://trello.com/c/Opyketf2/1691-migration-of-support-service-lambdas-to-aws-sdk-for-java-version-2-snyk-m-l).)

I'm taking the approach of upgrading each AWS module independently, in ascending order of how many apps use it.  At some point, for some apps, the jar size will swell to take v1 and v2 dependencies.  But they should then shrink as v1 dependencies are removed.

The main potential risk is that some vital piece of configuration has been missed.


project/Dependencies.scala
```diff
@@ -4,7 +4,8 @@ import sbtassembly.AssemblyPlugin.autoImport.{MergeStrategy, assemblyMergeStrate
 import sbtassembly.PathList
 
 object Dependencies {
-  val awsVersion = "1.11.879"
+  val awsVersion1 = "1.11.892"
+  val awsVersion2 = "2.15.19"
```
johnduffell, line 8: minor but could be more like 
```suggestion
  val awsV1Version = "1.11.892"
  val awsV2Version = "2.15.19"
```

johnduffell (approved):
yes I had a similar thing when moving to AWS v2 for dynamo, everything has changed slightly!  Looks good though, especially if it's still working afterwards 👍

project/Dependencies.scala
kelvin-chappell: That was my first thought but then it's like version1version.  A bit like a PIN number being a number number.
But if you think it's clearer I'll change it.

johnduffell: I just felt like it doesn't make sense the existing way - it makes sense to me to have a version number for each of version 1 and version 2.
But either way it's a minor point.

PR #744, opened 2020-10-22
AWS Api User Credentials Setter

david-pepper:
## Feature Description
Generate and set a password for each Aws API User in Salesforce and then update a Secret in Secrets Manager with the new password value

## Context
This feature has been developed to relieve some of the manual effort involved in API User Activation in Salesforce when a Sandbox Refresh has taken place.

## Process Flow
image

## Testing

1. Identify a test user in the sandbox
2. Alter the query in the lambda to query for that user only
3. Execute the lambda
4. navigate to secrets manager
5. Verify that an entry has been created for the user OR that an existing entry has been updated
6. Use the password stored in the secrets manager to log in as the user
7. If log in was successful, then the process has executed successfully

## Considerations

- The running user for this process will be generated at the point of sandbox generation, therefore at the point of deployment, there will be no credentials available

- Salesforce Password Policy is currently:
a minimum of 8 characters AND
at least 1 alpha character AND
at least 1 numeric character
	
An improvement to the solution would be if we could extract a regex representation from Salesforce programmatically for the Password Policy and inject this into the password generation mechanism

PR #748, opened 2020-11-03
contact-us-api: Cloud Formation fixes

ripecosta:
## What does this change?
This PR addresses 3 issues were were seeing in the `contact-us-api` related to the Cloud Formation template. It also adds a DEV stage so it's simpler to test in the future and switches to using the new secrets created by our newly created [User Credential Setter](https://github.com/guardian/support-service-lambdas/pull/744) or similarly named ones (for PROD).

### Problems
1. `authDomain` and `reqDomain` were being set to the same domain, which prevented the creation of cases in DEV and PROD environment.
2. A ghost stage (called 'Stage') was being created in the API Gateway.
3. Cloud Watch metrics were not being published under the expected name (should be published under API Gateway name according to [the documentation](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-metrics-and-dimensions.html#api-gateway-metricdimensions))

### Solutions
1. `authDomain` and `reqDomain` are set explicitly and independently. This is a temporary fix. Ideally the reqDomain should be obtained from Salesforce's authentication response.
2. This is a [known issue](https://github.com/aws/serverless-application-model/issues/191). To work around it the [`OpenApiVersion` property](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-resource-api.html#sam-api-openapiversion) is set to `2.0` on the AWS::Serverless::Api resource's property.
3. The Cloud Formation stack needs to be deleted and created again, now using the correct ApiName. The cause of this issue was that the API Gateway resource was initially created with no name and an explicit name was only set after as an update. When the stack was created and the resource had no explicit name so it assumed the stack's name instead. Later when an explicit name was set the metrics continued to be published under the original name instead of updating to the new name. AWS support has confirmed this is not the expected behaviour and have created an internal ticket for the API Gateway to investigate. In the meantime recreating the stack will fix it.

### DIFF
--- a/handlers/contact-us-api/cfn.yaml
+++ b/handlers/contact-us-api/cfn.yaml
@@ -16,13 +16,16 @@ Conditions:
 Mappings:
   StageMap:
     PROD:
-      Domain: gnmtouchpoint.my.salesforce.com
+      authDomain: gnmtouchpoint.my.salesforce.com
+      reqDomain: gnmtouchpoint.my.salesforce.com
       SecretsVersion: cfe10466-bd62-45f2-a61f-3fcf9c12ef9f
     CODE:
-      Domain: test.salesforce.com
+      authDomain: test.salesforce.com
+      reqDomain: gnmtouchpoint--DEV.my.salesforce.com
       SecretsVersion: 5a68b23e-6d75-4bb4-9131-a4d94c121010
     DEV:
-      Domain: test.salesforce.com
+      authDomain: test.salesforce.com
+      reqDomain: gnmtouchpoint--DEV.my.salesforce.com
       SecretsVersion: 89ac830e-166c-444a-a61b-72fb5173e761
 
 Resources:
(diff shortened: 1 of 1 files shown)


handlers/contact-us-api/cfn.yaml
```diff
@@ -93,7 +101,7 @@ Resources:
       ComparisonOperator: GreaterThanThreshold
       Dimensions:
         - Name: ApiName
-          Value: !Sub membership-${Stage}-contact-us-api
+          Value: !Sub contact-us-api-${Stage}-ApiGateway
```
mario-galic, line 104: `-ApiGateway` adds no information as this is already known from property `Namespace: AWS/ApiGateway`

ripecosta: I felt it was best to make it explicit since the name we were using before was not the API Gateway's name. But if you think it's a problem I'll change it.

kelvin-chappell (approved):
👍

PR #750, opened 2020-11-04
Fix Cloudwatch MetricDatum building

kelvin-chappell:
There's an alarm that tells us the SF dev env has not been cleared out recently.  It went off because the metric that gets written at the end of the lambda was missing the `stage` dimension.  The last dimension was overwriting any previous dimensions as a result of PR #747.

The problem was [here](https://github.com/guardian/support-service-lambdas/compare/kc-cloudwatch-fix?expand=1#diff-c0af23d9ebd1f190cebbae557689d125723a313754b209a7f64b23fa990c361eL51-L58) where the dimensions are repeatedly being recreated each time around the fold.


johnduffell (approved):
nice work, I guess that was a subtle difference in how the builder was working.  The new way is better though, even though it threw us a bit!  👍

PR #752, opened 2020-11-06
Disable debug logging of AWS SDK v2 code

kelvin-chappell:
AWS SDK v2 generates a lot of debug logging output by default, which fills the Cloudwatch logs with noise.
This reduces it to a clear signal.


mario-galic (approved):
Strange that `software.amazon.awssdk` is `DEBUG` by default...

kelvin-chappell:
It is very strange.

PR #753, opened 2020-11-09
Upgrade AWS SQS SDK to v2

kelvin-chappell:
This is the next step in the upgrade to v2, following on from #747.

This time it's an upgrade of the SQS module, which is used by:
* [zuora-callout-apis](https://github.com/guardian/support-service-lambdas/tree/main/handlers/zuora-callout-apis)
* [new-product-api](https://github.com/guardian/support-service-lambdas/tree/main/handlers/new-product-api)
* [batch-email-sender](https://github.com/guardian/support-service-lambdas/tree/main/handlers/batch-email-sender)

I've run the functional tests on these projects to prove that they work at runtime.

In Scala 2.13 we can convert Java futures to Scala futures using [FutureConverters](https://github.com/scala/scala/blob/v2.13.3/src/library/scala/jdk/FutureConverters.scala), which means the [AWSAsyncHandler](https://github.com/guardian/support-service-lambdas/blob/main/lib/effects-sqs/src/main/scala/com/gu/effects/sqs/AWSAsyncHandler.scala) becomes redundant.


ripecosta (approved):
LGTM 👍

lib/effects-sqs/src/main/scala/com/gu/effects/sqs/AwsSQSSend.scala
```diff
@@ -17,58 +17,71 @@ object AwsSQSSend extends LazyLogging {
...
 
 object aws {
   val ProfileName = "membership"
 
-  lazy val CredentialsProvider = new AWSCredentialsProviderChain(
-    new EnvironmentVariableCredentialsProvider,
-    new SystemPropertiesCredentialsProvider,
-    new ProfileCredentialsProvider(ProfileName),
-    new InstanceProfileCredentialsProvider(false),
-    new EC2ContainerCredentialsProviderWrapper
```
mario-galic, line 71: What happened to `xnstanceProfileCredentialsProvider` and `EC2ContainerCredentialsProviderWrapperx`?

lib/effects-sqs/src/main/scala/com/gu/effects/sqs/AwsSQSSend.scala
```diff
@@ -17,58 +17,71 @@ object AwsSQSSend extends LazyLogging {
...
+          logger.error(s"Failed to send message to $queueUrl due to:", throwable)
+          Failure(throwable)
+      }
+    } yield ()
   }
 
   def sendSync(queueName: QueueName)(payload: Payload): Try[Unit] = {
-    val (sqsClient: AmazonSQSAsync, queueUrl: String) = buildSqsClient(queueName)
+
+    val sqsClient = SqsClient
```
mario-galic, line 51: I realise this PR did not introduce it, but why are the clients created inside methods instead of once as member of singleton?

lib/effects-sqs/src/test/resources/logback-test.xml
```diff
@@ -0,0 +1,16 @@
+<configuration>
```
mario-galic, line 1: This is dreary. Might be worth asking AWS why is debug level the default?

mario-galic:
Regarding `batch-email-sender` I think it is worth performing full end-to-end test, that is, acquire using real email address with `+` postfix trick and then, say, book a holiday and see if the email actually arrives.

lib/effects-sqs/src/main/scala/com/gu/effects/sqs/AwsSQSSend.scala
kelvin-chappell: They don't seem to be relevant for credentials accessible to a lambda.

lib/effects-sqs/src/main/scala/com/gu/effects/sqs/AwsSQSSend.scala
kelvin-chappell: Yes I was just trying to make it work like-for-like.  But I can't see any good reason for rebuilding the client on every request.

lib/effects-sqs/src/test/resources/logback-test.xml
kelvin-chappell: I'll raise a ticket to see.

kelvin-chappell:
I can confirm that deploying this branch in Code and making a holiday stop on my sub has resulted in me receiving an email through `batch-email-sender`.

mario-galic (approved):

PR #754, opened 2020-11-10
Run test in production off the main thread

mario-galic:
## What does this change?

Make sure tests in prod do not slow down main business logic.

## How to test

Maybe measure response before and after with Postman by hitting `{{holidayStopApiUrl}}/potential/A-S00000000?startDate=2020-11-24&endDate=2020-12-02`

N/A

No risk.


kelvin-chappell (approved):
👍 
Do you suspect these tests are slowing Prod down?

mario-galic:
Indeed they do!

PR #758, opened 2020-11-16
Upgrade AWS S3 SDK to v2

kelvin-chappell:
This upgrades the AWS S3 module to v2.

Although it touches quite a few files, the actual change is quite small.
There's some scope for refactoring and centralising the calls to S3 but I've made this as small a change as I can.

I've tested it in various projects in Code.

The new code follows a similar format to the previously changed modules below.

Follows on from:
* Upgrade of Cloudwatch module: #747
* Upgrade of SQS module: #753

See also https://docs.aws.amazon.com/sdk-for-java/latest/migration-guide/whats-different.html

Perhaps if I were to do this again I'd concentrate on the s3-effects subproject and then change all the client code to use that.


mario-galic (approved):
I can double check tomorrow that contents of files written to S3 looks ok, for example, in zuora-datalake-export.

kelvin-chappell:
> I can double check tomorrow that contents of files written to S3 looks ok, for example, in zuora-datalake-export.

Thanks @mario-galic

PR #765, opened 2020-11-19
Remove explicit jackson databind dependency

kelvin-chappell:
Version 2.10.4 is now coming in as a transitive dependency so this explicit dependence seems to be redundant.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -70,7 +70,7 @@ lazy val zuora = library(project in file("lib/zuora"))
     effects % "test->test"
   )
   .settings(
-    libraryDependencies ++= Seq(okhttp3, playJson, scalatest, jacksonDatabind) ++ logging
+    libraryDependencies ++= Seq(okhttp3, playJson, scalatest) ++ logging
   )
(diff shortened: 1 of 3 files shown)


mario-galic (approved):

kelvin-chappell:
Although there's still a reference to bad Jackson coming through from the simple configuration project so I'll have a look at that first before merging this.

mario-galic:
Even so overriding a dependency to silence snyk is just exchanging one problem for another, as with override you are running software with dependencies against which it was not built and tested. JVM has no intelligence here, it nearly picks whatever is first on the classpath, and sbt will by default put the highest version on the classpath. I would still merge this.

kelvin-chappell:
> Even so overriding a dependency to silence snyk is just exchanging one problem for another, as with override you are running software with dependencies against which it was not built and tested. JVM has no intelligence here, it merely picks whatever is first on the classpath, and sbt will by default put the highest version on the classpath. I would still merge this.

I quite agree!

PR #766, opened 2020-11-27
Limit number of cancelled digital vouchers to process

kelvin-chappell:
We were seeing errors in the log like:
```
{"compositeResponse":[{"body":[{"errorCode":"LIMIT_EXCEEDED","message":"The request can’t contain more than 25 operations."}],"httpHeaders":{},"httpStatusCode":400,"referenceId":null}]}
```
when attempting to write results back to Salesforce.  
This is because the number of cancellations had built up following the upstream outage.

This change fixes the problem by limiting how many cancellations we attempt to process at one go.
As the processor runs every hour we'll have to have > 25 cancellations every hour over a prolonged period before the process runs into trouble.

This has been tested in Prod.

### DIFF
--- a/handlers/digital-voucher-cancellation-processor/src/main/scala/com/gu/digital_voucher_cancellation_processor/DigitalVoucherCancellationProcessorService.scala
+++ b/handlers/digital-voucher-cancellation-processor/src/main/scala/com/gu/digital_voucher_cancellation_processor/DigitalVoucherCancellationProcessorService.scala
@@ -2,14 +2,13 @@ package com.gu.digital_voucher_cancellation_processor
 
 import java.time.{Clock, Instant, LocalDate}
 
-import cats.{Monad, Show}
 import cats.data.{EitherT, NonEmptyList}
+import cats.implicits._
+import cats.{Monad, Show}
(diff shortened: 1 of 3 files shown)


handlers/digital-voucher-cancellation-processor/src/main/scala/com/gu/digital_voucher_cancellation_processor/DigitalVoucherCancellationProcessorService.scala
```diff
@@ -139,15 +136,18 @@ object DigitalVoucherCancellationProcessorService extends LazyLogging {
...
+    // Limit is because a Salesforce composite response cannot contain more than 25 operations
     s"""
        |SELECT
        |  Id,
        |  SF_Subscription__r.Id
        |FROM
        |  Digital_Voucher__c
-       |WHERE SF_Subscription__r.Cancellation_Effective_Date__c <= ${formatDate(today)}
-       |  AND Cancellation_Processed_At__c = null
+       |WHERE SF_Subscription__r.Cancellation_Effective_Date__c <= TODAY
```
mario-galic, line 147: Cool `TODAY` symbol works in Salesforce!

mario-galic (approved):

PR #768, opened 2020-11-30
Allow optional fields in Imovo redemption history

kelvin-chappell:
When a redemption fails, the address and postcode fields are left empty.


david-pepper (approved):
Looks good

PR #769, opened 2020-11-30
Standalone app to copy to lake an already completed Zuora job

mario-galic:
If for whatever reason zuora batch job succeeded but the CSV files were not copied over, we can try to use this small app to write them to data lake via local machine. This saves the following manual steps

1. Download CSV files `Account.csv`
1. Create metadata `Account.json` file which has correct row count
1. Upload raw csv file 
   `aws s3 cp  Account.csv s3://ophan-raw-zuora-increment-account/ --acl bucket-owner-read --profile membership`
1. Upload metadata file 
   `aws s3 cp  Account.json s3://ophan-raw-zuora-increment-account/metadata/Account.json  --acl bucket-owner-read --profile membership`


kelvin-chappell (approved):
👍

PR #797, opened 2020-12-03
Add alarm to digital voucher cancellation processor

kelvin-chappell:
This will set off an alarm when there's at least one error during the previous 5 minute period.


mario-galic:
Since this deviates from how alarms are declared in the other 20+ lambdas, have you confirmed it works?

kelvin-chappell:
Tested in Dev (with and without the Prod condition on the alarm).

It would actually have been a lot quicker to rewrite the whole thing in Yaml.  But thought I'd better just extend what was there.  We seem to be moving towards CDK in future.

mario-galic (approved):

PR #802, opened 2020-12-04
Update play-json-extensions to 0.42.0

mario-galic:
## What does this change?

In response to scala-steward https://github.com/guardian/support-service-lambdas/pull/799 because it contains breaking change

## How to test

Best to perform end-to-end test via Postman for email-batch-sender.

email-batch-sender end-to-end test succeeding. 

This change would affect `email-batch-sender` which has been known to break.


johnduffell (approved):
👏

mario-galic:
Successfully received email.

PR #808, opened 2020-12-07
Filter out debug entries from DV API log

kelvin-chappell:
The digital voucher API log has debug entries from the AWS SDK and from Netty.  They can obscure what's going on and are generally not very helpful.

This will filter them out.  AWS have no better suggestion for filtering them out.

### DIFF
--- a/handlers/batch-email-sender/src/main/resources/logback.xml
+++ b/handlers/batch-email-sender/src/main/resources/logback.xml
@@ -5,6 +5,10 @@
         </encoder>
     </appender>
 
+    <!-- see https://docs.aws.amazon.com/sdk-for-java/v2/developer-guide/java-dg-logging.html -->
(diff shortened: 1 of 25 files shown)


mario-galic (approved):
Does this mean this files will have to be duplicated 20+ times?

kelvin-chappell:
> Does this mean this files will have to be duplicated 20+ times?
Yes.  I suppose I might as well do them all in one go in this PR.

mario-galic:
Maybe there is a better way...maybe source generation at the point of packaging.

kelvin-chappell:
It's a one-off job and there are other longstanding debug entries leaking into the logs.  The Netty ones have been there for a while.  
All the handlers are already using logback: https://github.com/guardian/support-service-lambdas/blob/main/build.sbt#L253 so they should probably have explicit configuration anyway, which we'll need for our upcoming Json appender.

PR #809, opened 2020-12-07
Update play-json to 2.9.1

mario-galic:
## What does this change?

* Resolves scala-steward https://github.com/guardian/support-service-lambdas/pull/804
* Not sure what `JsonMatchersTest.scala` is supposed to be testing. It looks like it tests a test... 

Changes in `SalesforceToBrazeTransformations.scala` affect batch-email-sender which is worth end-to-end testing as we had issues with it in the past.


kelvin-chappell (approved):
👍

mario-galic:
batch-email-sender end-to-end test passed.

PR #810, opened 2020-12-07
Set correct today date in DiscountTest

mario-galic:
## What does this change?

This test would fail if executed individually with 

```
testOnly com.gu.zuora.subscription.DiscountTest
```

It seems the reason it succeeded when executed as part of all tests `sbt test` is that some previous test by luck set good enough mutable value of via `MutableCalendar.setFakeToday`.


kelvin-chappell (approved):
👍

PR #815, opened 2020-12-09
snyk monitor GitHub Action

mario-galic:
## What does this change?

 Runs `synk monitor` as GH Action in parallel with regular TeamCity build.

Confirmed `synk monitor` works via https://github.com/guardian/tip/pull/78


kelvin-chappell (approved):
👍

PR #839, opened 2020-12-18
[contact-us-api] Add catch-all alarm

ripecosta:
## What does this change?
Adds an alarm that will trigger if no requests are received for 6 hours. This should help us catch any silent issues that might happen.

## Why are we doing this?
A few weeks ago there was a change in MMA that broke the submission of the Contact Us form. This went unnoticed for over 3 days (2 of these were the weekend) because we did not have a way to catch these.

We don't currently have a method in place to test systems end-to-end and there's still a conversation to be had around how to do that. In the meantime this catch-all alarm will help us detect any such issues by alerting us if the it hasn't received any requests for 6 hours.

### DIFF
--- a/handlers/contact-us-api/README.md
+++ b/handlers/contact-us-api/README.md
@@ -34,7 +34,7 @@ Once the request has been processed a response with an appropriate status code w
 }
 ```
 
-## Errors
+## Errors & Alarms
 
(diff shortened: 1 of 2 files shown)


kelvin-chappell (approved):
👍

johnduffell (approved):
fill trick looks good 👍

PR #857, opened 2021-01-06
Log input on holiday-stop-api-failure

mario-galic:
## What does this change?

Old message looks something like

```
Failed to amend Holiday Stop Request for subscription SubscriptionName(A-S01243167) (contact SalesforceContactId(0035I000006Y5WMQA0)): CustomError(MULTIPLE ERRORS : List(LIMIT_EXCEEDED : The request can’t contain more than 25 operations.))
```

whilst new one will also log the `inputThatCausedError`. 

This information could be deduced from existing logs, however it is quite hard as logs can be out-of-order etc., so this PR attempts to simplify debugging a bit.

This extra error information is not returned to the client as part of response.


kelvin-chappell (approved):
👍

PR #858, opened 2021-01-06
Add GH action to test Scala-steward PRs

kelvin-chappell:
This adds an action to test Scala-steward PRs so that we can see if more work is needed before merging.
It will tell us if the code compiles and tests pass but won't tell us about runtime binary incompatibilities or if something dangerous has been imported into the repo.
So it gives more confidence but not complete confidence.

This is done as a GH action rather than in Teamcity to avoid polluting our general CI processes and generating potentially dangerous artefacts.

### DIFF
--- a/.github/workflows/scala-steward.yml
+++ b/.github/workflows/scala-steward.yml
@@ -7,7 +7,7 @@ on:
 jobs:
   test:
     # Assuming Scala-steward branches are always named 'update/...' and other branches aren't.
-    if: startsWith(github.head_ref, 'update/')
+#    if: startsWith(github.head_ref, 'update/')
 
     runs-on: ubuntu-latest
 
--- a/.github/workflows/scala-steward.yml
+++ b/.github/workflows/scala-steward.yml
@@ -12,10 +12,19 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - name: Echo name of source branch
+      - name: Echo significant env vars
         env:
            HEAD_REF: ${{ github.head_ref }}
-        run: echo "$HEAD_REF"
+           REF: ${{ github.ref }}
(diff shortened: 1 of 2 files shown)


.github/workflows/scala-steward.yml
```diff
@@ -0,0 +1,26 @@
+name: Test Scala-steward PRs
+
+on:
+  pull_request:
+    branches: [ main ]
+
+jobs:
+  test:
+    # Assuming Scala-steward branches are always named 'update/...' and other branches aren't.
```
mario-galic, line 9: Is it possible to use the branch name instead as it has `scala-steward` in it which makes it more explicit? For example

```
scala-steward:update/play-json-2.9.2
```

image

kelvin-chappell: [Line 17](https://github.com/guardian/support-service-lambdas/pull/858/files#diff-5f70f29f856428700109959debc5d1f950fa48d279caee7f9f0962a3caa63958R17) is to give me the name of the source branch to check that it's working but it doesn't seem to include the name of the fork so this is the best I can do I think.

Available variables are [here](https://docs.github.com/en/free-pro-team@latest/actions/reference/context-and-expression-syntax-for-github-actions).

mario-galic: Does something like this work:

```
if: ${{ github.event.pull_request.head.repo.full_name == 'scala-steward/support-service-lambdas' }}
```

kelvin-chappell: I thought I'd merge this and see what happens.  Then tighten it up when we get a dump of the extra environment info to find out what's available.  I'm not sure if `repo` refers to source or target.  The easiest way to find out is going to be to wait for another PR to arrive.

PR #873, opened 2021-01-18
Don't try to suspend today's digital vouchers

kelvin-chappell:
This is a workaround for the problem where CSRs can apply holiday stops retrospectively.
If we try to suspend a voucher for today it will fail as Imovo will only accept suspensions from tomorrow onwards.


ripecosta (approved):
👍

PR #885, opened 2021-01-28
add shell of the zuora revenue recogniser job

johnduffell:
This PR adds the shell of the lambda that will find unrecognised revenue in zuora, and recognise it where appropriate.  The code is missing at the moment as that will be in the next PR(s).

---

This is needed because for Digital sub gifts, we only recognise the revenue via the API on redemption.  So accounts will not see the revenue until the subscription itself is running for the recipient.

This means that if the subscription is never redeemed, or if it is cancelled or partially refunded for any reason, the revenue will remain undistributed in zuora, meaning the accounts will be incorrect.  This is used for financial reporting so must always be correct.

The initial aim will be to recognise all the revenue on unredeemed subs one year after purchase.  Also to distribute all refunds immediately (i.e. on the next run).

https://trello.com/c/xH92N3cI/3539-ds-gifts-revenue-recognition-cleanup-daily-job-0

## what does it do when run locally?
image


build.sbt
```diff
@@ -305,8 +310,14 @@ lazy val `zuora-sar` = lambdaProject(
 lazy val `dev-env-cleaner` = lambdaProject(
   "dev-env-cleaner",
   "Cleans up the salesforce to free up storage via 360 sync/zuora",
-  Seq(awsCloudwatch, catsEffect, circeParser, circe)
-).dependsOn(`zuora-reports`, handler, effectsDepIncludingTestFolder, testDep, `effects-s3`, `effects-lambda`)
+  Seq()
```
johnduffell, line 313: I don't know why it had cats effect and circe, must have been an error.  Cloudwatch is extracted as I needed to reuse it.

handlers/revenue-recogniser-job/cfn.yaml
```diff
@@ -0,0 +1,112 @@
...
+    Type: String
+    AllowedValues:
+      - PROD
+      - CODE
+    Default: CODE
+
+Mappings:
+  Stages:
+    Schedule:
+      CODE: 'rate(365 days)'
```
johnduffell, line 17: I don't think we need a CODE deployment running 24/7, but if we do leave it running, we probably don't need it to run often.

kelvin-chappell (approved):
👍

PR #899, opened 2021-02-05
Log failed testCase for testInProdPreviewPublications

mario-galic:
Needed to identify exact subscription that is failing.

Before

```
testInProdPreviewPublications failed because invoicing-api error: ZuoraApiFailure(DecodingFailure at .publicationsWithinRange: Attempt to decode value on failed cursor)
```

After

```
testInProdPreviewPublications failed A-S0000000?startDate=2020-11-13&endDate=2021-02-12 because invoicing-api error: ZuoraApiFailure(DecodingFailure at .publicationsWithinRange: Attempt to decode value on failed cursor)
```


kelvin-chappell (approved):
👍

PR #914, opened 2021-02-16
add query and 1 year writeoff function to recogniser cleanup

johnduffell:
This PR adds the query to find the DS gift subscriptions' revenue schedules that have not been redeemed within a year.  Also to find any refunded DS gift subscriptions' revenue schedules.
It also adds the code to recognise the revenue for the former subscriptions on today's date.
It does not yet distribute the refunds as this is more complicated and will come in the next PR.

This is needed so the finance team can correctly attribute the revenue in to the right months, even if the subscription doesn't follow the standard path of purchase and redemption.

Tested in DEV against DEV zuora using the Manual Test, working okay.

### DIFF
--- a/handlers/revenue-recogniser-job/src/main/scala/com/gu/recogniser/Handler.scala
+++ b/handlers/revenue-recogniser-job/src/main/scala/com/gu/recogniser/Handler.scala
@@ -113,8 +113,7 @@ class Steps(
          |from RevenueSchedule
          |where Rule = 'Digital Subscription Gift Rule'
          | AND (
-         |      UndistributedAmount < 0
-         |  OR (UndistributedAmount > 0 AND subscription.termstartdate <= '$oneYearAgo')
+         |     (UndistributedAmount > 0 AND subscription.termstartdate <= '$oneYearAgo')
          | )
          |""".stripMargin
     for {
--- a/handlers/revenue-recogniser-job/src/main/scala/com/gu/recogniser/Handler.scala
+++ b/handlers/revenue-recogniser-job/src/main/scala/com/gu/recogniser/Handler.scala
@@ -1,49 +1,20 @@
 package com.gu.recogniser
 
+import cats.syntax.all._
 import com.amazonaws.services.lambda.runtime._
 import com.gu.aws.AwsCloudWatch
 import com.gu.aws.AwsCloudWatch._
-import com.gu.effects.RawEffects
-import com.gu.util.config.Stage
+import com.gu.effects.{GetFromS3, RawEffects}
+import com.gu.recogniser.RevenueSchedule.csvFields
+import com.gu.util.config.{LoadConfigModule, Stage}
+import com.gu.util.zuora.{ZuoraRestConfig, ZuoraRestRequestMaker}
+import com.gu.zuora.reports._
+import com.gu.zuora.reports.aqua.ZuoraAquaRequestMaker
+import okhttp3.{Request, Response}
 
-import java.io.{ByteArrayInputStream, ByteArrayOutputStream, InputStream, OutputStream}
+import java.io.{InputStream, OutputStream}
 
 object Handler extends RequestStreamHandler {
-
-  def main(args: Array[String]): Unit = {
-    println("main: STARTING!")
-    // FOR TESTING
-    handleRequest(
-      new ByteArrayInputStream(Array[Byte]()), new ByteArrayOutputStream(), new Context {
-        override def getAwsRequestId: String = ???
-
-        override def getLogGroupName: String = ???
-
-        override def getLogStreamName: String = ???
-
-        override def getFunctionName: String = ???
-
-        override def getFunctionVersion: String = ???
-
-        override def getInvokedFunctionArn: String = ???
-
-        override def getIdentity: CognitoIdentity = ???
-
-        override def getClientContext: ClientContext = ???
-
-        override def getRemainingTimeInMillis: Int = ???
-
-        override def getMemoryLimitInMB: Int = ???
-
-        override def getLogger: LambdaLogger = new LambdaLogger {
-          override def log(message: String): Unit = println("LOG: " + message)
-
-          override def log(message: Array[Byte]): Unit = ???
-        }
-      }
-    )
-    println("main: FINISHED!")
-  }
 
   //referenced in cloudformation, change with care
   def handleRequest(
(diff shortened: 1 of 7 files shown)


handlers/revenue-recogniser-job/src/main/scala/com/gu/recogniser/Handler.scala
```diff
@@ -77,3 +52,87 @@ object Handler extends RequestStreamHandler {
...
+         |  OR (UndistributedAmount > 0 AND subscription.termstartdate <= '$oneYearAgo')
+         | )
+         |""".stripMargin
+    for {
+      undistributedRevenue <- blockingAquaQuery.executeQuery[RevenueSchedule](expiredGiftsAndRefunds).toDisjunction.leftMap(_.toString)
+      revenueSchedulesToDistribute <- undistributedRevenue.toList.sequence.leftMap(_.toString)
+      _ = revenueSchedulesToDistribute.map {
+        case RevenueSchedule(expiredCodeSchedule, amount) if amount > 0 =>
+          log(s"distribute on a single day: $expiredCodeSchedule $amount")
+          distributeRevenueOnSpecificDate.distribute(expiredCodeSchedule, today())
```
johnduffell, line 125: we do it "today" rather than "one year after the start date"
If it runs regularly, that should always be the same thing, but it's possible we might do a batch late, and they would be recognised late.

handlers/revenue-recogniser-job/src/main/scala/com/gu/recogniser/Handler.scala
```diff
@@ -77,3 +52,87 @@ object Handler extends RequestStreamHandler {
...
+        case RevenueSchedule(refundSchedule, amount) if amount < 0 =>
+          log(s"distribute refund: $refundSchedule $amount")
+        //TODO
```
johnduffell, line 128: this will come in the next PR

handlers/revenue-recogniser-job/src/test/scala/com/gu/recogniser/HandlerManualTest.scala
```diff
@@ -0,0 +1,199 @@
...
+      zuoraDeps = ZuoraRestRequestMaker(RawEffects.response, zuoraRestConfig)
+      subNumber <- ZuoraGiftSubscribe.subscribe(zuoraDeps).toDisjunction
+    } yield subNumber
+    println("result: " + actual)
+    println("****\nNOTE: wait a while as the revenue schedule isn't created straight away\n****")
+
+  }
+}
+
+object RunLambdaManualTest {
```
johnduffell, line 33: this is run to test it manually - any ideas about how to make this more obvious to people looking at this later?  I can add to the readme if it would be useful, or add a comment in the main handler?

handlers/revenue-recogniser-job/src/test/scala/com/gu/recogniser/HandlerManualTest.scala
```diff
@@ -0,0 +1,199 @@
...
+      |                "PaypalBaid": "B-23637766K5365543J",
+      |                "PaypalEmail": "test@paypal.com",
+      |                "PaypalType": "ExpressCheckout",
+      |                "Type": "PayPal"
+      |            },
+      |            "SubscriptionData": {
+      |                "RatePlanData": [
+      |                    {
+      |                        "RatePlan": {
+      |                            "ProductRatePlanId": "2c92c0f8778bf8f60177915b477714aa"
```
johnduffell, line 175: this is the new one time charge rate plan.

handlers/revenue-recogniser-job/src/main/scala/com/gu/recogniser/Handler.scala
```diff
@@ -34,13 +35,13 @@ object Handler extends RequestStreamHandler {
     } yield ()
     val _ = maybeSuccess.toTry.get // throws exception if something failed
     log("finished successfully - sending metric!")
-    putMetric(stage)
+    putMetric(stage).get
```
kelvin-chappell, line 38: Doesn't this `get` undermine the change in `putMetric` result type? ie. it will blow up here rather than in line 44

handlers/revenue-recogniser-job/src/test/scala/com/gu/recogniser/HandlerManualTest.scala
kelvin-chappell: Something in the readme might be best.  It would be the first place you'd look if you were trying to work out what the project does.

handlers/revenue-recogniser-job/src/main/scala/com/gu/recogniser/Handler.scala
johnduffell: I don't think so - you can treat any type as Unit, so it will just ignore the result rather than automatically calling `.get` on it if it happens to be a Try.  This was causing a warning hence why I noticed.

The reason I went ahead with the change, is if it fails silently, the alarm would indeed go off, but it wouldn't be clear why from the logs.  So this change will mean that once the alarm does go off, we will see in the logs that the lambda exploded and logged an exception, so we can realise what's going on.
This did bite us recently when I accidentally blew the metrics quota, and the IT-Test runner was logging a rate limit exceeded error, but apart from that everything seemed to be working.  Which was very confusing.

kelvin-chappell (approved):
👍

handlers/revenue-recogniser-job/src/main/scala/com/gu/recogniser/Handler.scala
johnduffell: I have proved it here: https://scalafiddle.io/sf/ijmen2i/1
```scala
import scala.util._
// Start writing your ScalaFiddle code here
def hi(): Unit = Failure(new RuntimeException("boom"))

val x = hi()
println("first is: " + x)


def hi2(): Try[Unit] = Failure(new RuntimeException("boom"))

val xx = hi2().get
println("second is: " + xx)
```
```
first is: undefined
ERROR: undefined
Error
    at $c_jl_RuntimeException.eval [as fillInStackTrace__jl_Throwable] (eval at  (https://scalafiddle.io/resultframe?theme=light:46:37), :2523:14)
.......

```

handlers/revenue-recogniser-job/src/test/scala/com/gu/recogniser/HandlerManualTest.scala
johnduffell: I've added a section to the readme

johnduffell:
I have checked in PROD and it seemed to run okay and found no subscriptions to distribute (as there are no gifts that weren't redeemed for a year)

PR #939, opened 2021-03-02
Run snyk as GitHub Actions

ripecosta:
## What does this change?
This PR adds GitHub Actions to run snyk. This mean we can get rid of snyk integrations elsewhere (TeamCity, GH webhooks). 

`snyk monitor` will run on any push to main. `snyk test` will run on all other branches. A ✅ &nbsp;next to the action means that there were no security vulnerabilities found. An ❌ &nbsp;means security vulnerabilities were found and should be addressed. These action will not block builds or merges but if security vulnerabilities are found they should be addressed as soon as possible.


kelvin-chappell (approved):
👍

PR #941, opened 2021-03-02
Run snyk as GitHub Actions

ripecosta:
## What does this change?
This PR adds GitHub Actions to run snyk. This mean we can get rid of snyk integrations elsewhere (TeamCity, GH webhooks). 

`snyk monitor` will run every day at 6AM and on any push to main. `snyk test` will run on all other branches. A ✅ &nbsp;next to the action in a PR means that there were no security vulnerabilities found. An ❌ &nbsp;means security vulnerabilities were found and should be addressed. These action will not block builds or merges but if security vulnerabilities are found they should be addressed as soon as possible.

### DIFF
--- a/.github/workflows/snyk.yml
+++ b/.github/workflows/snyk.yml
@@ -7,6 +7,7 @@ on:
   schedule:
     - cron: '0 6 * * *'
   push:
+  workflow_dispatch:
 
 jobs:
   security:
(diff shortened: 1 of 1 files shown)


arausuy (approved):

kelvin-chappell (approved):
Very nice!  👍

.github/workflows/snyk.yml
```diff
@@ -0,0 +1,31 @@
+# Run snyk monitor at 6 AM every day (reports vulnerabilities to snyk.io)
+# Run snyk monitor on every push into main branch (reports vulnerabilities to snyk.io)
+# Run snyk test on every push into branches that are not main
+name: Snyk
+
+on:
+  schedule:
+    - cron: '0 6 * * *'
+  push:
```
kelvin-chappell, line 9: Could also run it on demand with the [workflow_dispatch](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#manual-events) trigger, which is useful for testing.

ripecosta: Done!

PR #943, opened 2021-03-03
Snyk GitHub Actions corrections

ripecosta:
## What does this change?
Corrects the project name in the Snyk GitHub Actions as well the name of the branch name in the steps.

### DIFF
--- a/.github/workflows/snyk.yml
+++ b/.github/workflows/snyk.yml
@@ -28,5 +28,5 @@ jobs:
         env:
           SNYK_TOKEN: ${{ secrets.SNYK_TOKEN }}
         with:
-          args: --org=the-guardian-cuu --project-name=guardian/manage-frontend
+          args: --org=the-guardian-cuu --project-name=guardian/support-service-lambdas
           command: monitor
--- a/.github/workflows/snyk.yml
+++ b/.github/workflows/snyk.yml
@@ -28,5 +28,5 @@ jobs:
         env:
           SNYK_TOKEN: ${{ secrets.SNYK_TOKEN }}
         with:
-          args: --org=the-guardian-cuu --project-name=guardian/support-service-lambdas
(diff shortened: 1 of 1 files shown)


.github/workflows/snyk.yml
```diff
@@ -17,16 +17,16 @@ jobs:
...
 
       - name: Run snyk monitor and report vunerabilites to snyk.io
-        if: github.ref == 'ref/head/main'
+        if: github.ref == 'refs/heads/main'
         uses: snyk/actions/scala@0.3.0
         env:
           SNYK_TOKEN: ${{ secrets.SNYK_TOKEN }}
         with:
-          args: --org=the-guardian-cuu --project-name=guardian/manage-frontend
+          args: --org=the-guardian-cuu --project-name=guardian/support-service-lambdas
```
kelvin-chappell, line 31: According to [this](https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#github-context), the repo name is exposed as an env var.  So this might work:
```suggestion
          args: --org=the-guardian-cuu --project-name=${{github.repository}}
```

kelvin-chappell (approved):
👍

kelvin-chappell:
It's a pity the default branch name isn't also available in the Github Action environment but it doesn't seem to be.

ripecosta:
> It's a pity the default branch name isn't also available in the Github Action environment but it doesn't seem to be.
> Although you could create a [workflow template](https://docs.github.com/en/actions/learn-github-actions/sharing-workflows-with-your-organization) and include the default branch name in it for future use in other repos.

That's a great idea! There was a further improvement I wanted to make yesterday for this config but couldn't due to what I think is a bug in Snyk's GH Action implementation. This bug causes it to have no output when you supply a command (test or monitor). I'll wait for snyk's reply to the issue I raised before doing what you suggested.

PR #950, opened 2021-03-08
Add revenue recognition logic for DS gifts that are refunded or expired

johnduffell:
This PR adds extra logic to actually distribute the revenue for DS gifts that have been refunded, or that have expired.

This is needed so that finance team can correctly attribute the revenue from DS gifts even where they were not redeemed or were refunded.

This is very general now, so it would actually distribute the original schedule, as well as any refunds, so we can get rid of this logic: https://github.com/guardian/support-frontend/blob/e4837caea99cb1c63ad8bdd55a142b413d15af1f/support-services/src/main/scala/com/gu/zuora/ZuoraGiftService.scala#L103


handlers/revenue-recogniser-job/src/main/scala/com/gu/recogniser/Handler.scala
```diff
@@ -113,28 +106,142 @@ class Steps(
...
+        expiredGiftsAndRefunds
+      ).toDisjunction.leftMap(_.toString)
       numberNotDone = stillUndistributedRevenue.to(LazyList).length
       _ <- if (numberNotDone > 0) Left(s"there were still revenue schedules: $numberNotDone") else Right(())
     } yield ()
 
   }
 
+  private def distributeForQueryRow(revenueSchedulesToDistribute: List[RevenueScheduleAquaRow]): Either[String, Unit] =
+    revenueSchedulesToDistribute.zipWithIndex.drop(32).traverse {
```
johnduffell, line 130: `drop(32)` is temporary code while I am testing through the list

handlers/revenue-recogniser-job/src/main/scala/com/gu/recogniser/Handler.scala
```diff
@@ -100,41 +98,43 @@ object Steps {
...
-      stillUndistributedRevenue <- blockingAquaQuery.executeQuery[RevenueSchedule](expiredGiftsAndRefunds).toDisjunction.leftMap(_.toString)
-      numberNotDone = stillUndistributedRevenue.to(LazyList).length
-      _ <- if (numberNotDone > 0) Left(s"there were still revenue schedules: $numberNotDone") else Right(())
+      undistributedScheduled <- revenueSchedulesQuerier.execute()
+      (revenueSchedulesToDistributeToday, revenueSchedulesToDistributeRange) = partitionSchedules.partition(undistributedScheduled)
+      _ <- distributeForQueryRow(revenueSchedulesToDistributeToday, revenueSchedulesToDistributeRange)
+//      _ = log("now query again and assert there are none")
+//      stillUndistributedRevenue <- revenueSchedulesQuerier.execute()
+//      numberNotDone = stillUndistributedRevenue.to(LazyList).length
+//      _ <- if (numberNotDone > 0) Left(s"there were still revenue schedules: $numberNotDone") else Right(())
```
johnduffell, line 118: it woudl be nice to have a quick check to make sure none had been missed, but without doing all the getSubscription calls again, this is not easy.  I wonder if we should have a metric for the number of unredeemed-unexpired subscriptions, and alarm if it gets higher than a threshold? Or just wait for finance to notice!

handlers/revenue-recogniser-job/cfn.yaml
```diff
@@ -110,3 +110,29 @@ Resources:
...
+      AlarmActions:
+        - !Sub arn:aws:sns:${AWS::Region}:${AWS::AccountId}:reader-revenue-dev
+      AlarmName: !Join
+        - ' '
+        - - 'non urgent -'
+          - 'revenue recogniser job could not process a record'
+      AlarmDescription: !Join
+        - ' '
+        - - 'Impact - Accounts team will not see all revenue correctly for manually recognised subs'
+          - !FindInMap [ Constants, Alarm, Process ]
```
johnduffell, line 127: todo need to add the alarm to the doc - so that people know what this job is for!

handlers/revenue-recogniser-job/src/main/scala/com/gu/recogniser/Handler.scala
```diff
@@ -27,59 +26,53 @@ object Handler extends RequestStreamHandler {
...
-    putMetric(stage).get
+    putMetric(stage)
   }
 
   /*
   this is alarmed in the cfn
    */
-  def putMetric(stage: Stage): Try[Unit] = {
+  def error(stage: Stage, message: String) = {
+    println(s"FAILURE - sending metric: $message")
```
rupertbates, line 45: This message reads to me at first as 'failure sending metric' could we maybe rephrase to make it clearer what it actually means?

rupertbates (approved):
A little bit 🤯  but really well done for working through it!

PR #1004, opened 2021-04-13
Move the section regarding what to do when zuora export fails to the top

mario-galic:


paulbrown1982 (approved):
👍  Good idea, thank you for doing this!

PR #1007, opened 2021-04-15
Soft Opt-In Consent Setter

david-pepper:
## Problem Description
When a customer acquires or cancels a subscription, implicit consents ('Soft Opt-In Consents') should be set or unset in Identity depending on the product relevant to the subscription.

There are 3 relevant scenarios that the solution addresses:

- Customer acquires a subscription
- Customer cancels a subscription and has no other active subscriptions
- Customer cancels a subscription and 1 or more other active subscriptions

_Soft Opt-In Consents by Product Matrix_
image

### Examples

1. Customer acquires a subscription where the product is membership

Consents to be set: support_onboarding, similar_products, supporter_newsletter

image

2. Customer cancels a subscription where the product is subs card, and customer has no other active subscriptions

Consents to be unset: support_onboarding, similar_products, subscriber_preview, supporter_newsletter

image

3. Customer cancels a subscription where the product is subs card, and customer has another subscription which is Active and the product is membership

Consents to be unset: subscriber_preview

image

## Solution
A lambda will be scheduled to execute every 15 minutes to extract the relevant subscriptions from Salesforce, determine the relevant consents to set or unset depending on the subscription status and product, enact the updates in Identity, and write back success/failure status to Salesforce.

image

### Extraction
A query is performed for all subscriptions where the Soft_Opt_in_Status__c field is either 'Ready to process acquisition' or ‘Ready to process cancellation'. The strategy for surfacing these subscriptions is described in this related PR: [Salesforce Soft Opt In Fields and Layout changes](https://github.com/guardian/salesforce/pull/364).

### Setting consents
For each Identity Id on subscription records where the SF_Status__c is ‘Active’, a callout is made to Identity 
to set the consents associated with the product on the subscription.

### Unsetting consents
For all subscription records where the SF_Status__c is ‘Cancelled’, the associated Identity Ids are used to perform another query to Salesforce to obtain the Active Non-Gift subscriptions associated with each Identity Id.

A determination is made as to which consents should be unset per Identity Id and the relevant callouts per Identity Id are made to Identity to perform the updates.

#### Successful update to Identity
Subscription record in Salesforce will be updated as follows:

- Soft_Opt_in_Last_Stage_Processed__c updated to ‘Acquisition’/’Cancellation’ as appropriate
- Soft_Opt_in_Number_of_Attempts__c set to 0

#### Failed update to Identity
Subscription record in Salesforce will be updated as follows:

- Soft_Opt_in_Last_Stage_Processed__c remains unchanged
- Soft_Opt_in_Number_of_Attempts__c is incremented by 1

### Retries

- There will be a maximum of 5 retries per record. 
- The number of retries is represented in the Soft_Opt_in_Number_of_Attempts__c field on the Salesforce Subscription record. 
- Once this value reaches 5, the formula that determines the value for Soft_Opt_in_Status__c field will set the value to ‘Max number of attempts hit’, and these records will no longer be picked up by the start query. (Start query only picks up records where Soft_Opt_in_Status__c is  'Ready to process acquisition' or ‘Ready to process cancellation').

#### Notification of continual failures
A notification will be sent when the lambda increments the number of retries value from 4 to 5. This feature will be implemented in a subsequent PR.

## Handling System Limitations
Expected daily data volumes to process:

- Acquisitions ~700
- Cancellations ~400

#### REST GET
_Invoked when querying Salesforce for records to process_

The maximum number of records that can be returned in 1 response is 2000. For result sets that contain more than 2000 records, the results will be paginated into responses containing 2000 records with an endpoint provided to obtain the next set of records in the result set.

We decided to limit the number of results returned to 200 for the following reasons:

- We anticipate that for executions every 15 minutes, it is very unlikely that there will be more than 200 records to process.
- In the unlikely event that there are more than 200 records to process, given that the lambda will execute every 15 minutes - any surplus will be picked up in subsequent calls and would be processed within a relatively short period of time 
- When the logic is writing back the success/failure status to the SF Subscription records, the composite call to achieve this has a limitation of 200 records per update call. To process more than 200 records would incur further complexity in the solution to paginate through the extra records.

_Should the need arise in the future to process more than 200 records per transaction, the option to increase the limit and implement recursion to handle pagination is available to us._

#### REST GET
_Invoked when querying Salesforce for Active Non-gift records associated with cancellations_

There is no limitation applied to this query because we use the group by clause to only return the DISTINCT product per Identity Id. It is exceptionally unlikely that this will return a large recordset.
.
#### REST PUT
_Invoked when using composite request to write back success/failure status to Salesforce_

Maximum of 200 records per json body, and limited by the initial record set size as detailed above.

#### HTTP CALLOUTS

Per execution there will be a maximum of 3 callouts to Salesforce
= 12 callouts per hour
= 240 callouts per day

This is will within our daily limit of 420,000 callouts per day, with an average usage per day of approx 10%

#### EXECUTION TIME
Maximum execution time of a lambda is 15 minutes.

During performance testing we validated that execution time is less than 30 seconds from start to finish.

### Functional

Functional testing was performed on the following test scenarios:
	
#### Scenario 1. Customer acquires a subscription

**Description**
A new subscription is taken out by a customer. Customer has no other subscriptions.

**Expected outcome**
- The relevant permissions are set to true in identity
- Soft_Opt_in_Last_Stage_Processed__c field on sub in salesforce is updated to ‘acquisition’
- Soft_Opt_in_Number_of_Attempts__c field on sub in salesforce is set to 0
	
**Actual outcome**
As expected

#### Scenario 2. Customer cancels a subscription

**Description**
Customer cancels subscription. Customer has no other subscriptions.

**Expected outcome**

- The relevant permissions are set to false in identity
- Soft_Opt_in_Last_Stage_Processed__c field on sub in salesforce is updated to ‘cancellation’
- Soft_Opt_in_Number_of_Attempts__c field on sub in salesforce is set to 0

	
**Actual outcome**
As expected

#### Scenario 3. Customer cancels a subscription that has full consents overlap on another active subscription

**Description**
A customer has 2 subscriptions for 2 different products, and the products but have identical associated soft opt-in consents. Customer cancels 1 of the subscriptions.

**Expected outcome**

- No update to Identity takes place
- Soft_Opt_in_Last_Stage_Processed__c field on cancelled sub in salesforce is updated to ‘cancellation’
- Soft_Opt_in_Number_of_Attempts__c field on sub in salesforce is set to 0

	
**Actual outcome**
As expected

#### Scenario 4. Customer cancels a subscription that has partial consents overlap with product on another active subscription

**Description**
A customer has 2 subscriptions for 2 different products, and the products have one or more, but not all, overlapping soft opt-in consents. Customer cancels 1 of the subscriptions.

**Expected outcome**

- The consents that are represented on only 1 of the products are set to false in Identity
- Consents common to both products are not updated in Identity and should remain set to true
- Soft_Opt_in_Last_Stage_Processed__c field on  cancelled sub in salesforce is updated to ‘cancellation’
- Soft_Opt_in_Number_of_Attempts__c field on sub in salesforce is set to 0
	
**Actual outcome**
As expected

#### Scenario 5. Handling a failure to update Identity

**Description**
Identity callout to update consents fails and a response other than 2xx is returned

**Expected outcome**

- No update to Identity takes place
- Soft_Opt_in_Last_Stage_Processed__c field on cancelled sub in salesforce is not updated
- Soft_Opt_in_Number_of_Attempts__c field on cancelled sub in salesforce is incremented by 1

**Actual outcome**
As expected

#### Approach
The performance metric that we were interested in testing was execution time for processing 200 records.

There were 4 variations of 200 records that we were interested, as each involve distinct execution paths:

1. 200 acquisitions
2. 200 cancellations where customer does not have another active product
3. 200 cancellations where customer has another active product
4. 200 records containing a mixture of acquisitions and cancellations

#### Generation of Test Data
Test data was generated in batches of 200 using a postman runner to create Identity Guest Accounts, Salesforce Accounts and Contacts, and Zuora Billing Accounts and Subscriptions. This closely mirrors what occurs in the production environment,

#### Results
Execution time for each scenario was 21-26 seconds, which is a very comfortable time frame, allowing us a lot of flexibility for increasing or decreasing the frequency of lambda execution should the need arise

### Unit
Unit tests are self documenting in the scala tests folder


david-pepper:
[Soft Opt-in Identity Consent Setter](https://trello.com/c/WHfIHklA/2084-soft-opt-in-identity-consent-setter)

handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/models/SoftOptInConfig.scala
```diff
@@ -0,0 +1,69 @@
...
+    )).toRight(
+      SoftOptInError(
+        "SoftOptInConfig",
+        "Could not obtain all config."
+      )
+    )
+  }
+
+  def getConsentsByProductMapping(): Option[Map[String, Set[String]]] = {
+    fetchString(S3Location("kelvin-test", "ConsentsByProductMapping.json")) match {
```
arausuy, line 63: I'd suggest configging this before merging it in

handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/models/SoftOptInConfig.scala
```diff
@@ -0,0 +1,69 @@
...
+  sfClientSecret: String,
+  sfUsername: String,
+  sfPassword: String,
+  sfToken: String
+)
+case class IdentityConfig(identityUrl: String, identityToken: String)
+
+object SoftOptInConfig {
+
+  def get: Either[SoftOptInError, SoftOptInConfig] = {
```
arausuy, line 27: For consistency I'd add brackets into the signature or convert it into a val

handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/models/SoftOptInConfig.scala
```diff
@@ -0,0 +1,69 @@
...
+import scala.util.{Failure, Success}
+
+case class SoftOptInConfig(
+  sfConfig: SalesforceConfig,
+  sfApiVersion: String,
+  identityConfig: IdentityConfig,
+  consentsMapping: Map[String, Set[String]]
+)
+
+case class SalesforceConfig(
```
arausuy, line 15: Is there a shared version of this that exists somewhere else in support-service-lambdas?

handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/SalesforceConnector.scala
```diff
@@ -0,0 +1,97 @@
...
+            Right(())
+          case Left(error) =>
+            Left(SoftOptInError("SalesforceConnector", s"Could not decode SfCompositeRequest.Response: $error. String to decode: $result"))
+        }
+      }
+  }
+
+}
+
+object SalesforceConnector {
```
arausuy, line 75: Is there a shared auth object available in support-service-lambdas?

handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/Main.scala
```diff
@@ -0,0 +1,112 @@
...
+    val recordsToUpdate = getEnhancedCancSubs(cancSubs, activeSubs.records)
+      .map(sub => {
+        buildSfUpdateRequest(sub.cancelledSub, "Cancellation",
+          for {
+            consents <- consentsCalculator.getCancConsents(sub.cancelledSub.Product__c, sub.associatedActiveNonGiftSubs.map(_.Product__c).toSet)
+            _ <- sendCancConsentsIfPresent(identityConnector, sub.identityId, consents, consentsCalculator)
+          } yield ())
+      })
+
+    if (recordsToUpdate.isEmpty) Right(())
```
arausuy, line 62: Consistency around styles for if else would make this easier to revisit later

handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/ConsentsCalculator.scala
```diff
@@ -0,0 +1,58 @@
...
+      .get(productName)
+      .toRight(
+        SoftOptInError(
+          "ConsentsCalculator",
+          s"getAcqConsents couldn't find $productName in consentsMappings"
+        )
+      )
+  }
+
+  def getCancConsents(cancelledProductName: String, ownedProductNames: Set[String]): Either[SoftOptInError, Set[String]] = {
```
arausuy, line 20: Are we gaining anything but shortening cancellation/cancelled here?

handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/ConsentsCalculator.scala
```diff
@@ -0,0 +1,58 @@
+package com.gu.soft_opt_in_consent_setter
+
+import com.gu.soft_opt_in_consent_setter.models.SoftOptInError
+import io.circe.generic.auto._
+import io.circe.syntax.EncoderOps
+
+class ConsentsCalculator(consentsMappings: Map[String, Set[String]]) {
+
+  def getAcqConsents(productName: String): Either[SoftOptInError, Set[String]] = {
```
arausuy, line 9: Are we gaining anything byu shortening acquisitions here/

handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/ConsentsCalculator.scala
```diff
@@ -0,0 +1,58 @@
...
+              s"getCancConsents couldn't find $cancelledProductName in consentsMappings"
+            )
+          )
+          .flatMap(cancelledProductConsents =>
+            Right(cancelledProductConsents.diff(ownedProductConsents)))
+      })
+  }
+
+  def buildConsentsBody(consents: Set[String], state: Boolean): String = {
+    case class consentsObject(id: String, consented: Boolean)
```
arausuy, line 55: Camel case on the class name, also feels a bit off to have a scoped case class like this. Feels like there is an alternative approach, might even be as simple as lifting the case class outside of the method

handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/Main.scala
```diff
@@ -0,0 +1,112 @@
...
+    cancSubsIdentityIds = cancSubs.map(sub => sub.Buyer__r.IdentityID__c)
+
+    activeSubs <- sfConnector.getActiveSubs(cancSubsIdentityIds)
+    _ <- processCancSubs(cancSubs, activeSubs, identityConnector, sfConnector, consentsCalculator)
+  } yield ())
+    .left
+    .map(error => {
+      // TODO: Surface this error outside of the lambda
+      logger.error(s"${error.errorType}: ${error.errorDetails}")
+      println(error)
```
arausuy, line 34: println left in

handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/SalesforceConnector.scala
arausuy, line 75: I'm a little unsure of this pattern of an object doing auth and an instance doing the rest of the calls, why not just have one class that does both?

handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/Main.scala
```diff
@@ -0,0 +1,112 @@
...
+import io.circe.generic.auto._
+import io.circe.syntax.EncoderOps
+
+// TODO: introduce notifications when number of attempts is incremented to 5
+
+object Main extends App with LazyLogging {
+
+  (for {
+    config <- SoftOptInConfig.get
+    sfAuthDetails <- SalesforceConnector.auth(config.sfConfig, HttpRequestUtils.tryRequest)
```
arausuy, line 14: I've popped a comment below about it being a method on an object and an instance, if we're bailing out because we fail to get auth it makes more sense to create the connector, get the token and pass the token through methods. I think I understand the rationale for the current implementation but it doesn't quite feel right.

handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/Main.scala
```diff
@@ -0,0 +1,112 @@
...
+    sfConnector = new SalesforceConnector(sfAuthDetails, config.sfApiVersion, HttpRequestUtils.tryRequest)
+
+    allSubs <- sfConnector.getSfSubs()
+    identityConnector = new IdentityConnector(config.identityConfig, HttpRequestUtils.tryRequest)
+    consentsCalculator = new ConsentsCalculator(config.consentsMapping)
+
+    acqSubs = allSubs.records.filter(_.Soft_Opt_in_Status__c.equals("Ready to process acquisition"))
```
arausuy, line 21: Maybe move these out to vals outside of this, I skimmed over this thinking it was a log line

handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/Main.scala
```diff
@@ -0,0 +1,112 @@
...
+  def failedUpdateToRecordBody(sub: SFSubscription.Record): SFSubscription.UpdateRecord = {
+    SFSubscription.UpdateRecord(
+      Id = sub.Id,
+      Soft_Opt_in_Number_of_Attempts__c = sub.Soft_Opt_in_Number_of_Attempts__c.getOrElse(0) + 1,
+      Soft_Opt_in_Last_Stage_Processed__c = sub.Soft_Opt_in_Last_Stage_Processed__c
+    )
+  }
+
+  def getEnhancedCancSubs(cancSubs: Seq[SFSubscription.Record], associatedSubs: Seq[AssociatedSFSubscription.Record]): Seq[SFSubscription.EnhancedCancelledSub] = {
+    cancSubs.map(cancelledSub => {
```
arausuy, line 101: cancelledSubs as a val is a little clearer

handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/Main.scala
```diff
@@ -0,0 +1,112 @@
+package com.gu.soft_opt_in_consent_setter
```
arausuy, line 1: We have a lot of methods in a main class, refactoring it out into specific classes that display the responsibilities better would improve readability. The mixture of transforms and side effecting functions means this object contains a lot

handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/SfQueries.scala
```diff
@@ -0,0 +1,46 @@
+package com.gu.soft_opt_in_consent_setter
+
+object SfQueries {
+
+  def getAllSubsQuery: String = {
```
arausuy, line 5: this could be a val

handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/SalesforceConnector.scala
ripecosta: The idea was to avoid mutability in the class. To accomplish what we wanted to do (pass the config/authDetails to the class only once) we'd have to keep a variable that mutates over time. But perhaps there are better ways to do this?

handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/ConsentsCalculator.scala
ripecosta: Can you expand on why we shouldn't have a scoped class like that?
This class is only used as a helper to create the JSON string being returned. No other code needs it (or ever will) and only this function uses it.

handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/Main.scala
ripecosta: We plan to do this when we turn this into a lambda. Can I ask you to keep that in mind until our next PR?

handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/Main.scala
ripecosta: To make sure I understand, is your suggestions that we pass the auth details individually to each function in the class instead of passing it once when we instantiate the class?

handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/models/SoftOptInConfig.scala
ripecosta: Not that I have seen. Is there?

handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/SalesforceConnector.scala
ripecosta: Not that I have found. Is there?

ripecosta: Do you mean this?
https://github.com/guardian/support-service-lambdas/blob/main/lib/salesforce/core/src/main/scala/com/gu/salesforce/SFAuthConfig.scala

handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/models/SoftOptInConfig.scala
ripecosta: Do you mean this?
https://github.com/guardian/support-service-lambdas/blob/main/lib/salesforce/core/src/main/scala/com/gu/salesforce/SFAuthConfig.scala

ripecosta: We're now using them. Thanks for the nudge.

handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/SalesforceConnector.scala
ripecosta: We're now using them. Thanks for the nudge.

ripecosta: Me and @arausuy had a chat about this and would benefit from a third opinion. @kelvin-chappell would you mind sharing your expertise on this?

kelvin-chappell: You could move `auth` out of the companion object and into the `SalesforceConnector` class and declare it as a `lazy val` instead of a `def`.  You'd then have to have a `SalesforceConfig` as a field of the class instead of an `SfAuthDetails`.

kelvin-chappell: It's not necessarily an improvement though as logically the auth token is a prerequisite for the connector.  So if you can't get a token you want to fail fast.

handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/SalesforceConnector.scala
```diff
@@ -0,0 +1,97 @@
...
+import io.circe.generic.auto._
+import io.circe.parser.decode
+import scalaj.http.{Http, HttpOptions, HttpRequest, HttpResponse}
+
+class SalesforceConnector(sfAuthDetails: SfAuthDetails, sfApiVersion: String, runRequest: HttpRequest => Either[Throwable, HttpResponse[String]]) extends LazyLogging {
+
+  def doSfGetWithQuery(query: String): Either[SoftOptInError, String] = {
+    runRequest(
+      Http(s"${sfAuthDetails.instance_url}/services/data/$sfApiVersion/query/")
+        .option(HttpOptions.readTimeout(30000))
```
kelvin-chappell, line 15: This is repeated in line 48 so you could define a private class method that includes all the fields you include in every HTTP request and call that instead.

handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/ConsentsCalculator.scala
ripecosta: Spoke about this: We're not sure if the case class is redefined everything the function is called so it's safer that the class is defined outside the function.

handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/SalesforceConnector.scala
ripecosta: I figured this would be a good candidate for using an implicit class to add a method to the HttpRequest class, so that's what I did. Let me know what you think about this approach.

ripecosta: We spoke about this and came to the conclusion that using an implicit is overkill so I'm implementing a private function instead.

kelvin-chappell (approved):
Looking much better now.  I just had a few small comments but structure looks good and code is good. 👍

handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/Main.scala
ripecosta: We actually went ahead and made changes to the code in the main class. Have a second look and let us know if you have any thoughts.

handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/SalesforceConnector.scala
```diff
@@ -0,0 +1,130 @@
...
+  }
+
+  private def sfHttp(url: String): HttpRequest = {
+    Http(url)
+      .option(HttpOptions.readTimeout(30000))
+      .header("Authorization", s"Bearer ${sfAuthDetails.access_token}")
+      .header("Content-Type", "application/json")
+  }
+
+  def sendQueryReq(query: String): Either[Throwable, HttpResponse[String]] = {
```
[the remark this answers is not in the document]
ripecosta: Spoke to @kelvin-chappell about this point.
Ideally we'd want `sendQueryReq` and `handleQueryResp` to be private, as they are only used in this class.
However, making `handleQueryResp` private would make it harder to test and making `sendQueryReq` private by itself wouldn't make much sense because these two functions are closely related and it would create a weird rule (methods should be private unless external code needs to access them or they need to be tested).
There was also the option to make these functions private for the package, but in this case that would not bring any benefit.
We agreed that for this case it would be best to leave them as public.

handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/IdentityConnector.scala
```diff
@@ -0,0 +1,41 @@
...
+      Http(url)
+        .header("Content-Type", "application/json")
+        .header("Authorization", s"Bearer ${config.identityToken}")
+        .postData(body)
+        .method("PATCH")
+        .asString
+    ).toEither
+  }
+
+  def handleConsentsResp(response: Either[Throwable, HttpResponse[String]], errorDesc: String = ""): Either[SoftOptInError, Unit] = {
```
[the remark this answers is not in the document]
ripecosta: I understand your point but feel that turning it into an `Option[String]` would increase the complexity of the function code without offering a corresponding benefit. The intended result here is merely to offer two signatures for this method in a simplified manner, which it is doing. However, if the core logic of the function depended the presence of `errorDesc` I would agree that turning it into an `Option[String]` would be clearer.

handlers/soft-opt-in-consent-setter/src/test/scala/com.gu.soft_opt_in_consent_setter/ConsentsCalculatorTests.scala
```diff
@@ -0,0 +1,106 @@
...
+import org.scalatest.flatspec.AnyFlatSpec
+import org.scalatest.matchers.should
+
+class ConsentsCalculatorTests extends AnyFlatSpec with should.Matchers with EitherValues {
+
+  val calculator = new ConsentsCalculator(testConsentMappings)
+
+  // getAcquisitionConsents success cases
+  "getAcquisitionConsents" should "correctly return the mapping when a known product is passed" in {
+    calculator.getAcquisitionConsents("membership") shouldBe Right(membershipMapping)
```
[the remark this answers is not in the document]
ripecosta: Yes, it does. However if the function being tested returns a `Left`, the application throws an exception instead of providing a nice readable message of what it expected vs what it got.

handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/IdentityConnector.scala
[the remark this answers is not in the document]
kelvin-chappell: 👍

handlers/soft-opt-in-consent-setter/src/test/scala/com.gu.soft_opt_in_consent_setter/ConsentsCalculatorTests.scala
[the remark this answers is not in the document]
kelvin-chappell: 👍

PR #1015, opened 2021-04-21
Remove Payment Card query from Data Lake extract

david-pepper:
### Problem Description

We intend to delete the Payment_Card__c object and its records from Salesforce. 

From inspecting the logic it appears that the query is written but not executed by verifying that aws triggers have no actions associated with extracted data and there are no subsequent usages or references to payment card. This is corroborated by inspecting Bulk Data Load Jobs in Salesforce and observing that none are performed on the Payment Card object. 

References to this object in the Datalake extract logic are therefore unnecessary, so removing them here.


ripecosta (approved):
👍

PR #1035, opened 2021-05-05
Productionise Soft Opt-In Consent Setter Lambda

david-pepper:
## What does this change?
This PR contains the necessary files to create the Soft Opt-In Consent Setter Lambda and associated configuration for deployment

new files:
- cfn.yaml
- riffraff.yaml

Updated files:
- SoftOptInConfig.scala 
updated to dynamically obtain the stage variable used to retrieve the Stage-specific ConsentsByProductMapping.json file from S3

- Main 
updated to align with naming standards and a handler method has been added for the Lambda to hook into

## How to test and measuring success
- The configuration was used to deploy the solution to the CODE environment. 
- The Lambda was executed to operate on a subset of records in Salesforce. The Identity consents were successfully written and results written back to Salesforce as expected.

Notes:
- There is no Identity DEV environment - only CODE, which points to both Salesforce DEV and UAT
- This represented in the Stage mappings in cfn.yaml

### DIFF
--- a/handlers/soft-opt-in-consent-setter/cfn.yaml
+++ b/handlers/soft-opt-in-consent-setter/cfn.yaml
@@ -123,3 +123,9 @@ Resources:
             - Effect: Allow
               Action: cloudwatch:PutMetricData
               Resource: "*"
+        - Statement:
+            - Sid: readDeployedArtefact
+              Effect: Allow
+              Action: s3:GetObject
+              Resource:
+                - arn:aws:s3::*:membership-dist/*
--- a/handlers/soft-opt-in-consent-setter/cfn.yaml
+++ b/handlers/soft-opt-in-consent-setter/cfn.yaml
@@ -16,6 +16,13 @@ Mappings:
       Schedule: 'rate(12 hours)'
     CODE:
       Schedule: 'rate(365 days)'
+      SalesforceStage: DEV
+      IdentityStage: CODE
+      SalesforceUsername: SoftOptInConsentSetterAPIUser
+      AppName: AwsConnectorSandbox
+      AppSecretsVersion: b7b63e09-f39f-4e4f-80ef-9dd203c2d59b
+      SalesforceUserSecretsVersion: 08036ddd-9fc0-43b1-b478-4023a936751e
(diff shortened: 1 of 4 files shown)


handlers/soft-opt-in-consent-setter/cfn.yaml
```diff
@@ -0,0 +1,125 @@
...
+      Policies:
+        - Statement:
+            - Effect: Allow
+              Action: s3:GetObject
+              Resource:
+                - !Sub arn:aws:s3:::soft-opt-in-consent-setter/${Stage}/ConsentsByProductMapping.json
+        - Statement:
+            - Effect: Allow
+              Action: cloudwatch:PutMetricData
+              Resource: "*"
```
kelvin-chappell, line 125: As the membership-dist bucket will soon be private, will also need a policy to read from the distribution bucket.  Like this:
```
        - Statement:
            - Sid: readDeployedArtefact
              Effect: Allow
              Action: s3:GetObject
              Resource:
```

kelvin-chappell (approved):
Looks good! 👍

handlers/soft-opt-in-consent-setter/cfn.yaml
david-pepper: Thanks Kelvin, I've added this policy to cfn.yaml and pushed the change

ripecosta: I thought the CloudFormation got the S3 file and it would use the permissions from the account running the CloudFormation to fetch the artefact from S3. Is that not the case?

kelvin-chappell: When you tell a lambda where to find its code, the lambda has to have read access to the code.
So, as I understand it, the `UpdateS3Lambda` Riffraff step just nudges the lambda to update its code.

You could test this theory by logging in to the AWS console and try updating a lambda with code in an S3 bucket it's not allowed to read but you are allowed to read.  It should be the permission that the lambda has rather than your permissions that apply.

ripecosta: Got it. Thanks!

PR #1040, opened 2021-05-07
Move deployed zuora-retention jar

kelvin-chappell:
To make the zuora-retention stack consistent with other stacks in this repo, this moves its deployed jar from the `membership-dist` bucket to the `support-service-lambdas-dist` bucket.


ripecosta (approved):
👍

PR #1058, opened 2021-05-18
Upgrade STTP to STTP v3 in Imovo Client

kelvin-chappell:
This upgrade is required to fix a few Snyk vulnerabilities, as well as getting onto the latest code.
There will be more PRs to come to upgrade uses of STTP in other modules.


kelvin-chappell:
Superseded by #1190

PR #1084, opened 2021-06-03
Increase period of holiday stop processor alarm

kelvin-chappell:
It was 5 minutes but it's impossible for the lambda to fail twice in 5 minutes as each failure is terminal.
So increasing to 1 hour will give us a meaningful value.  Now the alarm should go off when there are two or more failures in an hour.


jambo1 (approved):
Looks good

PR #1086, opened 2021-06-04
Create payment-failure-comms skeleton

ripecosta:
## What does this change?
Creates a skeleton for the payment-failure-comms project, setting up basic CI/CD and CloudFormation config. This will allow easier testing of the project as it's built.

### DIFF
--- a/handlers/payment-failure-comms/cfn.yaml
+++ b/handlers/payment-failure-comms/cfn.yaml
@@ -1,0 +1,52 @@
+Transform: AWS::Serverless-2016-10-31
+
+Parameters:
+  Stage:
+    Description: Stage name
+    Type: String
+    AllowedValues:
+      - PROD
+      - CODE
+      - DEV
+    Default: CODE
+
+Conditions:
+  IsProd: !Equals [ !Ref Stage, PROD ]
+
+
+Resources:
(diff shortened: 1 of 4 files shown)


kelvin-chappell (approved):
👍

paulbrown1982 (approved):
👍  LGTM too

handlers/payment-failure-comms/cfn.yaml
```diff
@@ -0,0 +1,62 @@
...
+Conditions:
+  IsProd: !Equals [ !Ref Stage, PROD ]
+
+
+Resources:
+  LogGroup:
+    Type: AWS::Logs::LogGroup
+    Properties:
+      LogGroupName: !Sub /aws/lambda/${AppName}-${Stage}
+      RetentionInDays: 90
```
johnduffell, line 25: if you have 90 days, you must make sure not to log any potential PII, (otherwise these logs would need adding to the data map and fulfil subject access and erasure requests)
the other option is reduce to 28 days or less

ripecosta: Thanks for the information, John. I might need to log PII when there are errors (body of a request, for example), in which case I will change it to 28 days.

PR #1100, opened 2021-06-16
Remove API key config from payment failure comms

ripecosta:
## What does this change?
Removes the API key CoudFormation config from payment-failure-comms.

Zuora does not offer the feature to authenticate using an API key so we will need to authenticate using a different way. For now though as this is a Proof of Concept we will leave it open.


kelvin-chappell (approved):
😞

ripecosta:
> 😞

Don't be sad, Kelvin. This is just temporary. We shall not be defeated!

jambo1 (approved):
Looks good!

PR #1101, opened 2021-06-16
Add GW monthly as a supported product for holiday stops

kelvin-chappell:
This change should enable holiday stops to be set up in MMA and Salesforce for the new GW monthly subscription type.

Tested on manage.code.dev-theguardian.com.


ripecosta (approved):
👍

PR #1109, opened 2021-06-21
Add Config to payment-failure-comms

ripecosta:
## What does this change?
Add code to fetch config in the payment-failure-comms lambda. Also adds a Failure object that will be expanded in the future.

## How to test
To test send a message to the API Gateway URL and looks at the logs to confirm that the config was read correctly.


handlers/payment-failure-comms/src/main/scala/com.gu.payment_failure_comms/models/Config.scala
```diff
@@ -0,0 +1,20 @@
...
+  def apply(): Either[Failure, Config] = {
+    (for {
+      instanceUrl <- sys.env.get("brazeInstanceUrl")
+      bearerToken <- sys.env.get("brazeBearerToken")
+      zuoraAppId <- sys.env.get("zuoraAppIdForBraze")
+    } yield Config(
+      BrazeConfig(instanceUrl, bearerToken, zuoraAppId)
+    ))
+      .toRight(
+        ConfigFailure("Could not obtain all config.")
```
kelvin-chappell, line 17: Perhaps say which config you couldn't get.
(By putting the toRight on each statement in the for-comprehension.)

handlers/payment-failure-comms/src/main/scala/com.gu.payment_failure_comms/Handler.scala
```diff
@@ -2,13 +2,20 @@ package com.gu.payment_failure_comms
...
+import com.gu.payment_failure_comms.models.Config
 import com.gu.util.Logging
 
 class Handler extends RequestHandler[APIGatewayProxyRequestEvent, APIGatewayProxyResponseEvent] with Logging {
 
   def handleRequest(event: APIGatewayProxyRequestEvent, context: Context): APIGatewayProxyResponseEvent = {
-    logger.info("Received request with body: " + event.getBody)
+    for {
+      config <- Config()
+    } yield {
```
kelvin-chappell, line 13: If it's just a single statement, would be simpler as:
`Config().map(config => ...)`
but not much in it.

handlers/payment-failure-comms/src/main/scala/com.gu.payment_failure_comms/Handler.scala
```diff
@@ -2,13 +2,20 @@ package com.gu.payment_failure_comms
...
+      logger.info(s"Read config with: ${config.braze.instanceUrl} ${config.braze.bearerToken.charAt(0)} ${config.braze.zuoraAppId}")
```
kelvin-chappell, line 14: What's the idea with the first char of the token?

handlers/payment-failure-comms/src/main/scala/com.gu.payment_failure_comms/Handler.scala
ripecosta: For now it is a single statement but it will get expanded on the next PR so I figured I'd start using the for comprehension now.

handlers/payment-failure-comms/src/main/scala/com.gu.payment_failure_comms/Handler.scala
ripecosta: As a security precaution I avoided logging a token but still wanted to test I was getting it right so I printed the first char. However this was just for testing purposes and it will get removed in a subsequent PR.

kelvin-chappell (approved):
👍

handlers/payment-failure-comms/src/main/scala/com.gu.payment_failure_comms/models/Config.scala
ripecosta: That's a great idea! I've now done that.

handlers/payment-failure-comms/src/main/scala/com.gu.payment_failure_comms/models/Config.scala
```diff
@@ -6,15 +6,15 @@ case class BrazeConfig(instanceUrl: String, bearerToken: String, zuoraAppId: Str
 
 object Config {
   def apply(): Either[Failure, Config] = {
-    (for {
+    for {
```
kelvin-chappell, line 9: Not necessary, but you could even have had a small helper method here to avoid duplication:
```
def get(prop: String) = sys.env.get(prop).toRight(ConfigFailure(s"Could not obtain $prop"))
```
Then
```
...
instanceUrl <- get("brazeInstanceUrl")
...
```

ripecosta: Not necessary but you are right, it's so much cleaner that way. I'll do that separately since this one is already merged. Thanks!

ripecosta: PR opened here #1119

PR #1110, opened 2021-06-21
Add BrazeConnector to payment-failure-comms

ripecosta:
## What does this change?
Adds code for inserting Custom Events into Braze. Also decreases the log retention time to 14 because we cannot guarantee that Braze won't echo personal data in the event of an error (Thanks @johnduffell for the [tip](https://github.com/guardian/support-service-lambdas/pull/1086#discussion_r646508094)).

This code isn't actually in use right now. The next PR will plug everything together. Merging this separately to make it easier to review.

Intended use: A `CustomEvent` will be created based on the messages received from Zuora and `Config`. This will then be encoded into JSON and fed into `BrazeConnector.sendCustomEvent`. which will add it to Braze. `Unit` or `Failure` will be returned depending on whether the request was successful or not.

### DIFF
--- a/handlers/payment-failure-comms/cfn.yaml
+++ b/handlers/payment-failure-comms/cfn.yaml
@@ -36,7 +36,7 @@ Resources:
     Type: AWS::Logs::LogGroup
     Properties:
       LogGroupName: !Sub /aws/lambda/${AppName}-${Stage}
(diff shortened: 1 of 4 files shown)


kelvin-chappell (approved):
👍

PR #1119, opened 2021-06-23
Neater Config getter

ripecosta:
## What does this change?
Improvement to how we get env variables so it's cleaner and easier to read.

### DIFF
--- a/handlers/payment-failure-comms/src/main/scala/com.gu.payment_failure_comms/models/Config.scala
+++ b/handlers/payment-failure-comms/src/main/scala/com.gu.payment_failure_comms/models/Config.scala
@@ -6,6 +6,9 @@ case class BrazeConfig(instanceUrl: String, bearerToken: String, zuoraAppId: Str
 
 object Config {
   def apply(): Either[Failure, Config] = {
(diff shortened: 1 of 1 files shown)


handlers/payment-failure-comms/src/main/scala/com.gu.payment_failure_comms/models/Config.scala
```diff
@@ -7,14 +7,14 @@ case class BrazeConfig(instanceUrl: String, bearerToken: String, zuoraAppId: Str
...
+      instanceUrl <- getFromEnv("brazeInstanceUrl")
+      bearerToken <- getFromEnv("brazeBearerToken")
+      zuoraAppId <- getFromEnv("zuoraAppIdForBraze")
     } yield Config(
       BrazeConfig(instanceUrl, bearerToken, zuoraAppId)
     )
   }
+
+  private def getFromEnv(prop: String): Either[ConfigFailure, String] =
+    sys.env.get(prop).toRight(ConfigFailure(s"Could not obtain $prop"))
```
kelvin-chappell, line 19: You could nest this method inside the `apply` method if it isn't going to be used elsewhere

kelvin-chappell (approved):
👍

PR #1125, opened 2021-06-25
Soft Opt-In Consent Setter: Final Release

ripecosta:
## What does this change?
Releases the Soft Opt-In Consent Setter lambda in normal operation mode.

Reduces the rate at which the lambda runs to every 30 minutes. Logs shows the lambda never processed more than 26 subscriptions per run except for a peak of 200 that happened at midnight, meaning we can decrease the rate. We should continue to monitor though.

Reduces lambda memory size to 256MB. Logs show the most this lambda has used is 203MB so 256MB should a safe limit.


kelvin-chappell (approved):
👍

PR #1126, opened 2021-06-25
Payment Failure Comms: Add core logic

ripecosta:
## What does this change?
Adds core logic to Payment Failure Comms lambda.

The lambda expects to receive a JSON object with the following format:
```
{
    "identityId": String,
    "event": String, // "payment_failure" OR "payment_recovery"
    "properties": {
        "currency": String,
        "amount": Double
    }
}
```

It returns status code 200 for successful requests and 500 for unsuccessful ones.

Note: It currently uses a specific `brazeId` but that is to be changed in a subsequent PR so that the lambda will instead fetch it from Identity using the `IdentityId`


handlers/payment-failure-comms/src/main/scala/com.gu.payment_failure_comms/Handler.scala
```diff
@@ -2,20 +2,63 @@ package com.gu.payment_failure_comms
...
+        CustomEvent(
+          external_id = brazeId,
+          app_id = zuoraAppId,
+          name = request.event,
+          time = DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ss:SSSZ").format(ZonedDateTime.now),
+          properties = request.properties
+        )
+      )
+    )
+  }
```
kelvin-chappell, line 62: I would put this in a companion object to the `BrazeTrackRequest` case class.

handlers/payment-failure-comms/src/main/scala/com.gu.payment_failure_comms/Handler.scala
```diff
@@ -2,20 +2,63 @@ package com.gu.payment_failure_comms
...
+      request <- decodeMessage(event.getBody)
+      brazeRequest = convertToEventsRequest(
+        request = request,
+        brazeId = "1bd449f3-4326-4ae5-89dd-35cce2d50944",
+        zuoraAppId = config.braze.zuoraAppId)
+      _ <- BrazeConnector.sendCustomEvent(config.braze, brazeRequest.asJson.toString)
+    } yield ()) match {
+      case Right(_) =>
+        logger.info(s"The request was successful.")
+        new APIGatewayProxyResponseEvent().withStatusCode(200)
```
kelvin-chappell, line 31: Not much in it but this block could be included in the for-comprehension:
```
   ...
    _ <- logger.info(s"The request was successful.")
 } yield new APIGatewayProxyResponseEvent().withStatusCode(200)
```
Then do a `.left.map(...` on the result of the for-comprehension.

handlers/payment-failure-comms/src/main/scala/com.gu.payment_failure_comms/Handler.scala
```diff
@@ -2,20 +2,63 @@ package com.gu.payment_failure_comms
...
+
+    (for {
       config <- Config()
-    } yield {
-      logger.info(s"Read config with: ${config.braze.instanceUrl} ${config.braze.bearerToken.charAt(0)} ${config.braze.zuoraAppId}")
-      logger.info("Received request with body: " + event.getBody)
+      request <- decodeMessage(event.getBody)
+      brazeRequest = convertToEventsRequest(
+        request = request,
+        brazeId = "1bd449f3-4326-4ae5-89dd-35cce2d50944",
```
kelvin-chappell, line 25: Shouldn't this ID be in the Config?

ripecosta: This is ID is here only temporarily. The next PR will replace it with a lookup to IDAPI. It's here for testing pruposes.

kelvin-chappell: 👍

handlers/payment-failure-comms/src/main/scala/com.gu.payment_failure_comms/Handler.scala
ripecosta: If I did that wouldn't I still need the match (or equivalent) to extract the value from the Either?

kelvin-chappell: No, it would look like:
```
    .left.map { failure =>
        logger.error(s"An error occurred. ${failure.kind}: ${failure.details}")
        new APIGatewayProxyResponseEvent().withStatusCode(500)
    }
```
but it's not much different.

kelvin-chappell (approved):
👍

PR #1137, opened 2021-07-05
Raising memory for soft opt in consent setter to 512 due to out of me…

jambo1:
…mory erros in line with https://github.com/guardian/ophan-housekeeper/issues/4

## What does this change?
Soft opt-in-consent-setter lambda is getting out of memory errors due to Metaspace issues. It's believed to be a duplicate of this issue https://github.com/guardian/ophan-housekeeper/issues/4 so raising the memory from 256 to 512, as has been done elsewhere, should resolve the issue.


kelvin-chappell (approved):
👍

PR #1138, opened 2021-07-05
Switching contact us alarm to only match prod env to prevent noise alarms

jambo1:
## What does this change?
This alarm tells us no requests have been received in the last 6 hours to the contact us API. 

This has been running without issue in prod since December so alerting to no activity in other environments now seems to be creating noise. Leaving this alarm running in prod as we have no e2e tests for this at present.


kelvin-chappell (approved):
👍

PR #1153, opened 2021-07-15
Update cfn.yaml

jambo1:
Returning to openjdk 8 for testing of why export failed for out of memory heap error which happened after upgrading to amazon linux 2 image


kelvin-chappell (approved):
👍

kelvin-chappell (changes requested):
Superseded by #1154

PR #1155, opened 2021-07-15
Increase memory available for Java Corretto

kelvin-chappell:
The Salesforce export download step is failing with an out-of-memory error when using Corretto.
Increasing the memory available to the lambda sorts it out.

This has been tested in Prod.


jambo1 (approved):
Looks good 👍

PR #1178, opened 2021-08-24
Stripe payment intent webhook

TomPretty:
## What does this change?
Add a new project `stripe-webhook-endpoints` with a single endpoint `payment-intent-issues`. This endpoint listens to issues with stripe payment intents. This is so that we can respond to failed `SEPA` payments and update zuora accordingly. 

## TODO

- [x] send proper data to zuora reconciliation api
- [x] add alarm


handlers/stripe-webhook-endpoints/src/test/scala/com/gu/paymentIntentIssues/LambdaSpec.scala
```diff
@@ -0,0 +1,232 @@
+package com.gu.paymentIntentIssues
+
+import org.scalatest.matchers.should.Matchers
+import org.scalatest.flatspec.AnyFlatSpec
+import org.scalatest.Ignore
+import scala.jdk.CollectionConverters._
+import com.stripe.net.Webhook
+
+@Ignore
+class LambdaSpec extends AnyFlatSpec with Matchers {
```
TomPretty, line 10: We should think more generally what we want to test here with this webhook. This is here for now as a way to run locally. We **should not** merge this in this form.

handlers/stripe-webhook-endpoints/src/main/scala/com/gu/paymentIntentIssues/Lambda.scala
```diff
@@ -0,0 +1,129 @@
...
+    }
+    response
+  }
+
+  def loadConfig(identity: AppIdentity): EitherT[IO, Error, Config] =
+      ConfigLoader.loadConfig[IO, Config](identity).leftMap(e => ConfigLoadingError(e.message))
+
+  def getPayload(event: APIGatewayProxyRequestEvent, endpointSecret: String): Either[Error, String] =
+    for {
+      payload <- Option(event.getBody()).toRight(InvalidRequestError("Missing body"))
```
paulbrown1982, line 59: If `getBody()` returns a nullable string then this logic won't quite work if the string is empty rather than null.

handlers/stripe-webhook-endpoints/src/main/scala/com/gu/paymentIntentIssues/Lambda.scala
```diff
@@ -0,0 +1,129 @@
...
+
+  def getAccessToken(oauthConfig: ZuoraRestOauthConfig, backend: SttpBackend[Id, Nothing]): Either[Error, AccessToken] =
+    accessTokenGetResponseV2(oauthConfig, backend).left.map(e => ZuoraApiError(e.reason))
+
+  def queryPayments(paymentNumber: String, config: ZuoraRestConfig, backend: SttpBackend[Id, Nothing]): Either[Error, ZuoraPaymentQueryResponse] = {
+    implicit val b = backend
+
+    sttp.post(uri"${config.baseUrl}/action/query")
+      .header("Authorization", s"Bearer ${config.accessToken}")
+      .body(s"{\"queryString\": \"select Id from Payment where PaymentNumber = '$paymentNumber'\" }")
```
paulbrown1982, line 105: You can use tripe-quote strings `"""` to save the need to use `\"`

handlers/stripe-webhook-endpoints/src/test/scala/com/gu/paymentIntentIssues/LambdaSpec.scala
```diff
@@ -0,0 +1,232 @@
...
+
+    val result = Lambda.processEvent(json, config)
+
+    result should be(Right(()))
+  }
+
+  def getConfig =
+    for {
+      endpointSecret <- sys.env.get("endpointSecret")
+      zuoraBaseUrl = "https://rest.apisandbox.zuora.com/v1"
```
paulbrown1982, line 24: This URL will also need to come from an env config

handlers/stripe-webhook-endpoints/src/main/scala/com/gu/paymentIntentIssues/Lambda.scala
TomPretty: That's much nicer!

handlers/stripe-webhook-endpoints/src/test/scala/com/gu/paymentIntentIssues/LambdaSpec.scala
TomPretty: This is just hardcoded in the test - the actual lambda will read from ssm

handlers/stripe-webhook-endpoints/src/main/scala/com/gu/paymentIntentIssues/Lambda.scala
TomPretty: If body were an empty string, then we'd have `payload=""`. I think this would then fail at the `verifyHeader` call on line 61

handlers/stripe-webhook-endpoints/src/test/scala/com/gu/paymentIntentIssues/LambdaSpec.scala
TomPretty: Discussed offline - we've decided to keep this around for as it's a nice way to run locally, but will add some comments to explain that it's not really a test

TomPretty: Update again - we've now moved this from a spec file into a one for running locally `Local.scala`. We're still keeping it around for reference, but ideally we won't need to run it often

tomrf1 (approved):
great!

PR #1190, opened 2021-09-09
Upgrade STTP to v3

kelvin-chappell:
## What does this change?
This is an upgrade of the [STTP](https://sttp.softwaremill.com/en/latest) dependency from v1 to v3.
I've largely followed this [migration guide](https://blog.softwaremill.com/migrating-to-sttp-client-2-x-and-tapir-0-12-x-7956e6c79c52) to carry it out.
The main changes are:
* The STTP package name has changed.
* The names of some types have changed to avoid conflicts with Cats types.
* Calls are now made on the `basicRequest` object rather than the `sttp` object.
* There has been a change in the response model of an HTTP call.  From v2 onwards, it returns an `Either[Failure, Success]` rather than a success case only which makes some of the Json handling a little easier.
* The `AsyncHttpClientCatsBackend` now has to be given an underlying implementation, which here is always the `DefaultAsyncHttpClient`.
* The `CatsMonadError` has been removed from the library.

There are also some formatting changes that I didn't choose.  They were imposed by Scalariform.

I would like to have done this module by module but that was impossible due to the tight coupling of a lot of the codebase.  There was no single module that I could work on in isolation of all others so I had to do the whole thing in one go.

## How to test
There are three basic uses of STTP HTTP clients in the codebase:
* Imovo API calls
* Zuora API calls
* Salesforce API calls

I've tested them all by running these lambdas in Code:
* Holiday-stop credit processor
* Delivery-problem credit processor
* Digital voucher API
* Digital voucher suspension processor

That should give a reasonable degree of confidence as, between them, these lambdas make a variety of calls on each of the underlying APIs.

This should eliminate some vulnerabilities on the Snyk dashboard.

There are plenty of alarms on our apps and as there are three basic uses of STTP we should see any defects quite quickly.


johnduffell:
nice PR description 👏

handlers/delivery-records-api/src/test/scala/com/gu/delivery_records_api/DeliveryRecordsApiTest.scala
```diff
@@ -353,52 +358,56 @@ class DeliveryRecordsApiTest extends AnyFlatSpec with Matchers with EitherValues
...
-      SttpBackendStub[IO, Nothing](new CatsMonadError[IO])
+      SttpBackendStub[IO, Nothing](new CatsMonadAsyncError[IO])
         .stubAuth(config, auth)
 
     val app = createApp(salesforceBackendStub)
     val response = app.run(
       Request(
         method = Method.GET,
-        Uri(path = s"/delivery-records/${subscriptionNumber}")
+        Uri(path = s"/delivery-records/${ subscriptionNumber }")
```
johnduffell, line 386: is that scalariform? I've never seen that spacing before

lib/salesforce/sttp-client/src/test/resources/subscription-query-response1.json
```diff
@@ -9,7 +9,7 @@
         "url": "/services/data/v29.0/sobjects/SF_Subscription__c/a2F3E000001FhjuUAC"
       },
       "Id": "000000001",
-      "CreatedDate": "2019-11-18T16:59:24.000+0000",
+      "CreatedDate": "2019-11-18T16:59:24Z",
```
johnduffell, line 12: is this an actual change in what salesforce returns?

lib/salesforce/sttp-client/src/test/scala/com/gu/salesforce/sttp/SalesforceClientTest.scala
```diff
@@ -1,31 +1,30 @@
...
+
   case class FixtureParam(
-    config: SFAuthConfig,
-    auth: SalesforceAuth,
-    backendStub: SttpBackendStub[IO, Nothing]
-  )
+                           config: SFAuthConfig,
+                           auth: SalesforceAuth,
+                           backendStub: SttpBackendStub[IO, Nothing]
+                         )
```
johnduffell, line 27: does scalaform do this?  I'm not keen on any indentation depth that varies on the length of the name of the class/function as it easily gets messy

johnduffell (approved):
great work, I don't claim to understand STTP or cats effect etc, but it's interesting to have a look at the code!  I've picked out a few comments or questions but nothing to stop this PR 👍

lib/salesforce/sttp-client/src/test/scala/com/gu/salesforce/sttp/SalesforceClientTest.scala
```diff
@@ -1,31 +1,30 @@
...
 class SalesforceClientTest extends flatspec.FixtureAnyFlatSpec with Matchers {
 
+  private implicit val contextShift = IO.contextShift(ExecutionContext.global)
+
   case class FixtureParam(
-    config: SFAuthConfig,
-    auth: SalesforceAuth,
-    backendStub: SttpBackendStub[IO, Nothing]
-  )
+                           config: SFAuthConfig,
```
rupertbates, line 24: Was this Scalariform or Intellij? Doesn't look right either way!

kelvin-chappell:
@johnduffell and @rupertbates on the formatting I agree that it looks a bit strange.  Perhaps it's my Intellij settings if you say this isn't normal.  I'll check it again.

lib/salesforce/sttp-client/src/test/resources/subscription-query-response1.json
kelvin-chappell: Let me check this again.  I changed it when I was working on the tests because it relies on the parsing of Instants.  This is my `amm` session:
```
@ Instant.parse("2019-11-18T16:59:24Z")
res2: Instant = 2019-11-18T16:59:24Z

@ Instant.parse("2019-11-18T16:59:24.000+0000")
java.time.format.DateTimeParseException: Text '2019-11-18T16:59:24.000+0000' could not be parsed at index 23
  java.time.format.DateTimeFormatter.parseResolved0(DateTimeFormatter.java:2046)
  java.time.format.DateTimeFormatter.parse(DateTimeFormatter.java:1948)
  java.time.Instant.parse(Instant.java:395)
  ammonite.$sess.cmd3$.<clinit>(cmd3.sc:1)
```
But as you say, this doesn't match the Salesforce datetime format, which is still `2019-11-18T16:59:24.000+0000` so I'll have another look into this bit.

kelvin-chappell:
@rupertbates Thanks for the point about closing the clients.  That was an oversight.

lib/salesforce/sttp-client/src/test/scala/com/gu/salesforce/sttp/SalesforceClientTest.scala
kelvin-chappell: Rolled back in 23efc70a02425799479d5e2dc2dff59aca1852e8

lib/salesforce/sttp-client/src/test/scala/com/gu/salesforce/sttp/SalesforceClientTest.scala
kelvin-chappell: Rolled back in 23efc70a02425799479d5e2dc2dff59aca1852e8

johnduffell:
I think it might be intellij on the heavy indentation on case classes - generally I just defer to the standard rules, but I do remember having to find the setting for that when I got a new laptop.

kelvin-chappell:
> I think it might indeed be intellij on the heavy indentation on case classes - generally I just defer to the standard rules, but I do remember having to find the setting for that when I got a new laptop.

I had `Tools > Actions on Save > Reformat code` switched on, which is in an entirely different section to the Scala formatting page in Intellij.  It seems to be using some strange rules.  I've switched that off now anyway.  Usually I use scalafmt, which has sensible defaults.

lib/salesforce/sttp-client/src/test/resources/subscription-query-response1.json
kelvin-chappell: After some detective work, going backwards and forwards between main and this PR, I've discovered that there was a custom Instant decoder set up specifically to make this test pass and for no other reason.  It wasn't in scope anywhere else.  I've removed that in f125688 as it's misleading.  Testing in Code shows that Salesforce date-times are being decoded correctly by this branch.

PR #1193, opened 2021-09-10
Add cloudformation template to Riffraff artefacts

kelvin-chappell:
The cloudformation template for the Zuora callout APIs module was previously not being generated as an artefact that Riffraff could deploy.


jambo1 (approved):
Looks good!

PR #1195, opened 2021-09-13
fix snyk issue from CAS library

johnduffell:
Just while I was in bumping https://github.com/guardian/content-authorisation-common/pull/7
for https://github.com/guardian/subscriptions-frontend/pull/1347
I thought it would be sensible to bump it here too, as it's causing the same vuln here too https://app.snyk.io/vuln/SNYK-JAVA-COMMONSIO-1277109


rupertbates (approved):
👍

PR #1207, opened 2021-09-24
Support Saturday digital voucher rate plans

kelvin-chappell:
CSRs should be able to create Saturday and Saturday+ rate plans for digital vouchers.


graham228221 (approved):
Thanks Kelvin!

PR #1210, opened 2021-09-29
Don't schedule delivery credit processor in Dev

kelvin-chappell:
According to our cost management chart, the cost of the delivery credit processor has gone up since 9 Sep:

image

However, digging into it I found that all the extra cost was in the Dev stage:

image

Looking into the logs I found that there was a corresponding change in the behaviour of the Dev processor from that date.  It was timing out on every run and retrying twice on top of the scheduled run every 20 mins.  This was because of bad data in the Dev env, which we expect to crop up a lot.

So I've stopped the processor from running on a schedule in Dev but kept it running in Code to keep that stage as close to Prod as possible.  I've also stopped the processor from retrying on failure, as that seems unnecessary when it runs every 20 minutes anyway.

Deployed to Dev and Code.

### DIFF
--- a/handlers/delivery-problem-credit-processor/cfn.yaml
+++ b/handlers/delivery-problem-credit-processor/cfn.yaml
@@ -14,6 +14,7 @@ Parameters:
 
 Conditions:
   IsProd: !Equals [!Ref "Stage", "PROD"]
+  IsNotDev: !Not [!Equals [!Ref "Stage", "DEV"]]
 
 Mappings:
   StageMap:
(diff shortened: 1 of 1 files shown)


jambo1:
Does this change mean we won't have this lambda available in dev at all, or just that it won't run on the schedule in dev?

jambo1 (approved):

kelvin-chappell:
> Does this change mean we won't have this lambda available in dev at all, or just that it won't run on the schedule in dev?

It will be there but it won't run on a schedule so will need to be run manually if required.  But that seems quite sensible in a dev environment.

PR #1220, opened 2021-10-06
Update Github code-scanning result file

kelvin-chappell:
Unfortunately it looks as though I have to run the Snyk step twice as the action fails with error: 
`The following option combination is not currently supported: monitor + sarif-file-output`.


ripecosta (approved):
👍

PR #1258, opened 2021-10-29
Updating SNS topics & threshold for cloudwatch alarms

jambo1:
## What does this change?

Setting support-service lambdas cloudwatch alarms to the correct topics.

Updating threshold for holiday stops processor alarm to filter out more noise and only alert on systemic failures.


kelvin-chappell (approved):
Fast work! 🚀

PR #1264, opened 2021-11-02
make sure gw 6 for 6 don't get overlapping issues over christmas

johnduffell:
## What does this change?

This makes the main plan for 6 for 6 start and extra week after the first, to allow for the 7 weeks 6 issues problem at christmas
Also related
https://github.com/guardian/support-frontend/pull/3319

This change must be reverted along with the above PR, just before the catalog is changed back in zuora prod


rupertbates (approved):
👍

PR #1309, opened 2021-12-09
Raise threshold for highest expected discount

kelvin-chappell:
There are GW subs where each issue is worth 10.25 NZD.
This change ensures that they don't fail because they hit the threshold of expectation.


jambo1 (approved):

graham228221:
[CSR bug report (severity: 2) 502 is returned from holiday-stop-api for sub](https://trello.com/c/XPr8G45L/2744-csr-bug-report-severity-2-502-is-returned-from-holiday-stop-api-for-sub)

PR #1320, opened 2021-12-14
Sf emails exporter   group emails by case number and save as json in s3

david-pepper:
## What does this change?
Add logic to group the emails from Salesforce by Parent CaseNumber and save as json files to S3 with filename=casenumber

## How to test
run the lambda and verify that json files created in S3. 

Each Json file should:

- contain all emails relating to the Case

- be named by CaseNumber

## Coming Next
When writing to S3, rather than overwrite the file, append to it instead

### Previous PRs

[Create Basic Lambda - Sf emails to s3 exporter #1316](https://github.com/guardian/support-service-lambdas/pull/1316)

[Sf email exporter - authenticate with sf and get emails #1317](https://github.com/guardian/support-service-lambdas/pull/1317)

PR #1317, opened 2021-12-13
Sf email exporter - authenticate with sf and get emails

david-pepper:
## What does this change?
Adds the necessary authentication and retrieval mechanism for emails in Salesforce

## How to test
run the lambda and verify that the logger outputs valid email data

## Coming Next
Saving the emails as json to S3

PR #1316, opened 2021-12-13
Create Basic Lambda - Sf emails to s3 exporter

david-pepper:
**What does this change?**
Contains the Handler, cfn.yaml and riff-raff.yaml files for as basic lambda in AWS that will be iterated upon with business logic and tests in subsequent PRs.

**Lambda purpose**
The intention of the Lambda is to regularly query for emails from Salesforce which as marked as 'Ready for Export to S3'.

The records returned by the query will be converted to Json and saved to S3

The lambda should be able to GET and PUT files in new S3 bucket (emails-from-sf/{Stage})

Scheduling and Error logging will follow in subsequent dedicated PRs

PR #1323, opened 2021-12-15
Emails to s3 exporter - Append to files in S3 instead of simple overwrite

david-pepper:
## What does this change?
Add logic to append to S3 file if it already exists in S3

## Context
In Salesforce, the possibility exists that an email could be sent from a Case after the Case emails have been exported to S3.

We want to avoid an export on that email overwriting any file containing previously exported emails relating to that Case.

To ensure that previously exported emails do not get overwritten in S3, the changes in this PR will append the json representing the new email to any S3 file that matches the Parent CaseNumber

## How to test
**Verifying that a file is created in S3 when one does not already exist**
Create a case and a child email in Salesforce ensuring the email export_status__c field is ‘Ready for export to S3’

Run the lambda and verify that a new json file is created in S3 containing the email data as json, and the file is named by the casenumber

**Verifying that a file in S3 appended to when one already exists**

Complete the steps above for verifying that a file is created in S3 when one does not already exist

Add another email to the Case in Salesforce, ensuring the email export_status__c field is ‘Ready for export to S3’

Run the lambda and verify that the S3 file relating to the Case has been updated to contain the new email json from Salesforce and the original email json remains intact within the file

## Coming Next

Paginating through recordsets returned from Salesforce in the initial GET request

### Previous PRs

[Create Basic Lambda - Sf emails to s3 exporter #1316](https://github.com/guardian/support-service-lambdas/pull/1316)

[Sf email exporter - authenticate with sf and get emails #1317](https://github.com/guardian/support-service-lambdas/pull/1317)

[Sf emails exporter group emails by case number and save as json in s3 #1320](https://github.com/guardian/support-service-lambdas/pull/1320)

### DIFF
--- a/handlers/sf-emails-to-s3-exporter/src/main/scala/com/gu/sf_emails_to_s3_exporter/S3Connector.scala
+++ b/handlers/sf-emails-to-s3-exporter/src/main/scala/com/gu/sf_emails_to_s3_exporter/S3Connector.scala
@@ -2,10 +2,12 @@ package com.gu.sf_emails_to_s3_exporter
 
 import java.nio.charset.StandardCharsets
 
-import com.gu.effects.UploadToS3
+import com.gu.effects.{AwsS3, Key, UploadToS3}
 import com.typesafe.scalalogging.LazyLogging
 import software.amazon.awssdk.core.sync.RequestBody
-import software.amazon.awssdk.services.s3.model.PutObjectRequest
+import software.amazon.awssdk.services.s3.model.{ListObjectsRequest, PutObjectRequest}
+
+import scala.jdk.CollectionConverters.CollectionHasAsScala
 
 object S3Connector extends LazyLogging{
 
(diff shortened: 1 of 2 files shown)


handlers/sf-emails-to-s3-exporter/src/main/scala/com/gu/sf_emails_to_s3_exporter/S3Connector.scala
```diff
@@ -22,14 +28,69 @@ object S3Connector extends LazyLogging{
...
+    decodedCaseEmailsFromS3 match {
+      case Right(caseEmailsFromS3) => {
+        println("decoded from s3:" + caseEmailsFromS3)
+        val mergedEmails = mergeSfEmailsWithS3Emails(caseEmailsFromSf, caseEmailsFromS3)
+
+        writeEmailsJsonToS3(fileName, mergedEmails.asJson.toString())
+      }
+      case Left(error) => throw new RuntimeException(s"something went wrong $error")
+    }
+  }
```
kelvin-chappell, line 82: Is there any possibility of two processes writing to the same file concurrently?  Then you would lose information

david-pepper: No there is no possibility of that. The only writing to the s3 bucket will be via the lambda and it will process records sequentially within a single thread

handlers/sf-emails-to-s3-exporter/src/main/scala/com/gu/sf_emails_to_s3_exporter/S3Connector.scala
```diff
@@ -22,14 +28,69 @@ object S3Connector extends LazyLogging{
...
+    fileContentFromS3: Seq[EmailsFromSfResponse.Records]
+  ): Seq[EmailsFromSfResponse.Records] = {
+
+    val emailsThatExistInSfButNotS3 =
+      caseEmailsFromSf.filter(sfEmail =>
+        fileContentFromS3.count(S3Email =>
+          S3Email.Composite_Key__c == sfEmail.Composite_Key__c) == 0)
+
+    fileContentFromS3 ++ emailsThatExistInSfButNotS3
   }
```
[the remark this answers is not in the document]
david-pepper: Agreed. I'll add unit tests in a subsequent PR when the business logic has been completed

handlers/sf-emails-to-s3-exporter/src/main/scala/com/gu/sf_emails_to_s3_exporter/S3Connector.scala
kelvin-chappell: Then there's no problem as long as executions  of the lambda aren't run closely together and no-one tries to do a manual run of the lambda at the same time as a scheduled run

PR #1324, opened 2021-12-15
Emails to s3 exporter  - add pagination for larger recordsets

david-pepper:
## What does this change?
Enables processing of more than 200 emails per lambda execution

## Context
Queries via the REST API to Salesforce return a maximum of 200 records in one callout. If there are more than 200 records in the resultset returned by the query, the REST API provides the first 200, and a nextRecordsUrl to indicate the endpoint to hit that return the next set of 200 records, and so on.

The changes in this PR recursively retrieve the records and save to s3

## How to test

1. Ensure there are more than 200 emails to retrieve from Salesforce

2. Run the lambda.

3. Ensure that the emails are all saved to S3.

## Coming Next

Updating records in Salesforce that have been successfully saved to S3 to surface them for deletion

## Notes
We should still apply some limit based on anticipated upper boundary of the number of emails that may be processed in one day. To safely navigate the maximum 15 minute execution time, this limit will be applied to the query to ensure that the lambda runs for less than 10 minutes. The specific limitation will be determined in a subsequent piece of work to performance test the solution.

### Previous PRs

[Create Basic Lambda - Sf emails to s3 exporter #1316](https://github.com/guardian/support-service-lambdas/pull/1316)

[Sf email exporter - authenticate with sf and get emails #1317](https://github.com/guardian/support-service-lambdas/pull/1317)

[Sf emails exporter group emails by case number and save as json in s3 #1320](https://github.com/guardian/support-service-lambdas/pull/1320)

[Emails to s3 exporter - Append to files in S3 instead of simple overwrite #1323](https://github.com/guardian/support-service-lambdas/pull/1323)

### DIFF
--- a/handlers/sf-emails-to-s3-exporter/src/main/scala/com/gu/sf_emails_to_s3_exporter/Handler.scala
+++ b/handlers/sf-emails-to-s3-exporter/src/main/scala/com/gu/sf_emails_to_s3_exporter/Handler.scala
@@ -30,13 +30,13 @@ object Handler extends LazyLogging {
         for {
           emailsForExportFromSf <- getEmailsFromSfByQuery(successfulAuth)
         } yield {
-          processEmails(successfulAuth, emailsForExportFromSf)
+          saveEmailsToS3AndQueryForMoreIfTheyExist(successfulAuth, emailsForExportFromSf)
         }
       }
     }
   }
 
-  def processEmails(sfAuthDetails: SfAuthDetails, response: EmailsFromSfResponse.Response): Unit = {
+  def saveEmailsToS3AndQueryForMoreIfTheyExist(sfAuthDetails: SfAuthDetails, response: EmailsFromSfResponse.Response): Unit = {
 
     val sfEmailsGroupedByCaseNumber = response
       .records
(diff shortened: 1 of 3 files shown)


handlers/sf-emails-to-s3-exporter/src/main/scala/com/gu/sf_emails_to_s3_exporter/Handler.scala
```diff
@@ -14,25 +14,44 @@ object Handler extends LazyLogging {
...
+
+    response.done match {
+      case true => logger.info("Batch Complete")
+      case false => {
+        for {
+          nextPageEmails <- getEmailsFromSfByRecordsetReference(sfAuthDetails, response.nextRecordsUrl.get)
+        } yield {
+          saveEmailsToS3AndQueryForMoreIfTheyExist(sfAuthDetails, nextPageEmails)
+        }
       }
```
kelvin-chappell, line 55: Could be:
```suggestion
    if (response.done) logger.info("Batch Complete")
    else
      getEmailsFromSfByRecordsetReference(sfAuthDetails, response.nextRecordsUrl.get).map(nextPageEmails =>  saveEmailsToS3AndQueryForMoreIfTheyExist(sfAuthDetails, nextPageEmails))
```

kelvin-chappell (approved):
Looks good! 👍

PR #1353, opened 2022-01-14
Increase Salesforce read timeout threshold

kelvin-chappell:
We're seeing occasional timeouts when reading from Salesforce, which appears to be down to the time it takes to build up the query result set in Salesforce.  This is a short-term measure while we look at optimising the query.


graham228221 (approved):
Thanks Kelvin!

PR #1364, opened 2022-01-24
Replace email in s3 file if it already exists

david-pepper:
## What does this change?
When an email is exported from Salesforce to S3, a check is made as to whether or not the email already exists in the S3 file. The changes in this PR _replace_ the email in the S3 file if it already exists.

## Context
Previously if an email already existed in the S3 file, the logic would just do nothing and continue processing (i.e. it would not replace the email in the file because there was no need). 

This had the side effect of no writeback to Salesforce to indicate a successful export. We have decided to replace the email in the file rather than ignore it so that the datetime of the export can be written back to Salesforce to drive the state in the 'Export_Status__c' field and thus enable the deletion of the record in Salesforce once successfully exported

## How to test
- export an email from Salesforce to S3
- verify the email has been successfully saved in a file in S3
- Change a value on the email in Salesforce on one of the fields that are part of the export
- export the same email again
- verify that the email has been overwritten in the S3 file
- verify that the timestamp of the action has been written back to the email in Salesforce

### DIFF
--- a/handlers/sf-emails-to-s3-exporter/src/main/scala/com/gu/sf_emails_to_s3_exporter/Handler.scala
+++ b/handlers/sf-emails-to-s3-exporter/src/main/scala/com/gu/sf_emails_to_s3_exporter/Handler.scala
@@ -36,6 +36,7 @@ object Handler extends LazyLogging {
   }
 
   def processEmails(sfAuthDetails: SfAuthDetails, emailsDataFromSF: EmailsFromSfResponse.Response, bucketName: String): Any = {
+    logger.info(s"Start processing ${emailsDataFromSF.records.size} emails...")
 
     val emailIdsSuccessfullySavedToS3 = getEmailIdsSuccessfullySavedToS3(emailsDataFromSF, bucketName)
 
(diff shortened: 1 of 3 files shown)


kelvin-chappell (approved):
👍

handlers/sf-emails-to-s3-exporter/src/main/scala/com/gu/sf_emails_to_s3_exporter/Handler.scala
```diff
@@ -52,6 +53,7 @@ object Handler extends LazyLogging {
       )
     }
 
+    logger.info("More emails to retrieve from Salesforce:" + !emailsDataFromSF.done)
```
kelvin-chappell, line 56:
```suggestion
    logger.info("More emails to retrieve from Salesforce: " + !emailsDataFromSF.done)
```

handlers/sf-emails-to-s3-exporter/src/main/scala/com/gu/sf_emails_to_s3_exporter/S3Connector.scala
```diff
@@ -38,6 +41,8 @@ object S3Connector extends LazyLogging {
           .getOrElse(Seq[EmailsFromSfResponse.Records]())
           .exists(_.Composite_Key__c == caseEmail.Composite_Key__c)
 
+        logger.info(s" ${caseEmail.Composite_Key__c} already exists in file: ${emailAlreadyExistsInS3File} ")
```
kelvin-chappell, line 44:
```suggestion
        logger.info(s"${caseEmail.Composite_Key__c} already exists in file: ${emailAlreadyExistsInS3File}")
```

kelvin-chappell (approved):
👍

PR #1367, opened 2022-01-25
emails to s3 - fix env var names

david-pepper:
## What does this change?

Somehow the environment variable names fell out of sync with their references within the logic. This PR ensures they remain consistent between their definitions and their references

## How to test
- execute the lambda and verify that files are saved to s3 and successes are written back as a timestamps to EmailMessage.Most_Recent_Export__c


kelvin-chappell (approved):
👍

PR #1374, opened 2022-01-27
Increase timeout limit to 30 seconds for calls to sf

david-pepper:
## What does this change?

Intermittent timeouts are occurring in the 3 callouts to Salesforce. The changes in this PR increases the timeout to 30 seconds.

Calls to Salesforce:
- get emails by query [GET]
- get emails by recordset reference [GET]
- write back successes [PATCH]

## How to test
- Ensure there are at least 200 emails flagged as 'Ready for Export to S3' in Salesforce
- Run the lambda
- Verify that no timeouts occur


kelvin-chappell (approved):
You could avoid repeating the setting by building a skeletal custom Http function that includes the options you use with every call.
But this is fine. 👍

PR #1395, opened 2022-02-11
Remove gu.com reference

mchv:
## What does this change?

We should not be using that domain anymore.


kelvin-chappell (approved):
👍

johnduffell:
I did not know that, what changed - we use it in our test user email addresses to avoid emails being sent via braze https://github.com/guardian/membership-workflow/blob/51c1f21723649cb525bc1a343cab9ea2a901d84a/app/model/BrazeCampaignTriggerPayload.scala#L28

PR #1408, opened 2022-02-23
Order by createddate and set limit to 2000

david-pepper:
## What does this change?

Order by createddate instead of ParentId
- this increases the efficiency of the query because createddate is an indexed field, whereas ParentId is not

Increase limit of records to process to 2000 from 1000
- Under normal conditions there may be up to 1400 emails per day to export to S3 (usually this is significantly lower). 

From previous performance testing, processing of 2000 emails is estimated to take 6-7 mins, which is comfortably within the 15 minute lambda execution limit

## How to test
Execute  the lambda and verify that that no errors are encountered when the initial query is executed


graham228221 (approved):
Hopefully this helps a lot! If so this might be something to try with payment-failure-comms and soft-opt-in-consent-setter

david-pepper:
[Optimise Salesforce query](https://trello.com/c/vhAIUQpu/2943-optimise-salesforce-query)

PR #1418, opened 2022-03-01
Ensure that an account with an auto-cancelled subscription has no outstanding invoices

kelvin-chappell:
Previously, when a subscription had been auto-cancelled because it reached the end of the payment failure period without recovery it was left with an outstanding balance to pay on the final invoice.
In this PR, we transfer the balance of the negative invoice generated to balance the unpaid invoice to the credit account and then apply that amount to the unpaid invoice.  We call this a 'credit shuffle'.  The end result should be that all the invoices on the account have zero balances.

Tested on various accounts in Code and in the central sandbox.

### DIFF
--- a/handlers/zuora-callout-apis/src/main/scala/com/gu/autoCancel/AutoCancel.scala
+++ b/handlers/zuora-callout-apis/src/main/scala/com/gu/autoCancel/AutoCancel.scala
@@ -30,7 +30,8 @@ object AutoCancel extends Logging {
   /*
    * This process applies at the subscription level.  It will potentially run multiple times per invoice.
    * The cancellation call generates a balancing invoice that should be negative and the same amount
-   * as the amount outstanding for the invoice item corresponding to the subscription being processed.
+   * as the amount outstanding for the invoice items corresponding to the subscription being processed
+   * (there could be multiple invoice items per sub - it could include discounts and multi-day paper subs)
    * This means that after all subscriptions on an invoice have been cancelled, the balance of all
    * invoices should be 0.
    */
--- a/handlers/zuora-callout-apis/src/main/scala/com/gu/autoCancel/AutoCancel.scala
+++ b/handlers/zuora-callout-apis/src/main/scala/com/gu/autoCancel/AutoCancel.scala
@@ -1,17 +1,25 @@
 package com.gu.autoCancel
 
-import java.time.LocalDate
-
 import com.gu.autoCancel.AutoCancelSteps.AutoCancelUrlParams
 import com.gu.stripeCustomerSourceUpdated.TypeConvert._
 import com.gu.util.Logging
 import com.gu.util.reader.Types._
 import com.gu.util.resthttp.RestRequestMaker.Requests
-import com.gu.util.zuora.{SubscriptionNumber, ZuoraCancelSubscription, ZuoraUpdateCancellationReason}
+import com.gu.util.resthttp.Types.{ClientFailableOp, ClientSuccess, NotFound}
+import com.gu.util.zuora._
+import play.api.libs.json.{JsDefined, JsString, JsValue}
+
+import java.time.LocalDate
 
 object AutoCancel extends Logging {
 
-  case class AutoCancelRequest(accountId: String, subToCancel: SubscriptionNumber, cancellationDate: LocalDate)
+  case class AutoCancelRequest(
+    accountId: String,
+    subToCancel: SubscriptionNumber,
+    cancellationDate: LocalDate,
+    invoiceId: String,
+    invoiceAmount: BigDecimal
+  )
 
   def apply(requests: Requests)(acRequests: List[AutoCancelRequest], urlParams: AutoCancelUrlParams): ApiGatewayOp[Unit] = {
     logger.info(s"dryRun: ${urlParams.dryRun}")
@@ -22,14 +30,25 @@ object AutoCancel extends Logging {
     responses.head
   }
 
+  private def findBalancingInvoiceId(cancelSubscriptionResponse: JsValue): ClientFailableOp[String] =
+    cancelSubscriptionResponse \ "invoiceId" match {
+      case JsDefined(JsString(value)) => ClientSuccess(value)
+      case _ => NotFound(s"Subscription cancellation response '$cancelSubscriptionResponse' doesn't contain an invoice ID")
+    }
+
   private def executeCancel(requests: Requests, dryRun: Boolean)(acRequest: AutoCancelRequest): ApiGatewayOp[Unit] = {
-    val AutoCancelRequest(accountId, subToCancel, cancellationDate) = acRequest
+    val AutoCancelRequest(accountId, subToCancel, cancellationDate, invoiceId, invoiceAmount) = acRequest
     logger.info(s"Attempting to perform auto-cancellation on account: $accountId for subscription: ${subToCancel.value}")
     val zuoraUpdateCancellationReasonF = if (dryRun) ZuoraUpdateCancellationReason.dryRun(requests) _ else ZuoraUpdateCancellationReason(requests) _
     val zuoraCancelSubscriptionF = if (dryRun) ZuoraCancelSubscription.dryRun(requests) _ else ZuoraCancelSubscription(requests) _
+    val zuoraTransferToCreditBalanceF = if (dryRun) TransferToCreditBalance.dryRun(requests) _ else TransferToCreditBalance(requests) _
+    val zuoraApplyCreditBalanceF = if (dryRun) ApplyCreditBalance.dryRun(requests) _ else ApplyCreditBalance(requests) _
     val zuoraOp = for {
       _ <- zuoraUpdateCancellationReasonF(subToCancel).withLogging("updateCancellationReason")
-      _ <- zuoraCancelSubscriptionF(subToCancel, cancellationDate).withLogging("cancelSubscription")
+      cancellationResponse <- zuoraCancelSubscriptionF(subToCancel, cancellationDate).withLogging("cancelSubscription")
+      balancingInvoiceId <- findBalancingInvoiceId(cancellationResponse)
+      _ <- zuoraTransferToCreditBalanceF(balancingInvoiceId, invoiceAmount, "Auto-cancellation").withLogging("transferToCreditBalance")
+      _ <- zuoraApplyCreditBalanceF(invoiceId, invoiceAmount, "Auto-cancellation").withLogging("applyCreditBalance")
     } yield ()
     zuoraOp.toApiGatewayOp("AutoCancel failed")
   }
--- a/handlers/zuora-callout-apis/src/main/scala/com/gu/autoCancel/AutoCancel.scala
+++ b/handlers/zuora-callout-apis/src/main/scala/com/gu/autoCancel/AutoCancel.scala
@@ -45,8 +45,8 @@ object AutoCancel extends Logging {
     val zuoraOp = for {
       _ <- zuoraUpdateCancellationReasonF(subToCancel).withLogging("updateCancellationReason")
       cancellationResponse <- zuoraCancelSubscriptionF(subToCancel, cancellationDate).withLogging("cancelSubscription")
-      creditTransferInvoice <- zuoraGetInvoiceF(cancellationResponse.invoiceId)
-      creditTransferAmount = -creditTransferInvoice.Balance
+      cancellationNegativeInvoice <- zuoraGetInvoiceF(cancellationResponse.invoiceId)
+      creditTransferAmount = -cancellationNegativeInvoice.Balance
       _ <- zuoraTransferToCreditBalanceF(cancellationResponse.invoiceId, creditTransferAmount, "Auto-cancellation").withLogging("transferToCreditBalance")
       _ <- zuoraApplyCreditBalanceF(invoiceId, creditTransferAmount, "Auto-cancellation").withLogging("applyCreditBalance")
     } yield ()
--- a/handlers/zuora-callout-apis/src/main/scala/com/gu/autoCancel/AutoCancel.scala
+++ b/handlers/zuora-callout-apis/src/main/scala/com/gu/autoCancel/AutoCancel.scala
@@ -5,9 +5,7 @@ import com.gu.stripeCustomerSourceUpdated.TypeConvert._
 import com.gu.util.Logging
 import com.gu.util.reader.Types._
 import com.gu.util.resthttp.RestRequestMaker.Requests
-import com.gu.util.resthttp.Types.{ClientFailableOp, ClientSuccess, NotFound}
 import com.gu.util.zuora._
-import play.api.libs.json.{JsDefined, JsString, JsValue}
 
 import java.time.LocalDate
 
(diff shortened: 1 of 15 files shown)


handlers/zuora-callout-apis/src/main/scala/com/gu/autoCancel/AutoCancel.scala
```diff
@@ -1,17 +1,25 @@
...
 
 object AutoCancel extends Logging {
 
-  case class AutoCancelRequest(accountId: String, subToCancel: SubscriptionNumber, cancellationDate: LocalDate)
+  case class AutoCancelRequest(
+    accountId: String,
+    subToCancel: SubscriptionNumber,
+    cancellationDate: LocalDate,
+    invoiceId: String,
+    invoiceAmount: BigDecimal
```
johnduffell, line 21: we must remember to go in to zuora and add those two fields! (this can be done in advance of shipping the PR as they will be ignored)

kelvin-chappell: Actually I don't think we need to make any changes to the lambda input.

It comes in this format: 
https://github.com/guardian/support-service-lambdas/blob/main/handlers/zuora-callout-apis/src/main/scala/com/gu/autoCancel/AutoCancelCallout.scala#L5-L18

And then we transform it here:
https://github.com/guardian/support-service-lambdas/blob/main/handlers/zuora-callout-apis/src/main/scala/com/gu/autoCancel/AutoCancelDataCollectionFilter.scala#L16-L37
which is modified slightly in this PR to get the extra fields in

handlers/zuora-callout-apis/src/main/scala/com/gu/util/zuora/ZuoraCancelSubscription.scala
```diff
@@ -20,21 +20,18 @@ object ZuoraCancelSubscription extends LazyLogging {
...
-    Reads(_ => JsSuccess(()))
-
   private def toBodyAndPath(subscription: SubscriptionNumber, cancellationDate: LocalDate) =
     (SubscriptionCancellation(cancellationDate), s"subscriptions/${subscription.value}/cancel")
 
-  def apply(requests: Requests)(subscription: SubscriptionNumber, cancellationDate: LocalDate): ClientFailableOp[Unit] = {
+  def apply(requests: Requests)(subscription: SubscriptionNumber, cancellationDate: LocalDate): ClientFailableOp[JsValue] = {
     val (body, path) = toBodyAndPath(subscription, cancellationDate)
-    requests.put(body, path)
+    requests.put[SubscriptionCancellation, JsValue](body, path)
```
johnduffell, line 28: interesting to make this a JsValue rather than a modelled case class.  I'm not objecting on principle, but it does mean that we're relying on elsewhere in the code to understand the format of what was returned, which might muddy the waters?  Or would you consider the lambda so interdependent anyway that it doesn't need that separation?
In fact, it looks like we only need the `invoiceId` anyway, so this call could return a String, or a case class with a single field.

kelvin-chappell: You're right.  That would be much better.

handlers/zuora-callout-apis/src/main/scala/com/gu/autoCancel/AutoCancelDataCollectionFilter.scala
```diff
@@ -30,9 +29,10 @@ object AutoCancelDataCollectionFilter extends Logging {
         .withLogging("getAccountSubscriptions")
       subsToCancel <- filterNotActiveSubscriptions(subsOnAccount, subsOnInvoice).withLogging("filterNotActiveSubscriptions")
```
johnduffell, line 30: I'm thinking about failure modes, and if the balance shuffle fails, the subscription would have already been cancelled but zuora would receive a 500 error from us.  This means when zuora retries the callout, the not-active check might skip over the sub completely.
This is better than not autocancelling at all, and the risk is probably not high for this, but I thought I'd flag it anyway.

I think in reality it's all the more argument for a regular cleaner job that goes through and balances invoices that are missed.

handlers/zuora-callout-apis/src/main/scala/com/gu/util/zuora/ZuoraCreditTransfer.scala
```diff
@@ -0,0 +1,64 @@
...
+    val (body, path) = ZuoraCreditTransfer.toBodyAndPath(invoiceId, amount, increase = false, comment)
+    requests.put(body, path)
+  }
+
+  def dryRun(requests: Requests)(invoiceId: String, amount: BigDecimal, comment: String): ClientFailableOp[Unit] = {
+    val (body, path) = ZuoraCreditTransfer.toBodyAndPath(invoiceId, amount, increase = false, comment)
+    val msg = s"DryRun for ZuoraCreditTransfer: body=$body, path=$path"
+    logger.info(msg)
+    ClientSuccess(JsNull)
+  }
```
johnduffell, line 63: I realise this is following the existing pattern (that I started!) but it does seem to slightly defeat the purpose when there's a lot of (hopefully accurately) duplicated code between the dry run and the normal run.

kelvin-chappell: I think we could get rid of the dry-run completely if we use the central sandbox but, as you say, I'm just copying the existing pattern.  It does at least give you some confidence that you're making the right calls in the right order but arguably that should be being done in tests.

handlers/zuora-callout-apis/src/main/scala/com/gu/autoCancel/AutoCancelDataCollectionFilter.scala
kelvin-chappell: I'm a bit concerned about the brittleness of it too.  I prefer lambdas to do as little as possible.  Also the email step won't be reached if there's a failure.  Let's take this as the next step in an incremental improvement.

handlers/zuora-callout-apis/src/main/scala/com/gu/autoCancel/AutoCancel.scala
johnduffell: thanks Kelvin, I agree, I think I was confusing things hence my edit. 👍

handlers/zuora-callout-apis/src/main/scala/com/gu/autoCancel/AutoCancelDataCollectionFilter.scala
johnduffell: good point about the email step.  I agree about as little as possible (or idempodent).  Yes agree to look into it more in future - the EPIC that came out of the incident has some architectural improvements in terms of making some/all of these things repeatable cleaner-jobs rather than one-off triggered.

handlers/zuora-callout-apis/src/main/scala/com/gu/autoCancel/AutoCancelDataCollectionFilter.scala
```diff
@@ -30,9 +29,10 @@ object AutoCancelDataCollectionFilter extends Logging {
...
       cancellationDate <- getCancellationDateFromInvoice(invoiceId, accountSummary, now).withLogging("getCancellationDateFromInvoice")
+      invoiceAmount <- getInvoiceBalance(invoiceId, accountSummary).withLogging("getInvoiceBalance")
     } yield subsToCancel
       .map { subToCancel =>
-        AutoCancelRequest(accountId, subToCancel, cancellationDate)
+        AutoCancelRequest(accountId, subToCancel, cancellationDate, autoCancelCallout.invoiceId, invoiceAmount)
```
johnduffell, line 35:
```suggestion
        AutoCancelRequest(accountId, subToCancel, cancellationDate, invoiceId, invoiceAmount)
```

handlers/zuora-callout-apis/src/main/scala/com/gu/paymentFailure/ToMessage.scala
```diff
@@ -29,7 +29,6 @@ object ToMessage {
               first_name = paymentFailureCallout.firstName,
               last_name = paymentFailureCallout.lastName,
               primaryKey = PaymentId(paymentFailureCallout.paymentId),
-              price = price(paymentFailureInformation.amount, paymentFailureCallout.currency),
```
johnduffell, line 32: are we even triggered for payment failure any more?  If not all this payment failure related code can be deleted in favour of the Auto cancel code.

handlers/zuora-callout-apis/src/main/scala/com/gu/util/email/EmailSendSteps.scala
```diff
@@ -18,7 +18,6 @@ case class SubscriberAttributesDef(
   first_name: String,
   last_name: String,
   primaryKey: PrimaryKey,
-  price: String,
```
johnduffell, line 21: just for future reference, this price is not used in braze in the email template, so can be removed.

handlers/zuora-callout-apis/src/main/scala/com/gu/util/zuora/ZuoraCreditTransfer.scala
```diff
@@ -0,0 +1,71 @@
...
+import com.typesafe.scalalogging.LazyLogging
+import play.api.libs.json._
+
+object ZuoraCreditTransfer extends LazyLogging {
+
+  sealed trait AdjustmentType { def value: String }
+  object AdjustmentType {
+    case object Increase extends AdjustmentType { val value = "Increase" }
+    case object Decrease extends AdjustmentType { val value = "Decrease" }
+  }
```
johnduffell, line 15: simpler to write this way
```suggestion
  sealed abstract class AdjustmentType(val value: String)
  object AdjustmentType {
    case object Increase extends AdjustmentType("Increase")
    case object Decrease extends AdjustmentType("Decrease")
  }
```

handlers/zuora-callout-apis/src/main/scala/com/gu/autoCancel/AutoCancelDataCollectionFilter.scala
```diff
@@ -30,9 +29,10 @@ object AutoCancelDataCollectionFilter extends Logging {
         .withLogging("getAccountSubscriptions")
       subsToCancel <- filterNotActiveSubscriptions(subsOnAccount, subsOnInvoice).withLogging("filterNotActiveSubscriptions")
       cancellationDate <- getCancellationDateFromInvoice(invoiceId, accountSummary, now).withLogging("getCancellationDateFromInvoice")
+      invoiceAmount <- getInvoiceBalance(invoiceId, accountSummary).withLogging("getInvoiceBalance")
```
johnduffell, line 32: we don't want the total invoice amount, we only want the contribution from `subToCancel`
This might be tricky to work out for subs with multiple plans and charges (e.g. discounts, and newspaper subs)

johnduffell: might be better to get `invoiceAmount` from the invoice returned in `cancellationResponse` instead?

kelvin-chappell: That's a good idea.  I'll give it a go

handlers/zuora-callout-apis/src/main/scala/com/gu/paymentFailure/ToMessage.scala
kelvin-chappell: I think I'd rather leave that to another PR.  That's going to be removing a big chunk of code!

handlers/zuora-callout-apis/src/main/scala/com/gu/autoCancel/AutoCancel.scala
```diff
@@ -36,17 +35,20 @@ object AutoCancel extends Logging {
    * invoices should be 0.
    */
   private def executeCancel(requests: Requests, dryRun: Boolean)(acRequest: AutoCancelRequest): ApiGatewayOp[Unit] = {
-    val AutoCancelRequest(accountId, subToCancel, cancellationDate, invoiceId, invoiceAmount) = acRequest
+    val AutoCancelRequest(accountId, subToCancel, cancellationDate, invoiceId) = acRequest
```
johnduffell, line 38: I'm getting confused with all the invoices
```suggestion
    val AutoCancelRequest(accountId, subToCancel, cancellationDate, overdueInvoiceId) = acRequest
```

handlers/zuora-callout-apis/src/main/scala/com/gu/autoCancel/AutoCancel.scala
```diff
@@ -36,17 +35,20 @@ object AutoCancel extends Logging {
...
+    val zuoraGetInvoiceF = if (dryRun) ZuoraGetInvoice.dryRun(requests) _ else ZuoraGetInvoice(requests) _
     val zuoraTransferToCreditBalanceF = if (dryRun) TransferToCreditBalance.dryRun(requests) _ else TransferToCreditBalance(requests) _
     val zuoraApplyCreditBalanceF = if (dryRun) ApplyCreditBalance.dryRun(requests) _ else ApplyCreditBalance(requests) _
     val zuoraOp = for {
       _ <- zuoraUpdateCancellationReasonF(subToCancel).withLogging("updateCancellationReason")
       cancellationResponse <- zuoraCancelSubscriptionF(subToCancel, cancellationDate).withLogging("cancelSubscription")
-      _ <- zuoraTransferToCreditBalanceF(cancellationResponse.invoiceId, invoiceAmount, "Auto-cancellation").withLogging("transferToCreditBalance")
-      _ <- zuoraApplyCreditBalanceF(invoiceId, invoiceAmount, "Auto-cancellation").withLogging("applyCreditBalance")
+      creditTransferInvoice <- zuoraGetInvoiceF(cancellationResponse.invoiceId)
+      creditTransferAmount = -creditTransferInvoice.Balance
```
johnduffell, line 49:
```suggestion
      cancellationNegativeInvoice <- zuoraGetInvoiceF(cancellationResponse.invoiceId)
      creditTransferAmount = -cancellationNegativeInvoice.Balance
```

handlers/zuora-callout-apis/src/main/scala/com/gu/autoCancel/AutoCancel.scala
```diff
@@ -22,14 +27,28 @@ object AutoCancel extends Logging {
     responses.head
   }
 
+  /*
+   * This process applies at the subscription level.  It will potentially run multiple times per invoice.
+   * The cancellation call generates a balancing invoice that should be negative and the same amount
+   * as the amount outstanding for the invoice item corresponding to the subscription being processed.
```
johnduffell, line 33:
```suggestion
   * as the amount outstanding for the invoice items corresponding to the subscription being processed
   * (there could be multiple invoice items per sub - it could include discounts and multi-day paper subs)
```

handlers/zuora-callout-apis/src/main/scala/com/gu/autoCancel/AutoCancelSteps.scala
```diff
@@ -28,6 +28,10 @@ object AutoCancelSteps extends Logging {
     implicit val autoCancelUrlParamsReads: Reads[AutoCancelUrlParams] = json => wireReads.reads(json).map(_.toAutoCancelUrlParams)
   }
 
+  /*
+   * This process applies at the invoice level.
+   * Each invoice can have multiple invoice items applying to a different subscription.
```
johnduffell, line 33: just wondering if we should have a final check that the positive invoice is zero balance, once all the subs are cancelled?  seems overkill to make another API call for that, and we will (hopefully) add a general cleaner to deal with things left over

johnduffell (approved):
looks good, just a few final comments but I think this will work correctly, well done for getting all that ironed out 👏

handlers/zuora-callout-apis/src/main/scala/com/gu/autoCancel/AutoCancelSteps.scala
kelvin-chappell: I'm not sure how we could easily do that without making another call to fetch the final state of the account

PR #1424, opened 2022-03-03
Generate a balancing invoice when auto-cancelling a subscription: Attempt 2

kelvin-chappell:
Previously, the auto-cancel process wouldn't generate a balancing invoice. This made it impossible to do credit transfers to zero the last invoice for the cancelled subscription until a balancing invoice was generated on the next bill run.

Now, the balancing invoice will be generated immediately following the cancellation of the subscription. So this will enable us to zero the last invoice on the account for the cancelled subscription in another PR.

See https://www.zuora.com/developer/api-reference/#operation/PUT_CancelSubscription

The call to Zuora depends on a minimum version of the Zuora API, which has now been localised to the auto-cancelling lambda.

See also:
#1406
#1416 
#1421

### DIFF
--- a/handlers/zuora-callout-apis/src/main/scala/com/gu/autoCancel/AutoCancelHandler.scala
+++ b/handlers/zuora-callout-apis/src/main/scala/com/gu/autoCancel/AutoCancelHandler.scala
@@ -23,8 +23,6 @@ import scala.util.Try
 
 object AutoCancelHandler extends App with Logging {
 
-  private val zuoraApiMinorVersion = "315.0"
-
   def operationForEffects(
     stage: Stage,
     fetchString: StringFromS3,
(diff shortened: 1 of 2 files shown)


johnduffell (approved):
didn't know the api version was stored in the config!  seems a bit random for that to be the case given it's a code compatibility thing
I'd be tempted to put that in the code instead
but the PR is good and thanks for doing it 👍

handlers/zuora-callout-apis/src/main/scala/com/gu/autoCancel/AutoCancelHandler.scala
```diff
@@ -32,7 +32,9 @@ object AutoCancelHandler extends App with Logging {
   ): ApiGatewayOp[ApiGatewayHandler.Operation] = {
     val loadConfigModule = LoadConfigModule(stage, fetchString)
     for {
-      zuoraRestConfig <- loadConfigModule[ZuoraRestConfig].toApiGatewayOp("load zuora config")
+      zuoraRestConfig <- loadConfigModule[ZuoraRestConfig]
+        .toApiGatewayOp("load zuora config")
+        .map(_.copy(apiMinorVersion = Some("315.0")))
```
johnduffell, line 37: I think the "315.0" should be passed in on this line as a new parameter: 
https://github.com/guardian/support-service-lambdas/blob/1db3f62c255c0c710dba8f8d9f448b97b5af735b/handlers/zuora-callout-apis/src/main/scala/com/gu/util/zuora/ZuoraCancelSubscription.scala#L31

PR #1425, opened 2022-03-03
Support API version in call to Zuora

kelvin-chappell:
This is a change to the way we use the Zuora API minor version so that it's call-specific rather than context-specific.
This supersedes #1421.

Next steps:

- [ ] Raise a modified version of #1424 to use this new technique.
- [ ] Remove the functionality added in #1421

### DIFF
--- a/lib/restHttp/src/main/scala/com/gu/util/resthttp/RestRequestMaker.scala
+++ b/lib/restHttp/src/main/scala/com/gu/util/resthttp/RestRequestMaker.scala
@@ -44,6 +44,7 @@ object RestRequestMaker extends LazyLogging {
 
   trait RequestsPUT {
     def put[REQ: Writes, RESP: Reads](req: REQ, path: String): ClientFailableOp[RESP]
+    def put[REQ: Writes, RESP: Reads](req: REQ, path: String, apiVersion: String): ClientFailableOp[RESP]
   }
 
   sealed trait IsCheckNeeded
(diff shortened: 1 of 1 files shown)


lib/restHttp/src/main/scala/com/gu/util/resthttp/RestRequestMaker.scala
```diff
@@ -117,15 +118,21 @@ object RestRequestMaker extends LazyLogging {
...
         _ <- jsonIsSuccessful(bodyAsJson)
         respModel <- toResult[RESP](bodyAsJson)
       } yield respModel
     }
 
+    def put[REQ: Writes, RESP: Reads](req: REQ, path: String): ClientFailableOp[RESP] =
+      put(req, path, headers)
+
+    def put[REQ: Writes, RESP: Reads](req: REQ, path: String, apiVersion: String): ClientFailableOp[RESP] =
+      put(req, path, headers + ("zuora-version" -> apiVersion))
```
johnduffell, line 134: I suddenly thought is this Requests class zuora specific or covers other systems?  I think it's not zuora specific, why not pass the whole `("zuora-version" -> apiVersion)` as a parameter instead (we can abstract it out of we use it too many places)

johnduffell (approved):
I prefer this way although see the comment below about passing in the key as well as the value to avoid making this class zuora specific (unless it already is of course)

PR #1426, opened 2022-03-03
Apply Zuora API version at call level

kelvin-chappell:
This implements #1424 taking advantage of the method introduced in #1425.


johnduffell (approved):
yep that feels like it has the right things in the right place, thanks for updating that 👍 
Also thanks for making all these such small PRs, it makes it super simple to review and understand.

PR #1435, opened 2022-03-08
Backfill - run lambda every 15 mins

david-pepper:
## What does this change?
Schedule the lambda to run every 15 minutes.

From previous performance testing, we expect the processing time to be 6-7 minutes for each set of 2000 records returned by the start query

### DIFF
--- a/handlers/sf-emails-to-s3-exporter/cfn.yaml
+++ b/handlers/sf-emails-to-s3-exporter/cfn.yaml
@@ -323,6 +323,13 @@ Resources:
               Action: s3:GetObject
               Resource:
                 - arn:aws:s3::*:membership-dist/*
+      Events:
+        ScheduledRun:
+          Type: Schedule
+          Properties:
+            Schedule: !FindInMap [ StageMap, !Ref Stage, Schedule ]
+            Description: Runs SF Emails to S3 Exporter
+            Enabled: True
 
   failedS3WriteFileAlarm:
     Type: AWS::CloudWatch::Alarm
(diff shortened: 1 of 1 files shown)


kelvin-chappell:
You also need to add an `events` object to the lambda, like this: https://github.com/guardian/support-service-lambdas/blob/main/handlers%2Fsoft-opt-in-consent-setter%2Fcfn.yaml#L109-L115

handlers/sf-emails-to-s3-exporter/cfn.yaml
```diff
@@ -322,6 +323,13 @@ Resources:
               Action: s3:GetObject
               Resource:
                 - arn:aws:s3::*:membership-dist/*
+      Events:
+        ScheduledRun:
+          Type: Schedule
+          Properties:
+            Schedule: !FindInMap [ StageMap, !Ref Stage, Schedule ]
```
david-pepper, line 330: Could this line cause an issue if 'Schedule' is not present in the DEV Stage?

kelvin-chappell: Yes.  Unfortunately the events block isn't allowed to be conditional.  That's why [here](https://github.com/guardian/support-service-lambdas/blob/main/handlers/soft-opt-in-consent-setter/cfn.yaml#L25-L34) it's set to run once a year in non-Prod environments.

kelvin-chappell (approved):
That looks complete now 👍

david-pepper:
[Backfill of emails to s3](https://trello.com/c/MjZWS6af/2940-backfill-of-emails-to-s3)

PR #1443, opened 2022-03-11
Apply credit to all unpaid invoices during auto-cancellation

kelvin-chappell:
Previously, the auto-cancel would fail when an account has multiple unpaid invoices because we tried to apply the entire credit balance to the invoice that caused the cancellation.
In this PR, the credit balance is distributed among all the unpaid invoices on the account.
This works because the negative invoice generated on cancellation is always equal to the sum of balances of unpaid invoices.

This has been tested on accounts with single and multiple unpaid invoices.


handlers/zuora-callout-apis/src/main/scala/com/gu/util/zuora/ZuoraCreditTransfer.scala
```diff
@@ -57,15 +58,26 @@ object ApplyCreditBalance extends LazyLogging {
   implicit val unitReads: Reads[Unit] =
     Reads(_ => JsSuccess(()))
 
-  def apply(requests: Requests)(invoiceId: String, amount: BigDecimal, comment: String): ClientFailableOp[Unit] = {
+  def apply(requests: Requests)(invoices: Seq[Invoice], comment: String): ClientFailableOp[Unit] =
+    invoices.map(invoice =>
+      applyToInvoice(requests)(invoice.id, invoice.balance, comment)).collectFirst { case failure: ClientFailure => failure }
```
johnduffell, line 63: it will always apply the full balance to all negative invoices even if the `creditTransferAmount` is less than that

handlers/zuora-callout-apis/src/main/scala/com/gu/autoCancel/AutoCancel.scala
```diff
@@ -40,17 +42,34 @@ object AutoCancel extends Logging {
...
       _ <- zuoraTransferToCreditBalanceF(cancellationResponse.invoiceId, creditTransferAmount, "Auto-cancellation").withLogging("transferToCreditBalance")
-      _ <- zuoraApplyCreditBalanceF(overdueInvoiceId, creditTransferAmount, "Auto-cancellation").withLogging("applyCreditBalance")
+      _ <- zuoraApplyCreditBalanceF(invoicesToBalance.unpaidInvoices, "Auto-cancellation").withLogging("applyCreditBalance")
     } yield ()
     zuoraOp.toApiGatewayOp("AutoCancel failed")
   }
+
+  case class InvoicesToBalance(negativeInvoice: Invoice, unpaidInvoices: Seq[Invoice])
+
+  object InvoicesToBalance {
```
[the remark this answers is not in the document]
kelvin-chappell: That's a better name.  Didn't really think of that

handlers/zuora-callout-apis/src/main/scala/com/gu/autoCancel/AutoCancel.scala
```diff
@@ -40,17 +42,34 @@ object AutoCancel extends Logging {
...
+    def fromAccountSummary(accountSummary: AccountSummary):ClientFailableOp[InvoicesToBalance] =
+      for {
+        negativeInvoice <- accountSummary.invoices.find(_.balance < 0) match {
```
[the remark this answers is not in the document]
kelvin-chappell: I'll change this to look up by the known invoice ID rather than the balance.

handlers/zuora-callout-apis/src/main/scala/com/gu/util/zuora/ZuoraCreditTransfer.scala
kelvin-chappell: Expected behaviour here is going to need a bit of thinking about.  It will probably be easier if we can come up with some realistic test cases.

kelvin-chappell: In 4213795a8869b6f65b79944ddffde8d4a5774318, I'm returning a failure if the credit available is not enough to balance all the unpaid invoices.  As whatever we do there'll be at least one invoice with a non-zero balance in this scenario.

handlers/zuora-callout-apis/src/main/scala/com/gu/autoCancel/AutoCancel.scala
```diff
@@ -49,27 +48,27 @@ object AutoCancel extends Logging {
...
-  case class InvoicesToBalance(negativeInvoice: Invoice, unpaidInvoices: Seq[Invoice])
+  case class UnbalancedInvoices(negativeInvoice: Invoice, unpaidInvoices: Seq[Invoice])
 
-  object InvoicesToBalance {
-    def fromAccountSummary(accountSummary: AccountSummary):ClientFailableOp[InvoicesToBalance] =
+  object UnbalancedInvoices {
+    def fromAccountSummary(accountSummary: AccountSummary, idOfNegativeInvoice: String):ClientFailableOp[UnbalancedInvoices] =
       for {
-        negativeInvoice <- accountSummary.invoices.find(_.balance < 0) match {
+        negativeInvoice <- accountSummary.invoices.find(_.id == idOfNegativeInvoice) match {
```
johnduffell, line 64: something I have in my mind all the time is the account summary only shows the latest invoices.  I think this will be fine for us, but I am just keeping it in mind.

handlers/zuora-callout-apis/src/main/scala/com/gu/util/zuora/ZuoraCreditTransfer.scala
```diff
@@ -58,17 +58,22 @@ object ApplyCreditBalance extends LazyLogging {
   implicit val unitReads: Reads[Unit] =
     Reads(_ => JsSuccess(()))
 
-  def apply(requests: Requests)(invoices: Seq[Invoice], comment: String): ClientFailableOp[Unit] =
-    invoices.map(invoice =>
-      applyToInvoice(requests)(invoice.id, invoice.balance, comment)).collectFirst { case failure: ClientFailure => failure }
-      .getOrElse(ClientSuccess(()))
+  def apply(requests: Requests)(invoices: Seq[Invoice], creditAvailable: Double, comment: String): ClientFailableOp[Unit] = {
+    val totalUnpaidInvoiceBalance = invoices.map(_.balance).sum
+    if (totalUnpaidInvoiceBalance > creditAvailable)
```
johnduffell, line 63: Assuming the balance on an unpaid invoice is positive:
if there are two subscriptions, when we cancel the first one, only half(ish) the credit will be available.  So this would fail due to 2 subs unpaid and only one has been cancelled.

handlers/zuora-callout-apis/src/main/scala/com/gu/autoCancel/AutoCancel.scala
```diff
@@ -41,34 +41,49 @@ object AutoCancel extends Logging {
...
+    subToCancel: SubscriptionNumber, invoices: Seq[ItemisedInvoice], comment: String
+  ): ClientFailableOp[Unit] = {
+    invoices.map(invoice =>
+      invoice.invoiceItems.length match {
+        case 0 => GenericError(s"Invoice ${invoice.id} has no items")
+        case 1 => applyCreditBalance(invoice.id, invoice.balance, comment)
+        case _ =>
+          invoice.invoiceItems.find(_.subscriptionName == subToCancel.value) match {
+            case None => GenericError(s"Invoice ${invoice.id} isn't for subscription $subToCancel")
+            case Some(item) => applyCreditBalance(invoice.id, item.chargeAmount, comment)
```
johnduffell, line 69: so if there are multiple items we only apply the credit balance to the first item that matches the subscription name, how does that work if there are multiple items relating to one subscription, (e,g, discount, holiday credit, multi day paper sub)

johnduffell: if we use filter instead of find then total up the item amounts before applying the balance that might work better.  (As long as there are no rounding errors)  If we try to do each one separately it might struggle with the negative items.

kelvin-chappell: Latest commit sums up charges on relevant items and applies that to the credit balance.

handlers/zuora-callout-apis/src/main/scala/com/gu/util/zuora/ZuoraCreditTransfer.scala
kelvin-chappell: I've removed this check now as it wasn't adding any value to the message that Zuora gives when it tries to apply an amount to the credit balance that's greater than the credit available.

PR #1446, opened 2022-03-14
Avoid dependency on pre-synced subscription for Salesforce to raise cancellation case

kelvin-chappell:
To raise a cancellation case in Salesforce, we depend on the subscription that is being cancelled to already exist in Salesforce so that it can be looked up and have the case associated with it.  But when the sync from Zuora hasn't run in a while the subscription won't yet exist in Salesforce.

With this change, instead of looking up the associated subscription we just pass through the name of the subscription and let Salesforce sort out what to do with it.  This should mean that it will still be possible for the case to be successfully recorded in Salesforce and for the API call from MMA to return successfully even when the subscription has yet to be synced with Zuora.

Depends on:
- [x] Code to raise the case ready in Salesforce

### DIFF
--- a/handlers/cancellation-sf-cases-api/src/main/scala/com/gu/cancellation/sf_cases/RaiseCase.scala
+++ b/handlers/cancellation-sf-cases-api/src/main/scala/com/gu/cancellation/sf_cases/RaiseCase.scala
@@ -8,9 +8,9 @@ import com.gu.salesforce.SalesforceGenericIdLookup.{FieldName, LookupValue, SfOb
 import com.gu.salesforce.cases.SalesforceCase
 import com.gu.salesforce.cases.SalesforceCase.Create.WireNewCase
 import com.gu.salesforce.cases.SalesforceCase.GetMostRecentCaseByContactId.TGetMostRecentCaseByContactId
-import com.gu.salesforce.cases.SalesforceCase.{CaseId, CaseSubject, CaseWithId, ContactId, SubscriptionId}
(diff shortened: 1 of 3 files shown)


handlers/cancellation-sf-cases-api/src/main/scala/com/gu/cancellation/sf_cases/RaiseCase.scala
```diff
@@ -47,13 +47,13 @@ object RaiseCase {
 
   val buildWireNewCaseForSalesforce = (
     raiseCaseDetail: RaiseCaseDetail,
-    sfSubscriptionIdContainer: SubscriptionIdContainer,
+    subscriptionName: SubscriptionName,
```
johnduffell, line 50: can we delete SubscriptionIdContainer and its reads in this file completely now?

https://github.com/guardian/support-service-lambdas/blob/452b380e0d449c9d891c2dba911d63e17157275f/handlers/cancellation-sf-cases-api/src/main/scala/com/gu/cancellation/sf_cases/RaiseCase.scala#L34-L35

johnduffell (approved):
looks good, very simple change once we got through it! 👍

handlers/cancellation-sf-cases-api/src/main/scala/com/gu/cancellation/sf_cases/RaiseCase.scala
kelvin-chappell: Yes we can

PR #1447, opened 2022-03-14
Set toAddress and fromAddress to Option

david-pepper:
## What does this change?
We had assumed that FromAddress and ToAddress on EmailMessages can never be null in Salesforce, however a very small set of records contain a null for one of these fields. 

Null values on these fields are causing errors when parsing json to EmailsFromSfResponse.EmailRecord in scala, so I've updated their types to Option[String] from String

### DIFF
--- a/handlers/sf-emails-to-s3-exporter/src/main/scala/com/gu/sf_emails_to_s3_exporter/EmailsFromSfResponse.scala
+++ b/handlers/sf-emails-to-s3-exporter/src/main/scala/com/gu/sf_emails_to_s3_exporter/EmailsFromSfResponse.scala
@@ -10,7 +10,7 @@ object EmailsFromSfResponse {
   case class EmailRecord(
     Id: String,
     ParentId: String,
-    FromAddress: String,
+    FromAddress: Option[String] = None,
     BccAddress: Option[String] = None,
     CcAddress: Option[String] = None,
     FirstOpenedDate: Option[String] = None,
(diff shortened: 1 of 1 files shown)


handlers/sf-emails-to-s3-exporter/src/main/scala/com/gu/sf_emails_to_s3_exporter/EmailsFromSfResponse.scala
```diff
@@ -26,7 +26,7 @@ object EmailsFromSfResponse {
     Status: String,
```
graham228221, line 26: MessageDate is also nillable: https://developer.salesforce.com/docs/atlas.en-us.object_reference.meta/object_reference/sforce_api_objects_emailmessage.htm

not sure what scenario this would be missing from a production record (i think it defaults to createddate) but worth adding as an option?

handlers/sf-emails-to-s3-exporter/src/main/scala/com/gu/sf_emails_to_s3_exporter/EmailsFromSfResponse.scala
```diff
@@ -26,7 +26,7 @@ object EmailsFromSfResponse {
...
     Subject: Option[String] = None,
     TextBody: Option[String] = None,
-    ToAddress: String,
+    ToAddress: Option[String] = None,
     Composite_Key__c: Option[String] = None,
```
graham228221, line 30: We do want ONLY emails that have a composite key, so this doesn't need to be optional (but not a problem that it is, this requirement is handled elsewhere)

graham228221 (approved):

handlers/sf-emails-to-s3-exporter/src/main/scala/com/gu/sf_emails_to_s3_exporter/EmailsFromSfResponse.scala
david-pepper: Yes, that's a great shout. I've made MessageDate an option now as well

PR #1452, opened 2022-03-16
Add schedule to run at 3am and 3:30am every day

david-pepper:
## What does this change?
Schedule the Exporter Lambda to run at 3am and 3:30am every day

Each run will process 2000 records. We expect a maximum of 1500 emails per day to be surfaced for export, so the dual run will allow for up to 4000 emails per day to be exported (in the event of a sudden increase)

### Related Jobs and run times
02:00 **[Salesforce]** batch_CreateAsyncEmailExports - generates queue items for pickup by Lambda
03:00 **[AWS]** Lambda - Export Emails (this PR)
03:30 **[AWS]** Lambda - Export Emails (this PR)
04:00 **[Salesforce]** batch_genericRecordCleanup - Deletes emails that have been exported from Salesforce

### DIFF
--- a/handlers/sf-emails-to-s3-exporter/cfn.yaml
+++ b/handlers/sf-emails-to-s3-exporter/cfn.yaml
@@ -15,12 +15,14 @@ Parameters:
 Mappings:
   StageMap:
     PROD:
-      Schedule: 'rate(15 minutes)'
+      Schedule1: 'rate(cron(00 03 * * ? *))'
+      Schedule2: 'rate(cron(30 03 * * ? *))'
       SalesforceUsername: EmailsToS3APIUser
       AppName: SFEmailsToS3
       ToLowerCase: prod
     DEV:
-      Schedule: 'rate(365 days)'
+      Schedule1: 'rate(365 days)'
+      Schedule2: 'rate(364 days)'
       SalesforceUsername: EmailsToS3APIUser
       AppName: AwsConnectorSandbox
       ToLowerCase: dev


david-pepper:
[Implement scheduling for Lambda and Email Deletion job](https://trello.com/c/xKPf7hw5/2939-implement-scheduling-for-lambda-and-email-deletion-job)

handlers/sf-emails-to-s3-exporter/cfn.yaml
```diff
@@ -354,7 +362,7 @@ Resources:
       Period: 3600
       Statistic: Sum
       Threshold: 1
-      TreatMissingData: ignore
+      TreatMissingData: notBreaching
```
kelvin-chappell, line 365: Be careful not to miss an alarm going off because this will reset to OK before the problem has been fixed.

kelvin-chappell (approved):
It looks fine but if the lambda fails it won't have many chances to recover.  I don't know if that's a problem or not.  By default, a lambda will retry twice if it fails.

handlers/sf-emails-to-s3-exporter/cfn.yaml
david-pepper: Is there an alternative approach that we can take to indicate to the Lambda that the problem has been fixed to take the Alarm out of alarmin state?

kelvin-chappell: Actually in this case, this is fine because no data means no problems.
An alternative approach would be to have another condition for moving the alarm to OK state but that would be overkill here and have the same effect anyway.

handlers/sf-emails-to-s3-exporter/cfn.yaml
```diff
@@ -15,12 +15,14 @@ Parameters:
 Mappings:
   StageMap:
     PROD:
-      Schedule: 'rate(15 minutes)'
+      Schedule1: 'rate(cron(00 03 * * ? *))'
+      Schedule2: 'rate(cron(30 03 * * ? *))'
```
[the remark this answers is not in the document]
david-pepper: Thanks for this!

PR #1461, opened 2022-03-23
[WIP] add basic shell product-move lambda using zio and scala 3

johnduffell:
This PR adds a basic shell lambda for the new product-move API.

So far it uses Scala 3 and ZIO v1 (sttp, zio-logging, zio-json)

I have added code to read config from S3, and make an HTTP call to zuora.  It will then output the results to the logs.

It is WIP but just to get an idea of the direction I have been learning and request for comments on which approaches should be abandoned and any others I should be pusuing.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -408,6 +408,22 @@ lazy val `delivery-problem-credit-processor` = lambdaProject(
   )
 ).dependsOn(`credit-processor`, `salesforce-sttp-client`, effects)
 
+lazy val `product-move-api` = lambdaProject(
+  "product-move-api",
+  "moves a reader from one subscription to another.",
+  Seq(
+    zio,
+    awsEvents,
+    awsLambda,
+    "com.softwaremill.sttp.client3" %% "httpclient-backend-zio1" % "3.5.1"  exclude("org.scala-lang.modules","scala-collection-compat_2.13"),
+    "dev.zio" %% "zio-s3" % "0.3.7"  exclude("org.scala-lang.modules","scala-collection-compat_2.13"),
+    "com.softwaremill.sttp.client3" %% "zio1-json" % "3.5.1",
+    "dev.zio" %% "zio-logging-slf4j" % "0.5.14"
+  )
+)
+  .settings(scalaVersion := "3.1.1")
+  .dependsOn()
+
 lazy val `metric-push-api` = lambdaProject(
   "metric-push-api",
   "HTTP API to push a metric to cloudwatch so we can alarm on errors")
--- a/handlers/product-move-api/README.md
+++ b/handlers/product-move-api/README.md
@@ -1,0 +1,11 @@
+# Product move api
+This is a lambda API to take a subscription ID for a contribution
+and move the subscription to be a digital subscription.
+It would be called from the backend, either manage-frontend or
+possibly support-frontend for acquisition channel based switches.
+
+The intention is to cancel the existing sub in zuora and replace it
+with a fresh one in the same account.  There will also be a confirmation
+email sent via membership-workflow.
+
+Working doc here: https://docs.google.com/document/d/17_es85x-rVrS-ONmHUrX5lVS0bCcgk6ZxOkwvyCQ3xY/edit#heading=h.4hjwot4q08ip
--- a/handlers/product-move-api/src/main/resources/logback.xml
+++ b/handlers/product-move-api/src/main/resources/logback.xml
@@ -1,0 +1,15 @@
+<configuration>
+    <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
+        <encoder>
+            <pattern>%date %level - %message%n%xException</pattern>
+        </encoder>
+    </appender>
+
+    <!-- see https://docs.aws.amazon.com/sdk-for-java/v2/developer-guide/java-dg-logging.html -->
+    <logger name="software.amazon.awssdk" level="WARN"/>
+    <logger name="software.amazon.awssdk.request" level="DEBUG"/>
+
+    <root level="INFO">
+        <appender-ref ref="STDOUT" />
+    </root>
+</configuration>
--- a/handlers/product-move-api/src/main/scala/com/gu/productmove/Handler.scala
+++ b/handlers/product-move-api/src/main/scala/com/gu/productmove/Handler.scala
@@ -1,0 +1,165 @@
+package com.gu.productmove
+
+import com.amazonaws.services.lambda.runtime.*
+import com.amazonaws.services.lambda.runtime.events.{APIGatewayV2HTTPEvent, APIGatewayV2HTTPResponse}
+import software.amazon.awssdk.auth.credentials.*
+import software.amazon.awssdk.regions.Region
+import software.amazon.awssdk.services.s3.model.S3Exception
+import software.amazon.awssdk.utils.SdkAutoCloseable
+import sttp.capabilities.WebSockets
+import sttp.capabilities.zio.ZioStreams
+import sttp.client3.httpclient.zio.{HttpClientZioBackend, SttpClient, send}
+import sttp.client3.*
+import sttp.model.*
+import zio.blocking.*
+import zio.json.*
+import zio.s3.*
+import zio.s3.providers.*
+import zio.stream.ZTransducer
+import zio.{Has, *}
+import zio.logging._
+
+object Handler extends RequestHandler[APIGatewayV2HTTPEvent, APIGatewayV2HTTPResponse] {
+
+  val testInput: String = """{ "dummy": false }"""
+
+  // for testing
+  def main(args: Array[String]): Unit = {
+    val input = new APIGatewayV2HTTPEvent(
+      null,
+      null,
+      null,
+      null,
+      null,
+      null,
+      null,
+      null,
+      null,
+      testInput,
+      false,
+      null
+    )
+    val context = new Context {
+      override def getAwsRequestId: String = ???
+
+      override def getLogGroupName: String = ???
+
+      override def getLogStreamName: String = ???
+
+      override def getFunctionName: String = ???
+
+      override def getFunctionVersion: String = ???
+
+      override def getInvokedFunctionArn: String = ???
+
+      override def getIdentity: CognitoIdentity = ???
+
+      override def getClientContext: ClientContext = ???
+
+      override def getRemainingTimeInMillis: Int = ???
+
+      override def getMemoryLimitInMB: Int = ???
+
+      override def getLogger: LambdaLogger = ???
+    }
+    val response = handleRequest(input, context)
+    println("response: " + response)
+  }
+
+  // this is the main lambda entry point.  It is referenced in the cloudformation.
+  override def handleRequest(input: APIGatewayV2HTTPEvent, context: Context): APIGatewayV2HTTPResponse =
+    Runtime.default.unsafeRun(
+      run(input)
+        .provideCustomLayer(AwsS3.live ++ SttpLayer.live ++ LoggingLayer.live)
+    )
+
+  case class ExpectedInput(dummy: Boolean)
+
+  object ExpectedInput {
+    implicit val decoder: JsonDecoder[ExpectedInput] =
+      DeriveJsonDecoder.gen[ExpectedInput]
+  }
+
+  case class ZuoraRestConfig(
+    baseUrl: String,
+    username: String,
+    password: String,
+  )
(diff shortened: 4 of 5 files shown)


build.sbt
```diff
@@ -408,6 +408,22 @@ lazy val `delivery-problem-credit-processor` = lambdaProject(
...
+  "product-move-api",
+  "moves a reader from one subscription to another.",
+  Seq(
+    zio,
+    awsEvents,
+    awsLambda,
+    "com.softwaremill.sttp.client3" %% "httpclient-backend-zio1" % "3.5.1"  exclude("org.scala-lang.modules","scala-collection-compat_2.13"),
+    "dev.zio" %% "zio-s3" % "0.3.7"  exclude("org.scala-lang.modules","scala-collection-compat_2.13"),
+    "com.softwaremill.sttp.client3" %% "zio1-json" % "3.5.1",
+    "dev.zio" %% "zio-logging-slf4j" % "0.5.14"
```
johnduffell, line 421: I wasn't totally sure about which logger I should be using.  Since I'm on scala 3 there's a more limited choice of loggers on the scala side, but the java backends should work ok.
This post made me wonder whether to use an external library rather https://medium.com/@m.langer798/why-im-not-abandoning-slf4zio-in-favor-of-zio-logging-16d0ae70a1b9

handlers/product-move-api/src/main/resources/logback.xml
```diff
@@ -0,0 +1,15 @@
...
+    </appender>
+
+    <!-- see https://docs.aws.amazon.com/sdk-for-java/v2/developer-guide/java-dg-logging.html -->
+    <logger name="software.amazon.awssdk" level="WARN"/>
+    <logger name="software.amazon.awssdk.request" level="DEBUG"/>
+
+    <root level="INFO">
+        <appender-ref ref="STDOUT" />
+    </root>
+</configuration>
```
johnduffell, line 15: This is jist the standard logback config needed to stop AWS client spamming the console.

handlers/product-move-api/src/main/scala/com/gu/productmove/Handler.scala
```diff
@@ -0,0 +1,165 @@
...
+import zio.json.*
+import zio.s3.*
+import zio.s3.providers.*
+import zio.stream.ZTransducer
+import zio.{Has, *}
+import zio.logging._
+
+object Handler extends RequestHandler[APIGatewayV2HTTPEvent, APIGatewayV2HTTPResponse] {
+
+  val testInput: String = """{ "dummy": false }"""
```
johnduffell, line 24: this will eventually include subscription id, target product, target date and maybe more.

handlers/product-move-api/src/main/scala/com/gu/productmove/Handler.scala
```diff
@@ -0,0 +1,165 @@
...
+      zuoraConfig <- getConfig(stage)
+      subscriptionNumber = "A-S00339056"//DEV - for testing locally
+      sub <- send(
+        basicRequest
+          .get(uri"${zuoraConfig.baseUrl}/subscriptions/$subscriptionNumber")
+          .headers(Map(
+            "apiSecretAccessKey" -> zuoraConfig.password,
+            "apiAccessKeyId" -> zuoraConfig.username
+          ))
+      ).mapError(_.toString)
```
kelvin-chappell, line 115: I would model this as a `Zuora` service, which again is easier to understand in ZIO v2.

build.sbt
johnduffell: zio 2 has logging built in https://ziverge.com/blog/a-preview-of-logging-in-zio-2/
which addresses this point

build.sbt
```diff
@@ -408,6 +408,22 @@ lazy val `delivery-problem-credit-processor` = lambdaProject(
...
+  )
+)
+  .settings(scalaVersion := "3.1.1")
```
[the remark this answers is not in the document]
johnduffell: so just for reference, intellij couldn't find the "provide" method in scala 3, presumably because it's provided by a macro.
I have since updated to the latest version of intellij, and now it can find it no problem 😓

handlers/product-move-api/README.md
```diff
@@ -0,0 +1,11 @@
+# Product move api
+This is an API, implemented by a lambda, to take a subscription ID for a contribution
+and move the subscription to be a digital subscription.
+It would be called from the backend, either manage-frontend or
+possibly support-frontend for acquisition channel based switches.
+
+The intention is to cancel the existing sub in zuora and replace it
+with a fresh one in the same account.  There will also be a confirmation
```
paulbrown1982, line 8: i would envisage there eventually being two modes to this API, one which results in a new subscription (RC->DP upsell), the other which changes the plan within the existing subscription (newspaper upgrade/downgrade).

johnduffell: yes agree once we have RC->DS we can consider how to deal with the more complex products, and it can involve whatever orchestration in zuora etc that we see fit.
It would be ideal to be as consistent as possible, but the reason we have to cancel the zuora sub DS->RC is due to assumptions in members data api and possibly other places about the term end date.  I think those assumptions are more likely to be ok within a single product.

handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/ZuoraClientLive.scala
```diff
@@ -0,0 +1,99 @@
...
+        ).mapError(_.toString)
+      }
+    ZLayer.fromZIO(zuoraZio)
+  }
+
+}
+
+trait ZuoraClient {
+
+  def send[T](request: Request[T, ZioStreams with Effect[Task] with WebSockets]): ZIO[Any, String, Response[T]]
```
kelvin-chappell, line 66: I wonder what a junior dev would make of this type signature.  This is why I have misgivings about STTP and the effect is has on the clarity of the code

project/Dependencies.scala
```diff
@@ -66,6 +66,7 @@ object Dependencies {
 
   // Other
   val zio = "dev.zio" %% "zio" % "1.0.13"
+  val zio2 = "dev.zio" %% "zio" % "2.0.0-RC5"
```
kelvin-chappell, line 69:
```suggestion
  val zio2 = "dev.zio" %% "zio" % "2.0.0-RC6"
```

johnduffell: good idea - but turns out actually this is not possible yet due to no compatible version of `httpclient-backend-zio`
https://github.com/softwaremill/sttp/issues/1433

I get a binary incompatibility until 3.6.x is released of `httpclient-backend-zio`

handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/ZuoraClientLive.scala
johnduffell: resolved in our pairing session,  to remove some of the worst type junk. 
https://github.com/guardian/support-service-lambdas/blob/5dd3e9f5a167e41b607c2eea15f1e158ed728175/handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/rest/ZuoraClient.scala#L67

I think even if we stick with STTP, if we minimise the direct use of sttp in our application code (most of the time we just want to do a REST GET/POST against zuora with json payloads) then it should be pretty interchangable later if we ever decide to switch)

handlers/product-move-api/src/main/scala/com/gu/productmove/GuStageLive.scala
```diff
@@ -0,0 +1,16 @@
+package com.gu.productmove
+
+import zio.{ZIO, ZLayer}
+
+object GuStageLive {
+  enum Stage:
+    case PROD, CODE, DEV
+
+  val layer: ZLayer[Any, String, Stage] = {
```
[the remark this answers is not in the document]
johnduffell: oh interesting idea.
Before reading this, I realised failures in the provide layer were not handled and left to the unsafeRun to throw a FiberFailure.  As part of that I changed it so failures are an `Any`, and we just log and return a 500 error.
What is the benefit of modelling failures in the type system?  My feeling is at the moment any failures we would just check the logs rather than needing them to conform to a type.  I'm happy to convert them all to `String` rather than `Any`, but I'm unconvinced about a proper error type at this point.

handlers/product-move-api/src/main/scala/com/gu/productmove/GuStageLive.scala
```diff
@@ -0,0 +1,16 @@
...
+    val stageZIO: ZIO[Any, Throwable, Stage] = for {
+      stageString <- ZIO.attempt(sys.env.getOrElse("Stage", "DEV"))
```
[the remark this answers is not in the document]
johnduffell: done this, good pointer

handlers/product-move-api/src/main/scala/com/gu/productmove/AwsCredentialsLive.scala
```diff
@@ -0,0 +1,29 @@
...
+    }
+
+  private val lambdaCreds: Layer[InvalidCredentials, EnvironmentVariableCredentialsProvider] =
+    ZLayer.fromZIO {
+      IO.attempt(EnvironmentVariableCredentialsProvider.create())
+        .tap(c => attemptBlocking(c.resolveCredentials()))
+        .mapError(err => InvalidCredentials(err.getMessage))
+    }
+
+  val layer: Layer[InvalidCredentials, AwsCredentialsProvider] = lambdaCreds <> membershipProfile
```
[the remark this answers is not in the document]
johnduffell: good idea, I just copied the approach from the zio-s3 library but it does make sense to use the basic AWS functionalirty.

handlers/product-move-api/src/main/scala/com/gu/productmove/GuStageLive.scala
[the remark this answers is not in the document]
johnduffell: got it - let's keep that in mind and add it when we need it, thanks

kelvin-chappell (approved):
🚀 
I think maybe the logging could be abstracted out into a separate service but could do that later

project/Dependencies.scala
johnduffell: I managed this in the end, the `httpclient-backend-zio` was folded into the main library so I could remove that dep completely.
I then hit a logging issue which meant I had to use a snapshot build.  I will update once the next RC or final version is out.

handlers/product-move-api/src/main/scala/com/gu/productmove/ZIOApiGatewayRequestHandler.scala
```diff
@@ -45,21 +64,65 @@ trait ZIOApiGatewayRequestHandler extends RequestHandler[APIGatewayV2HTTPEvent,
...
+      cause: Cause[Any],
+      context: Map[FiberRef[_], Any],
+      spans: List[LogSpan],
+      annotations: Map[String, String]
+    ): Unit = {
+
+      val now = java.time.Instant.now().toString
+      val indentedMessage = message().replaceAll("\n", "\n ")
+
+      lambdaLogger.log(s"$now: $indentedMessage")
```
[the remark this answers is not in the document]
johnduffell: I have removed the timestamp and it looks ok still.  It wasn't using it anyway, they were slightly different by milliseconds.

I also played around with the multi line logs, and it's not possible to get multi line logs in a lambda, that only seems to apply to the aws logs agent.  So I have removed the indentation workaround as well.  It does mean that JSON and stack traces will likely be a mess, but if you click the magnifying glass in the cloudwatch console, it looks ok.

Interestingly AWS recommend logging in JSON format, which also ignores newlines, so just makes it really messy but with little benefit that I can see.

handlers/product-move-api/cfn.yaml
```diff
@@ -0,0 +1,66 @@
...
+    Description: Bucket to copy files to
+    Type: String
+    Default: membership-dist
+
+Conditions:
+  IsProd: !Equals [ !Ref Stage, "PROD" ]
+
+Globals:
+  Api:
+    OpenApiVersion: 3.0.1
```
johnduffell, line 26: > Note: Setting this property to any valid value will also remove the stage Stage that SAM creates.

https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-resource-api.html#sam-api-openapiversion

As referenced from https://github.com/aws/serverless-application-model/issues/191#issuecomment-515589603

PR #1489, opened 2022-04-11
holiday-stop-api make it return bad request if dates are invalid

johnduffell:
At the moment, if the user tries to shorten a holiday stop but it's too late, we return a 500 error.

This PR makes it return a 400 bad request instead.  This will prevent the alarm and be more appropriate for a REST API.

Also there is a separate change in salesforce to be done - if the CSR clicks a new date and presses the Amend button before the validate request has completed, the request will be made and cause this error.  If the CSR waits until the request has validated, the button would have been disabled.  This will be done separately to this PR.

### DIFF
--- a/lib/holiday-stops/src/main/scala/com/gu/salesforce/holiday_stops/SalesforceHolidayStopRequest.scala
+++ b/lib/holiday-stops/src/main/scala/com/gu/salesforce/holiday_stops/SalesforceHolidayStopRequest.scala
@@ -293,28 +293,26 @@ object SalesforceHolidayStopRequest extends Logging {
             ))(Json.writes[AddHolidayStopRequestDetailBody])
           )}
 
-      existingPublicationsThatWereToBeStopped
-        .filterNot(holidayStopRequestDetail =>
-          issuesData.map(_.issueDate).contains(holidayStopRequestDetail.Stopped_Publication_Date__c.value)
-        )
-        .map( holidayStopRequestDetail => {
-          if(holidayStopRequestDetail.Is_Actioned__c){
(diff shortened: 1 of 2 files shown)


lib/holiday-stops/src/main/scala/com/gu/salesforce/holiday_stops/SalesforceHolidayStopRequest.scala
```diff
@@ -261,9 +263,8 @@ object SalesforceHolidayStopRequest extends Logging {
       startDate: LocalDate,
       endDate: LocalDate,
       issuesData: List[IssueData],
-      existingPublicationsThatWereToBeStopped: List[HolidayStopRequestsDetail],
-      zuoraSubscription: Subscription
```
johnduffell, line 265: unused parameter

lib/holiday-stops/src/main/scala/com/gu/salesforce/holiday_stops/SalesforceHolidayStopRequest.scala
```diff
@@ -292,23 +293,23 @@ object SalesforceHolidayStopRequest extends Logging {
             ))(Json.writes[AddHolidayStopRequestDetailBody])
           )}
```
johnduffell, line 295: this is slightly clearer if you view it with [w=1](https://github.com/guardian/support-service-lambdas/pull/1489/files?w=1#diff-cfa6044d42c42cb2f880ba315b66e989a598a35853becb8826a04dd931ebf825R296)

PR #1495, opened 2022-04-14
Increase threshold for no-requests alarm

kelvin-chappell:
Increase threshold to trigger alarm when no requests for 12 hours, rather than 6.
Number of submissions has fallen significantly since live chat became available.


johnduffell (approved):
great to have the extra context in the descruiptuion, thanks 👍

PR #1504, opened 2022-04-21
Increase timeout for contact us form submissions

kelvin-chappell:
We're seeing errors in the contact-us lambda logs because form submits are timing out waiting for a response.
They do seem to write to Salesforce cases but don't return to client quickly enough.


graham228221:
[ALARM: "5XX rate from contact-us-api-PROD" in EU (Ireland)](https://trello.com/c/832EBNkC/3103-alarm-5xx-rate-from-contact-us-api-prod-in-eu-ireland)

david-pepper (approved):

PR #1506, opened 2022-04-25
Bump Salesforce API version to latest

kelvin-chappell:
Salesforce API v20 is about to be decommissioned so I've upgraded all references to it to the latest version, which is currently v54.0.
I've also reduced the number of references to the minimum I can without restructuring the project.


david-pepper (approved):
Looks good, thanks Kelvin

PR #1512, opened 2022-04-28
Resolve reference ID format error introduced by Salesforce API v 54

kelvin-chappell:
Since upgrading the Salesforce API to v 54, we see error entries in the log like this:
> Provided referenceId ('LinkDeliveryRecord-xyz') must start with a letter or a number, and it can contain only letters, numbers and underscores ('_').

This behaviour is documented in the [API developer guide](https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/resources_composite_subrequest_result.htm).

To fix it. I've modified the name of the offending reference ID.  These IDs aren't referred to in subsequent parts of the composite request so this seems safe to do.

### DIFF
--- a/handlers/delivery-records-api/src/main/scala/com/gu/delivery_records_api/SFApiDeliveryProblemCase.scala
+++ b/handlers/delivery-records-api/src/main/scala/com/gu/delivery_records_api/SFApiDeliveryProblemCase.scala
@@ -116,7 +116,7 @@ object SFApiCompositeCreateDeliveryProblem {
         Priority = detail.priority
       )
     )) ++ detail.deliveryRecords.map(deliveryRecord => SFApiCompositePart[SFApiCompositePartBody](
-      referenceId = s"LinkDeliveryRecord-${deliveryRecord.id}",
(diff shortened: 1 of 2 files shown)


johnduffell (approved):
nice debugging there! 👍

graham228221:
@kelvin-chappell @david-pepper so would this have **always** been silently failing part of the composite request? I.e. the Delivery records would not have been updated to reference the Case? 

And now this referenceId is fixed, this update should now work? Just wondering if we will see a change of behaviour in Salesforce.

kelvin-chappell:
> @kelvin-chappell @david-pepper so would this have **always** been silently failing part of the composite request? I.e. the Delivery records would not have been updated to reference the Case?

The bad reference IDs weren't being used elsewhere.  The parent case ID, which was being referred to by delivery records, was in the right format so should have been fine.  It's only the delivery record reference IDs that were bad.

> And now this referenceId is fixed, this update should now work? Just wondering if we will see a change of behaviour in Salesforce.

I don't think there'll be a change of behaviour.  Except that it will work again after today's short outage!

PR #1514, opened 2022-04-29
Upgrade transient jackson dependencies due to vulnerability CVE-2020-36518

eleanombre:
## What does this change?
Specifies transient dependencies so that every dependency that uses Jackson will use a version where this specific vulnerability has been fixed.

## How to test
`sbt dependencyTree` however in this project I temporarily installed the sbt dependency tree plugin (included `addDependencyTreePlugin` in `plugins.sbt`) in order to run `whatDependsOn com.fasterxml.jackson.core jackson-databind` to narrow dependencies down further. 

We will not be vulnerable to this High risk vulnerability.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -70,13 +70,15 @@ lazy val zuora = library(project in file("lib/zuora"))
     effects % "test->test"
   )
   .settings(
-    libraryDependencies ++= Seq(okhttp3, playJson, scalatest) ++ jacksonDependencies ++ logging
+    libraryDependencies ++= Seq(okhttp3, playJson, scalatest) ++ logging,
+    dependencyOverrides ++= jacksonDependencies
   )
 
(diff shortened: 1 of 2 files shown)


eleanombre:
SBT allows specifying [`depdencyOverrides`](https://www.scala-sbt.org/1.x/docs/Library-Management.html#Overriding+a+version) in project settings. I feel like this is a nicer solution as we then know that jackson (or other deps) aren't a direct dependency but an override. You can see the change in [8baa7b1](https://github.com/guardian/support-service-lambdas/pull/1514/commits/8baa7b18141df9d776225c643d8d16862eaecbb4). @guardian/retention do we prefer this? If yes I will use this method in changes from now.

kelvin-chappell (approved):
Good idea to use dependencyOverrides! 👍

PR #1519, opened 2022-05-04
Enable holiday stop operations that were broken by API upgrade

kelvin-chappell:
This is another bug introduced by #1509 
The solution is the same as #1512, ie. in a composite request ensure that reference IDs are well-formed.
In this case, these aren't referenced anywhere so they can be arbitrary.

### DIFF
--- a/lib/effects/src/test/scala/com/gu/effects/SFTestEffects.scala
+++ b/lib/effects/src/test/scala/com/gu/effects/SFTestEffects.scala
@@ -31,7 +31,7 @@ object SFTestEffects {
   def cancelSuccess(referenceId: String, price: Double) = (
     POSTRequest(
       s"/services/data/v$salesforceApiVersion/composite/",
-      s"""{"allOrNone":true,"compositeRequest":[{"method":"PATCH","url":"/services/data/v$salesforceApiVersion/sobjects/Holiday_Stop_Requests_Detail__c/HSD-1","referenceId":"CANCEL DETAIL : $referenceId","body":{"Actual_Price__c":$price,"Charge_Code__c":"ManualRefund_Cancellation"}}]}"""
+      s"""{"allOrNone":true,"compositeRequest":[{"method":"PATCH","url":"/services/data/v$salesforceApiVersion/sobjects/Holiday_Stop_Requests_Detail__c/HSD-1","referenceId":"CANCEL_DETAIL : $referenceId","body":{"Actual_Price__c":$price,"Charge_Code__c":"ManualRefund_Cancellation"}}]}"""
     ),
       HTTPResponse(200, """{ "compositeResponse" : [ { "httpStatusCode": 200 } ]}""".stripMargin)
   )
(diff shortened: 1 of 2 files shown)


lib/effects/src/test/scala/com/gu/effects/SFTestEffects.scala
```diff
@@ -31,7 +31,7 @@ object SFTestEffects {
   def cancelSuccess(referenceId: String, price: Double) = (
     POSTRequest(
       s"/services/data/v$salesforceApiVersion/composite/",
-      s"""{"allOrNone":true,"compositeRequest":[{"method":"PATCH","url":"/services/data/v$salesforceApiVersion/sobjects/Holiday_Stop_Requests_Detail__c/HSD-1","referenceId":"CANCEL DETAIL : $referenceId","body":{"Actual_Price__c":$price,"Charge_Code__c":"ManualRefund_Cancellation"}}]}"""
+      s"""{"allOrNone":true,"compositeRequest":[{"method":"PATCH","url":"/services/data/v$salesforceApiVersion/sobjects/Holiday_Stop_Requests_Detail__c/HSD-1","referenceId":"CANCEL_DETAIL : $referenceId","body":{"Actual_Price__c":$price,"Charge_Code__c":"ManualRefund_Cancellation"}}]}"""
```
johnduffell, line 34: how about the other spaces?
```suggestion
      s"""{"allOrNone":true,"compositeRequest":[{"method":"PATCH","url":"/services/data/v$salesforceApiVersion/sobjects/Holiday_Stop_Requests_Detail__c/HSD-1","referenceId":"CANCEL_DETAIL_$referenceId","body":{"Actual_Price__c":$price,"Charge_Code__c":"ManualRefund_Cancellation"}}]}"""
```

lib/effects/src/test/scala/com/gu/effects/SFTestEffects.scala
```diff
@@ -31,7 +31,7 @@ object SFTestEffects {
   def cancelSuccess(referenceId: String, price: Double) = (
```
johnduffell, line 31: is this used anywhere?

lib/holiday-stops/src/main/scala/com/gu/salesforce/holiday_stops/SalesforceHolidayStopRequest.scala
```diff
@@ -341,10 +341,10 @@ object SalesforceHolidayStopRequest extends Logging {
       val requestDetailParts = holidayStopRequestsDetails
         .map { requestDetail =>
           CompositePart(
-            "PATCH",
-            s"$holidayStopRequestsDetailSfObjectsBaseUrl/${requestDetail.Id.value}",
-            "CANCEL DETAIL : " + idGenerator,
-            Json.toJson(CancelHolidayStopRequestDetailBody(requestDetail.Actual_Price__c, requestDetail.Charge_Code__c))
+            method = "PATCH",
+            url = s"$holidayStopRequestsDetailSfObjectsBaseUrl/${requestDetail.Id.value}",
+            referenceId = "CANCEL_DETAIL : " + idGenerator,
```
johnduffell, line 346: `idGenerator` is a UUID so should be fine

johnduffell (approved):
ok as long as the other spaces are ok in them all?

lib/effects/src/test/scala/com/gu/effects/SFTestEffects.scala
kelvin-chappell: It isn't referenced anywhere but it's required by the composite request so it's just there for json serialising.

PR #1521, opened 2022-05-05
Add logging around Salesforce requests to isolate long-running query

kelvin-chappell:
This is to isolate a query that is frequently timing out.

We've wrapped logging around all the calls to Salesforce.  The next step is to optimise the query once it's found.

Also the log was full of AWS SDK debug entries so adding logback config should minimise that.


graham228221 (approved):
Thanks Kelvin!

jonflynng (approved):

PR #1524, opened 2022-05-06
Remove hyphens from reference IDs

kelvin-chappell:
Since upgrading Salesforce API to v 54, we've seen errors as reference IDs are no longer allowed to contain hyphens.

See:
#1506 
#1508 
#1509 
#1512 
#1519
etc.


graham228221:
[CSR bug report (severity: 2) JSON_PARSER_ERROR when editing a Holiday Stop in Salesforce (API upgrade related)](https://trello.com/c/5iQs7fFC/3127-csr-bug-report-severity-2-jsonparsererror-when-editing-a-holiday-stop-in-salesforce-api-upgrade-related)

PR #1541, opened 2022-05-24
update all lambdas to amazon java 11

johnduffell:
This PR updates all lambdas to use java11 runtime.  This is needed because Zio/scala3/sttp uses the built in java11 http client,  so it won't build on java 8.  If we flip the build over to java 11 there may be a class compatibility issue.

There are other solutions, for example adding a separate build for the product-move-api (which could even be a github action rather than a TC build)
Or switching sttp to use a non java11 backend.
Or even moving product-move-api to another repo.

I picked this approach because java8 is a legacy runtime and java11 is the current one according to AWS docs.  Also I think this is the quickest way to get this working, as I don't expect any other changes needed other than the runtime and build.

Once this is shipped, I can change TC to build for Java11:
> you can simply update the parameters in the config of the build by setting:
>
> env.JAVA_HOME to %env.JDK_11_0%

From https://docs.google.com/document/d/1ZR-YnaXCT5_gLVmTCeGs0mWd3KPaAozPjQK8uUzHZ9w/edit#heading=h.2c51544oeq0b

### DIFF
--- a/handlers/batch-email-sender/cfn.yaml
+++ b/handlers/batch-email-sender/cfn.yaml
@@ -119,7 +119,7 @@ Resources:
         - "LogAndWriteToEmailQueueRole"
         - Arn
       MemorySize: 1536
-      Runtime: java8.al2
+      Runtime: java11
       Timeout: 300
     DependsOn:
     - "LogAndWriteToEmailQueueRole"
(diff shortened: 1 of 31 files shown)


kelvin-chappell (approved):
That's another one off the health list! 👍

PR #1542, opened 2022-05-24
fix compile error in HSP for java 11

johnduffell:
folloiwing https://github.com/guardian/support-service-lambdas/pull/1541
I updated the build to use java11, then it flagged up a previously hidden compile error.
URIO doesn't allow a failure type in ZIO, however somehow in java 8 it was allowing an IOException.  In java 11 it correctly doesn't compile.  This fixes it.


kelvin-chappell (approved):
That's weird! 👍

PR #1544, opened 2022-05-25
Increase available memory for delivery credit processor

kelvin-chappell:
We're seeing metaspace capacity errors in the log on every run, which we've also seen in other lambdas recently.
So increasing memory available from 512 MB to 1024 MB.


paulbrown1982 (approved):
👍

jonflynng (approved):

PR #1546, opened 2022-05-25
use the right bucket for the lambda code

johnduffell:
I made it write to support-service-lambdas-dist but read from membership-dist.

This PR fixes it to use support-service-lambdas-dist in both places.

I will delete the erroneous artifact from membership-dist also


kelvin-chappell (approved):
👍

PR #1549, opened 2022-06-01
Initial spec of product movement API

kelvin-chappell:
This is the initial spec based on early discussions.  
It's likely to change based on the journeys through MMA and presentation data required.

The easiest way to view the spec is to copy it into the [Swagger Editor](https://editor.swagger.io/)

The benefit of Open API is that it can be opened in Postman and other tools, it can be validated and potentially be used for code generation.  It's really for published APIs but I was just experimenting with it.

We could consider using [Tapir](https://tapir.softwaremill.com/en/latest/) to generate the spec from the code in future.
The code and the spec are likely to drift apart very quickly without some autosync of some kind.

### DIFF
--- a/handlers/product-move-api/ProductMovementAPI.yaml
+++ b/handlers/product-move-api/ProductMovementAPI.yaml
@@ -9,7 +9,7 @@ info:
 
 paths:
 
-  /availableProductMoves/{subscriptionName}:
+  /available-product-moves/{subscriptionName}:
     get:
       summary: Gets available products that can be moved to from the given subscription.
       description:
@@ -41,13 +41,21 @@ paths:
         504:
           description: Timeout from an upstream system.
 
-  /productMove:
+  /product-move/{subscriptionName}:
     post:
       summary: Replaces the existing subscription with a new one.
       description:
         Cancels the existing subscription and replaces it with a new subscription
         to a different type of product.
         Also manages all the service comms associated with the movement.
+      parameters:
+        - name: subscriptionName
+          in: path
+          description: Name of subscription to be moved to a different product.
+          example: A-S000001
+          required: true
+          schema:
+            type: string
       requestBody:
         description: Definition of required movement.
         required: true
@@ -56,12 +64,8 @@ paths:
             schema:
               type: object
               required:
-                - sourceSubscriptionName
                 - targetProductId
               properties:
-                sourceSubscriptionName:
-                  description: Name of source subscription that is to be cancelled.
-                  type: string
                 targetProductId:
                   description: ID of target product that new subscription will be for.
                   type: string


handlers/product-move-api/ProductMovementAPI.yaml
```diff
@@ -0,0 +1,134 @@
...
+info:
+  title: Product Movement API
+  description:
+    API to facilitate replacing an existing subscription 
+    with another subscription to a different type of product.
+  version: 0.0.1
+
+paths:
+
+  /eligibles/{subName}:
```
johnduffell, line 12: doesn't quite seem clear what it does from the "eligibles", perhaps we can be a little more verbose as it won't be too costly.
```suggestion
  /availableSwitches/{subName}:
```
or `listSwitches`?
or other suggestions

handlers/product-move-api/ProductMovementAPI.yaml
```diff
@@ -0,0 +1,134 @@
...
+    get:
+      summary: Gets eligible target products for the given subscription.
+      description:
+        Returns an array of eligible products 
+        or an empty array if there aren't any for the given subscription.
```
johnduffell, line 17:
```suggestion
        Returns an array of eligible product switches for the given subscription, if any.
```

handlers/product-move-api/ProductMovementAPI.yaml
```diff
@@ -0,0 +1,134 @@
...
+      parameters:
+        - name: subName
+          in: path
+          description: Name of subscription whose eligibility for movement is to be checked.
```
johnduffell, line 21:
```suggestion
          description: Name of subscription whose eligibility for movement is to be checked. e.g. A-S000001
```

handlers/product-move-api/ProductMovementAPI.yaml
```diff
@@ -0,0 +1,134 @@
...
+paths:
+
+  /eligibles/{subName}:
+    get:
+      summary: Gets eligible target products for the given subscription.
+      description:
+        Returns an array of eligible products 
+        or an empty array if there aren't any for the given subscription.
+      parameters:
+        - name: subName
```
johnduffell, line 19:
```suggestion
        - name: subscriptionName
```

handlers/product-move-api/ProductMovementAPI.yaml
```diff
@@ -0,0 +1,134 @@
...
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: '#/components/schemas/product'
+        404:
+          description: No such subscription.
+        500:
+          description: Unable to respond.
```
johnduffell, line 37: do we need to document 500 error?

handlers/product-move-api/ProductMovementAPI.yaml
johnduffell: after a bit of [googleing](https://www.google.com/search?q=rest+best+practices) I wonder if we should have a list of possible switches
```suggestion
  /{subscriptionName}/switches:
```
and then a POST or PUT to the same endpoint to do a switch?

I'm not sure how well we fit in with the Restful principles, they seem to be more database oriented, but worth thinking about.

handlers/product-move-api/ProductMovementAPI.yaml
```diff
@@ -0,0 +1,134 @@
...
+      requestBody:
+        description: Definition of required movement.
+        required: true
+        content:
+          application/json:
+            schema:
+              type: object
+              required:
+                - sourceSubName
+                - targetProdId
```
johnduffell, line 55:
```suggestion
                - subscriptionName
                - targetProductId
```

handlers/product-move-api/ProductMovementAPI.yaml
```diff
@@ -0,0 +1,134 @@
...
+                  description: Name of source subscription that is to be cancelled.
+                  type: string
+                targetProdId:
+                  description: ID of target product that new subscription will be for.
+                  type: string
+                moveDate:
+                  description: Service start date of new subscription.
+                  example: 2022-01-01
+                  default: Next billing date
+                  type: string
```
johnduffell, line 67: I don't think we need this yet.
Also it sounds like the default will always be "today+14 days" based on discussions before I went away.  This works because there's no refund for contribution, so we could leave an overlap in zuora without causing an issue.
This has the added benefit of allowing siwtching annual contributors  more easily as well.

handlers/product-move-api/ProductMovementAPI.yaml
kelvin-chappell: I think `availableProductMoves` would be better as makes it consistently a `move` rather than a `switch`.
Then perhaps `productMove` as the actual move endpoint.

handlers/product-move-api/ProductMovementAPI.yaml
kelvin-chappell: Wanted to emphasise that we still return an (empty) array even if there aren't any eligible products.

handlers/product-move-api/ProductMovementAPI.yaml
kelvin-chappell: I think this can go explicitly in the `example` field

handlers/product-move-api/ProductMovementAPI.yaml
kelvin-chappell: If we can think of likely failing scenarios they should probably go as separate error codes such as timeouts and bad resource state but I didn't think there was anything specific to say about the 500 error as it's a kind of fall-through state.
Unless you mean the description as it stands seems a bit redundant.  It's just there to please the Open API spec.

handlers/product-move-api/ProductMovementAPI.yaml
kelvin-chappell: Yes I agree better to make these clearer.
But not sure if wanted to make clear that subscription name will change as result of move so perhaps call it `originSubscriptionName` or `initialSubscriptionName`?

handlers/product-move-api/ProductMovementAPI.yaml
kelvin-chappell: Done this.

handlers/product-move-api/ProductMovementAPI.yaml
kelvin-chappell: Changed this

handlers/product-move-api/ProductMovementAPI.yaml
kelvin-chappell: Removed the `moveDate` field

handlers/product-move-api/ProductMovementAPI.yaml
kelvin-chappell: Added some more specific error responses.

johnduffell: yeah sorry I meant I think we can leave it out, as anything can return a 500 error and there's no real solution

PR #1552, opened 2022-06-09
use fromFunction for layer construction

johnduffell:
This is a tiny PR, after reading https://softwaremill.com/structuring-zio-2-applications/
I realised we don't need so much boilerplate for creating a layer from a constructor.

I also took a second to make the braces vs indentation slightly more consistent.  I'm not sure what we will settle on, but it makes sense to try out the indentation based blocks and see how it feels.

### DIFF
--- a/handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/GetSubscription.scala
+++ b/handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/GetSubscription.scala
@@ -13,22 +13,16 @@ import sttp.model.Uri
 import zio.json.*
 import zio.{IO, RIO, Task, URLayer, ZIO, ZLayer}
 
-object GetSubscriptionLive {
-
-  val layer: URLayer[ZuoraGet, GetSubscription] =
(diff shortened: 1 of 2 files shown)


handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/GetSubscription.scala
```diff
@@ -13,22 +13,16 @@ import sttp.model.Uri
...
 import zio.{IO, RIO, Task, URLayer, ZIO, ZLayer}
 
-object GetSubscriptionLive {
-
-  val layer: URLayer[ZuoraGet, GetSubscription] =
-    ZLayer(ZIO.serviceWith[ZuoraGet](GetSubscriptionLive(_)))
-
-}
+object GetSubscriptionLive :
+  val layer: URLayer[ZuoraGet, GetSubscription] = ZLayer.fromFunction(GetSubscriptionLive(_))
```
johnduffell, line 17: this is the actual change - fromFunction is the simpler way of doing this.

kelvin-chappell (approved):
👍

PR #1570, opened 2022-06-16
add tapir and generate swagger to /docs/ route, and return stub data

johnduffell:
This PR adds tapir https://github.com/softwaremill/tapir to the router, so that we have control of the endpoints that we serve.
This means we don't have to write cloudformation to do the routing, and we can also serve swagger docs which must be up to date.

The endpoints are designed to generate effectively the same yaml as the hand crafted one.  It will also be self-served by the server itself.
The stub/hard coded data returned will be the same as assumed by the client side.

Next step is the add the implementation calls to zuora.


handlers/product-move-api/cfn.yaml
```diff
@@ -71,8 +71,8 @@ Resources:
         ProductMovement:
           Type: Api
           Properties:
-              Path: '/product-move'
-              Method: post
+              Path: '/{proxy+}'
+              Method: any
```
johnduffell, line 75: now all routes fire up the lambda, and it can decide itself to return 404 or route it wherever.
This saves us maintaining cloudformation.

build.sbt
```diff
@@ -439,7 +439,11 @@ lazy val `product-move-api` = lambdaProject(
     "com.softwaremill.sttp.client3" %% "zio-json" % "3.6.2",
     "dev.zio" %% "zio-logging-slf4j" % "2.0.0-RC10",
     "dev.zio" %% "zio-test" % zio2Version % Test,
-    "dev.zio" %% "zio-test-sbt" % zio2Version % Test
+    "dev.zio" %% "zio-test-sbt" % zio2Version % Test,
+    "com.softwaremill.sttp.tapir" %% "tapir-core" % tapirVersion,
+    "com.softwaremill.sttp.tapir" %% "tapir-json-zio" % tapirVersion,
+    "com.softwaremill.sttp.tapir" %% "tapir-aws-lambda" % tapirVersion,
```
johnduffell, line 445: this pulls in the whole of cats, which is not really needed as we are using ZIO.  I think we might be able to contribute back a tapir-aws-lambda-zio module to get around this, if they don't do it themselves.

handlers/product-move-api/deploy-CODE-cfn.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
```
johnduffell, line 1: just a handy script to avoid lots of clicking when developing the cfn

kelvin-chappell (approved):
I applaud the spirit of adventure and breaking ground.  But I also have mixed feelings about the added complexity to the code and the possible future problems with maintaining dependencies.  Let's go with it and weigh it up when we get the endpoints fully working and synchronised API docs.

Perhaps you should add an architectural note to the readme to say why we're using Tapir and what we hope to get out of it so that future generations can understand.

🚀

build.sbt
```diff
@@ -31,6 +31,29 @@ val scalaSettings = Seq(
   autoCompilerPlugins := true
 )
 
+val scala3Settings = Seq(
```
johnduffell, line 34: there were a load of errors caused by invalid scalac options, so I took the time to separate these.  There are some that could be kept in common, but I didn't so far.

build.sbt
```diff
@@ -439,14 +462,24 @@ lazy val `product-move-api` = lambdaProject(
...
     // needed for zio snapshot to get this PR https://github.com/zio/zio/pull/6775
     resolvers += "Sonatype OSS Snapshots" at "https://oss.sonatype.org/content/repositories/snapshots",
-    testFrameworks += new TestFramework("zio.test.sbt.ZTestFramework")
+    testFrameworks += new TestFramework("zio.test.sbt.ZTestFramework"),
+    genDocs := Def.taskDyn {
+      val targetPath = target.value.toString
+      Def.task {
+        (Compile / runMain).toTask(" com.gu.productmove.MakeDocsYaml " + targetPath + "/APIDocs.yaml").value
+      }
+    }.value
```
johnduffell, line 482: this is needed so we can run `sbt genDocs` and get the api docs written to a file in the target/ directory.  This might be necesary if we want to write the API docs for a branch, we might make them available from a build. 
Only sbt knows where the target directory is, and I didn't want to hard code the path in the scala code.

handlers/product-move-api/ProductMovementAPI.yaml
```diff
@@ -1,141 +1,144 @@
 openapi: 3.0.3
-
```
johnduffell, line 2: the changes in this file should be inconsequential, they were needed to format the file to be similar to the generated one, so I could use a diff to check them.
In reality this file could be deleted now, but I wanted to leave the diff here to show the sort of changes there will be in the file.

handlers/product-move-api/cfn.yaml
```diff
@@ -68,11 +68,31 @@ Resources:
...
+              Ref: ProductMoveApiGateway
+        Docs:
+          Type: Api
+          Properties:
+            Path: '/docs/{proxy+}'
+            Method: any
+            Auth:
+              ApiKeyRequired: false
+            RestApiId:
+              Ref: ProductMoveApiGateway
```
johnduffell, line 88: these had to be specifically called out at api gateway level in order to make them accessible via the browser wihtout an api key.

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/available/AvailableProductMovesEndpointTypes.scala
```diff
@@ -0,0 +1,106 @@
...
+
+@encodedName("billing")
+@description("Amount and frequency of billing.")
+case class Billing(
+  @description("Absolute amount that will be billed in pence or cents etc.\nEither this field or the percentage field will be populated.")
+  @encodedExample(1199)
+  amount: Option[Int],
+  @description("Percentage of standard amount that will be billed.\nThis field only makes sense if the billing object is attached to an introductory offer.\nEither this field or the amount field will be populated.")
+  @encodedExample(50)
+  percentage: Option[Int], // Either - amount or percentage?
```
johnduffell, line 59: this Billing structure might not make sense - if it's a product it can't have a percentage as a price, and there must be a currency.  However if it's a percentage discount it couldn't have a currency, or a frequency.  I'm not changing it in this PR but I think it could be tightened up, assuming we can generate a suitable yaml.

johnduffell:
[Update move-product endpoint to use stub structures and data](https://trello.com/c/dnEiUpHj/3200-update-move-product-endpoint-to-use-stub-structures-and-data)

johnduffell:
[Create second endpoint for "get-eligibility" operation](https://trello.com/c/BxbXcIqH/3199-create-second-endpoint-for-get-eligibility-operation)

build.sbt
paulbrown1982: Is that likely to slow down the Lambda's boot time? There is a real user experience benefit in in the eligibility API call responding as fast as possible.

johnduffell: it would slow it down in some way, but given it's not actually used, it may well be only in loading the bigger jar, rather than initialising the classes and JIT compliling them etc.
It's a great point about keeping an eye on the response times.  Although this will be a background call in manage, if we're using this while CSRs are on the phone, we might want to either make sure the lambda is kept warm (like support-workers are) or something more drastic.  A basic scala lambda is approaching 10s start time.

johnduffell: another interesting point is that although we're using the AWS lambda interpreter at the moment, we could change to a play interepreter and run it in an EC2 instance, without having to change the DSL code that defines the endpoints.
e.g. https://tapir.softwaremill.com/en/latest/server/play.html

handlers/product-move-api/ProductMovementAPI.yaml
johnduffell: I decided to delete it before merging, the file will be available "live" from the API at https://<id-here>.execute-api.eu-west-1.amazonaws.com/<stage-here>/docs/docs.yaml

project/Dependencies.scala
```diff
@@ -111,6 +112,8 @@ object Dependencies {
   // to resolve merge clash of 'module-info.class'
   // see https://stackoverflow.com/questions/54834125/sbt-assembly-deduplicate-module-info-class
   val assemblyMergeStrategyDiscardModuleInfo = assembly / assemblyMergeStrategy := {
+    case PathList("META-INF", "maven", "org.webjars", "swagger-ui", "pom.properties") =>
+      MergeStrategy.singleOrError
```
johnduffell, line 116: this addition is needed as per https://tapir.softwaremill.com/en/latest/docs/openapi.html#using-swaggerui-with-sbt-assembly

johnduffell:
tested in CODE and seems to be working (returning stub data)

PR #1587, opened 2022-06-27
Api spec updates to keep responses consistent with responses from members-data-api

jonflynng:
Currently members-data-api returns amounts in pence, and intervals use "month" and "year" rather than the plural "Months" and "Years".

Example of a response from members-data-api: https://github.com/guardian/manage-frontend/blob/70fb9fe78f18b0fc4e73d3a2524e887e3be9ba1f/app/client/fixtures/productDetail.ts#L525-L531

### DIFF
--- a/handlers/product-move-api/ProductMovementAPI.yaml
+++ b/handlers/product-move-api/ProductMovementAPI.yaml
@@ -119,7 +119,7 @@ components:
         - currency
       properties:
         amount:
-          type: number
+          type: integer
           description:
             Absolute amount that will be billed in pence.
             Either this field or the percentage field will be populated.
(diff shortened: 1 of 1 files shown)


handlers/product-move-api/ProductMovementAPI.yaml
```diff
@@ -118,11 +119,11 @@ components:
         - currency
       properties:
         amount:
-          type: number
+          type: integer
           description:
-            Absolute amount that will be billed.
+            Absolute amount that will be billed in pence.
```
kelvin-chappell, line 124: I don't know what the generic term for pence, cents, etc is.
```suggestion
            Absolute amount that will be billed in pence or cents etc.
```

kelvin-chappell (approved):
👍

PR #1620, opened 2022-07-14
update to zio 2.0.0

johnduffell:
This draft PR updates to zio 2 and sttp 3.7.2 (updates together needed to avoid binary incompatibility)

The compatible version of sttp was released here
https://github.com/softwaremill/sttp/pull/1469

The binary incompatibillity observed was that when you ran the lambda it failed to wire the layers with this error:
`java.lang.NoSuchMethodError: 'zio.ZIO zio.ZIO$.attempt(scala.Function0, java.lang.Object)'`


build.sbt
```diff
@@ -457,9 +457,9 @@ lazy val `product-move-api` = lambdaProject(
     zio2,
     awsEvents,
     awsLambda,
-    "com.softwaremill.sttp.client3" %% "zio" % "3.6.2"  exclude("org.scala-lang.modules","scala-collection-compat_2.13"),
+    "com.softwaremill.sttp.client3" %% "zio" % sttpVersion  exclude("org.scala-lang.modules","scala-collection-compat_2.13"),
     awsS3,
-    "com.softwaremill.sttp.client3" %% "zio-json" % "3.6.2",
+    "com.softwaremill.sttp.client3" %% "zio-json" % sttpVersion,
```
johnduffell, line 462: sttpVersion can be incremented when the new zio-2 compatible version number is known

johnduffell: this has been incremented now to 3.7.2

handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/rest/ZuoraClient.scala
```diff
@@ -36,7 +36,7 @@ object ZuoraClientLive {
     s"$basePath/$relativePath"
   }
 
-  val layer: ZLayer[AwsS3 with Stage with SttpClient, String, ZuoraClient] =
+  val layer: ZLayer[AwsS3 with Stage with SttpBackend[Task, Any], String, ZuoraClient] =
```
johnduffell, line 39: https://github.com/softwaremill/sttp/pull/1499

jonflynng (approved):
I tested this out by merging it into my `available-product-moves-endpoint` branch to observe any compiler errors. Looks all good, just had to remove the couple of `SttpClient` and `send` imports and it also seems that `setDateTime` has been removed from zio.test:

Screenshot 2022-08-12 at 17 13 22

I've used the `setTime` function instead.

project/Dependencies.scala
```diff
@@ -115,6 +115,7 @@ object Dependencies {
     case PathList("META-INF", "maven", "org.webjars", "swagger-ui", "pom.properties") =>
       MergeStrategy.singleOrError
     case PathList(ps @ _*) if ps.last == "module-info.class" => MergeStrategy.discard
+    case PathList(ps @ _*) if ps.last == "deriving.conf" => MergeStrategy.discard
```
johnduffell, line 118: I think we should ideally concatenate here for performance reasons https://github.com/scalaz/scalaz-deriving#deriving
I think this line may be copied in future to other projects to it would make sense to get it right first time.

johnduffell:
```suggestion
    case PathList(ps @ _*) if ps.last == "deriving.conf" => MergeStrategy.filterDistinctLines
```
> The config file is plain text with one line per wiring, formatted: fqn.TypeClass=fqn.DerivedTypeClass.method, comments start with #.

johnduffell: https://github.com/sbt/sbt-assembly
> MergeStrategy.filterDistinctLines also concatenates, but leaves out duplicates along the way. There is also an overload that accepts a Charset for reading the lines

PR #1621, opened 2022-07-14
Add zuora subscribe call to product-move API

jonflynng:
This is part of the product-movement-api which takes a user with a recurring contribution and switches them to a digital subscription.

 The endpoint takes in the current subscription name/Id and a targetProductId as input, which represents the ID of the product rate plan we want to subscribe the user to. We first call the `GetSubscription` service to obtain the user's `accountKey` as we need this for the subsequent subscribe call. We then call zuora's `/v1/subscriptions` ([docs here](https://www.zuora.com/developer/api-reference/#operation/POST_Subscription)) endpoint to subscribe the user to the new rateplan.

### DIFF
--- a/handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/move/ProductMoveEndpointTypes.scala
+++ b/handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/move/ProductMoveEndpointTypes.scala
@@ -1,9 +1,9 @@
 package com.gu.productmove.endpoint.move
 
 import com.gu.productmove.framework.InlineSchema.inlineSchema
-import com.gu.productmove.endpoint.available.{Currency, MoveToProduct}
+import com.gu.productmove.endpoint.available.{Currency, MoveToProduct, TimePeriod}
 import sttp.tapir.Schema
-import sttp.tapir.Schema.annotations.{description, encodedName}
+import sttp.tapir.Schema.annotations.{description, encodedExample, encodedName}
 import sttp.tapir.generic.{Configuration, Derived}
 import zio.json.{DeriveJsonDecoder, DeriveJsonEncoder, JsonDecoder, JsonEncoder}
 import sttp.tapir.Schema
@@ -39,5 +39,4 @@ object ProductMoveEndpointTypes {
   given JsonDecoder[Success] = DeriveJsonDecoder.gen[Success] // needed to keep tapir happy
   given JsonDecoder[NotFound] = DeriveJsonDecoder.gen[NotFound] // needed to keep tapir happy
   given JsonDecoder[OutputBody] = DeriveJsonDecoder.gen[OutputBody] // needed to keep tapir happy
-
 }
--- a/handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/move/ProductMoveEndpointTypes.scala
+++ b/handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/move/ProductMoveEndpointTypes.scala
@@ -1,9 +1,9 @@
 package com.gu.productmove.endpoint.move
 
 import com.gu.productmove.framework.InlineSchema.inlineSchema
-import com.gu.productmove.endpoint.available.{Currency, MoveToProduct, TimePeriod}
+import com.gu.productmove.endpoint.available.{Currency, MoveToProduct}
 import sttp.tapir.Schema
-import sttp.tapir.Schema.annotations.{description, encodedExample, encodedName}
+import sttp.tapir.Schema.annotations.{description, encodedName}
 import sttp.tapir.generic.{Configuration, Derived}
 import zio.json.{DeriveJsonDecoder, DeriveJsonEncoder, JsonDecoder, JsonEncoder}
 import sttp.tapir.Schema
--- a/handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/Subscribe.scala
+++ b/handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/Subscribe.scala
@@ -2,7 +2,7 @@ package com.gu.productmove.zuora
 
 import com.gu.productmove.AwsS3
 import com.gu.productmove.GuStageLive.Stage
-import com.gu.productmove.zuora.Subscribe.{CaseId, ChargeOverrides, CreateSubscriptionResponse, ProductRatePlanChargeId, ProductRatePlanId, SubscribeRequest, SubscribeToRatePlans, ZuoraAccountId}
+import com.gu.productmove.zuora.{CaseId, ChargeOverrides, CreateSubscriptionResponse, ProductRatePlanChargeId, ProductRatePlanId, SubscribeToRatePlans, ZuoraAccountId}
 import com.gu.productmove.zuora.GetSubscription.GetSubscriptionResponse
 import com.gu.productmove.zuora.rest.ZuoraGet
 import sttp.capabilities.zio.ZioStreams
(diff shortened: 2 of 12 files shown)


jonflynng:
[Add or reuse Zuora subscribe call](https://trello.com/c/tPFFBwCm/3205-add-or-reuse-zuora-subscribe-call)

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/move/ProductMoveEndpointTypes.scala
```diff
@@ -1,9 +1,9 @@
 package com.gu.productmove.endpoint.move
 
 import com.gu.productmove.framework.InlineSchema.inlineSchema
-import com.gu.productmove.endpoint.available.{Currency, MoveToProduct}
+import com.gu.productmove.endpoint.available.{Currency, MoveToProduct, TimePeriod}
 import sttp.tapir.Schema
-import sttp.tapir.Schema.annotations.{description, encodedName}
+import sttp.tapir.Schema.annotations.{description, encodedExample, encodedName}
```
johnduffell, line 6: were these needed to add?

handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/Subscribe.scala
```diff
@@ -0,0 +1,88 @@
...
+
+case class CreateSubscriptionResponse(subscriptionId: String)
+given JsonDecoder[CreateSubscriptionResponse] = DeriveJsonDecoder.gen[CreateSubscriptionResponse]
+
+case class ChargeOverrides(
+                            price: Option[Double],
+                            productRatePlanChargeId: String,
+                            triggerDate: Option[LocalDate],
+                            triggerEvent: Option[String]
+                          )
```
johnduffell, line 49: intellij does this sometimes for some reason and I forget the setting to fix it, but I prefer to stick to 2 space indent.  This means if the `ChargeOverrides` class is renamed we don't have to reformat the other lines
```suggestion
case class ChargeOverrides(
  price: Option[Double],
  productRatePlanChargeId: String,
  triggerDate: Option[LocalDate],
  triggerEvent: Option[String]
)
```

handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/Subscribe.scala
```diff
@@ -0,0 +1,88 @@
...
+                             contractEffectiveDate: LocalDate,
+                             customerAcceptanceDate: LocalDate,
+                             termType: String = "TERMED",
+                             renewalTerm: Int = 12,
+                             initialTerm: Int = 12,
+                             subscribeToRatePlans: List[SubscribeToRatePlans],
+                             AcquisitionCase__c: String,
+                             AcquisitionSource__c: String,
+                             CreatedByCSR__c: String
+                           )
```
johnduffell, line 71: same as above
```suggestion
case class SubscribeRequest(
  accountKey: String,
  autoRenew: Boolean = true,
  contractEffectiveDate: LocalDate,
  customerAcceptanceDate: LocalDate,
  termType: String = "TERMED",
  renewalTerm: Int = 12,
  initialTerm: Int = 12,
  subscribeToRatePlans: List[SubscribeToRatePlans],
  AcquisitionCase__c: String,
  AcquisitionSource__c: String,
  CreatedByCSR__c: String
)
```

handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/Subscribe.scala
```diff
@@ -0,0 +1,88 @@
...
+      date <- Clock.currentDateTime.map(_.toLocalDate)
+    } yield SubscribeRequest(
+      accountKey = zuoraAccountId,
+      contractEffectiveDate = date,
+      // customerAcceptanceDate = date.plusDays(14),    if there is a free trial, we simply add the 14 days onto the customerAcceptanceDate, no extra rateplans needed. To implement later on.
+      customerAcceptanceDate = date,
+      subscribeToRatePlans = List(SubscribeToRatePlans(productRatePlanId = targetProductId)),
+      AcquisitionCase__c = "case",
+      AcquisitionSource__c = "product-movement",
+      CreatedByCSR__c = "na"
```
johnduffell, line 86: not necessarily for this pr but these will need to be updated in future.
case and CSR would be blank at the moment.  And the source in future we need a matrix of movement vs standard and csr vs self service.  The plan is packing both pieces of info into one field.

handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/Subscribe.scala
```diff
@@ -0,0 +1,88 @@
...
+object SubscribeRequest {
+  given JsonEncoder[SubscribeRequest] = DeriveJsonEncoder.gen[SubscribeRequest]
+
+  def apply(zuoraAccountId: String, targetProductId: String): ZIO[Any, Nothing, SubscribeRequest] =
+    for {
+      date <- Clock.currentDateTime.map(_.toLocalDate)
+    } yield SubscribeRequest(
+      accountKey = zuoraAccountId,
+      contractEffectiveDate = date,
+      // customerAcceptanceDate = date.plusDays(14),    if there is a free trial, we simply add the 14 days onto the customerAcceptanceDate, no extra rateplans needed. To implement later on.
```
johnduffell, line 81: I understand - currently implemented start of service and first payment "today", but this will change as needed.

handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/Subscribe.scala
```diff
@@ -0,0 +1,88 @@
...
+                             initialTerm: Int = 12,
+                             subscribeToRatePlans: List[SubscribeToRatePlans],
+                             AcquisitionCase__c: String,
+                             AcquisitionSource__c: String,
+                             CreatedByCSR__c: String
+                           )
+object SubscribeRequest {
+  given JsonEncoder[SubscribeRequest] = DeriveJsonEncoder.gen[SubscribeRequest]
+
+  def apply(zuoraAccountId: String, targetProductId: String): ZIO[Any, Nothing, SubscribeRequest] =
```
johnduffell, line 75: I'd actually be tempted not to use `apply` here as it's not sucha simple constructor.  How about withTodaysDate?
```suggestion
  def withTodaysDate(zuoraAccountId: String, targetProductId: String): ZIO[Any, Nothing, SubscribeRequest] =
```
Another option to explore would be to pass in the dates as a normal parameter which would make this normal scala rather than ZIO.

handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/rest/ZuoraGet.scala
```diff
@@ -15,6 +15,13 @@ private class ZuoraGetLive(zuoraClient: ZuoraClient) extends ZuoraGet :
...
 
+  override def post[Request: JsonEncoder, Response: JsonDecoder](relativeUrl: Uri, input: Request): IO[String, Response] =
+    for {
+      response <- zuoraClient.send(basicRequest.contentType("application/json").body(input.toJson).post(relativeUrl))
+      parsedBody <- ZIO.fromEither(ZuoraRestBody.parseIfSuccessful[Response](response))
+    } yield parsedBody
+
 trait ZuoraGet:
   def get[T: JsonDecoder](relativeUrl: Uri): IO[String, T]
+  def post[Request: JsonEncoder, Response: JsonDecoder](relativeUrl: Uri, input: Request): IO[String, Response]
```
johnduffell, line 26: I see that get and post operations are in the same interface for now.  I'm happy to leave it like that for now as they are bread and butter operations - let's see how it goes.
If we need something more obscure like PATCH or DELETE I would be more in favour of separating it.

handlers/product-move-api/src/test/scala/com/gu/productmove/HandlerSpec.scala
```diff
@@ -3,57 +3,74 @@ package com.gu.productmove
...
           )
         )
-        for {
-          output <- ProductMoveEndpoint.runWithEnvironment(testInput)
-          requests <- TestGetSubscription.requests
+
+        (for {
+          output <- ProductMoveEndpoint.productMove(expectedSubNameInput, testPostData)
+          testRequests <- TestGetSubscription.requests
+          subscribeRequests <- MockSubscribe.requests
```
johnduffell, line 64: we should make sure the terminologify is consistent between Test and Mock (and Stub etc).  Not sure which is the "correct" one but if you know then rename the one that's incorrect.

handlers/product-move-api/src/test/scala/com/gu/productmove/HandlerSpec.scala
```diff
@@ -3,57 +3,74 @@ package com.gu.productmove
...
-        val testInput = ExpectedInput("false")
+      test("productMove endpoint") {
+        val expectedSubNameInput = "A-S00339056"
+        val testPostData = ExpectedInput("targetProductId")
+
+        val getSubscriptionResponse = GetSubscriptionResponse("subscriptionName", "zuoraAccountId")
+        val createSubscriptionResponse = CreateSubscriptionResponse("newSubscriptionName")
+
+        val GetSubscriptionStubs = Map(expectedSubNameInput -> getSubscriptionResponse)
+        val SubscribeStubs = Map(("zuoraAccountId", "targetProductId") -> createSubscriptionResponse)
```
johnduffell, line 23:
```suggestion
        val getSubscriptionStubs = Map(expectedSubNameInput -> getSubscriptionResponse)
        val subscribeStubs = Map(("zuoraAccountId", "targetProductId") -> createSubscriptionResponse)
```

johnduffell:
looks good, a few comments, also we will need to add some EffectsTest tests at some point, so we can make sure it's working against dev zuora even if we start updating http libraries or API version numbers in future.

Also if you could add a description if possible to the PR, at least to give a brief "what" and "why" and any links to other PRs (or documents if necessary).  This will help future people who don't have access to trello cards and the context is forgotten.

handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/Subscribe.scala
jonflynng: Ok, I'll have a google around this so it doesn't indent it lots on each re-format.

handlers/product-move-api/src/test/scala/com/gu/productmove/HandlerSpec.scala
jonflynng: This is the clearest definition I've found:

https://semaphoreci.com/community/tutorials/stubbing-and-mocking-with-mockito-2-and-junit
"A stub is a fake class that comes with preprogrammed return values. It’s injected into the class under test to give you absolute control over what’s being tested as input. A typical stub is a database connection that allows you to mimic any scenario without having a real database.

A mock is a fake class that can be examined after the test is finished for its interactions with the class under test. For example, you can ask it whether a method was called or how many times it was called. Typical mocks are classes with side effects that need to be examined, e.g. a class that sends emails or sends data to another external service."

So I'm assuming we should be calling them 'mocks'. Zio also refers to it as 'mocking' services. Cypress calls them spies however, so I don't know, but mocks sound fine to me.

handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/Subscribe.scala
jonflynng: I'll create a trello card

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/move/ProductMoveEndpointTypes.scala
jonflynng: hmmm, no, I didn't add anything to that file, so will remove them.

jonflynng:
I've added a trello card for adding integration tests and possibly E2E tests

handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/Subscribe.scala
jonflynng: I made it a ZIO so I could add an easy unit test for it in `SubscribeSpec.scala`, otherwise if we added it at a higher-level in the `create` function where it's being called, we'd have to mock the zuoraGet layer I think.

johnduffell: we can discuss if you like, I don't have strong feelings if it's simpler to keep it as a ZIO.  I do have stronger feelings on the name, nowadays I think apply should generally only be used for fairly trivial constructors, but we can also discuss that if you like.

jonflynng: Yep, I agree on the naming, have updated it.

johnduffell (approved):
great work, this has been interesting and I have learnt a lot more about ZIO, let's start getting these merged into main 👍

PR #1630, opened 2022-07-18
Add cancellation call to product-move-api

jonflynng:
This is part of the product-movement-api which takes a user with a recurring contribution and switches them to a digital subscription.

 This adds the API call to Zuora needed to cancel the user's current subscription, this is made before making the new subscribe call. The following Zuora endpoint is used: `/v1/subscriptions/A-S12345678/cancel`

https://www.zuora.com/developer/api-reference/#operation/PUT_CancelSubscription

### DIFF
--- a/handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/move/ProductMoveEndpoint.scala
+++ b/handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/move/ProductMoveEndpoint.scala
@@ -5,7 +5,7 @@ import com.gu.productmove.endpoint.move.ProductMoveEndpointTypes.*
 import com.gu.productmove.framework.ZIOApiGatewayRequestHandler.TIO
 import com.gu.productmove.framework.{LambdaEndpoint, ZIOApiGatewayRequestHandler}
 import com.gu.productmove.zuora.rest.{ZuoraClientLive, ZuoraGet, ZuoraGetLive}
-import com.gu.productmove.zuora.{GetSubscription, GetSubscriptionLive, Subscribe, SubscribeLive}
+import com.gu.productmove.zuora.{Cancellation, GetSubscription, GetSubscriptionLive, Subscribe, SubscribeLive}
 import com.gu.productmove.{AwsCredentialsLive, AwsS3Live, GuStageLive, SttpClientLive}
 import sttp.tapir.*
 import sttp.tapir.EndpointIO.Example
@@ -68,6 +68,7 @@ object ProductMoveEndpoint {
     for {
       _ <- ZIO.log("PostData: " + postData.toString)
       subscription <- GetSubscription.get(subscriptionName)
+      _ <- Cancellation.cancel(subscriptionName, subscription.chargedThroughDate)
       newSubscriptionId <- Subscribe.create(subscription.accountId, postData.targetProductId)
       _ <- ZIO.log("Sub: " + newSubscriptionId.toString)
     } yield Success(newSubscriptionId.subscriptionId, MoveToProduct(
--- a/handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/move/ProductMoveEndpoint.scala
+++ b/handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/move/ProductMoveEndpoint.scala
@@ -5,7 +5,7 @@ import com.gu.productmove.endpoint.move.ProductMoveEndpointTypes.*
 import com.gu.productmove.framework.ZIOApiGatewayRequestHandler.TIO
 import com.gu.productmove.framework.{LambdaEndpoint, ZIOApiGatewayRequestHandler}
 import com.gu.productmove.zuora.rest.{ZuoraClientLive, ZuoraGet, ZuoraGetLive}
-import com.gu.productmove.zuora.{Cancellation, GetSubscription, GetSubscriptionLive, Subscribe, SubscribeLive}
+import com.gu.productmove.zuora.{Cancellation, GetSubscription, GetSubscriptionLive, Subscribe, SubscribeLive, CancellationLive}
 import com.gu.productmove.{AwsCredentialsLive, AwsS3Live, GuStageLive, SttpClientLive}
 import sttp.tapir.*
 import sttp.tapir.EndpointIO.Example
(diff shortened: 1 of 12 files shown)


jonflynng:
[Add or reuse cancellation Zuora call](https://trello.com/c/z7aqPjNr/3204-add-or-reuse-cancellation-zuora-call)

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/move/ProductMoveEndpoint.scala
```diff
@@ -49,46 +49,52 @@ object ProductMoveEndpoint {
...
-        id = "123",
-        name = "Digital Pack",
-        billing = Billing(
-          amount = Some(1199),
-          percentage = None,
-          currency = Currency.GBP,
-          frequency = Some(TimePeriod(TimeUnit.month, 1)),
+      subscription <- GetSubscription.get(subscriptionName)
+
+      chargedThroughDate <- ZIO.fromOption(subscription.ratePlans.head.ratePlanCharges.head.chargedThroughDate).orElseFail(s"chargedThroughDate is null for subscription $subscriptionName.")
```
jonflynng, line 73: the chargedThroughDate comes through as null if the person is in a free trial (for a digital subscription) and has not made payment yet. Although we don't need to consider this in the MVP just thought I'd put it in as it was quick to do.

handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/Cancellation.scala
```diff
@@ -0,0 +1,47 @@
...
+import sttp.client3.*
+import sttp.client3.httpclient.zio.{HttpClientZioBackend, SttpClient, send}
+import sttp.client3.ziojson.*
+import sttp.model.Uri
+import zio.json.*
+import zio.{Clock, IO, RIO, Task, UIO, URLayer, ZIO, ZLayer}
+
+import java.time.LocalDate
+
+trait Cancellation:
```
johnduffell, line 19: I'm wondering if we should call these something like ZuoraCancel etc just to add some context to the class name.  I realise they are in package zuora.  What do you think?

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/move/ProductMoveEndpoint.scala
johnduffell: `.head` will throw an exception if there's no head, and I'm not sure how fromOption or flatMap would deal with that situation.
If we use `headOption` at each stage and flatMap it into the for comprehension then we can be sure that the error will be handled properly with an explicit message (as even if the exception is visible, it won't be clear which "head" failed from just the stack trace)

As a note, there will only be one (main) rate plan with one charge for contribution, but we need to beware for other products in future.  For example discounts would appear as a separate rate plan with a suitable discount charge.

handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/Cancellation.scala
```diff
@@ -0,0 +1,47 @@
...
+  def cancel(subscriptionNumber: String, chargedThroughDate: LocalDate): ZIO[Any, String, CancellationResponse]
```
johnduffell, line 20: let's change this to be in the context of the cancel call itself
```suggestion
  def cancel(subscriptionNumber: String, cancellationEffectiveDate: LocalDate): ZIO[Any, String, CancellationResponse]
```

handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/GetSubscription.scala
```diff
@@ -25,9 +27,30 @@ trait GetSubscription :
 
 object GetSubscription {
 
-  case class GetSubscriptionResponse(id: String, accountId: String)
+  case class GetSubscriptionResponse(id: String, accountId: String, ratePlans: List[RatePlan])
```
johnduffell, line 30: probably split this.  Also use 2 space indent on the other case classes in this PR.
```suggestion
  case class GetSubscriptionResponse(
    id: String,
    accountId: String,
    ratePlans: List[RatePlan]
  )
```

johnduffell:
looks good, thanks and sorry for the late review.  Just a few comments above, and also the same as the other PR about the PR description and the intendation.
And also we will add the EffectsTest for this at some point to make sure it's working long term.  Perhaps we can pair on that some time, as I'm not totally sure how it's going to pan out.  What would be fantastic would be to automate the EffectsTests similar to the IT tests on support-workers, but that might be a health job for the future.

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/move/ProductMoveEndpoint.scala
jonflynng: for `.head`, will there for always be at least one rateplan if the user can get to the cancellation page? Or are we just making it a bit more robust and future proof?

 As for multiple rateplans, yep, we need to take that into account for future products, will make a note of it.

handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/Cancellation.scala
jonflynng: If we have them as separate ZIO services then yes, I think that more specific naming is better, I'll change it.

handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/Cancellation.scala
jonflynng: good idea, thanks

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/move/ProductMoveEndpoint.scala
johnduffell: as discussed in the pairing earlier, it would be good to "fail fast" if there are the wrong number of rate plans, just in case this code is reused elsewhere.  Otherwise it might silently pick the wrong one.

johnduffell (approved):
nice work getting this PR through to completion👍 separating them has made it easier to review, although I accept it's been more difficult to orgnise the work

PR #1648, opened 2022-07-28
Don't run gifting lambda on 1st August 2022

paulbrown1982:
## What does this change?
Updating the gifting lambda to run exactly at 3am and 3pm UTC, and for now, exclude 1st day of month due to Finance project change happening 12am-2pm on 1 August.

## How to test
Release today and see that it runs at 3pm just fine.

Lambda does not deploy. Rollback PR will be merged and deployed. On 2nd August the 2-31 bit of the cron will be replaced with a *.

### DIFF
--- a/handlers/revenue-recogniser-job/cfn.yaml
+++ b/handlers/revenue-recogniser-job/cfn.yaml
@@ -15,7 +15,7 @@ Mappings:
   Stages:
     Schedule:
       CODE: 'rate(365 days)'
-      PROD: 'rate(12 hours)'
+      PROD: 'cron(0 3,15 2-31 * *)'
   Constants:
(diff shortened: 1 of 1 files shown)


rupertbates (approved):
👍

johnduffell:
why did we want it specifically to run at 3am and 3pm?

paulbrown1982:
@johnduffell - Previously the lambda was deployed at 3pm in the afternoon and so executed at 3pm and 3am. In order to exclude the Monday, I had to switch from a `rate` to a `cron`, and that required me to pick the time of day. I wanted to change as little of the ecosystem as possible, so I chose times which were as close as possible to the historical schedule.

PR #1660, opened 2022-08-09
Available Product Moves Endpoint

jonflynng:
This is the endpoint that returns which products the user may switch to. For this MVP, we are currently only allowing switching from a recurring contribution to a monthly digital subscription.


build.sbt
```diff
@@ -40,7 +40,7 @@ val scala3Settings = Seq(
     "-encoding", "UTF-8",
     "-feature",
     "-unchecked",
-    "-Xmax-inlines", "256"
+    "-Xmax-inlines", "256", "-Yretain-trees"
```
jonflynng, line 43: This is to fix a compiler error which John posted on Zio's github here: https://github.com/zio/zio/issues/7072

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/available/AvailableProductMovesEndpoint.scala
```diff
@@ -44,44 +53,121 @@ object AvailableProductMovesEndpoint {
...
+  private val freeTrialDays = 14
+
+
+  def getAvailableSwitchRatePlans(zuoraProductCatalogue: ZuoraProductCatalogue, ratePlanNames: List[String]): List[ZuoraProductRatePlan] =
+    val productRatePlans = for {
+      product <- zuoraProductCatalogue.products.filter(ratePlanNames contains _.name)
+      productRatePlan <- product.productRatePlans.toList
+    } yield productRatePlan
+
+    productRatePlans.find(_.name == "Digital Pack Monthly").toList
```
jonflynng, line 79: Do you have a preference between hardcoding the name of the produtRatePlan or the ID? I was thinking maybe the name would be better as the ID's differ between environments. We do it this way to select rateplans in the price-rise engine and haven't had any troubles (yet).

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/available/AvailableProductMovesEndpoint.scala
```diff
@@ -44,44 +53,121 @@ object AvailableProductMovesEndpoint {
...
-        ))
-      )
-    ))
+
+      // Kick off catalogue fetch in parallel
+      zuoraProductCatalogueFetch <- GetCatalogue.get.fork
+      subscription <- GetSubscription.get(subscriptionName)
+
+      // Next payment date
+      chargedThroughDate <- ZIO.fromOption(subscription.ratePlans.head.ratePlanCharges.head.chargedThroughDate).orElseFail(s"chargedThroughDate is null for subscription $subscriptionName.")
```
jonflynng, line 106: Ah, this shouldn't fail if the chargedThroughDate is None, it should succeed with an empty products list, I'll amend this.

build.sbt
johnduffell, line 43:
```suggestion
    "-Xmax-inlines", "256",
    "-Yretain-trees",
```

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/available/AvailableProductMovesEndpoint.scala
johnduffell: yeah it's a good point, traditionally we do the ID and have a Map to connect the stage with the id, but using the name is super clear in the code.
The ID is more definite, as we have had issues in the past where there have been products added which looked too similar to existing ones and caused the membership and subscriptions sites to go down.
I think in reality it would be ok in this case, but I noticed in all the PRs Rupert did he added the new IDs to the maps.  Also we do have the IDs in the New-product-api code that I was extracting.

So for consistency I would recommend sticking with the IDs if possible.

Does that help? We can always get more opinions if you feel strongly?

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/available/AvailableProductMovesEndpoint.scala
```diff
@@ -44,44 +53,121 @@ object AvailableProductMovesEndpoint {
...
+  def eligibilityCheck(accountNotEligible: Boolean, message: String): ZIO[Any, Unit, Unit] = {
+    if (accountNotEligible) {
+      for {
+        _ <- ZIO.log(message)
+        _ <- ZIO.fail(())
+      } yield ()
+    } else {
+      ZIO.succeed(())
+    }
+  }
```
johnduffell, line 92: Ignoring the possible scala 3 indentation style for now, but even in scala 2 style we could get rid of most of the braces if you like?
```suggestion
  def eligibilityCheck(accountNotEligible: Boolean, message: String): ZIO[Any, Unit, Unit] = 
    if (accountNotEligible) 
      for {
        _ <- ZIO.log(message)
        _ <- ZIO.fail(())
      } yield ()
    else
      ZIO.succeed(())
```

Also, we have a "Not" in the name of the variable.  Would it be clearer to flip the logic like to this?
```scala
  def eligibilityCheck(accountEligible: Boolean, message: String): ZIO[Any, Unit, Unit] = 
    if (accountEligible) 
      ZIO.succeed(())
    else
      for {
        _ <- ZIO.log(message)
        _ <- ZIO.fail(())
      } yield ()
```

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/available/AvailableProductMovesEndpoint.scala
```diff
@@ -44,44 +53,121 @@ object AvailableProductMovesEndpoint {
...
+
+  private[productmove] def runWithEnvironment(subscriptionName: String): ZIO[GetSubscription with GetCatalogue with GetAccount with Stage, String, OutputBody] = {
     for {
+      stage <- ZIO.service[Stage]
+      monthlyContributionRatePlanId = if (stage == Stage.DEV) "2c92c0f85a6b134e015a7fcd9f0c7855" else "2c92a0fc5aacfadd015ad24db4ff5e97"
```
johnduffell, line 97: I think this is the Gu Stage rather than the Zuora stage?  We never totally got the mappings between GU stages and Zuora stages clear in lambdas because it depends if they have a potential test user, but I think in this case Manage should let us know if it's a test user?
Might be one for another PR - it will come out when we try to test it with manage.

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/available/AvailableProductMovesEndpoint.scala
johnduffell: I think we also had a plan to make this fail if the length of `ratePlans` or `ratePlanCharges` was not exactly 1.  So that if this code is reused for other products, it will be clear that our current assumption is no longer the case.

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/available/AvailableProductMovesEndpoint.scala
```diff
@@ -44,44 +53,121 @@ object AvailableProductMovesEndpoint {
...
       GuStageLive.layer,
     )
 
-  private[productmove] def runWithEnvironment(subscriptionName: String): ZIO[GetSubscription, String, OutputBody] =
+  private val freeTrialDays = 14
+
+
+  def getAvailableSwitchRatePlans(zuoraProductCatalogue: ZuoraProductCatalogue, ratePlanNames: List[String]): List[ZuoraProductRatePlan] =
+    val productRatePlans = for {
+      product <- zuoraProductCatalogue.products.filter(ratePlanNames contains _.name)
```
johnduffell, line 75: I'm not against bracket free full stop, but generally I include the brackets etc for clarity in this situation
```suggestion
      product <- zuoraProductCatalogue.products.filter(ratePlanNames.contains(_.name))
```

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/available/AvailableProductMovesEndpoint.scala
```diff
@@ -44,44 +53,121 @@ object AvailableProductMovesEndpoint {
...
+        Monthly contribution
+        Only have one sub in the account
+      */
+      isEligible <-
+        (for {
+          _ <- eligibilityCheck(account.subscriptions.length > 1, s"More than one subscription for account for subscription: $subscriptionName")
+          _ <- eligibilityCheck(subscription.ratePlans.head.productRatePlanId != monthlyContributionRatePlanId,s"Subscription: $subscriptionName is not a monthly contribution")
+          _ <- eligibilityCheck(subscription.ratePlans.head.ratePlanCharges.head.currency != "GBP", s"Subscription: $subscriptionName not in GBP")
+          _ <- eligibilityCheck(paymentMethod.NumConsecutiveFailures > 0, s"User is in payment failure with subscription: $subscriptionName")
+          _ <- eligibilityCheck(account.basicInfo.defaultPaymentMethod.creditCardExpirationMonth.isBefore(today), s"card expired for subscription: $subscriptionName")
```
johnduffell, line 127: we would have to check the month and year here.  I can't remember whether we discussed deserialising the month and year fields into a LocalDate or similar before doing the comparison.  From this code it looks like it is, although it's still called `...Month`?

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/available/AvailableProductMovesEndpoint.scala
```diff
@@ -44,44 +53,121 @@ object AvailableProductMovesEndpoint {
...
+        Currency is GBP (initially on day 1 only?)
+        Monthly contribution
+        Only have one sub in the account
+      */
+      isEligible <-
+        (for {
+          _ <- eligibilityCheck(account.subscriptions.length > 1, s"More than one subscription for account for subscription: $subscriptionName")
+          _ <- eligibilityCheck(subscription.ratePlans.head.productRatePlanId != monthlyContributionRatePlanId,s"Subscription: $subscriptionName is not a monthly contribution")
+          _ <- eligibilityCheck(subscription.ratePlans.head.ratePlanCharges.head.currency != "GBP", s"Subscription: $subscriptionName not in GBP")
+          _ <- eligibilityCheck(paymentMethod.NumConsecutiveFailures > 0, s"User is in payment failure with subscription: $subscriptionName")
```
johnduffell, line 126: can we also add one to check the account balance is zero?  This might prevent people who are between an invoice and payment run, but I think it would be good to avoid broken accounts.

We also need to check the account Auto Pay is switched on.

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/available/AvailableProductMovesEndpoint.scala
johnduffell: I think this check will make sure that we don't allow DD or paypal.  Should we make the "is CC" check explicit?  We can check the payment gateway or similar on the Account object.

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/available/AvailableProductMovesEndpoint.scala
```diff
@@ -44,44 +53,121 @@ object AvailableProductMovesEndpoint {
...
+          _ <- eligibilityCheck(account.basicInfo.defaultPaymentMethod.creditCardExpirationMonth.isBefore(today), s"card expired for subscription: $subscriptionName")
+        } yield ()).isSuccess
+
+      availableProductMoves <- if (isEligible) {
+        for {
+          zuoraProductCatalogue <- zuoraProductCatalogueFetch.join
+
+          productRatePlans = getAvailableSwitchRatePlans(zuoraProductCatalogue, List("Digital Pack"))
```
johnduffell, line 134: this ties in with your previous comment about using the names or the IDs.

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/available/AvailableProductMovesEndpoint.scala
```diff
@@ -44,44 +53,121 @@ object AvailableProductMovesEndpoint {
...
+        User is not in payment failure or has unpaid invoices
+        Currency is GBP (initially on day 1 only?)
+        Monthly contribution
+        Only have one sub in the account
+      */
+      isEligible <-
+        (for {
+          _ <- eligibilityCheck(account.subscriptions.length > 1, s"More than one subscription for account for subscription: $subscriptionName")
+          _ <- eligibilityCheck(subscription.ratePlans.head.productRatePlanId != monthlyContributionRatePlanId,s"Subscription: $subscriptionName is not a monthly contribution")
+          _ <- eligibilityCheck(subscription.ratePlans.head.ratePlanCharges.head.currency != "GBP", s"Subscription: $subscriptionName not in GBP")
```
jonflynng, line 125: @johnduffell  instead of using `subscription.ratePlans.head.ratePlanCharges.head.currency`, we could use the currency property that comes from the `getAccount` response:

Screenshot 2022-08-09 at 12 21 20

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/available/AvailableProductMovesEndpoint.scala
jonflynng: @johnduffell ah yes, I believe you added the correct deserialization, but just forgot to rename the variable. I've renamed it now.

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/available/AvailableProductMovesEndpoint.scala
```diff
@@ -44,44 +53,121 @@ object AvailableProductMovesEndpoint {
...
+          _ <- eligibilityCheck(paymentMethod.NumConsecutiveFailures > 0, s"User is in payment failure with subscription: $subscriptionName")
+          _ <- eligibilityCheck(account.basicInfo.defaultPaymentMethod.creditCardExpirationMonth.isBefore(today), s"card expired for subscription: $subscriptionName")
+        } yield ()).isSuccess
+
+      availableProductMoves <- if (isEligible) {
+        for {
+          zuoraProductCatalogue <- zuoraProductCatalogueFetch.join
+
+          productRatePlans = getAvailableSwitchRatePlans(zuoraProductCatalogue, List("Digital Pack"))
+          moveToProduct <- ZIO.collectAll { productRatePlans.map(x => MoveToProduct.buildResponseFromRatePlan(subscriptionName, x, chargedThroughDate)) }
```
johnduffell, line 135: this line is a bit hard to read, not sure if we can split it up a bit and use foreach (the type signature looks right for here https://javadoc.io/static/dev.zio/zio_2.12/2.0.0/zio/ZIO$.html#foreach[R,E,A,B,Collection[+Element]%3C:Iterable[Element]](in:Collection[A])(f:A=%3Ezio.ZIO[R,E,B])(implicitbf:zio.BuildFrom[Collection[A],B,Collection[B]],implicittrace:zio.Trace):zio.ZIO[R,E,Collection[B]] )
```suggestion
          moveToProduct <- ZIO.foreach(productRatePlans) { productRatePlan =>
            MoveToProduct.buildResponseFromRatePlan(subscriptionName, productRatePlan, chargedThroughDate)
          }
```

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/available/AvailableProductMovesEndpoint.scala
jonflynng: @johnduffell thanks, all good suggestions, I'll amend it :)

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/available/AvailableProductMovesEndpoint.scala
```diff
@@ -44,44 +53,121 @@ object AvailableProductMovesEndpoint {
...
-            startDate = Some("2022-09-21")
-          ),
-          duration = TimePeriod(TimeUnit.month, 3)
-        ))
-      )
-    ))
+
+      // Kick off catalogue fetch in parallel
+      zuoraProductCatalogueFetch <- GetCatalogue.get.fork
+      subscription <- GetSubscription.get(subscriptionName)
```
johnduffell, line 103: if this call fails due to No such subscription, we need to recover it and return something like `AvailableProductMovesEndpointTypes.NotFound(s"Subscription $subscriptionName not found")`
As it is, the whole for comprehension either returns failure (causing a 5xx) or Success.

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/available/AvailableProductMovesEndpoint.scala
```diff
@@ -44,44 +53,121 @@ object AvailableProductMovesEndpoint {
...
+      frequency = Some(TimePeriod(TimeUnit.month, 1)),
+      startDate = Some("2022-09-21")
+    ),
+    trial = Some(Trial(14)),
+    introOffer = Some(Offer(
+      Billing(
+        amount = None,
+        percentage = Some(50),
+        currency = Currency.GBP, //FIXME doesn't make sense for a percentage
+        frequency = None, //FIXME doesn't make sense for a percentage
```
johnduffell, line 165: I remember these fixmes, I think it's like this because we tried to reuse the Billing object for discounts.

I think it would make more sense to have a Billing object without a percentage, and move the fields we do need for discounts into the Offer object.
This woudl require a bit of updating in the manage code to expect the flatter format.

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/available/AvailableProductMovesEndpoint.scala
johnduffell: good idea re getting it from the account.  There's only one currency for the whole account and it's permanenely fixed, so it would be the simplest way to do it.

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/available/AvailableProductMovesEndpoint.scala
jonflynng: yep, it works

johnduffell: nice.  so collectAll = sequence and forech = traverse, depending on whether you speak ZIO or cats style.

jonflynng:
[Product-switch: available-product-moves endpoint](https://trello.com/c/fX7T5Bpq/3206-product-switch-available-product-moves-endpoint)

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/available/AvailableProductMovesEndpoint.scala
jonflynng: there is a metrics object from the getAccount call like this:

`
"metrics": {
        "balance": 0.000000000,
        "totalInvoiceBalance": 0.000000000,
        "creditBalance": 0.000000000,
        "contractedMrr": 9.000000000
    },
`

Would I use the balance property here?

handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/GetAccount.scala
```diff
@@ -0,0 +1,116 @@
...
+  def getPaymentMethod(paymentMethodId: String): ZIO[GetAccount, String, PaymentMethodResponse]
+
+object GetAccount {
+  case class PaymentMethodResponse(NumConsecutiveFailures: Int)
+
+  given JsonDecoder[PaymentMethodResponse] = DeriveJsonDecoder.gen[PaymentMethodResponse]
+
+  case class GetAccountResponse(
+    basicInfo: BasicInfo,
+    billingAndPayment: BillingAndPayment,
```
johnduffell, line 45: you are getting "billingAndPayment" containing currency, but I see it in "basicInfo".  I also don't see "metrics" actually the account balance is also in basicInfo for me.  Do you know why the difference?
image

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/available/AvailableProductMovesEndpoint.scala
johnduffell: see my comment below about the "billingAndPayment" and "metrics" property.  I'm not getting it back when I use postman to get the account summary.

handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/GetAccount.scala
johnduffell: I think you may be using /v1/accounts/<id> rather than /v1/accounts/<id>/summary when you are using postman.  The code actually uses the summary (this is another reminder that we need to add some effectsTests at some point!)

jonflynng: yep, you're right, will amend

handlers/product-move-api/src/test/scala/com/gu/productmove/HandlerSpec.scala
```diff
@@ -169,20 +186,108 @@ object HandlerSpec extends ZIOSpecDefault {
...
+          NumConsecutiveFailures = 3
+        )
+
+        val createSubscriptionResponse = CreateSubscriptionResponse("newSubscriptionName")
+
+        val getSubscriptionStubs = Map(expectedSubNameInput -> getSubscriptionResponse)
+        val getAccountStubs = Map("accountNumber" -> getAccountResponse)
+        val getPaymentMethodStubs = Map("paymentMethodId" -> getPaymentMethodResponse)
+
+        val expectedOutput = AvailableProductMovesEndpointTypes.Success(body = List())
```
jonflynng, line 272: @johnduffell   Do you know if there is a way we can test which eligibility check failed here? As now I'm just checking it succeeded with an empty response.

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/available/AvailableProductMovesEndpoint.scala
jonflynng: Is this fine now that we added the `handleError` extension method to catch and log any errors from our zio services? Or do we want to explicitly pick up and log any 404's?

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/available/AvailableProductMovesEndpoint.scala
jonflynng: We could pair on this on the 22nd when I'm back to extract the IDs from the new-product-api and implement it

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/available/AvailableProductMovesEndpoint.scala
jonflynng: so I ended up adding the checks for this in their own separate for comprehension (see my last couple of commits) like so:

`
firstEligibilityChecks <-
        (for {
          _ <- succeedIfEligible(subscription.ratePlans.length == 1,s"Subscription: $subscriptionName has more than one ratePlan")
          _ <- succeedIfEligible(subscription.ratePlans.head.productRatePlanId == monthlyContributionRatePlanId,s"Subscription: $subscriptionName is not a monthly contribution")
          _ <- succeedIfEligible(subscription.ratePlans.head.ratePlanCharges.length == 1,s"Subscription: $subscriptionName has more than one ratePlan charge for ratePlan ${subscription.ratePlans.head}")
        } yield ()).isSuccess

      _ <- if (firstEligibilityChecks) ZIO.succeed(()) else ZIO.fail(Success(List()))
      `

Most users won't get passed these checks so it increases performance a bit by not having to make the `GetAccount` and `GetPaymentMethod` calls.

 When I was first writing the endpoint I liked having all the steps for it in one for comprehension so any newcomers wouldn't have to navigate around the codebase for different functions, but it is getting very big now so we should refactor it sometime.

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/available/AvailableProductMovesEndpoint.scala
johnduffell: as mentioned this work is being done separately so we can pair on it later

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/available/AvailableProductMovesEndpoint.scala
johnduffell: good point about newcomers, I think it's great to be thinking about that while we're developing the code

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/available/AvailableProductMovesEndpoint.scala
```diff
@@ -44,44 +53,149 @@ object AvailableProductMovesEndpoint {
...
-            startDate = Some("2022-09-21")
-          ),
-          duration = TimePeriod(TimeUnit.month, 3)
-        ))
-      )
-    ))
+
+      // Kick off catalogue fetch in parallel
+      zuoraProductCatalogueFetch <- GetCatalogue.get.fork
+      subscription <- GetSubscription.get(subscriptionName).handleError("GetSubscription")
```
johnduffell, line 113: actually we need to check if it's either a not found, or some other error coming back from zuora.  If it's not found from zuora, we must return a ZIO.fail(NotFound), rather than a internal server error.

handlers/product-move-api/src/test/scala/com/gu/productmove/HandlerSpec.scala
johnduffell: no that's a good point, because we don't return it.  We would have to either pack a tuple into the return type, or have some kind of special service in the ZIO resource? to log which one failed.
My feeling is it's going to be too troublesome and if we test each one either way it would detect any bugs.
But I'd be open to providing a special header or body field with more information, as that might even be useful for debugging in PROD.
Another question, is if we log it with ZIO.log, does that appear somewhere we can check it in the tests?

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/available/AvailableProductMovesEndpoint.scala
johnduffell: we are planning to do this in a separate PR to include the IntegrationTest stuff also.

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/available/AvailableProductMovesEndpoint.scala
johnduffell: just wanted to check if this comment was missed or you disagree?
```suggestion
      product <- zuoraProductCatalogue.products.filter(ratePlanNames.contains(_.name))
```

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/available/AvailableProductMovesEndpoint.scala
```diff
@@ -44,44 +53,149 @@ object AvailableProductMovesEndpoint {
...
+      } yield ()
+
+  def getCreditCardExpirationDate(wireDefaultPaymentMethod: WireDefaultPaymentMethod): Either[Unit, LocalDate] =
     for {
+      creditCardExpirationMonth <- wireDefaultPaymentMethod.creditCardExpirationMonth.toRight(())
+      creditCardExpirationYear <- wireDefaultPaymentMethod.creditCardExpirationYear.toRight(())
+    } yield LocalDate.of(creditCardExpirationYear, creditCardExpirationMonth, 1)
+
+  extension[R, E, A] (zio: ZIO[R, E, A])
+    def handleError(message: String) = zio.catchAll {
```
johnduffell, line 99: let's change this to failureAs500
```suggestion
    def failureAs500(message: String) = zio.catchAll {
```

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/available/AvailableProductMovesEndpoint.scala
johnduffell: let's think about this in a separate PR.  We can look at what others e.g. cancelation-sf-cases do for stage mappings to get an idea of what makes sense.

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/available/AvailableProductMovesEndpoint.scala
```diff
@@ -44,44 +53,149 @@ object AvailableProductMovesEndpoint {
...
-          duration = TimePeriod(TimeUnit.month, 3)
-        ))
-      )
-    ))
+
+      // Kick off catalogue fetch in parallel
+      zuoraProductCatalogueFetch <- GetCatalogue.get.fork
+      subscription <- GetSubscription.get(subscriptionName).handleError("GetSubscription")
+
+      firstEligibilityChecks <-
```
johnduffell, line 115:
```suggestion
      subscriptionIsEligible <-
```

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/available/AvailableProductMovesEndpoint.scala
```diff
@@ -44,44 +53,149 @@ object AvailableProductMovesEndpoint {
...
+          _ <- succeedIfEligible(subscription.ratePlans.head.productRatePlanId == monthlyContributionRatePlanId, s"Subscription: $subscriptionName is not a monthly contribution")
+          _ <- succeedIfEligible(subscription.ratePlans.head.ratePlanCharges.length == 1, s"Subscription: $subscriptionName has more than one ratePlan charge for ratePlan ${subscription.ratePlans.head}")
+        } yield ()).isSuccess
+
+      _ <- if (firstEligibilityChecks) ZIO.succeed(()) else ZIO.fail(Success(List()))
+
+      // Next payment date
+      chargedThroughDate <- ZIO.fromOption(subscription.ratePlans.head.ratePlanCharges.head.chargedThroughDate).orElse {
+        ZIO.log(s"chargedThroughDate is null for subscription $subscriptionName.").flatMap(_ => ZIO.fail(Success(List())))
+      }
```
johnduffell, line 127: rather than using `.head` in many places, which might case an Element not found exception or pick the wrong plan if there are multiple ones, we could like this
```suggestion
      ratePlan <- asSingle(subscription.ratePlans, "ratePlan")
      charge <- asSingle(ratePlan.ratePlanCharges, "ratePlanCharge")
      chargedThroughDate <- ZIO.fromOption(charge.chargedThroughDate).orElseFail(s"chargedThroughDate is null")
      }
```
and define this function (copied from my uncommitted changes)
```scala
  def asSingle[A](list: List[A], message: String): IO[String, A] =
    list match {
      case singlePlan :: Nil => ZIO.succeed(singlePlan)
      case wrongNumber => ZIO.fail(s"subscription can't be cancelled as we didn't have a single $message: ${wrongNumber.length}: $wrongNumber")
    }
```
Then it will fail with a nice error if there is some assumption that doesn't hold.

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/available/AvailableProductMovesEndpoint.scala
```diff
@@ -44,44 +53,149 @@ object AvailableProductMovesEndpoint {
...
+
+      // Kick off catalogue fetch in parallel
+      zuoraProductCatalogueFetch <- GetCatalogue.get.fork
+      subscription <- GetSubscription.get(subscriptionName).handleError("GetSubscription")
+
+      firstEligibilityChecks <-
+        (for {
+          _ <- succeedIfEligible(subscription.ratePlans.length == 1, s"Subscription: $subscriptionName has more than one ratePlan")
+          _ <- succeedIfEligible(subscription.ratePlans.head.productRatePlanId == monthlyContributionRatePlanId, s"Subscription: $subscriptionName is not a monthly contribution")
+          _ <- succeedIfEligible(subscription.ratePlans.head.ratePlanCharges.length == 1, s"Subscription: $subscriptionName has more than one ratePlan charge for ratePlan ${subscription.ratePlans.head}")
```
johnduffell, line 119: we can use the getSingle function defined in the below comment here as well.

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/available/AvailableProductMovesEndpoint.scala
```diff
@@ -44,44 +53,149 @@ object AvailableProductMovesEndpoint {
...
+
+      /*
+      Only show the switch if:
+        credit card is not expired (according to zuora)
+        User is not in payment failure or has unpaid invoices
+        Currency is GBP (initially on day 1 only?)
+        Monthly contribution
+        Account balance is 0
+      */
+      secondEligibilityChecks <-
```
johnduffell, line 147:
```suggestion
      accountIsEligible <-
```

handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/GetCatalogue.scala
```diff
@@ -0,0 +1,81 @@
...
+  name: String,
+  status: String,
+  productRatePlanCharges: Set[ZuoraProductRatePlanCharge]
+)
+
+object ZuoraProductRatePlan {
+  given JsonDecoder[ZuoraProductRatePlan] = DeriveJsonDecoder.gen[ZuoraProductRatePlan]
+}
+
+case class ZuoraProductRatePlanCharge(id: String, billingPeriod: Option[String], pricing: Set[ZuoraPricing])
```
johnduffell, line 77: I'd rather have this as a List, as then you get a consistent iteration order. (or you can do some kinds of stable iteration sets if you prefer the Set semantics)
```suggestion
case class ZuoraProductRatePlanCharge(id: String, billingPeriod: Option[String], pricing: List[ZuoraPricing])
```

johnduffell: the same comment applies to the other places that Set appears in the data types.

handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/GetAccount.scala
```diff
@@ -0,0 +1,112 @@
...
+/*
+* Don't use discount percentage from product catalogue,
+* because it can be overridden so the default value is unreliable.
+*/
+// price can be null, this is only for percentage discounts, so default to 0 rather than option handling
+case class ZuoraPricing(currency: String, price: BigDecimal = 0.000000000)
+
+object ZuoraPricing {
+  given JsonDecoder[ZuoraPricing] = DeriveJsonDecoder.gen[ZuoraPricing]
+}
```
johnduffell, line 112: I realise you've handled it in the business logic later, but on balance I would prefer it to be a deserialisation concern to keep the business logic more focussed.  It does require writing custom decoder though which can be a pain.  What do you think of something like this?
```suggestion
case class ZuoraPricing(currency: String, priceMinorUnits: Int)

object ZuoraPricing {

  private case class ZuoraPricingWire(currency: String, price: BigDecimal = 0.000000000)
  
  given JsonDecoder[ZuoraPricing] = DeriveJsonDecoder.gen[ZuoraPricingWire].map {
    case ZuoraPricingWire(currency, price) => ZuoraPricing(currency, (price.toDouble * 100).toInt)
  }
}
```

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/available/AvailableProductMovesEndpointTypes.scala
```diff
@@ -67,7 +83,7 @@ case class Billing(
   @encodedExample(50)
   percentage: Option[Int], // Either - amount or percentage?
   @validate[Currency](Validator.enumeration(List(Currency.GBP)))
-  currency: Currency,
+  currency: Option[Currency],    // optional if the ratePlan is a percentage discount
```
johnduffell, line 86: I think I covered this in a comment elsewhere - I think we need to make Billing to only cover actual billing charges hence the currency and amount are required and percentage is not allowed., and the discounts section can contain the relevant fields for discounting.  Either that or make Billing a sealed trait in the scala model, if we meed to keep the same structure of the json.

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/available/AvailableProductMovesEndpointTypes.scala
```diff
@@ -103,13 +119,20 @@ enum TimeUnit:
 object TimeUnit:
 
   given JsonCodec[TimeUnit] = summon[JsonCodec[String]].transform[TimeUnit]({
-    case "month" => TimeUnit.month
-    case "year" => TimeUnit.year
+    case "Month" => TimeUnit.month
+    case "Annual" => TimeUnit.year
   }, _.toString)
```
johnduffell, line 124: the encoder and decoder must be symmetric, so if you are making this change you need to also change toString to make it do init-caps.  Practically speaking, otherwise if we use tapir to generate a client, it won't be able to deserialise its own responses.

But I suspect this change is actually because you are reusing our external API object for the Zuora deserialisation.  I think we should have separate types for each, as our API needn't be tied to zuora's API.

(If we do decide to reuse it (but with separate Decoder in the Zuora client), it needs to move to a shared class, as this is specifically "AvailableProductMove Endpoint Types".)

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/move/ProductMoveEndpoint.scala
```diff
@@ -90,7 +90,7 @@ object ProductMoveEndpoint {
         Billing(
           amount = None,
           percentage = Some(50),
-          currency = Currency.GBP,//FIXME doesn't make sense for a percentage
+          currency = Some(Currency.GBP),//FIXME doesn't make sense for a percentage
```
johnduffell, line 93:
```suggestion
          currency = None
```

handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/GetAccount.scala
```diff
@@ -0,0 +1,112 @@
...
+    id: String
+  )
+
+  case class BasicInfo(
+    defaultPaymentMethod: WireDefaultPaymentMethod,
+    balance: BigDecimal,
+    currency: String
+  )
+
+  /* The zuora API has slightly different responses for ratePlans in the catalogue and when querying a subscription, e.g.: productRatePlanId in the subscription response and id in the catalogue response */
```
johnduffell, line 60: yes good spot - the catalog is the "product..." side and the subscription is the non-"product..." side, it's got its own model that's similar but not identical.

handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/GetAccount.scala
```diff
@@ -0,0 +1,112 @@
...
+    subscriptions: List[AccountSubscription],
+  )
+
+  case class AccountSubscription(
+    id: String
+  )
+
+  case class BasicInfo(
+    defaultPaymentMethod: WireDefaultPaymentMethod,
+    balance: BigDecimal,
```
johnduffell, line 56: we can use the same decoder/int trick here as mentioned above comment

handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/GetAccount.scala
```diff
@@ -0,0 +1,112 @@
...
+  creditCardExpirationDate: LocalDate,
+)
+
+case class WireDefaultPaymentMethod(
+  id: String,
+  creditCardExpirationMonth: Option[Int],
+  creditCardExpirationYear: Option[Int],
+)
+
+given JsonDecoder[WireDefaultPaymentMethod] = DeriveJsonDecoder.gen[WireDefaultPaymentMethod]
```
johnduffell, line 101: we can keep all the wrangling in the decoder rather than the business logic something like this:
```suggestion
given JsonDecoder[DefaultPaymentMethod] = DeriveJsonDecoder.gen[WireDefaultPaymentMethod].map {
  case WireDefaultPaymentMethod(id, month, year) =>
    DefaultPaymentMethod(id, LocalDate.of(creditCardExpirationYear, creditCardExpirationMonth, 1))
}
```

johnduffell, line 101: we can keep all the wrangling in the decoder rather than the business logic something like this:
```suggestion
object DefaultPaymentMethod {
  private case class WireDefaultPaymentMethod(
    id: String,
    creditCardExpirationMonth: Option[Int],
    creditCardExpirationYear: Option[Int],
  )

  given JsonDecoder[DefaultPaymentMethod] = DeriveJsonDecoder.gen[WireDefaultPaymentMethod].map {
    case WireDefaultPaymentMethod(id, month, year) =>
      DefaultPaymentMethod(id, LocalDate.of(creditCardExpirationYear, creditCardExpirationMonth, 1))
  }
}
```

handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/GetCatalogue.scala
johnduffell: billingPeriod here could use the ZuoraBillingPeriod enum that is defined in this file  - I guess that was the intention as it's not used elsewhere

handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/GetCatalogue.scala
```diff
@@ -0,0 +1,81 @@
...
+object GetCatalogue {
+  def get: ZIO[GetCatalogue, String, ZuoraProductCatalogue] = ZIO.serviceWithZIO[GetCatalogue](_.get)
+}
+
+enum ZuoraBillingPeriod:
+  case Month, Annual, Quarter
+
+object ZuoraBillingPeriod {
+  given JsonDecoder[ZuoraBillingPeriod] = DeriveJsonDecoder.gen[ZuoraBillingPeriod]
+}
```
johnduffell, line 48: is this used?

handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/GetCatalogue.scala
```diff
@@ -0,0 +1,81 @@
...
+)
+
+object ZuoraProduct {
+  given JsonDecoder[ZuoraProduct] = DeriveJsonDecoder.gen[ZuoraProduct]
+}
+
+case class ZuoraProductRatePlan(
+  id: String,
+  name: String,
+  status: String,
```
johnduffell, line 69: is this used?  and if so, should it be an enum (Active,Expired,Cancelled)

handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/rest/ZuoraClient.scala
```diff
@@ -74,19 +74,22 @@ trait ZuoraClient {
 // zuora commonly returns status = 200 and success = false.
 // this usually causes a deserialisation error and often makes it hard to understand what went wrong
 // This detects that and stops straight away.
+// the `/v1/object/` endpoint which we are using to get the user's payment method does not have a success flag, therefore if the success property does not exist, we decode the JSON as normal.
```
johnduffell, line 77: yes, in fact it returns a totally different response if it's failed:
```
{
    "size": 0,
    "done": true,
    "records": []
}
```

handlers/product-move-api/src/test/scala/com/gu/productmove/zuora/MockCatalogue.scala
```diff
@@ -0,0 +1,16 @@
...
+import com.gu.productmove.zuora.GetSubscription
+import com.gu.productmove.zuora.GetSubscription.GetSubscriptionResponse
+import zio.json.*
+import zio.{IO, Task, ZIO}
+
+import scala.io.Source
+
+class MockCatalogue extends GetCatalogue {
+  override def get: IO[String, ZuoraProductCatalogue] = {
+    val json = Source.fromResource("Catalogue.json").mkString
```
johnduffell, line 12: this is a common way to do things, but I am in principle uncomfortable with checking in (large) generated files as it's unreadable.  In my ideal world the IntegrationTests would cover using the real catalog from S3, and we would test using a minimal catalog for the purpose of our test.

Practically speaking I think it's fine to leave it like that for now, but if we automate the Integration tests we can get rid of this one and put in a minimal one in its place.

handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/rest/ZuoraClient.scala
```diff
@@ -74,19 +74,22 @@ trait ZuoraClient {
...
-          body.fromJson[A]
-        else Left(body)
-    } yield parsedResponse
+    val zuoraSuccessResponse = successDecoder.decodeJson(body)
+
+    zuoraSuccessResponse match {
+      case Right(res) =>
+        res.success match {
+          case Some(successFlag) => if (successFlag) body.fromJson[A] else Left(body)
+          case None => body.fromJson[A]
```
johnduffell, line 90: I'm not sure about this - zuora has so many incompatible ways to indicate failure I think we should be explicit about what kind of failure we're expecting.
In this case we actually need to check `size: 0` (otherwise we will get a deserialisation error if it's not found)

Off the top of my head, some endoints have `Success: true` and some `success: true` and some `size: 0`, so we need some positive way of indicating in the GetPaymentMethod code which kind of failure to check.

We could either do that by
1. passing in an enum when we call ZuoraGet.get (and then on into this parseIfSuccessfil code which would then pattern match) 
1. having a separate ZuoraGetSized (or whatever) which would be required by calls that expected that.

I slightly prefer the enum option to avoid an explosion of layers, but happy to hear your thoughts on this.

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/available/AvailableProductMovesEndpoint.scala
jonflynng: I actually got a compiler error when changing it to the suggested version:
Screenshot 2022-08-24 at 14 53 23

handlers/product-move-api/src/test/scala/com/gu/productmove/zuora/MockCatalogue.scala
jonflynng: Hmm, to be honest I thought creating a minimal catalog would take a long time to write, and be hard to maintain and keep it updated. Actually looking back on this, is it possible we don't mock this layer at all, and instead just fetch the real catalogue on each test from S3?

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/available/AvailableProductMovesEndpoint.scala
jonflynng: So for this we're checking there is just one ratePlan and one ratePlanCharge further up in the `subscriptionIsEligible` checks

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/available/AvailableProductMovesEndpoint.scala
jonflynng: Ah ok, yep, I'll add it in here then

handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/GetAccount.scala
jonflynng: Where would we perform the Option handling for this?

handlers/product-move-api/src/test/scala/com/gu/productmove/zuora/MockCatalogue.scala
jonflynng: I agree the IntegrationTests should cover this, and we can just use the real catalog from S3 in the test. I'll leave this for now just to have the tests working, but will amend it in the future EffectsTest PR.

handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/GetCatalogue.scala
jonflynng: no, have removed it.

handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/GetAccount.scala
jonflynng: Good suggestion, have amended

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/available/AvailableProductMovesEndpointTypes.scala
jonflynng: ah, I think I must have left this in here as I did end up using separate types for each.

handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/GetAccount.scala
jonflynng: Is it worth adding code to convert this one to an integer? As Scala can compare Int and BigDecimal

handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/rest/ZuoraClient.scala
jonflynng: is it possible instead of having to pass an enum in, we can match on multiple option types? like so:

```
case class ZuoraSuccess(success: Option[Boolean], size: Option[Int])

def parseIfSuccessful[A: JsonDecoder](body: String): Either[String, A] = {
  val successDecoder: JsonDecoder[ZuoraSuccess] = DeriveJsonDecoder.gen[ZuoraSuccess]
  val zuoraSuccessResponse = successDecoder.decodeJson(body)

  zuoraSuccessResponse match {
    case Right(res) =>
      (res.success, res.size) match {
        case (Some(successFlag), None) => if (successFlag) body.fromJson[A] else return Left(body)
        case (None, Some(size)) => if (size == 0) return Left(body) else body.fromJson[A]
        case (None, None) => body.fromJson[A]
      }

      body.fromJson[A]
    case Left(errorMessage) => Left(s"Error decoding json from zuora response: $errorMessage")
  }
}
```

handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/GetAccount.scala
jonflynng: updated this during pairing session

handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/rest/ZuoraClient.scala
jonflynng: updated this during pairing session

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/available/AvailableProductMovesEndpoint.scala
jonflynng: I think we didn't create a separate `PercentageDiscount` object because some products have discounts based on amount and not percentage.

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/available/AvailableProductMovesEndpoint.scala
johnduffell: my mistake, I didn't realise scala bracketed the underscores that way, try this instead
```suggestion
      product <- zuoraProductCatalogue.products.filter(product => ratePlanNames.contains(product.name))
```

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/available/AvailableProductMovesEndpointTypes.scala
jonflynng: I think we didn't create a separate PercentageDiscount object because some products have discounts based on amount and not percentage.

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/available/AvailableProductMovesEndpoint.scala
```diff
@@ -44,44 +53,143 @@ object AvailableProductMovesEndpoint {
...
-            startDate = Some("2022-09-21")
-          ),
-          duration = TimePeriod(TimeUnit.month, 3)
-        ))
-      )
-    ))
+
+      // Kick off catalogue fetch in parallel
+      zuoraProductCatalogueFetch <- GetCatalogue.get.fork
+      subscription <- GetSubscription.get(subscriptionName).mapErrorTo500("GetSubscription")
```
johnduffell, line 107:
```suggestion
      subscription <- GetSubscription.get(subscriptionName).mapErrorTo500("GetSubscription") // TODO add code to return 404 rather than 500 if it's not found
```

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/available/AvailableProductMovesEndpoint.scala
johnduffell: indeed some are not percentages, but I still think that should be inside the `Offer` class to represent all the different kinds of discount rather than in the Billing class
Either `Either[Percentage, Amount]` or an `enum DiscountValue: case Percentage(percentageOff: Int), Amount(amount: Int)` or similar.

I can't remember whether we decided to do in another PR though?  Perhaps we did!

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/available/AvailableProductMovesEndpoint.scala
```diff
@@ -44,44 +53,143 @@ object AvailableProductMovesEndpoint {
...
+          _ <- succeedIfEligible(subscription.ratePlans.head.productRatePlanId == monthlyContributionRatePlanId, s"Subscription: $subscriptionName is not a monthly contribution")
+          _ <- succeedIfEligible(subscription.ratePlans.head.ratePlanCharges.length == 1, s"Subscription: $subscriptionName has more than one ratePlan charge for ratePlan ${subscription.ratePlans.head}")
+        } yield ()).isSuccess
+
+      _ <- if (subscriptionIsEligible) ZIO.succeed(()) else ZIO.fail(AvailableMoves(List()))
+
+      // Next payment date
+      chargedThroughDate <- ZIO.fromOption(subscription.ratePlans.head.ratePlanCharges.head.chargedThroughDate).orElse {
+        ZIO.log(s"chargedThroughDate is null for subscription $subscriptionName.").flatMap(_ => ZIO.fail(AvailableMoves(List())))
+      }
```
johnduffell, line 121: based on this comment: https://github.com/guardian/support-service-lambdas/pull/1660#discussion_r953931130
I think we should change to this code roughly:

```suggestion

        ratePlan <- getSingleOrNotEligible(subscription.ratePlans, s"Subscription: $subscriptionName , ratePlan")
        _ <- succeedIfEligible(ratePlan.productRatePlanId == monthlyContributionRatePlanId, s"Subscription: $subscriptionName is not a monthly contribution")
        charge <- getSingleOrNotEligible(ratePlan.ratePlanCharges, s"Subscription: $subscriptionName , ratePlan charge for ratePlan $ratePlan")

      // Next payment date
      chargedThroughDate <- ZIO.fromOption(charge.chargedThroughDate).orElse {
        for {
          _ <- ZIO.log(s"chargedThroughDate is null for subscription $subscriptionName.")
          _ <- ZIO.fail(())
        } yield AvailableMoves(List())
      }

```

and add a function like this

```scala
def getSingleOrNotEligible[A](list: List[A], message: String): IO[AvailableMoves, A] =
    list match {
      case single :: Nil => ZIO.succeed(single)
      case wrongNumber => 
      for {
        _ <- ZIO.log(s"subscription can't be cancelled as we didn't have a single $message: ${wrongNumber.length}: $wrongNumber")
        _ <- ZIO.fail(())
      } yield AvailableMoves(List())
    }
```

and change the above named function
```scala
  def succeedIfEligible(accountEligible: Boolean, message: String): ZIO[Any, AvailableMoves, Unit] =
    if (accountEligible)
      ZIO.succeed(())
    else
      for {
        _ <- ZIO.log(message)
        _ <- ZIO.fail(())
      } yield AvailableMoves(List())
```

and we could probably extract this into a function as well as it's repeated 3 times at least
```scala
      for {
        _ <- ZIO.log(message)
        _ <- ZIO.fail(())
      } yield AvailableMoves(List())
```

handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/GetAccount.scala
johnduffell: that would be fine if they are both in pence.  But it might be dangerous if we mix values in pence with values in pounds in our business logic.
Personally I prefer to convert them all on the way in, then by the time we use them in business logic, there are no concerns about mixed units.  But I could happily use BigDecimal as long as the pounds/pence conversion is consistent.

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/available/AvailableProductMovesEndpointTypes.scala
```diff
@@ -51,9 +51,9 @@ object MoveToProduct {
 
   def buildResponseFromRatePlan(subscriptionName: String, productRatePlan: ZuoraProductRatePlan, chargedThroughDate: LocalDate): IO[OutputBody, MoveToProduct] =
     for {
-      billingPeriod <- ZIO.fromOption(productRatePlan.productRatePlanCharges.head.billingPeriod).handleError(s"billingPeriod is null for subscription: $subscriptionName")
-      pricing <- ZIO.fromOption(productRatePlan.productRatePlanCharges.head.pricing.find(_.currency == "GBP")).handleError(s"currency not found on ratePlanCharge")
-      price  = pricing.price.toFloat * 100
+      billingPeriod <- ZIO.fromOption(productRatePlan.productRatePlanCharges.head.billingPeriod).orElse(ZIO.log(s"billingPeriod is null for subscription: $subscriptionName").flatMap(_ => ZIO.fail(AvailableMoves(List()))))
+      pricing <- ZIO.fromOption(productRatePlan.productRatePlanCharges.head.pricing.find(_.currency == "GBP")).orElse(ZIO.log(s"currency not found on ratePlanCharge").flatMap(_ => ZIO.fail(AvailableMoves(List()))))
```
johnduffell, line 55: this pattern of log and fail with empty list could be reused (possibly via an `extension` method), as it's appearing in quite a few places.

johnduffell: we can pass in the charge from the code in the comment I made above, then we don't have to use .head any more.

handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/GetAccount.scala
```diff
@@ -89,16 +90,26 @@ object GetAccount {
...
+    creditCardExpirationYear: Option[Int],
+  )
 
-given JsonDecoder[WireDefaultPaymentMethod] = DeriveJsonDecoder.gen[WireDefaultPaymentMethod]
+  given JsonDecoder[DefaultPaymentMethod] = DeriveJsonDecoder.gen[WireDefaultPaymentMethod].map {
+    case WireDefaultPaymentMethod(id, month, year) =>
+      val creditCardExpirationDate = for {
+        creditCardExpirationMonth <- month
+        creditCardExpirationYear <- year
+      } yield LocalDate.of(creditCardExpirationYear, creditCardExpirationMonth, 1)
```
johnduffell, line 108: this might throw an exception, but I think it's unlikely enough not to worry about handling it, unless you prefer to be watertight?

> DateTimeException - if the value of any field is out of range, or if the day-of-month is invalid for the month-year

handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/rest/ZuoraClient.scala
```diff
@@ -77,19 +76,27 @@ trait ZuoraClient {
 // the `/v1/object/` endpoint which we are using to get the user's payment method does not have a success flag, therefore if the success property does not exist, we decode the JSON as normal.
```
johnduffell, line 76: this comment can be removed I think, or we can move it to the `ZuoraSuccessCheck` enum and update it to suit.

handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/rest/ZuoraClient.scala
```diff
@@ -77,19 +76,27 @@ trait ZuoraClient {
...
+
+  case class ZuoraSuccessLowercase(success: Boolean)
+  case class ZuoraSuccessSize(size: Option[Int])
+
+  def parseIfSuccessful[A: JsonDecoder](body: String, zuoraSuccessCheck: ZuoraSuccessCheck): Either[String, A] = {
+    val isSuccessful: Either[String, Unit] = zuoraSuccessCheck match {
+      case ZuoraSuccessCheck.SuccessCheckSize =>
+        for {
+          zuoraResponse <- DeriveJsonDecoder.gen[ZuoraSuccessSize].decodeJson(body)
+          isSuccessful <- if (zuoraResponse.size.isEmpty) Right(()) else Left(s"size = 0, body: $body")
```
johnduffell, line 90:
```suggestion
          succeeded = zuoraResponse.size.isEmpty // size field only exists if it's not found.
          isSuccessful <- if (succeeded) Right(()) else Left(s"size = 0, body: $body")
```

handlers/product-move-api/src/test/scala/com/gu/productmove/HandlerSpec.scala
```diff
@@ -40,7 +40,7 @@ object HandlerSpec extends ZIOSpecDefault {
 
   private val getAccountResponse = GetAccountResponse(
     BasicInfo(
-      WireDefaultPaymentMethod("paymentMethodId", Some(12), Some(2030)),
+      DefaultPaymentMethod("paymentMethodId", Some(LocalDate.of(2030, 12, 1))),
       balance = 0.000000000,
```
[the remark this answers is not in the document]
jonflynng: yep, have amended this

handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/GetAccount.scala
jonflynng: Ok, have changed this to decode any bigDecimals to integers

handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/GetAccount.scala
jonflynng: I think we can leave it for now

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/available/AvailableProductMovesEndpoint.scala
jonflynng: Have amended this :)

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/available/AvailableProductMovesEndpointTypes.scala
jonflynng: The charge from the subscription isn't used here as this is looking through the product catalogue

jonflynng: Have added a new function `logAndReturnEmptyResponse` that combines the logging and returning the empty list of products

jonflynng: Since it's coming direct from the product catalogue, my initial thoughts would be to leave the `.head` unhandled, but what do you think?

johnduffell: got it, didn't spot the product bit. 

I agree, it's reasonable not to explicitly check as the catalog is pretty static and consistent 👍

handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/GetAccount.scala
```diff
@@ -116,7 +130,7 @@ object DefaultPaymentMethod {
 * because it can be overridden so the default value is unreliable.
 */
 // price can be null, this is only for percentage discounts, so default to 0 rather than option handling
-case class ZuoraPricing(currency: String, price: BigDecimal = 0.000000000)
+case class ZuoraPricing(currency: String, priceMinorUnits: Int)
```
johnduffell, line 133: good idea on the naming, very explicit.

handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/GetCatalogue.scala
```diff
@@ -0,0 +1,72 @@
...
+)
+
+object ZuoraProduct {
+  given JsonDecoder[ZuoraProduct] = DeriveJsonDecoder.gen[ZuoraProduct]
+}
+
+case class ZuoraProductRatePlan(
+  id: String,
+  name: String,
+  productRatePlanCharges: Set[ZuoraProductRatePlanCharge]
```
johnduffell, line 61: please can we change this to List just to give it a stable iteration order/consistency

johnduffell: also a couple of others in this file too

johnduffell (approved):
great work getting this over the line Jon.  I've spotted a couple of "Set" in my final review which should ideally be changed to List, but overall I think this is a solid PR and should be clear and maintainable in future.
Thanks for going through the whole process and let's get this shipped 👏

handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/GetCatalogue.scala
jonflynng: Yep, will change

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/available/AvailableProductMovesEndpoint.scala
```diff
@@ -44,44 +53,151 @@ object AvailableProductMovesEndpoint {
       .serverLogic[TIO] { subscriptionName => run(subscriptionName).tapEither(result => ZIO.log("result tapped: " + result)).map(Right.apply) }
   }
 
+  // sub to test on: "A-S00334930"
   private def run(subscriptionName: String): TIO[OutputBody] =
-    runWithEnvironment(subscriptionName).provide(
+    runWithEnvironment("A-S00334930").provide(
```
johnduffell, line 58: I was looking at the code today and wondered why we were not using `subscriptionName`, turned out we missed this accidental commit all that time ago!

PR #1670, opened 2022-08-15
Reduce batch size to avoid Lambda timeout

graham228221:
## What does this change?
On Thursdays this lambda was regularly hitting the 15 minute lambda time limit. This was due to an increase in records on Thursday morning due to Home Deliveries including three days of records rather than 1.

We are reducing the batch size down from 120 to 66 (200/3) to try and ensure the lambda runs within 15 minutes.

### DIFF
--- a/handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/Salesforce.scala
+++ b/handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/Salesforce.scala
@@ -13,15 +13,15 @@ import com.typesafe.scalalogging.LazyLogging
 object Salesforce extends LazyLogging {
 
   /*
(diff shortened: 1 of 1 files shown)


johnduffell (approved):
nice work!

further ideas would be to use batch endpoints for SF aand/or Zuora where available.
To write back to SF between each Zuora call.
The Lambda could monitor it's available remaining time and quit in advance.
We could add some parallelism to the zuora calls, without getting too close to our limit.

For now let's see how this goes.

graham228221:
[ALARM: "URGENT 9-5 - PROD: Failed to process holiday stops" in EU (Ireland)](https://trello.com/c/bynfjfVg/3415-alarm-urgent-9-5-prod-failed-to-process-holiday-stops-in-eu-ireland)

PR #1695, opened 2022-08-31
new-product-api - Orchestrate Acquisition onto Product Prop

david-pepper:

### DIFF
--- a/handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/AddSupporterPlus.scala
+++ b/handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/AddSupporterPlus.scala
@@ -1,0 +1,160 @@
+package com.gu.newproduct.api.addsubscription
+
+import com.gu.effects.sqs.AwsSQSSend
+import com.gu.effects.sqs.AwsSQSSend.QueueName
+import com.gu.i18n.Currency
+import com.gu.newproduct.api.EmailQueueNames
+import com.gu.newproduct.api.addsubscription.TypeConvert._
+import com.gu.newproduct.api.addsubscription.email.supporterplus.SupporterPlusEmailDataSerialiser._
+import com.gu.newproduct.api.addsubscription.email.{EtSqsSend, SendConfirmationEmail, SupporterPlusEmailData}
+import com.gu.newproduct.api.addsubscription.validation.Validation._
+import com.gu.newproduct.api.addsubscription.validation._
+import com.gu.newproduct.api.addsubscription.validation.supporterplus.SupporterPlusValidations.ValidatableFields
+import com.gu.newproduct.api.addsubscription.validation.supporterplus.{GetSupporterPlusCustomerData, SupporterPlusAccountValidation, SupporterPlusCustomerData, SupporterPlusValidations, AmountLimits}
+import com.gu.newproduct.api.addsubscription.zuora.CreateSubscription.{ChargeOverride, SubscriptionName, ZuoraCreateSubRequest, ZuoraCreateSubRequestRatePlan}
+import com.gu.newproduct.api.addsubscription.zuora.GetAccount.SfContactId
+import com.gu.newproduct.api.addsubscription.zuora.GetAccount.WireModel.ZuoraAccount
+import com.gu.newproduct.api.addsubscription.zuora.GetAccountSubscriptions.WireModel.ZuoraSubscriptionsResponse
+import com.gu.newproduct.api.addsubscription.zuora.GetContacts.Contacts
+import com.gu.newproduct.api.addsubscription.zuora.GetContacts.WireModel.GetContactsResponse
+import com.gu.newproduct.api.addsubscription.zuora.GetPaymentMethod.{DirectDebit, PaymentMethod, PaymentMethodWire}
+import com.gu.newproduct.api.addsubscription.zuora.{GetAccount, GetAccountSubscriptions, GetContacts, GetPaymentMethod}
+import com.gu.newproduct.api.productcatalog.PlanId.MonthlySupporterPlus
+import com.gu.newproduct.api.productcatalog.ZuoraIds.{PlanAndCharge, ProductRatePlanId, ZuoraIds}
+import com.gu.newproduct.api.productcatalog.{AmountMinorUnits, Catalog, Plan, PlanId}
+import com.gu.util.apigateway.ApiGatewayResponse.internalServerError
+import com.gu.util.reader.AsyncTypes.{AsyncApiGatewayOp, _}
+import com.gu.util.reader.Types.{ApiGatewayOp, OptionOps}
+import com.gu.util.resthttp.RestRequestMaker.Requests
+import com.gu.util.resthttp.Types.ClientFailableOp
+
+import java.time.LocalDate
+import scala.concurrent.Future
+
+object AddSupporterPlus {
+  def steps(
+   getPlan: PlanId => Plan,
+   getCurrentDate: () => LocalDate,
+   getPlanAndCharge: PlanId => Option[PlanAndCharge],
+   getCustomerData: ZuoraAccountId => ApiGatewayOp[SupporterPlusCustomerData],
+   supporterPlusValidations: (SupporterPlusValidations.ValidatableFields, PlanId, Currency) => ValidationResult[AmountMinorUnits],
+   createSubscription: ZuoraCreateSubRequest => ClientFailableOp[SubscriptionName],
+   sendConfirmationEmail: (Option[SfContactId], SupporterPlusEmailData) => AsyncApiGatewayOp[Unit]
+ )(request: AddSubscriptionRequest): AsyncApiGatewayOp[SubscriptionName] = {
+    for {
+      customerData <- getCustomerData(request.zuoraAccountId).toAsync
+      SupporterPlusCustomerData(account, paymentMethod, subscriptions, contacts) = customerData
+      validatableFields = ValidatableFields(request.amountMinorUnits, request.startDate)
+      amountMinorUnits <- supporterPlusValidations(validatableFields, request.planId, account.currency).toApiGatewayOp.toAsync
+      acceptanceDate = request.startDate.plusDays(paymentDelayFor(paymentMethod))
+      planAndCharge <- getPlanAndCharge(request.planId).toApiGatewayContinueProcessing(internalServerError(s"no Zuora id for ${request.planId}!")).toAsync
+      chargeOverride = ChargeOverride(Some(amountMinorUnits), planAndCharge.productRatePlanChargeId, None)
+      zuoraCreateSubRequest = ZuoraCreateSubRequest(
+        request = request,
+        acceptanceDate = acceptanceDate,
+        ratePlans = List(
+          ZuoraCreateSubRequestRatePlan(
+            productRatePlanId = planAndCharge.productRatePlanId,
+            maybeChargeOverride = Some(chargeOverride)
+          )
+        )
+      )
+      subscriptionName <- createSubscription(zuoraCreateSubRequest).toAsyncApiGatewayOp("create monthly supporter plus")
+      supporterPlusEmailData = toSupporterPlusEmailData(
+        request = request,
+        currency = account.currency,
+        paymentMethod = paymentMethod,
+        firstPaymentDate = acceptanceDate,
+        contacts = contacts,
+        amountMinorUnits = amountMinorUnits,
+        plan = getPlan(request.planId),
+        currentDate = getCurrentDate()
+      )
+      _ <- sendConfirmationEmail(account.sfContactId, supporterPlusEmailData).recoverAndLog("send supporter plus confirmation email")
+    } yield subscriptionName
+  }
+
+  def wireSteps(
+    catalog: Catalog,
+    zuoraIds: ZuoraIds,
+    zuoraClient: Requests,
+    isValidStartDateForPlan: (PlanId, LocalDate) => ValidationResult[Unit],
+    createSubscription: ZuoraCreateSubRequest => ClientFailableOp[SubscriptionName],
+    awsSQSSend: QueueName => AwsSQSSend.Payload => Future[Unit],
+    emailQueueNames: EmailQueueNames,
+    currentDate: () => LocalDate
+  ): AddSubscriptionRequest => AsyncApiGatewayOp[SubscriptionName] = {
+
+    val planAndChargeForSupporterPlusPlanId = zuoraIds.apiIdToPlanAndCharge.get _
+    val supporterPlusIds = List(zuoraIds.supporterPlusZuoraIds.monthly.productRatePlanId, zuoraIds.supporterPlusZuoraIds.annual.productRatePlanId)
+    val getCustomerData = getValidatedSupporterPlusCustomerData(zuoraClient, supporterPlusIds)
+    val isValidSupporterPlusStartDate = isValidStartDateForPlan(MonthlySupporterPlus, _: LocalDate)
+    val validateRequest = SupporterPlusValidations(isValidSupporterPlusStartDate, AmountLimits.limitsFor()) _ //todo make sure amountLimits is using supporter plus
+    val supporterPlusSqsSend = awsSQSSend(emailQueueNames.supporterPlus)
+
+    val supporterPlusBrazeConfirmationSqsSend = EtSqsSend[SupporterPlusEmailData](supporterPlusSqsSend) _
+    val sendConfirmationEmail = SendConfirmationEmail(supporterPlusBrazeConfirmationSqsSend) _
+
+    AddSupporterPlus.steps(
+      getPlan = catalog.planForId,
+      getCurrentDate = currentDate,
+      getPlanAndCharge = planAndChargeForSupporterPlusPlanId,
+      getCustomerData = getCustomerData,
+      supporterPlusValidations = validateRequest,
+      createSubscription = createSubscription,
+      sendConfirmationEmail = sendConfirmationEmail
+    ) _
+
+  }
+
+  def paymentDelayFor(paymentMethod: PaymentMethod): Long = paymentMethod match {
+    case d: DirectDebit => 10l
+    case _ => 0l
+  }
+
+  def toSupporterPlusEmailData(
+     request: AddSubscriptionRequest,
+     plan: Plan,
+     currency: Currency,
+     paymentMethod: PaymentMethod,
+     firstPaymentDate: LocalDate,
+     contacts: Contacts,
+     amountMinorUnits: AmountMinorUnits,
+     currentDate: LocalDate
+   ) =
+    SupporterPlusEmailData(
+      accountId = request.zuoraAccountId,
+      currency = currency,
+      plan = plan,
+      paymentMethod = paymentMethod,
+      amountMinorUnits = amountMinorUnits,
+      firstPaymentDate = firstPaymentDate,
+      contacts = contacts,
+      created = currentDate
+    )
+
+  def getValidatedSupporterPlusCustomerData(
+   zuoraClient: Requests,
+   supporterPlusPlanIds: List[ProductRatePlanId]
+  ): ZuoraAccountId => ApiGatewayOp[SupporterPlusCustomerData] = {
+
+    val validateAccount = ValidateAccount.apply _ thenValidate SupporterPlusAccountValidation.apply _
+    val getValidatedAccount = GetAccount(zuoraClient.get[ZuoraAccount]) _ andValidateWith (
+      validate = validateAccount,
+      ifNotFoundReturn = Some("Zuora account id is not valid")
+    )
+    val getValidatedPaymentMethod = GetPaymentMethod(zuoraClient.get[PaymentMethodWire]) _ andValidateWith ValidatePaymentMethod.apply _
+    val validateSubs = ValidateSubscriptions(supporterPlusPlanIds, "Zuora account already has an active supporter plus subscription") _
+    val getValidatedSubs = GetAccountSubscriptions(zuoraClient.get[ZuoraSubscriptionsResponse]) _ andValidateWith validateSubs
+    val getContactsFromZuora = GetContacts(zuoraClient.get[GetContactsResponse]) _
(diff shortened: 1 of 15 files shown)


david-pepper:
[update new-product-api to orchestrate acquisition of new product prop](https://trello.com/c/trlucSMG/3437-update-new-product-api-to-orchestrate-acquisition-of-new-product-prop)

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/AddSupporterPlus.scala
```diff
@@ -0,0 +1,113 @@
...
+   getPlan: PlanId => Plan,
+   getCurrentDate: () => LocalDate,
+   getPlanAndCharge: PlanId => Option[PlanAndCharge],
+   getCustomerData: ZuoraAccountId => ApiGatewayOp[SupporterPlusCustomerData],
+   supporterPlusValidations: (SupporterPlusValidations.ValidatableFields, PlanId, Currency) => ValidationResult[AmountMinorUnits],
+   createSubscription: ZuoraCreateSubRequest => ClientFailableOp[SubscriptionName]
+ )(request: AddSubscriptionRequest): AsyncApiGatewayOp[SubscriptionName] = {
+    for {
+      customerData <- getCustomerData(request.zuoraAccountId).toAsync
+      SupporterPlusCustomerData(account, paymentMethod, subscriptions, contacts) = customerData
```
paulbrown1982, line 37: I've never seen this technique before: deconstructing a case class using its arguments! Is it a Scala 3 thing?

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/validation/supporterplus/GetSupporterPlusCustomerData.scala
```diff
@@ -0,0 +1,32 @@
...
+    getAccount: ZuoraAccountId => ApiGatewayOp[ValidatedAccount],
+    getPaymentMethod: PaymentMethodId => ApiGatewayOp[PaymentMethod],
+    getContacts: ZuoraAccountId => ApiGatewayOp[Contacts],
+    getAccountSubscriptions: ZuoraAccountId => ApiGatewayOp[List[Subscription]],
+    accountId: ZuoraAccountId
+  ) = for {
+    account <- getAccount(accountId)
+    paymentMethod <- getPaymentMethod(account.paymentMethodId)
+    accountSubscriptions <- getAccountSubscriptions(accountId)
+    contacts <- getContacts(accountId)
```
paulbrown1982, line 29: Not sure if this is necessary/beneficial for the customer's experience, but this method could be sped up by completing the 3 async calls (`getAccount`, `getAccountSubscriptions` and `getContacts`) in parallel rather than sequentially within the for comprehension (flatMap), only calling the `getPaymentMethod` once the `getAccount` call has returned.

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/AddSupporterPlus.scala
david-pepper: I don't know. I ripped it from the pre-existing Contribution logic. Think it was @pvighi who implemented it?

johnduffell: that's been around as long as I know, normally you would do it with tuples but it works for everything with an `unapply` method
`val (matches, nonMatches) = list.partitionBy(_.isUppercase)`

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/validation/supporterplus/GetSupporterPlusCustomerData.scala
```diff
@@ -0,0 +1,32 @@
...
+  } yield SupporterPlusCustomerData(account, paymentMethod, accountSubscriptions, contacts)
```
johnduffell, line 30: as per standup comments, you can implement it by a bit of a cheat with `val`s as follows
```suggestion
  ) = {
    val eventualAccount = getAccount(accountId)
    val eventualSubscriptions = getAccountSubscriptions(accountId)
    val eventualContacts = getContacts(accountId)
    for {
      account <- eventualAccount
      paymentMethod <- getPaymentMethod(account.paymentMethodId)
      accountSubscriptions <- eventualSubscriptions
      contacts <- eventualContacts
    } yield SupporterPlusCustomerData(account, paymentMethod, accountSubscriptions, contacts)
  }
```

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/AddSupporterPlus.scala
```diff
@@ -0,0 +1,113 @@
...
+      validatableFields = ValidatableFields(request.amountMinorUnits, request.startDate)
+      amountMinorUnits <- supporterPlusValidations(validatableFields, request.planId, account.currency).toApiGatewayOp.toAsync
+      acceptanceDate = request.startDate.plusDays(paymentDelayFor(paymentMethod))
+      planAndCharge <- getPlanAndCharge(request.planId).toApiGatewayContinueProcessing(internalServerError(s"no Zuora id for ${request.planId}!")).toAsync
+      chargeOverride = ChargeOverride(Some(amountMinorUnits), planAndCharge.productRatePlanChargeId, None)
```
johnduffell, line 42: something odd about the indentation here I think? maybe needs to remove one space?

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/validation/supporterplus/GetSupporterPlusCustomerData.scala
david-pepper: Many thanks to @johnduffell for his suggestion as to how to implement this!

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/AddSupporterPlus.scala
```diff
@@ -0,0 +1,113 @@
...
+      getPlanAndCharge = planAndChargeForSupporterPlusPlanId,
+      getCustomerData = getCustomerData,
+      supporterPlusValidations = validateRequest,
+      createSubscription = createSubscription,
+    ) _
+
+  }
+
+  def paymentDelayFor(paymentMethod: PaymentMethod): Long = paymentMethod match {
+    case d: DirectDebit => 10l
```
johnduffell, line 85: not a blocker but that does surprise me - our agreement with gocardless I was understanding it was only 2-3 days or so.  When we acquire through support-frontend for Recurring contribution there is no delay afaik.

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/AddSupporterPlus.scala
```diff
@@ -0,0 +1,113 @@
...
+    val getValidatedAccount = GetAccount(zuoraClient.get[ZuoraAccount]) _ andValidateWith (
+      validate = validateAccount,
+      ifNotFoundReturn = Some("Zuora account id is not valid")
+    )
+    val getValidatedPaymentMethod = GetPaymentMethod(zuoraClient.get[PaymentMethodWire]) _ andValidateWith ValidatePaymentMethod.apply _
+    val validateSubs = ValidateSubscriptions(supporterPlusPlanIds, "Zuora account already has an active supporter plus subscription") _
+    val getValidatedSubs = GetAccountSubscriptions(zuoraClient.get[ZuoraSubscriptionsResponse]) _ andValidateWith validateSubs
+    val getContactsFromZuora = GetContacts(zuoraClient.get[GetContactsResponse]) _
+    val getUnvalidatedContacts = getContactsFromZuora.andThen(_.toApiGatewayOp("getting contacts from Zuora"))
+    GetSupporterPlusCustomerData(
```
johnduffell, line 104: can we just inline GetSupporterPlusCustomerData as it just seems to be a for comprehension?  I realise it's got tests but I'm not sure if it's necessary to test something so simple on its own?

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/Handler.scala
```diff
@@ -38,10 +38,23 @@ object Handler extends Logging {
     ApiGatewayHandler(LambdaIO(inputStream, outputStream, context)) {
       Steps.operationForEffects(RawEffects.response, RawEffects.stage, GetFromS3.fetchString, SqsAsync.send(SqsAsync.buildClient), RawEffects.now)
     }
+
+  def main(args: Array[String]): Unit = {
```
johnduffell, line 42: not sure what the something is - would this just wire up the dependencies and waiting for an input stream/request body?
```suggestion
  // run this locally to something?
  def main(args: Array[String]): Unit = {
```

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/validation/supporterplus/AmountLimits.scala
```diff
@@ -0,0 +1,58 @@
...
+import com.gu.i18n.Currency._
+import com.gu.newproduct.api.productcatalog.PlanId
+import com.gu.newproduct.api.productcatalog.PlanId.AnnualSupporterPlus
+
+case class SupporterPlusLimits(monthly: AmountLimits, annual: AmountLimits)
+
+case class AmountLimits(min: Int, max: Int)
+
+object AmountLimits {
+
```
johnduffell, line 13: stripping off the `00` from each via a utility constructor would make it much easier to check they are reasonable
```suggestion
  def fromMajorUnits(min: Int, max: Int) = AmountLimits(min * 100, max * 100)
```

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/validation/supporterplus/AmountLimits.scala
```diff
@@ -0,0 +1,58 @@
...
+  val gbp = SupporterPlusLimits(
+    monthly = AmountLimits(min = 1200, max = 16600),
+    annual = AmountLimits(min = 11900, max = 200000)
+  )
```
johnduffell, line 17: stripping off the `00` from each via a utility constructor would make it much easier to check they are reasonable
```suggestion
  def fromMajorUnits(min: Int, max: Int) = AmountLimits(min * 100, max * 100)

  val gbp = SupporterPlusLimits(
    monthly = AmountLimits.fromMajorUnits(min = 12, max = 166),
    annual = AmountLimits.fromMajorUnits(min = 119, max = 2000)
  )
```
etc

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/validation/supporterplus/SupporterPlusValidations.scala
```diff
@@ -0,0 +1,32 @@
...
+import com.gu.newproduct.api.addsubscription.validation.{ValidationResult}
+import com.gu.newproduct.api.addsubscription.validation.Validation._
+import com.gu.newproduct.api.productcatalog.{AmountMinorUnits, PlanId}
+
+object SupporterPlusValidations {
+
+  case class ValidatableFields(amountMinorUnits: Option[AmountMinorUnits], startDate: LocalDate)
+
+  //TODO ensure upper threshold is being applied
+  //TODO format amount in response
```
johnduffell, line 15: I think at least one of these is done now?

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/AddSupporterPlus.scala
```diff
@@ -0,0 +1,113 @@
...
+   getPlanAndCharge: PlanId => Option[PlanAndCharge],
+   getCustomerData: ZuoraAccountId => ApiGatewayOp[SupporterPlusCustomerData],
+   supporterPlusValidations: (SupporterPlusValidations.ValidatableFields, PlanId, Currency) => ValidationResult[AmountMinorUnits],
+   createSubscription: ZuoraCreateSubRequest => ClientFailableOp[SubscriptionName]
+ )(request: AddSubscriptionRequest): AsyncApiGatewayOp[SubscriptionName] = {
+    for {
+      customerData <- getCustomerData(request.zuoraAccountId).toAsync
+      SupporterPlusCustomerData(account, paymentMethod, subscriptions, contacts) = customerData
+      validatableFields = ValidatableFields(request.amountMinorUnits, request.startDate)
+      amountMinorUnits <- supporterPlusValidations(validatableFields, request.planId, account.currency).toApiGatewayOp.toAsync
```
johnduffell, line 39: what's the different between request.amountMinorUnits and this amountMinorUnits?

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/validation/supporterplus/SupporterPlusAccountValidation.scala
```diff
@@ -0,0 +1,12 @@
+package com.gu.newproduct.api.addsubscription.validation.supporterplus
+
+import com.gu.newproduct.api.addsubscription.validation.{Failed, Passed, ValidatedAccount, ValidationResult}
+
+object SupporterPlusAccountValidation {
```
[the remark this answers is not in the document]
david-pepper: This validation logic is identical to that of the Contribution product, as well as Guardian Weekly. Digipack has what looks to me to be a slightly different implementation of the same logic. So yes, this logic is duplicated across the products and it's arguable that centralising the logic would be the optimum way to proceed,  but also arguable that each product has its own tract of account validation to be applied or updated should the product-specific validation rules change. Both are compelling arguments and for me personally neither of them trump the other, so I am inclined to opt for the 'leaving it as-is' option.

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/AddSupporterPlus.scala
david-pepper: The logic has been replicated from the Contribution product which has been in play for several years, so given the consistent approach I think it is 'good enough', however I'm open an alternative approach if this is something we feel strongly about

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/AddSupporterPlus.scala
david-pepper: Looks like request.amountMinorUnits is an attribute on a Case Class (AddSubscriptionRequest) and amountMinorUnits on line 39 is just a local variable? Again, this has been replicated from the Contribution logic, so to understand the nuance here it would make sense to pair with a Scala developer

david-pepper:
@pvighi As the original developer of the new-product-api lambda (and in @johnduffell's absence) please could you add your thoughts to this PR? You'll notice that I've directly replicated a lot of the logic in play for Contribution product because of its similarity to the new product prop.

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/AddSupporterPlus.scala
pvighi, line 37: minor: you might be able to remove subscriptions and contacts here if they are not used below.
something like:
```scala
SupporterPlusCustomerData(account, paymentMethod, _, _) = customerData
```

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/AddSupporterPlus.scala
pvighi: it looks like they are pretty much the same, the validation returns either an error or the amount. Technically amountMinorUnits on line 39 is guaranteed to have passed the validation unlike request.amountMinorUnits, but because this is in a for comprehension it would not continue execution it if the validation fails.

 Maybe at some point in the history of the code you based this on the validation could return a modified  amount or something ?

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/Handler.scala
pvighi: I'm confused as to what is this new main method is supposed to do. It looks like it only wires the code but it never makes it run.

To run this locally we have [this code](https://github.com/guardian/support-service-lambdas/blob/68c025ada78ab38c8bbf06526194fa63b083ace4/handlers/new-product-api/src/test/scala/manualTest/AddSubscriptionManualTest.scala#L9) that pretty much calls the handler but it passes a hardcoded request json in the inputstream and it prints out the response from the outputstream

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/validation/supporterplus/SupporterPlusAccountValidation.scala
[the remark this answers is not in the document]
pvighi: I think maybe when the code was originally written it was not clear that the Account validation would end up being the same for all products except paper, or maybe it changed over time. It's not a ton of code so it's not that bad but we could consider replacing them all with a single `DigitalAccountValidation`. Of course this is not an issue introduced by this PR so it can be left for later

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/validation/supporterplus/SupporterPlusValidations.scala
```diff
@@ -0,0 +1,26 @@
...
+import com.gu.i18n.Currency
+import com.gu.newproduct.api.addsubscription.validation.{ValidationResult}
+import com.gu.newproduct.api.addsubscription.validation.Validation._
+import com.gu.newproduct.api.productcatalog.{AmountMinorUnits, PlanId}
+
+object SupporterPlusValidations {
+
+  case class ValidatableFields(amountMinorUnits: Option[AmountMinorUnits], startDate: LocalDate)
+
+  def apply(isValidStartDate: LocalDate => ValidationResult[Unit], limitsFor: (PlanId, Currency) => AmountLimits)(
```
pvighi, line 14: Unless I'm missing something it seems like this is the same as contributions. Could we use the same code if it's meant to be the same ? Maybe we can rename it to something that applies to both.
This code was not previously duplicated and it's more complicated that the Accounts one.

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/validation/supporterplus/AmountLimits.scala
david-pepper: Done

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/validation/supporterplus/SupporterPlusValidations.scala
david-pepper: Yes, both are done and comments are removed

image


image

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/validation/supporterplus/SupporterPlusValidations.scala
graham228221: Just to counter - if/when "Supporter Plus" actually launches it is likely to look quite different, so keeping "Supporter Plus" separate in most places at least gives us a skeleton to work from if/when we get more information about what it will look like.

We are currently being pushed to get the current version of Supporter Plus ready to go, and this current version is almost identical to contributions in most regards. But we are in the odd position of knowing that it will change _at some point_!

I'd also argue for whatever the least work is, as due to the uncertainty any or even all this work may be wasted effort - so get it done to meet the current goal, and we worry about tidying up later if and when we get actual finalised requirements for the new product.

KentGrm (approved):

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/AddSupporterPlus.scala
david-pepper: Ok I'll keep it the same as contributions for now

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/validation/supporterplus/SupporterPlusAccountValidation.scala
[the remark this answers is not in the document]
david-pepper: Thanks Patricio. One for the future.

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/Handler.scala
david-pepper: Yes you're right this was an attempt to run the logic and verify that it ran without error, but it really didn't work very well since it needs input of some sort and I haven't figured out how to do that yet. Thanks for the pointer towards AddSubscriptionManualTest as an alternative

david-pepper:
Thanks everyone for their insights and time on this PR. It has surfaced a few interesting areas of discussion, mainly around the  duplication of logic and how it can or should be applied to nascent concepts (which in this case is a new product). Another area of interest here is how robustly should we stick to established development patterns when enhancing or extending the logic, and how does the domain-specific experience of the developer making the changes factor into the equation?

I am happy that the changes in this PR are 'good enough' to satisfy the requirements set out initially. Should it be decided that the duplicated logic should be coalesced into one reusable unit, then it is simple enough in nature that it would not be too taxing to do so. I would suggest that the next time we revisit this logic, either for bug-fixing, extension or enhancement, that we take into account if it makes sense to centralise the logic rather than expand it outwards (or not).

PR #1697, opened 2022-09-01
Update scala-steward.yml

GHaberis:
## What does this change?

This action isn't currently running on `gu-scala-steward-public-repos` pull-requests, this change should mean it does from now on.


rupertbates (approved):
👍

PR #1700, opened 2022-09-02
Update scala-steward.yml

GHaberis:
## What does this change?

This is still skipping `gu-scala-steward-public-repos` PRs, it looks like we need to append "[bot]" to the username for it to run on these PRs.


johnduffell (approved):
fair enough!

PR #1706, opened 2022-09-09
Update api spec

jonflynng:
## What does this change?

 Changes the `product-move` endpoint to only return the new subscription number, removing the product detail with billing and information about the offer.

## Why?

 To return the entire product detail we would need to write some more code and make another Zuora call. [manage-frontend](https://github.com/guardian/manage-frontend) already has this product detail which is returned from the previous endpoint; `available-product-moves`. We will have to amend manage-frontend to carry the state from this endpoint between the different pages within the product-switch journey, probably via `react-router`.

## link the PR from manage-frontend when done


handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/available/AvailableProductMovesEndpointTypes.scala
```diff
@@ -57,7 +57,7 @@ object MoveToProduct {
 
       introOffer = Offer(Billing(amount = None, percentage = Some(50), currency = None, frequency = None, startDate = Some(localDateToString.format(chargedThroughDate))), TimePeriod(TimeUnit.month, 3))
       newPlan = Billing(amount = Some(price.toInt), percentage = None, currency = Some(GBP), frequency = Some(TimePeriod(TimeUnit.fromString(billingPeriod), 1)), startDate = Some(localDateToString.format(chargedThroughDate.plusDays(90))))
-    } yield MoveToProduct(id = subscriptionName, name = "Digital Pack", trial = Some(Trial(dayCount = 14)), introOffer = Some(introOffer), billing = newPlan)
+    } yield MoveToProduct(id = productRatePlan.id, name = "Digital Pack", trial = Some(Trial(dayCount = 14)), introOffer = Some(introOffer), billing = newPlan)
```
jonflynng, line 60: Small fix, we need the `productRatePlanId` here

PR #1756, opened 2022-10-31
new-product-api lambda - Update lower boundary thresholds for supporter plus

david-pepper:
## What does this change?
Update the lower boundary thresholds for each currency for qualification to Supporter Plus product:

image

### Suggested future work
These values should be stored in config that is accessible by the logic, allowing for more straightforward updates should the thresholds change in future, which they probably will (inflation etc)


david-pepper:
[Supporter Plus pricing in SF: Update Lambdas](https://trello.com/c/wXCv3XYd/3586-supporter-plus-pricing-in-sf-update-lambdas)

graham228221 (approved):

PR #1767, opened 2022-11-16
[Product-switching] Handle refunds

jonflynng:
Create a new lambda and SQS queue to handle product-switching refunds. The Spike to create these refunds in Zuora is [here](https://docs.google.com/document/d/10__G7q9HMUZEe3aUcIPc66D48qZYfwz_bme6cqZ0vus/edit#heading=h.vzw0y9b1pb7h). **Put description here of what the endpoint used to do and what it now does**

If the customer needs to be refunded (denoted by a negative `totalDeltaMrr` property in the [SubscriptionUpdate response](https://github.com/guardian/support-service-lambdas/pull/1767/files#diff-64c1833ca500163ac63ffb6fe3e328c5eb0edb1e6de05ca8ace23930f94bdecaR40)), we [send a message to an SQS queue](https://github.com/guardian/support-service-lambdas/pull/1767/files#diff-8279c56bbcc6ef4ba4156a1092ef21917f961403ecd46fd162cbea1b190688b5R135-R139), which in turn triggers the new lambda to process the refund.

**Testing**

to fill in.

PR #1768, opened 2022-11-16
Email field updates for product-switching

jonflynng:
**Note:** This PR is based off [this PR](https://github.com/guardian/support-service-lambdas/pull/1767/files), so please merge that one before looking at this one. I'll then merge `main` branch into this one so the diff will be much smaller.


jonflynng:
[Update email fields in product-switch endpoint](https://trello.com/c/YApd8UYj/3608-update-email-fields-in-product-switch-endpoint)

charleycampbell (approved):
Looks good to me!

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/move/ProductMoveEndpoint.scala
```diff
@@ -77,13 +78,20 @@ object ProductMoveEndpoint {
       error => ZIO.fail(s"$message failed with: $error")
     }
 
+  extension(billingPeriod: BillingPeriod)
+    def value =
+      billingPeriod match {
```
jonflynng, line 83: @rupertbates is it worth adding a default here? I don't think it should be anything other than Monthly or Annual as that is handled in the json decoding.

PR #1783, opened 2022-12-07
Fix billing account remover bug

david-pepper:
### Context
There is an error occurring when the Billing Account Remover Lambda executes in DEV

**What is the Billing Account Remover Lambda?**
This lambda is intended to:

- run daily
- grab billing accounts from Salesforce that are marked for removal
- sever the link between the Account in Salesforce and the Billing Account in Zuora (i.e. remove the foreign Id from the Billing Account in Zuora, and set its status to 'Cancelled')
- Write back any errors to Salesforce

**Error**

```
{
  "errorMessage": "An error occurred: ",
  "errorType": "java.lang.RuntimeException",
  "stackTrace": [
    "com.gu.sf_billing_account_remover.BillingAccountRemover$.$anonfun$processBillingAccounts$10(BillingAccountRemover.scala:121)",
    "scala.util.Either$LeftProjection.foreach(Either.scala:549)",
    "com.gu.sf_billing_account_remover.BillingAccountRemover$.processBillingAccounts(BillingAccountRemover.scala:121)",
    "com.gu.sf_billing_account_remover.BillingAccountRemover$.lambda(BillingAccountRemover.scala:347)",
    "com.gu.sf_billing_account_remover.BillingAccountRemover.lambda(BillingAccountRemover.scala)",
    "sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)",
    "sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)",
    "sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)",
    "java.lang.reflect.Method.invoke(Method.java:498)"
  ],
  "cause": {
    "errorMessage": "Attempt to decode value on failed cursor: DownField(access_token)",
    "errorType": "io.circe.DecodingFailure$$anon$2",
    "stackTrace": []
  }
}
```

**Root Cause**
The global variable salesforceApiVersion was not permeating throughout the code, resulting in the following endpoint being generated:

```
https://gnmtouchpoint--dev1.sandbox.my.salesforce.com/services/data/vnull/query/
```
This is because the solution extends App which extends DelayedInit which which means that fields of the object will not have been initialized before the main method has been executed.

**Fix**
I have hardcoded the API version into the endpoint to resolve the issue

**Comments**
Ideally, the api version would be stored in a piece of config in one location, where it would be accessed by every application that uses the Salesforce REST API. There is a decision to be made about how we should go about this. Until we get to a decision and an implementation plan, hardcoding the value here is good enough.

### Notes
This is a good opportunity to update our authentication process to align with our security policy for integrations with Salesforce to have 1 user and 1 Connected App per integration.

I've created a new [User](https://gnmtouchpoint.my.salesforce.com/0055I000004BZRE?noredirect=1&isUserEntityOverride=1) in Salesforce, as well as a new [Connected App](https://gnmtouchpoint.my.salesforce.com/app/mgmt/forceconnectedapps/forceAppDetail.apexp?applicationId=06P5I000000c71v&id=0Ci5I000000PBtb)

The User Credentials and Connected App secrets have been added to secrets in AWS Secrets Manager:

**Users**
[DEV/Salesforce/User/BillingAccountRemoverAPIUser](https://eu-west-1.console.aws.amazon.com/secretsmanager/secret?name=DEV%2FSalesforce%2FUser%2FBillingAccountRemoverAPIUser&region=eu-west-1)
[PROD/Salesforce/User/BillingAccountRemoverAPIUser](https://eu-west-1.console.aws.amazon.com/secretsmanager/secret?name=PROD%2FSalesforce%2FUser%2FBillingAccountRemoverAPIUser&region=eu-west-1)

**Connected App**
[PROD/Salesforce/ConnectedApp/BillingAccountRemover](https://eu-west-1.console.aws.amazon.com/secretsmanager/secret?name=PROD%2FSalesforce%2FConnectedApp%2FBillingAccountRemover&region=eu-west-1)

### DIFF
--- a/handlers/sf-billing-account-remover/src/main/scala/com/gu/sf_billing_account_remover/BillingAccountRemover.scala
+++ b/handlers/sf-billing-account-remover/src/main/scala/com/gu/sf_billing_account_remover/BillingAccountRemover.scala
@@ -11,7 +11,9 @@ import scalaj.http._
 import scala.util.Try
 
 object BillingAccountRemover extends App with LazyLogging {
-  
+
+  val salesforceApiVersion = "54.0"
+
   //Salesforce
   case class SfAuthDetails(access_token: String, instance_url: String)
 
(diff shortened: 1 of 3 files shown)


handlers/sf-billing-account-remover/src/main/scala/com/gu/sf_billing_account_remover/BillingAccountRemover.scala
```diff
@@ -172,7 +168,7 @@ object BillingAccountRemover extends App with LazyLogging {
   }
 
   def doSfGetWithQuery(sfAuthDetails: SfAuthDetails, query: String): String = {
-    Http(s"${sfAuthDetails.instance_url}/services/data/v$salesforceApiVersion/query/")
+    Http(s"${sfAuthDetails.instance_url}/services/data/v54.0/query/")
```
graham228221, line 171: Would this not fix it?

```suggestion
    Http(s"${sfAuthDetails.instance_url}/services/data/v${salesforceApiVersion}/query/")
```

graham228221: If not, using v54 is fine for the next few years :)

david-pepper: I think you're right! I've readded the variable and usage with the correct syntax. Nice spot

rupertbates: Does  this work? I can't see why it would as `$salesforceApiVersion` is valid syntax, you only need the curly braces when the thing you're injecting is multipart eg `${user.name}`

david-pepper: Rupert is correct, it didn't work. The syntax was correct all along. I've reverted it now so that the api version is hardcoded

david-pepper:
[Salesforce Retention Policy Solution Build - Bug Fix](https://trello.com/c/iTffuR8P/3696-salesforce-retention-policy-solution-build-bug-fix)

david-pepper:
After some investigation with @rupertbates we discovered that the salesforceApiVersion variable does not get initialised when expected. This is because the solution extends App which extends DelayedInit which which means that fields of the object will not have been initialized before the main method has been executed.

image

PR #1786, opened 2022-12-07
Move stripe customer update endpoint into the stripe-webhooks project & fix cfn

rupertbates:
## What does this change?

We recently had an issue where the zuora-callout-apis project deployment caused an update of the cloudformation stack. Why this happened is not clear because the project has been auto-deploying for years without the cloudformation changing, however, the result was that the `stripeCustomerSourceUpdated` endpoint which receives webhook updates from Stripe whenever a user changes or updates their payment information broke because the cloudformation representing it was incorrect (there were no deploys in API gateway). The aws infrastructure must have been manually deployed at some point in the past and the riffraff deploy overwrote it with an undeployed api gateway setup.

While investigating this it became clear that also this code doesn't really belong alongside the zuora callouts code and would be better situated in the stripe-webhook-endpoints project. This also is a newer project which uses SAML rather than simple cloudformation so is much simpler and less verbose.

Most of the changes in this PR are just moving files about along with a few changes to namespaces and imports.


handlers/stripe-webhook-endpoints/src/main/scala/com/gu/util/zuora/ZuoraGetAccountSummary.scala
```diff
@@ -0,0 +1,77 @@
+package com.gu.util.zuora
```
rupertbates, line 1: This was moved from zuora-callout-apis as well but for some reason git didn't pick that up and is treating it as a new file

handlers/stripe-webhook-endpoints/src/test/scala/com/gu/stripeCustomerSourceUpdated/TestData.scala
```diff
@@ -0,0 +1,31 @@
+package com.gu.stripeCustomerSourceUpdated
```
rupertbates, line 1: This was moved from zuora-callout-apis as well but for some reason git didn't pick that up and is treating it as a new file

handlers/zuora-callout-apis/cloudformation.yaml
```diff
@@ -186,65 +186,6 @@ Resources:
         - ZuoraAutoCancelLambda
         - ZuoraAutoCancelProxyResource
 
-
```
rupertbates, line 189: Delete the old cloudformation out of zuora-callout-apis

handlers/zuora-callout-apis/src/main/scala/com/gu/autoCancel/AutoCancel.scala
```diff
@@ -1,7 +1,7 @@
 package com.gu.autoCancel
 
 import com.gu.autoCancel.AutoCancelSteps.AutoCancelUrlParams
-import com.gu.stripeCustomerSourceUpdated.TypeConvert._
+import com.gu.util.TypeConvert._
```
rupertbates, line 4: All the changes below this point are because the  `TypeConvert` class was needed by both the code that moved to stripe-webhook-endpoints and the code that remained in zuora-callout-apis, so I have duplicated it and changed the namespace slightly

handlers/stripe-webhook-endpoints/cfn.yaml
```diff
@@ -62,3 +62,44 @@ Resources:
             Properties:
                 Path: '/payment-intent-issue'
                 Method: post
+  CustomerUpdatedLambda:
```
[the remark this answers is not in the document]
KentGrm: Only question I have is where each of the old end points are declared in this new cfn? I can only see one but I believe there were three?

KentGrm (approved):

PR #1787, opened 2022-12-07
Revert "Move stripe customer update endpoint into the stripe-webhooks project & fix cfn"

rupertbates:
Reverts guardian/support-service-lambdas#1786

Investigating a problem with the new-product-api so rolling this back to eliminate it as a cause


graham228221 (approved):

graham228221:
[CSR bug report (severity: 4) Salesforce cannot connect to new-product-api](https://trello.com/c/Lo2e4ZLu/3702-csr-bug-report-severity-4-salesforce-cannot-connect-to-new-product-api)

rupertbates:
This is not needed, we found the cause and it was unrelated to this PR

PR #1794, opened 2022-12-12
Reduce batch size for Delivery Credit Processor

graham228221:
## What does this change?
This reduces the batch size for the delivery-problem-credit-processor so that the batch can be completed within the 15 minute Lambda limit.

Reducing the batch size from 300 to 150 reduces the number of credits that can be processed in a day from 20,000 to about 10,000, assuming each batch runs successfully. This should be well within expectations.

We are currently processing a larger than usual number of delivery credits (around 2,000 due to a weather related supplier issue) and are noticing that each full batch of 300 records is failing at the 15 minute time limit.


david-pepper (approved):

graham228221:
[Manually credit all Home Delivery Subs due delivery on 12 December 2022 ❄️❄️❄️](https://trello.com/c/slXYM5hs/3708-manually-credit-all-home-delivery-subs-due-delivery-on-12-december-2022-%E2%9D%84%EF%B8%8F%E2%9D%84%EF%B8%8F%E2%9D%84%EF%B8%8F)

PR #1798, opened 2022-12-19
Increase RAM for Stripe Card Updater Lambda to speed it up

paulbrown1982:
## What does this change?
Increases RAM (and hence CPU power) to 1536MB to help ensure Lambda completes before 29s (API Gateway's Limit - https://stackoverflow.com/questions/54299958/how-can-i-set-the-aws-api-gateway-timeout-higher-than-30-seconds). Average maximum (green line) appears to be too close to 29s.
Screenshot 2022-12-19 at 10 50 03

## How to test
Deploy to CODE via RiffRaff.

Deploy via RiffRaff and observe reduction Stripe errors.

Low - increasing Lambda memory is standard stuff.

## Images

N/A

## Accessibility

N/A

### DIFF
--- a/handlers/stripe-webhook-endpoints/cfn.yaml
+++ b/handlers/stripe-webhook-endpoints/cfn.yaml
@@ -69,7 +69,7 @@ Resources:
       Description: A lambda for handling customer updates
       Runtime: java11
       Handler: com.gu.stripeCardUpdated.Lambda::apply
-      MemorySize: 1356
+      MemorySize: 1536
       Timeout: 900
       Environment:
         Variables:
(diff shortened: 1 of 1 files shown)


handlers/stripe-webhook-endpoints/cfn.yaml
```diff
@@ -69,7 +69,7 @@ Resources:
       Description: A lambda for handling customer updates
       Runtime: java11
       Handler: com.gu.stripeCardUpdated.Lambda::apply
-      MemorySize: 512
+      MemorySize: 1024
```
rupertbates, line 72: The previous lambda has 1536 so probably good to go with that

rupertbates (approved):
👍

PR #1801, opened 2022-12-20
Adding in logging to the Zuora library 

charleycampbell:
## What does this change?
Adding in logging to the Zuora library to log the response and request bodies when hitting the subscriptionGetResponse and subscriptionUpdateResponse. These logs will help with looking into the issue that is occurring around holiday stops and negative invoices incorrectly being created.

We will need to check in Cloudwatch to make sure that it is logging as expected when in production

### DIFF
--- a/lib/zuora-core/src/main/scala/com/gu/zuora/Zuora.scala
+++ b/lib/zuora-core/src/main/scala/com/gu/zuora/Zuora.scala
@@ -5,6 +5,7 @@ import com.typesafe.scalalogging.LazyLogging
 import io.circe.generic.auto._
 import sttp.client3._
 import sttp.client3.circe._
+import io.circe.syntax._
 
 import scala.annotation.tailrec
 
@@ -57,7 +58,7 @@ object Zuora extends LazyLogging {
       .mapResponse(_.left.map(e => ZuoraApiFailure(e.getMessage)))
       .send(backend)
 
-    logger.info(s"subscriptionGetResponse for sub ${subscriptionName.value} is ${response.body.toString}")
+    logger.info(s"subscriptionGetResponse for sub ${subscriptionName.value} is ${response.body.asJson}")
     response.body
   }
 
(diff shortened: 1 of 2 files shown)


lib/zuora-core/src/main/scala/com/gu/zuora/Zuora.scala
```diff
@@ -51,17 +51,21 @@ object Zuora {
...
   def subscriptionGetResponse(config: ZuoraConfig, accessToken: AccessToken, backend: SttpBackend[Identity, Any])(subscriptionName: SubscriptionName): ZuoraApiResponse[Subscription] = {
-    basicRequest.get(uri"${config.baseUrl}/subscriptions/${subscriptionName.value}")
+    val response = basicRequest.get(uri"${config.baseUrl}/subscriptions/${subscriptionName.value}")
       .header("Authorization", s"Bearer ${accessToken.access_token}")
       .response(asJson[Subscription])
       .mapResponse(_.left.map(e => ZuoraApiFailure(e.getMessage)))
       .send(backend)
-      .body
+
+    logger.info(s"subscriptionGetResponse for sub ${subscriptionName.value} is ${response.body.toString}")
```
rupertbates, line 60: You might get more readable output with something like 
```suggestion
    import io.circe.syntax._ //stick this at the top of the file
    logger.info(s"subscriptionGetResponse for sub ${subscriptionName.value} is ${response.body.asJson}")
```

lib/zuora-core/src/main/scala/com/gu/zuora/Zuora.scala
```diff
@@ -74,7 +78,9 @@ object Zuora {
           else Left(ZuoraApiFailure(errMsg(status.reasons.map(_.mkString).getOrElse(""))))
       }
       .send(backend)
-      .body.left.map(failure => ZuoraApiFailure(errMsg(failure.reason)))
+      logger.info(s"response for subscriptionUpdateResponse is ${response.body.toString}")
```
rupertbates, line 81: Same as above

rupertbates (approved):
Have made one suggestion but otherwise 👍

charleycampbell:
Closing pull request as the test on DEV to check the logging was unsuccessful and the ticket has been moved back onto the backlog

PR #1824, opened 2023-01-18
Add new 'Previous Product Name' property to Salesforce record

jonflynng:
A new property `Previous_Product_Name__c` will be added to the `Subscription_Rate_Plan_Update__c` Salesforce record when performing a product switch. DRR have asked for this property to be available as it will be used in targeted onboarding via Braze. It will also be used by the [soft-opt-in-consent-setter](https://github.com/guardian/support-service-lambdas/tree/main/handlers/soft-opt-in-consent-setter).


handlers/product-move-api/src/main/scala/com/gu/productmove/salesforce/CreateRecord.scala
```diff
@@ -36,6 +36,7 @@ object CreateRecord {
       Type__c: String = "Product Switch",
       Source__c: String = "MMA",
       Previous_Amount__c: BigDecimal,
+      Old_Product__c: String,
```
rupertbates, line 39: Why is it  `Old_Product__c` and not `Previous_Product_Name__c` to match the rate plan field?

jonflynng:
[Soft opt-ins for product switching](https://trello.com/c/otV25Ji1/3782-soft-opt-ins-for-product-switching)

handlers/product-move-api/src/main/scala/com/gu/productmove/salesforce/CreateRecord.scala
jonflynng: yep. @graham228221 is it ok if we can refactor this in Salesforce?

graham228221: @jonflynng no problem, you should see `Previous_Product_Name__c` instead of `Old_Product__c`.

jonflynng: @rupertbates it has been renamed

rupertbates (approved):

PR #1830, opened 2023-01-23
adjust soft opt-ins for product switches

jonflynng:
## Currently

Currently the lambda consists of two main functions `processAcquiredSubs` and `processCancelledSubs`. These functions run through newly acquired and cancelled subs and adjusts the soft opt-in consents according to [this mapping](https://s3.console.aws.amazon.com/s3/object/soft-opt-in-consent-setter?region=eu-west-1&prefix=PROD/ConsentsByProductMapping.json).


handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/ConsentsCalculator.scala
```diff
@@ -6,19 +6,34 @@ import io.circe.syntax.EncoderOps
 
 class ConsentsCalculator(consentsMappings: Map[String, Set[String]]) {
 
-  private case class ConsentsObject(id: String, consented: Boolean)
+  case class ConsentsObject(id: String, consented: Boolean)
 
-  def getAcquisitionConsents(productName: String): Either[SoftOptInError, Set[String]] = {
+  def getSoftOptInsByProduct(productName: String): Either[SoftOptInError, Set[String]] = {
```
jonflynng, line 11: just renaming here as we re-use this function for the new functionality.

handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/models/sfSubRecordUpdate.scala
```diff
@@ -15,26 +15,36 @@ case class SFSubRecordUpdateRequest(records: Seq[SFSubRecordUpdate]) {
 
 object SFSubRecordUpdate {
 
-  def apply(sub: SFSubRecord, softOptInStage: String, updateResult: Either[SoftOptInError, Unit]): SFSubRecordUpdate = {
+  def apply(
```
jonflynng, line 18: This had to be refactored as it took in a `SFSubRecord` before, but we use data from a different Salesforce record for product switches.

handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/SfQueries.scala
```diff
@@ -36,6 +36,31 @@ object SfQueries {
     """.stripMargin
   }
 
+  val getProductSwitchSubsToProcessQuery: String = {
+    val limit = 50
```
jonflynng, line 40: randomly chose 50. Any thoughts on how many records we should process at a time?

handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/ConsentsCalculator.scala
```diff
@@ -6,19 +6,34 @@ import io.circe.syntax.EncoderOps
...
     consentsMappings
       .get(productName)
       .toRight(
         SoftOptInError(
           "ConsentsCalculator",
-          s"getAcquisitionConsents couldn't find $productName in consentsMappings",
+          s"getConsentsByProduct couldn't find $productName in consentsMappings",
```
rupertbates, line 17:
```suggestion
          s"getSoftOptInsByProduct couldn't find $productName in consentsMappings",
```

handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/Handler.scala
```diff
@@ -73,14 +87,83 @@ object Handler extends LazyLogging {
...
+      newProductName: String,
+      allProductsForUser: Set[String],
+      consentsCalculator: ConsentsCalculator,
+  ): Either[SoftOptInError, String] = {
+    import consentsCalculator._
+
+    for {
+      oldProductSoftOptIns <- getSoftOptInsByProduct(oldProductName)
+      newProductSoftOptIns <- getSoftOptInsByProduct(newProductName)
+      currentProductSoftOptIns <- getSoftOptInsByProducts(allProductsForUser)
```
jonflynng, line 125: need some better naming of variables here. This represents all soft opt-ins for every product a user has.

handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/Handler.scala
```diff
@@ -73,14 +87,83 @@ object Handler extends LazyLogging {
...
+      allOtherProductSoftOptIns <- getSoftOptInsByProducts(allProductsForUser - newProductName)
```
jonflynng, line 126: need some better naming of variables here. This represents all soft opt-ins for every product a user has minus the new product they have switched to.

handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/SfQueries.scala
```diff
@@ -36,6 +36,31 @@ object SfQueries {
...
+       |    SF_Subscription__r.Soft_Opt_in_Number_of_Attempts__c,
+       |    Buyer__r.IdentityID__c
+       |FROM
+       |    Subscription_Rate_Plan_Update__c
+       |WHERE
+       |    SF_Subscription__r.Soft_Opt_in_Number_of_Attempts__c < 5 AND
+       |    SF_Subscription__r.Soft_Opt_in_Eligible__c = true AND
+       |    SF_Subscription__r.SF_Status__c = 'Active' AND
+       |    SF_Subscription__r.Soft_Opt_in_Last_Stage_Processed__c = 'Acquisition'
+       |LIMIT
```
graham228221, line 59: This logic would mean that a subscription could only be processed as a "switch" once. If we introduce further switch journeys (e.g. Supporter Plus to Recurring Contribution) it may be necessary to facilitate multiple "switch" soft opt-in processing.

This might mean having an additional `Soft_Opt_in_Processed__c` field on `Subscription_Rate_Plan_Update__c` which is updated in Salesforce record writeback so that we know whether a specific switch has been processed.

On the other hand, this might be ok as a V1 as it wouldn't likely be an issue until more switch journeys are enabled.

handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/SfQueries.scala
graham228221: `getSubsToProcessQuery` gets 200, and there's no extra complexity, so it should be ok to have the same limit for both?

jonflynng:
[Soft opt-ins for product switching](https://trello.com/c/otV25Ji1/3782-soft-opt-ins-for-product-switching)

handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/SfQueries.scala
jonflynng: True, let's do that as part of this PR, do you mind adding the field?

graham228221: Sure, I think we should also add a `Soft_Opt_in_Number_of_Attempts__c` to `Subscription_Rate_Plan_Update__c` then on a failure you can write back to this in a similar way. What do you think?

I'll create a `Soft_Opt_in_Processed__c` field as a boolean, so on successful writeback you should change this to `TRUE`.

This would change the query to:

```suggestion
       |WHERE
       |    Soft_Opt_in_Number_of_Attempts__c < 5 AND
       |    Soft_Opt_in_Processed__c = false
       |    SF_Subscription__r.SF_Status__c = 'Active' AND
       |LIMIT
```

jonflynng: sounds good

graham228221: You should now see `Soft_Opt_In_Processed__c` (boolean) and `Soft_Opt_in_Number_of_Attempts__c` (integer) in the DEV Salesforce sandbox

handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/ConsentsCalculator.scala
```diff
@@ -6,19 +6,34 @@ import io.circe.syntax.EncoderOps
...
 
+  def getSoftOptInsByProducts(productNames: Set[String]): Either[SoftOptInError, Set[String]] = {
+    productNames
+      .foldLeft[Either[SoftOptInError, Set[String]]](Right(Set())) { (acc, ownedProductName) =>
+        consentsMappings
+          .get(ownedProductName)
+          .toRight(
+            SoftOptInError(
+              "ConsentsCalculator",
+              s"getCancellationConsents couldn't find $ownedProductName in consentsMappings",
```
graham228221, line 30:
```suggestion
              s"getSoftOptInsByProducts couldn't find $ownedProductName in consentsMappings",
```

graham228221 (commented):
I think this looks good, but I'll leave it to a Scala dev to approve if that's ok!

david-pepper:
_The general algorithm is as follows:

If the user has something turned off, it should remain turned off in the switch, even if, at the point of acquisition it would normally be turned on. We should also remove any consents where necessary._

So basically, if an permission is opted out, keep it opted out. What happens for permissions that are opted in on?

david-pepper:
In the 'What's New' section there are two links: 

[Salesforce query](https://github.com/guardian/support-service-lambdas/pull/1830/files#diff-8eda28b3273f518446cfde3c6dd8a9ae6bad6cdc2e8b9582d2f76bac7937cd80R35) to fetch data for product switches and a [processProductSwitchSubs](https://github.com/guardian/support-service-lambdas/pull/1830/files#diff-8eda28b3273f518446cfde3c6dd8a9ae6bad6cdc2e8b9582d2f76bac7937cd80R54)

Neither of these links take me where I would expect them to.

Salesforce query takes me here:
image

processProductSwitchSubs takes me here:
image

jonflynng:
> Would it have been possible to update the original query to also grab subs that have underwent a switch? And then to filter out those subs from the returned recordset and process accordingly?

As SOQL can't do joins i don't know how I would do that in one query.

jonflynng:
> In the 'What's New' section there are two links:
> 
> [Salesforce query](https://github.com/guardian/support-service-lambdas/pull/1830/files#diff-8eda28b3273f518446cfde3c6dd8a9ae6bad6cdc2e8b9582d2f76bac7937cd80R35) to fetch data for product switches and a [processProductSwitchSubs](https://github.com/guardian/support-service-lambdas/pull/1830/files#diff-8eda28b3273f518446cfde3c6dd8a9ae6bad6cdc2e8b9582d2f76bac7937cd80R54)
> 
> Neither of these links take me where I would expect them to.
> 
> Salesforce query - I would expect it to take me to the query, but it takes me here: image
> 
> processProductSwitchSubs - I would expect it to take me to the function, but it takes me here: image

Yeah, these links are outdated as i changed the code.

jonflynng:
> _The general algorithm is as follows:_
> 
> _If the user has something turned off, it should remain turned off in the switch, even if, at the point of acquisition it would normally be turned on. We should also remove any consents where necessary._
> 
> So basically, if an permission is opted out, keep it opted out. What happens for permissions that are opted in on?

Keep it switched on, unless the new product does not have that soft opt-in, then turn it off.

rupertbates (approved):

graham228221 (approved):
Thanks Jon!

PR #1832, opened 2023-01-24
Product move alarm

rupertbates:
## What does this change?

This PR adds two alarms to the product move api so that we can tell when it is failing. 

They are triggered by a single 5XX Api Gateway response or multiple  4XXs over a time period - this period is quite low at the moment so we may need to tweak it if it is too noisy.

## How to test
To test this I removed the `isProd` restriction in the cloudformation template for these alarms, deployed them to dev and then triggered a 500 response by renaming a [config file in S3](s3://gu-reader-revenue-private/membership/support-service-lambdas/DEV/).
Next time we have an issue with either the product-move api or the supporter plus cancellation api, we should receive an alarm immediately.

### DIFF
--- a/handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/cancel/SubscriptionCancelEndpoint.scala
+++ b/handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/cancel/SubscriptionCancelEndpoint.scala
@@ -128,16 +128,6 @@ object SubscriptionCancelEndpoint {
   private def subIsWithinFirst14Days(now: LocalDate, contractEffectiveDate: LocalDate) =
     now.isBefore(contractEffectiveDate.plusDays(15)) // This is 14 days from the day after the sub was taken out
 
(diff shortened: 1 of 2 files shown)


handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/cancel/SubscriptionCancelEndpoint.scala
```diff
@@ -128,16 +128,6 @@ object SubscriptionCancelEndpoint {
   private def subIsWithinFirst14Days(now: LocalDate, contractEffectiveDate: LocalDate) =
     now.isBefore(contractEffectiveDate.plusDays(15)) // This is 14 days from the day after the sub was taken out
 
-  private def getEffectiveCancellationDate(
```
rupertbates, line 131: This was unused so I removed it

pawel-krupinski-guardian (approved):
Looks fine

PR #1842, opened 2023-02-02
Use refund lambda in S+ cancellation

jonflynng:
Upon refunding customers when cancelling a supporter plus subscription within 14 days, a negative invoice was being generated by Zuora:

Screenshot 2023-02-02 at 15 24 01

 This happens if a sub is cancelled before the end of the current billing period. To get around this, we do a refund without adjusting any invoice items on the original invoice. We then transfer the generated negative invoice to credit, and use that credit to pay off the original invoice, which now contains a positive balance due to the refund. The refund lambda we use for product switches does exactly this.


jonflynng:
[Fix Zuora generating a negative invoice on S+ cancellations](https://trello.com/c/kPwFu7bg/3833-fix-zuora-generating-a-negative-invoice-on-s-cancellations)

rupertbates (approved):

PR #1844, opened 2023-02-06
Convert `invoiceId` in Zuora cancellation response to Option

jonflynng:


handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/cancel/SubscriptionCancelEndpoint.scala
```diff
@@ -176,7 +176,7 @@ object SubscriptionCancelEndpoint {
       _ <-
         if (shouldBeRefunded)
           SQS
-            .queueRefund(RefundInput(subscriptionName, cancellationResponse.invoiceId, charge.price))
+            .queueRefund(RefundInput(subscriptionName, cancellationResponse.invoiceId.get, charge.price))
```
jonflynng, line 179: no error handling as the negative invoice should always be here as the customer is being refunded.

rupertbates: You could put `invoiceId.isDefined` in the if statement above

rupertbates (approved):

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/cancel/SubscriptionCancelEndpoint.scala
jonflynng: but then no error will be triggered if they should be refunded but there is no negative invoice attached. I will throw an error inside this if statement if no negative invoice exists.

PR #1861, opened 2023-02-10
Use paidAmount instead of totalDeltaMrr

jonflynng:
## What does this change?

The amount paid/refunded during switch was being incorrected calculated by using the `totalDeltaMrr` property received from the subscription update response from Zuora. This is not reliable for annual subs. Instead, we use the `paidAmount` property present in the response if an amount has been paid upon switch. 

 After discussion, it appears we would have to calculate the refund amount the same way we are doing in the `preview` function, by querying a list of invoices and calculating it manually from that list. This could be done via the `billing-preview` endpoint. However, after further discussion, we have decided to remove any code related to refunds as we believe it's unlikely these will be necessary in the future. As a reminder, refunds are only possible if the customer lowers their price on switch, which is not yet possible in MMA.

### DIFF
--- a/handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/move/ProductMoveEndpoint.scala
+++ b/handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/move/ProductMoveEndpoint.scala
@@ -222,8 +222,8 @@ object ProductMoveEndpoint {
                 first_name = account.billToContact.firstName,
                 last_name = account.billToContact.lastName,
                 currency = account.basicInfo.currency.symbol,
-                price = price.toString,
-                first_payment_amount = paidAmount.toString,
+                price = price.setScale(2, BigDecimal.RoundingMode.FLOOR).toString,
+                first_payment_amount = paidAmount.setScale(2, BigDecimal.RoundingMode.FLOOR).toString,
                 date_of_first_payment = todaysDate.format(DateTimeFormatter.ofPattern("d MMMM uuuu")),
                 payment_frequency = billingPeriod,
                 contribution_cancellation_date = todaysDate.format(DateTimeFormatter.ofPattern("d MMMM uuuu")),
@@ -242,7 +242,7 @@ object ProductMoveEndpoint {
       .queueSalesforceTracking(
         SalesforceRecordInput(
           subscriptionName,
-          price,
+          BigDecimal(ratePlanCharge.price),
           price,
           currentRatePlan.productName,
           currentRatePlan.ratePlanName,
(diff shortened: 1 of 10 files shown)


handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/move/ProductMoveEndpoint.scala
```diff
@@ -241,23 +240,26 @@ object ProductMoveEndpoint {
         SalesforceRecordInput(
           subscriptionName,
           price,
+          price,
```
graham228221, line 243: This is using the "after" amount for both `Previous_Amount__c` and `New_Amount__c`:

Screenshot 2023-02-10 at 17 08 22

Is it possible to get the previous amount? It might be acceptable to leave this it out if this is difficult.

jonflynng: yep, i forgot about contributions we're upselling. I can get the `price` from the current rate plan charge of the sub.

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/move/ProductMoveEndpoint.scala
```diff
@@ -241,23 +240,26 @@ object ProductMoveEndpoint {
...
           "Supporter Plus",
           todaysDate,
           todaysDate,
-          updateResponse.totalDeltaMrr.abs,
+          updateResponse.paidAmount,
         ),
       )
       .fork
 
+    /*
```
jonflynng, line 254: @rupertbates  have commented this out for now as we won't need it for V1.

jonflynng:
[Confirmation email formatting fixes and Salesforce record fix](https://trello.com/c/jYHilY1a/3873-confirmation-email-formatting-fixes-and-salesforce-record-fix)

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/move/ProductMoveEndpoint.scala
```diff
@@ -206,10 +206,12 @@ object ProductMoveEndpoint {
     updateResponse <- SubscriptionUpdate
       .update(subscription.id, ratePlanCharge.billingPeriod, price, currentRatePlan.id)
       .addLogMessage("SubscriptionUpdate")
-    totalDeltaMrr = updateResponse.totalDeltaMrr
 
     todaysDate <- Clock.currentDateTime.map(_.toLocalDate)
     billingPeriod <- ratePlanCharge.billingPeriod.value
+
+    paidAmount = updateResponse.paidAmount.getOrElse(BigDecimal(0))
```
jonflynng, line 213: this may be `null` in the edge case a person switches just before they get billed, also assuming they are not changing the price of their subscription. They would then get refunded the entire month/year which would then be used to pay off the next invoice, meaning no immediate payment will be needed. Therefore, in this case, we set the `paidAmount` to 0.

PR #1873, opened 2023-02-20
Return error message from `supporter-plus-cancel` endpoint

jonflynng:
This keeps the design consistent with the `product-move` endpoint. It also allows us to document different error responses in the API documentation.

 This has been tested in DEV.


rupertbates (approved):
👍

PR #1876, opened 2023-02-21
Add dead letter queue to refund lambda

jonflynng:
Adds a dead letter queue to the refund lambda, after the message has been tried and failed twice it will be put in this queue. An alarm will be triggered when a message is added to the queue.

### DIFF
--- a/handlers/product-move-api/cfn.yaml
+++ b/handlers/product-move-api/cfn.yaml
@@ -183,7 +183,7 @@ Resources:
     Type: AWS::SQS::Queue
     Properties:
       VisibilityTimeout: 3000
-      QueueName: !Sub product-switch-refund-queue-${Stage}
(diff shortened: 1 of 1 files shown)


handlers/product-move-api/cfn.yaml
```diff
@@ -184,6 +184,14 @@ Resources:
     Properties:
       VisibilityTimeout: 3000
       QueueName: !Sub product-switch-refund-${Stage}
+      RedrivePolicy:
+        deadLetterTargetArn: !GetAtt RefundDeadLetterQueue.Arn
+        maxReceiveCount: 2
```
jonflynng, line 189: retries the refund twice before being put in the dead letter queue

rupertbates (approved):
👍

PR #1879, opened 2023-02-23
Only send product move alarms to retention

rupertbates:
## What does this change?

We only want to send product move related alarms to retention, not the whole of reader revenue


undergroundquizscene (approved):
Thanks!

PR #1880, opened 2023-02-23
Fix `nextPaymentDate` not being calculated for annual subs

jonflynng:
## What does this change?

When attempting to calculate the `nextPaymentDate` for annual subscriptions, an error occurs because only the invoice item for the current term (12 months) is obtained. As a result, we are unable to access the next invoice needed to determine the `nextPaymentDate`. To address this issue, we have developed a workaround by adjusting the `currentTerm` property in the API call to 24 months. This enables us to retrieve two invoices for annual subscriptions that span two years, providing us with the necessary information to accurately calculate the `nextPaymentDate`.

Tested in DEV.


rupertbates (approved):
Nice

PR #1885, opened 2023-02-27
soft-opt-in-consent-setter: add support for in-app purchases

jonflynng:
## What does this change?

Adds a [new lambda](https://github.com/guardian/support-service-lambdas/pull/1885/files#diff-e74c9f0afd9cc901a6ccb53fef2049d7eefab5f803f00c7a2ea59e0be8a0b16dR128-R204) in the cloudformation stack, with the entry point in the `HandlerIAP.scala` file. The new lambda consumes events from an [SQS queue](https://github.com/guardian/support-service-lambdas/pull/1885/files#diff-e74c9f0afd9cc901a6ccb53fef2049d7eefab5f803f00c7a2ea59e0be8a0b16dR206-R214). A [dead-letter queue](https://github.com/guardian/support-service-lambdas/pull/1885/files#diff-e74c9f0afd9cc901a6ccb53fef2049d7eefab5f803f00c7a2ea59e0be8a0b16dR216-R220) has been set up, any messages which error [3 times](https://github.com/guardian/support-service-lambdas/pull/1885/files#diff-e74c9f0afd9cc901a6ccb53fef2049d7eefab5f803f00c7a2ea59e0be8a0b16dR214) will be sent to it. `MpapiConnector.scala` is used to query any mobile subscriptions a customer may have.

 Many of the changes in this PR is due to the [change in the error class](https://github.com/guardian/support-service-lambdas/pull/1885/files#diff-ee80d3c5893b5fa3a3c22f7fe18547e63654d17058d690b872f58a58b4918af1R3).

Things to do before merging:

- [x] Mobile purchases API (MPAPI) secrets added to secrets manager
- [x] In-App Purchase soft opt-ins added to the consents mapping file in S3
- [x] Tested in DEV

### DIFF
--- a/handlers/soft-opt-in-consent-setter/cfn.yaml
+++ b/handlers/soft-opt-in-consent-setter/cfn.yaml
@@ -16,18 +16,21 @@ Mappings:
       Schedule: 'rate(30 minutes)'
       SalesforceStage: PROD
       IdentityStage: PROD
+      MpapiStage: PROD
       SalesforceUsername: SoftOptInConsentSetterAPIUser
       AppName: TouchpointUpdate
     CODE:
       Schedule: 'rate(365 days)'
       SalesforceStage: CODE
       IdentityStage: CODE
+      MpapiStage: CODE
       SalesforceUsername: SoftOptInConsentSetterAPIUser
       AppName: AwsConnectorSandbox
     DEV:
       Schedule: 'rate(365 days)'
       SalesforceStage: DEV
       IdentityStage: CODE
+      MpapiStage: CODE
       SalesforceUsername: SoftOptInConsentSetterAPIUser
       AppName: AwsConnectorSandbox
 
@@ -90,6 +93,14 @@ Resources:
             !Sub
             - '{{resolve:secretsmanager:${IdentityStage}/Identity/SoftOptInConsentAPI:SecretString:identityToken}}'
             - IdentityStage: !FindInMap [ StageMap, !Ref Stage, IdentityStage ]
+          mpapiUrl:
+            !Sub
+            - '{{resolve:secretsmanager:${MpapiStage}/MobilePurchasesAPI/User/asdfasdfa:SecretString:mpapiUrl}}'
+            - MpapiStage: !FindInMap [ StageMap, !Ref Stage, IdentityStage ]
+          mpapiToken:
+            !Sub
+            - '{{resolve:secretsmanager:${MpapiStage}/MobilePurchasesAPI/User/asdfasdf:SecretString:mpapiToken}}'
+            - MpapiStage: !FindInMap [ StageMap, !Ref Stage, IdentityStage ]
       Events:
         ScheduledRun:
           Type: Schedule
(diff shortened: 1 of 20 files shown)


handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/models/SoftOptInError.scala
```diff
@@ -1,3 +1,3 @@
 package com.gu.soft_opt_in_consent_setter.models
 
-case class SoftOptInError(errorType: String, errorDetails: String)
+case class SoftOptInError(message: String) extends Exception(message)
```
jonflynng, line 3: I've changed the error class to extend `Exception`. This is because many of the libraries we use (like Circe) returns an error class that extends `Exception`. This therefore unifies all the error types in this lambda. All errors are passed up to the `handleRequest` function in the `Handler` file and handled via the `handleError` function. I also don't see the need for having two properties, `errorType` and `errorDetails`, as they just get concatenated into one string further up in the `Handler`.

handlers/soft-opt-in-consent-setter/cfn.yaml
```diff
@@ -113,6 +117,117 @@ Resources:
               Resource:
                 - arn:aws:s3::*:membership-dist/*
 
+  LambdaFunctionIAP:
+    Type: AWS::Serverless::Function
+    Condition: IsDev
+    Properties:
+      Description: Updates Identity Soft Opt-In Consents upon Acquisition and Cancellation of Subscriptions in Salesforce
```
graham228221, line 124:
```suggestion
      Description: Updates Identity Soft Opt-In Consents upon Acquisition and Cancellation of Subscriptions based on a queue populated by Salesforce and the Mobile Purchases API
```

handlers/soft-opt-in-consent-setter/cfn.yaml
```diff
@@ -113,6 +117,117 @@ Resources:
...
+      Handler: com.gu.soft_opt_in_consent_setter.HandlerIAP::handleRequest
+      CodeUri:
+        Bucket: support-service-lambdas-dist
+        Key: !Sub membership/${Stage}/soft-opt-in-consent-setter/soft-opt-in-consent-setter.jar
+      MemorySize: 512
+      Runtime: java11
+      Environment:
+        Variables:
+          Stage: !Ref Stage
+          sfApiVersion: v46.0
```
graham228221, line 135: Could we update this to the latest version pleeease? It would nice to move this into config or maybe in secrets manager alongside credentials for all lambdas at some point.
```suggestion
          sfApiVersion: v56.0
```

handlers/soft-opt-in-consent-setter/cfn.yaml
```diff
@@ -113,6 +117,117 @@ Resources:
               Resource:
                 - arn:aws:s3::*:membership-dist/*
 
+  LambdaFunctionIAP:
+    Type: AWS::Serverless::Function
+    Condition: IsDev
+    Properties:
+      Description: Updates Identity Soft Opt-In Consents upon Acquisition and Cancellation of Subscriptions in Salesforce
+      FunctionName: !Sub soft-opt-in-consent-setter-IAP-${Stage}
```
graham228221, line 125: Would it be better to have a more generic name that doesn't reference IAP? `soft-opt-in-consent-setter-v2-` maybe

handlers/soft-opt-in-consent-setter/cfn.yaml
```diff
@@ -138,6 +253,31 @@ Resources:
...
+      ComparisonOperator: GreaterThanOrEqualToThreshold
+      Dimensions:
+        - Name: FunctionName
+          Value: !Ref LambdaFunctionIAP
+      EvaluationPeriods: 2
+      MetricName: Errors
+      Namespace: AWS/Lambda
+      Period: 3600
+      Statistic: Sum
+      Threshold: 5
```
graham228221, line 278: Does this equate to `5 errors in a two hour period`? I wonder if it's a better idea to alarm on error rate or dead letter queue size if messages will be added to the queue ad hoc rather than in scheduled batches?

handlers/soft-opt-in-consent-setter/cfn.yaml
jonflynng: Yeah, was planning to rename later on once retired.

handlers/soft-opt-in-consent-setter/cfn.yaml
jonflynng: Have updated it to alarm on dead letter queue build up.

PR #1888, opened 2023-03-03
Fix missing contentLength header from Zuora

rupertbates:
## What does this change?
This fixes a bug caused by an issue described [in this PR](https://github.com/guardian/support-frontend/pull/4470)

It seems that Zuora have made a change to their response format so that they are now compressed by default. This causes an issue with some HTTP clients whereby the `Content-length` header is no longer set which in turn is causing our application code to fail.

The solution is to add an additional header `"Accept-Encoding" -> "identity"` to the request.

### DIFF
--- a/handlers/revenue-recogniser-job/src/test/scala/com/gu/recogniser/HandlerManualTest.scala
+++ b/handlers/revenue-recogniser-job/src/test/scala/com/gu/recogniser/HandlerManualTest.scala
@@ -63,7 +63,7 @@ object RunQueryManualTest {
   def main(args: Array[String]): Unit = {
 
     val actual = for {
-      zuoraRestConfig <- LoadConfigModule(Stage("PROD"), GetFromS3.fetchString)[ZuoraRestConfig]
+      zuoraRestConfig <- LoadConfigModule(Stage("DEV"), GetFromS3.fetchString)[ZuoraRestConfig]
       downloadRequests = ZuoraAquaRequestMaker(RawEffects.downloadResponse, zuoraRestConfig)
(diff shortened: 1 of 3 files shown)


lib/zuora-reports/src/main/scala/com/gu/zuora/reports/aqua/ZuoraAquaRequestMaker.scala
```diff
@@ -25,6 +25,7 @@ object ZuoraAquaRequestMaker extends LazyLogging {
     new RestRequestMaker.Requests(
       headers = Map(
         "Authorization" -> s"Basic $encodedCredentials",
+        "Accept-Encoding" -> "identity",
```
rupertbates, line 28: This is the fix

handlers/revenue-recogniser-job/src/test/scala/com/gu/recogniser/HandlerManualTest.scala
```diff
@@ -63,7 +63,7 @@ object RunQueryManualTest {
   def main(args: Array[String]): Unit = {
 
     val actual = for {
-      zuoraRestConfig <- LoadConfigModule(Stage("PROD"), GetFromS3.fetchString)[ZuoraRestConfig]
+      zuoraRestConfig <- LoadConfigModule(Stage("DEV"), GetFromS3.fetchString)[ZuoraRestConfig]
```
undergroundquizscene, line 66: What’s the reason for changing this?

undergroundquizscene (approved):
Looks good, thanks for linking to the previous explanations!

PR #1895, opened 2023-03-13
Product switching from salesforce - set case id and user id on switch record

rupertbates:
## What does this change?

When switching a recurring contributor to supporter plus through Salesforce, we want to record the CSR user id and case id in the 'Subscription Rate Plan Update' record which we write to SF at the end of the process.

## [Trello card](https://trello.com/c/8WMWYrGO/3944-csr-product-switch-wizard-receive-and-store-csr-name-and-case-number-on-switch)

### DIFF
--- a/handlers/product-move-api/get-secrets-as-env-vars.sh
+++ b/handlers/product-move-api/get-secrets-as-env-vars.sh
@@ -27,3 +27,19 @@ while read -r invoicingApiKey; read -r invoicingApiUrl; do
   #  https://stackoverflow.com/questions/32760584/in-intellij-how-do-i-set-default-environment-variables-for-new-test-configurati
   echo "invoicingApiUrl=$invoicingApiUrl;invoicingApiKey=$invoicingApiKey"
 done
+
+# Salesforce
+aws secretsmanager get-secret-value --secret-id arn:aws:secretsmanager:eu-west-1:865473395570:secret:DEV/Salesforce/User/SupportServiceLambdas | jq -c '[.SecretString | fromjson]' | \
+jq -M -r '.[] | .url, .client_id, .client_secret, .username, .password, .token' | \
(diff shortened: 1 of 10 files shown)


david-pepper:
[CSR Product Switch Wizard - Receive and store CSR Name and Case Number on Switch](https://trello.com/c/8WMWYrGO/3944-csr-product-switch-wizard-receive-and-store-csr-name-and-case-number-on-switch)

handlers/product-move-api/get-secrets-as-env-vars.sh
```diff
@@ -27,3 +27,19 @@ while read -r invoicingApiKey; read -r invoicingApiUrl; do
   #  https://stackoverflow.com/questions/32760584/in-intellij-how-do-i-set-default-environment-variables-for-new-test-configurati
   echo "invoicingApiUrl=$invoicingApiUrl;invoicingApiKey=$invoicingApiKey"
 done
+
+# Salesforce
+aws secretsmanager get-secret-value --secret-id arn:aws:secretsmanager:eu-west-1:865473395570:secret:DEV/Salesforce/User/SupportServiceLambdas | jq -c '[.SecretString | fromjson]' | \
```
david-pepper, line 32: Are these secrets being retrieved dynamically at run time, or are they retrieved once at the point of stack creation and then stored in env vars on the lambda?

david-pepper:
All 3 values are being set as expected in Salesforce:

image

david-pepper (approved):
Looks great, thanks Rupert

handlers/product-move-api/get-secrets-as-env-vars.sh
rupertbates: This script is just for use locally to enable us to run integration tests which need these secrets. The lambdas get them from env vars as you say.

PR #1897, opened 2023-03-14
Switching batch-email-sender to use a general-purpose braze-comms queue.

pawel-krupinski-guardian:


rupertbates (approved):

pawel-krupinski-guardian:
> What is the roll-out plan for this then, replace queues one by one and then decommission them all? Also, nice improvement on exporting the ARN of the queue in the other PR and importing it here, I didn't know that was possible, which is why I referenced resources directly by ARN.

Yes, replace all on the sending side (still couple places left after this change), then remove consumers in workflow, then delete the queues.

PR #1909, opened 2023-03-23
Supporter plus V2 product switching

rupertbates:
## What does this change?
This PR  enables product switching from a recurring contribution onto supporter plus V2. 

V2 changes the original supporter plus product by adding two separate charges to the product rate plan, one for the cost of the subscription which is fixed and an additional contribution charge which is variable and may be zero.

For product switching from a recurring contribution, this means that we need to work out how much of the amount which the user has chosen to pay is above the threshold so that we can treat that as a contribution (add it to the contribution charge in Zuora)

There is a new boolean flag added `SubscriptionUpdate.SwitchToV2SupporterPlus` which determines whether to switch to V1 or V2, this is currently set to false so that we can merge these changes without making the code active.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -50,8 +50,8 @@ lazy val scalafmtSettings = Seq(
 
 // fixme this whole file needs splitting down appropriately
 
-lazy val EffectsTest = config("effectsTest") extend (Test) describedAs ("run the edge tests")
-lazy val HealthCheckTest = config("healthCheck") extend (Test) describedAs ("run the health checks against prod/code")
+lazy val EffectsTest = config("effectsTest") extend Test describedAs "run the edge tests"
+lazy val HealthCheckTest = config("healthCheck") extend Test describedAs "run the health checks against prod/code"
 val testSettings = inConfig(EffectsTest)(Defaults.testTasks) ++ inConfig(HealthCheckTest)(Defaults.testTasks) ++ Seq(
   Test / testOptions += Tests.Argument("-l", "com.gu.test.EffectsTest"),
   Test / testOptions += Tests.Argument("-l", "com.gu.test.HealthCheck"),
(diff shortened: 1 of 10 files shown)


build.sbt
```diff
@@ -553,7 +559,7 @@ lazy val `product-move-api` = lambdaProject(
       s"aws lambda update-function-code --function-name product-switch-salesforce-tracking-$stage --s3-bucket $s3Bucket --s3-key $s3Path --profile membership --region eu-west-1".!!
     }
   }
-  .dependsOn(`zuora-models`)
+  .dependsOn(`zuora-models`, `new-product-api`)
```
rupertbates, line 562: new product api already has the pricing information which we need for working out contribution amounts

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/move/ProductMoveEndpoint.scala
```diff
@@ -168,14 +175,21 @@ object ProductMoveEndpoint {
         currentRatePlan.ratePlanCharges,
         s"Subscription: $subscriptionName has more than one ratePlanCharge",
       )
+      currency <- ZIO
+        .fromOption(ratePlanCharge.currencyObject)
+        .orElseFail(
+          s"Missing or unknown currency ${ratePlanCharge.currency} on rate plan charge in rate plan ${currentRatePlan.id} ",
+        )
```
rupertbates, line 182: To get pricing information we need to know the currency the user is paying in

handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/SubscriptionUpdate.scala
```diff
@@ -20,21 +20,24 @@ import sttp.client3.ziojson.*
 import sttp.model.Uri
 import zio.json.*
 import zio.{Clock, IO, RIO, Task, UIO, URLayer, ZIO, ZLayer}
-import com.gu.util.config
 
 import java.time.LocalDate
+
+val SwitchToV2SupporterPlus = false
```
rupertbates, line 26: Leave this set to false until we are ready to put V2 live

handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/SubscriptionUpdate.scala
```diff
@@ -140,60 +155,133 @@ case class SubscriptionUpdateInvoice(
     taxAmount: BigDecimal,
     invoiceItems: List[SubscriptionUpdateInvoiceItem],
 )
-case class SupporterPlusRatePlanIds(ratePlanId: String, ratePlanChargeId: String)
+case class SupporterPlusRatePlanIds(
+    ratePlanId: String,
+    subscriptionRatePlanChargeId: String,
+    contributionRatePlanChargeId: Option[String],
+)
```
rupertbates, line 162: We will have a rate plan charge id for the contribution charge if the `SwitchToV2SupporterPlus` flag is true otherwise we won't, this can change to a String once we remove the V1 code

build.sbt
johnduffell: this causes a load of extra "logback" logging spam as it complains there are two "logback.xml" on the classpath.
we need to extract the relevant file into `lib` rather than pulling in another handler as a dependency

PR #1949, opened 2023-05-01
Use Secrets Manager client at contact_us_api

shtukas:
We recently became aware of the fact that instructions of the type 

```
'{{resolve:secretsmanager:${SalesforceStage}/Salesforce/ConnectedApp/${AppName}:SecretString:clientId::${AppSecretsVersion}}}'
```

are leaking the values it resolves. The migration from resolving them from the Cloud Formation to the running app using the AWS Secrets Manager client has already been done for the price migration engine: https://github.com/guardian/price-migration-engine/pull/852

The other system owned by Engine that is leaking is support-service-lambdas. In this case the situation is more difficult because support-service-lambdas is a large collection of lambdas and 9 of them are leaking. In this PR, we migrate the first: `contact-us-api`.

The approach we are using is very similar to the one we used for the engine. Introducing types (case classes) to represent the JSON data stored in Secrets Manager. Extract that data and make it available where it is needed, which in essence means removing the use of the environments variables, and once that is done we can remove the lines from the cloud formation.


gnm-notona (approved):
Nice work!

PR #1950, opened 2023-05-02
Use Secrets Manager at digital-voucher-suspension-processor

shtukas:
This change is the second stage of the ongoing effort to prevent credentials leaks from support-service-lambdas. Here we are updating digital-voucher-suspension-processor. 

For a full context of that change see the first PR description: https://github.com/guardian/support-service-lambdas/pull/1949

Notes. 

1. An interesting difference between the first PR and this one is that in the first one, we modified `ContactUsConfig` and replaced its use with `SalesforceSecrets` and `MembersDataAPISecrets`. Whereas in this case this would have led to a much deeper and much more time consuming refactoring. Instead, here,  we use `Secrets` to build the same `Config`. Also in this case we have `ImovoSecrets` as third source of secrets (which corresponds to a third JSON string/object in AWS Secrets Manager)

2. Once the entire process of updating all the lambdas is completed, there might be an opportunity for a global refactoring across  support-service-lambdas and remove code duplication across the lambdas (it looks like they are reading from the same small set of secrets). Code duplication is currently happening because we are updating them in independent PRs.

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -71,7 +71,7 @@ object Dependencies {
   val tapirVersion = "1.4.0"
   val enumeratum = "com.beachape" %% "enumeratum" % "1.7.2"
   val scalaXml = "org.scala-lang.modules" %% "scala-xml" % "2.1.0"
-  val stripe = "com.stripe" % "stripe-java" % "22.18.0"
+  val stripe = "com.stripe" % "stripe-java" % "22.20.0"
   val parallelCollections = "org.scala-lang.modules" %% "scala-parallel-collections" % "1.0.4"
 
   // Testing


gnm-notona (approved):

graham228221:
[SECURITY: Stop resolving credentials via cfn in sf-billing-account-remover](https://trello.com/c/rG4ZJZcY/158-security-stop-resolving-credentials-via-cfn-in-sf-billing-account-remover)

graham228221:
> there might be an opportunity for a global refactoring across  support-service-lambdas and remove code duplication across the lambdas (it looks like they are reading from the same small set of secrets). 

@shtukas just flagging that this depends somewhat on the third party system - for the i-movo API we'll have a single set of credentials across all our systems, but for Salesforce we aim to have a discreet set of credentials for each Lambda (each with its own permissions/access). Zuora I'd imagine is probably a bit of halfway house.

shtukas:
Thanks @graham228221 That makes sense. I wasn't necessarily referring to the exact secrets themselves (although I did notice that credentials are sometimes shared from one lambda to another -- something we can improve later), but mostly the Secrets object and its functionality, which can be abstracted :)

graham228221:
ah, I see now, that makes sense - thanks @shtukas!

shtukas:
Three successful secrets retrieval.

Screenshot 2023-05-19 at 15 41 42

PR #1954, opened 2023-05-05
Mj/introduce gu cdk for new product api

michaelbjacobson:


.github/workflows/ci.yml
```diff
@@ -6,61 +6,110 @@ on:
...
           cache: 'sbt'
-      - name: Build
-        run: |
-          LAST_TEAMCITY_BUILD=6410
-          export GITHUB_RUN_NUMBER=$(( $GITHUB_RUN_NUMBER + $LAST_TEAMCITY_BUILD ))
-          sbt "project ${{ matrix.subproject }}" riffRaffUpload
\ No newline at end of file
+
+      - name: Build subproject JAR
+        run: sbt "project ${{ matrix.subproject }}" assembly
```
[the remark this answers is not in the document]
jacobwinch: I'm also a bit unsure how the correct build number is set now that we aren't setting `env` vars - would you mind explaining that?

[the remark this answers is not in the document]
michaelbjacobson: @jacobwinch good shout about manually adding the tests, I'll do that now. Re. the build number, my understanding is that it's set by the Github Actions run number. I've now added an offset of 7000 (the old build number rounded up to the nearest thousand) to the RiffRaff upload action, so it will skip a few numbers but the latest build will always have the highest number assigned to it.

.github/workflows/ci.yml
```diff
@@ -6,6 +6,51 @@ on:
...
+        run: sbt "project ${{ matrix.subproject }}" test assembly
+
+      - name: Generate CFN templates from GuCDK
+        run: bash ./script/cfn.sh
+
+      - name: Upload to Riff-Raff
+        uses: guardian/actions-riff-raff@v2
+        with:
+          projectName: support-service-lambdas::${{ matrix.subproject }}
+          buildNumberOffset: 7000
```
michaelbjacobson, line 45: @rupertbates FYI

.github/workflows/ci.yml
```diff
@@ -6,6 +6,51 @@ on:
...
+    runs-on: ubuntu-latest
+    permissions:
+      id-token: write
+      contents: read
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@v1
```
akash1810, line 23: Would recommend using [v2](https://github.com/aws-actions/configure-aws-credentials/blob/main/CHANGELOG.md#200-2023-03-06) which uses Node 16 (an LTS version) following https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/.

script/cfn.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+
+set -e
+set -x
+
+DIR=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
+ROOT_DIR="${DIR}/.."
+cd "$ROOT_DIR/cdk"
+
+yarn install
```
akash1810, line 10: Recommend the [`--frozen-lockfile` flag](https://classic.yarnpkg.com/lang/en/docs/cli/install/#toc-yarn-install-frozen-lockfile) which makes CI more deterministic:

> Running yarn with no command will run yarn install, passing through any provided flags.
> 
> If you need reproducible dependencies, which is usually the case with the continuous integration systems, you should pass --frozen-lockfile flag.

michaelbjacobson:
Thanks @akash1810, I've implemented those recommendations now, but I've limited the changes to the new 'best practice' build job.

michaelbjacobson:
[New Product API GuCDK Migration - Step 1](https://trello.com/c/5qCaGcXF/4147-new-product-api-gucdk-migration-step-1)

.github/workflows/ci.yml
johnduffell: this isn't quite right as you can see below, LAST_TEAMCITY_BUILD=6410
This means that for some reason the earlier builds go to the top
image

but the latest one actually has to be found in the build log and typed in manually
image

PR #1955, opened 2023-05-05
Get rid of DEV environment from AWS & point all CODE lambdas at the main Zuora and SF test environments

rupertbates:
## What does this change?
This PR removes any usage of the secondary Zuora and Salesforce test environment which is variously called CODE or UAT and points all CODE usages to the primary test environment which has sometimes been called DEV but from now on will be called CODE.

This diagram shows the new environments, their usages and their relationships with external systems:
Supporter Revenue Environments(3)

There are more details [in this document](https://docs.google.com/document/d/1wg1kMk7U7ht3AmPpsPUimtz1FkVbu5BWjijjme2598s/edit#)

### Changes
- Change config files in S3, parameter store and secrets manager to point CODE environment to our primary dev environments in Zuora and Salesforce. [This spreadsheet](https://docs.google.com/spreadsheets/d/1sIFGmwPtSIeT1cSWrAi9GkXYCvaAl3jIhX66veCHRaE/edit#gid=0) documents the changes.
- Change environment/stage classes to reflect the new definitions - DEV refers to local machine only, CODE is our AWS test environment and PROD is production.
- Update any readmes


shtukas:
Amazing 🎉. Thank you!

PR #1980, opened 2023-05-19
Identity retention Lambda uses BigQuery to find active products

gnm-notona:
## What does this change?
Uses a BigQuery query to get the product holdings instead of querying Zuora directly.  This means we can include other products - stripe patrons, eventbrite, mobile apps and single contributions - in the decision about whether to retain an identity_id.

Also extends the existing response with additional fields: `effectiveDeletionDate` and `responseValidUntil`.

See README.md for details of the new response format. 

https://trello.com/c/YQ1p8GC5/192-identity-retention-api-use-bigquery-to-determine-deletion-date


shtukas (approved):

shtukas:
@raphaelkabo Merging now. FYI.

PR #1952, opened 2023-05-05
Use Secrets Manager at holiday-stop-api

shtukas:
This change is the third of the ongoing effort to prevent credentials leaks from support-service-lambdas. Here we are updating holiday-stop-api. 

For a full context of that change see the first PR description: https://github.com/guardian/support-service-lambdas/pull/1949

PR #1995, opened 2023-05-23
Recover Stage as Environment variable

shtukas:
### holiday-stop-processor

When I made this change https://github.com/guardian/support-service-lambdas/pull/1953 , I indicated in the PR description that the fragment of the CF I was removing is not used. That was only partially true.

`InvoicingApiUrl` and `InvoicingApiKey` are definitively not used in that Lambda and the opinion that they might have been the result of copy pasting or decommission is maybe true. 

On the other hand, the `Stage` although not used "directly" seems to be read by `com.gu.util.config.Stage` which is a dependency. 

This was not detected in CODE testing (where the default CODE value was used), but only in PROD. 

### holiday-stop-api
https://github.com/guardian/support-service-lambdas/pull/1952 suffered from the same problem, so we fixed it as well.


rupertbates (approved):
👍

PR #2008, opened 2023-06-20
New switch campaign for membership to recurring contribution

jonflynng:
Includes new case class with required data points.


rupertbates (approved):
Put a more detailed description in so that if you come back to this PR in isolation you know what it is doing and why.

PR #2011, opened 2023-06-30
Remove redundant price check

jonflynng:
We are calculating the amount the user is currently paying on the server-side but it is also being passed in as the `price` property from the frontend. This removes the calculation and just uses the property passed from the frontend.


handlers/product-move-api/src/main/scala/com/gu/productmove/switchtype/ToRecurringContribution.scala
```diff
@@ -56,14 +56,6 @@ object ToRecurringContribution {
         )
       (activeRatePlan, activeRatePlanCharge) = activeRatePlanAndCharge
 
-      currency <- ZIO
```
jonflynng, line 59: currency was not being used

shtukas (approved):

PR #2027, opened 2023-07-25
Update supporter product data from move subscription fix lambda

rupertbates:
## What does this change?
Currently the sf-move-subscription lambda, which reassigns subscriptions from one identity account to another, does not update the SupporterProductData dynamo table which drives the members-data-api attribute service. This means that even though a subscription has been reassigned the new user does not get the benefit of it.

This PR fixes that issue by adding a new record into SupporterProductData when the switch occurs. The way it  does this is by adding an item to the supporter-product-data SQS queue which is then processed by a lambda in the support-frontend repository. 

Currently we do not delete the old record on switch, I don't think this is a problem as the fact that the user is asking to switch is usually because they do not use the other account and even if they do end up using digital benefits on two accounts this is not a huge problem and will anyway only be until the next renewal date. 

If we do decide that we should remove the old record we can do this in a future PR

### DIFF
--- a/handlers/sf-move-subscriptions-api/src/main/scala/com/gu/sf/move/subscriptions/api/UpdateSupporterProductDataService.scala
+++ b/handlers/sf-move-subscriptions-api/src/main/scala/com/gu/sf/move/subscriptions/api/UpdateSupporterProductDataService.scala
@@ -13,20 +13,8 @@ trait UpdateSupporterProductData {
 class UpdateSupporterProductDataService(queueName: QueueName) extends UpdateSupporterProductData with LazyLogging {
 
   def combineErrorsOrUnit(list: List[Either[String, Unit]]): Either[String, Unit] = {
-    list.foldLeft[Either[String, Unit]](Right(())) { (acc, either) =>
-      acc match {
-        case Left(errors) =>
-          either match {
-            case Left(error) => Left(s"$errors, $error")
-            case Right(_) => acc
-          }
-        case Right(_) =>
-          either match {
-            case Left(error) => Left(s"$error")
-            case Right(_) => acc
-          }
-      }
-    }
+    val errors = list.collect { case Left(error) => error }
+    if (errors.isEmpty) Right(()) else Left(errors.mkString(", "))
   }
   override def update(
       subscription: Subscription,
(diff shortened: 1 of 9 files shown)


handlers/sf-move-subscriptions-api/src/main/scala/com/gu/sf/move/subscriptions/api/UpdateSupporterProductDataService.scala
```diff
@@ -0,0 +1,59 @@
...
+            case Left(error) => Left(s"$errors, $error")
+            case Right(_) => acc
+          }
+        case Right(_) =>
+          either match {
+            case Left(error) => Left(s"$error")
+            case Right(_) => acc
+          }
+      }
+    }
```
johnduffell, line 29: I assume you're trying to avoid cats/implicits here -
haven't compiled but something like this would be more in line with the intention
```suggestion
    val errors = list.collect { case Left(error) => error }
    if (errors.isEmpty) Right(()) else Left(errors.mkString(", "))
```

handlers/sf-move-subscriptions-api/src/main/scala/com/gu/sf/move/subscriptions/api/UpdateSupporterProductDataService.scala
```diff
@@ -0,0 +1,59 @@
...
+          ratePlan.productRatePlanId,
+          ratePlan.ratePlanName,
+          subscription.termEndDate,
+          subscription.contractEffectiveDate,
+        ),
+      )
+      .map(item => Payload(item.asJson.noSpaces))
+      .map(payload => sqsSync(payload).toEither.left.map(_.getMessage))
+
+    combineErrorsOrUnit(events)
```
johnduffell, line 51: not sure if it's common to have more than one here with only one failing - if there are a few to send it won't be clear which one the error is from. consider having e.g.  `0: OK, 1: IllegalArgumentException, 2: CredentialException, 3: OK, 4: OK` so we can debug?

johnduffell: on the flip side, if it's usually going to be a test with no credentials, having a page of errors to tell you that, fail fast could be better (given I don't think we're running in parallel)

johnduffell (approved):
looks good a couple of comments though. 
Also no new tests or metric/alarms - do we need to be alerted if this stops working, other than people writing in to userhelp to let us know?

rupertbates:
> looks good a couple of comments though. Also no new tests or metric/alarms - do we need to be alerted if this stops working, other than people writing in to userhelp to let us know?

This is only called from Salesforce I believe so a CSR would raise a ticket for a failure

PR #1957, opened 2023-05-11
Use Secrets Manager at product-move-api

shtukas:
This change is part of the ongoing effort to prevent credentials leaks from support-service-lambdas. Here we are updating product-move-api. 

For a full context of that change see the first PR description: https://github.com/guardian/support-service-lambdas/pull/1949 

Note that this only updates `move-product-${Stage}`, we will do `product-switch-salesforce-tracking-${Stage}` in another PR.

PR #2029, opened 2023-07-27
Fix Refund lambda secretsmanager access

graham228221:
Gives the product-switch-refund Lambda secrets manager access so that it can communicate with the invoicing-api

After fixing access, the Lambda would then fail with another error: `Failure while parsing json string: upickle.core.AbortException: missing keys in dictionary: invoicingApiUrl, invoicingApiKey at index 140` - This was due to the secrets manager entry having capitalised keys (`InvoicingApiUrl` and `InvoicingApiKey`).

Note that we don't have any kind of health check or automated test that would spot configuration issues such as this, so this has been missing since this PR https://github.com/guardian/support-service-lambdas/pull/1957/files#diff-5da588a0a6da3f0826731e2fd10146bf806da88d2308953ed5bb8a0346297ad2L152 - it was only spotted after a colleague complained that they didn't receive their refund

### DIFF
--- a/handlers/product-move-api/cfn.yaml
+++ b/handlers/product-move-api/cfn.yaml
@@ -155,6 +155,14 @@ Resources:
             - sqs:GetQueueAttributes
             - sqs:ReceiveMessage
           Resource: "*"
+      - Statement:
+        - Effect: Allow
+          Action:
+            - secretsmanager:DescribeSecret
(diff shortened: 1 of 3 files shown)


johnduffell (approved):
thanks for doing this! 👍

shtukas (approved):

PR #2034, opened 2023-08-01
Product move refund tax

rupertbates:
## What does this change?
The product move refund functionality had incorrect logic for working out the amount to refund as it was not taking account of any tax paid, meaning that customers would be refunded less than they originally paid.

This PR fixes that bug, however the changes are quite extensive because of the affect on the invoice item adjustments that are made after the refund. 

To adjust the tax portion of an invoice item we need a separate invoice item adjustment with the source type of 'Tax' (rather than 'InvoiceDetail') and the source id of the taxation item associated with the invoice to be adjusted.
[API reference is here](https://www.zuora.com/developer/api-references/older-api/operation/Object_POSTInvoiceItemAdjustment/#!path=SourceType&t=request)

Unfortunately this means that we have to make an additional call to Zuora to query the details of the taxation item.

I have also added a lot of tests for various parts of the logic as it is much more complex than I'd like!


graham228221 (approved):

handlers/product-move-api/src/main/scala/com/gu/productmove/Secrets.scala
```diff
@@ -88,7 +88,7 @@ object SecretsLive extends Secrets {
   }
 
   def getStage: ZIO[Any, ErrorResponse, String] =
-    ZIO.fromOption(sys.env.get("Stage")).mapError(_ => SecretsError("Failure while extracting stage from environment"))
+    ZIO.succeed(sys.env.getOrElse("Stage", "CODE"))
```
rupertbates, line 91: Providing a default here makes it easier to run integration tests locally

PR #2035, opened 2023-08-01
explicitly define 5XX alarm and set more accurate alarm name

michaelbjacobson:
This explicitly creates the 5XX alarm, as opposed to using the out-of-the-box configuration from GuCDK, giving us the ability to adjust the name of the alarm to better reflect its function.


rupertbates (approved):
👍

PR #2036, opened 2023-08-08
Align the S+ contribution charge with the subscription charge

rupertbates:
## What does this change?
There is currently a bug in the new-product-api when a user buys a supporter plus sub with Direct Debit - the subscription charge in the rate plan is set up with a customerAcceptanceDate of today + 10 days to allow for the DD to be set up before we charge it, whereas the contribution charge which is set up via a charge override with a trigger date of today. This means that the two charges are billed on different days.

There is an subscription with an example of this [here](https://www.zuora.com/platform/subscriptions/8a128be6899d0b8a0189b0a9bcc574e1)

This PR fixes this by setting the trigger date of the contribution charge to the same date as the customerAcceptanceDate.


rupertbates:
After more investigation I have tracked this down to a misconfiguration of the S+ contribution charge on the monthly product rate plan. The trigger condition on it was set to ‘Upon contract effective’ whereas the subscription charge has its trigger condition set correctly to ‘Upon customer acceptance’.
I have now fixed this misconfiguration so the issue will not reoccur.

PR #2023, opened 2023-07-10
Endpoint to update supporter plus amount

jonflynng:
An endpoint to update the Supporter Plus amount, for both V1 and V2.

Tested on CODE on V1 and V2 subs.

PR #2037, opened 2023-08-10
add validation and multi currency to supporter plus change amount

johnduffell:
This PR follows on from https://github.com/guardian/support-service-lambdas/pull/2023

Just some extra code to validate the amounts and also to make sure we subtract the right amount for the "fixed" base price.

Also included are some bonus changes due to warnings from intellij and tweaks to reduce the massive verbosity of the logs.  Check the inline comments for more detail.

TODO add some tests if necessary

trello https://trello.com/c/3utmrhbj/4204-mma-allow-supporter-to-change-value-above-threshold

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,4 +1,4 @@
-import Dependencies._
+import Dependencies.*
 
 val scala2Settings = Seq(
   ThisBuild / scalaVersion := "2.13.10",
@@ -365,7 +365,7 @@ lazy val `identity-retention` = lambdaProject(
       case "META-INF/kotlin-stdlib.kotlin_module" => MergeStrategy.discard
       case "META-INF/kotlin-stdlib-common.kotlin_module" => MergeStrategy.discard
       case x =>
-        val oldStrategy = (assemblyMergeStrategy).value
+        val oldStrategy = assemblyMergeStrategy.value
         oldStrategy(x)
     },
   )
@@ -584,8 +584,8 @@ lazy val `product-move-api` = lambdaProject(
     // product-move-api needs its own deploy task currently because firstly it's Scala 3 so the jar path is different to
     // other projects and secondly the jar is too large to deploy with the aws cli --zip-file parameter so we need to use S3
     deployTo := {
-      import scala.sys.process._
-      import complete.DefaultParsers._
+      import scala.sys.process.*
+      import complete.DefaultParsers.*
       val jarFile = assembly.value
 
       val Seq(stage) = spaceDelimited("<arg>").parsed
@@ -598,7 +598,7 @@ lazy val `product-move-api` = lambdaProject(
       s"aws lambda update-function-code --function-name product-switch-salesforce-tracking-$stage --s3-bucket $s3Bucket --s3-key $s3Path --profile membership --region eu-west-1".!!
     }
   }
-  .dependsOn(`zuora-models`, `new-product-api`)
+  .dependsOn(`zuora-models`, `effects-sqs`, `new-product-api`)
 
 lazy val `metric-push-api` =
   lambdaProject("metric-push-api", "HTTP API to push a metric to cloudwatch so we can alarm on errors")
@@ -726,8 +726,8 @@ lazy val `stripe-webhook-endpoints` = lambdaProject(
     2. run `deployTo CODE`
    */
   deployTo := {
-    import scala.sys.process._
-    import complete.DefaultParsers._
+    import scala.sys.process.*
+    import complete.DefaultParsers.*
     val jarFile = assembly.value
 
     val Seq(stage) = spaceDelimited("<arg>").parsed
@@ -766,8 +766,8 @@ lazy val deployAwsLambda =
     "Directly update AWS lambda code from your local machine instead of via RiffRaff for faster feedback loop",
   )
 deployAwsLambda := {
-  import scala.sys.process._
-  import complete.DefaultParsers._
+  import scala.sys.process.*
+  import complete.DefaultParsers.*
   val Seq(name, stage) = spaceDelimited("<arg>").parsed
   s"aws lambda update-function-code --function-name $name-$stage --zip-file fileb://handlers/$name/target/scala-2.13/$name.jar --profile membership --region eu-west-1".!
 }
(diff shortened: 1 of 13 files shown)


build.sbt
```diff
@@ -1,4 +1,4 @@
-import Dependencies._
+import Dependencies.*
```
johnduffell, line 1: was a warning

build.sbt
```diff
@@ -598,7 +598,7 @@ lazy val `product-move-api` = lambdaProject(
       s"aws lambda update-function-code --function-name product-switch-salesforce-tracking-$stage --s3-bucket $s3Bucket --s3-key $s3Path --profile membership --region eu-west-1".!!
     }
   }
-  .dependsOn(`zuora-models`, `new-product-api`)
+  .dependsOn(`zuora-models`, `effects-sqs`, `new-product-api`)
```
johnduffell, line 601: this was added because we need the queue name and were getting it as a transitive dependency.  We maybe shouldn't pull the whole thing in just for a getEnv call but that's a job for another day.

handlers/product-move-api/src/main/resources/logback.xml
```diff
@@ -7,7 +7,7 @@
 
     <!-- see https://docs.aws.amazon.com/sdk-for-java/v2/developer-guide/java-dg-logging.html -->
     <logger name="software.amazon.awssdk" level="WARN"/>
-    <logger name="software.amazon.awssdk.request" level="DEBUG"/>
+    <logger name="software.amazon.awssdk.request" level="WARN"/>
```
johnduffell, line 10: prevent a load of logging every time the AWS client makes a call.  Do we need this logging or not?  This goes into the JAR so may also affect when it's running in AWS.

handlers/product-move-api/src/main/scala/com/gu/productmove/SttpClientLive.scala
```diff
@@ -24,10 +24,11 @@ object SttpClientLive {
 
   private class SttpLogger extends Logger[Task] {
 
-    override def apply(level: logging.LogLevel, message: => String): Task[Unit] = ZIO.log("STTP Backend: " + message)
+    override def apply(level: logging.LogLevel, message: => String): Task[Unit] = 
+      ZIO.logDebug("STTP Backend: " + message)
```
johnduffell, line 28: change these to "debug" so we can switch them off by default when runnning locally.
it was logging the entire content of the http requests and responses to the console meaning any logging added manually was getting swamped.  The log level can be changed in the runTest of the LambdaEndpoint class.

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/updateamount/UpdateSupporterPlusAmountEndpoint.scala
```diff
@@ -154,123 +114,4 @@ object UpdateSupporterPlusAmountEndpoint {
       .tapEither(result => ZIO.log(s"OUTPUT: $subscriptionName: " + result))
   } yield Right(res)
 
-  private def getSupporterPlusCharge(
```
johnduffell, line 157: moved all this out to another file to make it clearer which is the external interface and which is the logic

handlers/product-move-api/src/main/scala/com/gu/productmove/framework/LambdaEndpoint.scala
```diff
@@ -25,19 +28,29 @@ object LambdaEndpoint {
...
+      case Exit.Failure(cause) =>
+        println(s"program failed with error")
+        val shortcause = cause.filter {
+          case _: Cause.Interrupt => includeInterruptedCause
+          case _ => true
+        }
+        println(shortcause.prettyPrint)
+        println("(Cause filtered for clarity)")
+
+    println("for more verbose logging change constants in LambdaEndpoint.scala")
```
johnduffell, line 53: when running the main method locally in intellij, there were huge amounts of logs, so this file includes some tweaks to get rid of almost all of it so we can focus on the actual logging done by the lambda.  I wasn't sure where to document it, so I stuck to this println as it's more likly to be read

handlers/product-move-api/src/main/scala/com/gu/productmove/switchtype/RecurringContributionToSupporterPlus.scala
```diff
@@ -274,8 +276,6 @@ object RecurringContributionToSupporterPlus {
         )
     }
 
-    given JsonDecoder[SubscriptionUpdatePreviewResponse] = DeriveJsonDecoder.gen[SubscriptionUpdatePreviewResponse]
```
johnduffell, line 277: these were duplicated in quite a few places.  I'm not sure if the update subscription function really should be generic on the return type, if we change it to be a fixed type we can get rid of all these implicits/givens in the application code.

handlers/product-move-api/src/main/resources/logback.xml
rupertbates: I'm not sure what this logging looks like. Is it likely to be useful in diagnosing errors?

handlers/product-move-api/src/main/scala/com/gu/productmove/SttpClientLive.scala
rupertbates: We should keep that request level logging when running in AWS, it's very useful

johnduffell: ok will make sure it's kept that way.  The logger run in AWS is set up to log without filtering the log levels so it should be unaffected https://github.com/guardian/support-service-lambdas/blob/1b7e0ee7131c016038bf2b2a25a26c86baf8836c/handlers/product-move-api/src/main/scala/com/gu/productmove/framework/ZIOApiGatewayRequestHandler.scala#L121

handlers/product-move-api/src/main/scala/com/gu/productmove/switchtype/RecurringContributionToSupporterPlus.scala
rupertbates: sounds sensible, if we don't need it to be generic let's simplify

rupertbates (approved):
With a few comments

PR #2038, opened 2023-08-13
dependency updates

shtukas:
Scala Steward attempted a PR ( https://github.com/guardian/support-service-lambdas/pull/1999 ) that is failing. Here we perform one update manually. 

- update support-internationalisation from 0.15 to 0.16

Note: It is surprising that Stewart was attempting to upgrade cats-effect from 2.5.5 to 3.5.0, considering that this is a major upgrade which is actually breaking.


paulbrown1982 (approved):
👍

PR #2040, opened 2023-08-14
Improve product move refund lambda alarm

rupertbates:
## What does this change?
Previously the only alarm for the refund lambda was on the dead letter queue of the queue which triggers it. However this is not particularly useful unless the dead letter queue is regularly cleared because once it goes into alarm it will stay there and no further notifications will be received.

This PR changes the alarm to trigger on lambda failure to ensure that we recieve a notification each time there is an error.

### DIFF
--- a/handlers/product-move-api/cfn.yaml
+++ b/handlers/product-move-api/cfn.yaml
@@ -263,12 +263,15 @@ Resources:
       Statistic: Sum
       Threshold: 1
       TreatMissingData: notBreaching
-      
+
   RefundLambdaFailureAlarm:
     Type: AWS::CloudWatch::Alarm
+    Condition: IsProd
     Properties:
-      AlarmName: LambdaFailureAlarm
-      AlarmDescription: Alarm triggered when the refund Lambda fails
+      AlarmActions:
+        - !Sub arn:aws:sns:${AWS::Region}:${AWS::AccountId}:retention-dev
+      AlarmName: !Sub An error in the refund lambda ${Stage}. Please check the logs to diagnose and redrive the product-switch-refund-dead-letter-${Stage} SQS queue when ready.
+      ComparisonOperator: GreaterThanOrEqualToThreshold
       Namespace: AWS/Lambda
       MetricName: Errors
       Dimensions:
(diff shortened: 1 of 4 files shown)


michaelbjacobson (approved):
👍🏼

handlers/product-move-api/cfn.yaml
```diff
@@ -272,16 +272,15 @@ Resources:
...
-      EvaluationPeriods: 1
-      MetricName: ApproximateNumberOfMessagesVisible
-      Namespace: AWS/SQS
+        - Name: FunctionName
+          Value: !Ref RefundLambda
       Statistic: Sum
+      Period: 60
+      EvaluationPeriods: 1
       Threshold: 1
-      TreatMissingData: notBreaching
```
johnduffell, line 284: missing data still is not breaching, not sure if it makes a difference here though https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/AlarmThatSendsEmail.html#alarms-and-missing-data

johnduffell (approved):

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/updateamount/UpdateSupporterPlusAmountEndpoint.scala
```diff
@@ -113,7 +113,9 @@ object UpdateSupporterPlusAmountEndpoint {
         .in(subscriptionNameCapture)
         .in(
           jsonBody[ExpectedInput].copy(info =
-            EndpointIO.Info.empty[ExpectedInput].copy(description = Some("Definition of the updated total amount of regular payment")),
+            EndpointIO.Info
+              .empty[ExpectedInput]
+              .copy(description = Some("Definition of the updated total amount of regular payment")),
```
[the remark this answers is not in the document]
rupertbates: scalafmt was failing in the commit hook without this

handlers/product-move-api/cfn.yaml
rupertbates: Yeah it doesn't seem to

PR #2041, opened 2023-08-15
Use invoice item date in adjustments

rupertbates:
## What does this change?
This fixes a bug found with the product move refund lambda

The bug is related to a time zone issue where if we try to run an invoice adjustment in the hour after midnight (during BST) the invoice adjustment date will be before the invoice date for the invoice we are trying to adjust. 
The reason is that the code which is building the invoice item adjustments is using UTC when it builds them.

In fact we don't need to generate a date in code for these adjustments at all, we can just use the date from the invoice itself so that is what I have done in this PR.

[Cloudwatch logs are here](https://eu-west-1.console.aws.amazon.com/cloudwatch/home?region=eu-west-1#logsV2:log-groups/log-group/$252Faws$252Flambda$252Finvoicing-api-refund-PROD/log-events/2023$252F08$252F14$252F$255B$2524LATEST$255D2707ed4a3f8f46cc940db7970c305bab)


johnduffell:
> There are some more details about the bug in this doc
can you copy them into the PR if possible so we can see them in future?

johnduffell (approved):
thanks 👍

PR #2046, opened 2023-08-31
National Failed Deliveries - generate json for composite request

david-pepper:
## Introduction
Generate json required to update delivery records and create cases in Salesforce via composite api callout*

_*integration with Salesforce is not part of these changes. It will follow in a subsequent PR._

## Background
- Processing failed deliveries is a required feature of the National Delivery project to begin delivering papers outside the M25

**Previous PRs**
#2045 Create Lambda for processing failed national deliveries

### Example of expected json
image

## Outline of Changes

image

- Add types.ts with new types: Request, SubRequest, Body
- Add RequestBuilder.ts with methods (+ helpers) to build requests
- Add FileIO.ts with methods to get + filter file contents

## Testing

**Generating the json**
Using a test csv file with the following contents:
image

I executed the logic and output the generated json (see example above).

**Verifying the Json body is constructed correctly**
I made a callout using Postman with the generated Json body to the following endpoint ```https://gnmtouchpoint--dev1.sandbox.my.salesforce.com/services/data/v58.0/composite```and verified that the outcomes were as expected:
- Delivery records were updated in Salesforce 
- Cases were created in Salesforce 
- SF_Subscription__c field on case record was updated 
- Case__c field on delivery record was updated 

### Coming next 
- Getting the file from S3


handlers/failed-national-deliveries-processor/package.json
```diff
@@ -8,5 +8,9 @@
   },
   "devDependencies": {
     "esbuild": "^0.19.2"
+  },
+  "dependencies": {
+    "csv-parse": "^5.4.1",
+    "csv-parser": "^3.0.0"
```
andrewHEguardian, line 14: can we remove this dependency if it's not being used? (looks like only `csv-parse` is)

handlers/failed-national-deliveries-processor/src/RequestBuilder.ts
```diff
@@ -0,0 +1,86 @@
+import { Request, SubRequest, FileRow } from "./types";
+
+export function generateRequestsFromFailedDeliveryRows(failedDeliveryRows : FileRow[]) : Request[] {
+	var requests : Request[] = [];
+		
+	failedDeliveryRows.forEach(
+		failedDeliveryRow => requests.push(generateRequest(failedDeliveryRow))
```
andrewHEguardian, line 7: this could be a `map` I think

david-pepper: Would using a map be beneficial?

handlers/failed-national-deliveries-processor/src/types.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export type SubRequest = {
+    method: string
+    url: string
+    referenceId: string
+    body?: Body
+}
+
+export type Body = {
+	Case__c?: string
```
andrewHEguardian, line 14: this doesn't particularly bother me but I noticed the tabbing here is slightly different to the above definitions - there might be a common guardian eslint-config (and tsconfig) you could use if you wanted.

(Ignore this for now)

handlers/failed-national-deliveries-processor/src/RequestBuilder.ts
andrewHEguardian: The benefits would be: gets rid of a `var` assignment, makes the whole function a one liner.
Style wise / how easily understood it is I'll leave up to you

david-pepper: YES! Done in this [commit](https://github.com/guardian/support-service-lambdas/pull/2046/commits/c8b7985bcea58cc4951e9664c634dfc2b7b79687), thanks for pairing with me on it

handlers/failed-national-deliveries-processor/src/types.ts
david-pepper: Weird - they are aligned in my IDE:

image

handlers/failed-national-deliveries-processor/package.json
david-pepper: Done

handlers/failed-national-deliveries-processor/src/RequestBuilder.ts
```diff
@@ -0,0 +1,80 @@
...
+
+function generateDeliveryUpdateUrl(compositeKey : string) : string {
+	return '/services/data/v58.0/sobjects/Delivery__c/Composite_Key__c/'+compositeKey;
+}
+
+function generateCompositeKey(subName : string, deliveryDate : string) : string {
+	const year = deliveryDate.substring(6,10);
+	const month = deliveryDate.substring(3,5);
+	const day = deliveryDate.substring(0,2);
+	return subName + '-' + year + '-' + month + '-' + day;
```
JoeMitchellGuardian, line 72: Why do you generate the composite key with dashes "-" and then replace them with underscores "_" any time this composite key is used? Would it not be easier to simply generate the composite key using underscores initially (so you don't need the removeInvalidCharsForReferenceId function?

handlers/failed-national-deliveries-processor/src/types.ts
andrewHEguardian: I think one is spaces and one is a tab, and github happens to not align them

handlers/failed-national-deliveries-processor/src/RequestBuilder.ts
david-pepper: They are not replaced **_every_** time the composite key is used. The composite key is used here in 2 ways - 

1. As a composite key to identify the record
2. As a part of the reference Id in the composite subrequest. 

The dashes are the standard for the composite key structure on Delivery__c records, and always has been, so:

- when composite key is being used as a composite key the dashes are present
- when composite key is being used as a component of the referenceId in the composite request, dashes are not allowed, so they are replaced with underscores

JoeMitchellGuardian: Ah ok, thank you for clarifying my understanding of this!

david-pepper:
I've made some updates to :

- add a helper class for low-level methods to build the Request
- add config class that holds the sf api version only once (used for url building), which should ease the load when version 58.0 becomes deprecated and needs updating

undergroundquizscene (approved):

handlers/failed-national-deliveries-processor/src/RequestBuilder.ts
```diff
@@ -0,0 +1,80 @@
...
+
+	compositeRequests.push(createJsonForPatchDeliveryWithNotes(fileRow.sub, fileRow['Delivery Date'], compositeKey))
+	compositeRequests.push(createJsonForGetSFSubFromDelivery(fileRow.sub, fileRow['Delivery Date'], compositeKey))
+	compositeRequests.push(createJsonForCreateCase(fileRow.sub, fileRow['Delivery Date'], compositeKey))
+	compositeRequests.push(createJsonForUpdateDeliveryWithCase(fileRow.sub, fileRow['Delivery Date'], compositeKey))
+	
+	return {
+		allOrNone:true, 
+		compositeRequest:compositeRequests
+	};
```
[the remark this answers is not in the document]
david-pepper: Yes you're right! It's a bit silly with all that overhead isn't it? :) [Thank you](https://github.com/guardian/support-service-lambdas/pull/2046/commits/a63c53c218d75186b362d03e94b3c99116aeed32)!

PR #2052, opened 2023-09-04
only look at active rateplans when allowing product move

johnduffell:
If a customer who has already switched, tries to preview or perform another switch, they will be rejected with `Subscription: ${subscriptionName.value} has more than one ratePlan`
However there is no reason to look at removed rate plans.

This PR changes it so that removed rate plans will be ignored.


rupertbates (approved):
Looks good but can we add a test for it?

johnduffell:
good idea on the test, I won't get a chance to add one imminently so it's either take on the tech debt or block the PR for now. (or someone else adds the test!)

PR #2056, opened 2023-09-21
Delivery records api refactor

johnduffell:
This PR refactors the delivery records API lambda.

I have a task to add national delivery support to this lambda.

Although the code overall was easy to follow, there was a lot of nesting, use of apply methods, and lack of package structure/mixing of loosely related code.

This PR is a pre-work opportunistic refactor to make sure it is easy to maintain going forwards.  It also helped me to get up to speed with the framework and codebase.

Putting this out for review separately because opinions may vary on how far to go tinkering with the code in advance.

There should be no functionality changes, all tests pass and TBC that it works the same in CODE


handlers/delivery-records-api/src/main/scala/com/gu/delivery_records_api/DeliveryRecordApiRoutes.scala
```diff
@@ -1,152 +1,126 @@
...
+import org.http4s.circe.CirceEntityEncoder._
+import org.http4s.dsl.Http4sDsl
+import org.http4s.{HttpRoutes, Request, Response}
+
+import java.time.LocalDate
 
 case class DeliveryRecordApiRoutesError(message: String)
 
-object DeliveryRecordApiRoutes {
+class DeliveryRecordApiRoutes[F[_]: Effect](createDeliveryProblemService: CreateDeliveryProblemService[F], getDeliveryRecordsService: GetDeliveryRecordsService[F]) {
```
johnduffell, line 20: reduce the nesting level - before, all the functions were nested inside an apply method, now they are merely nested in the class.

handlers/delivery-records-api/src/main/scala/com/gu/delivery_records_api/DeliveryRecordsApiApp.scala
```diff
@@ -18,33 +20,35 @@ final case class DeliveryRecordsApiError(message: String)
...
     for {
       salesforceClient <- SalesforceClient(sttpBackend, config)
         .leftMap(error => DeliveryRecordsApiError(error.toString))
-    } yield createLogging()(DeliveryRecordApiRoutes(DeliveryRecordsService(salesforceClient)))
-  }
-
-  def createLogging(): HttpRoutes[IO] => HttpRoutes[IO] = {
+    } yield {
+      val create = new CreateDeliveryProblemServiceImpl(salesforceClient)
+      val get = new GetDeliveryRecordsServiceImpl(salesforceClient)
```
johnduffell, line 40: here we have moved away from a `DeliveryRecordsService` that mixes the two operations, to separate implementations for each.  This is because there was little overlap between the two operations.

handlers/delivery-records-api/src/main/scala/com/gu/delivery_records_api/DeliveryRecordsApiApp.scala
```diff
@@ -18,33 +20,35 @@ final case class DeliveryRecordsApiError(message: String)
...
+      val routes = new DeliveryRecordApiRoutes(create, get).routes
+      addRequestLogging(routes)
+    }
+
+  private def addRequestLogging(routes: HttpRoutes[IO]): HttpRoutes[IO] =
```
johnduffell, line 45: renamed and made it into a "normal" rather than paramterless function

handlers/delivery-records-api/src/main/scala/com/gu/delivery_records_api/service/DeliveryRecordsServiceError.scala
```diff
@@ -0,0 +1,11 @@
+package com.gu.delivery_records_api.service
```
johnduffell, line 1: moved into the service package and split out into own file

handlers/delivery-records-api/src/main/scala/com/gu/delivery_records_api/service/createproblem/CreateDeliveryProblemService.scala
```diff
@@ -0,0 +1,46 @@
+package com.gu.delivery_records_api.service.createproblem
```
johnduffell, line 1: this file is only the interface and its data types

handlers/delivery-records-api/src/main/scala/com/gu/delivery_records_api/service/createproblem/CreateDeliveryProblemService.scala
```diff
@@ -0,0 +1,46 @@
...
+
+import java.time.LocalDate
+
+trait CreateDeliveryProblemService[F[_]] {
+
+  def createDeliveryProblemForSubscription(
+      subscriptionNumber: String,
+      contact: Contact,
+      detail: CreateDeliveryProblem,
+  ): EitherT[F, DeliveryRecordServiceError, Unit]
```
johnduffell, line 15: change this to Unit as the response wasn't used

handlers/delivery-records-api/src/main/scala/com/gu/delivery_records_api/service/createproblem/SFApiCreateDeliveryProblem.scala
```diff
@@ -9,6 +7,74 @@ import io.circe.Encoder
 import io.circe.generic.auto._
 import io.circe.syntax._
 
+import java.time.{LocalDate, LocalDateTime}
+
+object SFApiCreateDeliveryProblem {
```
johnduffell, line 12: moved this to the top, named the file and object the same, and renamed apply to create method.  This is the code that creates the SF request body object from the input data.

johnduffell: yeah I used to love the apply method, but @rupertbates was less in favour as it's just a bit surprising/harder to navigate.  I think in the end we settled upon using apply if it's a simple smart constructor, but not if it's actual logic.

As an aside, now that scala 3 has top level methods, we can have the best of both worlds - although only the most recently created lambdas are using scala 3.
We haven't migrated any existing over as far as I know, mainly because you had to choose your libraries carefully to get scala 3 macros.  I might have a play on 10% time as that's probably changed now.

PR #2058, opened 2023-09-25
new product api refactor

johnduffell:
Some comments from Emily on the previous PR https://github.com/guardian/support-service-lambdas/pull/2044#discussion_r1336022703
Basically to remove the "homeDelivery" prefix from classes that are wholly about home delivery, etc.  This was a hangover from the refactor where they were previously all in one long function.

I decided to push as a separate PR just to keep things simpler.

No functionality changes, and the tests all still pass.

### DIFF
--- a/handlers/new-product-api/src/main/scala/com/gu/newproduct/api/productcatalog/plans/DigitalPackPlans.scala
+++ b/handlers/new-product-api/src/main/scala/com/gu/newproduct/api/productcatalog/plans/DigitalPackPlans.scala
@@ -7,21 +7,18 @@ import java.time.LocalDate
 
 class DigitalPackPlans(today: LocalDate) {
 
-  private val DigiPackFreeTrialPeriodDays = 14
-  private val DigiPackStartDateWindowSize = WindowSizeDays(90)
+  private val FreeTrialPeriodDays = 14
 
-  private val digiPackWindowRule = WindowRule(
-    startDate = today.plusDays(DigiPackFreeTrialPeriodDays.toLong),
-    maybeSize = Some(DigiPackStartDateWindowSize),
-  )
-
-  private val digipackStartRules = StartDateRules(
-    windowRule = digiPackWindowRule,
+  private val startRules = StartDateRules(
+    windowRule = WindowRule(
+      startDate = today.plusDays(FreeTrialPeriodDays.toLong),
+      maybeSize = Some(WindowSizeDays(90)),
+    ),
   )
 
   val planInfo: List[(PlanId, PlanDescription, StartDateRules, BillingPeriod)] = List(
-    (DigipackAnnual, PlanDescription("Annual"), digipackStartRules, Annual),
-    (DigipackMonthly, PlanDescription("Monthly"), digipackStartRules, Monthly),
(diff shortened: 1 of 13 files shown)


handlers/new-product-api/src/main/scala/com/gu/newproduct/api/productcatalog/plans/DigitalVoucherPlans.scala
```diff
@@ -5,81 +5,77 @@ import com.gu.newproduct.api.productcatalog._
...
-  private val DigitalVoucherStartDateWindowSize = WindowSizeDays(1)
-
-  private def digitalVoucherStartDateRule(daysOfWeek: List[DayOfWeek]) =
+  private def startDateRule(daysOfWeek: List[DayOfWeek]) =
     StartDateRules(
       daysOfWeekRule = Some(DaysOfWeekRule(daysOfWeek)),
       windowRule = WindowRule(
         startDate = getStartDateFromFulfilmentFiles(ProductType.NewspaperDigitalVoucher, daysOfWeek),
-        maybeSize = Some(DigitalVoucherStartDateWindowSize),
+        maybeSize = Some(WindowSizeDays(1)),
```
johnduffell, line 16: window size inlined as the wrapper makes it obvious enough

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/productcatalog/plans/DigitalPackPlans.scala
```diff
@@ -7,21 +7,18 @@ import java.time.LocalDate
 
 class DigitalPackPlans(today: LocalDate) {
 
-  private val DigiPackFreeTrialPeriodDays = 14
-  private val DigiPackStartDateWindowSize = WindowSizeDays(90)
```
johnduffell, line 11: I inlined the window size because it's wrapped so it's clear, but not the free trial period as we don't actually wrap that.

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/productcatalog/plans/DigitalVoucherPlans.scala
```diff
@@ -5,81 +5,77 @@ import com.gu.newproduct.api.productcatalog._
 
 import java.time.{DayOfWeek, LocalDate}
 
-class DigitalVoucherPlans(
-  getStartDateFromFulfilmentFiles: (ProductType, List[DayOfWeek]) => LocalDate
-) {
+class DigitalVoucherPlans(getStartDateFromFulfilmentFiles: (ProductType, List[DayOfWeek]) => LocalDate) {
```
johnduffell, line 8: ideally could make a proper trait for `getStartDateFromFulfilmentFiles` rather than writing the function type out long hand.  Maybe for a later PR.

rupertbates (approved):
Got to say I still find it complicated to follow, but it's an improvement 👍

johnduffell:
Agree @rupertbates , perhaps refactoring is a process not a destination.  I'd like to switch out all the functional `blah => etc` to be more OO style traits instead, that would make it much easier to naviate in Intellij

PR #2059, opened 2023-09-26
Ad/single contribution record queues

AndreaDiotallevi:
## What are you doing in this PR?

- Add new Scala project `single-contributions-processor`
- Add new GuCDK stack `SingleContributionsProcessor` with the following resources:
  -  SQS queue (`single-contributions-processor-queue-CODE` / `single-contributions-processor-queue-PROD`) to store single contribution messages coming via support-frontend
  - Dead Letter Queue (`dead-letters-single-contributions-processor-queue-CODE` / `dead-letters-single-contributions-processor-queue-PROD`) to handle messages that cannot be successfully processed by the normal processing flow
- Add project to the existing CI build pipeline

[**Trello Card**](https://trello.com/c/RbcNlxbD/334-single-contribution-record-epic-create-sqs-main-queue-and-dead-letter-queue)

## Why are you doing this?

This is part 2 of the epic [Create a single contribution record in Salesforce via support-payment-api](https://trello.com/c/qibQVaR2/295-create-a-single-contribution-record-in-salesforce-via-support-frontend-payment-api).

The pull request for part 1 is [here](https://github.com/guardian/support-frontend/pull/5283).

## Architecture

1. After a single contribution is made via support-frontend, a message is sent to the SQS queue

2. The message payload is:

```scala
case class Message(
     contributionData: ContributionData,
)
```
3. A lambda polls the SQS queue and creates a single contribution record in Salesforce (lambda not yet implemented):

single-contribution-cropped-2

## Notes

- The SQS queues are encrypted using SQS-managed encryption keys (SSE-SQS):

```typescript
import { Queue, QueueEncryption } from 'aws-cdk-lib/aws-sqs';

new Queue(this, `${APP_NAME}-queue`, {
	encryption: QueueEncryption.SQS_MANAGED,
	...otherProperties,
});
```

## How to test

For testing refer to the previous pull request description [here](https://github.com/guardian/support-frontend/pull/5283).

## Next steps

Create a lambda function that polls the message and creates a single contribution record in Salesforce.


handlers/single-contributions-processor/.gitignore
```diff
@@ -0,0 +1,2 @@
+/.bsp/
+target/
```
johnduffell, line 2: I'm not 100% up on gitignore practices - do we need the gitignore at this level and as well as the top level?

build.sbt
```diff
@@ -383,6 +383,12 @@ lazy val `new-product-api` = lambdaProject(
   )
   .dependsOn(zuora, handler, `effects-sqs`, effectsDepIncludingTestFolder, testDep, `zuora-models`, `config-core`)
 
+lazy val `single-contributions-processor` = lambdaProject(
+  "single-contributions-processor",
+  "Create a single contribution record in Salesforce",
+  Seq(awsSecretsManager, circe, circeParser, scalajHttp, upickle),
```
johnduffell, line 389: I'm not sure if this list is copy and paste, or your intention, but as a point of interest I've been looking at scala toolkit, which is supposed to take the decision making out of choosing all the usual stuff (http librari, json deserialiser, test framework) and just give a standard out of the box experience.

I am sure there will be plenty more chances to try it on new lambdas, but it would be good to know if it's a convention the team could get behind (as compared with a different style for each lambda). So feel free to take a look or go with your current intention.

The link is here https://docs.scala-lang.org/toolkit/introduction.html

johnduffell (approved):
looks good and thanks for breaking things down nicely as it's much easier to review.

I've added some comments, some are suggestions and some are more for your interest.

I'm not totally up to speed with CDK but if it produces the right thing it can't be too far wrong!   I was a bit surprised about the encryption, I thought always that SQS was encrypted at rest and transit by default.

handlers/single-contributions-processor/.gitignore
AndreaDiotallevi: Good point.

For context, that file was generated by this command I ran to create a new project:
`sbt new scala/scala-seed.g8`

I was meant to ask at some point, what is the best way to create a new Scala sub-project - with a `sbt` command or manually?

I checked and that `.gitignore` file doesn’t have any effect anyways.

I have deleted it [here](https://github.com/guardian/support-service-lambdas/commit/6deb9a8e7e30b86c40f520c8bf8d702e61946769).

handlers/single-contributions-processor/riff-raff.yaml
```diff
@@ -0,0 +1,15 @@
...
+  - CODE
+  - PROD
+deployments:
+  single-contributions-processor-cloudformation:
+    type: cloud-formation
+    app: single-contributions-processor
+    parameters:
+      templateStagePaths:
+        CODE: single-contributions-processor-CODE.template.json
+        PROD: single-contributions-processor-PROD.template.json
```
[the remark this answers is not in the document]
AndreaDiotallevi: Interesting question.

I have checked a few stacks in the membership account (`membership-CODE-soft-opt-in-consent-setter`, `contributions-store-queue-CODE`) and they all seem to be deployed in one stack.

This would also be my preferred approach for a few reasons:
- simplicity: the resources are defined in one place so there is less cognitive load involved
- SRP: they change for the same reason since the SQS queue is a single purpose queue and the lambda’s only job is to poll the queue
- atomic updates: when you update the CloudFormation stack, it can make atomic updates to all resources defined in the template, ensuring consistency (if for instance you change the queue name, you don’t have the problem of having to update the name in a different CloudFormation file)
- CloudFormation updates only what has changed: so if you have to update only the lambda, the deployment is not gonna take much longer than having a separate template file

If the app becomes complex, we can always split them in separate stacks, but for now I can’t see any advantage.

Another option would be to use nested stacks. For instance, you have a main stack and nested stacks like:
- Queues
- Functions
- Alarms
- Roles

At the same time though, I am not sure the complexity this brings is justified at this point.

I am interested in knowing your thoughts?

build.sbt
AndreaDiotallevi: Yes, this is a copy-paste from another project as a placeholder.

I like conventions and I much prefer to have a standard for all lambdas to follow, so I like this idea.

I didn’t know this tool you mentioned - the Scala toolkit. I’ll have a read. And it was interested to see the demo earlier today during 10%. Let me know when you get to a point with your experimentation when we can test it!

AndreaDiotallevi:
Thank you John for going through the PR and for the useful comments!

> sounds good I would just make the point that processor is a really general word, if I understand correctly, actually it's not processing the contribution itself it's just putting a record in salesforce? Maybe call it something like `single-contribution-salesforce-writes` ?

That’s a very good point. In fact, I wasn’t happy with the final name I chose. I like your suggestion `single-contribution-salesforce-writes`. I have updated. I will also update the other pull request to match the queue name.

> I'm not totally up to speed with CDK but if it produces the right thing it can't be too far wrong! I was a bit surprised about the encryption, I thought always that SQS was encrypted at rest and transit by default.

I had noticed that most of the projects in this repo use CloudFormation templates and only one project uses the GuCDK. The only reason I used the CDK is because I thought that was the “latest” / “most recent” way of doing IAC. If you think we should keep it consistent with the majority of the projects in this mono-repo, I am happy to convert it.

Regarding the SQS encryption, I have checked the docs and you are right. Since October 2022, every new SQS is encrypted by default using SSE-SQS keys. Link [here](https://aws.amazon.com/blogs/compute/announcing-server-side-encryption-with-amazon-simple-queue-service-managed-encryption-keys-sse-sqs-by-default/). So I have updated the code by removing the explicit definition for encryption. To note, this change is not retroactive, so queues created before that time will need to be updated.

build.sbt
johnduffell: sure sounds good 👍 don't let that get in the way of your current tasks

johnduffell (approved):
sounds good to me! 👍 let me know if you need anything else

PR #2065, opened 2023-10-02
don't delay DD subs by 10 days

johnduffell:
(don't merge until after new product api refactor PR is approved and merged https://github.com/guardian/support-service-lambdas/pull/2058 )

We have always had a 10 day delay on DD contributions via new product-api [1] .  However this is not technically needed under the DD guarantee as it's a specifically agreed payment rather than an updated recurring amount.

This code was copied into the Supporter plus code and it was spotted during testing as an incorrect "fulfilment start date" in salesforce.

I checked with Matt and he didn't see a reason for the delay.  The PR [2] implicied you might have suggested it originally @paulbrown1982 so just adding in case you have an opinion.

[1] https://github.com/guardian/support-service-lambdas/blob/6417309fd415126e9f349b081edd27e63a9ef006/handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/Handler.scala#L53
[2] https://github.com/guardian/support-service-lambdas/pull/164/files#diff-bae1d5b1d39da9219c297afd04d53a576bae5713674d9ce3e4fdb6f21bf64583R53

### DIFF
--- a/handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/AddContribution.scala
+++ b/handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/AddContribution.scala
@@ -49,7 +49,7 @@ object AddContribution {
         request.planId,
         account.currency,
       ).toApiGatewayOp.toAsync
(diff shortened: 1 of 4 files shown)


rupertbates (approved):
👍

PR #2068, opened 2023-10-05
Add Effects tests for delivery records api national delivery and version bumps

johnduffell:
As part of the national delivery work, we need to make sure the delivery records api can support "National delivery"  devliery records in salesforce.

This PR adds some effects tests to make sure it's working.  Ideally they would create the national delivery sub and the delivery record as part of the test, so that it won't break later.  I will add a ticket to do that in trello.

It also involves some small version bumps and tidy up along the way.

The diffx API has changed a bit in the version bump so I have fixed up the compile in other modules that use it.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,7 +1,7 @@
 import Dependencies.*
 
 val scala2Settings = Seq(
-  ThisBuild / scalaVersion := "2.13.10",
+  ThisBuild / scalaVersion := "2.13.12",
   version := "0.0.1",
   organization := "com.gu",
   scalacOptions ++= Seq(
@@ -19,13 +19,14 @@ val scala2Settings = Seq(
     "-Ywarn-dead-code",
     "-Ywarn-numeric-widen",
     "-Ywarn-value-discard",
+    "-Ytasty-reader",
   ),
   Test / fork := true,
   autoCompilerPlugins := true,
 )
 
 val scala3Settings = Seq(
-  scalaVersion := "3.2.2",
+  scalaVersion := "3.3.1",
   version := "0.0.1",
   organization := "com.gu",
   scalacOptions ++= Seq(
@@ -216,7 +217,7 @@ lazy val `effects-lambda` = library(project in file("lib/effects-lambda"))
     dependencyOverrides ++= jacksonDependencies,
   )
 
-lazy val `config-core` = library(project in file("lib/config-core"))
+lazy val `config-core` = library(project in file("lib/config-core"), scala3Settings)
 
 lazy val `config-cats` = library(project in file("lib/config-cats"))
   .settings(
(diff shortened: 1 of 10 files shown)


build.sbt
```diff
@@ -19,13 +19,14 @@ val scala2Settings = Seq(
...
     "-Ywarn-value-discard",
+    "-Ytasty-reader",
   ),
   Test / fork := true,
   autoCompilerPlugins := true,
 )
 
 val scala3Settings = Seq(
-  scalaVersion := "3.2.2",
+  scalaVersion := "3.3.1",
```
johnduffell, line 29: scala 3.3 is the LTS release

build.sbt
```diff
@@ -216,7 +217,7 @@ lazy val `effects-lambda` = library(project in file("lib/effects-lambda"))
     dependencyOverrides ++= jacksonDependencies,
   )
 
-lazy val `config-core` = library(project in file("lib/config-core"))
+lazy val `config-core` = library(project in file("lib/config-core"), scala3Settings)
```
johnduffell, line 220: no reason for this project not to be scala 3

build.sbt
```diff
@@ -636,13 +636,22 @@ lazy val `fulfilment-date-calculator` = lambdaProject(
...
   "API for accessing delivery records in Salesforce",
-  Seq(http4sDsl, http4sCirce, http4sServer, circe, sttpAsyncHttpClientBackendCats, scalatest),
+  Seq(
+    http4sDsl,
+    http4sCirce,
+    http4sServer,
+    circe,
+    sttpAsyncHttpClientBackendCats,
+    scalatest,
+    diffx
```
johnduffell, line 646: added diffx to get much clearer diffs of nested case classes in intellij and SBT

build.sbt
```diff
@@ -636,13 +636,22 @@ lazy val `fulfilment-date-calculator` = lambdaProject(
...
+  ),
 ).dependsOn(
   `effects-s3`,
   `config-core`,
   `salesforce-sttp-client`,
   `salesforce-sttp-test-stub` % Test,
   `http4s-lambda-handler`,
+  testDep,
```
johnduffell, line 654: this gives us the EffectsTest annotation

project/Dependencies.scala
```diff
@@ -79,7 +79,7 @@ object Dependencies {
   val parallelCollections = "org.scala-lang.modules" %% "scala-parallel-collections" % "1.0.4"
 
   // Testing
-  val diffx = "com.softwaremill.diffx" %% "diffx-scalatest" % "0.8.3" % Test
+  val diffx = "com.softwaremill.diffx" %% "diffx-scalatest-should" % "0.9.0" % Test
```
johnduffell, line 82: this has been split into two jars since 0.9 and I think the old one has gone out of use

rupertbates (approved):
Looks good, I'd never seen diffx before so that's interesting 👍

PR #2071, opened 2023-10-10
Add auth to New Product API

michaelbjacobson:
This PR adds an API key and requires it for access to the two New Product API endpoints.

### DIFF
--- a/cdk/lib/new-product-api.ts
+++ b/cdk/lib/new-product-api.ts
@@ -5,7 +5,7 @@ import {GuStack} from "@guardian/cdk/lib/constructs/core";
 import {GuLambdaFunction} from "@guardian/cdk/lib/constructs/lambda";
 import type {App} from "aws-cdk-lib";
 import {Duration, Fn} from "aws-cdk-lib";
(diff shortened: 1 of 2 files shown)


cdk/lib/new-product-api.ts
```diff
@@ -126,6 +128,30 @@ export class NewProductApi extends GuStack {
...
+      name: `new-product-api-usage-plan-${this.stage}`,
+      apiStages: [
+        {
+          api: newProductApi.api,
+          stage: newProductApi.api.deploymentStage
+        }
+      ]
+    })
+
+    const apiKey = new ApiKey(this, "NewProductApiKey", {
```
rupertbates, line 142: Is the key from the old stack still around?

rupertbates (approved):
Looks good 👍

cdk/lib/new-product-api.ts
michaelbjacobson: Nope, the old keys got deleted during the migration to CDK.

johnduffell:
looks good although very minor point I would have called the commit "add auth to new-product-api" rather than initial commit as it makes it easier to understand the history

PR #2077, opened 2023-10-18
Migrate batch-email-sender to CDK (Step 1)

michaelbjacobson:
This PR begins the process of migrating `batch-email-sender` to CDK, by importing the raw Cloudformation YAML template into the CDK, and then generating new templates for `PROD` and `CODE` environments.


andrewHEguardian (approved):
Looks good to me! Nice incremental approach

PR #2079, opened 2023-10-19
Dual stack cloudformation (Step 2)

michaelbjacobson:
This PR uses the GuCDK to define a duplicate stack of resources, alongside the current YAML template defined ones. This is a non-destructive process and has been tested manually in `CODE` ahead of merging.

Related PRs:
- [Step 1](https://github.com/guardian/support-service-lambdas/pull/2077)

PR #2082, opened 2023-10-23
ensures unique names and descriptions for alarms, even if they're dis…

michaelbjacobson:
Follow-up to #2079. Alarms defined using the GuCDK are created in both `CODE` and `PROD` environments, but only enabled in `PROD`. Therefore the names need to be unique per environment. This is a new requirement as we only created these alarms in `PROD` when the stack was defined using a YAML template.


andrewHEguardian (approved):
👍

PR #2093, opened 2023-10-30
GuCDK Migration Clean Up 2

michaelbjacobson:
This PR removes the (now redundant) cfn.yaml template for batch email sender. It also removes the `-CDK` suffix from cloudformation resource names.


andrewHEguardian (approved):
Looks good 👏

PR #2097, opened 2023-10-31
Create a new term for subscriptions during product switch

rupertbates:
## What does this change?

We have noticed that the current product switch from recurring contribution to supporter plus has an issue, which is that because the charge is aligned to the subscription end date, unless the switch happens on a term start date the billing amount and date will be unpredictable.

Screenshot 2023-11-01 at 10 25 33

This PR updates the switch mechanism fairly substantially to avoid this, changes are:
- We check the term start date of the subscription to be switched before we begin - if it is not today we also reduce the length of the current term to today and then renew the sub for a new 12 month term. For subs where the term start date is today we skip this step.
- I have removed the requirement for the client to pass in a `checkChargeAmountBeforeUpdate` parameter which then caused us to run an additional preview call to check whether to collect the amount payable or write it off if it was below $0.50. We now just check the balance of the invoice created by the switch for this amount.
- I have moved the collection of payment to a separate api call so that we can separate out the logic involved with it (whether to collect or write off the amount owing) from the updates to the subscription that are being carried out.

## Test Subscriptions
Subscription switched on the day it was taken out - no renewal required:
https://apisandbox.zuora.com/platform/subscriptions/8ad093fb8bc82179018bc86b11d25702

Subscription switched part way through the term - renewal required:
https://apisandbox.zuora.com/platform/subscriptions/8ad097048bc82163018bc86dba57092c

## How to test
- Find a recurring contribution
- Switch it to a Supporter Plus subscription using 
```
curl --location 'https://zvobga46la.execute-api.eu-west-1.amazonaws.com/CODE/product-move/[MY-SUBSCRIPTION-NUMBER]' \
  --header 'x-api-key: [MY-API-KEY]' \
--data '{
    "price": "120",
    "preview": false,
    "checkChargeAmountBeforeUpdate": true
}'
```
- Check that the subscription has been switched to a Supporter Plus sub, that the term start date is today and the term end date is in 12 months time and that the outstanding balance on the account is zero.


johnduffell:
as an aside, we should probably add a proper domain instead of "https://zvobga46la.execute-api.eu-west-1.amazonaws.com" as per the other API gateway lambdas https://github.com/guardian/support-service-lambdas/blob/9f77d88755977913af410699980231ec2c74625c/cdk/bin/cdk.ts#L12

build.sbt
```diff
@@ -306,7 +306,7 @@ def lambdaProject(
   Project(projectName, file(s"handlers/$projectName"))
     .enablePlugins(RiffRaffArtifact)
     .configs(EffectsTest, HealthCheckTest)
-    .settings(scalaSettings, testSettings)
+    .settings(scalaSettings, testSettings, scalafmtSettings)
```
johnduffell, line 309: not related to this PR, but I've noticed intellij fighting with itself in some cases about whether it should do 2 space or 4 space indentation since we've been using scalafmt.  I wasn't sure if 4 space is the standard in some cases and intellij/scalafmt is trying to guess from the existing file contents?

handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/TermRenewal.scala
```diff
@@ -0,0 +1,143 @@
...
+  During this process two invoices will be generated - a negative one for the full amount which represents the fact
+  that we have reduced the current term to zero days from today's payment and then a positive one for the full amount
+  which covers the new 12 month term.
+
+  In order to balance these out, we can use the `applyCreditBalance` parameter in the renewal request.
+   */
+
+  override def startNewTermFromToday[R: JsonDecoder](
+      subscriptionName: SubscriptionName,
+  ): ZIO[Stage with GetSubscription, ErrorResponse, R] = for {
```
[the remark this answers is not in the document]
johnduffell: in fact, we don't seem to ever use the response, so we could just return Unit instead.  As long as we're effectively checking the success status inside within this function

handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/rest/ZuoraGet.scala
```diff
@@ -52,7 +53,11 @@ trait ZuoraGet:
       input: Request,
       zuoraSuccessCheck: ZuoraSuccessCheck = ZuoraSuccessCheck.SuccessCheckLowercase,
   ): IO[ErrorResponse, Response]
-  def put[Request: JsonEncoder, Response: JsonDecoder](relativeUrl: Uri, input: Request): IO[ErrorResponse, Response]
+  def put[Request: JsonEncoder, Response: JsonDecoder](
```
johnduffell, line 56: if we're going to have put/post on this it should probably be renamed to ZuoraRestClient or somerthing.
The reason they were originally separate ZuoraGet/ZuoraPost/etc was to simplify mocking, but if it's causing a problem that way I could go either way on that.

handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/TermRenewal.scala
```diff
@@ -0,0 +1,143 @@
...
+            ),
+          ),
+        ),
+      ),
+    )
+    zuoraGet.put[AmendTermLengthRequest, R](
+      relativeUrl = uri"action/amend",
+      input = requestBody,
+      zuoraSuccessCheck = ZuoraSuccessCheck.None,
+    )
```
johnduffell, line 93: do we want to check anything in the field contents on top of that it all deserialised?

handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/TermRenewal.scala
```diff
@@ -0,0 +1,143 @@
...
+    subscription <- GetSubscription.get(subscriptionName)
+    _ <- amendTermEndDateToToday(SubscriptionId(subscription.id), subscription.termStartDate)
+    response <- renewSubscription(subscriptionName)
+  } yield response
+
+  private def amendTermEndDateToToday[R: JsonDecoder](
+      subscriptionId: SubscriptionId,
+      termStartDate: LocalDate,
+  ): ZIO[Stage, ErrorResponse, R] = {
+    val today = LocalDate.now
```
johnduffell, line 72: is this going to match zuora's timezone? e.g. BST (would be good if zuora had a get-current-time endpoint)

johnduffell: we get the current date later in the code using a different mechanism, I think that's for the thank you email and various tracking purposes. 
https://github.com/guardian/support-service-lambdas/blob/85e06f49f996618be588664e093f3e50cc5d7361/handlers/product-move-api/src/main/scala/com/gu/productmove/switchtype/RecurringContributionToSupporterPlus.scala#L415

handlers/product-move-api/src/main/scala/com/gu/productmove/switchtype/RecurringContributionToSupporterPlus.scala
```diff
@@ -391,6 +403,9 @@ object RecurringContributionToSupporterPlus {
         amountPayableToday,
       )
 
+      // Start a new term when we do the switch to avoid issues with billing dates
+      _ <- TermRenewal.startNewTermFromToday[RenewalResponse](subscriptionName)
```
johnduffell, line 407: given that we are now doing a renewal , the subscription update on line 393 isn't going to be the last word on charges.  We are running billing (and collection) which will produce some invoices (line 388) but we are immediately extending the term which is going to update the term, and then tomorrow some more invoices will be produced?

I'm not sure how this has panned out in testing but it does seem like a grey area.

I think we should probably do the amendment and switch without running billing, and then run billing afterwards, in order to get a single invoice that we can collect straight away as we wish.  And we won't need the 50p/50cents check any more.

handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/TermRenewal.scala
```diff
@@ -0,0 +1,143 @@
...
+)
+case class AmendTermLengthRequest(
+    requests: List[AmendmentRequest],
+)
+case class AmendmentResponse(results: List[AmendmentResult])
+case class AmendmentResult(AmendmentIds: List[String], Success: Boolean)
+case class RenewalRequest(
+    contractEffectiveDate: LocalDate,
+    applyCreditBalance: Boolean = true, // When we shorten
+    runBilling: Boolean = true,
```
johnduffell, line 133: Happy to go either way but might be worth putting these booleans in the code rather than slightly hidden in the default values?

rupertbates:
> as an aside, we should probably add a proper domain instead of "https://zvobga46la.execute-api.eu-west-1.amazonaws.com" as per the other API gateway lambdas
> 
> https://github.com/guardian/support-service-lambdas/blob/9f77d88755977913af410699980231ec2c74625c/cdk/bin/cdk.ts#L12

Agreed, different PR though

build.sbt
rupertbates: Yes Intellij definitely gets a bit confused about it sometimes, import optimisation is another place where it fights with scalafmt

handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/rest/ZuoraGet.scala
rupertbates: Yes I thought that as well but it's been like this for a while. I think I'd favour ZuoraClient over separate classes for GET and POST because of the boilerplate involved with ZIO. It will cause a lot of noise if I do it in this PR though

handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/TermRenewal.scala
rupertbates: I've added a check

handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/TermRenewal.scala
rupertbates: Also generally I think default values can lead to unintended behaviour

handlers/product-move-api/src/main/scala/com/gu/productmove/switchtype/RecurringContributionToSupporterPlus.scala
```diff
@@ -298,30 +299,37 @@ object RecurringContributionToSupporterPlus {
...
         )
       } yield ()
-    } else ZIO.succeed(())
+    } else {
+      println("collectPayment is true, invoice does not need adjusting")
+      ZIO.succeed(())
+    }
+
+  private def getNewTermLengthInDays(today: LocalDate, termStartDate: LocalDate): String =
+    Math.max(ChronoUnit.DAYS.between(termStartDate, today).toInt, 1).toString
```
johnduffell, line 332: so if they try to switch on the day they took it out (which I guess is mainly an issue for testing!) then we just push it on a day?

rupertbates: Yes, I've tested that scenario and it all appears correct. There's an example here: https://apisandbox.zuora.com/platform/subscriptions/8ad093fb8bae945e018bafaaab9b7888

rupertbates: The reason we do that is that Zuora won't let you set a term length to zero

handlers/product-move-api/src/main/scala/com/gu/productmove/switchtype/RecurringContributionToSupporterPlus.scala
```diff
@@ -377,35 +386,56 @@ object RecurringContributionToSupporterPlus {
             amount = previewResponse.asInstanceOf[PreviewResult].amountPayableToday
           } yield amount
         } else ZIO.succeed(BigDecimal(1))
-      collectPayment = !(amountPayableToday > BigDecimal(0) && amountPayableToday < BigDecimal(0.50))
+      collectPayment = amountPayableToday < 0 || amountPayableToday >= 0.5
```
johnduffell, line 388: just want to check that this preview is right?

I don't think we need to do that anyway, we can probably make the code less coupled -
I think we can just skip the preview completely, and once we've renewed the subscription (without collect=true) we can have a look at the actual resulting invoice and either adjust or take the payment.

handlers/product-move-api/src/main/scala/com/gu/productmove/switchtype/RecurringContributionToSupporterPlus.scala
johnduffell: that is a bit annoying - as the term will then be one day out from the charge.  Would it be better, if the term starts today and is 12 months anyway, to just not touch it at all?

handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/CreatePayment.scala
```diff
@@ -0,0 +1,106 @@
...
+object CreatePaymentLive:
+  val layer: URLayer[ZuoraGet, CreatePayment] = ZLayer.fromFunction(CreatePaymentLive(_))
+
+private class CreatePaymentLive(zuoraGet: ZuoraGet) extends CreatePayment:
+  /*
+  Start a new term for this subscription from today.
+  This is to avoid problems with charges not aligning correctly with the term and resulting in unpredictable
+  billing dates and amounts.
+
+  Uses https://www.zuora.com/developer/api-references/api/operation/PUT_RenewSubscription/
```
johnduffell, line 51: this whole comment is outdated

rupertbates: Remnant of copy and paste

handlers/product-move-api/src/main/scala/com/gu/productmove/switchtype/RecurringContributionToSupporterPlus.scala
rupertbates, line 1: I reordered a lot of functions in this class which is why the diff is so big

handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/GetAccount.scala
```diff
@@ -80,35 +82,26 @@ object GetAccount {
...
-              case _ =>
-                BasicInfo(
-                  defaultPaymentMethod,
-                  Some(id),
-                  sfContactId__c,
-                  (balance.toDouble * 100).toInt,
-                  currencyCodetoObject(currency),
-                )
-            }
         }
```
johnduffell, line 105: maybe there used to be a difference but not any more?
```suggestion
        BasicInfo(
          id = id,
          defaultPaymentMethod = defaultPaymentMethod,
          IdentityId__c = identityId.filter(_ != ""),
          sfContactId__c = sfContactId__c,
          balance = (balance.toDouble * 100).toInt,
          currency = currencyCodetoObject(currency),
        )
```

handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/TermRenewal.scala
```diff
@@ -0,0 +1,96 @@
...
+
+  Uses https://www.zuora.com/developer/api-references/api/operation/PUT_RenewSubscription/
+   */
+
+  override def renewSubscription(
+      subscriptionName: SubscriptionName,
+      runBilling: Boolean,
+  ): ZIO[Stage with GetSubscription, ErrorResponse, RenewalResponse] = for {
+    _ <- ZIO.log(s"Attempting to renew subscription $subscriptionName")
+    today <- Clock.currentDateTime.map(_.toLocalDate)
```
johnduffell, line 57: I checked this out regarding timezones and I think it will be ok as long as we only run this code in GMT/BST region of AWS.

"Zuora observes Daylight Savings Time, if applicable."
https://knowledgecenter.zuora.com/Zuora_Central_Platform/Tenant_Management/A_Administrator_Settings/Managing_Your_tenant_profile

"To get current date time in the current timezone the currentDateTime function returns a ZIO effect containing OffsetDateTime"
https://zio.dev/reference/services/clock/
.... toLocalDate "Gets the LocalDate part of this date-time."
https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/time/OffsetDateTime.html#toLocalDate()

handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/TermRenewal.scala
```diff
@@ -0,0 +1,96 @@
...
+    response <- renewSubscription(subscriptionName, today, runBilling)
+    _ <- ZIO.log(s"Successfully renewed subscription $subscriptionName")
+  } yield response
+
+  private def renewSubscription(
```
johnduffell, line 62: not sure about naming but consider calling the two functions different things
```suggestion
  private def renewSubscriptionFromDate(
```

handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/TermRenewal.scala
```diff
@@ -0,0 +1,96 @@
...
+      subscriptionName: SubscriptionName,
+      contractEffectiveDate: LocalDate,
+      runBilling: Boolean,
+  ): ZIO[Stage, ErrorResponse, RenewalResponse] = {
+    val requestBody = RenewalRequest(
+      contractEffectiveDate,
+      collect = None,
+      runBilling = runBilling,
```
johnduffell, line 70: I can't remember if we discussed this last time but this is always true so we could just hard code here
```suggestion
      runBilling = true,
```

handlers/product-move-api/src/test/scala/com/gu/productmove/zuora/CreatePaymentSpec.scala
```diff
@@ -0,0 +1,41 @@
...
+              ZuoraClientLive.layer,
+              CreatePaymentLive.layer,
+              SttpClientLive.layer,
+              SecretsLive.layer,
+            )
+
+        } yield {
+          assert(true)(equalTo(true))
+        }
+      } @@ TestAspect.ignore,
```
johnduffell, line 39: what is this test for?  I know that "Live" means contact the actual DEV zuora, but I'm not sure if this test should be active, or is it just there for manual testing/setup?  If so it might need to be better to put in in a specific package and add to the readme?

handlers/product-move-api/src/test/scala/com/gu/productmove/HandlerSpec.scala
```diff
@@ -379,11 +407,14 @@ object HandlerSpec extends ZIOSpecDefault {
...
           ZLayer.succeed(new MockGetSubscription(getSubscriptionStubs())),
           ZLayer.succeed(new MockSubscriptionUpdate(subscriptionUpdatePreviewStubs, subscriptionUpdateStubs)),
+          ZLayer.succeed(new MockTermRenewal(termRenewalStubs)),
           ZLayer.succeed(new MockSQS(sqsStubs)),
           ZLayer.succeed(new MockDynamo(dynamoStubs)),
           ZLayer.succeed(new MockGetAccount(getAccountStubs, getPaymentMethodStubs)),
           ZLayer.succeed(new MockInvoiceItemAdjustment(invoiceItemAdjustmentStubs)),
           ZLayer.succeed(new MockGetInvoiceItems(getInvoiceItemsStubs)),
+          ZLayer.succeed(new MockGetInvoice(getInvoiceStubs)),
+          ZLayer.succeed(new MockCreatePayment(CreatePaymentResponse(Some(true)))),
```
johnduffell, line 417: I know these tests are ignored (temporarily?) but do we need to check the `mutableStore` as part of the test assertions, to make sure it really makes the calls it should?

handlers/product-move-api/src/main/scala/com/gu/productmove/switchtype/RecurringContributionToSupporterPlus.scala
johnduffell: ok thanks for the warning, I'll do my best, I will try to comment on new code rather than reordered stuff!

handlers/product-move-api/src/main/scala/com/gu/productmove/switchtype/RecurringContributionToSupporterPlus.scala
```diff
@@ -331,77 +206,93 @@ object RecurringContributionToSupporterPlus {
...
+          adjustNonCollectedInvoice(
+            invoiceId,
+            supporterPlusRatePlanIds,
+            subscriptionName,
+            invoiceBalance,
+          ).as(BigDecimal(0))
+        }
 
-      account <- accountFuture.join
+      _ <- ZIO.log(s"The amount paid on switch by the customer was $paidAmount")
```
johnduffell, line 289: glad to see all the extra logging you've added in 👍

handlers/product-move-api/src/main/scala/com/gu/productmove/switchtype/RecurringContributionToSupporterPlus.scala
```diff
@@ -471,6 +361,208 @@ object RecurringContributionToSupporterPlus {
...
+    val newTermLength = getNewTermLengthInDays(today, termStartDate)
+    for {
+      updateRequestBody <- getRatePlans(billingPeriod, currency, currentRatePlanId, price).map {
+        case (addRatePlan, removeRatePlan) =>
+          SwitchProductUpdateRequest(
+            add = addRatePlan,
+            remove = removeRatePlan,
+            currentTerm = Some(newTermLength),
+            currentTermPeriodType = Some("Day"),
+            // We will run the billing during the renewal call to reduce the number of invoices
```
johnduffell, line 413:
```suggestion
            // We will run the billing during the renewal call to ensure there's exactly one invoice produced for the switch
```

handlers/product-move-api/src/main/scala/com/gu/productmove/switchtype/RecurringContributionToSupporterPlus.scala
```diff
@@ -471,6 +361,208 @@ object RecurringContributionToSupporterPlus {
...
+            runBilling = Some(false),
+            // We will collect via a separate create payment call if the amount payable is not too small
+            collect = None,
+            preview = Some(false),
+          )
+      }
+      updateResponse <- SubscriptionUpdate.update[SubscriptionUpdateResponse](subscriptionName, updateRequestBody)
+      renewalResult <- TermRenewal.renewSubscription(
+        subscriptionName,
+        // Run billing here as this is the last change being made
```
johnduffell, line 423:
```suggestion
        // this is the last change so run billing to get one invoice to cover all the changes
```

handlers/product-move-api/src/main/scala/com/gu/productmove/switchtype/RecurringContributionToSupporterPlus.scala
```diff
@@ -471,6 +361,208 @@ object RecurringContributionToSupporterPlus {
...
+      .mapError(x => InternalServerError(x))
+
+  def getRatePlans(
+      billingPeriod: BillingPeriod,
+      currency: Currency,
+      ratePlanIdToRemove: String,
+      price: BigDecimal,
+  ): ZIO[Stage, ErrorResponse, (List[AddRatePlan], List[RemoveRatePlan])] =
+    for {
+      date <- Clock.currentDateTime.map(_.toLocalDate)
```
johnduffell, line 508: I'm honestly never sure whether we're supposed to pass in the date, or just use this "built in" zio service whereever we need it?

handlers/product-move-api/src/test/scala/com/gu/productmove/zuora/CreatePaymentSpec.scala
rupertbates: It's just there to aid local development, ideally this would be an effects test which you could run with a different command but I'm not aware of a way to do that with ZIO

handlers/product-move-api/src/test/scala/com/gu/productmove/HandlerSpec.scala
rupertbates: Those tests are not ignored, the mocked ones run the ones with live classes don't

handlers/product-move-api/src/main/scala/com/gu/productmove/switchtype/RecurringContributionToSupporterPlus.scala
```diff
@@ -331,77 +206,93 @@ object RecurringContributionToSupporterPlus {
...
-        productSwitchRatePlanIds.supporterPlusRatePlanIds,
-        subscriptionName,
-        amountPayableToday,
-      )
+      /*
+        If the amount payable today is less than 0.50, we don't want to collect payment because Stripe has a minimum
+        charge of 50 cents. Instead we write-off the invoices in the `adjustNonCollectedInvoices` function.
+       */
+      paidAmount <-
+        if (invoiceBalance < 0 || invoiceBalance >= 0.5) {
```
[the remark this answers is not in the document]
rupertbates: Yeah I don't know either, just retaining the existing logic here with the assumption that someone has included it for a reason

handlers/product-move-api/src/main/scala/com/gu/productmove/switchtype/RecurringContributionToSupporterPlus.scala
rupertbates: I think it achieves the same thing either way, which is to enable passing in a different date when needed for testing purposes

handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/TermRenewal.scala
rupertbates: I've left it like this because it makes it more explicit at the call site that the renewal call is where the invoices are generated

handlers/product-move-api/src/main/scala/com/gu/productmove/switchtype/RecurringContributionToSupporterPlus.scala
rupertbates: This was the route I took in the end

handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/TermRenewal.scala
johnduffell: yeah good point, the whole process is so complicated so keeping the code explicit is a good idea

handlers/product-move-api/src/test/scala/com/gu/productmove/zuora/CreatePaymentSpec.scala
johnduffell: ok fair enough happy to leave it as is or perhaps could we put them in com.gu.productmove.manual or something, and perhaps you could add a comment or something maybe in the readme? perhaps in a separate PR.

---
for reference...
It looks like ZIO does support running by tag with different parameters to scalatest's way.  https://zio.dev/reference/test/running-tests/  I would call these Manual scripts, not even Effects tests.
It does expose, given how much of our stuff relies on integrations (rather than fancy code) we have a bit of a weak process for testing those interfaces.  I'm still tempted by some kind of request/response recording into S3, and expiring them regularly (or hashing any PII) so that we can replay the last x weeks of production as a effects tests.  One for 10% time maybe.

johnduffell (approved):
nice work persisiting on that one, it's one of those quick jobs that turns into a mountain (of false summits..).

PR #2080, opened 2023-10-19
feat: write lambda polling SQS and creating a single contribution record in Salesforce

AndreaDiotallevi:
## What does this change?

- Implement lambda handler polling the SQS queue and creating a single contribution record in Salesforce
- Add Salesforce auth credentials secrets in AWS Secrets Manager
- Add policy statement to lambda role policy to allow `secretsmanager:GetSecretValue` of Salesforce auth credentials
- (health) Delete all files related to the obsolete project failed-national-deliveries-processor

[**Trello Card**](https://trello.com/c/JCxMwBRv/335-single-contribution-record-create-lambda-to-create-a-single-contribution-record-in-salesforce)

## Why are you doing this?

This is part of the epic [Create a single contribution record in Salesforce via support-payment-api](https://trello.com/c/qibQVaR2/295-create-a-single-contribution-record-in-salesforce-via-support-frontend-payment-api).

Previous work:

1. [Create single-contribution-salesforce-writes project with SQS queues to handle single contribution messages](https://github.com/guardian/support-service-lambdas/pull/2059)
2. [Add rule to acquisition bus to send payment-api messages to the single-contribution-salesforce-writes-queue](https://github.com/guardian/support-service-lambdas/pull/2066)
3. [Add fields (postalCode, state, email) to AcquisitionDataRow](https://github.com/guardian/support-frontend/pull/5345)

## Architecture
Screenshot 2023-10-23 at 10 41 47

## Project structure

I have attempted to create a standardised folder structure:

- scala
  - models
    - Errors
  - handlers
    - CreateSalesforceSingleContributionRecordHandler
  - services
    - awsSecretsManager
      - GetSecretValue
    - jsonDecoder
      - DecodeJson
    - salesforce
      - CreateSingleContributionRecord
      - GetAccessToken
      - GetApiDomain
- test (to be completed) 

## Design notes
  
- The lambda handler is separated from the core logic as suggested [here](https://docs.aws.amazon.com/lambda/latest/dg/best-practices.html)
- CloudWatch alarm strategy needs to be added to the diagram
- PROD Salesforce secrets to be added before deploying to production

## How to test

1. Make a single contribution from the [CODE site](https://support.code.dev-theguardian.com/uk/contribute)
2. From the Salesforce sandbox dashboard, check that the contribution record has been created [here](https://gnmtouchpoint--dev1.sandbox.lightning.force.com/lightning/r/Report/00O9E000001A5vkUAC/view)

## How can we measure success?

[...]

PR #2101, opened 2023-11-02
fix: make identityId optional

AndreaDiotallevi:
## What does this change?

This makes the `identityId` field optional, to avoid breaking the lambda when the identityId is not present in the event.

```scala
case class PaymentApiMessageDetail(
    amount: Double,
    contributionId: String,
    country: String,
    currency: String,
    eventTimeStamp: String,
    identityId: Option[String],
    paymentId: String,
    paymentProvider: String,
    postalCode: Option[String],
    state: Option[String],
    email: String,
)
```

This is a fix relative to this [PR](https://github.com/guardian/support-service-lambdas/pull/2080).


rupertbates (approved):
👍

PR #2104, opened 2023-11-02
Cancellation Salesforce Cases GuCDK Migration Step 1

michaelbjacobson:
This first PR spins up a dual stack, per the usual migration pattern, ie. defining duplicate resources using CDK and including existing, template-defined resources using `CfnInclude`.

### DIFF
--- a/cdk/bin/cdk.ts
+++ b/cdk/bin/cdk.ts
@@ -4,6 +4,7 @@ import { BatchEmailSender } from "../lib/batch-email-sender";
 import type { NewProductApiProps } from "../lib/new-product-api";
 import { NewProductApi } from "../lib/new-product-api";
 import { APP_NAME as SINGLE_CONTRIBUTION_SALESFORCE_WRITES_APP_NAME, SingleContributionSalesforceWrites } from "../lib/single-contribution-salesforce-writes";
+import {CancellationSfCasesApi} from "../lib/cancellation-sf-cases-api";
 
 const app = new App();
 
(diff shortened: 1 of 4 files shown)


cdk/bin/cdk.ts
```diff
@@ -1,49 +1,67 @@
...
+const certificateId = "c1efc564-9ff8-4a03-be48-d1990a3d79d2"
+const membershipApisDomain = "membership.guardianapis.com"
 
 export const codeProps: NewProductApiProps = {
     stack: "membership",
     stage: "CODE",
-    domainName: "new-product-api-code.membership.guardianapis.com",
-    hostedZoneId: "Z1E4V12LQGXFEC",
-    certificateId: "c1efc564-9ff8-4a03-be48-d1990a3d79d2",
+    domainName: `new-product-api-code.${membershipApisDomain}`,
```
johnduffell, line 17: this is a good pattern that we started using to give our stuff a proper domain, so it's good that we are retaining this.

cdk/bin/cdk.ts
```diff
@@ -1,49 +1,67 @@
...
     stack: "membership",
     stage: "PROD"
 });
 
+new CancellationSfCasesApi(app, "cancellation-sf-cases-api-CODE", {
+    stack: "membership",
+    stage: "CODE",
+    domainName: `cancellation-sf-cases-api-code.${membershipApisDomain}`,
+    certificateId,
+    hostedZoneId,
```
johnduffell, line 55: can we keep the ordering consistent, e.g. above there are some with stack and stage at the end

johnduffell (approved):
looks good, nothing stands out, and we've got the security key set up by the looks!

PR #2110, opened 2023-11-17
Add setup for Typescript lambdas

rupertbates:
## What does this change?
This PR introduces a framework for managing Typescript lambdas within support-service-lambdas. 
It uses [pnpm workspaces](https://pnpm.io/workspaces) to handle common dependencies, project interdependencies etc. see the readme for more info.

It also adds a new Typescript api-gateway triggered lambda `discount-api` which uses this new project setup as an example, currently this lambda does nothing other than return a successful 200 response with the body 'Hello World'.


shtukas (approved):

cdk/bin/cdk.ts
```diff
@@ -1,62 +1,94 @@
...
-import type { NewProductApiProps } from "../lib/new-product-api";
-import { NewProductApi } from "../lib/new-product-api";
-import { APP_NAME as SINGLE_CONTRIBUTION_SALESFORCE_WRITES_APP_NAME, SingleContributionSalesforceWrites } from "../lib/single-contribution-salesforce-writes";
+import 'source-map-support/register';
+import { App } from 'aws-cdk-lib';
+import { BatchEmailSender } from '../lib/batch-email-sender';
+import { CancellationSfCasesApi } from '../lib/cancellation-sf-cases-api';
+import { DiscountApi } from '../lib/discount-api';
+import type { NewProductApiProps } from '../lib/new-product-api';
+import { NewProductApi } from '../lib/new-product-api';
```
rupertbates, line 7: Formatting changes in this file are because we are now using the standard Guardian prettier format

handlers/HOWTO-create-lambda.md
```diff
@@ -2,7 +2,22 @@
...
+1. set up [continuous deployment](https://riffraff.gutools.co.uk/deployment/continuous) for future deploys (optional)
+
+## Typescript lambdas
+### Api Gateway with a custom domain name
+1. copy the discount-api lambda and the associated CDK file discount-api.ts
+1. add the new project to [pnpm-workspace.yaml](https://github.com/guardian/support-service-lambdas/blob/647299d94b471d0065030bce692a4631078346bb/pnpm-workspace.yaml) so that it can be built from the root of the project
+1. add the new project to [ci-typescript.yml](https://github.com/guardian/support-service-lambdas/blob/647299d94b471d0065030bce692a4631078346bb/.github/workflows/ci-typescript.yml)
+1. delete all the code from the handler (index.ts) that you don't need
+1. search and replace all sf-contact-merge in the code/packages/cdk with your chosen name
+1. push the branch and deploy to code (see below)
```
rupertbates, line 17: I'm going to try and improve this approach using code generation at some point

handlers/discount-api/src/index.ts
```diff
@@ -0,0 +1,15 @@
...
+	APIGatewayProxyResult,
+	Handler,
+} from 'aws-lambda';
+
+export const handler: Handler = async (
+	event: APIGatewayProxyEvent,
+): Promise<APIGatewayProxyResult> => {
+	console.log(`Input is ${JSON.stringify(event)}`);
+	return await Promise.resolve({
+		body: 'Hello World',
```
rupertbates, line 12: This lambda just returns a 200 with the body 'Hello World' atm

.github/workflows/ci-typescript.yml
```diff
@@ -0,0 +1,51 @@
...
+          aws-region: eu-west-1
+
+      - name: Generate CFN templates from GuCDK
+        run: bash ./script/cfn.sh
+
+      - name: Upload to Riff-Raff
+        uses: guardian/actions-riff-raff@v2
+        with:
+          projectName: support-service-lambdas::${{ matrix.subproject }}
+          buildNumberOffset: 7000
```
johnduffell, line 43: why 7000? presumably it won't stay in sync with the scala build numbers unless we combine them?

.github/workflows/ci-typescript.yml
```diff
@@ -0,0 +1,51 @@
+name: CI-Typescript
```
johnduffell, line 1: do we need to add anything to the snyk workflow as well? or will it pick up the TS automatically?

README.md
```diff
@@ -1,13 +1,51 @@
...
+  }
+}
+```
+
+Dependencies can also be managed either at the root workspace level or in individual lambdas. To add to the workspace
+use:
+```shell
+pnpm --workspace-root add dayjs
+```
+this should only be used for dependencies which really are necessary for all sub-projects.
```
johnduffell, line 41: not sure how many dependencies would really be needed for all? unless it's for build/deployment or some kind of snyk etc.  Maybe deemphasise the root workspace level in this section?

handlers/HOWTO-create-lambda.md
johnduffell: good idea, also the less code that needs to be generated the better (if we can have defaults for most things)

handlers/discount-api/riff-raff.yaml
```diff
@@ -0,0 +1,25 @@
+stacks:
```
johnduffell, line 1: this file can go in the workflow under the actions-riff-raff section, or there's one automatically generated by CDK in the output folder (I think the cdk one is recommended by devx)

johnduffell (approved):
great I think this is a big step forward, I am not sure how IDEs are going to handle it, but I suspect it won't be a big issue?  Thanks for doing all the research and I look forward to getting rid of some separate repos! 👍

michaelbjacobson (approved):
Brilliant!

.github/workflows/ci-typescript.yml
rupertbates: Started with a copy of the scala one and didn't change it

handlers/discount-api/riff-raff.yaml
rupertbates: It's easier to generate one as a separate file rather than trying to programatically edit the build file, and I don't think the cdk one  supports api gateway lambdas currently from looking at the docs.

.github/workflows/ci-typescript.yml
rupertbates: Great question, I've checked and the existing one will pick up node dependencies as well

PR #2112, opened 2023-11-20
Code generation for new Typescript Lambdas

rupertbates:
## What does this change?
This PR adds a node script to make it simple to add a new Typescript lambda with an API Gateway frontend and a custom domain to the repo. 

## Details
This script uses the [hygen.io](https://www.hygen.io/) code generator to create all the files needed for a new Typescript lambda:
- a package.json a riff-raff.yaml, a tsconfig.json and a src\index.ts file
- a new CDK lib definition for your new lambda as well as a new test and snapshot file
- a new instance of your CDK lib in the root cdk.ts file
- a new subproject entry in the Github build file ci-typescript.yml

## Usage
At the root of the repo ensure you have run `pnpm install`, then run `pnpm new-lambda`. You will be prompted to enter the name of your new lambda eg. `my-new-api` and the script will then generate the necessary files.

To test your new lambda push your branch and deploy to code (take note of the information [here](https://github.com/guardian/support-service-lambdas/blob/224191949d7b3108ab473398433ad76671a1948d/handlers/HOWTO-create-lambda.md#important-deploying-a-lambda-with-riffraff-for-the-first-time)).
You should then be able to call your lambda with cUrl 
```shell
curl --location 'https://my-new-api-code.membership.guardianapis.com/prod'
```
and receive a 200 response with the body 'Hello World'.
 <video loop src="https://github.com/guardian/support-service-lambdas/assets/181371/9f824400-7282-435f-9f73-d8a42a226628">  video </video> 

## TODO
- It would be good to add a generator for a plain lambda (without API Gateway)
- Include authentication in template


shtukas (approved):

_templates/new-lambda/api-gateway/riff-raff.ejs.t
```diff
@@ -0,0 +1,31 @@
+---
```
johnduffell, line 1: can we use the one that CDK produces as part of the CDK build?

cdk/lib/__snapshots__/discount-api.test.ts.snap
```diff
@@ -175,8 +175,8 @@ exports[`The Discount API stack matches the snapshot 1`] = `
     },
     "DNSRecord": {
       "Properties": {
-        "HostedZoneId": "Z1E4V12LQGXFEC",
-        "Name": "discount-api.code.membership.guardianapis.com",
+        "HostedZoneId": "Z3KO35ELNWZMSX",
+        "Name": "discount-api.code.support.guardianapis.com",
```
johnduffell, line 179: my reservation is having two standards, if we're already using membership.guardianapis.com in this repo, why add a new one? Or is the aim to move them all over?

package.json
```diff
@@ -4,17 +4,19 @@
 		"package": "pnpm --stream -r run package",
 		"type-check": "pnpm --stream -r run type-check",
 		"lint": "pnpm --stream -r run lint",
-		"clean": "rm -rf .parcel-cache && pnpm -r run clean"
+		"clean": "rm -rf .parcel-cache && pnpm -r run clean",
+		"new-lambda": "pnpm hygen new-lambda api-gateway && cd ./cdk && yarn lint --fix && yarn test -u && git add ./lib/__snapshots__/*.test.ts.snap"
```
johnduffell, line 8: what's this line for?

_templates/new-lambda/api-gateway/cdk-lib.ejs.t
```diff
@@ -0,0 +1,128 @@
...
+
+		const s3InlinePolicy: Policy = new Policy(this, 'S3 inline policy', {
+			statements: [
+				new PolicyStatement({
+					effect: Effect.ALLOW,
+					actions: ['s3:GetObject'],
+					resources: ['arn:aws:s3::*:membership-dist/*'],
+				}),
+			],
+		});
```
akash1810, line 124: IIUC this policy is created to download the application artifact? GuCDK provides this as standard. The GuCDK one is also more tightly scoped to `bucket/stack/stage/app/*`, following the principle of least privilege more closely.

akash1810 (commented):
Added a slight drive-by review... I landed here via a conversation in Chat 👋🏽 .

_templates/new-lambda/api-gateway/riff-raff.ejs.t
rupertbates: Will check

package.json
rupertbates: That's the script that runs the code generator

cdk/lib/__snapshots__/discount-api.test.ts.snap
rupertbates: I'd like to move them over ideally. Membership hasn't been a thing for what 5 years? We use support everywhere else so why not standardise on it here?

_templates/new-lambda/api-gateway/riff-raff.ejs.t
rupertbates: It's a bit more complicated to add it because we have individually deployable projects whereas the cdk one currently is set up to generate one riff-raff.yaml file for the whole cdk build. 

It may be possible to make it work using the [advanced configuration options](https://github.com/guardian/cdk/tree/main/src/riff-raff-yaml-file#advanced-usage) but it's not straightforward and I'm not convinced there's a huge benefit currently.

pnpm-workspace.yaml
```diff
@@ -1,5 +1,5 @@
 packages:
   # lambdas
-  - 'handlers/discount-api'
+  - 'handlers/*'
```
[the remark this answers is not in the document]
rupertbates: Yeah the package.json I think, that's where the project name is defined

PR #2115, opened 2023-11-23
use official ZIO-lambda library in product-move-api

johnduffell:
I noticed that the docs endpoint for product-move-api was returning a 500 error
```
Failed with: Die(scala.MatchError: InputStreamRangeBody (of class sttp.tapir.RawBodyType$InputStreamRangeBody$),Stack trace for thread "zio-fiber-227": at com.gu.productmove.framework.TIOInterpreter.given_MonadError_TIO.$anon.flatMap(TIOInterpreter.scala:39) at com.gu.productmove.framework.TIOInterpreter.given_MonadError_TIO.$anon.handleWrappedError(TIOInterpreter.scala:46) at com.gu.productmove.framework.TIOInterpreter.given_MonadError_TIO.$anon.map(TIOInterpreter.scala:37) at com.gu.productmove.framework.TIOInterpreter.given_MonadError_TIO.$anon.flatMap(TIOInterpreter.scala:39) at com.gu.productmove.framework.TIOInterpreter.given_MonadError_TIO.$anon.map(TIOInterpreter.scala:37) at com.gu.productmove.framework.ZIOApiGatewayRequestHandler.handleWithLoggerAndErrorHandling(ZIOApiGatewayRequestHandler.scala:120) at com.gu.productmove.framework.ZIOApiGatewayRequestHandler.handleWithLoggerAndErrorHandling(ZIOApiGatewayRequestHandler.scala:121) at com.gu.productmove.framework.ZIOApiGatewayRequestHandler.handleWithLoggerAndErrorHandling(ZIOApiGatewayRequestHandler.scala:122))
```

When I looked into it, it was because I hadn't covered all the cases in the custom ZIO/aws lambda interpreter I wrote for tapir early last year.

Rather than adding the one case, I realised that a few months ago they added an official interpreter if we bumped to a later version of tapir. https://github.com/softwaremill/tapir/pull/2975

So I switched over to the new one, which worked fine.  I also had to change to the more standard ZIO Task rather than using our own custom TIO, which meant changing the error type from Any to Throwable.  This might help our error stack traces, which can often be a bit unclear with other types.

Tested in CODE and working again
Feel free to take a look yourself at the docs https://product-move-api-code.support.guardianapis.com/docs/
Also tested the available moves endpoint
{{product-movement-baseurl}}/available-product-moves/A-S00737090
response data looks ok
```json
[
    {
        "id": "2c92c0f84bbfec8b014bc655f4852d9d",
        "name": "Digital Pack",
        "billing": {
            "amount": 1499,
            "currency": {
                "code": "GBP",
                "symbol": "£"
            },
            "frequency": {
                "name": {
                    "month": {}
                },
                "count": 1
            },
            "startDate": "2024-03-27"
        },
        "trial": {
            "dayCount": 14
        },
        "introOffer": {
            "billing": {
                "percentage": 50,
                "startDate": "2023-12-28"
            },
            "duration": {
                "name": {
                    "month": {}
                },
                "count": 3
            }
        }
    }
]
```


build.sbt
```diff
@@ -568,6 +568,7 @@ lazy val `product-move-api` = lambdaProject(
     "com.softwaremill.sttp.tapir" %% "tapir-core" % tapirVersion,
     "com.softwaremill.sttp.tapir" %% "tapir-json-zio" % tapirVersion,
     "com.softwaremill.sttp.tapir" %% "tapir-aws-lambda" % tapirVersion,
+    "com.softwaremill.sttp.tapir" %% "tapir-aws-lambda-zio" % tapirVersion,
```
johnduffell, line 571: this is the new library added in https://github.com/softwaremill/tapir/pull/2975

handlers/product-move-api/src/main/scala/com/gu/productmove/AwsCredentialsLive.scala
```diff
@@ -13,8 +13,8 @@ object AwsCredentialsLive {
 
   private val ProfileName = "membership"
 
-  val layer: Layer[String, AwsCredentialsProvider] =
-    ZLayer.scoped(ZIO.fromAutoCloseable(ZIO.attempt(impl))).mapError(_.toString)
+  val layer: Layer[Throwable, AwsCredentialsProvider] =
+    ZLayer.scoped(ZIO.fromAutoCloseable(ZIO.attempt(impl)))
```
johnduffell, line 17: the official zio interpreter expects Throwable as the error, rather than Any, so I had to make everything comply.  That is the majority of the changes.

handlers/product-move-api/src/main/scala/com/gu/productmove/SQS.scala
```diff
@@ -44,11 +44,11 @@ object SQS {
 }
 
 object SQSLive {
-  val layer: ZLayer[AwsCredentialsProvider with Stage, ErrorResponse, SQS] =
+  val layer: RLayer[AwsCredentialsProvider & Stage, SQS] =
     ZLayer.scoped(for {
       stage <- ZIO.service[Stage]
       sqsClient <- initializeSQSClient().mapError(ex =>
-        InternalServerError(s"Failed to initialize SQS Client with error: $ex"),
+        new Throwable(s"Failed to initialize SQS Client with error", ex),
```
johnduffell, line 51: we were returning InternalServerError but I don't know if it was actually getting used directly, or just converted into an internal server error later anyway

handlers/product-move-api/src/main/scala/com/gu/productmove/Handler.scala
```diff
@@ -1,78 +1,139 @@
...
-object Handler extends ZIOApiGatewayRequestHandler {
+object Handler
+    extends ZIOApiGatewayRequestHandler(
+      List(
+        AvailableProductMovesEndpoint.server,
+        ProductMoveEndpoint.server,
+        UpdateSupporterPlusAmountEndpoint.server,
+        SubscriptionCancelEndpoint.server,
+      ),
+    )
```
johnduffell, line 33: I removed all the test stuff from the handler and made the endpoints a constructor parameter.

This was because with inheritance and vals I was getting issues with initialisation order, and `server` was coming out as null.
Rather than just going down the lazy val option (which worked) I decided to separate/simplify appropriately.

handlers/product-move-api/src/main/scala/com/gu/productmove/Handler.scala
```diff
@@ -1,78 +1,139 @@
...
+  private val devSubscriptionNumber = "A-S00737111"
 
   @main
   // run this to test locally via console with some hard coded data
-  def testProductMove(): Unit = super.runTest(
+  def testProductMove(): Unit = runTest(
     "POST",
-    "/product-move/A-S123",
-    Some(ProductMoveEndpointTypes.ExpectedInput(49.99, false, None, None).toJson),
+    "/product-move/recurring-contribution-to-supporter-plus/" + devSubscriptionNumber,
```
johnduffell, line 43: this was not matching the actual routes definiton.

handlers/product-move-api/src/main/scala/com/gu/productmove/Handler.scala
```diff
@@ -1,78 +1,139 @@
...
+  // this will output the HTML to the console
+  def testDocsHtml(): Unit = {
+    runTest(
+      "GET",
+      "/docs/",
+      None,
+    )
+  }
+
+  // for testing
```
johnduffell, line 75: this functions moved out of the superclass `ZIOApiGatewayRequestHandler`

handlers/product-move-api/src/main/scala/com/gu/productmove/framework/ZIOApiGatewayRequestHandler.scala
```diff
@@ -1,173 +1,56 @@
...
+    Unsafe.unsafe { implicit unsafe =>
+      runtime.unsafe
+        .run(
+          for {
+            allBytes <- ZIO.attempt(input.readAllBytes())
+            _ = input.close()
+            str = new String(allBytes, StandardCharsets.UTF_8)
+            _ <- ZIO.log("input: " + str)
+            loggedOutputStream = new ByteArrayOutputStream()
+            _ <- handler.process[AwsRequestV1](new ByteArrayInputStream(allBytes), loggedOutputStream)
```
johnduffell, line 46: we have to use V1 because we are still using the 1.0 integration
https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-lambda.html

I'm not sure if we should be updating or not, it doesn't seem to encourage it in the documentation.

handlers/product-move-api/src/main/scala/com/gu/productmove/framework/ZIOApiGatewayRequestHandler.scala
```diff
@@ -1,173 +1,56 @@
...
-    val route: Route[TIO] = TIOInterpreter().toRoute(server ++ swaggerEndpoints)
-    val routedTask: TIO[AwsResponse] = route(awsRequest)
+  // this is the main lambda entry point.  It is referenced in the cloudformation.
+  override def handleRequest(input: InputStream, output: OutputStream, context: Context): Unit = {
     val runtime = Runtime.default
-    Unsafe.unsafe { implicit u =>
-      runtime.unsafe.run(
-        routedTask
-          .provideLayer(Runtime.removeDefaultLoggers)
-          .provideLayer(Runtime.addLogger(new AwsLambdaLogger(context.getLogger))),
```
johnduffell, line 121: I removed the lambda logger support, this means we use the default logging.
I'm not sure this was actually still working anyway, as looking in PROD I can't see the timestamps being generated.
Ultimately everything that goes to stdout will end up in the cloudwatch logs, it might just not look nice.

handlers/product-move-api/src/main/scala/com/gu/productmove/framework/ZIOApiGatewayRequestHandler.scala
```diff
@@ -1,173 +1,56 @@
...
-import java.nio.charset.Charset
-import java.util.Base64
-import scala.concurrent.Future
+import java.io.{ByteArrayInputStream, ByteArrayOutputStream, InputStream, OutputStream}
+import java.nio.charset.StandardCharsets
 import scala.jdk.CollectionConverters.*
-import scala.util.{Success, Try}
 
-object ZIOApiGatewayRequestHandler {
+abstract class ZIOApiGatewayRequestHandler(val server: List[ServerEndpoint[Any, Task]]) extends RequestStreamHandler {
```
johnduffell, line 20: this class has basically been rewritten according to the example
https://github.com/softwaremill/tapir/blob/master/serverless/aws/lambda-zio-tests/src/main/scala/sttp/tapir/serverless/aws/ziolambda/tests/ZioLambdaHandlerImpl.scala
with the addition of request and response stream logging

handlers/product-move-api/src/main/scala/com/gu/productmove/Secrets.scala
```diff
@@ -65,20 +62,11 @@ object Secrets {
   def getSalesforceSSLSecrets: RIO[Secrets, SalesforceSSLSecrets] =
     ZIO.environmentWithZIO[Secrets](_.get.getSalesforceSSLSecrets)
 }
-object SecretsLive extends Secrets {
+class SecretsLive(secretsClient: SecretsManagerClient) extends Secrets {
```
johnduffell, line 65: this is updated to match how the S3 client works, the credentials come from the AwsCredentialsLive.layer

handlers/product-move-api/src/test/scala/com/gu/productmove/zuora/InvoiceItemAdjustmentSpec.scala
```diff
@@ -84,7 +85,7 @@ object InvoiceItemAdjustmentSpec extends ZIOSpecDefault {
           ),
         )
         assert(adjustments.length)(equalTo(2)) &&
-          assert(adjustments.head.AdjustmentDate.getDayOfMonth)(equalTo(15))
+        assert(adjustments.head.AdjustmentDate.getDayOfMonth)(equalTo(15))
```
johnduffell, line 88: I don't agree with this, but scalafmt complains otherwise
```
[warn] scalafmt: ~/code/support-service-lambdas/handlers/product-move-api/src/test/scala/com/gu/productmove/zuora/InvoiceItemAdjustmentSpec.scala isn't formatted properly!
```

handlers/product-move-api/src/main/scala/com/gu/productmove/SQS.scala
rupertbates: Can you check? We obvs want to maintain the existing behaviour

johnduffell: yes that's a really good point I will check that for sure

johnduffell: yes, existing code we end up with `Any` as the failure type in the ZIO, and that ends up here in the code
https://github.com/guardian/support-service-lambdas/blob/1b7e0ee7131c016038bf2b2a25a26c86baf8836c/handlers/product-move-api/src/main/scala/com/gu/productmove/framework/ZIOApiGatewayRequestHandler.scala#L126
So no matter whether we called it InternalServerError or BadRequest etc, it it would still be an internal server error.

If/where we map the error to a ZIO.success then we could use it as a response.
I think the new (official) way of using Throwable rather than my way of using Any as the error type is better as it's pretty clear that a Throwable will be a 500 error, and if we try to use a BadRequest or something, expecting it will really be a bad request, it won't compile unless we recover it into a success.

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/available/AvailableProductMovesEndpoint.scala
```diff
@@ -74,14 +74,13 @@ object AvailableProductMovesEndpoint {
...
           |""".stripMargin)
-      .serverLogic[TIO] { subscriptionName =>
+      .serverLogic[Task] { subscriptionName =>
         run(subscriptionName).tapEither(result => ZIO.log("result tapped: " + result)).map(Right.apply)
       }
   }
 
-  // sub to test on: "A-S00334930"
-  private def run(subscriptionName: String): TIO[OutputBody] =
-    runWithEnvironment(SubscriptionName("A-S00334930")).provide(
```
johnduffell, line 84: this shouldn't have been hard coded, it should have used the one passed from the client!

PR #2081, opened 2023-10-19
call paperround during acquisition to get email data

johnduffell:
This PR follows on from https://github.com/guardian/support-service-lambdas/pull/2076 and is off the same branch

This makes us call paper round API to get the details that we need to send to braze for the thank you email.

It's a bit rough as I haven't really had time to tidy it.

It's not tested (beyond the little bits of test code that are there)

It also includes a new rest/form post client and some tweaks to how the config is loaded to match more modern scala standards.

PR #2116, opened 2023-11-29
Don't log HTTP client headers

paulbrown1982:
## What does this change?
Don't log HTTP client headers as these may contain secrets. Although this is safe in priciple due to access being equivalent to all tools in the AWS account (CloudWatch & Systems / Secrets Manager), we have no protection from some other developer in the department configuring our AMI base recipes to ship these logs somewhere out of our AWS account, thus creating a security incident.

## How to test

1. Everything still compiles and unit tests pass.

1. No headers sent to logs.

### DIFF
--- a/lib/effects/src/main/scala/com/gu/effects/Http.scala
+++ b/lib/effects/src/main/scala/com/gu/effects/Http.scala
@@ -28,7 +28,7 @@ object Http extends Logging {
     { request: Request =>
       val maybeBodySummary = Option(request.body).map(bodySummary)
       logger.info(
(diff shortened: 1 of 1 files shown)


johnduffell (approved):
this was introduced as part of https://github.com/guardian/support-service-lambdas/pull/2081 / https://github.com/guardian/support-service-lambdas/commit/2959f63fed068e99e49e82a90a3c79e0073bb57a#diff-b7d16b7ee4bc898f8696cbe9961b946850c4fb1bf3f390551c10cd2db7af4d53R31

PR #2117, opened 2023-11-29
Discount api

rupertbates:
## What does this change?
This PR updates the discount-api project which was just a stub previously to be a fully functioning api. There are a large number of new files as this is the first Typescript api in support-service-lambdas so there is a lot of generic code to handle things like stages, configuration, zuora access etc.

The initial purpose of this api is to enable us to apply discounts to existing digital subscriptions as a 'save' mechanism ie. to prevent users from cancelling. 
There are two endpoints `/preview-discount` and `/apply-discount`
### preview-discount
```shell
curl --location 'https://discount-api-code.support.guardianapis.com/preview-discount' \
--header 'x-api-key: API_GATEWAY_KEY' \
--data '{
    "subscriptionNumber": "A-S00752803",
    "discountProductRatePlanId": "2c92c0f962cec7990162d3882afc52dd"
}'
```
Returns
```json
{
    "valid": true,
    "discountedPrice": 11.24
}
```
Or
```json
{
    "valid": false
}
```
### add-discount
```shell
curl --location 'https://discount-api-code.support.guardianapis.com/apply-discount' \
--header 'x-api-key: API_GATEWAY_KEY' \
--data '{
    "subscriptionNumber": "A-S00752573",
    "discountProductRatePlanId": "2c92c0f962cec7990162d3882afc52dd"
}'
```
Returns a 200 with the body of `Success` or a 400 or 500 response with an error message in the body eg. 400 - `Subscription A-S00752803 has status Cancelled`


handlers/discount-api/src/catalog/catalog.ts
```diff
@@ -0,0 +1,63 @@
...
+} from './catalogSchema';
+import { catalogSchema } from './catalogSchema';
+
+const client = new S3Client(awsConfig);
+
+export async function getCatalogFromS3(stage: Stage): Promise<Catalog> {
+	console.log('getCatalogFromS3');
+	const command = new GetObjectCommand({
+		Bucket: 'gu-zuora-catalog',
+		Key: `PROD/Zuora-${stage}/catalog.json`,
```
andrewHEguardian, line 19: should this always be `PROD`? looks like it's staged within that but I find that a bit confusing

handlers/discount-api/src/catalog/catalog.ts
```diff
@@ -0,0 +1,63 @@
...
+	});
+
+	const response = await client.send(command);
+	const body = await response.Body?.transformToString();
+	if (!body) {
+		throw new Error(
+			'Response body was undefined when fetching the Catalog from S3',
+		);
+	}
+	return catalogSchema.parse(JSON.parse(body));
```
andrewHEguardian, line 29: what happens if we don't successfully parse it?

handlers/discount-api/src/eligibilityChecker.ts
```diff
@@ -0,0 +1,133 @@
...
+
+		const eligibleRatePlans: RatePlan[] = subscription.ratePlans.filter(
+			(ratePlan) =>
+				eligibleProductRatePlans?.includes(ratePlan.productRatePlanId),
+		);
+
+		if (eligibleRatePlans.length > 1) {
+			throw new Error(
+				`Subscription ${subscription.subscriptionNumber} has more than one eligible rate plan
+				 for ${discountProductRatePlanId}. I don't know whether this should be allowed so fail for now`,
```
andrewHEguardian, line 78: is this still TBD?

handlers/discount-api/src/catalog/catalog.ts
rupertbates: Yes that PROD is the stage of the lambda which does the extract the second stage is the Zuora stage it extracts from.

handlers/discount-api/src/catalog/catalog.ts
rupertbates: It throws an exception and we return a 500

handlers/discount-api/src/eligibilityChecker.ts
rupertbates: I can't think of any situation in which this would happen, I've just put this check in so that if it does we can investigate

handlers/discount-api/src/eligibilityChecker.ts
```diff
@@ -0,0 +1,133 @@
...
+			);
+		}
+
+		if (eligibleRatePlans.length === 0 || !eligibleRatePlans[0]) {
```
andrewHEguardian, line 82: just trying to understand this - so the array could be empty or have a null/undefined element in it?

handlers/discount-api/src/eligibilityChecker.ts
```diff
@@ -0,0 +1,133 @@
...
+		const chargePrices = this.catalog.getCatalogPriceOfCharges(
+			ratePlan.productRatePlanId,
+			currency,
+		);
+		const totalPrice = sum(chargePrices, (i) => i);
+
+		// Work out how much the cost of the next invoice will be
+		const billingPreview = await getBillingPreview(
+			this.zuoraClient,
+			dayjs().add(13, 'months'),
```
andrewHEguardian, line 110: why do we look 13 months into the future?

handlers/discount-api/src/zuora/billingPreview.ts
```diff
@@ -0,0 +1,44 @@
...
+
+	const groupedInvoiceItems = groupBy(billingPreview.invoiceItems, (item) =>
+		item.serviceStartDate.toISOString(),
+	);
+	const nextInvoiceItems = checkDefined(
+		groupedInvoiceItems[nextInvoiceDate],
+		'No invoice items found for next invoice date',
+	);
+
+	return [...nextInvoiceItems];
```
andrewHEguardian, line 43: why are you spreading it into an array?

handlers/discount-api/src/eligibilityChecker.ts
rupertbates: Typescript thinks so, there may well be a more elegant way to do this but I'm not sure what it is.

andrewHEguardian (approved):
Looks good 👍 👍 👏👏

handlers/discount-api/src/zuora/billingPreview.ts
rupertbates: I can't remember, I think it was originally an iterator but looking at it again now it's not needed

handlers/discount-api/src/eligibilityChecker.ts
rupertbates: Just to ensure that there will definitely be at least one invoice. It probably doesn't need to be so far in the future but I just sort by date and take the top one anyway so it doesn't really matter either way.

handlers/discount-api/src/eligibilityChecker.ts
andrewHEguardian: ah, it looks like the `noUncheckedIndexedAccess` flag in the TSconfig makes this the case

rupertbates: Yes it does but I still would have thought that the fact that we have ensured that the length of the array is 1 plus the fact that it's type is `RatePlan[]` would be enough that we wouldn't need it.

PR #2126, opened 2023-12-06
Don't try to cancel a non-active subscription

rupertbates:
## What does this change?
We are getting alarms for attempted cancellation of already cancelled subscriptions. I'm not sure how this is possible on the front end but this change validates these attempts and fails with a bad request (400) rather than an internal server error (500) so we will no longer receive alarms for them.


johnduffell:
nice work! glad you got back to this card 👏

PR #2129, opened 2023-12-11
Widen scope of test emails that get cancelled in Zuora CODE (and removed from Salesforce sandbox)

graham228221:
For a few months, we have noticed the rate that the Salesforce sandbox has been filling up has been increasing. More recently, the Conversions Team have introduced more end-to-end tests that would be adding more data to Zuora/Salesforce.

The dev-env-cleaner Lambda is intended to remove test data from the Salesforce Sandbox, by cancelling both subscriptions and accounts in Zuora. We found that it had a few issues that this PR addresses.

#### Changes the filter for both queries to start including test data generated from new [support-frontend playwright tests](https://github.com/guardian/support-frontend/tree/main/support-e2e/tests)

These new tests will create up to 30 new brand new subscriptions for each deployment on support-frontend, which would substantially increase the rate of test data accumulation in our Salesforce Sandbox.

The query is updated to account for these new tests. Previously it would only cancel accounts for `integration-test@thegulocal.com` and `test@thegulocal.com`. Now it will find and cancel any accounts matching emails `%@thegulocal.com` or `test%...%@theguardian.com`. This should pick up all test data from the support-frontend deployment tests, plus users running tests locally or using other `@thegulocal.com` test users.

#### For Customer Accounts with a credit balance, only removes the CRM Id instead of cancelling the account

The Lambda was getting stuck with Accounts with a Credit Balance, as Zuora does not allow these to be cancelled. To rectify this would require manual invoice adjustments to remove the Balance. 

Instead I've updated the Lambda to remove their CRM ID instead of cancelling the Account, if the Credit Balance is not 0. This will ensure that the associated data is still removed from Salesforce.

#### Run Lambda four times a day, instead of twice

Within the 15 minute time limit, the Lambda can do about 1,000 records (subs + customer accounts) depending on the response time from the Zuora requests. So 1,000 new customer accounts and subs can be cleaned in 24 hours

Running twice is probably fine, but with the support-frontend tests now adding 30 accounts/subscriptions per merge it's easy to imagine more than 1,000 being generated on a particularly busy day (imagine many dependabot/snyk PRs all being merged together).

The rate of test data accumulation in the Salesforce sandbox should be significantly reduced, leading to fewer instances of the Salesforce developers having to manually clear out data.

After testing these changes in CODE, the data storage in the Salesforce sandbox was reduced from 106% to 96% - the lowest it has been for a while (the hard upper limit is 110%):
Screenshot 2023-12-12 at 15 50 27

We can reduce this further with other clean up efforts, but the changes are clearly effective. 

This is only dealing with test data, so there is no risk to live customer data. 

There is a risk that required test data is inadvertently deleted, but developers can easily avoid this by avoiding using `@gulocal.com`. The lambda also only runs every 6 hours, so test data will persist for a while.

### DIFF
--- a/handlers/dev-env-cleaner/src/main/scala/com/gu/cleaner/Handler.scala
+++ b/handlers/dev-env-cleaner/src/main/scala/com/gu/cleaner/Handler.scala
@@ -8,6 +8,7 @@ import com.amazonaws.services.lambda.runtime._
 import com.gu.aws.AwsCloudWatch
 import com.gu.aws.AwsCloudWatch._
 import com.gu.cleaner.CancelAccount.CancelAccountRequest
+import com.gu.cleaner.RemoveAccountCrm.RemoveAccountCrmRequest
 import com.gu.cleaner.CancelSub._
 import com.gu.effects.{GetFromS3, RawEffects}
 import com.gu.util.config.{LoadConfigModule, Stage}
@@ -80,6 +81,7 @@ object Handler extends RequestStreamHandler {
       requests = ZuoraRestRequestMaker(response, zuoraRestConfig)
       cancelSub = CancelSub(log, requests)
       cancelAccount = CancelAccount(log, requests)
+      removeAccountCrm = RemoveAccountCrm(log, requests)
       downloadRequests = ZuoraAquaRequestMaker(downloadResponse, zuoraRestConfig)
       aquaQuerier = Querier.lowLevel(downloadRequests) _
       getJobResult = GetJobResult(downloadRequests.get[AquaJobResponse]) _
@@ -89,6 +91,7 @@ object Handler extends RequestStreamHandler {
         downloadRequests,
         cancelSub,
         cancelAccount,
+        removeAccountCrm,
         () => RawEffects.now().toLocalDate,
       )
     } yield ()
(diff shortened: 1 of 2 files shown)


david-pepper (approved):

handlers/dev-env-cleaner/src/main/scala/com/gu/cleaner/Handler.scala
```diff
@@ -128,22 +131,23 @@ class Steps(log: String => Unit) {
...
+      removeAccountCrm: RemoveAccountCrm,
       today: () => LocalDate,
   ): Either[Throwable, Unit] = {
     val subs_to_cancel = "subs_to_cancel"
     val subsQuery = AquaQuery(
       subs_to_cancel,
       """select Id, TermEndDate
         |from Subscription
-        |where (billtocontact.WorkEmail = 'integration-test@thegulocal.com' OR billtocontact.WorkEmail = 'test@thegulocal.com') and Status = 'Active' and account.Status = 'Active'
+        |where (billtocontact.WorkEmail LIKE '%@thegulocal.com' OR (billtocontact.WorkEmail LIKE 'test%' AND billtocontact.WorkEmail LIKE '%@theguardian.com')) and Status = 'Active' and account.Status = 'Active'
```
johnduffell, line 142: I know it's late but can't you do 
`billtocontact.WorkEmail LIKE 'test%@theguardian.com'`

graham228221: Ah totally, yes that works too! I was testing this bit-by-bit in ZOQL so I didn't clock this.

PR #2130, opened 2023-12-13
prevent disabling similar_guardian_products when responding to a cancellation notification

shtukas:
This implements a change in the way we process cancellations. We no longer remove all associated consents, but leave `similar_guardian_products` active.

For this, and to avoid affecting the delicate balance of the `ConsentsCalculator`, we simply give it a method which only remove `similar_guardian_products` from the set of consents that are going to be turned off. That method is then called just before `sendCancellationConsents`

### DIFF
--- a/handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/ConsentsCalculator.scala
+++ b/handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/ConsentsCalculator.scala
@@ -63,4 +63,14 @@ class ConsentsCalculator(consentsMappings: Map[String, Set[String]]) {
     consents.map(ConsentsObject(_, state)).asJson.toString()
   }
 
+  def removeSimilarGuardianProductFromSet(consents: Set[String]): Set[String] = {
+    // This method was added during https://github.com/guardian/support-service-lambdas/pull/2130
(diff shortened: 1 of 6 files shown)


rupertbates (approved):
As discussed, looks good to me from a code point of view.

shtukas:
Note to self: Do not touch until 2024. Requires confirmation from DP.

handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/ConsentsCalculator.scala
```diff
@@ -63,4 +63,14 @@ class ConsentsCalculator(consentsMappings: Map[String, Set[String]]) {
     consents.map(ConsentsObject(_, state)).asJson.toString()
   }
 
+  def removeSimilarGuardianProductFromSet(consents: Set[String]): Set[String] = {
```
rupertbates, line 66: What's the reason for removing a consent from the set, rather than just having a different set for cancellations? Is it so that we only have to keep one set up to date, or is there something else?

rupertbates (approved):

paulbrown1982 (approved):
👍 DP have agreed we can do this and that the logic within https://github.com/guardian/consent-autolapse takes control from today. No backfill.

PR #2133, opened 2023-12-18
Discount length in preview

rupertbates:
## What does this change?
This PR updates the discount api so that it returns the length that a discount will be applicable for in the preview request. 

The new response body will be similar to 
```json
{
    "discountedPrice": 11.24,
    "upToPeriods": 3,
    "upToPeriodsType": "Months"
}
```
`upToPeriodsType` comes directly from Zuora and can be `Months, Years, Days, Weeks` or `Billing_Periods`

It also changes the preview request behaviour so that an ineligible request returns a 400 response rather than a 200 with a boolean in the body.


andrewHEguardian (approved):
Looks good to me 👍

PR #2134, opened 2023-12-18
Check subscription belongs to the signed-in user

rupertbates:
## What does this change?
When checking the eligibility of a digital subscription user for a discount we want to ensure that the subscription we would be discounting actually belongs to the currently signed in user. This PR adds that check

### DIFF
--- a/handlers/discount-api/src/endpoints/discountEndpoint.ts
+++ b/handlers/discount-api/src/endpoints/discountEndpoint.ts
@@ -1,3 +1,4 @@
+import type { APIGatewayProxyEventHeaders } from 'aws-lambda';
 import dayjs from 'dayjs';
 import type { Stage } from '../../../../modules/stage';
 import { sum } from '../arrayFunctions';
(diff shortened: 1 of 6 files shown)


handlers/discount-api/src/endpoints/discountEndpoint.ts
```diff
@@ -38,12 +47,17 @@ export const discountEndpoint = async (
...
 
-	console.log('Getting billing preview for the subscription');
-	const billingPreview = await getBillingPreview(
-		zuoraClient,
-		dayjs().add(13, 'months'),
-		subscription.accountNumber,
+	console.log(
+		'Check sub is owned by user & get billing preview for the subscription',
 	);
+	const [, billingPreview] = await Promise.all([
```
rupertbates, line 53: These calls can happen in parallel as they don't depend on each other

andrewHEguardian (approved):
👍  looks good 🔒

PR #2136, opened 2023-12-20
Move reusable components into modules folder

rupertbates:
## What does this change?
Up until now we have only had one Typescript lambda and that has had all the functionality it needs (zuora client, accessing the catalog etc.) within it. This PR breaks reusable functionality out into a shared `modules` folder so that other lambdas can reuse that code.


andrewHEguardian (approved):
🎄🎁

johnduffell:
@rupertbates it looks like this caused snyk to pick up the typescript stuff, but it's not running the appropriate "npm install" type of command before hand.
I'm not sure what best practice is for snyk stuff, but I feel like the TS code isn't being checked (but the CDK looks like it is)
(snyk was further failing because of the Java 21 update which @MaelGNM and I fixed this morning)

PR #2141, opened 2024-01-02
Update TS lambda code generation templates

rupertbates:
## What does this change?
Support service lambdas has code generation templates written using [Hygen.io](https://www.hygen.io/) which enables the easy creation of a new Typescript lambda with an API Gateway frontend using a simple node script.

This PR makes the following changes to that mechanism:
- update the templates and prompt to optionally create a new api key and make the api gateway endpoint require that key for authentication
- add a test directory in the newly created lambda with placeholders for a unit test and an integration test
- add a `test` and `it-test` script to the new package.json
- improve several readme files


_templates/new-lambda/api-gateway/cdk-lib.ejs.t
```diff
@@ -11,6 +11,9 @@ import type { GuStackProps } from '@guardian/cdk/lib/constructs/core';
 import { GuStack } from '@guardian/cdk/lib/constructs/core';
 import type { App } from 'aws-cdk-lib';
 import { Duration } from 'aws-cdk-lib';
+<% if (includeApiKey === 'Y'){ %>
```
rupertbates, line 14: If the user has decided that the new lambda should have api key authentication

tomrf1 (approved):
👍 looks good

PR #2145, opened 2024-01-12
Fix the code which works out which discount to apply to a given sub

rupertbates:
## What does this change?
We have a bug in the discount-api code which tries to work out which discount to apply to the provided subscription.

The way it does this is to look at the rate plans on the sub to work out the billing period and then pick either the monthly or annual discount based on that. 

However for annual subscriptions which have a discount applied (even if that discount is expired) we have been picking the discount rate plan when trying to work out the billing period. The billing period of the discount is monthly so we have been picking the wrong discount.

The eligiblilty check then fails because the monthly discount cannot be applied to annual subs and the subscription is flagged as ineligible for a discount.

## The fix
This PR just introduces a check that the rate plan we are looking at does not belong to the `Discounts` product. We will need a more flexible and robust way of doing this in future but that will involve building a lot of code which knows about different product types so is out of scope at the moment.


andrewHEguardian (approved):
Great find 👍

PR #2146, opened 2024-01-15
Fix the discount api eligibility check for price risen subscriptions

rupertbates:
## What does this change?
Fixes bug with error message 'Subscription A-S0XXXXXXXX has more than one eligible rate plan` in the digital subscription serve journey.

The issue was caused by the way the price rise process works - it uses a Zuora `update` call to remove the old digisub product rate plan and then re-add it at which point the new catalog price is picked up. However when this sub is then fetched from Zoura by the discount-api, both rate plans are present in the response leading to the error above.

The fix for this is to filter out rate plans which have a `lastChangeType` of `Remove` when looking for the digital sub rate plan.


andrewHEguardian (approved):
👍

PR #2148, opened 2024-01-17
Explicitly specify currentTermPeriodType for discount preview

rupertbates:
## What does this change?
Follows on from  #2147 - that PR added a fix for trying to apply discounts outside the current term of a subscription by extending the current term to the next billing date. However it did this for all subs, even ones where the next billing date is not outside the current term. This is uncecessary so this PR changes the behaviour so that we only change the term when we need to.


modules/zuora/src/addDiscount.ts
```diff
@@ -58,6 +71,7 @@ export const previewDiscount = async (
 		// 14 days after the contract effective date (acquisition date/when the term begins) to provide a free
 		// trial period, for annual subs in particular the next billing date is going to be outside the current term.
 		currentTerm: 24,
+		currentTermPeriodType: 'Month',
```
rupertbates, line 74: It is safer to specify the `currentTermPeriodType` we want in the call to Zuora rather than assuming it is `Month`

PR #2149, opened 2024-01-25
Introduce a Typescript module to handle modelling our Product catalog

rupertbates:
This PR adds a model to describe Guardian products in Typescript and to map between that model and the Zuora catalog.

Readme description included below:

----------------
# Product Module
This module defines a model to describe our product structure and the mapping between that structure and the Zuora catalog which holds pricing information.

There are three main types involved in our product definitions:
## `ProductFamilyKey` 
This represents a family or category of products which we sell, available values are 
- `GuardianWeekly`
- `Newspaper`
- `Digital`
## `ZuoraProductKey`
This maps onto a particular product defined in the Zuora catalog. Each ZuoraProductKey belongs to a product family so it is a generic type: 
```typescript
type ZuoraProductKey<PF extends ProductFamilyKey>
``` 
available values for each product family key are:
#### Newspaper:
- `NationalDelivery`
- `HomeDelivery`
- `SubscriptionCard`
#### GuardianWeekly
- `Domestic`
- `RestOfWorld`
#### Digital
- `DigitalSubscription`
- `SupporterPlus`
- `Contribution`

## `ProductRatePlanKey`
This maps to an individual product rate plan. Each ProductRatePlanKey belongs to a specific product family and zuora product so it is also a generic type with the signature:
```typescript
type ProductRatePlanKey<
	PF extends ProductFamilyKey,
	ZP extends ZuoraProductKey<PF>,
>
```
Examples of product options for particular product families and Zuora products are:

#### Newspaper
- `HomeDelivery`
  - `Saturday`
  - `Sunday`
  - `Weekend`
  - `Sixday`
  - `Everyday`
- `NationalDelivery`
  - `Weekend`
  - `Sixday`
  - `Everyday` // National delivery doesn't have a Saturday and Sunday option

#### Digital
- `DigitalSubscription`
  - `Monthly`
  - `Annual`
  - `OneYearGift`
  - `ThreeMonthGift`
- `SupporterPlus`
  - `Monthly`
  - `Annual`

This diagram shows the mapping between this type model and the Zuora catalog:
Screenshot 2024-01-30 at 18 15 38

# Usage
By providing a `ProductFamilyKey`, `ZuoraProductKey` and `ProductRatePlanKey` we can map to any [product rate plan](https://knowledgecenter.zuora.com/Zuora_Central_Platform/API/G_SOAP_API/E1_SOAP_API_Object_Reference/ProductRatePlan) in the Zuora catalog and from that we can retrieve the pricing information for that particular configuration. For instance if I want to find the GBP price to get the Newspaper delivered on a Saturday I can use the following:
```typescript
import { getZuoraCatalog } from '@modules/catalog/catalog';
import { getProductRatePlan } from '@modules/product/productCatalogMapping';

const stage = 'CODE';
const catalog = await getZuoraCatalog(stage);

const productRatePlan = getProductRatePlan(
    stage,
    'Newspaper',
    'HomeDelivery',
    'Saturday',
);
const price = catalog.getCatalogPrice(productRatePlan.id, 'GBP');
// price is 19.99

```
The way that the key types `ProductFamilyKey`, `ZuoraProductKey` and `ProductRatePlanKey` are defined in a dependent hierarchy ensures that we can only use options which are appropriate for the given product

```typescript
import {getProductRatePlanId} from "./productToCatalogMapping";

const productRatePlan = getProductRatePlan(
    'CODE',
    'GuardianWeekly',
    'HomeDelivery', // TS2345: Argument of type "Digital" is not assignable to parameter of type "RestOfWorld" | "Domestic"
    'Monthly',
);
```
The product rate plan object also holds information about the charges that plan contains, and that charge info is also strongly typed to the particular product variant.
```typescript
import {getProductRatePlanId} from "./productToCatalogMapping";

const mondayRatePlanChargeId = getProductRatePlan(
      'CODE', 
      'Newspaper', 
      'NationalDelivery', 
      'Everyday'
).charges.Monday

// If we try to access a charge which doesn't exist we will get a compile error 
const sundayRatePlanChargeId = getProductRatePlan(
        'CODE',
        'Newpaper',
        'NationalDelivery',
        'Sixday'
).charges.Sunday // TS2339: Property Sunday does not exist on type


```
## Implementation
The mapping between our object model and Zuora's catalog is defined in two json files `codeCatalogMapping.json` and `prodCatalogMapping.json`.  These are actually generated from the zuora catalog itself using the functions in `catalogMappingGeneration.ts`. 

Whenever the catalog changes these json files can be regenerated by running the command
```shell
pnpm --filter product generateMapping
```
from the root of the repository.


modules/product/src/product.ts
```diff
@@ -0,0 +1,113 @@
...
+		case 'Contribution':
+			return ['Digital'] as const;
+		case 'Newspaper':
+			return newspaperDeliveryOptions;
+		case 'GuardianWeekly':
+			return guardianWeeklyDeliveryOptions;
+	}
+};
+
+export const getProductOptionsForProduct = (product: Product) => {
```
AndreaDiotallevi, line 100: Minor comment - would it be better to use an object rather than switch statements?

modules/product/src/productToCatalogMapping.ts
```diff
@@ -0,0 +1,343 @@
...
+	[P in Product]: {
+		[DO in DeliveryOption<P>]: {
+			[PO in ProductOption<P>]: { productRatePlanId: string };
+		};
+	};
+};
+
+const productOptionNotAvailable =
+	'Product option not available for this product';
+const codeProductToCatalogMapping: ProductToCatalogMapping = {
```
AndreaDiotallevi, line 25: This is gonna be super useful for other projects as well.
I wonder if it is a use case for becoming a npm package? Or maybe it's overkill.

modules/product/src/productToCatalogMapping.ts
```diff
@@ -0,0 +1,343 @@
...
+	DigitalSubscription: {
+		Digital: {
+			Monthly: {
+				productRatePlanId: '2c92c0f84bbfec8b014bc655f4852d9d',
```
AndreaDiotallevi, line 29: Does it mean that every time we change the Zuora catalogue, we need to update those ids manually?
I wonder whether we should place these objects in their own files and auto-generate them (maybe in the CI?) with the Zuora API.

modules/product/src/productToCatalogMapping.ts
rupertbates: It might be worth it, ideally I think things which need to use it should live in this monorepo, but I'm sure there will be some that can't.

modules/product/src/productToCatalogMapping.ts
rupertbates: I think you're right, a next step would be to find a better way to get hold of these from the catalogue but I think that should be a separate PR.

AndreaDiotallevi (approved):

jamesgorrie (approved):
This all looks good. I wonder if there would be some value for clients by exporting the values and types, but maybe we can look at adding these while integrating this lib. 

Before you'd committed the data/type mapping, [I had a go at being  expressive of what would pop out the other side for the client](https://gist.github.com/jamesgorrie/8e01f854c5b4975d0b22e0bdfadf988a). Not a suggestion but thought it was interesting.

modules/product/src/productToCatalogMapping.ts
rupertbates: I have added a script now to enable us to regenerate these objects using 
```shell
pnpm --filter product generateMapping
```

modules/product/src/productCatalogMapping.ts
```diff
@@ -289,54 +247,64 @@ const prodProductToCatalogMapping: ProductToCatalogMapping = {
...
+	PF extends ProductFamilyKey,
+	ZP extends ZuoraProductKey<PF>,
+> = keyof (typeof prodMapping)[PF][ZP];
+type ProductRatePlanObject = { productRatePlanId: string };
 
-export const getProductRatePlanId = <T extends ProductFamily>(
+export const getProductRatePlanId = <
+	P extends ProductFamilyKey,
+	ZP extends ZuoraProductKey<P>,
+	PRP extends ProductRatePlanKey<P, ZP>,
```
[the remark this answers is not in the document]
rupertbates: Yes I think it could be. It would be good if we could standardise on these as a product definition across all our TS codebases. I'll maybe look to how we could do that as a next step.

PR #2153, opened 2024-02-02
build: create new application "salesforce-disaster-recovery"

AndreaDiotallevi:
## What does this change?

[Trello](https://trello.com/c/UJCw9g4x/581-create-salesforce-disaster-recovery-project-with-initial-state-machine-cdk)

- Create new application called `salesforce-disaster-recovery`
- Create new CDK stack called `SalesforceDisasterRecovery` with a placeholder state machine + lambda
- Create C4 model diagram for the new project using the Structurizr DSL

## Why are you doing this?

This is the first PR for the KR1 of the [Q4 FY 23/24 OKR - Disaster Recovery for Salesforce](https://docs.google.com/document/d/1UFDM33Yhl0cgHcIDIfWIkq8V7ZiYNc32wy0SGdfmILo/edit#heading=h.d6f3idk6a29j).

landscape

## Design document

This PR implements points 1 & 2 of the solution outlined in this [document](https://docs.google.com/document/d/1_KxFtfKU3-3-PSzaAYG90uONa05AVgoBmyBDyu5SC5c/edit#heading=h.jyzfkcwrdmx7).

## Next

Implement first state `CreateSalesforceQueryJob`.


cdk/lib/salesforce-disaster-recovery.ts
```diff
@@ -0,0 +1,57 @@
...
+
+export class SalesforceDisasterRecovery extends GuStack {
+	constructor(scope: App, id: string, props: GuStackProps) {
+		super(scope, id, props);
+
+		const app = 'salesforce-disaster-recovery';
+		const runtime = Runtime.NODEJS_20_X;
+		const fileName = `${app}.zip`;
+		const timeout = Duration.seconds(300);
+		const memorySize = 1024;
```
rupertbates, line 17: Timout and memory are probably fine for this lambda but when you get to the ones that do the heavy lifting I'd increase them both.

rupertbates (approved):
Looks good, great PR description.

cdk/lib/salesforce-disaster-recovery.ts
AndreaDiotallevi: Good point. I will note this down for when I'll actually start using lambdas.

PR #2155, opened 2024-02-06
Add billing period to new-product-api

rupertbates:
## What does this change?
As part of the work to enable Salesforce to apply discounts to subscriptions at the point of acquisition we need to provide the billing period and price of payment plans in the product catalog endpoint of the new-product-api.


JoeMitchellGuardian (approved):
Looks good to me! Thanks for the quick turnaround on this!

PR #1945, opened 2023-04-28
introduce error response types

jonflynng:
## Accessibility

-   [ ] [Tested with screen reader](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#screen-reader)
-   [ ] [Navigable with keyboard](https://github.com/guardian/accessibility/blob/main/people-and-technology/02-physical.md#keyboard)
-   [ ] [Colour contrast passed](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#contrast)
-   [ ] [The change doesn't use only colour to convey meaning](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#use-of-colour)

PR #2156, opened 2024-02-06
add x-identity-id check to product move lambda plus ZIO refactoring

johnduffell:
This PR updates product-move-api to check x-identity-id header from manage-frontend, and make sure the specifiied subscription id matches with the identity id.  This will secure against any privilege escalation attack being possible against these endpoints.

I've also taken the opportunity to do some tidyup as the old home-grown AWS-lambda -> tapir -> zio did not support HTTP headers, and there is an official one now.  That meant updating the tapir version, and some other tidyups also followed (stopping using the ZIO environment as it just confuses people, and other scala 3 syntax improvements)

Testing in CODE:
- S+ to recurring contribution and back again - https://apisandbox.zuora.com/platform/subscriptions/8ad097048ebd45bc018ec227abb148fc
- cancellation of S+ with reason
- amount change S+


build.sbt
```diff
@@ -145,7 +145,6 @@ lazy val `holiday-stops` = library(project in file("lib/holiday-stops"))
       playJson,
       scalatest,
       scalaCheck,
-      playJsonExtensions,
```
johnduffell, line 148: this library  is no longer maintained and  was causing a packaging conflict.  Although the packagine conflict could have been patched, it leaves us open to binary incompatibilieis.  It turned out that most of the time we were using features that play-json has natively, and only one place actually benefited from it, so I removed it completely and rewrote the place that needed it in a less good way.

build.sbt
```diff
@@ -608,7 +606,7 @@ lazy val `product-move-api` = lambdaProject(
       s"aws lambda update-function-code --function-name product-switch-salesforce-tracking-$stage --s3-bucket $s3Bucket --s3-key $s3Path --profile membership --region eu-west-1".!!
     }
   }
-  .dependsOn(`zuora-models`, `effects-sqs`, `new-product-api`)
```
johnduffell, line 611: this dependency should not really have been there, lambda handler projects should only depend on lib projects.  I just copied the small amount of code that was depended on, and that shaves quite a bit off the jar size and removes a lot of dependencies from the classpath.

handlers/batch-email-sender/src/main/scala/com/gu/batchemailsender/api/batchemail/BrazeSqsMessage.scala
```diff
@@ -74,12 +73,51 @@ case class BrazeSqsMessage(
...
-  implicit val stoppedCreditSummaryFormat = Jsonx.formatCaseClass[StoppedCreditSummary]
-  implicit val digitalVoucherFormat = Jsonx.formatCaseClass[DigitalVoucher]
-  implicit val emailPayloadSubscriberAttributesWriter = Jsonx.formatCaseClass[BrazeApiTriggerProperties]
-  implicit val emailPayloadContactAttributesWriter = Jsonx.formatCaseClass[EmailPayloadContactAttributes]
-  implicit val emailPayloadToWriter = Jsonx.formatCaseClass[EmailPayloadTo]
-  implicit val emailToSendWriter = Jsonx.formatCaseClass[BrazeSqsMessage]
+  implicit val stoppedCreditSummaryFormat = Json.writes[StoppedCreditSummary]
+  implicit val digitalVoucherFormat = Json.writes[DigitalVoucher]
+
+  // has to be done manually due to the 22 field limit
```
johnduffell, line 79: play-json-extensions helped us out here in the past.  Luckily there are a few unit tests, so I could check I hadn't misnamed or missed any fields.  This is going to be more annoying to maintain, but overall it's a win.

handlers/cancellation-sf-cases-api/src/test/scala/com/gu/cancellation/sf_cases/RaiseCaseTest.scala
```diff
@@ -0,0 +1,37 @@
+package com.gu.cancellation.sf_cases
+
+import com.gu.cancellation.sf_cases.RaiseCase._
+import org.scalatest.EitherValues
+import org.scalatest.flatspec.AnyFlatSpec
+import org.scalatest.matchers.should.Matchers
+import play.api.libs.json.Json
+
+class RaiseCaseTest extends AnyFlatSpec with Matchers with EitherValues {
```
johnduffell, line 9: just added a quick test to make sure play-json deals with value classes in the same way as play-json-extensions did

handlers/product-move-api/src/main/scala/com/gu/productmove/AwsCredentialsLive.scala
```diff
@@ -16,7 +16,7 @@ object AwsCredentialsLive {
   val layer: Layer[Throwable, AwsCredentialsProvider] =
     ZLayer.scoped(ZIO.fromAutoCloseable(ZIO.attempt(impl)))
 
-  private def impl: AwsCredentialsProviderChain =
+  def impl: AwsCredentialsProviderChain =
```
johnduffell, line 19: this PR moves us away from ZIO Layers and towards normal parameters/classes.  This means exposing the actual values publically, and deprecating the ZLayer or removing it completely where possible.  Hopefully we can complete this cleanup more thoroughly over time.
This does have the disadvantage of losing resource safety i.e. things won't be closed down properly when the lambda becomes idle.

handlers/product-move-api/src/main/scala/com/gu/productmove/AwsS3Live.scala
```diff
@@ -27,9 +25,9 @@ object AwsS3Live {
 
 }
 
-private class AwsS3Live(s3Client: S3Client) extends AwsS3:
+class AwsS3Live(s3Client: S3Client) extends AwsS3:
 
-  override def getObject(bucket: String, key: String): ZIO[Any, ErrorResponse, String] =
+  override def getObject(bucket: String, key: String): Task[String] =
```
johnduffell, line 30: the lambdas have a model (where we are the server) where they return either a success or an `ErrorResponse` (meaning 4xx or 5xx)
For some reason this model was reused inappropriately in cases where we were the client, so if AWS calls failed, they would return some kind of error which you may expect to get returned directly, but usually they would get converted to a 500 error and logged.  This was confusing and inappropriate, so I have changed it so we just use the ZIO `Task` type, which uses a normal `Throwable` as the error type.

This has the advantage of totally decluttering the code, and also meaning it's very clear that any upstream unexpected errors will be logged and returned as a 5xx.

handlers/product-move-api/src/main/scala/com/gu/productmove/SQS.scala
```diff
@@ -4,170 +4,190 @@ import com.gu.effects.sqs.AwsSQSSend.EmailQueueName
...
 
 trait SQS {
-  def sendEmail(message: EmailMessage): ZIO[Any, ErrorResponse, Unit]
+  def sendEmail(message: EmailMessage): Task[Unit]
 
-  def queueRefund(refundInput: RefundInput): ZIO[Any, ErrorResponse, Unit]
+  def queueRefund(refundInput: RefundInput): Task[Unit]
 
-  def queueSalesforceTracking(salesforceRecordInput: SalesforceRecordInput): ZIO[Any, ErrorResponse, Unit]
+  def queueSalesforceTracking(salesforceRecordInput: SalesforceRecordInput): Task[Unit]
```
johnduffell, line 29: this class is very big, I'm sure all three of these don't need to be in the same class, but that's not something for this PR

handlers/product-move-api/src/main/scala/com/gu/productmove/SQS.scala
```diff
@@ -4,170 +4,190 @@ import com.gu.effects.sqs.AwsSQSSend.EmailQueueName
...
-      stage <- ZIO.service[Stage]
-      sqsClient <- initializeSQSClient().mapError(ex =>
-        new Throwable(s"Failed to initialize SQS Client with error", ex),
-      )
-      emailQueueName = EmailQueueName.value
-      emailQueueUrlResponse <- getQueue(emailQueueName, sqsClient)
-      refundQueueUrlResponse <- getQueue(s"product-switch-refund-${stage.toString}", sqsClient)
-      salesforceTrackingQueueUrlResponse <- getQueue(s"product-switch-salesforce-tracking-${stage.toString}", sqsClient)
-    } yield new SQS {
-      override def sendEmail(message: EmailMessage): ZIO[Any, ErrorResponse, Unit] =
```
johnduffell, line 58: the changes here are mostly indentation to make it into a normal class with parameters rather than an object that returns an anonymous class

handlers/product-move-api/src/main/scala/com/gu/productmove/SQS.scala
```diff
@@ -181,6 +201,7 @@ case class toRCEmailPayloadProductSwitchAttributes(
     currency: String,
     subscription_id: String,
 ) extends EmailPayloadAttributes
+    derives JsonEncoder
```
johnduffell, line 204: in scala 3 you can just write `derives JsonEncoder` or whatever, rather than writing a separate given/implicit

handlers/product-move-api/src/main/scala/com/gu/productmove/Secrets.scala
```diff
@@ -72,22 +69,17 @@ class SecretsLive(secretsClient: SecretsManagerClient) extends Secrets {
     secretsClient.getSecretValue(GetSecretValueRequest.builder().secretId(secretId).build()).secretString()
   }
 
-  def getStage: Task[String] =
-    ZIO.succeed(sys.env.getOrElse("Stage", "CODE"))
```
johnduffell, line 76: this wouldn't have picked up the mocked stage in any tests, so changed to a parameter to match other clients.

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/available/AvailableProductMovesEndpoint.scala
```diff
@@ -228,8 +219,9 @@ object AvailableProductMovesEndpoint {
       _ <- ZIO.log("done")
     } yield AvailableMoves(moveToProduct)
 
-    output.catchAll { failure =>
-      ZIO.succeed(failure)
+    output.catchAll {
+      case failure: OutputBody => ZIO.succeed(failure)
+      case other: Throwable => ZIO.fail(other)
```
johnduffell, line 224: we used to insert .mapErrorTo500 into the for comprehension every time we got back a throwable.
With scala 3 union types, you can have the error type as `OutputBody | Throwable` and then deal with it here at the end.
This declutters the important logic in the for comprehension, but still allows shortcutting in the case of both errors and early completed responses.

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/cancel/SubscriptionCancelEndpoint.scala
```diff
@@ -93,6 +91,7 @@ object SubscriptionCancelEndpoint {
               .copy(description = Some("Information to describe the nature of the cancellation")),
           ),
         )
+        .in(header[IdentityId]("x-identity-id"))
```
johnduffell, line 94: this is telling tapir that identity id is required, it will put the value in the case class for us at runtime

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/cancel/SubscriptionCancelEndpoint.scala
```diff
@@ -115,146 +114,42 @@ object SubscriptionCancelEndpoint {
...
+    stage <- ZIO.service[Stage]
+    zuoraSetCancellationReason <- ZIO.service[ZuoraSetCancellationReason]
+    res <- new SubscriptionCancelEndpointSteps(
+      getSubscription,
+      getAccount,
+      getSubscriptionToCancel,
+      zuoraCancel,
+      sqs,
+      stage,
+      zuoraSetCancellationReason,
```
johnduffell, line 137: now SubscriptionCancelEndpointSteps is a normal class with some parameters, rather than a function that returns a ZIO that has to have loads of "Layers" provided

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/cancel/SubscriptionCancelEndpointSteps.scala
```diff
@@ -0,0 +1,172 @@
+package com.gu.productmove.endpoint.cancel
```
johnduffell, line 1: this was moved out from the above SubscriptionCancelEndpoint.scala, note that it has quite a few changes to do the identity id check and to remove all the ZIO Layers

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/cancel/SubscriptionCancelEndpointSteps.scala
```diff
@@ -0,0 +1,172 @@
...
+      postData: ExpectedInput,
+      identityId: IdentityId,
+  ): Task[OutputBody] = {
+    val maybeResult: IO[OutputBody | Throwable, Success] = for {
+      subscriptionAccount <- assertSubscriptionBelongsToIdentityUser(
+        getSubscription,
+        getAccount,
+        subscriptionName,
+        Some(identityId),
+      )
```
johnduffell, line 48: this is the important new line.  If the request is naughty, the function will return an HTTP error to the client.

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/cancel/SubscriptionCancelEndpointSteps.scala
```diff
@@ -0,0 +1,172 @@
...
+          ZIO.succeed(RefundResponse("Success", ""))
+
+      _ <- ZIO.log(s"Attempting to update cancellation reason on Zuora subscription")
+      _ <- zuoraSetCancellationReason
+        .update(
+          subscriptionName,
+          subscription.version + 1,
+          postData.reason,
+        ) // Version +1 because the cancellation will have incremented the version
+      _ <- sqs.sendEmail(EmailMessage.cancellationEmail(subscriptionAccount._2, cancellationDate))
```
johnduffell, line 110: now the code uses the account that was fetched during the identity check above, rather than fetching a fresh copy.  This means the extra check won't affect the performance as much.

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/move/ProductMoveEndpoint.scala
```diff
@@ -1,109 +1,101 @@
...
       ExpectedInput(1, false, None, None),
+      None,
     ),
   )
 
   val server: sttp.tapir.server.ServerEndpoint.Full[
     Unit,
     Unit,
-    (String, String, ProductMoveEndpointTypes.ExpectedInput),
+    (SwitchType, SubscriptionName, ProductMoveEndpointTypes.ExpectedInput, Option[IdentityId]),
```
johnduffell, line 60: this file has some general tidy up but the main points are
- the identity id is optional, as it can be called by manage and salesforce.
- the two strings now have proper types so that once they're in the door the compiler won't let you mix them up

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/move/ProductMoveEndpointSteps.scala
```diff
@@ -0,0 +1,81 @@
...
+    getSubscription: GetSubscription,
+    getAccount: GetAccount,
+    subscriptionName: SubscriptionName,
+    maybeIdentityId: Option[IdentityId],
+): IO[Throwable | BadRequest, (GetSubscription.GetSubscriptionResponse, GetAccount.GetAccountResponse)] =
+  for {
+    subscription <- getSubscription.get(subscriptionName)
+    account <- getAccount.get(subscription.accountNumber)
+    _ <- (maybeIdentityId, account.basicInfo.IdentityId__c) match
+      case (None, _) => ZIO.unit
```
johnduffell, line 28: on review, if it's a request from salesforce, we don't actually need to fetch the subscription or account here.  We do still need to make them available as we return them to the caller.  I wonder if it's worth the speedup for CSRs for it only to fetch if necessary here OR if the caller needs it.

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/move/ProductMoveEndpointSteps.scala
```diff
@@ -0,0 +1,81 @@
...
+    maybeResult.catchAll(recoverErrors)
+
+end ProductMoveEndpointSteps
+
+// update a subscription, integration error https://developer.zuora.com/api-references/api/operation/PUT_Subscription/
+private val PutSubscriptionIntegrationError = 53500099
+
+def recoverErrors(err: OutputBody | Throwable): Task[OutputBody] = err match {
+  case failure: OutputBody => ZIO.succeed(failure)
+  case ZuoraClientError(_, reasons) if reasons.exists(_.code == PutSubscriptionIntegrationError) =>
```
johnduffell, line 71: there was a funny bit of logic in the zuora client to turn that error code into a TransactionError which would hopefully get all the way back to our client as the right HTTP error code.
This created a bit of a cyclical dependency (Zuora client shouldn't know about the error response types of an upstream lambda)
I changed it so the zuora client simply returns the error codes and then this lambda can decide to look out for what it wants and convert it at this point.

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/move/switchtype/GetRatePlans.scala
```diff
@@ -0,0 +1,116 @@
+package com.gu.productmove.endpoint.move.switchtype
```
johnduffell, line 1: this code was pulled out from RecurringContributionToSupporterPlus as it was pretty standalone and that file was so big that github isn't really happy to show the diff

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/move/switchtype/GetRatePlans.scala
```diff
@@ -0,0 +1,116 @@
...
+      catalogPlanId: PlanId,
+      currency: Currency,
+  ): Task[AmountMinorUnits] =
+    for {
+      productRatePlanId <- ZIO
+        .fromEither(
+          zuoraIdsForStage(config.Stage(stage.toString)).map(_.apiIdToRateplanId(catalogPlanId)),
+        )
+        .mapError(new Throwable(_))
+      prices <- getCatalogue.get
```
johnduffell, line 85: this now uses the (existing) GetCatalogue functionality rather than depending on new-product-api

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/move/switchtype/RecurringContributionToSupporterPlus.scala
```diff
@@ -94,27 +33,35 @@ case class ProductSwitchRatePlanIds(
...
-      with GetAccount
-      with SQS
-      with Dynamo
-      with Stage,
-    OutputBody,
-  ] = {
-    (for {
+      subscription: GetSubscriptionResponse,
+      account: GetAccountResponse,
+  ): Task[OutputBody] =
```
johnduffell, line 62: the code here has been heavily de-zio-`Layer`ed but the main change is it takes the subscription and account as parameters, as they have already been fetched during the identity id check.  This avoids two identical calls to zuora.

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/move/switchtype/ToRecurringContribution.scala
```diff
@@ -1,55 +1,51 @@
...
+    subscriptionUpdate: SubscriptionUpdate,
+    sqs: SQS,
+    stage: Stage,
+) extends ToRecurringContribution {
+  def run(
+      subscriptionName: SubscriptionName,
+      postData: ExpectedInput,
+      subscription: GetSubscriptionResponse,
+      account: GetAccountResponse,
+  ): Task[OutputBody] =
```
johnduffell, line 41: similar to the above file, this has been de-`Layer`ed and account and subscription that were fetched in the identity check are now passed in to avoid double hitting zuora

handlers/product-move-api/src/main/scala/com/gu/productmove/framework/RequestMapper.scala
```diff
@@ -1,62 +0,0 @@
-package com.gu.productmove.framework
-
-import com.amazonaws.services.lambda.runtime.events.APIGatewayV2WebSocketEvent
-import sttp.model.Uri
-import sttp.tapir.serverless.aws.lambda.{AwsHttp, AwsRequest, AwsRequestContext}
-
-import scala.jdk.CollectionConverters.*
-
-object RequestMapper {
```
johnduffell, line 9: the official tapir ZIO lambda library has the equivalent functionality that's more complete, so rather than adding something to handle headers so we can get x-identity-id, I removed this and use the official code.

handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/rest/ZuoraClient.scala
```diff
@@ -68,14 +61,13 @@ private class ZuoraClientLive(baseUrl: Uri, sttpClient: SttpBackend[Task, Any],
           )
           .copy(uri = absoluteUri),
       )
-      .mapError(e => InternalServerError(e.toString))
-      .map(_.body.left.map(e => InternalServerError(e)))
+      .map(_.body.left.map(e => new Throwable("response failure: " + e)))
```
johnduffell, line 64: the case where the request is failed with an exception now passes through , the case where it succeeds but we get back a non 2xx (body.left) turns the response into a Throwable, and in the case where response is 200 OK but the field success=false, it will return the new `ZuoraClientError` exception

handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/rest/ZuoraGet.scala
```diff
@@ -5,27 +5,27 @@ import com.gu.productmove.zuora.rest.ZuoraRestBody.ZuoraSuccessCheck
 import sttp.client3.basicRequest
 import sttp.model.Uri
 import zio.json.*
-import zio.{IO, ULayer, URLayer, ZIO, ZLayer}
+import zio.{IO, RIO, Task, ULayer, URLayer, ZIO, ZLayer}
 
 object ZuoraGetLive:
```
johnduffell, line 10: this has get/post/etc in it so in future it should be renamed to something more generic

handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/rest/ZuoraGet.scala
```diff
@@ -5,27 +5,27 @@ import com.gu.productmove.zuora.rest.ZuoraRestBody.ZuoraSuccessCheck
...
   val layer: URLayer[ZuoraClient, ZuoraGet] = ZLayer.fromFunction(ZuoraGetLive(_))
 
-private class ZuoraGetLive(zuoraClient: ZuoraClient) extends ZuoraGet:
+class ZuoraGetLive(zuoraClient: ZuoraClient) extends ZuoraGet:
 
-  override def get[T: JsonDecoder](
+  override def get[Response: JsonDecoder](
```
johnduffell, line 15: `Response` is still a type parameter but it's one that's a bit more descriptive than just `T`

handlers/product-move-api/src/test/scala/com/gu/productmove/HandlerSpec.scala
```diff
@@ -135,75 +143,82 @@ object HandlerSpec extends ZIOSpecDefault {
...
-          ZLayer.succeed(new MockCreatePayment(CreatePaymentResponse(Some(true)))),
-          ZLayer.succeed(Stage.valueOf("CODE")),
-        )
-      } @@ TestAspect.ignore, // TODO: make the code which fetches the catalog price a dependency so it can be mocked
+          assert(mockGetInvoiceItems.requests)(equalTo(List("89ad8casd9c0asdcaj89sdc98as"))) &&
+          assert(mockInvoiceItemAdjustment.requests)(equalTo(List(invoiceItemAdjustmentInputs))) &&
+          assert(mockSQS.requests)(hasSameElements(List(emailMessageLowCharge, salesforceRecordInput4))) &&
+          assert(mockDynamo.requests)(equalTo(List(supporterRatePlanItem2)))
+        }
+      } @@ TestAspect.ignore,
```
johnduffell, line 182: these probably should be unignored now that the catalog can be mocked, but since they haven't been run in the build, there's something else needs fixing there

handlers/product-move-api/src/test/scala/com/gu/productmove/endpoint/cancel/SubscriptionCancelEndpointStepsSpec.scala
```diff
@@ -0,0 +1,98 @@
...
+        identityOfMMAUser = "baddieIdentity",
+      ).map(wasCalled => assertTrue(!wasCalled))
+    },
+    test("can't get through if no identity id in zuora") {
+      runCheck(
+        identityInZuora = None,
+        identityOfMMAUser = "baddieIdentity",
+      ).map(wasCalled => assertTrue(!wasCalled))
+    },
+  )
```
johnduffell, line 39: this is the important test to make sure it's only possible to get through with the right identity id

handlers/product-move-api/src/test/scala/com/gu/productmove/endpoint/move/ProductMoveEndpointSpec.scala
```diff
@@ -0,0 +1,116 @@
...
+        identityInZuora = Some("johnIdentity"),
+        identityOfMMAUser = Some("johnIdentity"),
+      ).map(wasCalled => assertTrue(wasCalled))
+    },
+    test("can get through from SF (zuora should always have an id)") {
+      runCheck(
+        identityInZuora = Some("johnIdentity"),
+        identityOfMMAUser = None, // called from SF
+      ).map(wasCalled => assertTrue(wasCalled))
+    },
```
johnduffell, line 43: this is the extra test needed for this endpoint to make sure saleforce/csrs can get through

handlers/product-move-api/src/test/scala/com/gu/productmove/zuora/ProductMoveSpec.scala
```diff
@@ -1,85 +1,113 @@
...
+import zio.test.Assertion.*
+import zio.test.*
 
-import java.time._
+import java.time.*
 
-object ProductMoveSpec extends ZIOSpecDefault {
+object ProductMoveSpec {
 
-  override def spec: Spec[TestEnvironment with Scope, Any] =
```
johnduffell, line 17: there were quite a few of these "tests" to just run the lambda locally so I have updated them to just be normal main method which just removes a bit of the boilerplate.

handlers/sf-gocardless-sync/src/main/scala/com/gu/sf_gocardless_sync/SyncSharedObjects.scala
```diff
@@ -1,37 +1,37 @@
 package com.gu.sf_gocardless_sync
 
-import ai.x.play.json.Jsonx
+import play.api.libs.json.Json
 
 object SyncSharedObjects {
 
   case class GoCardlessMandateID(value: String) extends AnyVal
-  implicit val formatMandateID = Jsonx.formatInline[GoCardlessMandateID]
+  implicit val formatMandateID = Json.valueFormat[GoCardlessMandateID]
```
johnduffell, line 8: this is exactly the same thing only it comes from the core library rather than a third party extension library

lib/zuora-models/src/main/scala/com/gu/newproduct/api/addsubscription/validation/supporterplus/AmountLimits.scala
```diff
@@ -56,6 +56,7 @@ object AmountLimits {
     case NZD => nzd
     case CAD => cad
     case EUR => eur
+    case _ => ???
```
johnduffell, line 59: warning.  This will still fail at runtime, but at least it's not a compiler warning about non exchastive match

project/Dependencies.scala
```diff
@@ -120,6 +119,7 @@ object Dependencies {
      * So it's redundant in a binary artefact.
      */
     case PathList("codegen-resources", _*) => MergeStrategy.discard
+    case PathList("META-INF", "FastDoubleParser-LICENSE") => MergeStrategy.discard
```
johnduffell, line 122: there was an assembly error due to a shaded version in AWS client and a non shaded version of this  outside.  Licence files are probably not needed in the artifact jar that we use internally

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/move/ProductMoveEndpointSteps.scala
graham228221: I was going to ask if we should be checking whether this would disrupt the requests from Salesforce, but sounds like this is all accounted for already. Thanks!

handlers/product-move-api/src/main/scala/com/gu/productmove/AwsS3Live.scala
rupertbates: I think the original idea behind this was that the place where the error happened was where you had the context to decide if it was a 4xx or 5xx, how are you making that decision now?

handlers/product-move-api/src/main/scala/com/gu/productmove/Secrets.scala
```diff
@@ -96,46 +88,37 @@ class SecretsLive(secretsClient: SecretsManagerClient) extends Secrets {
...
 
-  def getInvoicingAPISecrets: Task[InvoicingAPISecrets] = {
-    for {
-      stg <- getStage
-      secretId: String = s"${stg}/InvoicingApi"
-      secretJsonString = getJSONString(secretId)
-      secrets <- parseInvoicingAPISecretsJSONString(secretJsonString)
-    } yield secrets
-  }
+  def getInvoicingAPISecrets: Task[InvoicingAPISecrets] =
```
rupertbates, line 91: Do we really want to go to significant whitespace? It was the thing that basically made me give up on Scala!

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/available/AvailableProductMovesEndpoint.scala
```diff
@@ -93,8 +92,6 @@ object AvailableProductMovesEndpoint {
       SecretsLive.layer,
     )
 
-  private val freeTrialDays = 14
```
rupertbates, line 96: Not sure what this was doing, but why is it no longer needed?

handlers/product-move-api/src/main/scala/com/gu/productmove/Secrets.scala
johnduffell: yeah agreed I don't mind it for short functions but it really messes up copy and paste etc for longer ones. In that case we need to turn it off in the compiler as I think intellij is happy to allow it otherwise.  I'll check what the incantation is.  Didn't know it was that much of a deal breaker for you!

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/available/AvailableProductMovesEndpoint.scala
johnduffell: good question - looking back at the commit that added it, it wasn't used in the first place
https://github.com/guardian/support-service-lambdas/commit/46b8412fc2c62ae0fc4d494a26b152c0c135353a#diff-e76bd6b369bfd0a16fdef7e2cfb73c4da147ebaa575035ec0a2092a36dd5375bR69

handlers/product-move-api/src/main/scala/com/gu/productmove/AwsS3Live.scala
johnduffell: That's interesting because in general would have gone the other way around - the full context comes from what the code is trying to achieve.  So may or may not be enough context at the low level.

in practice in this codebase, every low level code was just catching all exceptions and converting them to InternalServerError immediately. (The only one that didn't is redundant, see [1] below)

In terms of my proposal devs are free to decide at any point the response code, but if they want 5xx it removes all the `recover...` and `ZIO[....]` boilerplate:
- if we simply want a 5xx error use Throwable (i.e. `Task[A]`). This covers all of our low level code, clients etc.
- if we don't have enough context to decide yet, use a subclass of Throwable,containing useful information (i.e. `Task[A]`).  This would cover the redundant case I mentioned above.
- if we know what to return, add ErrorResponse via union types ( `IO[Throwable | ErrorResponse, A]`). This covers the top level function and its helper functions.
Any residual uncaught throwables just get converted to InternalServerError at the top level.

Options in place of this PR would be :

1. (no change to existing) keep the existing ErrorResponse and the lines of code to recover every throwable into an InternalServerError etc. (good if we have many non 5xx lambda response codes needed to be decided in the lower level code - http clients, AWS clients etc)
2. revert back to String as it was before the PR mentioned below, since we don't collect payments during product switching any more (`String` is simpler than `ErrorResponse`)
3. (this PR) all failures are Throwables unless we have enough context, use ErrorResponse once we know what response we want to return (good if the higher level code - handler etc - generally catches then decides on response codes and content, and everything else is a 5xx)
4. the support-workers ErrorHandler approach - make all responses a Throwable https://github.com/guardian/support-frontend/blob/main/support-workers/src/main/scala/com/gu/support/workers/exceptions/ErrorHandler.scala#L27-L35 (to save needing a union type.  I'm not sure if tapir[2] would handle that.)
5. something else?

----

[1]
The only difference from that (and the reason those types were introcued https://github.com/guardian/support-service-lambdas/pull/1945 ) was the ZuoraClient which checked all responses for 53500099 (which means "integration error during PUT subscription request" i.e. payment failed) and returned TransationError which maps to http 430.
See https://github.com/guardian/support-service-lambdas/pull/2156/files#diff-d7b7f0a738adb5daee9fc36f3ee2d6c696b8503488370a42f5b34da207e85a6aL128

In fact, that code is redundant because we don't collect a small payment during product switching any more.

I did keep the redundant code just to avoid making this PR even bigger. 

[2] tapir: Having said that we're not really taking advantage of the generated openAPI.yaml docs or the built in swagger Live API explorer etc for tapir https://product-move-api.support.guardianapis.com/docs/ It might have been a nice alternative to PostMan if it wasn't totally separate.

handlers/product-move-api/src/main/scala/com/gu/productmove/Secrets.scala
johnduffell: rewritten in this pr: https://github.com/guardian/support-service-lambdas/pull/2222

rupertbates: It was just the fact that Scala 3 was supposed to simplify and standardise writing Scala, and then at the last minute they decide to make such a massive change to the language in the face of a lot of opposition for no discernable benefit. It was just the last straw for me - was like the story of the scorpion and the frog!

rupertbates (approved):
Might have been good to split the security work and the refactor into two separate PRS?

johnduffell:
> Might have been good to split the security work and the refactor into two separate PRs?

yes in fact there probably should have been 3
1. using the official zio-tapir-lambda instead of our own home grown
2. de-`ZLayer`ing
3. the actual identity id check (depends on 1 but not 2)

In the long term I'd like to get rid of ZIO in this lambda as I the only time I've been clear on the benefits is when I've just had it explained to me.

PR #2157, opened 2024-02-07
feat: implement state to create Salesforce query job

AndreaDiotallevi:
## What does this change?

[Trello](https://trello.com/c/W9IMHSNM/584-create-salesforce-query-job)

Implement the first machine state to create a Salesforce query job for retrieving the new IDs.

Screenshot 2024-02-06 at 16 01 47

### Details

  - Create a connection from the AWS console for both [CODE](https://eu-west-1.console.aws.amazon.com/events/home?region=eu-west-1#/connections/salesforce-disaster-recovery-CODE-salesforce-api) and [PROD](https://eu-west-1.console.aws.amazon.com/events/home?region=eu-west-1#/connections/salesforce-disaster-recovery-PROD-salesforce-api) which handles the Salesforce authentication and secret creation

  - Reference the connection ARN in the state machine

  - Configure API url and parameters directly in the state machine IAC (no need for a lambda!)

  ```typescript
stateJson: {
        Type: 'Task',
        Resource: 'arn:aws:states:::http:invoke',
        Parameters: {
                ApiEndpoint: `${props.salesforceApiDomain}/services/data/v59.0/jobs/query`,
                Method: 'POST',
                Authentication: {
	                ConnectionArn: salesforceApiConnectionArn,
                },
                RequestBody: {
	                operation: 'query',
	                'query.$': '$.query',
                },
        },
},
  ```

  - The state expects a required input to filter Salesforce accounts
  ```json
  {
    "query": "SELECT Id, Zuora__Zuora_Id__c, Zuora__Account__c, Contact__c from Zuora__CustomerAccount__c"
  }
  ```

- Rename previously created lambda to `placeholder-lambda` for future use cases

## Why are you doing this?

This is the second PR for the KR1 of the [Q4 FY 23/24 OKR - Disaster Recovery for Salesforce](https://docs.google.com/document/d/1UFDM33Yhl0cgHcIDIfWIkq8V7ZiYNc32wy0SGdfmILo/edit#heading=h.d6f3idk6a29j).

landscape

## Design document

This PR implements point 3 of the solution outlined in this [document](https://docs.google.com/document/d/1_KxFtfKU3-3-PSzaAYG90uONa05AVgoBmyBDyu5SC5c/edit#heading=h.jyzfkcwrdmx7).

## How to test

You can test the CODE state machine from the AWS console [here](https://eu-west-1.console.aws.amazon.com/states/home?region=eu-west-1#/statemachines/view/arn%3Aaws%3Astates%3Aeu-west-1%3A865473395570%3AstateMachine%3Asalesforce-disaster-recovery-CODE).

Note - Learn [how to test step functions](https://docs.aws.amazon.com/step-functions/latest/dg/sfn-local.html).

## Previous

- [https://github.com/guardian/support-service-lambdas/pull/2153](https://github.com/guardian/support-service-lambdas/pull/2153)

## Next

- [https://github.com/guardian/support-service-lambdas/pull/2158](https://github.com/guardian/support-service-lambdas/pull/2158)


SarahJessica (approved):
Looks good. Well done Andrea 😄

johnduffell:
> Note - Learn [how to test step functions](https://docs.aws.amazon.com/step-functions/latest/dg/sfn-local.html).

I like this, one of my main complaints about adding more logic to step functions in the past has been the inability to unit test or even run the logic without pushing to AWS each time and poking through the console.  Perhaps this goes some way towards making it more testable without too much rigmarole!

cdk/lib/salesforce-disaster-recovery.ts
```diff
@@ -30,28 +41,85 @@ export class SalesforceDisasterRecovery extends GuStack {
 			environment,
 		};
 
-		const createSalesforceQueryJobLambda = new GuLambdaFunction(
+		new GuLambdaFunction(this, 'PlaceholderLambda', {
```
rupertbates, line 44: Could you get rid of this now? Or are you going to use it in later steps?

rupertbates:
This is a really interesting approach, have you checked what the output looks like if the request to SF fails for whatever reason?

rupertbates (approved):

cdk/lib/salesforce-disaster-recovery.ts
AndreaDiotallevi: Good point.
I have removed the IAC for the lambda, but I had to keep the script that generates the zip file since it is required by guardian/actions-riff-raff@v2.
It is probably conceptually wrong that we still need to have a zip event when we don't need it, so we could potentially add a conditional to the Github actions yaml to handle this, but it is probably for a separate PR.

AndreaDiotallevi:
> This is a really interesting approach, have you checked what the output looks like if the request to SF fails for whatever reason?

Good question!

For the moment I have only used Retry and it just returns the error status code, which you can use to handle retries and catch states.

Screenshot 2024-02-09 at 12 00 26

If you Catch the error instead, you should be able to access a Cause object with more details, but I haven't tested yet. I will let you know once I get there.

[Link](https://docs.aws.amazon.com/step-functions/latest/dg/concepts-error-handling.html#error-handling-error-output)

PR #2160, opened 2024-02-13
DRAFT

AndreaDiotallevi:
DRAFT


handlers/salesforce-disaster-recovery/package.json
```diff
@@ -10,6 +10,11 @@
 		"type-check": "tsc --noEmit"
 	},
 	"devDependencies": {
-		"@types/aws-lambda": "^8.10.129"
+		"@types/aws-lambda": "^8.10.129",
+		"aws-sdk-client-mock": "^3.0.1"
```
jamesgorrie, line 14: As we'll need these in CI - what's the reason for them being `devDependencies`? My understanding is that we the only difference is that you can install `prod-only` is you want to save filesize, but I think these are required?

Not massively related to this PR, just curious.

handlers/salesforce-disaster-recovery/src/services/salesforce.ts
```diff
@@ -0,0 +1,66 @@
...
+	};
+}
+
+export const generateSalesforceAccessToken = async ({
+	credentials,
+}: {
+	credentials: SalesforceOauthCredentials;
+}): Promise<string> => {
+	try {
+		const formData = new URLSearchParams([
```
jamesgorrie, line 16: Potentially a confusingly named `const` as [`URLSearchParams`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) are similar, but different to [`FormData`](https://developer.mozilla.org/en-US/docs/Web/API/FormData).

I think in this instance it might be more correct to use `FormData` - super interesting to know it works irrespective. Ah, Web APIs.

handlers/salesforce-disaster-recovery/test/services/s3.test.ts
```diff
@@ -0,0 +1,39 @@
+import { PutObjectCommand, S3Client } from '@aws-sdk/client-s3';
+import { mockClient } from 'aws-sdk-client-mock';
```
jamesgorrie, line 2: Not used these before, they look super interesting.

jamesgorrie (approved):
Does what it says on the tin, and does it nicely!

cdk/bin/cdk.ts
```diff
@@ -105,6 +105,8 @@ new SalesforceDisasterRecovery(app, 'salesforce-disaster-recovery-CODE', {
 	salesforceApiDomain: 'https://gnmtouchpoint--dev1.sandbox.my.salesforce.com',
 	salesforceApiConnectionResourceId:
 		'salesforce-disaster-recovery-CODE-salesforce-api/c8d71d2e-9101-439d-a3e2-d8fa7e6b155f',
+	salesforceOauthSecretName:
+		'events!connection/salesforce-disaster-recovery-CODE-salesforce-api/e2792d75-414a-48f3-89a1-5e8eac15f627',
```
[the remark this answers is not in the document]
AndreaDiotallevi: Good question!

Those are the secret IDs automatically generated and handled by AWS after you create a [connection](https://eu-west-1.console.aws.amazon.com/events/home?region=eu-west-1#/connections/salesforce-disaster-recovery-CODE-salesforce-api).

The events!connection/* prefix is the internal convention AWS has for those generated secrets.

I believe they decided to prefix it with the word "events" since the connections are used by [api destinations](https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-api-destinations.html) which get given an event as input.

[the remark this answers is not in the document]
AndreaDiotallevi: Worth to note that, since I could not use the http:invoke feature, I had to use a lambda.
And I didn't want to re-create the secrets for the OAuth.
So I re-used those ones.

handlers/salesforce-disaster-recovery/package.json
AndreaDiotallevi: That's an interesting question.

I believe `esbuild` just packages the files that are used by the handler.
So in theory it should already exclude for instance the `aws-sdk-client-mock` package since it is not used.

In general, I think separating dependencies between dev and prod is a good practice in Node projects - although it does not seem to have any practical effect in this case.

In the past I have worked on a Node project with Docker and the flow was:
1. install everything
2. run tests
3. remove dev dependencies and keep only prod

Actually I have written a [blog post](https://www.andreadiotallevi.com/blog/how-to-create-a-production-image-for-a-node-typescript-app-using-docker-multi-stage-builds) about this!

handlers/salesforce-disaster-recovery/src/services/salesforce.ts
AndreaDiotallevi: Thank you - I have replaced it with the proper FormData().

handlers/salesforce-disaster-recovery/test/services/s3.test.ts
AndreaDiotallevi: I agree! It was super easy to make it work.

Here is an interesting [article](https://aws.amazon.com/blogs/developer/mocking-modular-aws-sdk-for-javascript-v3-in-unit-tests/).

cdk/bin/cdk.ts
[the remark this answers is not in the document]
AndreaDiotallevi: I see your point.

Usually when you pass environment ARNs or IDs in the CDK, it is implicit that you are referencing an "external" resource.

But I can see that it can be unclear in this case since it uses a different pattern.

Do you suggest it's clearer to recreate the secret for the lambda functions, rather than re-using the one created by the connection implicitly by AWS?

I am happy to amend it if so.

PR #2161, opened 2024-02-15
update play-json to scala 3 compatible version and remove unsupported…

johnduffell:
This updates us to a scala 3 compatible version of play json, it also removes an unsupported library which has not been updated for years.

### DIFF
--- a/handlers/cancellation-sf-cases-api/src/test/scala/com/gu/cancellation/sf_cases/RaiseCaseTest.scala
+++ b/handlers/cancellation-sf-cases-api/src/test/scala/com/gu/cancellation/sf_cases/RaiseCaseTest.scala
@@ -1,0 +1,37 @@
+package com.gu.cancellation.sf_cases
+
+import com.gu.cancellation.sf_cases.RaiseCase._
+import org.scalatest.EitherValues
+import org.scalatest.flatspec.AnyFlatSpec
+import org.scalatest.matchers.should.Matchers
+import play.api.libs.json.Json
+
+class RaiseCaseTest extends AnyFlatSpec with Matchers with EitherValues {
+
+  it should "deserialise input" in {
+
+    val testString = "Membership"
+
+    val jsonData = s""""$testString""""
+
+    val expected = ProductName(testString)
+    val actual = Json.fromJson[ProductName](Json.parse(jsonData))
+
+    actual.asEither.value shouldEqual expected
+
+  }
+
+  it should "serialise class" in {
+
+    val testString = "Membership"
+
+    val expected = s""""$testString""""
+
+    val testData = ProductName(testString)
+    val actual = Json.stringify(Json.toJson(testData))
+
+    actual shouldEqual expected
+
+  }
+
+}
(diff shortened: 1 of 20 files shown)


lib/zuora-core/src/test/scala/com/gu/zuora/subscription/SubscriptionDataEchoLegacyIntegrationTest.scala
```diff
@@ -47,7 +47,7 @@ class SubscriptionDataEchoLegacyIntegrationTest extends AnyFlatSpec {
       subscriptionFile = "EchoLegacySubscription.json",
       startDate = startDate,
       expectedIssueData = expectedIssueData,
-      expectedTotalAnnualIssueLimitPerSubscription = 12,
+      expectedTotalAnnualIssueLimitPerSubscription = 10,
```
johnduffell, line 50: these two tests have been failing for a while due to these tests not being run in the automated build since we switched to assembling the jars individually in GHA instead of doing them all in one build in teamcity
We should probably find a way to get these running again.

handlers/cancellation-sf-cases-api/src/test/scala/com/gu/cancellation/sf_cases/RaiseCaseTest.scala
```diff
@@ -0,0 +1,37 @@
...
+    val expected = s""""$testString""""
+
+    val testData = ProductName(testString)
+    val actual = Json.stringify(Json.toJson(testData))
+
+    actual shouldEqual expected
+
+  }
+
+}
```
jamesgorrie, line 37: I can't really tell if this is testing any functionality of ours, or just checking the Play JSON API works?

jamesgorrie (approved):
It looks as though the project is using a mixture of Play JSON and Circe for decoding. I might be wrong as that's at initial glance. 

If we are - should we potentially look at choosing one?

handlers/cancellation-sf-cases-api/src/test/scala/com/gu/cancellation/sf_cases/RaiseCaseTest.scala
johnduffell: it was making sure that we still serialise value classes in the same way as we did before.  So it's more like checking we are using the right play json API, rather than testing play json itself!
I did think about putting an entire request in and deserialising the whole thing and checking it all, as we don't seem to have that test, but I went for the slimmed down version.
If you think a full test on the actual request payload would be better instead or in addition I can add another test maybe in another PR?

johnduffell:
> It looks as though the project is using a mixture of Play JSON and Circe for decoding. I might be wrong as that's at initial glance.
> 
> If we are - should we potentially look at choosing one?

yes and there's some upickle in there too  I think. And now there's even some typescript in there too!  This repo has suffered from being very flexible and letting people implement new lambdas in a new way each time, which always sounds like a good thing at the time, but there's no obvious point where old things are brought up to date.
Perhaps with some better integration testing and more time spent refactoring we could keep things more consistent, I would fully support that.

PR #2162, opened 2024-02-16
Test users

rupertbates:
## What does this change?
This PR adds a module which allows creation, cancellation and deletion of test subscriptions in CODE Zuora from the command line.

Usage examples from the README:
### Create a subscription
```shell
pnpm -filter test-users createSubscription
```
Result will be: 
```json
[
  {
    Success: true,
    SubscriptionNumber: 'A-S00815422',
    AccountNumber: 'A00826604'
  }
]
```
This will create a monthly digital subscription
### Cancel a subscription
```shell
pnpm -filter test-users cancelSubscription A-S00815422
```
Result will be:
```json
{ success: true }
```

### Delete an account
```shell
pnpm -filter test-users deleteAccount A00826604
```
Result will be:
```json
{ success: true }
```

### DIFF
--- a/modules/test-users/README.md
+++ b/modules/test-users/README.md
@@ -1,5 +1,5 @@
 ## Test Users
-This module provides an easy way to create, cancel and delete test subscriptions and accounts in CODE Zuora from the command line.
+This module provides an easy way to create, cancel and delete test subscriptions and accounts in CODE Zuora from the command line. You will need to have fresh Janus credentials.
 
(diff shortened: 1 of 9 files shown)


rupertbates:
> I have tested the cli commands and all works nicely.
> 
> I wonder whether we should add an optional command line parameter (or any other option) to be able to specify CSBX as well?

Yes I think there are lots of things we could add, this should just be a start

PR #2166, opened 2024-02-22
Product catalog lambda

rupertbates:
## Accessibility


handlers/generate-product-catalog/src/index.ts
```diff
@@ -0,0 +1,35 @@
...
+		Key: `${stage}/product-catalog.json`,
+		ContentType: 'application/json',
+		Body: JSON.stringify(productCatalog, null, 2),
+	});
+
+	try {
+		const response = await client.send(command);
+		console.log('Response from S3 was ', response);
+	} catch (err) {
+		console.error('An error occurred while writing the catalog to S3: ', err);
```
jamesgorrie, line 33: Do we alert on these errors?

modules/arrayFunctions.ts
```diff
@@ -36,3 +36,9 @@ export const getSingleOrThrow = <T>(
 
 export const findDuplicates = <T>(array: T[]) =>
 	array.filter((item, index) => array.indexOf(item) !== index);
+export const distinct = <T>(array: T[]) => Array.from(new Set(array));
+export const arrayToObject = <T>(array: Array<Record<string, T>>) => {
```
jamesgorrie, line 40: I wonder if this could be renamed to something that describes it? I mention it as it's taken me quite a while to figure that out.

From what I can tell, it's taking an array of objects (as the types say) then taking the key of each element in each object and setting it as a top level value on a new object.

Is this more like a `flatten`?

modules/product-catalog/src/generateProductCatalog.ts
```diff
@@ -0,0 +1,96 @@
...
+	return allPrices.reduce<PricingObject>((acc, price) => {
+		const total = (acc[price.currency] ?? 0) + price.price;
+		acc[price.currency] = roundPriceToTwoDecimalPlaces(total);
+		return acc;
+	}, {});
+};
+const getProductRatePlanCharges = (
+	productRatePlanCharges: ZuoraProductRatePlanCharge[],
+) => {
+	return arrayToObject(
```
jamesgorrie, line 40: Wondering if this would be simpler and less steps to follow with a reduce, [as you do with the prices](https://github.com/guardian/support-service-lambdas/pull/2166/files#diff-5de6b5e9c1b138278c692011beb5220687ea9e2b3f17bf9dc692d16b65384d88R31)?

```ts
productRatePlanCharges.reduce((acc, productRatePlanCharge) => {
  const productRatePlanChargeName = getProductRatePlanChargeKey(productRatePlanCharge.name);

  acc[productRatePlanChargeName]: { id: productRatePlanCharge.id }
})
```

modules/product-catalog/src/generateTypeObject.ts
```diff
@@ -0,0 +1,94 @@
...
+				productRatePlanCharge.name,
+			);
+			return {
+				[productRatePlanChargeName]: {},
+			};
+		}),
+	);
+};
+const getCurrenciesForProduct = (productRatePlan: ZuoraProductRatePlan) =>
+	distinct(
```
jamesgorrie, line 33: When do we expect duplicates?

modules/zuora-catalog/test/catalog.test.ts
```diff
@@ -0,0 +1,20 @@
+import { ZuoraCatalogHelper } from '@modules/zuora-catalog/zuoraCatalog';
+import type { ZuoraCatalog } from '@modules/zuora-catalog/zuoraCatalogSchema';
+import { zuoraCatalogSchema } from '@modules/zuora-catalog/zuoraCatalogSchema';
+import code from './fixtures/catalog-code.json';
+import prod from './fixtures/catalog-prod.json';
+
+test('catalogSchema', () => {
+	const codeCatalog: ZuoraCatalog = zuoraCatalogSchema.parse(code);
+	expect(codeCatalog.products.length).toEqual(39);
```
jamesgorrie, line 9: Are these safe, especially as we might start to have a more flexible catalog? We could as `toBeGreaterThanOrEqual`?

modules/zuora-catalog/test/catalog.test.ts
```diff
@@ -0,0 +1,20 @@
...
+
+	const prodCatalog: ZuoraCatalog = zuoraCatalogSchema.parse(prod);
+	expect(prodCatalog.products.length).toEqual(19);
+});
+
+test('getCatalogPriceForCurrency', () => {
+	const codeCatalog = new ZuoraCatalogHelper(zuoraCatalogSchema.parse(code));
+	const digiSubAnnualId = '2c92c0f94bbffaaa014bc6a4212e205b';
+	const price = codeCatalog.getCatalogPrice(digiSubAnnualId, 'GBP');
+	expect(price).toEqual(149);
```
jamesgorrie, line 19: Similar to the above - this feels like we're tying ourself to potentially dynamic data. It wouldn't be hard to change, but it feels like a change in an external system shouldn't break these tests.

modules/product-catalog/src/api.ts
```diff
@@ -0,0 +1,22 @@
...
+		stage === 'PROD'
+			? 'https://product-catalog.guardianapis.com/product-catalog.json'
+			: 'https://product-catalog.code.dev-guardianapis.com/product-catalog.json';
+	const response = await fetch(url, {
+		method: 'GET',
+	});
+
+	const json = await response.json();
+	if (response.ok) {
+		return productCatalogSchema.parse(json) as ProductCatalog;
```
jamesgorrie, line 16: Do we need type coercion here as the parsing should return it in that type?

modules/product-catalog/src/generateTypeObject.ts
```diff
@@ -0,0 +1,94 @@
...
+	return arrayToObject(arrayVersion);
+};
+
+const writeTypesToFile = async () => {
+	const prodCatalog = await getZuoraCatalogFromS3('PROD');
+	const types = generateTypeObject(prodCatalog);
+	const typesString = JSON.stringify(types, null, 2);
+
+	fs.writeFileSync(
+		'./src/typeObject.ts',
```
jamesgorrie, line 87: Thinking - this name is quite vague. Maybe something like `ProductCatalogWithNoPricing` (really not sure about that but it's descriptive)

jamesgorrie (approved):
All look good. A couple of questions around gaps in my knowledge, and a suggestion of simplifying the transforming by not using `arrayToObject`, but reduce similarly to other areas in the code base.

handlers/generate-product-catalog/src/index.ts
rupertbates: No, good idea though.

modules/product-catalog/src/api.ts
rupertbates: I'll move this to specifying the return type on the function explicitly. We do need to use the shorthand somewhere or we just get the zod type which is the same but much more verbose.

modules/product-catalog/src/generateTypeObject.ts
rupertbates: Because there are multiple charges on lots of product rate plans and we're just taking the currency from them all, there will be dupes

modules/zuora-catalog/test/catalog.test.ts
rupertbates: Same as above

modules/product-catalog/src/generateTypeObject.ts
rupertbates: The thing is it's only there to use to construct types. I'd like to get rid of it all together really but I couldn't think of a way to do so that didn't involve loads more code so I've left it.

PR #2167, opened 2024-02-22
when testing a handler, also run tests in all depended on modules

johnduffell:
Further to the previous PR, it turns out that since the switch to GHA, we haven't been running any of the "lib" tests in the build.  See https://github.com/guardian/support-service-lambdas/pull/2161#discussion_r1491370947

This PR fixes it up so when you test a handler (or lib) it will run all depended on tests at the same time.


build.sbt
```diff
@@ -65,8 +65,20 @@ val testSettings = inConfig(EffectsTest)(Defaults.testTasks) ++ inConfig(HealthC
...
+    scalaSettings: SettingsDefinition = scala2Settings,
+) =
+  theProject
+    .settings(scalaSettings, testSettings, scalafmtSettings)
+    .configs(EffectsTest, HealthCheckTest)
+    .dependsOn(deps *)
+    .settings(
+      Test / test := ((Test / test) dependsOn (deps.map(cp => cp.project / Test / test) *)).value,
+      Test / testOnly := ((Test / testOnly) dependsOn (deps.map(_.project / Test / test) *)).evaluated,
+      Test / testQuick := ((Test / testQuick) dependsOn (deps.map(_.project / Test / test) *)).evaluated,
```
johnduffell, line 80: these are the important additions to make it run the depended on modules' tests

build.sbt
```diff
@@ -318,47 +364,61 @@ def lambdaProject(
       riffRaffArtifactResources += (file(s"handlers/$projectName/$cfName"), s"cfn/$cfName"),
       dependencyOverrides ++= jacksonDependencies,
       libraryDependencies ++= dependencies ++ logging,
+      Test / test := ((Test / test) dependsOn (deps.map(_.project / Test / test) *)).value,
+      Test / testOnly := ((Test / testOnly) dependsOn (deps.map(_.project / Test / test) *)).evaluated,
+      Test / testQuick := ((Test / testQuick) dependsOn (deps.map(_.project / Test / test) *)).evaluated,
```
johnduffell, line 369: the test task chaining needs to happen in both the libs and the handlers, so there's a small amount of duplication

build.sbt
```diff
@@ -318,47 +364,61 @@ def lambdaProject(
...
 }
 
-// FIXME: This seems to be non-standard
-// FIXME: Why is the name in sub-project build.sbt support-service-lambda
-// FIXME: Why is riff-raff not refering to CF?
-lazy val `zuora-callout-apis` = library(project in file("handlers/zuora-callout-apis"))
-  .enablePlugins(RiffRaffArtifact)
-  .settings(libraryDependencies += scalaMock)
-  .dependsOn(zuora, handler, effectsDepIncludingTestFolder, `effects-sqs`, testDep)
+lazy val `zuora-callout-apis` = lambdaProject(
```
johnduffell, line 373: I fixed this up to be standard, rather than applying the test task dependency separately

build.sbt
```diff
@@ -298,12 +342,14 @@ def lambdaProject(
     projectName: String,
     projectDescription: String,
     dependencies: Seq[sbt.ModuleID] = Nil,
+    deps: Seq[ClasspathDependency] = Nil,
```
tjmw, line 345: I wonder if there's a name for this new arg which would make a clearer distinction between `dependencies` and `deps` which sound very similar? (Not sure that it would be though - naming things is hard!)

build.sbt
```diff
@@ -318,47 +364,61 @@ def lambdaProject(
...
+    testDep,
+  ),
 )
 
 lazy val `digital-subscription-expiry` = lambdaProject(
   "digital-subscription-expiry",
   "check digital subscription expiration for authorisation purposes",
   Seq(contentAuthCommon),
-).dependsOn(zuora, handler, effectsDepIncludingTestFolder, testDep)
+  Seq(zuora, handler, effectsDepIncludingTestFolder, testDep),
```
tjmw, line 403: When we have a dependency which is shared by multiple SBT projects (for example, `zuora` here is shared by multiple SBT projects), will the tests for that dependency end up being run multiple times (once per consumer)?

tjmw (approved):
This looks good to me! (Though I'm not super well versed in SBT).

For my own knowledge - is this pattern a common one in multi project builds?

build.sbt
johnduffell: good point, I've changed it
```scala
    externalDependencies: Seq[sbt.ModuleID] = Nil,
    projectDependencies: Seq[ClasspathDependency] = Nil,
```

build.sbt
johnduffell: that's a good point.- the answer is that yes it will during the automated build, as we would expect
but I guess the downside is that if you run `sbt test` locally it might run them once in their own right and once for each handler or other project.
I wonder if we should have a key called `testIncludingDeps` and get the build to call that, but leave the original test task untouched?

johnduffell:
> is this pattern a common one in multi project builds?

I don't know! I just asked on the scala channel and someone suggested that.  I wonder what happens on support-frontend, as that has a lot of shared modules as well.

build.sbt
johnduffell: just to update this, actually SBT seems to magically work out that something is shared, and only runs it once.  (I tried running `sbt test` locally and it only ran `get auth SF correctly` test once.)

johnduffell: the answer is here - it's an intentional feature of most build tools
 https://www.scala-sbt.org/1.x/docs/Task-Graph.html#Benefits+of+hybrid+flow-based+programming:~:text=First%20is%20de%2Dduplication.%20With%20flow%2Dbased%20programming%2C%20a%20task%20is%20executed%20only%20once%20even%20when%20it%20is%20depended%20by%20multiple%20tasks.%20For%20example%2C%20even%20when%20multiple%20tasks%20along%20the%20task%20graph%20depend%20on%20Compile%20/%20compile%2C%20the%20compilation%20will%20be%20executed%20exactly%20once.

PR #2168, opened 2024-02-22
make salesforce-core scala 3 ready

johnduffell:
This is just a small PR to make salesforce-core module possible to compile against scala 3 (but it still compiles against 2.13 in reality)

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -95,7 +95,7 @@ lazy val zuora = library(project in file("lib/zuora"))
 lazy val `salesforce-core` = library(project in file("lib/salesforce/core"))
   .dependsOn(`config-core`)
   .settings(
-    libraryDependencies ++= Seq(playJson) ++ logging,
+    libraryDependencies ++= Seq(playJson),
     dependencyOverrides ++= jacksonDependencies,
   )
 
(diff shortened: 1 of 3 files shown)


build.sbt
```diff
@@ -95,7 +95,7 @@ lazy val zuora = library(project in file("lib/zuora"))
 lazy val `salesforce-core` = library(project in file("lib/salesforce/core"))
   .dependsOn(`config-core`)
   .settings(
-    libraryDependencies ++= Seq(playJson) ++ logging,
```
johnduffell, line 98: unused

lib/salesforce/core/src/main/scala/com/gu/salesforce/SFAuthConfig.scala
```diff
@@ -14,7 +14,7 @@ case class SFAuthConfig(
 
 object SFAuthConfig {
   implicit val reads: Reads[SFAuthConfig] = Json.reads[SFAuthConfig]
-  implicit val location = ConfigLocation[SFAuthConfig](path = "sfAuth", version = 1)
+  implicit val location: ConfigLocation[SFAuthConfig] = ConfigLocation[SFAuthConfig](path = "sfAuth", version = 1)
```
johnduffell, line 17: implicits must have explicit type in scala 3

tomrf1 (approved):

PR #2170, opened 2024-02-27
Change noisy alarm in discount-api

rupertbates:
## What does this change?
Changes the alarm in the discount api to trigger on 500 errors not 400s as it did previously. 

400s are expected from the discount api as it responds with them when a user is not eligible for a discount due to one of a number of reasons so we do not want to alarm on it.

### DIFF
--- a/cdk/lib/discount-api.ts
+++ b/cdk/lib/discount-api.ts
@@ -125,9 +125,9 @@ export class DiscountApi extends GuStack {
 		const alarmDescription = (description: string) =>
 			`Impact - ${description}. Follow the process in https://docs.google.com/document/d/1_3El3cly9d7u_jPgTcRjLxmdG2e919zCLvmcFCLOYAk/edit`;
 
-		new GuAlarm(this, 'ApiGateway4XXAlarmCDK', {
+		new GuAlarm(this, 'ApiGateway5XXAlarmCDK', {
 			app,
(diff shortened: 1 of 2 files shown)


tomrf1 (approved):

cdk/lib/discount-api.ts
```diff
@@ -125,9 +125,9 @@ export class DiscountApi extends GuStack {
 		const alarmDescription = (description: string) =>
 			`Impact - ${description}. Follow the process in https://docs.google.com/document/d/1_3El3cly9d7u_jPgTcRjLxmdG2e919zCLvmcFCLOYAk/edit`;
 
-		new GuAlarm(this, 'ApiGateway4XXAlarmCDK', {
+		new GuAlarm(this, 'ApiGateway5XXAlarmCDK', {
 			app,
-			alarmName: alarmName('API gateway 4XX response'),
+			alarmName: alarmName('API gateway 5XX response'),
 			alarmDescription: alarmDescription(
 				'Discount api received an invalid request',
```
johnduffell, line 132:
```suggestion
				'Discount api had something go wrong internally during a request',
```

johnduffell: actually this is impact
```suggestion
				'Discount api could not display/apply a discount to an existing subscription via Manage',
```

johnduffell (approved):
with a small tweak

PR #2182, opened 2024-03-01
Enable discount at point of acquisition in new-product-api

LAKSHMIRPILLAI:
## What does this change?

As part of the new 3 tier product proposition, we need to be able to add discounts to Supporter Plus subscriptions at the point of acquisition in Salesforce.

This PR updates the AddSubscriptionRequest object with a discountRatePlanId  ( an Optional ProductRatePlanId) and is passed to Zuora along with the other ratePlans  when ever  a new subscription is added to an existing Zuora account.This change is done for products  like - SupporterPlus, Guardian Weekly, Digital Subs, Newspaper

### DIFF
--- a/handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/AddDigipackSub.scala
+++ b/handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/AddDigipackSub.scala
@@ -49,15 +49,32 @@ class AddDigipackSub(
     zuoraRatePlanId <- getZuoraRateplanId(request.planId)
       .toApiGatewayContinueProcessing(internalServerError(s"no Zuora id for ${request.planId}!"))
       .toAsync
+    ratePlans = request.discountRatePlanId
+      .map(id =>
+        List(
+          ZuoraCreateSubRequestRatePlan(
+            productRatePlanId = id,
+            maybeChargeOverride = None,
+          ),
+          ZuoraCreateSubRequestRatePlan(
+            maybeChargeOverride = None,
+            productRatePlanId = zuoraRatePlanId,
+          ),
+        ),
+      )
+      .getOrElse(
+        List(
+          ZuoraCreateSubRequestRatePlan(
+            maybeChargeOverride = None,
+            productRatePlanId = zuoraRatePlanId,
+          ),
+        ),
+      )
+
     createSubRequest = ZuoraCreateSubRequest(
       request = request,
       acceptanceDate = request.startDate,
-      ratePlans = List(
-        ZuoraCreateSubRequestRatePlan(
-          maybeChargeOverride = None,
-          productRatePlanId = zuoraRatePlanId,
-        ),
-      ),
+      ratePlans = ratePlans,
     )
     subscriptionName <- createSubscription(createSubRequest).toAsyncApiGatewayOp("create digiPack subscription")
     plan = getPlan(request.planId)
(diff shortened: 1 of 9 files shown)


handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/AddGuardianWeeklySub.scala
```diff
@@ -143,25 +143,53 @@ object AddGuardianWeeklySub {
...
-          acceptanceDate = request.startDate.plusWeeks(6),
-          ratePlans = List(
-            ZuoraCreateSubRequestRatePlan(
-              productRatePlanId = zuora6for6RatePlanAndCharge.productRatePlanId,
-              maybeChargeOverride = Some(
-                ChargeOverride(
-                  amountMinorUnits = None,
-                  productRatePlanChargeId = zuora6for6RatePlanAndCharge.productRatePlanChargeId,
-                  triggerDate = Some(request.startDate),
+
```
graham228221, line 146: I think you're only implementing the discountRatePlanId changes for the case `if (request.planId == sixForSixPlanId)`, but it should be implemented for the `else` as well.

graham228221: You could also actually move your chances so that they are ONLY in the `else` block because 6for6 already has a discount and wouldn't/shouldn't have another acquisition discount applied on top. 

Even better, we could remove 6for6 completely as it is not offered to customers any more. But that's probably outside the scope of your PR!

LAKSHMIRPILLAI: yes, thank you for spotting this.I have updated it for the else block now, keeping the changes for sixForSixPlanId .But I will remove the 6for6 code completely in a follow up  PR

JoeMitchellGuardian (approved):
I'm happy to approve this! If you want to wait for Rupert or a server side developer to take a look before merging then that's fine too!

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/AddDigipackSub.scala
```diff
@@ -49,15 +49,32 @@ class AddDigipackSub(
...
+        ),
+      )
+      .getOrElse(
+        List(
+          ZuoraCreateSubRequestRatePlan(
+            maybeChargeOverride = None,
+            productRatePlanId = zuoraRatePlanId,
+          ),
+        ),
+      )
```
rupertbates, line 72:
```suggestion
        ratePlans = List(
      ZuoraCreateSubRequestRatePlan(
        maybeChargeOverride = None,
        productRatePlanId = zuoraRatePlanId,
      ),
    ) ++ request.discountRatePlanId
      .map(id =>
        List(
          ZuoraCreateSubRequestRatePlan(
            productRatePlanId = id,
            maybeChargeOverride = None,
          ),
        ),
      )
      .getOrElse(Nil)
```

I think this is actually better because we don't have to define the second rate plan twice

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/AddGuardianWeeklySub.scala
rupertbates: I would remove it from the the 6 for 6 part because as Graham says we would not apply a discount to 6 for 6 even if we did still sell it.

PR #2163, opened 2024-02-19
feat: implement state to update Zuora accounts concurrently

AndreaDiotallevi:
## What does this change?

[Trello](https://trello.com/c/IrT4D7KK/596-implement-state-update-zuora-accounts-in-batches)

- Add `actionUpdate` function to Zuora support-service-lambda shared module for updating 50 accounts at once
- Use [Map state in Distributed mode](https://docs.aws.amazon.com/step-functions/latest/dg/use-dist-map-orchestrate-large-scale-parallel-workloads.html) to loop through the CSV saved in S3 and update the Zuora accounts

Screenshot 2024-03-02 at 12 08 09

Screenshot 2024-03-02 at 11 40 18

- The distributed map state also saves the results to [S3](https://s3.console.aws.amazon.com/s3/buckets/salesforce-disaster-recovery-code?region=eu-west-1&bucketType=general&prefix=2024-03-02T12%3A07%3A34.654Z/a4f28c2d-959f-4e90-bc55-250e603c356a/&showversions=false). See [Result Writer](https://docs.aws.amazon.com/step-functions/latest/dg/input-output-resultwriter.html) for details.

Screenshot 2024-03-02 at 12 20 50

- We can adjust concurrency from 1 to 10,000

Screenshot 2024-03-02 at 11 40 35

Screenshot 2024-03-02 at 11 42 49

Screenshot 2024-03-02 at 11 41 14

### Why add concurrency?

I have tested the workflow on CODE and it took ~8 minutes for ~7000 records (3-4 minutes per API call, which updates 50 record at once in a batch update).

Which means for PROD with 2 millions account it could take ~40 hours (rough approximation that doesn't take into account potentially faster api in PROD).

On CODE, with concurrency 40, it takes 30 seconds.

## Why are you doing this?

This is the 6th PR for the KR1 of the [Q4 FY 23/24 OKR - Disaster Recovery for Salesforce](https://docs.google.com/document/d/1UFDM33Yhl0cgHcIDIfWIkq8V7ZiYNc32wy0SGdfmILo/edit#heading=h.d6f3idk6a29j).

For context, this application is the implementation of the "Salesforce Disaster Recovery" system below.

landscape

## Design document

This PR implements point 7 of the solution outlined in this [document](https://docs.google.com/document/d/1_KxFtfKU3-3-PSzaAYG90uONa05AVgoBmyBDyu5SC5c/edit#heading=h.jyzfkcwrdmx7).

## How to test

[...]

PR #2159, opened 2024-02-09
feat: implement state to get Salesforce query job status

AndreaDiotallevi:
## What does this change?

[Trello](https://trello.com/c/VmJywoHI/594-implement-state-get-salesforce-query-job-status)

Implement conditional logic based on the status of the Salesforce query.

Screenshot 2024-02-13 at 15 20 31

### Details

- Use HTTP task state to call Salesforce API from state machine
- Use Choice state to branch logic based on `$.ResponseBody.state`

```typescript
.next(
	new Choice(this, 'IsSalesforceQueryJobCompleted')
		.when(
			Condition.stringEquals('$.ResponseBody.state', 'JobComplete'),
			saveSalesforceQueryResultToS3,
		)
		.otherwise(waitForSalesforceQueryJobToComplete),
),
```

- Create placeholder Pass state for next state `SaveSalesforceQueryResultToS3`

## Why are you doing this?

This is the 4th for the KR1 of the [Q4 FY 23/24 OKR - Disaster Recovery for Salesforce](https://docs.google.com/document/d/1UFDM33Yhl0cgHcIDIfWIkq8V7ZiYNc32wy0SGdfmILo/edit#heading=h.d6f3idk6a29j).

For context, this application is the implementation of the "Salesforce Disaster Recovery" system below.

landscape

## Design document

This PR implements point 5 of the solution outlined in this [document](https://docs.google.com/document/d/1_KxFtfKU3-3-PSzaAYG90uONa05AVgoBmyBDyu5SC5c/edit#heading=h.jyzfkcwrdmx7).

## How to test

You can test the CODE state machine from the AWS console [here](https://eu-west-1.console.aws.amazon.com/states/home?region=eu-west-1#/statemachines/view/arn%3Aaws%3Astates%3Aeu-west-1%3A865473395570%3AstateMachine%3Asalesforce-disaster-recovery-CODE).

Note - Learn [how to test step functions](https://docs.aws.amazon.com/step-functions/latest/dg/sfn-local.html).

## Previous

- [https://github.com/guardian/support-service-lambdas/pull/2153](https://github.com/guardian/support-service-lambdas/pull/2153)
- [https://github.com/guardian/support-service-lambdas/pull/2157](https://github.com/guardian/support-service-lambdas/pull/2157)
- [https://github.com/guardian/support-service-lambdas/pull/2158](https://github.com/guardian/support-service-lambdas/pull/2158)

## Next

Implement invoke lambda state `SaveSalesforceQueryResultToS3`.

PR #2158, opened 2024-02-09
feat: implement state to wait for Salesforce query job to complete

AndreaDiotallevi:
## What does this change?

[Trello](https://trello.com/c/E2vbXsXl/587-implement-state-wait-for-salesforce-query-job-to-complete)

Implement wait state `WaitForSalesforceQueryJobToComplete` of the Salesforce disaster recovery state machine.

Screenshot 2024-02-09 at 14 23 12

### Details

- Wait 1 second for CODE
- Wait 30 seconds for PROD

## Why are you doing this?

This is the 3rd PR for the KR1 of the [Q4 FY 23/24 OKR - Disaster Recovery for Salesforce](https://docs.google.com/document/d/1UFDM33Yhl0cgHcIDIfWIkq8V7ZiYNc32wy0SGdfmILo/edit#heading=h.d6f3idk6a29j).

For context, this application is the implementation of the "Salesforce Disaster Recovery" system below.

landscape

## Design document

This PR implements point 4 of the solution outlined in this [document](https://docs.google.com/document/d/1_KxFtfKU3-3-PSzaAYG90uONa05AVgoBmyBDyu5SC5c/edit#heading=h.jyzfkcwrdmx7).

## How to test

You can test the CODE state machine from the AWS console [here](https://eu-west-1.console.aws.amazon.com/states/home?region=eu-west-1#/statemachines/view/arn%3Aaws%3Astates%3Aeu-west-1%3A865473395570%3AstateMachine%3Asalesforce-disaster-recovery-CODE).

Note - Learn [how to test step functions](https://docs.aws.amazon.com/step-functions/latest/dg/sfn-local.html).

## Previous

- [https://github.com/guardian/support-service-lambdas/pull/2153](https://github.com/guardian/support-service-lambdas/pull/2153)
- [https://github.com/guardian/support-service-lambdas/pull/2157](https://github.com/guardian/support-service-lambdas/pull/2157)

## Next

Implement state `GetSalesforceQueryJobStatus`.

PR #2185, opened 2024-03-05
feat: save failed rows to S3

AndreaDiotallevi:
## What does this change?

[Trello](https://trello.com/c/0MRNBgDu/637-handle-errors-and-failures)

Save failed rows to S3 file called `failed-rows.csv`.

Screenshot 2024-03-07 at 10 58 13

In more details you can see in the screenshot below:

- folder that contains results from the distributed map state

- `failed-rows.csv` is the summary CSV containing the failed rows

- `query-result.csv` is the initial CSV from Salesforce

Screenshot 2024-03-07 at 10 33 20

- the step `GetMapResults` uses the S3 SDK integration to get the `manifest.json` file which contains the result files of the distributed map

- the step `SaveFailedRowsToS3` filters all the result arrays and keeps only the rows which success `false`

Screenshot 2024-03-07 at 10 31 17

## Why are you doing this?

This is the 7th PR for the KR1 of the [Q4 FY 23/24 OKR - Disaster Recovery for Salesforce](https://docs.google.com/document/d/1UFDM33Yhl0cgHcIDIfWIkq8V7ZiYNc32wy0SGdfmILo/edit#heading=h.d6f3idk6a29j).

landscape

## Design document

This PR implements point 8 of the solution outlined in this [document](https://docs.google.com/document/d/1_KxFtfKU3-3-PSzaAYG90uONa05AVgoBmyBDyu5SC5c/edit#heading=h.jyzfkcwrdmx7).

## How to test

You can test the CODE state machine from the AWS console [here](https://eu-west-1.console.aws.amazon.com/states/home?region=eu-west-1#/statemachines/view/arn%3Aaws%3Astates%3Aeu-west-1%3A865473395570%3AstateMachine%3Asalesforce-disaster-recovery-CODE).

You can run unit tests with `pnpm test` inside `handlers/salesforce-disaster-recovery`.

## Previous

- [https://github.com/guardian/support-service-lambdas/pull/2153](https://github.com/guardian/support-service-lambdas/pull/2153)
- [https://github.com/guardian/support-service-lambdas/pull/2157](https://github.com/guardian/support-service-lambdas/pull/2157)
- [https://github.com/guardian/support-service-lambdas/pull/2158](https://github.com/guardian/support-service-lambdas/pull/2158)
- [https://github.com/guardian/support-service-lambdas/pull/2159](https://github.com/guardian/support-service-lambdas/pull/2159)
- [https://github.com/guardian/support-service-lambdas/pull/2160](https://github.com/guardian/support-service-lambdas/pull/2160)
- [https://github.com/guardian/support-service-lambdas/pull/2163](https://github.com/guardian/support-service-lambdas/pull/2163)

## Next

Create stack for CSBX environment.


handlers/salesforce-disaster-recovery/src/handlers/saveFailedRowsToS3.ts
```diff
@@ -0,0 +1,49 @@
...
+	resultFiles: Array<{ Key: string; Size: number }>;
+	filePath: string;
+}): Promise<void> => {
+	const { resultFiles, filePath } = event;
+
+	const bucketName = process.env.S3_BUCKET;
+
+	if (!bucketName) {
+		throw new Error('Environment variables not set');
+	}
```
rupertbates, line 18:
```suggestion
const bucketName = checkDefined(
		process.env.S3_BUCKET,
		'Environment variables not set',
	);
```
Up to you, but I find this a bit more readable

handlers/salesforce-disaster-recovery/src/handlers/updateZuoraAccounts.ts
```diff
@@ -3,29 +3,27 @@ import { ZuoraClient } from '@modules/zuora/zuoraClient';
 import { type AccountRow, batchUpdateZuoraAccounts } from '../services';
 
 export const handler = async (event: { Items: AccountRow[] }) => {
+	const { Items } = event;
+
 	const stage = process.env.STAGE;
 
 	if (!stage) {
 		throw new Error('Environment variables not set');
 	}
```
rupertbates, line 12:
```suggestion
const stage = checkDefined(
		process.env.STAGE,
		'Environment variables not set',
	);
```
You could use the same pattern here.

rupertbates (commented):
So if there are failures, it doesn't fail the step function? Will we alert on them, or is that not needed during a disaster recovery?

handlers/salesforce-disaster-recovery/src/handlers/updateZuoraAccounts.ts
AndreaDiotallevi: Ah nice utility function. I have updated the code.

AndreaDiotallevi:
> So if there are failures, it doesn't fail the step function?

Yes.
The reason is that since we update Zuora accounts in batches of 50, we might have 49 successes and 1 failure.
So, the step function success is more related to the fact that the batch has been successfully "processed".

> Will we alert on them, or is that not needed during a disaster recovery?

That is a good question.
We might want to alert or send an summary email - I will check if it is needed.
At the moment, a file is created with the accounts that have failed.

rupertbates:
> > So if there are failures, it doesn't fail the step function?
> 
> Yes. The reason is that since we update Zuora accounts in batches of 50, we might have 49 successes and 1 failure. So, the step function success is more related to the fact that the batch has been successfully "processed".
> 
> > Will we alert on them, or is that not needed during a disaster recovery?
> 
> That is a good question. We might want to alert or send an summary email - I will check if it is needed. At the moment, a file is created with the accounts that have failed.
> 
> Do you think we should fail the step function at the end if we have failures?
> 
> Or do you think we should fail the batches that have at least a failure? It actually might be clearer now that you mentioned that. (We could add the `ToleratedFailurePercentage` to 100.) However, the problem is that if we fail the entire batch, we are not able to see on the result S3, the individual results of the accounts. I might be able to place the accounts that failed in the error potentially - I'll give it a test!

I think it might be good to fail the step function after the SaveFailedRowToS3 stage so that it's really clear to whoever triggers it that there are failed rows, otherwise there is a risk that they see that the step function succeeded and never check S3.

AndreaDiotallevi:
> I think it might be good to fail the step function after the SaveFailedRowToS3 stage so that it's really clear to whoever triggers it that there are failed rows, otherwise there is a risk that they see that the step function succeeded and never check S3.

@rupertbates 

Sounds like a good idea.

Shall I implement that as a separate PR?

rupertbates:
> > I think it might be good to fail the step function after the SaveFailedRowToS3 stage so that it's really clear to whoever triggers it that there are failed rows, otherwise there is a risk that they see that the step function succeeded and never check S3.
> 
> @rupertbates
> 
> Sounds like a good idea.
> 
> Shall I implement that as a separate PR?

Yes, makes sense

rupertbates (approved):
👍 good work

PR #2187, opened 2024-03-06
Remove the top level `products` object from the product-catalog

rupertbates:
## What does this change?

This PR removes the top level `products` array from the product catalog as it wasn't adding any benefits and was just making code which uses the catalog more verbose.


jamesgorrie (approved):
Fab

PR #2191, opened 2024-03-08
Use pnpm as the package manager for cdk rather than yarn

rupertbates:
## What does this change?
Ever since #2110 we have used pnpm as a package manager within support-service-lambdas. However the cdk build was still using yarn. 

This PR changes the cdk project so it is now just another project in the pnpm workspace, meaning that we can now do things like `pnpm install` or `pnpm test` to install dependencies or run tests for all projects including the cdk project, rather than having to separately cd to the cdk directory and run `yarn install` & `yarn test`.

It also means that the cdk project shares typescript version and a number of dependency versions with all the other projects, rather than needing to be kept up to date separately.


.github/workflows/ci-typescript.yml
```diff
@@ -35,9 +35,6 @@ jobs:
           role-to-assume: ${{ secrets.GU_RIFF_RAFF_ROLE_ARN }}
           aws-region: eu-west-1
 
-      - name: Generate CFN templates from GuCDK
```
rupertbates, line 38: Don't think this was needed anyway

cdk/lib/generate-product-catalog.ts
```diff
@@ -43,8 +43,8 @@ export class GenerateProductCatalog extends GuStack {
 			app: app,
 		});
 
-		const zuoraCatalogBucketName: string = 'gu-zuora-catalog';
-		const productCatalogBucketName: string = 'gu-product-catalog';
+		const zuoraCatalogBucketName = 'gu-zuora-catalog';
+		const productCatalogBucketName = 'gu-product-catalog';
```
rupertbates, line 47: Linting fixes

cdk/package.json
```diff
@@ -9,26 +9,19 @@
     "format": "prettier --write \"{lib,bin}/**/*.ts\"",
     "lint": "eslint lib/** bin/** --ext .ts --no-error-on-unmatched-pattern",
     "synth": "cdk synth --path-metadata false --version-reporting false",
-    "diff": "cdk diff --path-metadata false --version-reporting false"
+    "diff": "cdk diff --path-metadata false --version-reporting false",
+    "package": "pnpm build && pnpm lint && pnpm test && pnpm synth"
   },
   "devDependencies": {
     "@guardian/cdk": "52.3.0",
-    "@guardian/eslint-config-typescript": "5.0.0",
```
rupertbates, line 16: All the dependencies and config which have been removed from this file now lives in the workspace

modules/test-users/src/create.ts
```diff
@@ -122,6 +121,10 @@ export const createAccountAndSubscription = async (
 
 void (async () => {
 	const zuoraClient = await ZuoraClient.create(stage);
-	const response = await createAccountAndSubscription(zuoraClient);
+	const productCatalog = await getProductCatalogFromApi(stage);
+	const response = await createAccountAndSubscription(
+		zuoraClient,
+		productCatalog,
+	);
```
jamesgorrie, line 128: ☝️ These changes look like they might be related to some other work rather than linting?

jamesgorrie (approved):
Tidies it up really nicely! Just a question on whether some work creeped into `modules/test-users/src/create.ts`?

modules/test-users/src/create.ts
rupertbates: This code hadn't been updated when the product catalog work was merged so it was out of date. I think that didn't get picked up because this project is not build in CI. We should probably change that but it can be in another PR

PR #2192, opened 2024-03-12
feat: fail batches that have errors

AndreaDiotallevi:
DRAFT


cdk/lib/salesforce-disaster-recovery.ts
```diff
@@ -191,6 +193,7 @@ export class SalesforceDisasterRecovery extends GuStack {
 				stateJson: {
 					Type: 'Map',
 					MaxConcurrency: 10,
+					ToleratedFailurePercentage: 100,
```
rupertbates, line 196: Maybe put a comment here to explain why we are ignoring errors?

AndreaDiotallevi: Good point - I have added a comment

rupertbates (approved):
👍 Looks good

PR #2194, opened 2024-03-13
use arm64 runtime for all cloudformed lambdas

johnduffell:
ARM 64 is cheaper than x86, and similar speed.  It's well used and tested amongst our main servers for several years.  The product-move-api is the only one that uses it from the lambdas.
background info 
- https://docs.aws.amazon.com/lambda/latest/dg/foundation-arch.html
- https://filia-aleks.medium.com/aws-lambda-battle-x86-vs-arm-graviton2-perfromance-3581aaef75d9

This PR switches all our lambdas over
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-lambda-function.html#cfn-lambda-function-architectures

I will test a couple in CODE but I don't have any reservations.  I think any issues might show would be possible cases with memory usage, similar to what we had when switching java8 -> java8.al2

I also happened to notice that java17 and java21 is available, I will look into any benefits for scala code.

### DIFF
--- a/handlers/cancellation-sf-cases-api/cfn.yaml
+++ b/handlers/cancellation-sf-cases-api/cfn.yaml
@@ -85,6 +85,8 @@ Resources:
       MemorySize: 1536
       Runtime: java11
       Timeout: 300
+      Architectures:
+        - arm64
     DependsOn:
       - CancellationSFCasesApiRole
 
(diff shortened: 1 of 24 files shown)


tomrf1 (approved):

johnduffell:
tested in CODE with auto cancel lambda via API gatway - it seems to be slightly slower cold start based on one attempt.  Then it reduces to 100-130ms approx.

PR #2195, opened 2024-03-13
compile and run against java 21 LTS

johnduffell:
Further to updating to arm64 architecture https://github.com/guardian/support-service-lambdas/pull/2194

This PR updates to java 21 LTS.  This seems to have bumped performance significantly with support-workers https://github.com/guardian/support-frontend/pull/5792 , so I am replicating it here.

I have also bumped the version we compile with in GHA, in case that improves compilation performance.

I tested auto cancel in CODE and it seems to have improved the cold start noticably

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -9,7 +9,6 @@ val scala2Settings = Seq(
     "-encoding",
     "UTF-8",
     "-feature",
-    "-target:jvm-1.8",
     "-language:existentials",
     "-language:higherKinds",
     "-language:implicitConversions",
@@ -860,8 +859,8 @@ def genDocsImpl(makeDocsClassName: String): Def.Initialize[Task[Unit]] = {
 initialize := {
   val _ = initialize.value
   assert(
-    List("1.8", "11").contains(sys.props("java.specification.version")),
-    "Java 8 or 11 is required for this project.",
+    sys.props("java.specification.version") == "21",
+    "Java 21 is required for this project, download corretto-21 from File->Project structure.",
   )
 }
 
--- a/.github/workflows/ci-scala.yml
+++ b/.github/workflows/ci-scala.yml
@@ -105,7 +105,7 @@ jobs:
         uses: actions/setup-java@v4
         with:
           distribution: 'corretto'
-          java-version: '11'
+          java-version: '21'
           cache: 'sbt'
 
       - name: Build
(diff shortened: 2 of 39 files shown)


build.sbt
```diff
@@ -9,7 +9,6 @@ val scala2Settings = Seq(
     "-encoding",
     "UTF-8",
     "-feature",
-    "-target:jvm-1.8",
```
johnduffell, line 12: this parameter was causing a lot of warnings, also it wasn't really relevant any more.  It would only be needed if we were compiling on a later VM and running on an earlier one.

build.sbt
```diff
@@ -860,8 +859,8 @@ def genDocsImpl(makeDocsClassName: String): Def.Initialize[Task[Unit]] = {
 initialize := {
   val _ = initialize.value
   assert(
-    List("1.8", "11").contains(sys.props("java.specification.version")),
-    "Java 8 or 11 is required for this project.",
+    sys.props("java.specification.version") == "21",
+    "Java 21 is required for this project, download corretto-21 from File->Project structure.",
```
johnduffell, line 863: I've seen problems locally where people have been on a later version than is supported by either SBT/play/scala, so I have kept this in and updated it.
If it's causing issues locally we can rethink.

tomrf1 (approved):

rtyley:
Looks like we need to bump the version of Java used by our Scala Steward workflow as well:

https://github.com/guardian/scala-steward-public-repos/actions/runs/8278707245/job/22651625930

> java.lang.AssertionError: assertion failed: **Java 21** is required for this project, download corretto-21 from File->Project structure.
>   	at scala.Predef$.assert(Predef.scala:223)
>   	at $cd99ddf78d62fbdae579$.$anonfun$$sbtdef$1(/home/runner/scala-steward/workspace/repos/guardian/support-service-lambdas/build.sbt:863)
>   	at $cd99ddf78d62fbdae579$.$anonfun$$sbtdef$1$adapted(/home/runner/scala-steward/workspace/repos/guardian/support-service-lambdas/build.sbt:859)
>   	at scala.Function1.$anonfun$compose$1(Function1.scala:49)

johnduffell:
good spot @rtyley I have raised a PR to update scala steward, however it is likely to fail on other projects where they are not up to recent versions of sbt, it's needing 2.12.18 or later to work https://docs.scala-lang.org/overviews/jdk-compatibility/overview.html

PR #2196, opened 2024-03-19
Don't create a failed lambda alarm in code for generate-product-catalog

rupertbates:
## What does this change?
We don't want alarms when the CODE version of the generate-product-catalog lambda fails


johnduffell (approved):
nice and simple 👍

jamesgorrie (approved):
Like it - much more readable than the CF conditional way of doing this.

PR #2199, opened 2024-03-22
build: add pre-push hook

AndreaDiotallevi:
## What does this change?

This adds a pre-push hook that runs the cdk lint & tests before push. 

This can hopefully shorten the feedback loop between development and deployment when we forget to update the snapshot tests or when there are linting errors.

### DIFF
--- a/.husky/pre-push
+++ b/.husky/pre-push
@@ -1,0 +1,1 @@
+cd cdk && yarn test
--- a/.husky/pre-push
+++ b/.husky/pre-push
@@ -1,2 +1,1 @@
 pnpm --filter cdk lint && pnpm --filter cdk test
-pnpm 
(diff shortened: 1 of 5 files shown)


.husky/pre-push
```diff
@@ -0,0 +1 @@
+cd cdk && yarn lint && yarn test
```
rupertbates, line 1:
```suggestion
pnpm --filter cdk lint && pnpm --filter cdk test
pnpm 
```

We are not using yarn to build CDK since #2191

rupertbates (approved):
Great idea 👏 👏 👏

.husky/pre-push
AndreaDiotallevi: Good point!

.husky/pre-push
```diff
@@ -0,0 +1 @@
+pnpm --filter cdk lint && pnpm --filter cdk test
```
jamesgorrie, line 1: With linting - most IDEs support fix on save, might be worth looking at that, it shortens the feedback loop even more.

PR #2201, opened 2024-03-26
Product switching

rupertbates:
## What does this change?
This PR rewrites the recurring contribution to supporter plus product-move lambda to reduce errors and improve maintainability. 

## Changes
- The switch process now removes any future dated amount update amendments before trying to switch product - the failure to do this in the previous version has been causing errors since launch.
- It uses Zuora's newer Orders api rather than the now deprecated subscribe and amend api. This is now Zuora's recommended approach for these sorts of operations and [will be the only supported method from Feb 2026](https://knowledgecenter.zuora.com/Releases/ZProduct_and_feature_end_of_support_plan).
- It makes all the necessary changes to the sub in a single call to increase performance, reduce the possibility of partial failures and reduce the number of intermediate versions on the subscription.
- Moves to Typescript from Scala ZIO which is a better fit for us strategically moving forward.

## Next Steps
This PR will leave the existing api in place and create a new one, to switch over to this will require a change to manage-frontend.


tomrf1:
can we have a readme?

cdk/lib/product-switch-api.ts
```diff
@@ -0,0 +1,188 @@
...
+			functionName: nameWithStage,
+			fileName: `${app}.zip`,
+			handler: 'index.handler',
+			runtime: Runtime.NODEJS_18_X,
+			memorySize: 1024,
+			timeout: Duration.seconds(300),
+			environment: commonEnvironmentVariables,
+			// Create an alarm
+			monitoringConfiguration: {
+				http5xxAlarm: { tolerated5xxPercentage: 5 },
```
tomrf1, line 51: this seems high, are failures expected often?

handlers/product-switch-api/src/contributionToSupporterPlus.ts
```diff
@@ -0,0 +1,280 @@
...
+				],
+			},
+		],
+	};
+};
+
+export const buildSwitchRequestBody = (
+	orderDate: Dayjs,
+	productSwitchInformation: SwitchInformation,
+) => {
```
tomrf1, line 211: is it worth adding return types to some of these functions?

handlers/product-switch-api/src/productSwitchEmail.ts
```diff
@@ -0,0 +1,67 @@
...
+		dayjs(),
+		emailAddress,
+		firstName,
+		lastName,
+		currency,
+		switchInformation.input.price,
+		firstPaymentAmount,
+		billingPeriod,
+		subscriptionNumber,
+		identityId,
```
tomrf1, line 63: with this many params is it better for `buildEmailMessage` to receive an object?

modules/email/src/email.ts
```diff
@@ -4,6 +4,28 @@ import { awsConfig } from '@modules/aws/config';
...
+	{
+		[K in Keys]-?: Required<Pick<T, K>> &
+			Partial<Record<Exclude<Keys, K>, undefined>>;
+	}[Keys];
+
+// The EmailMessageWithUserId type ensures that the message has either a SfContactId or an IdentityUserId
+export type EmailMessageWithUserId = Either<
+	EmailMessage,
+	'SfContactId' | 'IdentityUserId'
+>;
```
tomrf1, line 20: This Either definition is a little mind bending, is it equivalent to the following?
```
type EmailMessageWithUserId = EmailMessage & ({ SfContactId: string } | { IdentityUserId: string });
```

jamesgorrie: Another way to avoid the mind bend

```ts
type EmailMessageToIdentityUser = {
	To: EmailPayload;
	DataExtensionName: DataExtensionName;
	IdentityUserId: string;
}
type EmailMessageToSfUser = {
	To: EmailPayload;
	DataExtensionName: DataExtensionName;
	SfContactId: string;
}

type EmailMessage = EmailMessageToIdentityUser | EmailMessageToSfUser
```

jamesgorrie (approved):
The code seems good - so happy to approve from that POV. I'll need to go through it again to see if I understand the Zuora flow as it seems like there are quite a few steps.

cdk/lib/product-switch-api.ts
rupertbates: I'm going to rework the alarming actually

modules/email/src/email.ts
rupertbates: Ok I've gone for 
```typescript
type EmailMessage = {
	To: EmailPayload;
	DataExtensionName: DataExtensionName;
};

export type EmailMessageWithSfContactId = EmailMessage & {
	SfContactId: string;
};

export type EmailMessageWithIdentityUserId = EmailMessage & {
	IdentityUserId: string;
};

export type EmailMessageWithUserId =
	| EmailMessageWithSfContactId
	| EmailMessageWithIdentityUserId;
```

handlers/product-switch-api/src/productSwitchEmail.ts
rupertbates: I don't think it's worth it because they are all being destructured out of an object in the calling function, putting them back into another object feels like overkill

cdk/bin/cdk.ts
```diff
@@ -147,23 +148,45 @@ new GenerateProductCatalog(app, 'generate-product-catalog-PROD', {
...
+	'stripe-webhook-endpoints-PROD',
+	stripeWebhookEndpointsProdProps,
+);
 
-new StripeWebhookEndpoints(app, "stripe-webhook-endpoints-CODE",stripeWebhookEndpointsCodeProps);
-new StripeWebhookEndpoints(app, "stripe-webhook-endpoints-PROD",stripeWebhookEndpointsProdProps);
+new ProductSwitchApi(app, 'product-switch-api-CODE', {
+	stack: 'support',
+	stage: 'CODE',
+	domainName: `product-switch-api-code.${supportApisDomain}`,
```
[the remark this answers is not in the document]
rupertbates: It's the way that we used to do it before the GuCname construct worked with NS1 and I've not changed it. It would be good to (and particularly to change the code generators) but I'm going to leave it for a different PR.

cdk/lib/product-switch-api.ts
```diff
@@ -0,0 +1,188 @@
...
+		// ---- API-triggered lambda functions ---- //
+		const lambda = new GuApiLambda(this, `${app}-lambda`, {
+			description:
+				'An API Gateway triggered lambda generated in the support-service-lambdas repo',
+			functionName: nameWithStage,
+			fileName: `${app}.zip`,
+			handler: 'index.handler',
+			runtime: Runtime.NODEJS_18_X,
+			memorySize: 1024,
+			timeout: Duration.seconds(300),
```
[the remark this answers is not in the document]
rupertbates: It's being cautious because Zuora's response times can be unpredictable and the lambda is carrying out quite a few requests. 

This is the same number as the old version of the API used - hopefully this version should be quite a bit quicker because it uses an api which allows us to make fewer requests but better to be safe. 

Really I don't think it matters how big this number is because the lambda will run for as long as it takes, it only becomes an issue if this number is too low because then we'll be failing to service users.

modules/email/src/email.ts
jamesgorrie: Nice and easy to understand.

cdk/bin/cdk.ts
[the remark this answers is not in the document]
jamesgorrie: Yea, makes sense to do it as a universal thing - thanks for explaining.

cdk/lib/product-switch-api.ts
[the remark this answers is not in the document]
jamesgorrie: Makes sense - I was wondering if every lambda run hit close to this would it be a cost issue. But sounds like not, and this could sometimes be the case.

handlers/product-switch-api/src/contributionToSupporterPlus.ts
```diff
@@ -0,0 +1,287 @@
...
+		productSwitchInformation.catalog.supporterPlus.subscriptionChargeId,
+		productSwitchInformation.account.id,
+		productSwitchInformation.account.defaultPaymentMethodId,
+	);
+
+	await sendThankYouEmail(paidAmount, productSwitchInformation);
+
+	await sendSalesforceTracking(paidAmount, productSwitchInformation);
+
+	await sendToSupporterProductData(productSwitchInformation);
```
tjmw, line 51: No big deal either way, but do these need to happen in serial? Could this be:

```suggestion
	await Promise.all([
                sendThankYouEmail(paidAmount, productSwitchInformation),
                sendSalesforceTracking(paidAmount, productSwitchInformation),
                sendToSupporterProductData(productSwitchInformation),
         ]);
```

handlers/product-switch-api/src/productSwitchEmail.ts
tjmw: I don't feel super strongly about this, but for me any time there are > 2 or 3 args I tend to prefer passing as an object simply so that I don't have to think about argument ordering. With ordered args it feels easy to make a mistake and get two args with the same type the wrong way around (say `emailAddress` and `firstName` in this example). Whereas with named args (an object) that feels less likely.

handlers/product-switch-api/test/contributionToSupporterPlusIntegration.test.ts
```diff
@@ -0,0 +1,132 @@
...
+import { adjustNonCollectedInvoice } from '../src/payment';
+import { getSwitchInformationWithOwnerCheck } from '../src/switchInformation';
+
+const jestConsole = console;
+beforeEach(() => {
+	global.console = console;
+});
+afterEach(() => {
+	global.console = jestConsole;
+});
```
tjmw, line 22: Out of interest what effect does this have? I'm always interested in ways of making test output less noisy!

handlers/product-switch-api/src/contributionToSupporterPlus.ts
rupertbates: I've added a bunch, it was worth doing because it made me build a schema for the Zuora orders api which will be useful I think.

handlers/product-switch-api/src/productSwitchEmail.ts
rupertbates: Makes sense, I'll go with the majority opinion!

handlers/product-switch-api/test/contributionToSupporterPlusIntegration.test.ts
rupertbates: Without it:
```
  console.log
    Checking subscription A-S00527544 is owned by the currently logged in user

      at getSwitchInformationWithOwnerCheck (src/switchInformation.ts:107:10)

  console.log
    fetching zuora bearer token for stage: CODE

      at BearerTokenProvider.fetchZuoraBearerToken (../../modules/zuora/src/bearerTokenProvider.ts:34:11)
```

With it:
```
Checking subscription A-S00527544 is owned by the currently logged in user
fetching zuora bearer token for stage: CODE
```

handlers/product-switch-api/src/contributionToSupporterPlus.ts
rupertbates: Good point 👍

handlers/product-switch-api/test/contributionToSupporterPlusIntegration.test.ts
tjmw: ✨ Thanks, I'm definitely going to steal this!

rupertbates: It's much better isn't it? The console.log stuff really makes it hard to read

PR #2202, opened 2024-03-26
Add scripts to create and update stacks, and update lambdas

rupertbates:
## What does this change?
This PR adds 3 shell scripts to the root of the repo:
## create-stack.sh 
This script will build the lambda and cdk for the provided project, upload the lambda package to S3, and then call `create-stack` with the AWS Cli. It will fail if the stack already exists, in this case use update-stack.

Using this script when deploying a lambda for the first time will avoid the need to run two deploys (the first a partial one) through RiffRaff.

Usage: `./create-stack.sh [project name]` 
eg. `./create-stack.sh discount-api`

## update-stack.sh 
This script will build the cdk for the provided project and then call `update-stack` with the AWS Cli. It will fail if the stack does not already exist, in this case use create-stack.

Usage: `./update-stack.sh [project name]` 
eg. `./update-stack.sh discount-api`

## update-lambda.sh
This script will build the provided project, copy the zip file to S3 and then call `update-function-code` with the AWS Cli to update the lambda.

Usage: `./create-lambda.sh [project name]` 
eg. `./create-lambda.sh discount-api`


akash1810:
Any reason to not use Riff-Raff for this?

johnduffell (approved):
nice, this will be very handy.
There is already an `sbt deployToCode` that does the lambda part only at least for some lambdas, I'm not sure whether the two could be combined to make a one stop script to cover all cases.

johnduffell:
> It's significantly quicker during development. Also using the create script gets round the problem RiffRaff has where it needs the files to exist in S3 before it can do the first deploy, so you need to do a partial deploy and then a full one.

agree, also constantly having to push a commit, watch the build, then go off to riffraff is much more context switching than just typing in a command where you are already.

create-stack.sh
```diff
@@ -0,0 +1,42 @@
...
+zipFile="./handlers/$PROJECT_NAME/target/$PROJECT_NAME.zip"
+
+echo "Copying lambda package to S3"
+aws s3 cp $zipFile s3://$s3Bucket/$s3Path --profile membership --region eu-west-1
+
+echo "Building CDK"
+pnpm --filter cdk package
+
+echo "Creating stack"
+aws cloudformation create-stack \
```
akash1810, line 34: Should the standard `Stack`, `Stage`, and `App` tags be added to the stack too? By [default](https://riffraff.gutools.co.uk/docs/magenta-lib/types#cloudformation), Riff-Raff uses these tags to identify which CFN stack to deploy to.

create-stack.sh
```diff
@@ -0,0 +1,42 @@
...
+  exit 2
+fi
+
+PROJECT_NAME="$1"
+echo "Creating stack $PROJECT_NAME"
+
+echo "Building lambda package"
+pnpm --filter $PROJECT_NAME package
+
+s3Bucket="membership-dist"
```
akash1810, line 23: Worth using the `/account/services/artifact.bucket` SSM Parameter, removing a bucket name from the repository?

create-stack.sh
akash1810, line 34: Does this need the `--region` flag?

update-stack.sh
```diff
@@ -0,0 +1,29 @@
...
+  echo "please provide the project name as an argument, eg. ./update-stack.sh discount-api"
+  exit 2
+fi
+
+PROJECT_NAME="$1"
+echo "Updating stack $PROJECT_NAME"
+
+pnpm --filter cdk package
+
+aws cloudformation update-stack \
```
akash1810, line 21: Does this need the `--region` flag?

create-stack.sh
rupertbates: Good idea 👍

PR #2205, opened 2024-03-28
Set soft-opt-in DLQ retention to 10 days

tomrf1:
To give us more time to process failed events
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-sqs-queue.html


jamesgorrie (approved):
You might want to leave a note/comment in the code but not sure it's worth it

PR #2207, opened 2024-03-28
revert to java 11 due to JWT issue

johnduffell:
we are seeing "Invalid JWT Signature" errors when this is called when it's trying to connect to google.

This is causing 5xx errors to our client (identity)

Reverting the JVM so that we can investigate/write more tests before we move forward.

### DIFF
--- a/handlers/identity-retention/cfn.yaml
+++ b/handlers/identity-retention/cfn.yaml
@@ -77,7 +77,7 @@ Resources:
                 - "IdentityRetentionRole"
                 - Arn
             MemorySize: 1536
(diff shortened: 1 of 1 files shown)


jamesgorrie (approved):
Assuming there's no integration issues with this (other lambdas being on a separate version)?

johnduffell:
> Assuming there's no integration issues with this (other lambdas being on a separate version)?

yes this is only this one lambda, reverting back to what it used to be on.
scalac seems to output java 8 bytecode by default so there's no issues there.

PR #2174, opened 2024-02-29
 Recreate existing CFN template using CDK constructs for stripe-webhook-endpoints

LAKSHMIRPILLAI:
## What does this change?

This PR is to recreate existing CFN template using CDK constructs, while retaining existing infrastructure for stripe-webhook-endpoints.

For apps where avoiding downtime is critical, a dual-stack migration strategy forGuCDK Migration is recommended . The idea is to duplicate the key parts of the stack using the pattern, test, and, once confident, switch DNS to point to the new stack.
In this phase we create a new version of your app's infrastructure, according to CDK’s best practices. So, we should have a functional copy of the application, which can be used for testing purposes. Riff-Raff will have been configured to deploy to both versions of the infrastructure, so you can operate in this state for as long as you need to.

[Trello card](https://trello.com/c/mXJhc0b9/255-migration-of-stripe-webhook-endpoints-cloudformation-recreate-existing-cfn-template-using-cdk-constructs-while-retaining-existin)

In more detail, this change:

Makes the lambdas using cdk constructs
Adds an API gateway endpoint for calling it
Adds necessary config for it to be called from stripe accounts (GNM Membership)
Changes existing stripe service calls to specify which account they’re calling (so that, for example, we can look up a customer in the right account, depending on which account’s webhook called the lambda)
We’ve tested this by deploying it to CODE, creating the necessary parameter store config for CODE, and triggering events on the stripe accounts in test mode, which show the calls succeeding.

## Post-Deployment Steps

Once this change is deployed, there are some post-deployment steps, which need to happen in order:

- Create a (non-test-mode) webhook (initially disabled) to call the payment-intent-issue event lambda:
  - [ ] On the GNM Membership account
- Create a (non-test-mode) webhook (initially disabled) to call the customer update event lambda:
  - [ ] On the GNM Membership account
- Add the following AWS Parameter Store entries
  - [x] /CODE/support/stripe-webhook-endpoints/endpointSecret
  - [x] /CODE/support/stripe-webhook-endpoints/zuoraBaseUrl
  - [x] /CODE/support/stripe-webhook-endpoints/zuoraClientId
  - [x] /CODE/support/stripe-webhook-endpoints/zuoraSecret
  - [x] /PROD/support/stripe-webhook-endpoints/endpointSecret 
  - [x] /PROD/support/stripe-webhook-endpoints/zuoraBaseUrl
  - [x] /PROD/support/stripe-webhook-endpoints/zuoraClientId 
[...]

PR #2210, opened 2024-04-04
Recreate existing CFN template using CDK constructs for stripe-webhook-endpoints and update riff raff yaml

LAKSHMIRPILLAI:
## What does this change?

This PR is to recreate existing CFN template using CDK constructs, while retaining existing infrastructure for stripe-webhook-endpoints.

> For apps where avoiding downtime is critical, a dual-stack migration strategy forGuCDK Migration is recommended . The idea is to duplicate the key parts of the stack using the pattern, test, and, once confident, switch DNS to point to the new stack.
> In this phase we create a new version of your app's infrastructure, according to CDK’s best practices. So, we should have a functional copy of the application, which can be used for testing purposes. Riff-Raff will have been configured to deploy to both versions of the infrastructure, so you can operate in this state for as long as you need to.

[Trello card](https://trello.com/c/mXJhc0b9/255-migration-of-stripe-webhook-endpoints-cloudformation-recreate-existing-cfn-template-using-cdk-constructs-while-retaining-existin)

In more detail, this change:

- Makes the lambdas using cdk constructs
- Adds an API gateway endpoint for calling it
- Adds necessary config for it to be called from stripe accounts (GNM Membership)
- Changes existing stripe service calls to specify which account they’re calling (so that, for example, we can look up a customer in the right account, depending on which account’s webhook called the lambda)

We’ve tested this by deploying it to CODE, creating the necessary parameter store config for CODE, and triggering events on the stripe accounts in test mode, which show the calls succeeding.

## Post-Deployment Steps

Once this change is deployed, there are some post-deployment steps, which need to happen in order:

- Create a (non-test-mode) webhook (initially disabled) to call the payment-intent-issue event lambda:
  - [ ] On the GNM Membership account
- Create a (non-test-mode) webhook (initially disabled) to call the customer update event lambda:
  - [ ] On the GNM Membership account
- Add the following AWS Parameter Store entries
  - [x] /CODE/membership/stripe-webhook-endpoints/endpointSecret
  - [x] /CODE/membership/stripe-webhook-endpoints/zuoraBaseUrl
  - [x] /CODE/membership/stripe-webhook-endpoints/zuoraClientId
  - [x] /CODE/membership/stripe-webhook-endpoints/zuoraSecret
  - [x] /PROD/membership/stripe-webhook-endpoints/endpointSecret 
  - [x] /PROD/membership/stripe-webhook-endpoints/zuoraBaseUrl
  - [x] /PROD/membership/stripe-webhook-endpoints/zuoraClientId 
  - [x] /PROD/membership/stripe-webhook-endpoints/zuoraSecret 
  - [ ] Enable the new webhooks
- Remove these AWS Parameter Store entries:
  - [ ] /CODE/membership/payment-intent-issues/endpointSecret
  - [ ] /CODE/membership/payment-intent-issues/zuoraBaseUrl
  - [ ] /CODE/membership/payment-intent-issues/zuoraClientId
  - [ ] /CODE/membership/payment-intent-issues/zuoraSecret
  - [ ] /PROD/membership/payment-intent-issues/endpointSecret
  - [ ] /PROD/membership/payment-intent-issues/zuoraBaseUrl
  - [ ] /PROD/membership/payment-intent-issues/zuoraClientId
  - [ ] /PROD/membership/payment-intent-issues/zuoraSecret

This was previously done in the PR[ #2174](https://github.com/guardian/support-service-lambdas/pull/2174) (which also changed the stack name to `support`) before but had to be reverted in #2200  as it caused a failed deployment .The cause was the riff-raff.yaml  still has the old stack name.  This PR does not include the changes to update the stack name.


cdk/lib/stripe-webhook-endpoints.ts
```diff
@@ -1,15 +1,222 @@
...
+import {ComparisonOperator, Metric} from "aws-cdk-lib/aws-cloudwatch";
+import {Effect,  Policy, PolicyStatement} from "aws-cdk-lib/aws-iam";
+import {Runtime} from "aws-cdk-lib/aws-lambda";
+import {CfnRecordSet} from "aws-cdk-lib/aws-route53";
+import {CfnInclude} from "aws-cdk-lib/cloudformation-include";
+
+export interface StripeWebhookEndpointsProps extends GuStackProps {
+    stack: string;
+    stage: string;
+    deployBucket: string;
```
johnduffell, line 19: is this deployBucket prop used anywhere?  I can't spot it (maybe it's coming from SSM instead?)

cdk/lib/stripe-webhook-endpoints.ts
```diff
@@ -1,15 +1,222 @@
...
+
+        // ---- Existing CFN template ---- //
+        const yamlTemplateFilePath = path.join(__dirname, "../..", "handlers/stripe-webhook-endpoints/cfn.yaml");
         new CfnInclude(this, "YamlTemplate", {
             templateFile: yamlTemplateFilePath,
         });
+
+        // ---- API-triggered lambda functions ---- //
+
+        //PaymentIntentIssuesLambda
```
johnduffell, line 39: probably don't need these comments as it's clear from the const name (and the comments may get outdated)

cdk/lib/stripe-webhook-endpoints.ts
```diff
@@ -1,15 +1,222 @@
...
+                    path: "/customer-updated",
+                    httpMethod: "POST",
+                    lambda: customerUpdatedLambda,
+                },
+            ],
+            // Create an alarm
+            monitoringConfiguration: {
+                snsTopicName: "conversion-dev",
+                http5xxAlarm: {
+                    tolerated5xxPercentage: 1,
```
johnduffell, line 95: I doubt it will make a difference here, but can we make it alarm on one 5xx error, rather than a percentage?

cdk/lib/stripe-webhook-endpoints.ts
```diff
@@ -1,15 +1,222 @@
...
+        });
+
+        new CfnBasePathMapping(this, "BasePathMapping", {
+            domainName: cfnDomainName.ref,
+            // Uncomment the lines below to reroute traffic to the new API Gateway instance
+            restApiId: stripeWebhookEndpointsApi.api.restApiId,
+            stage: stripeWebhookEndpointsApi.api.deploymentStage.stageName,
+            // Uncomment the lines below to reroute traffic to the old (existing) API Gateway instance
+            // restApiId: yamlDefinedResources.getResource("ServerlessRestApi").ref,
+            // stage: props.stage,
```
johnduffell, line 167: is that true? do we not just switch over the endpoint urls in the stripe console?

cdk/lib/stripe-webhook-endpoints.ts
```diff
@@ -1,15 +1,222 @@
...
+                        `arn:aws:s3:::gu-reader-revenue-private/membership/support-service-lambdas/${props.stage}/zuoraRest-${props.stage}.*.json`,
+                        `arn:aws:s3:::gu-reader-revenue-private/membership/support-service-lambdas/${props.stage}/trustedApi-${props.stage}.*.json`,
+                        `arn:aws:s3:::gu-reader-revenue-private/membership/support-service-lambdas/${props.stage}/stripe-${props.stage}.*.json`,
+                    ],
+                }),
+            ],
+        });
+
+        paymentIntentIssuesLambda.role?.attachInlinePolicy(ssmInlinePolicy);
+        paymentIntentIssuesLambda.role?.attachInlinePolicy(s3InlinePolicy);
```
johnduffell, line 218: consistency
```suggestion
        paymentIntentIssuesLambda.role?.attachInlinePolicy(s3InlinePolicyForPaymentIntentIssues);
```

johnduffell (approved):
looks good, I've put a few small comments for you to check. 👍 

I had a look at the old one and I couldn't find any alarms? either in the CFN or in AWS!  So it's good that you've spotted and added them.
Thanks for going back over and making that PR again, I've learned more about CDK etc from all this!

cdk/lib/stripe-webhook-endpoints.ts
LAKSHMIRPILLAI: I think the GuApiGatewayWithLambdaByPath  props has [monitoringConfiguration](https://guardian.github.io/cdk/interfaces/patterns.GuApiGatewayWithLambdaByPathProps.html#monitoringConfiguration): [NoMonitoring](https://guardian.github.io/cdk/interfaces/constructs_cloudwatch.NoMonitoring.html) | [ApiGatewayAlarms](https://guardian.github.io/cdk/interfaces/patterns.ApiGatewayAlarms.html)

And  Interface ApiGatewayAlarms property http5xxAlarm has only tolerated5xxPercentage

cdk/lib/stripe-webhook-endpoints.ts
LAKSHMIRPILLAI: Usually we follow the dual stack method- running old and new resources together for sometime and monitor.As a next step, once everything is in place and working properly, we will remove this

cdk/lib/stripe-webhook-endpoints.ts
LAKSHMIRPILLAI: Okay, I will remove this, thanks for spotting!

PR #2211, opened 2024-04-05
Updated the product move refund lambda queue behaviour to have retries

LAKSHMIRPILLAI:
## Accessibility

### DIFF
--- a/handlers/product-move-api/cfn.yaml
+++ b/handlers/product-move-api/cfn.yaml
@@ -305,13 +305,14 @@ Resources:
   RefundQueue:
     Type: AWS::SQS::Queue
     Properties:
-      VisibilityTimeout: 3000
+      VisibilityTimeout: 86400
+      MessageRetentionPeriod: 604800
       QueueName:
         Fn::Sub: product-switch-refund-${Stage}
       RedrivePolicy:
         deadLetterTargetArn:
           Fn::GetAtt: RefundDeadLetterQueue.Arn
-        maxReceiveCount: 1
+        maxReceiveCount: 14
   RefundLambdaRole:
     Type: AWS::IAM::Role
     Properties:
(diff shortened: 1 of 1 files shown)


handlers/product-move-api/cfn.yaml
```diff
@@ -305,13 +305,14 @@ Resources:
   RefundQueue:
     Type: AWS::SQS::Queue
     Properties:
-      VisibilityTimeout: 3000
+      VisibilityTimeout: 43200
+      MessageRetentionPeriod: 604800
```
johnduffell, line 309: is it possible to put a comment saying how many hours/days this is?
```suggestion
      VisibilityTimeout: 43200 # 12 hours
      MessageRetentionPeriod: 604800 # 7 days
```

johnduffell:
good idea to keep trying - however this means if we break something it will take a whole week to find out - is that acceptable?
could we queue DD failures on that specific error, but not other failures - if that's not too much work?

handlers/product-move-api/cfn.yaml
rupertbates: I wonder if we should make it a few minutes longer than 7 days as well, just to make sure we get the final retry before the end of the message retention period

rupertbates:
> good idea to keep trying - however this means if we break something it will take a whole week to find out - is that acceptable? could we queue DD failures on that specific error, but not other failures - if that's not too much work?

Can you think of a way to do that? I guess we could only let the lambda fail for these types of errors and for all other types, trigger a different type of alarm, but then we wouldn't get the message saved to the dead letter queue

rupertbates (approved):
A couple of suggestions from John and me but I think we should merge this and monitor how it performs.

handlers/product-move-api/cfn.yaml
LAKSHMIRPILLAI: Yes, I have added the comment and also added 30 min to message retention period

PR #2212, opened 2024-04-05
Fix a rounding error whilst summing prices

rupertbates:
## What does this change?
We noticed a bug in the new-product-api where it was returning the incorrect price for subscription card Sixday subs.
This was caused by the code which sums the charges on that particular product rate plan when converting fractional (Double) prices to minor units prices. 
The code 
```scala
def doubleToMinorUnitsInt(amount: Double): Int = (amount * 100).toInt
```
when given a value of 9.79 returns 978, not as you might reasonably expect 979 and this led to the incorrect prices.

To fix this issue, this PR converts the amount to be multiplied to a BigDecimal before multiplication.
Screenshot 2024-04-05 at 15 48 06


GHaberis (approved):
Looks good

JoeMitchellGuardian (approved):

PR #2214, opened 2024-04-08
add new executable app to generate PROD CAS tokens

johnduffell:
Previously there was a simple test in the CAS lambda to generate a PROD token (if you pasted in the secret)

This works well but generally it's nice to keep secrets away from source code.

This adds a simple "App" to generate a token by fetching the secret from S3, meaning it's usable by the whole team without risk of committing the secret.

Also I updated the readme to signpost that.

### DIFF
--- a/handlers/digital-subscription-expiry/README.md
+++ b/handlers/digital-subscription-expiry/README.md
@@ -1,6 +1,6 @@
 # digital subscription expiry lambdas
 
-## HOWTO generate an emergency token
+## HOWTO generate an emergency token (CAS code) for customer service teams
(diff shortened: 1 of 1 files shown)


handlers/digital-subscription-expiry/README.md
```diff
@@ -1,5 +1,11 @@
 # digital subscription expiry lambdas
 
+## HOWTO generate an emergency token
```
rupertbates, line 3:
```suggestion
## HOWTO generate an emergency token (CAS code)
```

PR #2215, opened 2024-04-09
feat: implement weekly health check flow for the Salesforce Disaster Recovery state machine

AndreaDiotallevi:
## What does this change?

[Trello](https://trello.com/c/stuJuxvX/771-implement-health-check-for-re-syncing-procedure-system)

- Create new application `salesforce-disaster-recovery-health-check` in charge of running weekly health checks on the Salesforce Disaster Recovery state machine:

  - An AWS rule is triggered every Monday at 6AM

  - A function starts an execution of the state machine:
    - the query filters only active accounts created yesterday (e.g. `SELECT Id, Zuora__Zuora_Id__c, Zuora__Account__c, Contact__c FROM Zuora__CustomerAccount__c WHERE CreatedDate = YESTERDAY AND Zuora__Status__c = 'Active'`)

    - A custom execution name is passed so that the state machine can detect whether the execution is a health check (e.g. `health-check-20240411T111548720Z`)
  - the same function sends a notification if the state machine is not healthy (same SNS topic as main application)

- Amend the main `salesforce-disaster-recovery` state machine:

  - Improve state names:
    - `ConstructEmailData` -> `ConstructNotificationData`

    - `SendProcedureSummaryEmail` -> `SendCompletionNotification`

  - Add choice state to skip sending the completion notification if the execution is for health check. The state machine uses intrinsic functions to extract the execution name from the context (e.g. `Condition.stringMatches('$$.Execution.Name', 'health-check-*')`)

## Why are you doing this?

This new testing application makes sure the PROD Salesforce Disaster Recovery re-syncing procedure is available at all times (e.g. expired auth credentials or amended Zuora permission).

Thanks to @johnduffell  for the idea.

## Architecture

<div align="center">
  <table>
    <tr>
      <td align="center" valign="middle">
        Screenshot 2024-04-11 at 13 04 55
      </td>
      <td align="center" valign="middle">
       Screenshot 2024-04-11 at 13 05 20
      </td>
    </tr>
  </table>
</div>

## How to test

Execute the CODE lambda function from the AWS console [here](https://eu-west-1.console.aws.amazon.com/lambda/home?region=eu-west-1#/functions/salesforce-disaster-recovery-health-check-CODE?tab=testing).

Find and monitor the state machine execution [here](https://eu-west-1.console.aws.amazon.com/states/home?region=eu-west-1#/statemachines/view/arn%3Aaws%3Astates%3Aeu-west-1%3A865473395570%3AstateMachine%3Asalesforce-disaster-recovery-CODE).

Screenshot 2024-04-11 at 15 24 28

<div align="center">
Screenshot 2024-04-11 at 15 26 22
</div>


cdk/bin/cdk.ts
```diff
@@ -161,3 +162,19 @@ new AlarmsHandler(app, 'alarms-handler-PROD', {
 	stack: 'support',
 	stage: 'PROD',
 });
+new SalesforceDisasterRecoveryHealthCheck(
+	app,
+	'salesforce-disaster-recovery-health-check-CODE',
+	{
+		stack: 'membership',
```
rupertbates, line 169: I'd like us to use the 'support' stack going forward. Up to you if you want to change that here though.

handlers/salesforce-disaster-recovery-health-check/src/handlers/salesforceDisasterRecoveryHealthCheck.ts
```diff
@@ -0,0 +1,89 @@
...
+
+		let status = 'RUNNING';
+		while (status === 'RUNNING') {
+			const describeExecutionResponse = await describeExecution({
+				executionArn,
+			});
+
+			if (describeExecutionResponse.status === undefined) {
+				throw new Error('Execution status is undefined');
+			}
```
rupertbates, line 58:
```suggestion
			checkDefined(describeExecutionResponse.status, 'Execution status is undefined');	
```

handlers/salesforce-disaster-recovery-health-check/test/firstIntegration.test.ts
```diff
@@ -0,0 +1,10 @@
+/**
```
rupertbates, line 1: Delete this file

handlers/salesforce-disaster-recovery-health-check/test/firstTest.test.ts
```diff
@@ -0,0 +1,7 @@
+/**
```
rupertbates, line 1: Delete this file

rupertbates (approved):
Really good to see this put in place 🥇

cdk/bin/cdk.ts
AndreaDiotallevi: Ah that clarifies it - I wasn't sure.

I might make this change in a separate PR as I need to check if it breaks some hyperlinks.

In general it should be easy to refactor since the state machine will have the same name, it is just the stack that changes.

Regarding the tests, the build failed without tests. So that was probably a smell I needed to add tests. So I have added tests for the handler function.

cdk/lib/salesforce-disaster-recovery.ts
```diff
@@ -370,6 +374,13 @@ export class SalesforceDisasterRecovery extends GuStack {
 			},
 		);
 
+		const haveAllRowsSuccedeed = new Choice(this, 'HaveAllRowsSuccedeed')
+			.when(
+				Condition.numberEquals('$.failedRowsCount', 0),
+				new Succeed(this, 'AllRowsHaveSuccedeed'),
```
johnduffell, line 380: there's a typo here in "succeeded", I guess it is the same in all the places that matter though!

cdk/lib/salesforce-disaster-recovery-health-check.ts
```diff
@@ -0,0 +1,73 @@
...
+			handler: 'salesforceDisasterRecoveryHealthCheck.handler',
+			functionName: `${app}-${this.stage}`,
+			rules: [
+				{
+					schedule: Schedule.cron({
+						minute: '0',
+						hour: '6',
+						weekDay: 'MON',
+						month: '*',
+						year: '*',
```
johnduffell, line 48: I'd probably make it daily, but weekly is fine I'm sure for something rarely used.
Having said that, if we make it every 12 hours, then we can have an alarm that goes off if it doesn't run (similar to the "no X in Y hours" alarms that we have in support-workers.  And I think we have a similar one for IT tests not running.)

cdk/lib/salesforce-disaster-recovery.ts
AndreaDiotallevi: Good spot! I will amend on separate PR since I have merged already. Thanks

cdk/lib/salesforce-disaster-recovery-health-check.ts
AndreaDiotallevi: That was a discussion I had with Graham and he suggested, from the SF side, that weekly was more than enough, given that we might need this system 0 / 1 times over the next few years potentially.

What would be the benefit of having the "no X in Y hours" alarm in this case? At the moment if the state machine fails, a notification is sent, so potentially you would know.

johnduffell: sounds like you've thought about it pretty thoroughly so I'm happy to go with the existing.

To answer your question the benefit would just be that if the state machine stopped kicking off for any reason we would know.  I agree that's less likely than conversions stopping for a particular product - but we would probably never(?) notice if the state machine stopped triggering wheras conversion loss would be noticed within a week.

PR #2216, opened 2024-04-10
Add some scripts to the test-users project

rupertbates:
This PR adds three new scripts to the test-users project to carry out the following actions:
- create a monthly recurring contribution
- create an annual recurring contribution
- update the amount of an existing contribution

See readme for more information

### DIFF
--- a/modules/prettyPrint.ts
+++ b/modules/prettyPrint.ts
@@ -1,0 +1,3 @@
+export const prettyPrint = (object: object) => {
+	return JSON.stringify(object, null, 2);
+};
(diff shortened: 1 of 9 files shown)


AndreaDiotallevi:
Useful functions!

Are these functions different than functions we may use in the main code?

If not, we might want to define these functions inside the main Zuora module and just import them in the test-users module, rather than defining them in the test-users module directly.

rupertbates:
> Useful functions!
> 
> Are these functions different than functions we may use in the main code?
> 
> If not, we might want to define these functions inside the main Zuora module and just import them in the test-users module, rather than defining them in the test-users module directly.

Yes, I did start off doing that, but they are not that generally useful without a bit more refinement so I've left them in the test project for now

PR #2217, opened 2024-04-12
Add email module

rupertbates:
## What does this change?
This PR adds a module to enable sending of emails from Typescript code via Braze and [membership-workflow](https://github.com/guardian/membership-workflow/)

See the integration test for usage.

### DIFF
--- a/modules/email/src/email.ts
+++ b/modules/email/src/email.ts
@@ -1,0 +1,39 @@
+import type { SendMessageCommandOutput } from '@aws-sdk/client-sqs';
+import { SendMessageCommand, SQSClient } from '@aws-sdk/client-sqs';
+import { awsConfig } from '@modules/aws/config';
+import type { Stage } from '@modules/stage';
+
+export type EmailPayload = {
+	Address: string; // email address
+	ContactAttributes: {
+		SubscriberAttributes: Record<string, string>;
+	};
+};
+
+export type DataExtensionName =
+	| 'SV_RCtoSP_Switch'
+	| 'subscription-cancelled-email';
+
+export type EmailMessage = {
+	To: EmailPayload;
+	DataExtensionName: DataExtensionName;
+	SfContactId: string;
+	IdentityUserId?: string;
+};
+
+export const sendEmail = async (
+	stage: Stage,
+	emailMessage: EmailMessage,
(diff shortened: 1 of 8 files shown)


modules/email/src/email.ts
```diff
@@ -0,0 +1,39 @@
...
+	const queueName = `braze-emails-${stage}`;
+	const client = new SQSClient(awsConfig);
+
+	const command = new SendMessageCommand({
+		QueueUrl: queueName,
+		MessageBody: JSON.stringify(emailMessage),
+	});
+
+	const response = await client.send(command);
+	console.log(response);
```
jamesgorrie, line 37: Do we want to remove this, or give it some context?

Maybe some error handling might be more robust?

modules/email/test/emailIntegrationTest.test.ts
```diff
@@ -0,0 +1,38 @@
...
+		To: {
+			Address: emailAddress,
+			ContactAttributes: {
+				SubscriberAttributes: {
+					first_name: 'Support',
+					last_name: 'Service Lambdas',
+					currency: '£',
+					price: '10',
+					first_payment_amount: '7',
+					date_of_first_payment: dayjs().format('DD MMMM YYYY'),
```
jamesgorrie, line 26: I wonder if could use [toLocaleString](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleString) to be specific to a person's local way of reading dates?

This does allow quite a lot of options via [DateTimeformat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/DateTimeFormat).

jamesgorrie: Actually looking at this - could we use `Date.now` as it's just for testing? Seems like an unnecessary dependency.

jamesgorrie (approved):
Great - I wonder if we could handle more cases (errors etc) in from the response?

modules/email/src/email.ts
rupertbates: On errors, I think I'd prefer for them to be handled higher up the call stack. Maybe we could add some more context here but for now I think I'm going to leave as is until there's a clear benefit to adding it.

PR #2220, opened 2024-04-15
feat: make state machine final states more clear

AndreaDiotallevi:
## What does this change?

This makes the state machine final state more clear. In more detail, the final states are:
- `NoAccountsToResync` if the CSV is empty - this was required since an empty CSV breaks the distributed map state
- `HealthCheckSuccessful` when the health check is successful (notifications are skipped)
- `AllAccountsHaveBeenResynced` when all CSV rows have have been updated
- `SomeAccountsHaveFailedToUpdate - Open state input 'failedRowsFileUrl' to debug` when at least one account failed to update - the state name is improve to drive the user to the S3 file for debugging

In addition, the state machine does not fail if some accounts have failed since it is not the state machine itself that failed. That state name has been improved to drive the user to the next manual debugging step.

Also, the notification text has been improved by adding a link to the runbook.

Screenshot 2024-04-19 at 10 09 07

Screenshot 2024-04-19 at 10 08 20


johnduffell (approved):
thanks for updating that! 👍

graham228221 (approved):
Nice, thanks for updating the final states - definitely think it's not a state machine failure to have some failed rows in the final result

PR #2221, opened 2024-04-17
Add billingPeriod to product-catalog

jamesgorrie:
Adds the `billingPeriod` to the `ratePlan` of a product.

This is unconventional to current practice as we hoist this value from the `ratePlanCharges` up to the `ratePlan` to be more useful to the client where we don't really interact with the `ratePlanCharges`.

This is possible due to the fact that they are currently always the same across `ratePlanCharges`, and we have no plans on changing that.

We do error if they are, and this should alert via the normal channels.

## How to test
Tests have been run and snapshots updated.

This is also additive, and the fields optional, so they should have no upstream implications.


jamesgorrie:
@rupertbates the types here are quite free AKA optional strings.

We would tighten it to:
```ts
{ "billingPeriod": "Annual" } |
{ "billingPeriod": "Month" } |
{ "billingPeriod": "Quarter" } |
{ "billingPeriod": "Specific_Weeks", "specificBillingPeriod": 6, }
```

But I also wonder if we should transform the text to `"Monthly"` and `"Quarterly"`, as this language is used quite heavily through the code base. This would allow for quicker and hopefully smoother take-up of this API.

We could then also transform `{ "billingPeriod": "Specific_Weeks", "specificBillingPeriod": 6, }` => `{ "billingPeriod": "SixWeekly" }`.

I might have a go at this for the reason mentioned above of ease of implementation into the current codebase.

modules/product-catalog/src/generateProductCatalog.ts
```diff
@@ -50,6 +50,36 @@ const getProductRatePlanCharges = (
 		}),
 	);
 };
+
+/**
+ * We use this to hoist billing period from the `productRatePlanCharges` up into the `ratePlans`
+ * if-and-only-if all the charges have the same `billingPeriod` and `specificBillingPeriod`.
```
rupertbates, line 56: I'm not aware that we ever use `specificBillingPeriod` for anything, do we need it here?

modules/product-catalog/test/__snapshots__/generateProductCatalog.test.ts.snap
```diff
@@ -173,6 +181,7 @@ exports[`Generated product catalog matches snapshot 1`] = `
         },
       },
       "SixWeekly": {
+        "billingPeriod": "Specific_Weeks",
```
rupertbates, line 184: Is this what you are expecting here?

modules/product-catalog/test/__snapshots__/generateProductCatalog.test.ts.snap
```diff
@@ -187,8 +196,10 @@ exports[`Generated product catalog matches snapshot 1`] = `
           "NZD": 6,
           "USD": 6,
         },
+        "specificBillingPeriod": 6,
```
rupertbates, line 199: Ah, now I see the point of `specificBillingPeriod`

modules/product-catalog/src/productCatalog.ts
```diff
@@ -32,6 +32,8 @@ type ProductRatePlan<
 	charges: {
 		[PRPC in ProductRatePlanChargeKey<P, PRP>]: ProductRatePlanCharge;
 	};
+	billingPeriod?: string;
```
rupertbates, line 35: Does this need to be optional?

modules/product-catalog/src/generateProductCatalog.ts
rupertbates: Ignore this!

modules/product-catalog/src/productCatalog.ts
jamesgorrie: There is an instance in which `billingPeriod` is optional - when a payment is a one off.

I've seen this on gift payments.

jamesgorrie:
@rupertbates thanks for the 👀 on this - I have refactored it out so we use language that is more like the source data `'Annual' | 'Month' | 'Quarter'` and added `SixWeeks` as another option.

It'll mean a slight refactor to the clients of this, but hopefully just some string checking 

(There's a little more than that in `support-workers` as we decode / encode with string discerners)

modules/product-catalog/test/__snapshots__/generateProductCatalog.test.ts.snap
jamesgorrie: I've refactored this out to now be `SixWeeks`

modules/product-catalog/src/productCatalog.ts
jamesgorrie: What I think this model describes is "If `billingPeriod` is here, we will take money repeatedly, if not, we'll take it once and that's all"

rupertbates: Strange that it's there for OneYearGift but not ThreeMonthGift...

modules/product-catalog/src/productCatalogSchema.ts
```diff
@@ -14,6 +14,7 @@ export const productCatalogSchema = z.object({
 					AUD: z.number(),
 				}),
 				charges: z.object({ Subscription: z.object({ id: z.string() }) }),
+				billingPeriod: z.string().optional(),
```
rupertbates, line 17:
```suggestion
				billingPeriod: z.enum(BillingPeriodValues).optional(),
```

modules/product-catalog/test/__snapshots__/generateProductCatalog.test.ts.snap
jamesgorrie: This is not longer relevant as we don't include this `ratePlan` in the catalog.

jamesgorrie:
@rupertbates this has been simplified by removing the need to support `SixWeeks` `ratePlans` ✨.

modules/product-catalog/src/zuoraToProductNameMappings.ts
```diff
@@ -21,13 +21,11 @@ const zuoraCatalogToProductRatePlanKey: Record<string, string> = {
 	'GW Oct 18 - Annual - ROW': 'Annual',
 	'GW Oct 18 - Monthly - ROW': 'Monthly',
 	'GW Oct 18 - Quarterly - ROW': 'Quarterly',
-	'GW Oct 18 - Six for Six - ROW': 'SixWeekly',
 	'GW GIFT Oct 18 - 1 Year - ROW': 'OneYearGift',
 	'GW GIFT Oct 18 - 3 Month - ROW': 'ThreeMonthGift',
 	'GW Oct 18 - Annual - Domestic': 'Annual',
 	'GW Oct 18 - Monthly - Domestic': 'Monthly',
 	'GW Oct 18 - Quarterly - Domestic': 'Quarterly',
-	'GW Oct 18 - Six for Six - Domestic': 'SixWeekly',
```
jamesgorrie, line 30: These have been removed to avoid having to support `SixWeeks` as we don't actually sell this product any more.

modules/zuora/src/zuoraSchemas.ts
```diff
@@ -50,7 +50,7 @@ export const zuoraSubscriptionSchema = z.object({
 					currency: z.string(),
 					effectiveStartDate: z.coerce.date(),
 					effectiveEndDate: z.coerce.date(),
-					billingPeriod: z.enum(BillingPeriodValues),
+					billingPeriod: z.nullable(z.enum(BillingPeriodValues)),
```
rupertbates, line 53: Isn't this optional rather than nullable?

rupertbates (approved):
Great 👍

PR #2222, opened 2024-04-18
rewrite and enforce braces in scala 3

johnduffell:
This PR adds back in and enforces braces in scala 3 code.  This is because it has similar issues to copying into yaml files - copy and pasting lines into the code gets very difficult as you have to also correct indentation, and it's far easier to chuck some code in and keep track of and correct the braces than it is to keep track of the indentation on every single line.  Also the code could actually do the wrong thing if you get the indentation wrong on a single line.

I've raised it as a separate PR into the previous PR https://github.com/guardian/support-service-lambdas/pull/2156 so that it reduces the noise.  If that's not the case we can merge this PR into that one.

To do it I just added `-no-indent` and `-rewrite` and recompiled the code. (see https://docs.scala-lang.org/scala3/reference/other-new-features/indentation.html#settings-and-rewrites-1 ).


build.sbt
```diff
@@ -37,6 +37,7 @@ val scala3Settings = Seq(
     "-Xmax-inlines",
     "256",
     "-Yretain-trees",
+    "-no-indent",
```
johnduffell, line 40: this is the important change.  Intellij seems to pick the setting up automatically and then it adds red underlines on any indentation based syntax

rupertbates (approved):
Great idea 👏 👏 👏

shtukas (approved):

PR #2225, opened 2024-04-19
alarms-handler: app mappings

tomrf1:
Adds support for sending alarms to different google chat channels based on the app name.

- When a message is received, get the ARN of the alarm and fetch the tags for that alarm
- If an `App` tag is present then we attempt to resolve that to a team
- The team is then mapped to a webhook url

If we cannot resolve to a webhook url then the SRE channel is used by default.

### DIFF
--- a/cdk/lib/alarms-handler.ts
+++ b/cdk/lib/alarms-handler.ts
@@ -49,7 +49,7 @@ export class AlarmsHandler extends GuStack {
 				STACK: this.stack,
 				STAGE: this.stage,
 				GROWTH_WEBHOOK: buildWebhookParameter('GROWTH').valueAsString,
-				'P&E_WEBHOOK': buildWebhookParameter('P&E').valueAsString,
+				'PP_WEBHOOK': buildWebhookParameter('PP').valueAsString,
 				VALUE_WEBHOOK: buildWebhookParameter('VALUE').valueAsString,
 				SRE_WEBHOOK: buildWebhookParameter('SRE').valueAsString,
 			},
--- a/cdk/lib/alarms-handler.ts
+++ b/cdk/lib/alarms-handler.ts
@@ -49,7 +49,7 @@ export class AlarmsHandler extends GuStack {
 				STACK: this.stack,
 				STAGE: this.stage,
 				GROWTH_WEBHOOK: buildWebhookParameter('GROWTH').valueAsString,
-				'PP_WEBHOOK': buildWebhookParameter('PP').valueAsString,
+				PP_WEBHOOK: buildWebhookParameter('PP').valueAsString,
 				VALUE_WEBHOOK: buildWebhookParameter('VALUE').valueAsString,
 				SRE_WEBHOOK: buildWebhookParameter('SRE').valueAsString,
 			},
--- a/cdk/lib/alarms-handler.ts
+++ b/cdk/lib/alarms-handler.ts
@@ -30,11 +30,10 @@ export class AlarmsHandler extends GuStack {
 			},
 		});
 
-		const parameters = {
-			webhook: new GuStringParameter(this, `${app}-webhook`, {
-				description: 'Google Chat webhook URL',
-			}),
-		};
+		const buildWebhookParameter = (team: string): GuStringParameter =>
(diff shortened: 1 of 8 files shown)


handlers/salesforce-disaster-recovery/package.json
```diff
@@ -10,12 +10,12 @@
 		"type-check": "tsc --noEmit"
 	},
 	"devDependencies": {
-		"@types/aws-lambda": "^8.10.129",
+		"@types/aws-lambda": "^8.10.137",
```
tomrf1, line 13: necessary because of conflicts with the aws library I added elsewhere

cdk/lib/alarms-handler.ts
```diff
@@ -5,6 +5,7 @@ import { GuLambdaFunction } from '@guardian/cdk/lib/constructs/lambda';
 import type { App } from 'aws-cdk-lib';
 import { Duration } from 'aws-cdk-lib';
 import { ComparisonOperator } from 'aws-cdk-lib/aws-cloudwatch';
+import {Policy, PolicyStatement} from "aws-cdk-lib/aws-iam";
```
AndreaDiotallevi, line 8: It looks like it misses prettier formatting?

handlers/alarms-handler/src/alarmMappings.ts
```diff
@@ -0,0 +1,33 @@
...
+    if (!team) {
+        console.log(`No team found for app: ${appName}, defaulting to SRE`);
+        return 'SRE';
+    }
+    return team;
+}
+
+export const buildWebhookMappings = (): Record<Team, string> => {
+    const getEnvironmentVariable = (team: Team): string => {
+        const prefix = team.toUpperCase();
```
AndreaDiotallevi, line 21: Minor comment, if we have the teams in uppercase above in Team, you can avoid this uppercase step.

handlers/alarms-handler/src/cloudwatch.ts
```diff
@@ -0,0 +1,21 @@
...
+import type {Tag} from "@aws-sdk/client-cloudwatch/dist-types/models/models_0";
+
+const getTags = (alarmArn: string): Promise<Tag[]> => {
+    const client = new CloudWatchClient({ region: "eu-west-1" });
+
+    const request = new ListTagsForResourceCommand({
+        ResourceARN: alarmArn
+    });
+
+    return client.send(request).then((response: ListTagsForResourceCommandOutput) => response.Tags ?? []);
```
AndreaDiotallevi, line 14: This is nice!

handlers/alarms-handler/src/index.ts
```diff
@@ -1,42 +1,52 @@
...
+		return webhookMappings[teamName];
+	}
+}
+
+const processMessage = async (rawMessage: string): Promise<void> => {
+	const message = JSON.parse(rawMessage) as AlarmMessage;
+
+	const webhookUrl = await getWebhookUrl(message);
+
+	const text = `*ALARM:* ${
```
AndreaDiotallevi, line 30: Here if we don't keep the try catch block, we will not be able to use this for SNS publish messages that don't have a structure (the message is just text).

AndreaDiotallevi (approved):

AndreaDiotallevi:
> Adds support for sending alarms to different google chat channels based on the app name.

I am glad this solution has worked out!

cdk/lib/alarms-handler.ts
tomrf1: true. I just run `pnpm format` in the /cdk dir and it's made big changes to all files, so I'd like to keep that change separate from this PR!

handlers/alarms-handler/src/alarmMappings.ts
tomrf1: Fair point, it's better to be uppercase everywhere, I'll make that change

handlers/alarms-handler/src/index.ts
tomrf1: I've refactored things to validate the payload, and if it's not a cloudwatch alarm event then we just send the message to the SRE channel

PR #2229, opened 2024-05-01
feat: improve Zuora to product name mapping

AndreaDiotallevi:
## What does this change?

This improves the Zuora to product name mapping to match what the team is using in the frontend.

[CODE catalog](https://product-catalog.code.dev-guardianapis.com/product-catalog.json).


rupertbates:
Can you update the Zod catalog schema to reflect these changes please?
https://github.com/guardian/support-service-lambdas/blob/d8c56fd8f728bf16c9522d795835e48813f27578/modules/product-catalog/src/productCatalogSchema.ts#L4

AndreaDiotallevi:
> Can you update the Zod catalog schema to reflect these changes please?

Good point! Thanks I have updated now.

modules/product-catalog/src/productCatalogSchema.ts
```diff
@@ -184,6 +184,70 @@ export const productCatalogSchema = z.object({
 				}),
 				billingPeriod: z.enum(BillingPeriodValues).optional(),
 			}),
+			SupporterPlusWithGuardianWeeklyRowMonthly: z.object({
```
jamesgorrie, line 187: Couple things on the naming here (maybe I should have picked up on this earlier).

Could we remove `SupporterPlus`, as we are already part of that product?

I am not too fussy - but we use `Row` here, and `RestOfWorld` elsewhere. I wonder if we remove `SupporterPlus` would we be able to standardise on `GuardianWeeklyRestOfWorld`?

jamesgorrie (commented):
A comment on the naming

modules/product-catalog/src/productCatalogSchema.ts
AndreaDiotallevi: @jamesgorrie 

I have updated to those ones, what do you think?

- GuardianWeeklyRestOfWorldMonthly
- GuardianWeeklyDomesticMonthly
- GuardianWeeklyRestOfWorldAnnual
- GuardianWeeklyDomesticAnnual

AndreaDiotallevi: @jamesgorrie Tagging you again in case you missed this.

rupertbates (approved):
I'm not 100% sure on the naming of these new plans but let's go with it for now

jamesgorrie (approved):

PR #2235, opened 2024-05-03
alarms-handler: make AlarmDescription nullish

tomrf1:
this field can be null, and zod's `optional` doesn't cover this.


tjmw (approved):
Nice!

PR #2244, opened 2024-05-10
Allow alarms from mobile AWS account

tomrf1:
We need alarms in the mobile aws account to publish messages to the SNS topic in the membership account


cdk/lib/alarms-handler.ts
```diff
@@ -1,16 +1,16 @@
-import { GuAlarm } from '@guardian/cdk/lib/constructs/cloudwatch';
```
tjmw, line 1: When I run `pnpm --filter cdk format` it wants these spaces, is your editor maybe formatting differently?

tjmw (approved):
Looks good! (With a formatting question).

PR #2251, opened 2024-05-14
feat: use message attributes to extract app metadata from SNS publish messages

AndreaDiotallevi:
## What does this change?

[Trello](https://trello.com/c/MXFzyYBm/850-sre-add-app-metadata-for-sns-publish-messages-not-only-cloudwatch-alarms)

- Restructure the lambda handler to take into account MessageAttributes coming from a SNS publish message, to extract the app name for mapping purposes.

  Example from the `salesforce-disaster-recovery` application:
  ```typescript
    TopicArn: "...",
    Message: "...",
    MessageAttributes: {
	    app: {
		    DataType: 'String',
		    StringValue: app,
	    },
	    stage: {
		    DataType: 'String',
		    StringValue: this.stage,
	    },
    },
  ```

- Refactor the code to make it more testable
- Add a test for the lambda handler
- Amend `salesforce-disaster-recovery` and `salesforce-disaster-recovery-health-check` applications to send message attributes on SNS publish

## How to test

From `handlers/alarms-handler`, run `pnpm test`.


handlers/alarms-handler/src/index.ts
```diff
@@ -1,42 +1,74 @@
 import type { SNSEventRecord, SQSEvent } from 'aws-lambda';
 import { z } from 'zod';
-import { buildWebhookMappings, getTeam } from './alarmMappings';
+import { getTeam, getTeamWebhookUrl } from './alarmMappings';
 import { getAppNameTag } from './cloudwatch';
 
-const webhookMappings = buildWebhookMappings();
+const cloudWatchAlarmEventRecord = z.object({
```
tjmw, line 6: Are we using this schema to validate the data we receive anywhere? I can see places where we `JSON.parse` and cast, but can't see anywhere we use the schema(s). Sorry if I'm missing something!

tomrf1: Let's stick to using the zod schemas for parsing (as before). It means we get a useful validation error if the payload doesn't match what's expected

AndreaDiotallevi: Thank you for the comments - I have updated now.

Let me know if there is anything else you would like to raise.

tomrf1 (approved):
👍

handlers/alarms-handler/src/cloudwatch.ts
```diff
@@ -2,25 +2,22 @@ import {
...
 };
 
-export const getAppNameTag = (
+export const getAppNameTag = async (
 	alarmArn: string,
 ): Promise<string | undefined> => {
-	return getTags(alarmArn).then((tags: Tag[]) => {
-		return tags.find((tag: Tag) => tag.Key === 'App')?.Value;
-	});
+	const tags = await getTags(alarmArn);
```
tjmw, line 21: 👏 I like these changes to use async/await.

tjmw (approved):
Looks good!

handlers/alarms-handler/src/alarmMappings.ts
```diff
@@ -46,27 +46,13 @@ const appToTeamMappings: Record<string, Team> = Object.entries(
...
 
-export const getTeam = (appName: string): Team => {
-	const team = appToTeamMappings[appName];
-	if (!team) {
-		console.log(`No team found for app: ${appName}, defaulting to SRE`);
-		return 'SRE';
-	}
-	return team;
+export const getTeam = (appName?: string): Team => {
+	return appName ? appToTeamMappings[appName] ?? 'SRE' : 'SRE';
```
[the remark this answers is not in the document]
AndreaDiotallevi: Good suggestion - I have updated.

(Although not exactly like your suggested code because of some temperamental Eslint warning.)

```typescript
export const getTeam = (appName?: string): Team => {
	if (appName && appToTeamMappings[appName]) {
		return appToTeamMappings[appName] as Team;
	}

	return 'SRE';
};
```

PR #2253, opened 2024-05-15
Handle mobile-purchases alarms

tomrf1:
We need to receive alarms from the mobile aws account.

The lambda now checks the `AWSAccountId` of a cloudwatch alarm event. If it's the mobile aws account then it uses the role provided by the mobile-purchases cloudformation. This is because it needs cross-account permissions to fetch the tags for the alarm.

See https://github.com/guardian/mobile-purchases/pull/1511


cdk/lib/alarms-handler.ts
```diff
@@ -91,8 +111,14 @@ export class AlarmsHandler extends GuStack {
 			new PolicyStatement({
 				effect: Effect.ALLOW,
 				actions: ['sns:Publish'],
-				principals: [new AccountPrincipal(mobileAccountId.valueAsString)],
+				// Setting principal to mobileAccountId doesn't work, so we have to restrict the account in the conditions below
```
tomrf1, line 114: this is strange, but is consistent with examples I've seen, e.g. https://repost.aws/knowledge-center/sns-topic-organization-accounts-publish

handlers/alarms-handler/src/alarmMappings.ts
```diff
@@ -19,13 +19,26 @@ const teamToAppMappings: Record<Team, string[]> = {
...
+		'mobile-purchases-google-subscription-status',
+		'mobile-purchases-apple-subscription-status',
+		'mobile-purchases-apple-pubsub',
+		'mobile-purchases-feast-apple-pubsub',
+		'mobile-purchases-google-pubsub',
+		'mobile-purchases-feast-google-pubsub',
+		'mobile-purchases-apple-pubsub',
+		'mobile-purchases-google-pubsub',
+		'mobile-purchases-feast-apple-pubsub',
+		'mobile-purchases-delete-user-subscription',
```
tjmw, line 32: The Value team are actively working on some of these currently. I wonder what a good way of keeping us in the loop on these alarms would be? Just manually pinging us on chat maybe? (And if that doesn't work, we can adapt, possibly send to multiple channels or something).

handlers/alarms-handler/src/cloudwatch.ts
```diff
@@ -2,25 +2,54 @@ import {
...
+			process.env['MOBILE_ROLE_ARN'],
+			'MOBILE_ROLE_ARN environment variable not set',
+		);
+		const credentials = fromTemporaryCredentials({
+			params: { RoleArn: roleArn },
+		});
+		return new CloudWatchClient({ region: 'eu-west-1', credentials });
+	}
+
+	return new CloudWatchClient({ region: 'eu-west-1' });
```
tjmw, line 30: Total nitpick, but since this is the default/happy path would it be clearer to return early if the `awsAccountId` _isn't_ the mobile account?

handlers/alarms-handler/src/alarmMappings.ts
tomrf1: Good point. I can make the change we discussed previously to support multi-team ownership.
Would Value want to receive alarms for all of these apps?

tjmw: Oops sorry I missed the question here - yes I think Value would want to receive them all for now 👍

PR #2264, opened 2024-05-20
SOI consent setter - get SOI product from mobile-purchases response

tjmw:
## What does this change?

Instead of assuming the product is `InAppPurchase`, get the SOI product name from the mobile-purchases response. We now have different products with different SOIs.

### DIFF
--- a/handlers/soft-opt-in-consent-setter/src/test/scala/com.gu.soft_opt_in_consent_setter/ConsentsCalculatorTests.scala
+++ b/handlers/soft-opt-in-consent-setter/src/test/scala/com.gu.soft_opt_in_consent_setter/ConsentsCalculatorTests.scala
@@ -228,7 +228,7 @@ class ConsentsCalculatorTests extends AnyFlatSpec with should.Matchers with Eith
     Handler.buildProductSwitchConsents(
       "guardianweekly",
       "newspaper",
-      Set("newspaper", "mobilesubscription"),
+      Set("newspaper", "InAppPurchase"),
       calculator,
     ) shouldBe Right("""[
         |  {
@@ -246,7 +246,7 @@ class ConsentsCalculatorTests extends AnyFlatSpec with should.Matchers with Eith
     HandlerIAP.buildProductSwitchConsents(
       "guardianweekly",
       "newspaper",
-      Set("newspaper", "mobilesubscription"),
+      Set("newspaper", "InAppPurchase"),
       calculator,
     ) shouldBe Right("""[
         |  {
--- a/handlers/soft-opt-in-consent-setter/src/test/scala/com.gu.soft_opt_in_consent_setter/HandlerTests.scala
+++ b/handlers/soft-opt-in-consent-setter/src/test/scala/com.gu.soft_opt_in_consent_setter/HandlerTests.scala
@@ -26,7 +26,7 @@ class HandlerTests extends AnyFunSuite with Matchers with MockFactory {
   test(testName = "processProductSwitchSub should handle product switch event correctly") {
     val mobileSubscriptions = MobileSubscriptions(
       List(
-        MobileSubscription(true),
+        MobileSubscription(true, "InAppPurchase"),
       ),
     )
 
@@ -72,7 +72,7 @@ class HandlerTests extends AnyFunSuite with Matchers with MockFactory {
   test(testName = "processAcquiredSub should handle acquisition event correctly") {
     val mobileSubscriptions = MobileSubscriptions(
       List(
-        MobileSubscription(true),
+        MobileSubscription(true, "InAppPurchase"),
       ),
     )
 
@@ -103,7 +103,7 @@ class HandlerTests extends AnyFunSuite with Matchers with MockFactory {
   test(testName = "processCancellation should handle supporter plus cancellation while owning an IAP") {
     val mobileSubscriptions = MobileSubscriptions(
       List(
-        MobileSubscription(true),
+        MobileSubscription(true, "InAppPurchase"),
       ),
     )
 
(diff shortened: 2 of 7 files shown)


handlers/soft-opt-in-consent-setter/src/test/scala/com.gu.soft_opt_in_consent_setter/testData/ConsentsCalculatorTestData.scala
```diff
@@ -20,7 +19,7 @@ object ConsentsCalculatorTestData {
     "newspaper" -> newspaperMapping,
     "guardianweekly" -> guWeeklyMapping,
     "testproduct" -> testProductMapping,
-    "mobilesubscription" -> testMobileSubscriptionMapping,
+    "FeastInAppPurchase" -> testFeastMobileSubscriptionMapping,
```
tjmw, line 22: This actually isn't a case of removing one with the other. Some tests were using `mobilesubscription` and some were using the `InAppPurchase` consent group defined just above, but they represent the same thing. I found this confusing so have consolidated on `InAppPurchase`. The `FeastInAppPurchase` is used by the new tests added in this PR.

handlers/soft-opt-in-consent-setter/src/test/scala/com.gu.soft_opt_in_consent_setter/HandlerTests.scala
```diff
@@ -177,4 +177,45 @@ class HandlerTests extends AnyFunSuite with Matchers with MockFactory {
...
+    val mobileSubscriptionsIncludingFeast = MobileSubscriptions(
+      List(
+        MobileSubscription(true, "FeastInAppPurchase"),
+      ),
+    )
+
+    mockSendConsentsReq
+      .expects(
+        "someIdentityId",
+        "[\n  {\n    \"id\" : \"supporter_newsletter\",\n    \"consented\" : false\n  },\n  {\n    \"id\" : \"digital_subscriber_preview\",\n    \"consented\" : false\n  }\n]",
```
johnduffell, line 191: it's hard to read this - you can use triple quotes to make this a bit more readable, then when you paste multi line text in, (at least intellij) will use .stripMargin to indent it nicely too.

johnduffell:
```suggestion
        """[
          | {
          |   "id" : "supporter_newsletter",
          |   "consented" : false
          | },
          | {
          |   "id" : "digital_subscriber_preview",
          |   "consented" : false
          | }
          |]""".stripMargin,
```

handlers/soft-opt-in-consent-setter/src/test/scala/com.gu.soft_opt_in_consent_setter/HandlerTests.scala
```diff
@@ -177,4 +177,45 @@ class HandlerTests extends AnyFunSuite with Matchers with MockFactory {
 
     result shouldBe Right(())
   }
+
+  test(testName = "processCancellation should handle supporter plus cancellation while owning a Feast IAP") {
```
johnduffell, line 181: "handle" is a bit vague - it looks like this test is expecting it to (only?) remove consents for "supporter_newsletter" and "digital_subscriber_preview" (and leave the rest? unchanged?)

handlers/soft-opt-in-consent-setter/src/test/scala/com.gu.soft_opt_in_consent_setter/HandlerTests.scala
```diff
@@ -177,4 +177,45 @@ class HandlerTests extends AnyFunSuite with Matchers with MockFactory {
...
+
+    val result = processCancelledSub(
+      testMessageBody,
+      mockSendConsentsReq,
+      mockGetMobileSubscriptions,
+      calculator,
+      mockSfConnector,
+    )
+
+    result shouldBe Right(())
```
johnduffell, line 219: do we check that the side effecting operation (mockSendConsentsReq) was actually called?

handlers/soft-opt-in-consent-setter/src/test/scala/com.gu.soft_opt_in_consent_setter/ConsentsCalculatorTests.scala
```diff
@@ -228,7 +228,7 @@ class ConsentsCalculatorTests extends AnyFlatSpec with should.Matchers with Eith
     Handler.buildProductSwitchConsents(
       "guardianweekly",
       "newspaper",
-      Set("newspaper", "mobilesubscription"),
+      Set("newspaper", "InAppPurchase"),
```
johnduffell, line 231: this value was ignored before right? and now must match the (something?)

handlers/soft-opt-in-consent-setter/src/test/scala/com.gu.soft_opt_in_consent_setter/HandlerTests.scala
tjmw: Good spot, I've made the test description clearer (hopefully!)

handlers/soft-opt-in-consent-setter/src/test/scala/com.gu.soft_opt_in_consent_setter/ConsentsCalculatorTests.scala
tjmw: It wasn't ignored, it's just that in the test product consents mappings `InAppPurchase` and `mobilesubscription` both map to the same consents, which confused me, so I've consolidated into one (going by the names, I think they both conceptually represented the same thing too, i.e. it's not a coincidence that they match).

handlers/soft-opt-in-consent-setter/src/test/scala/com.gu.soft_opt_in_consent_setter/HandlerTests.scala
tjmw: Yeah I think that's covered by this expectation:

```
    mockSendConsentsReq
      .expects(...)
```

If I change the expected payload (or break the implementation to not call the mocked method) the test fails.

PR #2267, opened 2024-05-21
Send mobile-purchases alarms to Growth and Value teams

tomrf1:
For now we want mobile-purchases alarms to go to both teams

### DIFF
--- a/handlers/alarms-handler/src/alarmMappings.ts
+++ b/handlers/alarms-handler/src/alarmMappings.ts
@@ -2,6 +2,20 @@ import { checkDefined } from '@modules/nullAndUndefined';
 
 type Team = 'VALUE' | 'GROWTH' | 'PP' | 'SRE';
 
+const sharedMobilePurchasesApps = [
+	'mobile-purchases-google-oauth',
+	'mobile-purchases-google-subscription-status',
+	'mobile-purchases-apple-subscription-status',
+	'mobile-purchases-apple-pubsub',
+	'mobile-purchases-feast-apple-pubsub',
+	'mobile-purchases-google-pubsub',
+	'mobile-purchases-feast-google-pubsub',
+	'mobile-purchases-apple-pubsub',
+	'mobile-purchases-google-pubsub',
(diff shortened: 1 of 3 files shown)


handlers/alarms-handler/src/alarmMappings.ts
```diff
@@ -49,22 +54,28 @@ const teamToAppMappings: Record<Team, string[]> = {
...
-);
+const buildAppToTeamMappings = (): Record<string, Team[]> => {
+	const mappings: Record<string, Team[]> = {};
+
+	for (const [team, apps] of Object.entries(teamToAppMappings)) {
+		for (const app of apps) {
+			const teams = mappings[app] ?? [];
+			teams.push(team as Team);
+
+			mappings[app] = teams;
```
tjmw, line 65: I wonder if this is any clearer? Maybe it makes the fact that we're mutating/pushing onto `mappings[app]` a bit more obvious?

```suggestion
			mappings[app] ||= [];
			mappings[app].push(team);
```

tjmw (approved):
Nice! 👍

handlers/alarms-handler/src/alarmMappings.ts
tjmw: Ah right yeah good point - final decision is yours!

tjmw: There's also [`??=`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_assignment) which is maybe slightly more correct here.

tomrf1: I'm starting to wonder if there are too many operators in the language now!

tomrf1: typescript seems to need an intermediary value - when I use either of these operators, it doesn't believe that `mappings[app]` exists -

Screenshot 2024-05-21 at 14 40 39

PR #2270, opened 2024-05-21
Remove duplicate discount-api alarm

rupertbates:
## What does this change?
The discount-api has two alarms defined for 5XX responses so we get duplicate emails whenever there is a failure. This removes the percentage one which is a bit less easy to understand in my opinion.


johnduffell (approved):
yeah I've never liked those percentage alarms that we seem to be moving to

PR #2275, opened 2024-05-22
Add supporter plus V1 to the product catalog

rupertbates:
## What does this change?
Up until now the product catalog has not had the old V1 supporter plus rate plans in it because we no longer sell them. However we do still have subscriptions on those plans, so to work with them as we will need to do in the new-update-supporter-plus lambda which I am about to add, we will need to add them to the catalog.

I have given them ugly names (V1LegacyMonthly and V1LegacyAnnual) so it is very clear that they are no longer current.


modules/product-catalog/src/productCatalogSchema.ts
```diff
@@ -152,6 +152,38 @@ export const productCatalogSchema = z.object({
 	}),
 	SupporterPlus: z.object({
 		ratePlans: z.object({
+			V1LegacyMonthly: z.object({
```
jamesgorrie, line 155: We could use `deprecated` to indicate not to bring it back to life?

Should we leave a note here describing that? Could be useful to surface to clients of the API

jamesgorrie (approved):
Just a thought on naming and socialisation of what this is

modules/product-catalog/src/productCatalogSchema.ts
rupertbates: How would you do that?

jamesgorrie: Sorry, I wasn't very clear, could we call it `V1DeprecatedMonthly`.

`OpenAPI` [use a deprecated field](https://swagger.io/specification/) - maybe we could do the same? It gives us something we could filter by in terms of surfacing or allowing products to be bought in future. 

Maybe overkill though.

rupertbates: I've updated the names, I'm going to hold off on the OpenAPI stuff fn

PR #2288, opened 2024-05-28
Increase the message retention period of the refund lambda dead letter queue

rupertbates:
## What does this change?
The product move refund SQS queue and lambda are set up to retry for seven days as described [in this PR](https://github.com/guardian/support-service-lambdas/pull/2211). After that point failed messages are moved to a dead letter queue so that they can be inspected and then hopefully 'redrived' to carry out the refund. 

Unfortunately the message retention period on the dead letter queue is currently only four days, and as [described here](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-dead-letter-queues.html#understanding-message-retention-periods) the retention period of the dead letter queue starts from when the message is added to the ORIGINAL queue. This means that messages which have been retrying for seven days on the original queue will disappear almost immediately when they are added to the dead letter queue.

To fix this issue, this PR increases the message retention period of the dead letter queue to fourteen days.

### DIFF
--- a/handlers/product-move-api/cfn.yaml
+++ b/handlers/product-move-api/cfn.yaml
@@ -91,6 +91,7 @@ Resources:
   RefundDeadLetterQueue:
     Type: AWS::SQS::Queue
     Properties:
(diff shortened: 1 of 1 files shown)


johnduffell (approved):
nice good improvement to the description too
👍

PR #2289, opened 2024-05-28
Migrate 'update supporter plus amount' functionality to new Typescript lambda

rupertbates:
## What does this change?
This PR creates a new microservice update-supporter-plus-amounts.support.guardianapis.com to replace the endpoint in the product-move-api. The old endpoint failed a lot, mostly with annual subscriptions (see reason below), this fixes that bug.

The reasons for breaking this out into a new api are:
- the functionality is not really related to product switching, it was bundled in there for historical reasons
- it is the next step in moving all the product-move-api functionality out of ZIO and into Typescript
- it uses the newer Zuora Orders api rather than the now deprecated Subscribe and Amend api

## More details on the Annual bug: 
Annual subscriptions which were switched to supporter plus before the product-move-api was changed to start a new term on switch have a next billing date which is after the term end date. As Zuora doesn't allow amendments beyond the term end date, this means that any attempt to update the amount paid on a S+ annual sub in this state will fail with the error message 'The Contract effective date should not be later than the term end date of the basic subscription'.
See error number 9 in [this document](https://docs.google.com/document/d/1iHQ_f0t7mmyuUth7U3F4u6fkp4Q4OOjMiZy95e2NAdU/edit#heading=h.tkudho29f285) for more details.

## Testing
This PR has been tested on code in conjunction with [this manage-fronted PR](https://github.com/guardian/manage-frontend/pull/1342)


handlers/product-switch-api/src/schemas.ts
```diff
@@ -66,93 +66,3 @@ export const zuoraGetAmendmentResponseSchema = z.object({
 export type ZuoraGetAmendmentResponse = z.infer<
 	typeof zuoraGetAmendmentResponseSchema
 >;
-
```
rupertbates, line 69: All of this has moved to a file in the zuora module so it is usable from the new project

modules/zuora/src/orders.ts
```diff
@@ -0,0 +1,102 @@
+export type ProcessingOptions = {
```
rupertbates, line 1: This is all the code which was moved, see comment above

modules/zuora/test/fixtures/subscription.json
```diff
@@ -0,0 +1,247 @@
+{
```
rupertbates, line 1: Just a test fixture

modules/zuora/test/zuoraSchemas.test.ts
```diff
@@ -0,0 +1,9 @@
+import { zuoraSubscriptionSchema } from '@modules/zuora/zuoraSchemas';
+import subscriptionJson from './fixtures/subscription.json';
+
+test('ChargedThroughDate is null in the model when it is null in Zuora', async () => {
```
rupertbates, line 4: Previously when this date was null in the response from zuora, it was defaulting to a min value date in the model

modules/zuora/src/zuoraSchemas.ts
```diff
@@ -52,7 +53,7 @@ export const zuoraSubscriptionSchema = z.object({
 					effectiveEndDate: z.coerce.date(),
 					billingPeriod: z.nullable(z.enum(BillingPeriodValues)),
 					processedThroughDate: z.coerce.date(),
-					chargedThroughDate: z.coerce.date(),
+					chargedThroughDate: z.coerce.date().nullable(),
```
rupertbates, line 56: This can be null in Zuora, previously when null it was just defaulting here to a min value date

handlers/update-supporter-plus-amount/test/fixtures/subscription.json
```diff
@@ -0,0 +1,398 @@
+{
```
rupertbates, line 1: Just a test fixture

handlers/product-switch-api/src/index.ts
```diff
@@ -20,8 +20,8 @@ export const handler: Handler = async (
 };
 
 const routeRequest = async (event: APIGatewayProxyEvent) => {
-	const parsedUrlPath = parseUrlPath(event.path);
 	try {
+		const parsedUrlPath = parseUrlPath(event.path);
```
rupertbates, line 24: Seems like this should be inside the try as it can fail

shtukas (approved):

cdk/lib/update-supporter-plus-amount.ts
```diff
@@ -0,0 +1,191 @@
...
+		};
+
+		// ---- API-triggered lambda functions ---- //
+		const lambda = new GuApiLambda(this, `${app}-lambda`, {
+			description:
+				'An API Gateway triggered lambda to carry out supporter plus amount updates',
+			functionName: nameWithStage,
+			fileName: `${app}.zip`,
+			handler: 'index.handler',
+			runtime: Runtime.NODEJS_18_X,
```
johnduffell, line 45: should we be using 20 for new things? https://aws.amazon.com/blogs/compute/node-js-20-x-runtime-now-available-in-aws-lambda/

cdk/lib/update-supporter-plus-amount.ts
```diff
@@ -0,0 +1,191 @@
...
+			memorySize: 1024,
+			timeout: Duration.seconds(300),
```
johnduffell, line 47: if this is an interactive endpoint, 5 minutes seems long.  is the idea it will complete in the background if zuora is slow even if the browser gives up?

cdk/lib/update-supporter-plus-amount.ts
```diff
@@ -0,0 +1,191 @@
...
+			environment: commonEnvironmentVariables,
+			monitoringConfiguration: {
+				noMonitoring: true,
```
johnduffell, line 50: (not for this PR) but if we have a pattern of turning off monitoring and adding our own alarm, should we extract that as some kind of function/pattern?

cdk/lib/update-supporter-plus-amount.ts
```diff
@@ -0,0 +1,191 @@
...
+			},
+			app: app,
+			api: {
+				id: nameWithStage,
+				restApiName: nameWithStage,
+				description: 'API Gateway created by CDK',
```
johnduffell, line 56: this doesn't seem useful should we just have a blank description?

cdk/lib/update-supporter-plus-amount.ts
```diff
@@ -0,0 +1,191 @@
...
+						`arn:aws:sqs:${this.region}:${this.account}:supporter-product-data-${this.stage}`,
+					],
+				}),
+			],
+		});
+
+		lambda.role?.attachInlinePolicy(s3InlinePolicy);
+		lambda.role?.attachInlinePolicy(secretsManagerPolicy);
+		lambda.role?.attachInlinePolicy(sqsPolicy);
+	}
```
johnduffell, line 190: I guess overall it just feels like for all the supposed benefits of CDK we are adding a lot of similar code for a new lambda, just like in the old CFN days, I thought the promise was that we could have a pattern that would get rid of most of the noise?

handlers/update-supporter-plus-amount/riff-raff.yaml
```diff
@@ -0,0 +1,25 @@
+stacks:
```
johnduffell, line 1: rumour has it that GuCDK generates this file for us in its output directory (in such a way that you don't need to do the riffraff shuffle the first time you deploy)

handlers/update-supporter-plus-amount/src/index.ts
```diff
@@ -0,0 +1,62 @@
...
+	try {
+		console.log(`Input is ${JSON.stringify(event)}`);
+		const subscriptionNumber = getSubscriptionNumberFromUrl(event.path);
+		const eventBody = checkDefined(event.body, 'No request body provided');
+		const requestBody = requestBodySchema.parse(JSON.parse(eventBody));
+		const identityId = checkDefined(
+			event.headers['x-identity-id'],
+			'Identity ID not found in request',
+		);
+		const zuoraClient = await ZuoraClient.create(stage);
```
johnduffell, line 30: maybe it's early to invoke [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself), but all the handler lines above this are going to be common to every lambda called by API gateway (and possibly the one below and fetching the subscription and checking the identity id) so should we have a wrapper handler for that?

handlers/update-supporter-plus-amount/src/index.ts
```diff
@@ -0,0 +1,62 @@
...
+		console.log('Caught error in index.ts ', error);
+		if (error instanceof ValidationError) {
+			console.log(`Validation failure: ${error.message}`);
+			return {
+				body: error.message,
+				statusCode: 400,
+			};
+		} else {
+			return {
+				body: JSON.stringify(error),
```
johnduffell, line 57: don't return the error to the client in case it's got internal details, log it and return a generic "check the logs" style error

handlers/update-supporter-plus-amount/src/index.ts
rupertbates: Yes I think that will make sense at some point, I don't want to add complexity before it becomes clear what the useful patterns are though

handlers/update-supporter-plus-amount/riff-raff.yaml
rupertbates: I looked into this a while back and there was a reason why I stayed with the riff-raff.yaml at that point, but it would be worth revisiting

handlers/update-supporter-plus-amount/src/updateSupporterPlusAmount.ts
```diff
@@ -0,0 +1,216 @@
...
+			planHasSeparateContributionCharge: false,
+		},
+		[productCatalog.SupporterPlus.ratePlans.V1DeprecatedAnnual.id]: {
+			productRatePlan:
+				productCatalog.SupporterPlus.ratePlans.V1DeprecatedAnnual,
+			charge:
+				productCatalog.SupporterPlus.ratePlans.V1DeprecatedAnnual.charges
+					.Subscription,
+			planHasSeparateContributionCharge: false,
+		},
```
johnduffell, line 70: how many people are still on the old charge?  are there just some annual people who haven't quite hit the anniversary when it will switch?

handlers/update-supporter-plus-amount/src/updateSupporterPlusAmount.ts
```diff
@@ -0,0 +1,216 @@
...
+				const { productRatePlan, planHasSeparateContributionCharge } =
+					productData;
+				const contributionCharge = checkDefined(
+					ratePlan.ratePlanCharges.find(
+						(charge) =>
+							charge.productRatePlanChargeId === productData.charge.id,
+					),
+					'Contribution charge not found in rate plan',
+				);
+				acc.push({
```
johnduffell, line 87: if you're using push, that mutates the original array, so using reduce complicates it.  perhaps you could just use foreach or similar if you're relying on mutation, then it would be clear what's happening?

handlers/update-supporter-plus-amount/src/updateSupporterPlusAmount.ts
rupertbates: No there are quite a few monthlys as well. I'm not sure why, probably issues with the price migration engine

handlers/update-supporter-plus-amount/src/updateSupporterPlusAmount.ts
```diff
@@ -0,0 +1,216 @@
...
+		},
+	};
+
+	const productRatePlans = ratePlans
+		.filter((ratePlan) => ratePlan.lastChangeType !== 'Remove')
+		.reduce((acc: SupporterPlusData[], ratePlan: RatePlan) => {
+			const productData = supporterPlusProductData[ratePlan.productRatePlanId];
+			if (productData !== undefined) {
+				const { productRatePlan, planHasSeparateContributionCharge } =
+					productData;
```
johnduffell, line 79: why pull these two out up front if they're only used once at the end?

handlers/update-supporter-plus-amount/src/updateSupporterPlusAmount.ts
```diff
@@ -0,0 +1,216 @@
...
+			productRatePlan:
+				productCatalog.SupporterPlus.ratePlans.V1DeprecatedAnnual,
+			charge:
+				productCatalog.SupporterPlus.ratePlans.V1DeprecatedAnnual.charges
+					.Subscription,
+			planHasSeparateContributionCharge: false,
+		},
+	};
+
+	const productRatePlans = ratePlans
```
johnduffell, line 73: something like this..
```suggestion
	const updatableChargeData = ratePlans
```

handlers/update-supporter-plus-amount/src/updateSupporterPlusAmount.ts
```diff
@@ -0,0 +1,216 @@
...
+					contributionCharge,
+					planHasSeparateContributionCharge,
+				});
+			}
+			return acc;
+		}, []);
+
+	if (productRatePlans.length !== 1 || productRatePlans[0] === undefined) {
+		throw new Error(
+			`Expected 1 rate plan for Supporter Plus, got ${ratePlans.length}`,
```
johnduffell, line 99: need to know which , at least the ID if this would be overkill:
```suggestion
			`Expected 1 rate plan for Supporter Plus, got ${ratePlans.length}: ${productRatePlans}`,
```

handlers/update-supporter-plus-amount/src/updateSupporterPlusAmount.ts
rupertbates: Good point, I've removed the mutation

handlers/update-supporter-plus-amount/src/updateSupporterPlusAmount.ts
```diff
@@ -0,0 +1,216 @@
...
+	}
+
+	const supporterPlusData = getSupporterPlusPlans(
+		productCatalog,
+		subscription.ratePlans,
+	);
+
+	const billingPeriod = checkDefined(
+		supporterPlusData.productRatePlan.billingPeriod,
+		`Billing period was undefined in product rate plan ${prettyPrint(supporterPlusData.productRatePlan)}`,
```
johnduffell, line 177: is this somethign we see, or just an abundance of caution?

handlers/update-supporter-plus-amount/src/updateSupporterPlusAmount.ts
rupertbates: It makes the object returned at the end neater
```typescript
{
    ratePlan,
    productRatePlan,
    contributionCharge,
    planHasSeparateContributionCharge,
}
```
rather than
```typescript
{ 
    ratePlan,
    productRatePlan: productData.productRatePlan,
    contributionCharge,
    planHasSeparateContributionCharge : productData.planHasSeparateContributionCharge,
}
```
Just personal preference, but I think it's neater.

handlers/update-supporter-plus-amount/test/fixtures/subscription.json
```diff
@@ -0,0 +1,398 @@
...
+  "IPCountry__c": null,
+  "CancelledBy__c": null,
+  "dummy__c": null,
+  "PromotionCode__c": null,
+  "OriginalSubscriptionStartDate__c": null,
+  "LegacyContractStartDate__c": null,
+  "CancellationReason__c": null,
+  "billToContact": {
+    "id": "8ad09be48b6b4f7d018b6d422e3835bc",
+    "address1": "Kings Place - giftee",
```
johnduffell, line 79: probably delete that to reduce confusion
```suggestion
    "address1": "Kings Place",
```

handlers/update-supporter-plus-amount/test/fixtures/subscription.json
```diff
@@ -0,0 +1,398 @@
...
+    "city": "London",
+    "country": "United Kingdom",
+    "county": null,
+    "fax": null,
+    "state": null,
+    "postalCode": "N1 9GU",
+    "firstName": "SupportPostDeployTestF",
+    "lastName": "SupportPostDeployTestL",
+    "nickname": null,
+    "workEmail": "postdeploytest@theguardian.com",
```
johnduffell, line 121: not sure if this is going to frustrate people searching for post deploy test related stuff in future? could be worth changing

handlers/update-supporter-plus-amount/test/updateSupporterPlusAmountIntegration.test.ts
```diff
@@ -0,0 +1,66 @@
...
+
+	const result = await updateSupporterPlusAmount(
+		zuoraClient,
+		productCatalog,
+		identityId,
+		subscriptionNumber,
+		newPaymentAmount,
+	);
+
+	expect(result.newAmount).toEqual(newPaymentAmount);
```
johnduffell, line 40: is result.newAmount from a billing preview or just returned what we passed in?
for a more thorough test we could pick a random amount within the range and check the billing preview before and after

handlers/update-supporter-plus-amount/test/updateSupporterPlusAmountIntegration.test.ts
```diff
@@ -0,0 +1,66 @@
...
+});
+
+test('We can send a thank you email', async () => {
```
johnduffell, line 43: that's a handy test!  a comment would be useful explaining how to run it on its own.  otherwise you're going to get emails every time someone runs the it-tests

modules/zuora/test/fixtures/subscription.json
```diff
@@ -0,0 +1,247 @@
...
+    "city": null,
+    "country": "United Kingdom",
+    "county": null,
+    "fax": null,
+    "state": null,
+    "postalCode": null,
+    "firstName": "simon",
+    "lastName": "haley",
+    "nickname": null,
+    "workEmail": "simon.haley.freelancer+rc37@observer.co.uk",
```
johnduffell, line 90: probably shouldn't be committing other people's names

handlers/update-supporter-plus-amount/src/updateSupporterPlusAmount.ts
johnduffell: fair enough I agree its neater but for me not as clear, but either way in that case only pull them out just before they're used?  I was looking through trying to work out why only two out of three fields were pulled out and where they were used

handlers/update-supporter-plus-amount/src/updateSupporterPlusAmount.ts
```diff
@@ -0,0 +1,219 @@
...
+					{
+						ratePlan,
+						productRatePlan,
+						contributionCharge,
+						planHasSeparateContributionCharge,
+					},
+				];
+			}
+			return acc;
+		}, []);
```
johnduffell, line 98: having looked at the code, reduce is overkill anyway, all you are doing is mapping and filtering at the same time, you're not using the preceding values in any way.
If there's no flatMap or similar in typescript it could be done with filter and map
```suggestion
    .filter((ratePlan) => supporterPlusProductData[ratePlan.productRatePlanId] !== undefined)
		.map((acc: SupporterPlusData[], ratePlan: RatePlan) => {
			const productData = supporterPlusProductData[ratePlan.productRatePlanId];
			const contributionCharge = checkDefined(
				ratePlan.ratePlanCharges.find(
					(charge) =>
						charge.productRatePlanChargeId === productData.charge.id,
				),
				'Contribution charge not found in rate plan',
			);
			const { productRatePlan, planHasSeparateContributionCharge } = productData;
			return {
				ratePlan,
				productRatePlan,
				contributionCharge,
				planHasSeparateContributionCharge,
			};
		}, []);
```

johnduffell (approved):
I've left a tonne of comments as I was getting my head around it but overall it's good to see how things are going, thanks for sharing 👍

handlers/update-supporter-plus-amount/src/updateSupporterPlusAmount.ts
rupertbates: Not sure actually, I don't think it should it should ever be undefined but I'm not going to investigate right now

handlers/update-supporter-plus-amount/src/updateSupporterPlusAmount.ts
rupertbates: I've refactored to a filter and map

handlers/update-supporter-plus-amount/src/updateSupporterPlusAmount.ts
```diff
@@ -0,0 +1,216 @@
...
+		throw new ValidationError(
+			`Subscription ${subscriptionNumber} does not belong to identity ID ${identityIdFromRequest}`,
+		);
+	}
+	const currency = account.billingAndPayment.currency;
+	if (!isProductCurrency('SupporterPlus', currency)) {
+		throw new Error(`Unsupported currency ${currency}`);
+	}
+
+	const supporterPlusData = getSupporterPlusPlans(
```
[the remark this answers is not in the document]
rupertbates: We need a few things as well as the charge though hence the data name

handlers/update-supporter-plus-amount/src/updateSupporterPlusAmount.ts
```diff
@@ -0,0 +1,225 @@
...
+	applyFromDate: Dayjs,
+): Dayjs | undefined => {
+	if (applyFromDate.isAfter(dayjs(subscription.termEndDate))) {
+		// This will cause an error in Zuora "The Contract effective date should not be later than the term end date of the basic subscription."
+		// because you can't add an update after the current term end date.
+		// This has probably happened because the subscription was updated without a new term being started,
+		// so the charge is no longer aligned with the term. To fix this we can start a new term from
+		// the point at which the current term started.
+		const potentialNewTermStartDate = dayjs(
+			contributionCharge.effectiveStartDate,
```
johnduffell, line 144: this is trying to clean up the subscription back to how it would have been if the correct term changes were made at the time of the switch. (in order to align the term with the new S+ rateplan rather than the old contribution)
Unfortunately this is only possible soon after the switch happened, because if there has been an auto renewal since, the term start date will be later than the switch date.

I think the least messy solution is that if a term amendment is needed, we should extend the current term to reach up to the charged through date (then there will be no need for the additional (renewal) amendment)
Then the on the day the updated amount kicks in, the bill run will renew the subscription automatically.

johnduffell: actually, the above PR #2888 didn't work completely, as some subscriptions latest term might not have even started if it's been given extra renewals e.g. price rise engine.
A new PR will remove support for realigning terms completely: https://github.com/guardian/support-service-lambdas/pull/3079

PR #2292, opened 2024-05-30
extract and name discount-api ids

johnduffell:
This PR is part of the work to add 2 months free to discount api.

This PR extracts the digi sub IDs and gives them names.

This improves clarity of the code.

Perhaps there is already a catalog map somewhere but I couldn't find it?  But this can be extracted to zuora-catalog moule and reused in future.

### DIFF
--- a/handlers/discount-api/src/productToDiscountMapping.ts
+++ b/handlers/discount-api/src/productToDiscountMapping.ts
@@ -39,6 +39,29 @@ export type Discount = {
 	eligibleProductRatePlanIds: string[];
 };
 
+type Product = 'digiSub'
+
+const productRatePlanIds: {
+	[K in 'CODE'|'PROD']: {
+		[K in Product]: {[K in BillingPeriod]: string};
+	};
+} = {
+    CODE: {
+		digiSub: {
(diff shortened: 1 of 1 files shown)


handlers/discount-api/src/productToDiscountMapping.ts
```diff
@@ -39,6 +39,29 @@ export type Discount = {
 	eligibleProductRatePlanIds: string[];
 };
 
+type Product = 'digiSub'
```
johnduffell, line 42: wasn't sure if there was an existing enum?

handlers/discount-api/src/productToDiscountMapping.ts
```diff
@@ -39,6 +39,29 @@ export type Discount = {
...
+
+const productRatePlanIds: {
+	[K in 'CODE'|'PROD']: {
```
johnduffell, line 45: couldn't reuse the existing Stage type because it has CSBX which is a copy of PROD.  I'm not totally sure how CSBX relates to prod in terms of IDs but I left it out for now on the basis that they will be identical every time we copy across?

rupertbates (approved):

PR #2302, opened 2024-06-04
remove unused fields

johnduffell:
I noticed two fields were unused so I removed them.
```
	effectiveStartDate: string;
	effectiveEndDate: string;
```


rupertbates (approved):
👍

PR #2303, opened 2024-06-04
Alarm mappings digital voucher suspension processor

jamesgorrie:
- Adds `digital-voucher-suspension-processor` to `PP` alarm mapping.
- Bumps the `EvaluationPeriods: 2` as we expect some runs to fail, and they should be picked up by the next run.

### DIFF
--- a/handlers/alarms-handler/src/alarmMappings.ts
+++ b/handlers/alarms-handler/src/alarmMappings.ts
@@ -79,12 +79,15 @@ const teamToAppMappings: Record<Team, string[]> = {
 		'frontend',
 		'it-test-runner',
 		'stripe-intent',
+		'workers',
+
+		// support-service-lambdas
+		'digital-voucher-suspension-processor',
 
 		// other
 		'canonical-config',
 		'salesforce-disaster-recovery',
 		'salesforce-disaster-recovery-health-check',
-		'workers',
 	],
 };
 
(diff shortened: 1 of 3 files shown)


handlers/alarms-handler/src/alarmMappings.ts
```diff
@@ -79,12 +79,15 @@ const teamToAppMappings: Record<Team, string[]> = {
...
+		'workers',
+
+		// support-service-lambdas
+		'digital-voucher-suspension-processor',
 
 		// other
 		'canonical-config',
 		'salesforce-disaster-recovery',
 		'salesforce-disaster-recovery-health-check',
-		'workers',
```
jamesgorrie, line 87: I just moved this up to the relevant repo.

handlers/digital-voucher-suspension-processor/cfn.yaml
```diff
@@ -123,15 +123,18 @@ Resources:
...
       EvaluationPeriods: 1
       MetricName: Errors
       Namespace: AWS/Lambda
-      Period: 3600
+      # We set this to 3 hours as the lambda runs hourly
+      # And we do expect intermittent failures from external APIs
+      # This should pick-up on any sustained issues
+      Period: 10800
       Statistic: Sum
       Threshold: 3
```
graham228221, line 139: I think that this would mean we would still get an alarm if there are three errors within a single hourly run (as long as they occur within a three hour evaluation period).

I think we should set:

`EvaluationPeriods` to 3
`Period` to 3600
`Threshold` to 1

That would mean we get an alarm if there is at least one error in three consecutive hourly evaluation periods. There's an example [here](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/AlarmThatSendsEmail.html#alarm-evaluation).

graham228221 (approved):

PR #2304, opened 2024-06-04
add free two months plan and allow for supporter plus monthly

johnduffell:
This PR is part of the cancellation save work for supporter plus.

This adds links between supporter plus monthly to the newly added 2 months free plan.

The means that people calling the endpoint will be offered the appropriate save period.

Tested in CODE on A-S00894132 and it previewed and added the 100% discount correctly
https://apisandbox.zuora.com/platform/subscriptions/8ad093fb8fe33df8018fe7cfb31a5054

I've also added a couple of tests for the "happy" cases, I wasn't sure whether to try to cover too much especially in the IT test.


handlers/discount-api/src/productToDiscountMapping.ts
```diff
@@ -40,26 +39,26 @@ export type Discount = {
 	upToPeriodsType: string;
 };
 
-type Product = 'digiSub';
-
-const catalog: {
-	[K in 'CODE' | 'PROD']: {
-		[K in Product]: { [K in BillingPeriod]: string };
-	};
-} = {
```
johnduffell, line 49: I had to remove the type because if it wanted all the billing periods for all products.

If I made them optional, the code lower down didn't compile because it assumed any might be undefined.

If I remove the type completely, it infers the exact shape and compiles as appropriate.

handlers/discount-api/src/productToDiscountMapping.ts
```diff
@@ -93,6 +98,12 @@ const Discounts: {
 			upToPeriods: 12,
 			upToPeriodsType: 'Months',
 		},
+		cancellationFree2Mo: {
+			productRatePlanId: '8a1299c28fb956e8018fe2c0e12c3ae4',
+			name: 'Cancellation Save Discount - Free for 2 months',
+			upToPeriods: 2,
+			upToPeriodsType: 'Months',
+		},
```
johnduffell, line 106: https://www.zuora.com/apps/Product.do?method=view&id=2c92a0ff5345f9200153559c6d2a3385

handlers/discount-api/src/productToDiscountMapping.ts
```diff
@@ -79,6 +78,12 @@ const Discounts: {
 			upToPeriods: 12,
 			upToPeriodsType: 'Months',
 		},
+		cancellationFree2Mo: {
+			productRatePlanId: '8ad081dd8fd3d9df018fe2b6a7bc379d',
+			name: 'Cancellation Save Discount - Free for 2 months',
+			upToPeriods: 2,
+			upToPeriodsType: 'Months',
+		},
```
johnduffell, line 86: https://apisandbox.zuora.com/apps/Product.do?method=view&id=2c92c0f85721ff7c01572940171363db

handlers/discount-api/src/productToDiscountMapping.ts
```diff
@@ -40,26 +39,26 @@ export type Discount = {
...
 	},
 	PROD: {
 		digiSub: {
 			Month: '2c92a0fb4edd70c8014edeaa4eae220a',
 			Quarter: '2c92a0fb4edd70c8014edeaa4e8521fe',
 			Annual: '2c92a0fb4edd70c8014edeaa4e972204',
 		},
+		supporterPlus: {
+			Month: '8a128ed885fc6ded018602296ace3eb8',
+		},
```
johnduffell, line 61: https://www.zuora.com/apps/Product.do?method=view&id=8a12865b8219d9b4018221061563643f

handlers/discount-api/src/discountEndpoint.ts
```diff
@@ -104,10 +104,10 @@ const getDiscountPreview = async (
 		discount.productRatePlanId,
 	);
 
-	if (!previewResponse.success || previewResponse.invoiceItems.length != 2) {
+	if (!previewResponse.success || previewResponse.invoiceItems.length < 2) {
```
johnduffell, line 107: supporter plus has 2 charges (base plus contribution)
I wasn't sure how valuable this check was/what it's protecting us from? but I kept it in a loose form.  Should I keep it in the limit 2->3 or is minimum 2 ok?  Or remove completely?

handlers/discount-api/test/fixtures/request-bodies/digitalSub-subscribe-body-old-price.ts
```diff
@@ -57,7 +58,7 @@ export const digiSubSubscribeBody = (
 					RatePlanData: [
 						{
 							RatePlan: {
-								ProductRatePlanId: '2c92c0f84bbfec8b014bc655f4852d9d',
+								ProductRatePlanId: catalog.CODE.digiSub.Month,
```
johnduffell, line 61: I updated the existing code to use the const, just so it's a little clearer what's going on.

handlers/discount-api/test/fixtures/request-bodies/supporterplus-subscribe-body-tier2.ts
```diff
@@ -0,0 +1,69 @@
+import { zuoraDateFormat } from '@modules/zuora/common';
```
johnduffell, line 1: this file is mostly a copy of the above one, rather than adding a parameter, but there's definitely scope for more reuse.

handlers/discount-api/test/helpers.ts
```diff
@@ -17,10 +17,72 @@ export const createDigitalSubscription = async (
...
+				BankCode: string;
+				LastName: string;
+			};
+			BillToContact: {
+				FirstName: string;
+				Country: string;
+				LastName: string;
+				WorkEmail: string;
+			};
+		}[];
```
johnduffell, line 81: this type is a bit wild but it doesn't seem to exist in any defined form, it's just from the data.  I wasn't sure what to do here?

handlers/discount-api/test/eligibilityChecker.test.ts
```diff
@@ -61,3 +64,22 @@ test('error', () => {
...
+	);
+	const eligibilityChecker = new EligibilityChecker(catalog);
+	expect(
+		dayjs(
+			eligibilityChecker.getNextBillingDateIfEligible(
+				billingPreview,
+				getDiscountableRatePlan(sub),
+			),
+		).format('YYYY-MM-DD'),
+	).toEqual('2024-07-04');
```
tjmw, line 84: This is totally subjective so feel free to ignore! I'm a fan of the [arrange-act-assert](https://wiki.c2.com/?ArrangeActAssert) pattern in tests, just to make clear what's happening at each phase. What do you think about this slight refactoring?

```suggestion
        // Arrange
        // ..other setup
	const eligibilityChecker = new EligibilityChecker(catalog);
	
	// Act
	const nextDate = eligibilityChecker.getNextBillingDateIfEligible(
	    billingPreview,
	    getDiscountableRatePlan(sub),
	);
	
	// Assert
	expect(dayjs(nextDate).format('YYYY-MM-DD')).toEqual('2024-07-04');
```

handlers/discount-api/test/previewDiscountIntegration.test.ts
```diff
@@ -120,3 +121,44 @@ test('Subscriptions on the new price are eligible', async () => {
 	);
 	expect(cancellationResult.success).toEqual(true);
 }, 30000);
+
+test('Supporter Plus subscriptions are eligible', async () => {
+	const zuoraClient = await ZuoraClient.create(stage);
+
+	console.log('Creating a new S+ subscription');
```
tjmw, line 128: Should this `console.log` still be here?

tjmw (approved):
Looks really good!

I left a few minor comments/suggestions to take or leave, no blockers though!

handlers/discount-api/test/previewDiscountIntegration.test.ts
johnduffell: I think so, but I haven't quite got my head around the logging standards in TS yet.  The logging does seem very spammy in the tests.  This was copied from the "creating a digital subscription" code and follows the pattern.

handlers/discount-api/src/discountEndpoint.ts
johnduffell: funny thing is it only supported digi sub before which has one charge.
So the 2 referred to the single digi sub charge plus the new discount charge.

Now it can be 3 as we have S+ with two charges plus one new discount charge.

Maybe we could look at the subscription and count the charges, and then check for n+1 items in the preview, but I'm not 100% sure what that check was intended to assert.  Perhaps it's just checking that the new charge is being planned to trigger on the right date, otherwise it wouldn't appear in the invoice?

@rupertbates any thoughts?

handlers/discount-api/test/eligibilityChecker.test.ts
johnduffell: yep I'm totally on board with that, I never heard that way of describing it but I do try to follow it, it was a just a bit copy and paste at the moment.
I will check the other tests and keep them all consistent with your suggestion.

johnduffell: although that's an interesting one for the tests that are expected to throw - because catching an exception is not referentially transparent, it's not easy to pull out the "Act" and the "Assert".
`Try` is a very scala-y thing so I'm guessing that or similar is not the done thing in TS?

handlers/discount-api/src/discountEndpoint.ts
rupertbates: I think it's just following the principle of 'set out your expectations and if the situation you're in doesn't match them fail at that point rather than trying to fudge it'. Seems like a reasonable thing to do to me but if it becomes overly complicated when you add S+ into the mix I wouldn't be too upset to lose it.

handlers/discount-api/src/productToDiscountMapping.ts
rupertbates: You could type it more strongly using the types in the product-catalog module:

```typescript
import type { ProductBillingPeriod } from '@modules/product-catalog/productCatalog';


export const catalog: {
	[S in 'CODE' | 'PROD']: {
		[P in 'DigitalSubscription' | 'SupporterPlus']: {
			[BP in ProductBillingPeriod<P>]: string;
		};
	};
} = {
	CODE: {
		DigitalSubscription: {
			Month: '2c92c0f84bbfec8b014bc655f4852d9d',
			Quarter: '2c92c0f84bbfec58014bc6a2d43a1f5b',
			Annual: '2c92c0f94bbffaaa014bc6a4212e205b',
		},
		SupporterPlus: {
			Month: '8ad08cbd8586721c01858804e3275376',
			Annual: '',
		},
	},
	PROD: {
		DigitalSubscription: {
			Month: '2c92a0fb4edd70c8014edeaa4eae220a',
			Quarter: '2c92a0fb4edd70c8014edeaa4e8521fe',
			Annual: '2c92a0fb4edd70c8014edeaa4e972204',
		},
		SupporterPlus: {
			Month: '8a128ed885fc6ded018602296ace3eb8',
			Annual: '',
		},
	},
};
```
But then you would have to provide a value for all the billing periods which each product supports and it looks like you don't want to add one for S+ annual, so up to you if you think this is worthwhile. I suspect probably not.

handlers/discount-api/test/helpers.ts
rupertbates: If you felt adventurous you could upgrade this to use the orders api and define the types outlined here:
https://developer.zuora.com/v1-api-reference/api/operation/POST_Order/
in 
modules/zuora/orders.ts

No pressure though obviously!

handlers/discount-api/test/previewDiscountIntegration.test.ts
rupertbates: On a separate point, I'm not sure how useful all these integration tests are. I was experimenting with setting up a new subscription for every test run, carrying out the action under test and then cancelling the subs, but I think it was of limited use ultimately because you need subs which are in a number of different states to test effectively and that's not really possible to do like this.

johnduffell: fair enough well it's another string to the bow.  I was a bit surprised there were no "end to end" style integration tests that pass in the raw payload and expect the output.

I wonder if you've seen my PR here ? : https://github.com/guardian/support-service-lambdas/pull/2307
we can use it as stated to collect requests in PROD for a short time period, but we could also use it to capture tests that we do manually in CODE (and even run locally in DEV as integration tests) and store them permanently.  Perhaps not quite as easy to read and update as normal tests, but we could improve that.

handlers/discount-api/src/productToDiscountMapping.ts
johnduffell: thanks for the pointer, in this case I think the type just needs to be inferred as then it only compiles if everything it actually uses statically is present.  Since this is not used dynamically, that should be enough.
The one that needs to be right and isn't checked by the compiler is the one that is keyed off the ID in the subscription, but that's just business logic that we need to get right.

handlers/discount-api/src/discountEndpoint.ts
johnduffell: I see, so fail fast and loudly principle rather than leaving zuora in a mess.  I will think about whether to count the charges (on non removed rate plans) later.

johnduffell:
thanks for the comments and suggestions, one comment warranted more work at this stage and it's waiting for review in this PR: https://github.com/guardian/support-service-lambdas/pull/2309

handlers/discount-api/test/eligibilityChecker.test.ts
tjmw: > although that's an interesting one for the tests that are expected to throw - because catching an exception is not referentially transparent, it's not easy to pull out the "Act" and the "Assert".

Yeah totally, I definitely find it harder to follow AAA in these cases.

> Try is a very scala-y thing so I'm guessing that or similar is not the done thing in TS?

True, it's not a language feature, but you do occasionally see abstractions which attempt to provide something similar, it's not very common.

PR #2306, opened 2024-06-05
cancel concurrent builds to avoid double-deploys

johnduffell:
We have noticed some rate exceeded errors when multiple PRs are merged at a similar time.

This is because we have so many stacks, and if two deploys are deploying all the stacks at the same time, AWS errors out and the deploy fails at random.

@akash1810 has a solution which he suggested "I think you're talking about the [concurrency](https://docs.github.com/en/actions/using-jobs/using-concurrency) setting? This is [how Service Catalogue is configured](https://github.com/guardian/service-catalogue/blob/9b826a8153893285599863eea8df849043cff6a7/.github/workflows/ci.yml#L15-L20)."

This PR changes the GHA workflow as suggested so that it only kicks off a new build once the previous one has either finished or been cancelled.

I tested this with adding a dummy commit to this PR and it did cancel the previous build
https://github.com/guardian/support-service-lambdas/actions/runs/9380999928
image

### DIFF
--- a/.github/workflows/ci-scala.yml
+++ b/.github/workflows/ci-scala.yml
@@ -8,7 +8,7 @@ on:
 
 # Ensure we only ever have one build running at a time.
 # If we push twice in quick succession, the first build will be stopped once the second starts.
-# This avoids multiple deploys happening in quick succession causing AWS rate exceeded errors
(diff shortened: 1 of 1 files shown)


jamesgorrie (approved):
🤞

PR #2307, opened 2024-06-05
add code to record and play back tests from real data

johnduffell:
This is a 10% time PR so although I'm convinced it's a good idea, there may need to be more thought applied.

At the moment we always try our best to write tests manually, based on all the typical and error cases we can think of.
Unfortunately sometimes we miss adding tests when we add new products or codepaths, even if they are very common ones.  Or the tests have a gap in them meaning they don't truly test end to end.

This PR tries to address that by recording 2 weeks worth of real requests and responses into S3, and letting us regression test the API e.g. when we bump libraries or refactor.

Test data is recorded here for CODE (and correspondingly for DEV): 
https://eu-west-1.console.aws.amazon.com/s3/buckets/gu-reader-revenue-logs?prefix=CODE/discount-api/&region=eu-west-1&bucketType=general

There are a  couple of issues where the code is non determinstic.
First is this piece of code where the first rejected promise cancels the other and fails the overall promise.  This means sometimes we don't have a billing preview in the store, so then that can fail first when rerunning the test. Not sure how to handle that off hand
https://github.com/guardian/support-service-lambdas/blob/2bc50abd03a6f00e04be0edfc20f1c708709eb7d/handlers/discount-api/src/discountEndpoint.ts#L55-L62

Also there is the date used in a few places, so the test might fail if we move on to the next month between recording the rerunning the tests.  I think that isn't an issue that needs to be fixed in this PR but we would need record and pass the time in rather than getting it during the code execution.


handlers/discount-api/src/index.ts
```diff
@@ -1,32 +1,49 @@
...
-	Handler,
-} from 'aws-lambda';
+import type { APIGatewayProxyEvent, APIGatewayProxyResult } from 'aws-lambda';
 import { discountEndpoint } from './discountEndpoint';
+import { FetchInterface, RequestLogger } from '@modules/zuora/requestLogger';
 
 const stage = process.env.STAGE as Stage;
-export const handler: Handler = async (
+export const handler: (
+	event: APIGatewayProxyEvent,
```
johnduffell, line 9: We only need the `event`, not the other values e.g. `context`, but when I call the handler explicitly from the test it requires those other values.
Rather than magicking them up, I just changed the type.

handlers/discount-api/src/index.ts
```diff
@@ -1,32 +1,49 @@
...
+) => Promise<APIGatewayProxyResult> = async (
 	event: APIGatewayProxyEvent,
 ): Promise<APIGatewayProxyResult> => {
 	console.log(`Input is ${JSON.stringify(event)}`);
-	const response = await routeRequest(event);
+	const requestLogger = new RequestLogger(stage || 'DEV');
+	requestLogger.setRequest(JSON.stringify(event));
```
johnduffell, line 15: the request logger gets a copy of the incoming event to store, and if it's running locally it places them in `DEV` key in the bucket.

handlers/discount-api/test/recordedE2e.test.ts
```diff
@@ -0,0 +1,68 @@
...
+	expect(response.statusCode).toEqual(requestPlayback.response.statusCode);
+	expect(response.body).toEqual(requestPlayback.response.body);
+}
+test('call the lambda with all recorded data', async () => {
+	const keys = await getKeys('DEV');
+	for (const key of keys) {
+		const requestPlayback = await RequestPlayback.loadKey(key);
+		await testSingle(requestPlayback);
+	}
+});
```
johnduffell, line 22: this test iterates through all the saved values in S3 (for DEV) and runs tests against them all. e.g. 
https://eu-west-1.console.aws.amazon.com/s3/buckets/gu-reader-revenue-logs?prefix=DEV/discount-api/&region=eu-west-1&bucketType=general

handlers/discount-api/test/recordedE2e.test.ts
```diff
@@ -0,0 +1,68 @@
...
+
+test('call the lambda with recorded data', async () => {
+	// load data from s3
+	const requestPlayback = await RequestPlayback.load('DEV', '1717622177150');
+	await testSingle(requestPlayback);
+});
```
johnduffell, line 28: this test just runs the one stored data execution by its timestamp (as a number)

handlers/discount-api/test/recordedE2e.test.ts
```diff
@@ -0,0 +1,68 @@
...
+	expect(response.body).toEqual(expectedBody);
+}, 30000);
+
+test('call the lambda to record data - apply', async () => {
+	const testDataObject = JSON.parse(applyTestData);
+	// const expected = '';
+	const response = await handler(testDataObject);
+	expect(response.statusCode).toEqual(200);
+	expect(response.body).toEqual('Success');
+}, 30000);
```
johnduffell, line 44: these plus the data below were useful for running locally to record some data to S3.
They should probably be ignored or written as a proper integration test that creates its own subscription and cancels it afterwards.

modules/zuora/src/bearerTokenProvider.ts
```diff
@@ -40,14 +42,17 @@ export class BearerTokenProvider {
...
 			['client_secret', this.credentials.clientSecret],
 			['grant_type', 'client_credentials'],
-		]);
+		]).toString();
 
-		const response = await fetch(url, {
+		const response = await this.doFetch.execute(url, {
 			method: 'POST',
+			headers: {
+				'Content-type': 'application/x-www-form-urlencoded',
```
johnduffell, line 50: this header needs to be explicitly set - I think if you pass in `URLSearchParams` it does it automatically.

modules/zuora/src/bearerTokenProvider.ts
```diff
@@ -40,14 +42,17 @@ export class BearerTokenProvider {
 			['client_id', this.credentials.clientId],
 			['client_secret', this.credentials.clientSecret],
 			['grant_type', 'client_credentials'],
-		]);
+		]).toString();
```
johnduffell, line 45: `URLSearchParams` doesn't serialise to anything when I try to write it to S3 so I have to make it a json type (i.e. string)

modules/zuora/src/bearerTokenProvider.ts
```diff
@@ -40,14 +42,17 @@ export class BearerTokenProvider {
...
+		const response = await this.doFetch.execute(url, {
 			method: 'POST',
+			headers: {
+				'Content-type': 'application/x-www-form-urlencoded',
+			},
 			body: formData,
 		});
 
-		const json = await response.json();
+		const json = JSON.parse(response.text);
```
johnduffell, line 55: it's hard to preserve the types if I try to keep it as json when writing/reading it to the file in s3, so I just store the string and make peolpe parse it when they get it out.  It might be possible to add a helper extension method but I didn't know how off hand.

modules/zuora/src/requestLogger.ts
```diff
@@ -0,0 +1,196 @@
...
+
+	outgoingFetch: LogRecord[] = [];
+	request?: string;
+
+	setRequest = (request: string) => {
+		if (this.request) console.log('Error: request is already set');
+		this.request = request;
+	};
+
+	setResponse = async (response: ResponseData) => {
```
johnduffell, line 28: response might be better as a string? - for API gateway (this lambda) we always have a statusCode and a body, but that wouldn't be the case if we extend this to SQS or time triggered lambdas later.

handlers/discount-api/test/recordedE2e.test.ts
```diff
@@ -0,0 +1,68 @@
...
+import { getKeys, RequestPlayback } from '@modules/zuora/requestLogger';
+
+async function testSingle(requestPlayback: RequestPlayback) {
+	const response = await routeRequest(
+		JSON.parse(requestPlayback.request),
+		requestPlayback,
+	);
+	console.log(`Response is ${response.statusCode}\n${response.body}`);
+	expect(response.statusCode).toEqual(requestPlayback.response.statusCode);
+	expect(response.body).toEqual(requestPlayback.response.body);
```
johnduffell, line 14: it's a bit tricky to compare the body, as we don't care about the order of the keys.  At the moment it seems to work ok, but it might have to get cleverer if a refactor changes the order of things around.

modules/zuora/src/zuoraClient.ts
```diff
@@ -78,14 +85,12 @@ export class ZuoraClient {
...
 			body,
 		});
-		const json = await response.json();
+		const json = JSON.parse(response.text);
 		console.log('Response from Zuora was: ', JSON.stringify(json, null, 2));
 
 		if (response.ok) {
 			return schema.parse(json);
 		} else {
-			console.error(response.text);
```
johnduffell, line 87: this would just crash it, as if you already did response.json, then response.text fails with an error.

johnduffell:
this is still a WIP but it's being broken down into smaller PRs

PR #2308, opened 2024-06-06
Return a 400 Bad Request response from the product-switch-api if a user tries to switch an already switched subscription

rupertbates:
## What does this change?
We regularly see attempts to carry out recurring contribution to supporter plus switches for subscriptions which have already been switched. 

This is probably an issue with manage-frontend but we definitely don't want to alarm on it as there is no action to take, so this PR returns a 400 when this happens rather than a 500.


shtukas (approved):

johnduffell:
why not return 400 for any subscription that's not valid?  is it just a temporary thing because we want to keep the alarm unless it's this bug that we know about already/aren't going to fix soon?

PR #2310, opened 2024-06-06
Make BillingPeriod an option in the GetSubscription response

rupertbates:
## What does this change?
The GetSubscriptionResponse model used by the product-move-api expects that all rateplan charge objects will have a billing period set, however some (for instance the old membership Friend rate plan) have a null billing period and this was causing crashes.

This PR makes billing period optional in the model


AndreaDiotallevi:
Looks good.

A question: is it a valid solution to instead updating the old rate plans to add the billing period?

AndreaDiotallevi (approved):

rupertbates:
> Looks good.
> 
> A question: is it a valid alternative solution to update the old rate plans to add the billing period instead?

I don't think so, it would be a large amount of work and has the potential to screw things up. Much safer to just handle the null.

johnduffell:
> Looks good.
> 
> A question: is it a valid alternative solution to update the old rate plans to add the billing period instead?

You can't as "friend" is a one time charge, so it doesn't have a billing period by definition.  Although all friends should be cancelled by now, they still exist if people have upgraded (and then switched to contribution perhaps and maybe on to S+)

PR #2315, opened 2024-06-10
Add new Tier Three product into product catalog

rupertbates:
## What does this change?
We are changing the way we represent the new tier three product in Zuora as described here: https://docs.google.com/document/d/1NOata7_3YNsZ97bd312QQ9MYJrmko55CaO4D77ExUhI/edit

This PR updates the generate-product-catalog lambda to add this new product into the product catalog.


modules/product-catalog/src/zuoraToProductNameMappings.ts
```diff
@@ -9,6 +9,7 @@ const zuoraCatalogToProductKey: Record<string, string> = {
 	'Newspaper Digital Voucher': 'SubscriptionCard',
 	Contributor: 'Contribution',
 	'Newspaper Delivery': 'HomeDelivery',
+	'Tier Three': 'TierThree',
```
jamesgorrie, line 12: I am still a little wary of this - are the business using any other language like "Premium"?

rupertbates: Premium tier has been mentioned once, but that is has been used in the apps previously so could be confusing. T3 seems to be staying the distance, but I agree it isn't ideal.

rupertbates: Apparently the business name for it is print + digital, but that clashes with the old print + digital newspaper product. We could go for print + digital 2024 or V2 or something but I don't really like those too much either!

jamesgorrie: Yea, maybe something that is vague and can change is alright.

jamesgorrie (approved):
Lovely

PR #2317, opened 2024-06-10
Add new fields to discount-api responses

johnduffell:
This PR adds the new fields needed in discount-api preview and apply discount responses.  cc @rBangay 

The idea is to display more information in the flow about the actual dates involved, so that the user can understand what they are agreeing to.

As always I'm a TS novice so no comment is too small to make!

Tested and working in CODE on a new supporter plus.
```
{
    "nextPaymentDate": "2024-09-11"
}
```

### DIFF
--- a/handlers/discount-api/src/discountEndpoint.ts
+++ b/handlers/discount-api/src/discountEndpoint.ts
@@ -4,24 +4,30 @@ import { checkDefined } from '@modules/nullAndUndefined';
 import type { Stage } from '@modules/stage';
 import { addDiscount, previewDiscount } from '@modules/zuora/addDiscount';
 import { getBillingPreview } from '@modules/zuora/billingPreview';
+import { zuoraDateFormat } from '@modules/zuora/common';
 import { getAccount } from '@modules/zuora/getAccount';
 import { getSubscription } from '@modules/zuora/getSubscription';
 import { ZuoraClient } from '@modules/zuora/zuoraClient';
 import type { ZuoraSubscription } from '@modules/zuora/zuoraSchemas';
 import { getZuoraCatalog } from '@modules/zuora-catalog/S3';
 import type { APIGatewayProxyEventHeaders } from 'aws-lambda';
+import type { ManipulateType } from 'dayjs';
 import dayjs from 'dayjs';
 import { EligibilityChecker } from './eligibilityChecker';
 import type { Discount } from './productToDiscountMapping';
 import { getDiscountFromSubscription } from './productToDiscountMapping';
 import { applyDiscountSchema } from './requestSchema';
+import type {
+	ApplyDiscountResponseBody,
+	EligibilityCheckResponseBody,
+} from './responseSchema';
 
 export const discountEndpoint = async (
 	stage: Stage,
 	preview: boolean,
 	headers: APIGatewayProxyEventHeaders,
 	body: string | null,
-) => {
+): Promise<EligibilityCheckResponseBody | ApplyDiscountResponseBody> => {
 	const zuoraClient = await ZuoraClient.create(stage);
 	const catalog = await getZuoraCatalog(stage);
 	const eligibilityChecker = new EligibilityChecker(catalog);
@@ -87,6 +93,7 @@ export const discountEndpoint = async (
 			subscription.termEndDate,
 			dateToApply,
 			discount.productRatePlanId,
+			subscription.accountNumber,
 		);
 	}
 };
@@ -96,7 +103,7 @@ const getDiscountPreview = async (
 	subscriptionNumber: string,
 	nextBillingDate: Date,
 	discount: Discount,
-) => {
+): Promise<EligibilityCheckResponseBody> => {
 	const previewResponse = await previewDiscount(
 		zuoraClient,
 		subscriptionNumber,
@@ -116,13 +123,23 @@ const getDiscountPreview = async (
 		(item) => item.chargeAmount + item.taxAmount,
 	);
 
+	const firstDiscountedPaymentDate = zuoraDateFormat(dayjs(nextBillingDate));
+	if (discount.upToPeriodsType !== 'Months') {
+		throw new Error(
+			'only month discount is supported, consider using a billing preview',
+		);
+	}
+	const unit: ManipulateType = 'month';
+	const nextNonDiscountedPaymentDate = zuoraDateFormat(
+		dayjs(nextBillingDate).add(discount.upToPeriods, unit),
+	);
+
 	return {
-		body: JSON.stringify({
-			discountedPrice,
-			upToPeriods: discount.upToPeriods,
-			upToPeriodsType: discount.upToPeriodsType,
-		}),
-		statusCode: 200,
+		discountedPrice,
+		upToPeriods: discount.upToPeriods,
+		upToPeriodsType: discount.upToPeriodsType,
+		firstDiscountedPaymentDate,
+		nextNonDiscountedPaymentDate,
 	};
 };
 
@@ -133,7 +150,8 @@ const applyDiscount = async (
 	termEndDate: Date,
 	nextBillingDate: Date,
 	discountProductRatePlanId: string,
-) => {
+	accountNumber: string,
+): Promise<ApplyDiscountResponseBody> => {
 	console.log('Apply a discount to the subscription');
 	const discounted = await addDiscount(
 		zuoraClient,
@@ -146,12 +164,76 @@ const applyDiscount = async (
 
 	if (discounted.success) {
 		console.log('Discount applied successfully');
+	} else {
+		throw new Error('discount was not applied: ' + JSON.stringify(discounted));
 	}
+
+	const billingPreviewAfter = await getBillingPreview(
+		zuoraClient,
+		dayjs().add(13, 'months'),
+		accountNumber,
+	);
+
+	const nextPaymentDate = getNextNonFreePaymentDate(
+		billingPreviewAfter.invoiceItems.map((entry) => {
+			return {
+				date: entry.serviceStartDate,
+				amount: entry.chargeAmount + entry.taxAmount,
+			};
+		}),
+	);
+
 	return {
-		body: 'Success',
-		statusCode: 200,
+		nextPaymentDate,
 	};
 };
+
+export function getNextNonFreePaymentDate(
+	invoiceItems: Array<{ date: Date; amount: number }>,
+) {
+	const grouped = groupBy(invoiceItems, (ii) => {
+		return zuoraDateFormat(dayjs(ii.date));
+	});
+
+	const ordered = Object.entries(grouped).sort();
+
+	const firstNonFree = ordered.find((entry) => {
+		return (
+			entry[1]
+				.map((i) => {
+					return i.amount;
+				})
+				.reduce((a, b) => {
+					return a + b;
+				}) > 0
+		);
+	});
+
+	if (!firstNonFree) {
+		throw new Error('could not find a non free payment in the preview');
+	}
+
+	const nextPaymentDate = firstNonFree[0];
+	return nextPaymentDate;
+}
+
+function groupBy<A>(array: A[], f: (elem: A) => string): Record<string, A[]> {
+	const accu: Record<string, A[]> = {};
+	for (const elem of array) {
+		const keyToUse = f(elem);
+		let group: A[];
+		const maybeGroup: A[] | undefined = accu[keyToUse];
+		if (!maybeGroup) {
+			group = [];
+			accu[keyToUse] = group;
+		} else {
+			group = maybeGroup;
+		}
+		group.push(elem);
+	}
+	return accu;
+}
+
 const checkSubscriptionBelongsToIdentityId = async (
 	zuoraClient: ZuoraClient,
 	subscription: ZuoraSubscription,
--- a/handlers/discount-api/src/discountEndpoint.ts
+++ b/handlers/discount-api/src/discountEndpoint.ts
@@ -18,21 +18,115 @@ import type { APIGatewayProxyEventHeaders } from 'aws-lambda';
 import type { ManipulateType } from 'dayjs';
 import dayjs from 'dayjs';
 import { EligibilityChecker } from './eligibilityChecker';
-import type { Discount } from './productToDiscountMapping';
 import { getDiscountFromSubscription } from './productToDiscountMapping';
-import { applyDiscountSchema } from './requestSchema';
 import type {
 	ApplyDiscountResponseBody,
 	EligibilityCheckResponseBody,
 } from './responseSchema';
 
-export const discountEndpoint = async (
+export const previewDiscountEndpoint = async (
 	stage: Stage,
-	preview: boolean,
 	headers: APIGatewayProxyEventHeaders,
-	body: string | null,
-): Promise<EligibilityCheckResponseBody | ApplyDiscountResponseBody> => {
+	subscriptionNumber: string,
+): Promise<EligibilityCheckResponseBody> => {
 	const zuoraClient = await ZuoraClient.create(stage);
+
+	const { discount, dateToApply } = await getDiscountToApply(
+		stage,
+		headers,
+		subscriptionNumber,
+		zuoraClient,
+	);
+
+	console.log('Preview the new price once the discount has been applied');
+	const previewResponse = await previewDiscount(
+		zuoraClient,
+		subscriptionNumber,
+		dayjs(dateToApply),
+		discount.productRatePlanId,
+	);
+
+	if (!previewResponse.success || previewResponse.invoiceItems.length < 2) {
+		throw new Error(
+			'Unexpected data in preview response from Zuora. ' +
+				'We expected at least 2 invoice items, one for the discount and at least one for the main plan',
+		);
+	}
+
+	const discountedPrice = sum(
+		previewResponse.invoiceItems,
+		(item) => item.chargeAmount + item.taxAmount,
+	);
+
+	const firstDiscountedPaymentDate = zuoraDateFormat(dayjs(dateToApply));
+	if (discount.upToPeriodsType !== 'Months') {
+		throw new Error(
+			'only discounts measured in months are supported in this version of discount-api',
+		);
+	}
+	const unit: ManipulateType = 'month';
+	const nextNonDiscountedPaymentDate = zuoraDateFormat(
+		dayjs(dateToApply).add(discount.upToPeriods, unit),
+	);
+
+	return {
+		discountedPrice,
+		upToPeriods: discount.upToPeriods,
+		upToPeriodsType: discount.upToPeriodsType,
+		firstDiscountedPaymentDate,
+		nextNonDiscountedPaymentDate,
+	};
+};
+
+export const applyDiscountEndpoint = async (
+	stage: Stage,
+	headers: APIGatewayProxyEventHeaders,
+	subscriptionNumber: string,
+): Promise<ApplyDiscountResponseBody> => {
+	const zuoraClient = await ZuoraClient.create(stage);
+
+	const { subscription, discount, dateToApply } = await getDiscountToApply(
+		stage,
+		headers,
+		subscriptionNumber,
+		zuoraClient,
+	);
+
+	console.log('Apply a discount to the subscription');
+	const discounted = await addDiscount(
+		zuoraClient,
+		subscriptionNumber,
+		dayjs(subscription.termStartDate),
+		dayjs(subscription.termEndDate),
+		dayjs(dateToApply),
+		discount.productRatePlanId,
+	);
+
+	if (discounted.success) {
+		console.log('Discount applied successfully');
+	} else {
+		throw new Error('discount was not applied: ' + JSON.stringify(discounted));
+	}
+
+	const billingPreviewAfter = await getBillingPreview(
+		zuoraClient,
+		dayjs().add(13, 'months'),
+		subscription.accountNumber,
+	);
+
+	const nextPaymentDate = getNextNonFreePaymentDate(
+		billingPreviewToRecords(billingPreviewAfter),
+	);
+
+	return { nextPaymentDate };
+};
+
+async function getDiscountToApply(
+	stage: Stage,
+	headers: APIGatewayProxyEventHeaders,
+	subscriptionNumber: string,
+	zuoraClient: ZuoraClient,
+) {
 	const catalog = await getZuoraCatalog(stage);
 	const eligibilityChecker = new EligibilityChecker(catalog);
 
@@ -41,15 +135,8 @@ export const discountEndpoint = async (
 		'Identity ID not found in request',
 	);
 
-	const requestBody = applyDiscountSchema.parse(
-		JSON.parse(checkDefined(body, 'No body was provided')),
-	);
-
 	console.log('Getting the subscription details from Zuora');
-	const subscription = await getSubscription(
-		zuoraClient,
-		requestBody.subscriptionNumber,
-	);
+	const subscription = await getSubscription(zuoraClient, subscriptionNumber);
 
 	if (subscription.status !== 'Active') {
 		throw new ValidationError(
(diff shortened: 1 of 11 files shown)


handlers/discount-api/src/discountEndpoint.ts
```diff
@@ -4,24 +4,30 @@ import { checkDefined } from '@modules/nullAndUndefined';
...
+	EligibilityCheckResponseBody,
+} from './responseSchema';
 
 export const discountEndpoint = async (
 	stage: Stage,
 	preview: boolean,
 	headers: APIGatewayProxyEventHeaders,
 	body: string | null,
-) => {
+): Promise<EligibilityCheckResponseBody | ApplyDiscountResponseBody> => {
```
johnduffell, line 30: previously this was returning an object saying status 200 and the body already serialised (or throw an exception).  This made it harder to write the tests and reduced the type checking.  Although it did ensure that the it-tests tested the serialisers as well as the code (if that's a thing in typescript?)
I have introduced a union type here so that we know what it will return, and index.js (i.e. the handler) will stringify it.

handlers/discount-api/src/discountEndpoint.ts
```diff
@@ -87,6 +93,7 @@ export const discountEndpoint = async (
 			subscription.termEndDate,
 			dateToApply,
 			discount.productRatePlanId,
+			subscription.accountNumber,
```
johnduffell, line 96: needed to do a billing preview after the discount is applied so we can confirm the date back to the user.

handlers/discount-api/src/discountEndpoint.ts
```diff
@@ -116,13 +123,23 @@ const getDiscountPreview = async (
...
+	const firstDiscountedPaymentDate = zuoraDateFormat(dayjs(nextBillingDate));
+	if (discount.upToPeriodsType !== 'Months') {
+		throw new Error(
+			'only month discount is supported, consider using a billing preview',
+		);
+	}
+	const unit: ManipulateType = 'month';
+	const nextNonDiscountedPaymentDate = zuoraDateFormat(
+		dayjs(nextBillingDate).add(discount.upToPeriods, unit),
+	);
```
johnduffell, line 135: at the moment I just add the relevant amount of months to predict what zuora will do.  This speeds things up as it saves a zuora call, however zuora is always the best place to find out what will be charged.

handlers/discount-api/src/discountEndpoint.ts
```diff
@@ -116,13 +123,23 @@ const getDiscountPreview = async (
...
-			discountedPrice,
-			upToPeriods: discount.upToPeriods,
-			upToPeriodsType: discount.upToPeriodsType,
-		}),
-		statusCode: 200,
+		discountedPrice,
+		upToPeriods: discount.upToPeriods,
+		upToPeriodsType: discount.upToPeriodsType,
+		firstDiscountedPaymentDate,
+		nextNonDiscountedPaymentDate,
```
johnduffell, line 142: these are the two new fields on the preview call

handlers/discount-api/src/discountEndpoint.ts
```diff
@@ -146,12 +164,76 @@ const applyDiscount = async (
 
 	if (discounted.success) {
 		console.log('Discount applied successfully');
+	} else {
+		throw new Error('discount was not applied: ' + JSON.stringify(discounted));
```
johnduffell, line 168: not sure why we logging and returning 200 anyway if it wasn't applied

handlers/discount-api/src/discountEndpoint.ts
```diff
@@ -146,12 +164,76 @@ const applyDiscount = async (
...
 	}
+
+	const billingPreviewAfter = await getBillingPreview(
```
johnduffell, line 171: we could save a call to zuora and use the same prediction as in the preview call, however I'd rather confirm the actual preview from zuora at this stage (and latency is less of an issue once the user has converted)

handlers/discount-api/src/discountEndpoint.ts
```diff
@@ -146,12 +164,76 @@ const applyDiscount = async (
...
+	);
+
 	return {
-		body: 'Success',
-		statusCode: 200,
+		nextPaymentDate,
 	};
 };
+
+export function getNextNonFreePaymentDate(
```
johnduffell, line 191: I wasn't sure whether to break out this function into another file - thoughts?

handlers/discount-api/src/discountEndpoint.ts
```diff
@@ -146,12 +164,76 @@ const applyDiscount = async (
...
+		if (!maybeGroup) {
+			group = [];
+			accu[keyToUse] = group;
+		} else {
+			group = maybeGroup;
+		}
+		group.push(elem);
+	}
+	return accu;
+}
```
johnduffell, line 235: theres a lodash function that does similar available: https://lodash.com/docs/#groupBy - I'm tempted to switch to using that in a later PR, as Tom has said lodash is a well used library and I like my scala style higher order functions!

handlers/discount-api/test/addDiscountIntegration.test.ts
```diff
@@ -0,0 +1,62 @@
+/**
```
johnduffell, line 1: the diff is not ideal here - this was actually called applyDiscountIntegration.test.ts before, but it was not testing that, it was testing addDiscount, so I renamed.
The new applyDiscountIntegration is testing applyDiscount functionality and was missing before.

handlers/discount-api/src/discountEndpoint.ts
```diff
@@ -146,12 +164,76 @@ const applyDiscount = async (
...
+		accountNumber,
+	);
+
+	const nextPaymentDate = getNextNonFreePaymentDate(
+		billingPreviewAfter.invoiceItems.map((entry) => {
+			return {
+				date: entry.serviceStartDate,
+				amount: entry.chargeAmount + entry.taxAmount,
+			};
+		}),
```
johnduffell, line 183: I had to map the InvoiceItem array down to this simplified object in order to make the tests usefully readable - see the new test below: https://github.com/guardian/support-service-lambdas/pull/2317/files#diff-5c3f53328f250b8667d469ecef097de27a5ec4c9c7390d280434f750c0261278

handlers/discount-api/test/helpers.ts
```diff
@@ -80,11 +81,20 @@ export const createSubscription = async (
...
+	const subscribeResponse: ZuoraSubscribeResponse = await zuoraClient.post(
+		path,
+		body,
+		zuoraSubscribeResponseSchema,
+	);
+
+	return checkDefined(
+		subscribeResponse[0]?.SubscriptionNumber,
+		'SubscriptionNumber was undefined in response from Zuora',
+	);
```
johnduffell, line 97: this was repeated in every it-test that creates a sub so I moved it here

handlers/discount-api/src/discountEndpoint.ts
jamesgorrie: There's a few reasons I would try to avoid return types where I can, the TL;DR is that if you use `as const` on objects, you often get the compiler using stricter inferred types.

Trying to find some written literature on this - but here is a good video on it https://www.youtube.com/watch?v=I6V2FkW1ozQ

handlers/discount-api/src/discountEndpoint.ts
```diff
@@ -96,7 +103,7 @@ const getDiscountPreview = async (
 	subscriptionNumber: string,
 	nextBillingDate: Date,
 	discount: Discount,
-) => {
+): Promise<EligibilityCheckResponseBody> => {
```
jamesgorrie, line 106: Not completely related to this PR - but would it make sense for the naming of the return type and method to align e.g. `const elegibilityCheck = async() //...`?

handlers/discount-api/src/discountEndpoint.ts
```diff
@@ -116,13 +123,23 @@ const getDiscountPreview = async (
 		(item) => item.chargeAmount + item.taxAmount,
 	);
 
+	const firstDiscountedPaymentDate = zuoraDateFormat(dayjs(nextBillingDate));
+	if (discount.upToPeriodsType !== 'Months') {
+		throw new Error(
+			'only month discount is supported, consider using a billing preview',
```
jamesgorrie, line 129: I'm not very clear on this error - what is `"month discount"`?

handlers/discount-api/src/discountEndpoint.ts
```diff
@@ -116,13 +123,23 @@ const getDiscountPreview = async (
...
+		);
+	}
+	const unit: ManipulateType = 'month';
```
jamesgorrie, line 132: Do you need to add the type or have it as a value as it's only used in 1 place?

handlers/discount-api/src/discountEndpoint.ts
```diff
@@ -133,7 +150,8 @@ const applyDiscount = async (
 	termEndDate: Date,
 	nextBillingDate: Date,
 	discountProductRatePlanId: string,
-) => {
+	accountNumber: string,
+): Promise<ApplyDiscountResponseBody> => {
```
jamesgorrie, line 154: Same as mentioned above, inferring return types can actually lead to stricter typing.

handlers/discount-api/src/discountEndpoint.ts
```diff
@@ -146,12 +164,76 @@ const applyDiscount = async (
...
-		body: 'Success',
-		statusCode: 200,
+		nextPaymentDate,
 	};
 };
+
+export function getNextNonFreePaymentDate(
+	invoiceItems: Array<{ date: Date; amount: number }>,
+) {
+	const grouped = groupBy(invoiceItems, (ii) => {
```
jamesgorrie, line 194: `ii` makes it feel a little like a name for an index, whereas I think this is `invoiceItem`?

handlers/discount-api/src/discountEndpoint.ts
```diff
@@ -146,12 +164,76 @@ const applyDiscount = async (
...
+		return (
+			entry[1]
+				.map((i) => {
+					return i.amount;
+				})
+				.reduce((a, b) => {
+					return a + b;
+				}) > 0
+		);
+	});
```
jamesgorrie, line 210: Could this logic be reduced to
```ts
// Sort by date, find first amount > 0
const firstNonFree = invoiceItems
  .sort((a, b) => b.date - a.date)
  .find(invoiceItem => invoiceItem.amount > 0)
```

I might be missing something in your logic though and over simplified.

handlers/discount-api/src/discountEndpoint.ts
jamesgorrie: Agreed!

handlers/discount-api/test/helpers.ts
jamesgorrie: This does obfuscate some logic that you'll just have to know exists here, whereas where it was before, it was essentially a proxy for the zuora API and would also make it harder if you want to access any of the rest of the model.

Not too fussed either way, just pointing out a potential case of early abstraction.

handlers/discount-api/src/discountEndpoint.ts
rupertbates: There's a groupBy function here: https://github.com/guardian/support-service-lambdas/blob/b848e1b6565d7fdcbf9dd3663d17d501927cf30a/modules/arrayFunctions.ts#L7

jamesgorrie (commented):
Added a few TS comments. I haven't approved as i am unfamiliar with the domain. I'll dig a little deeper from that perspective in a bit.

handlers/discount-api/src/discountEndpoint.ts
johnduffell: yeah I did notice that the compiler works out some (super complex) structural types that are more strict, however explicit types are also like unit tests in a way, i.e. without them if we deviate from a inferred type, it would just infer a different type (which would happily Json.stringify) wheras if we specify explicitly, then it will check it and fail to compile if it's wrong.

I will look at the video later though as I'm sure there's a lot of nuance.

The actual change on this line is to make it return the structure rather than a stringified json, as strings of json have no type checking.

handlers/discount-api/src/discountEndpoint.ts
johnduffell: I realised that the only thing using that groupBy already is also grouping invoice items by date, so I did a wider rationalisation of the two bits of code just now (in the second commit on this PR)

handlers/discount-api/test/helpers.ts
johnduffell: yeah that's a good point, although I would say this isn't really logic it's more like assertions of postconditions of the zuora response.
you could think of it more akin to the deserialiser checking that the response is valid on a slightly higher level than just the json checked by "JSON" and fields checked by "zod".
The scala way I'm used to is more like check things centrally in the deserialisers and ideally the type system remembers it after that.  In typescript the philosophy seems to be more like check it when you need it, in this codebase it seems to be more repetitive.

handlers/discount-api/src/discountEndpoint.ts
johnduffell: Thanks for the video, it's interesting 

It sounds like`as const` is akin to defining an enum using a sealed trait and case objects in scala (here is more like an Either or a sealed trait with case classes.

Specifically, Typescript always infers the most accurate type it can from what it knows, however, (as per the video and your comment) it would see `"hello"` and assume you want to treat it as a string.  If you write `as const` on the line with `"hello"` in then it will say the type is `"hello"` only rather than string.

Yes it's stricter, but I'm not sure quite how it applies here.

The rest of the video seems to be basically saying that covariance is classed as lying in typescript? and furthermore the type checker might compile invalid code - what does this even mean?
image

It's quite a culture shift from scala where I treat types as little inline unit tests and a way to localise the scope of errors, and intellij encourages them.

johnduffell: although to be fair, if you write the above in scala it also compiles, however the rules of overloading determine which function it resolves to at compile and run time.
https://scastie.scala-lang.org/sSGlJJAKS4aHkHUKiZD3SQ
It's still confusing, but it is well defined what will happen.

handlers/discount-api/src/discountEndpoint.ts
johnduffell: I thought that too - there are two purposes of this endpoint, one it to check eligibility, and if so, it returns a preview, so I guess the naming wasn't made consistent.

handlers/discount-api/src/discountEndpoint.ts
johnduffell: that's just the type that "add" takes (in dayjs library). So it means anyone editing this in future will get the error on the right line, and it also allows auto complete to work 
image

handlers/discount-api/src/discountEndpoint.ts
johnduffell: also, as a point directly related to your suggestion to infer the type - the response from this is simply passed to `JSON.stringify` which can take anything at all, so if we don't have a type here, then it will literally just serialise whatever is passed back, whether we missed, misspelled, or passed back the wrong thing.
Presumably that means that we should either specify the type here or on the line before we call JSON.stringify?

handlers/discount-api/src/discountEndpoint.ts
johnduffell: if we infer the type, then it will also compile if we (incorrectly) return a `EligibilityCheckResponseBody` rather than the correct `ApplyDiscountResponseBody`

handlers/discount-api/src/discountEndpoint.ts
johnduffell: refactor has combined this with an existing function in billingPreview.ts

handlers/discount-api/src/discountEndpoint.ts
johnduffell: thanks this was moved in the refactor but I have applied your suggestion there

handlers/discount-api/src/discountEndpoint.ts
johnduffell: yes this logic was a bit messy so that would have helped - but actually I have completely refactored this in order to use an existing invoice grouping code suggested by rupert, and furthermore I've added a scala style `sortBy` function to avoid having to subtract dates

handlers/discount-api/src/discountEndpoint.ts
```diff
@@ -3,26 +3,130 @@ import { ValidationError } from '@modules/errors';
...
-import type { Discount } from './productToDiscountMapping';
 import { getDiscountFromSubscription } from './productToDiscountMapping';
-import { applyDiscountSchema } from './requestSchema';
+import type {
+	ApplyDiscountResponseBody,
+	EligibilityCheckResponseBody,
+} from './responseSchema';
 
-export const discountEndpoint = async (
+export const previewDiscountEndpoint = async (
```
johnduffell, line 27: before there was one function to cover preview and apply, which took a boolean to tell it which function to call (apply or preview).

I guess the reasoning was because preview mode was originally intended to do the same as apply mode apart from the final amendment call.  However in reality it returns a union type and does some different things.

Now the apply vs preview function are separate, so the union type is no longer needed.

The common code in the old common function has been extracted into `getDiscountToApply` which is called from either.

I think this gives a better separation between the outside interface, while still allowing sharing of the appropriate code.

handlers/discount-api/test/previewDiscountIntegration.test.ts
```diff
@@ -19,24 +19,13 @@ test("Subscriptions which don't belong to the provided identity Id are not eligi
...
-
-	const subscriptionNumber = checkDefined(
-		subscribeResponse[0]?.SubscriptionNumber,
-		'SubscriptionNumber was undefined in response from Zuora',
-	);
-
-	const requestBody = {
-		subscriptionNumber: subscriptionNumber,
-		preview: true,
-	};
```
johnduffell, line 32: after this change, the it tests are simplified (good) but also not testing the deserialisers any more (bad?).  incidentally, `preview: true,` was unused anyway.

modules/zuora/src/billingPreview.ts
```diff
@@ -21,22 +21,55 @@ export const getBillingPreview = async (
...
+}
+
+export function billingPreviewToRecords(billingPreviewAfter: BillingPreview) {
+	return billingPreviewAfter.invoiceItems.map((entry) => {
+		return {
+			date: entry.serviceStartDate,
+			amount: entry.chargeAmount + entry.taxAmount,
+		};
+	});
+}
```
johnduffell, line 75: I wanted to write some unit tests that were readable. Which are in handlers/discount-api/test/getNextPaymentDate.test.ts 
In order to make them readable, I had to have a reasonable function signature rather than what zuora returns.
This gets the messy zuora data structure into the reasonable data.
Then I can test the bits I want to, and not this one.

handlers/discount-api/src/discountEndpoint.ts
rupertbates: sorry I still don't see why you wouldn't just pass 'month' into the add function?

johnduffell: oh me neither, I misunderstood the original comment, I've inlined it now.

 github comment suggestions can help illustrate things as my typescript terminology may be lacking! 
image

handlers/discount-api/test/applyDiscountIntegration.test.ts
```diff
@@ -1,70 +1,49 @@
 /**
```
johnduffell, line 1: this is actually a new test with the old name - the old test was renamed and is just above handlers/discount-api/test/addDiscountIntegration.test.ts

handlers/discount-api/test/helpers.ts
johnduffell: oh sorry I realise you meant we are actually extracting the subscription number as well as checking it.  Yes that's a good point.  I was less fussed because it's a test but I'll have a play.

johnduffell: ok I've renamed this to `subscribe` and made it a basic proxy, and moved this extractor to separate helper methods.

handlers/discount-api/src/discountEndpoint.ts
johnduffell: hopefully the final comment on this thread - I wrote a wrapper for JSON.stringify with a type parameter, so we can at least ensure the type at the point of serialisation, while still inferring the types in the rest of the code.
I think that's a good enough tradeoff between having enough type checking and usual typescript conventions.

modules/zuora/src/billingPreview.ts
```diff
@@ -21,22 +21,55 @@ export const getBillingPreview = async (
...
+	return ordered;
+}
+
+export function billingPreviewToRecords(billingPreviewAfter: BillingPreview) {
+	return billingPreviewAfter.invoiceItems.map((entry) => {
+		return {
+			date: entry.serviceStartDate,
+			amount: entry.chargeAmount + entry.taxAmount,
+		};
+	});
```
tjmw, line 74: If you wanted to, you could omit the `return` keyword in a simple arrow function like this:

```suggestion
	return billingPreviewAfter.invoiceItems.map((entry) => ({
		date: entry.serviceStartDate,
		amount: entry.chargeAmount + entry.taxAmount,
	}));
```

modules/zuora/src/billingPreview.ts
```diff
@@ -21,22 +21,55 @@ export const getBillingPreview = async (
...
+		return zuoraDateFormat(dayjs(invoiceItem.date));
+	});
+
+	const ordered = sortBy(Object.entries(grouped), (item) => item[0]).map(
+		(item) => {
+			return {
+				date: new Date(item[0]),
+				total: sumNumbers(item[1].map((item) => item.amount)),
+			};
+		},
```
tjmw, line 63: You could also omit `return` in these functions if you wanted (see example below).

modules/zuora/src/billingPreview.ts
```diff
@@ -21,22 +21,55 @@ export const getBillingPreview = async (
...
-		groupedInvoiceItems[nextInvoiceDate],
-		'No invoice items found for next invoice date',
+
+	const nextPaymentDate = zuoraDateFormat(dayjs(firstNonFree.date));
+
+	return nextPaymentDate;
+}
+
+function getOrderedInvoiceData(
+	invoiceItems: Array<{ date: Date; amount: number }>,
```
tjmw, line 51: This type appears in a few places - would it be worth extracting and giving it a name?

modules/arrayFunctions.ts
```diff
@@ -17,6 +17,16 @@ export const groupBy = <T>(
 	}, {});
 };
 
+export const sortBy = <T>(array: T[], fn: (item: T) => string): T[] => {
```
tjmw, line 20: Would it be worth adding a test for this? Partly as documentation (as well as verifying the behaviour).

handlers/discount-api/test/getNextPaymentDate.test.ts
```diff
@@ -0,0 +1,65 @@
+import { getNextNonFreePaymentDate } from '@modules/zuora/billingPreview';
+
+test('getNextPaymentDate fails if there are no payments in the preview', () => {
```
tjmw, line 3: Would the description be more accurate?

```suggestion
test('getNextPaymentDate fails if there are no non-free payments in the preview', () => {
```

handlers/discount-api/test/getNextPaymentDate.test.ts
```diff
@@ -0,0 +1,65 @@
...
+		{ date: new Date('2020-03-01'), amount: 10 },
+		{ date: new Date('2020-02-01'), amount: 10 },
+		{ date: new Date('2020-02-01'), amount: 0 },
+		{ date: new Date('2020-01-01'), amount: 0 },
+	];
+
+	const actual = getNextNonFreePaymentDate(items);
+
+	expect(actual).toEqual('2020-02-01');
+});
```
tjmw, line 65: ❤️  These tests!

handlers/discount-api/test/getNextPaymentDate.test.ts
johnduffell: better still, flip the language completely
`getNextNonFreePaymentDate fails if all payments are free`

modules/arrayFunctions.ts
johnduffell: I had a go and it looks like arrayFunctions isn't in a module so it isn't easy to write a test.  I think this will probably be best in another PR to make a module and add tests for that (and groupBy)

johnduffell: added in https://github.com/guardian/support-service-lambdas/pull/2325

handlers/discount-api/src/responseSchema.ts
```diff
@@ -4,8 +4,16 @@ export const previewDiscountSchema = z.object({
...
+	firstDiscountedPaymentDate: z.string(),
+	nextNonDiscountedPaymentDate: z.string(),
 });
 
 export type EligibilityCheckResponseBody = z.infer<
 	typeof previewDiscountSchema
 >;
+
+export const applyDiscountSchema = z.object({
+	nextPaymentDate: z.string(),
```
tjmw, line 16:
```suggestion
	nextNonDiscountedPaymentDate: z.string(),
```
?

handlers/discount-api/src/discountEndpoint.ts
jamesgorrie: > so if we don't have a type here, then it will literally just serialise whatever is passed back, whether we missed, misspelled, or passed back the wrong thing.

Yea, this makes a tonne of sense as to why you might want to type here. Adding the type param to `JSON.stringify()` is also a nice touch!

tjmw (approved):

handlers/discount-api/src/responseSchema.ts
johnduffell: as discussed the naming improvements are in here: https://github.com/guardian/support-service-lambdas/pull/2330

PR #2318, opened 2024-06-12
use a verb to name the check/get method

johnduffell:
"check" methods are confusing to me as they sound like assertions, but actually if the assertion is passing we return the value.

This PR only has one change (although it hits a lot of files) it renames the `check...` method to `get...` to make it more readable as to what it's doing

Opinions welcome (hence it's a separate PR) as I'm not sure if this is a standard naming convention for TS code?


tjmw (approved):
Nice! I think the new name is clearer 👍

PR #2322, opened 2024-06-12
send confirmation email when suspending s+

johnduffell:
This PR is part of the cancellation saves work.

As part of this we need to send out an email to confirm what the user should expect.

This PR adds the code to send the email with the suitable fields.

This is a rough wireframe of what the email has, although it will also use their first name at least
image

### DIFF
--- a/handlers/discount-api/src/discountEndpoint.ts
+++ b/handlers/discount-api/src/discountEndpoint.ts
@@ -17,8 +17,8 @@ import { getZuoraCatalog } from '@modules/zuora-catalog/S3';
 import type { APIGatewayProxyEventHeaders } from 'aws-lambda';
 import dayjs from 'dayjs';
 import { EligibilityChecker } from './eligibilityChecker';
-import type { Discount } from './productToDiscountMapping';
 import { getDiscountFromSubscription } from './productToDiscountMapping';
+import { sendCancellationDiscountEmail } from './sendCancellationDiscountConfirmationEmail';
 
 export const previewDiscountEndpoint = async (
 	stage: Stage,
@@ -35,93 +35,6 @@ export const previewDiscountEndpoint = async (
 	);
 
 	console.log('Preview the new price once the discount has been applied');
-	return await getDiscountPreview(
-		zuoraClient,
-		subscriptionNumber,
-		dateToApply,
-		discount,
-	);
-};
-
-export const applyDiscountEndpoint = async (
-	stage: Stage,
-	headers: APIGatewayProxyEventHeaders,
-	subscriptionNumber: string,
-) => {
-	const zuoraClient = await ZuoraClient.create(stage);
-
-	const { subscription, discount, dateToApply } = await getDiscountToApply(
-		stage,
-		headers,
-		subscriptionNumber,
-		zuoraClient,
-	);
-	return await applyDiscount(
-		zuoraClient,
-		subscriptionNumber,
-		subscription.termStartDate,
-		subscription.termEndDate,
-		dateToApply,
-		discount.productRatePlanId,
-		subscription.accountNumber,
-	);
-};
-
-async function getDiscountToApply(
-	stage: Stage,
-	headers: APIGatewayProxyEventHeaders,
-	subscriptionNumber: string,
-	zuoraClient: ZuoraClient,
-) {
-	const catalog = await getZuoraCatalog(stage);
-	const eligibilityChecker = new EligibilityChecker(catalog);
-
-	const identityId = getIfDefined(
-		headers['x-identity-id'],
-		'Identity ID not found in request',
-	);
-
-	console.log('Getting the subscription details from Zuora');
-	const subscription = await getSubscription(zuoraClient, subscriptionNumber);
-
-	if (subscription.status !== 'Active') {
-		throw new ValidationError(
-			`Subscription ${subscription.subscriptionNumber} has status ${subscription.status}`,
-		);
-	}
-
-	console.log(
-		'Check sub is owned by user & get billing preview for the subscription',
-	);
-	const [, billingPreview] = await Promise.all([
-		checkSubscriptionBelongsToIdentityId(zuoraClient, subscription, identityId),
-		getBillingPreview(
-			zuoraClient,
-			dayjs().add(13, 'months'),
-			subscription.accountNumber,
-		),
-	]);
-
-	console.log('Working out the appropriate discount for the subscription');
-	const { discount, nonDiscountRatePlan } = getDiscountFromSubscription(
-		stage,
-		subscription,
-	);
-
-	console.log('Checking this subscription is eligible for the discount');
-	const dateToApply = eligibilityChecker.getNextBillingDateIfEligible(
-		billingPreview,
-		nonDiscountRatePlan,
-	);
-	return { subscription, discount, dateToApply };
-}
-
-const getDiscountPreview = async (
-	zuoraClient: ZuoraClient,
-	subscriptionNumber: string,
-	dateToApply: Date,
-	discount: Discount,
-) => {
 	const previewResponse = await previewDiscount(
 		zuoraClient,
 		subscriptionNumber,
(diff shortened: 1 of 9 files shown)


handlers/discount-api/src/discountEndpoint.ts
```diff
@@ -35,12 +35,42 @@ export const previewDiscountEndpoint = async (
 	);
 
 	console.log('Preview the new price once the discount has been applied');
-	return await getDiscountPreview(
```
johnduffell, line 38: I've inlined the `getDiscountPreview` function in the end as there was no reason to keep it separate any more, and I needed to inline the similar `getDiscountToApply` function below so I could get all the right account etc fields to the email code without excessive numbers of parameters

handlers/discount-api/src/discountEndpoint.ts
```diff
@@ -50,21 +80,49 @@ export const applyDiscountEndpoint = async (
...
+				firstName: account.billToContact.firstName,
+				lastName: account.billToContact.lastName,
+				identityId: account.basicInfo.identityId,
+			})
+		: undefined;
+
+	return {
+		emailPayload,
+		response: { nextPaymentDate: zuoraDateFormat(dayjs(nextPaymentDate)) },
+	};
```
johnduffell, line 125: this is the main change for this PR - we generate the payload if the discount requires an email, and return it back to the index.ts handler (why do we call it "index" rather than "handler"?  I would prefer something like "cancellationSaveHandler.ts" as it's easier to find using the Open File finder.)

handlers/discount-api/src/productToDiscountMapping.ts
```diff
@@ -37,6 +37,7 @@ export type Discount = {
 	name: string;
 	upToPeriods: number;
 	upToPeriodsType: string;
+	sendEmail: boolean;
```
johnduffell, line 40: I wasn't 100% sure whether to make whether to send an email a property of the discount definition or not.  It seems clearer than putting it in the logic at the moment, but I'm sure before long we will have several different kinds of email with different fields, meaning we would need a refactor later anyway.

handlers/discount-api/test/applyDiscountIntegration.test.ts
```diff
@@ -7,32 +7,87 @@ import { ZuoraClient } from '@modules/zuora/zuoraClient';
...
+	const cancellationResult = await cancelSubscription(
+		zuoraClient,
+		subscriptionNumber,
+		dayjs().add(1, 'month'),
+		true,
+	);
+	expect(cancellationResult.success).toEqual(true);
+}, 30000);
+
+test('digi subs can have a discount but dont get an email', async () => {
```
johnduffell, line 69: added a new test for digi sub so we can be sure they don't get an email

as we add more it tests, we could combine the preview and apply tests to save doing the setup and teardown twice?

handlers/discount-api/test/getDiscountFromSubscription.test.ts
```diff
@@ -1,13 +1,18 @@
...
+import {
+	Discount,
+	getDiscountFromSubscription,
+} from '../src/productToDiscountMapping';
 import json from './fixtures/digital-subscriptions/get-discount-test.json';
 
 test('getDiscountFromSubscription should return an annual discount for an annual sub', () => {
 	const sub = zuoraSubscriptionSchema.parse(json);
-	expect(getDiscountFromSubscription('PROD', sub).discount).toEqual({
+	const expected: Discount = {
```
johnduffell, line 10: without adding this type, the code compiles, but the test fails at runtime.

handlers/discount-api/src/discountEndpoint.ts
```diff
@@ -50,21 +80,49 @@ export const applyDiscountEndpoint = async (
...
+		dayjs(subscription.termEndDate),
+		dayjs(dateToApply),
 		discount.productRatePlanId,
+	);
+
+	if (discounted.success) {
+		console.log('Discount applied successfully');
+	} else {
+		throw new Error('discount was not applied: ' + JSON.stringify(discounted));
+	}
```
tjmw, line 99: Totally subjective, but just to reduce the nesting (and hopefully increase readability) I'd probably put the throw in a condition and then have the console.log after:

```suggestion
	if (!discounted.success) {
	        throw new Error('discount was not applied: ' + JSON.stringify(discounted));
	}
	
	console.log('Discount applied successfully');
```

handlers/discount-api/src/discountEndpoint.ts
```diff
@@ -50,21 +80,49 @@ export const applyDiscountEndpoint = async (
...
+
+	const billingPreviewAfter = await getBillingPreview(
+		zuoraClient,
+		dayjs().add(13, 'months'),
```
tjmw, line 103: What's the significance of the 13 here? Would it be worth pulling out to a const to give it a name?

handlers/discount-api/src/sendCancellationDiscountConfirmationEmail.ts
```diff
@@ -0,0 +1,41 @@
...
+export type EmailFields = {
+	firstDiscountedPaymentDate: dayjs.Dayjs;
+	nextNonDiscountedPaymentDate: dayjs.Dayjs;
+	emailAddress: string;
+	firstName: string;
+	lastName: string;
+	identityId: string;
+};
+
+export const sendCancellationDiscountEmail = ({
```
tjmw, line 14: I don't think this actually _sends_ the email right? Would it be worth removing/replacing the `send` verb in the name here?

handlers/discount-api/src/sendCancellationDiscountConfirmationEmail.ts
```diff
@@ -0,0 +1,41 @@
...
+	firstDiscountedPaymentDate,
+	nextNonDiscountedPaymentDate,
+	emailAddress,
+	firstName,
+	lastName,
+	identityId,
+}: EmailFields) => {
+	const emailMessage: EmailMessageWithUserId = {
```
tjmw, line 22: What's the relationship between `EmailFields` and `EmailMessageWithUserId`. The thing we're returning is an `EmailMessageWithUserId`. As an aside - do we need to assign to a const just to return that const? What do you think about returning the object directly?

handlers/discount-api/test/applyDiscountIntegration.test.ts
tjmw: Is the setup/teardown expensive? If not I'd vote to keep things separate. In general I think it's nicer to test fewer things in a given `test` block, as it reduces the cognitive load when the test fails and you have to figure out which bit failed.

handlers/discount-api/test/getDiscountFromSubscription.test.ts
tjmw: That's very odd! What'd the failure you see at runtime?

handlers/discount-api/test/applyDiscountIntegration.test.ts
johnduffell: yes, it has to create a subscription in sandbox zuora, then cancel it afterwards.  The it-tests take the order of 30 seconds despite there not being many.
In the past we've tried to rely on ready made subscriptions in sandbox env, but that makes the tests flaky as we sometimes clean up zuora (or the subscriptions pass a crucial date)

handlers/discount-api/test/getDiscountFromSubscription.test.ts
johnduffell: oh sorry that wasn't clear - I added a new field to the Discount object, and everything was fine until I actually ran the test.  When I added the type, it had a type error.
I am still struggling to get my head around the whole "infer types is safer than specifying them" in typescript because my experience in scala (and so far in typescript) is the opposite.

handlers/discount-api/src/discountEndpoint.ts
johnduffell: not sure as that was moved code rather than something I added.  I  wonder if in turn it was moved from elsewhere.

I think traditionally we would get 13 months to be sure we cover at least 2 payments even if it's an annual subscription.

In this case 12 would be enough as we only need the next payment, not the next two.

I'll add a comment I think, as I can't think of a useful name
```suggestion
		dayjs().add(13, 'months'), // 13 months gives us minimum 2 payments even on an Annual sub
```

handlers/discount-api/src/sendCancellationDiscountConfirmationEmail.ts
johnduffell: good point, I'll rename it (and the file) to `generateCancellationDiscountConfirmationEmail`

handlers/discount-api/src/sendCancellationDiscountConfirmationEmail.ts
johnduffell: good question - I was just basing it on a copy of this:
https://github.com/guardian/support-service-lambdas/blob/d8cc2659afb482f8bc707f7cd486cf7a6294261d/handlers/update-supporter-plus-amount/src/sendEmail.ts#L10-L19
with some modifications to make it easier/safer to test

good point about returning directly, that was left over from the above modifications, I'l fix

johnduffell: and sorry so EmailFields is the typed parameters needed to send the email, whereas EmailMessageWithUserId is the thing that's ready to go into SQS, so everything is a string and in the right structure for membership-workflow's receiver.

tjmw: Oh sorry I'd misread the function signature, thanks for explaining!

tjmw (approved):
👍

PR #2324, opened 2024-06-13
make the module tests run in the automated build

johnduffell:
I noticed that one of the tests was failing locally but there were no issues with the build.

After investigation it seemed that the build only runs "package" which doesn't exist on the modules so it's skipped.
The tsc and eslint at the root search out all files in both locations, so those checks were not skipped.

This PR updates it so that pnpm test is run at the root, which should exist on all modules as well as the handlers.

There is an efficiency issue where all the builds are doing everything, this would also mean an issue in one lambda would stop them all building, but that can be dealt with if/when it causes an issue.

### DIFF
--- a/.github/workflows/ci-typescript.yml
+++ b/.github/workflows/ci-typescript.yml
@@ -39,7 +39,7 @@ jobs:
           node-version-file: '.nvmrc'
           cache: 'pnpm'
       - run: pnpm install
-      - run: pnpm --filter modules test
+      - run: pushd modules && pnpm test; popd
       - run: pnpm package
 
       - name: Configure AWS credentials
--- a/.github/workflows/ci-typescript.yml
+++ b/.github/workflows/ci-typescript.yml
@@ -39,6 +39,7 @@ jobs:
           node-version-file: '.nvmrc'
           cache: 'pnpm'
       - run: pnpm install
+      - run: pushd modules && pnpm test; popd
       - run: pnpm package
 
       - name: Configure AWS credentials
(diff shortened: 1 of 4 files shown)


modules/zuora/test/billingPreview.test.ts
```diff
@@ -1,10 +1,15 @@
-import { getNextInvoiceItems } from '../src/billingPreview';
```
johnduffell, line 1: this was the line that was broken in the test

modules/jest.config.js
```diff
@@ -0,0 +1,10 @@
+/** @type {import('ts-jest').JestConfigWithTsJest} */
```
johnduffell, line 1: this and the file below are simply to make sure tests in the modules root can run

.github/workflows/ci-typescript.yml
```diff
@@ -39,6 +39,7 @@ jobs:
           node-version-file: '.nvmrc'
           cache: 'pnpm'
       - run: pnpm install
+      - run: pushd modules && pnpm test; popd
```
rupertbates, line 42: does `pnpm --filter modules test` work? If so it's more idiomatic

rupertbates (approved):

.github/workflows/ci-typescript.yml
johnduffell: no
```
Run pnpm --filter modules test
No projects matched the filters in "/home/runner/work/support-service-lambdas/support-service-lambdas"
```

PR #2332, opened 2024-06-18
move identity id check to happen first

johnduffell:
In discount api the identity id check is important to make sure people cant access other people's subscriptions.

I noticed that the logic was a bit mixed together, meaning that it was easy to accidentally reveal things about the subscription.

This already meant that you could find out whether someone else's subscription was active or cancelled.

I have cleaned up the code into a single method which should be run first, and then all code that might reveal information about the subscription can come after that.

Unfortunately because the code is sequential it will not fetch the billing preview and account in parallel any more, I'm sure there's a way of doing it but I wonder whether that would be more confusing (and risky) than just taking the performance hit.

### DIFF
--- a/handlers/discount-api/src/discountEndpoint.ts
+++ b/handlers/discount-api/src/discountEndpoint.ts
@@ -27,10 +27,15 @@ export const previewDiscountEndpoint = async (
 ) => {
 	const zuoraClient = await ZuoraClient.create(stage);
 
+	const { subscription } = await getSubscriptionIfBelongsToIdentityId(
+		headers,
+		zuoraClient,
+		subscriptionNumber,
+	);
+
 	const { discount, dateToApply } = await getDiscountToApply(
 		stage,
-		headers,
-		subscriptionNumber,
+		subscription,
 		zuoraClient,
 	);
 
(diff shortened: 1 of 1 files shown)


handlers/discount-api/src/discountEndpoint.ts
```diff
@@ -27,10 +30,15 @@ export const previewDiscountEndpoint = async (
 ) => {
 	const zuoraClient = await ZuoraClient.create(stage);
 
-	const { discount, dateToApply } = await getDiscountToApply(
-		stage,
+	const { subscription } = await getSubscriptionIfBelongsToIdentityId(
 		headers,
+		zuoraClient,
 		subscriptionNumber,
+	);
```
johnduffell, line 37: `getSubscriptionIfBelongsToIdentityId` is the new function to only do the identity check

handlers/discount-api/src/discountEndpoint.ts
```diff
@@ -27,10 +30,15 @@ export const previewDiscountEndpoint = async (
...
+
+	const { discount, dateToApply } = await getDiscountToApply(
+		stage,
+		subscription,
 		zuoraClient,
 	);
```
johnduffell, line 43: now `getDiscountToApply` only does the business logic part

handlers/discount-api/src/discountEndpoint.ts
```diff
@@ -144,23 +154,37 @@ async function getDiscountToApply(
...
+		subscription.accountNumber,
 	);
-	const [account, billingPreview] = await Promise.all([
-		checkSubscriptionBelongsToIdentityId(zuoraClient, subscription, identityId),
-		getBillingPreview(
-			zuoraClient,
-			dayjs().add(13, 'months'),
-			subscription.accountNumber,
-		),
-	]);
```
johnduffell, line 163: the getAccount and getBillingPreview used to happen in parallel, this has been lost by disentangling the code.
I could pass a function `(accountId) => Promise[A]` in to `getSubscriptionIfBelongsToIdentityId` and return an A as well that's been fetched in parallel but that seems a bit more scala and a bit less typescript!

rupertbates (approved):
👍

PR #2334, opened 2024-06-19
Add a `check-formatting` pnpm script

tjmw:
## What does this change?

I've noticed that sometimes the TS code formatting in this repo drifts from the prettier config so I'd like to start checking formatting in CI, failing the build if there are files which aren't formatting correctly.

This adds a script `check-formatting` which can be run for the whole project:

`$ pnpm check-formatting`

or for a specific workspace:

`$ pnpm --filter alarms-handler check-formatting`

In #2230 I added a script to format the codebase, but on reflection I don't think this was very `pnpm`-ish. I think delegating to individual workspace `package.json`s is the way to go (so you can run for a specific workspace). I'll follow up with a PR to move over to this approach for formatting.

Then two more PRs:

* Fix the current formatting for the project
* Enforce formatting in CI.

## Accessibility

### DIFF
--- a/_templates/new-lambda/api-gateway/package.ejs.t
+++ b/_templates/new-lambda/api-gateway/package.ejs.t
@@ -12,7 +12,8 @@ sh: git add handlers/<%=lambdaName%>/package.json
     "type-check": "tsc --noEmit",
     "build": "esbuild --bundle --platform=node --target=node18 --outfile=target/index.js src/index.ts",
     "lint": "eslint src/**/*.ts",
-    "package": "pnpm type-check && pnpm lint && pnpm test && pnpm build && cd target; zip -qr <%=lambdaName%>.zip ./*.js"
+    "package": "pnpm type-check && pnpm lint && pnpm test && pnpm build && cd target; zip -qr <%=lambdaName%>.zip ./*.js",
+    "check-formatting": "prettier --check **/*.ts"
   },
   "devDependencies": {
     "@types/aws-lambda": "^8.10.129"
--- a/cdk/package.json
+++ b/cdk/package.json
@@ -7,7 +7,7 @@
     "test": "jest",
     "test-update": "jest -u",
     "format": "prettier --write \"{lib,bin}/**/*.ts\"",
-    "check-formatting": "prettier --check \"{lib,bin}/**/*.ts\"",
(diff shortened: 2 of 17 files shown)


rupertbates (approved):
Nice

jamesgorrie (approved):
Great - could we add a `fix-formatting`?

tjmw:
> Great - could we add a `fix-formatting`?

Yep that's coming up! There actually is already a `format` command in the root package.json, but it's not very pnpm-y. I'm going to migrate it so that it can be run per workspace if needed, to mirror `check-formatting`.

jamesgorrie:
> > Great - could we add a `fix-formatting`?
> 
> Yep that's coming up! There actually is already a `format` command in the root package.json, but it's not very pnpm-y. I'm going to migrate it so that it can be run per workspace if needed, to mirror `check-formatting`.

@tjmw - what's the benefit of having it per package vs moving this command to the root given formatting is a repo-wide concern?

jamesgorrie:
> and gives more flexibility

I suppose my question is do we get more benefit from being flexible or standardising. When someone starts a new package, should they not have to think about how they want to implement this and just get it for free?

tjmw:
> > and gives more flexibility
> 
> I suppose my question is do we get more benefit from being flexible or standardising. When someone starts a new package, should they not have to think about how they want to implement this and just get it for free?

👍 Yeah it's a good question. Right now you get a sensible default for free from the `new-lambda` script, so you don't have to think about it, unless you want to do something different for some reason.

johnduffell:
I think we should enforce things everywhere as long as they can also be set up to format compliantly in the IDEs
A more general point - At the moment we have individual GHA builds for each lambda, so ideally we could only check the relevant code for each lambda in its own build (at the moment, all the builds build all the TS code, so if there's an issue anywhere, they all fail, which is not ideal.)

tjmw:
> A more general point - At the moment we have individual GHA builds for each lambda, so ideally we could only check the relevant code for each lambda in its own build (at the moment, all the builds build all the TS code, so if there's an issue anywhere, they all fail, which is not ideal.)

Ha yes I've just realised this today and am experimenting with some alternatives.

tjmw:
@johnduffell I've attempted to improve things here: #2341

PR #2341, opened 2024-06-19
Be more focused in the build

tjmw:
## What does this change?

Before:

* All tests etc for all workspaces run multiple times - once per subproject in the matrix
* When the tests failed for a particular workspace, it wasn't the package step that failed but the "Upload to Riff-Raff" step, which feels a bit accidental (only failed because there's no zip file). See [this run](https://github.com/guardian/support-service-lambdas/actions/runs/9582507082/job/26421631106) as an example.

After:

* Only run the tests etc for the current subproject in the matrix (and the cdk project, which we always need)
* When the tests fail, the `package` step fails, so we fail faster and it's clearer where the issue is

## Accessibility

### DIFF
--- a/.github/workflows/ci-typescript.yml
+++ b/.github/workflows/ci-typescript.yml
@@ -41,7 +41,7 @@ jobs:
           cache: 'pnpm'
       - run: pnpm install
       - run: pushd modules && pnpm test; popd
-      - run: pnpm package
+      - run: pnpm --filter ${{ matrix.subproject }} package
 
       - name: Configure AWS credentials
         uses: aws-actions/configure-aws-credentials@v4
--- a/.github/workflows/ci-typescript.yml
+++ b/.github/workflows/ci-typescript.yml
@@ -41,6 +41,7 @@ jobs:
           cache: 'pnpm'
       - run: pnpm install
       - run: pushd modules && pnpm test; popd
+      - run: pnpm --filter cdk package
       - run: pnpm --filter ${{ matrix.subproject }} package
 
       - name: Configure AWS credentials
--- a/_templates/new-lambda/api-gateway/package.ejs.t
+++ b/_templates/new-lambda/api-gateway/package.ejs.t
@@ -12,7 +12,7 @@ sh: git add handlers/<%=lambdaName%>/package.json
     "type-check": "tsc --noEmit",
     "build": "esbuild --bundle --platform=node --target=node18 --outfile=target/index.js src/index.ts",
     "lint": "eslint src/**/*.ts",
-    "package": "pnpm type-check && pnpm lint && pnpm test && pnpm build && cd target; zip -qr <%=lambdaName%>.zip ./*.js",
+    "package": "pnpm type-check && pnpm lint && pnpm test && pnpm build && cd target && zip -qr <%=lambdaName%>.zip ./*.js",
     "check-formatting": "prettier --check **/*.ts",
     "fix-formatting": "prettier --write **/*.ts"
   },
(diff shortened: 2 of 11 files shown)


_templates/new-lambda/api-gateway/package.ejs.t
```diff
@@ -12,7 +12,7 @@ sh: git add handlers/<%=lambdaName%>/package.json
     "type-check": "tsc --noEmit",
     "build": "esbuild --bundle --platform=node --target=node18 --outfile=target/index.js src/index.ts",
     "lint": "eslint src/**/*.ts",
-    "package": "pnpm type-check && pnpm lint && pnpm test && pnpm build && cd target; zip -qr <%=lambdaName%>.zip ./*.js",
+    "package": "pnpm type-check && pnpm lint && pnpm test && pnpm build && cd target && zip -qr <%=lambdaName%>.zip ./*.js",
```
tjmw, line 15: Previously if the `pnpm test` command failed, the overall exit code for `package` was 0. I guess this must be becuase the `zip` command was somehow succeeding. This resulted in the `package` step in the GH action succeeding which is a bit misleading. Then we'd fail later on in the upload to Riff Raff step.

.github/workflows/ci-typescript.yml
```diff
@@ -40,7 +40,8 @@ jobs:
           cache: 'pnpm'
       - run: pnpm install
       - run: pushd modules && pnpm test; popd
```
johnduffell, line 42: > Only run the tests etc for the current subproject in the matrix (and the cdk project, which we always need)

we will also run all the tests for all the modules

johnduffell (commented):
can you check that the type-check/lint will check the modules still, as before it was relying ont he top level `package` command in the package.json to search out all the ts files bc eslint or tsc.
We might have to add an extra step to do this similar to the way we run the tests.

modules/product-catalog/src/api.ts
```diff
@@ -14,6 +14,7 @@ export const getProductCatalogFromApi = async (
 	});
 
 	const json = await response.json();
+	return 1;
```
rupertbates, line 17: Was this an accidental commit?

tjmw: Ha yes, I wanted to understand how TS errors in `modules` were being surfaced (see John's comment above). I'll revert it now...

.github/workflows/ci-typescript.yml
tjmw: 👍  Yeah - I think that's okay. Maybe as a future alternative we could run the module tests in a separate step and fail early if they're broken.

johnduffell (approved):
turns out linting is (already) not running on the modules, so we're not losing anything, thanks for making it better Tom 👍

tjmw:
> turns out linting is (already) not running on the modules, so we're not losing anything, thanks for making it better Tom 👍

Thanks John. Yep linting of modules doesn't currently happen in CI (either in main or this branch) - one for a future PR I think. Type checking of modules does happen, via the consuming handler. This means that if a module is used by two handlers, both will fail the TS check - I think that's okay. It also means if a module is defined but not used by anything it won't be type checked in CI.

rupertbates (approved):
This is great work, thanks for picking it up 👏👏👏

.github/workflows/ci-typescript.yml
tjmw: @johnduffell I've changed this in #2359 where I've introduced a separate job for the common stuff (module tests, module formatting)

PR #2343, opened 2024-06-20
validate supporter plus with additional checks for cancellation offer

johnduffell:
So far I have added support for Supporter Plus monthly to the discount api.

This PR adds additional validation checks as required, so we don't offer the discount where people are not eligible.


handlers/discount-api/src/discountEndpoint.ts
```diff
@@ -24,10 +27,11 @@ export const previewDiscountEndpoint = async (
 	stage: Stage,
 	headers: APIGatewayProxyEventHeaders,
 	subscriptionNumber: string,
+	today: dayjs.Dayjs,
```
johnduffell, line 30: now we pass in "today" so that in the integration tests, we can validate "subscription must be at least 2 months old" by pretending we are in the future.

handlers/discount-api/src/discountEndpoint.ts
```diff
@@ -24,10 +27,11 @@ export const previewDiscountEndpoint = async (
...
 ) => {
 	const zuoraClient = await ZuoraClient.create(stage);
 
-	const { subscription } = await getSubscriptionIfBelongsToIdentityId(
+	const { subscription, account } = await getSubscriptionIfBelongsToIdentityId(
```
johnduffell, line 34: we need the account to check that there's no outstanding balance (which means they are in payment failure although the check is not 100% if they are on DD)

handlers/discount-api/src/discountEndpoint.ts
```diff
@@ -169,33 +178,50 @@ async function getSubscriptionIfBelongsToIdentityId(
...
+			break;
+		case 'AtCatalogPrice':
+			eligibilityChecker.assertNextPaymentIsAtCatalogPrice(
+				await catalog(),
+				billingPreview,
+				discountableProductRatePlanId,
+				account.metrics.currency,
+			);
+			break;
+	}
```
johnduffell, line 219: this is the new product-specific validation

handlers/discount-api/src/discountEndpoint.ts
```diff
@@ -169,33 +178,51 @@ async function getSubscriptionIfBelongsToIdentityId(
...
+	const catalog = () => getZuoraCatalog(stage);
+	const eligibilityChecker = new EligibilityChecker(
+		subscription.subscriptionNumber,
+	);
 
-	if (subscription.status !== 'Active') {
-		throw new ValidationError(
-			`Subscription ${subscription.subscriptionNumber} has status ${subscription.status}`,
-		);
-	}
```
johnduffell, line 181: moved into eligibilityChecker.getNextBillingDateIfEligible in

handlers/discount-api/src/eligibilityChecker.ts
```diff
@@ -1,60 +1,108 @@
...
+			subscription.contractEffectiveDate.toDateString(),
+		);
+		eligibilityChecker.assertValidState(
+			subscription.ratePlans.every(
+				(rp) => rp.productRatePlanId !== discountProductRatePlanId,
+			),
+			validationRequirements.notAlreadyUsed,
+			discountProductRatePlanId,
+		);
+	};
```
johnduffell, line 84: this is the new supporter plus specific validation - at least 2 months old, and hasn't already used this discount

handlers/discount-api/src/eligibilityChecker.ts
```diff
@@ -1,60 +1,108 @@
...
+			`No next invoice found for account containing ${this.subscriptionNumber}`,
 		);
 
-		if (nextInvoiceTotal.total < totalPrice) {
+		this.assertValidState(
+			nextInvoiceTotal >= catalogPrice,
+			validationRequirements.atLeastCatalogPrice + ' of ' + catalogPrice,
+			nextInvoiceTotal + ' ' + currency,
+		);
+	};
```
johnduffell, line 62: this is the existing validation that only appleis to Digi sub - has to be at the latest catalog price

handlers/discount-api/src/eligibilityChecker.ts
```diff
@@ -1,60 +1,108 @@
...
+					'>' +
+					` (was ${actual})`,
 			);
 		}
-		return getIfDefined(
-			nextInvoiceTotal.date,
-			'No next invoice date found in next invoice items',
-		);
 	};
 }
```
johnduffell, line 99: this is a simple new function just to make sure that all validation failures appear in the same format

handlers/discount-api/src/eligibilityChecker.ts
```diff
@@ -1,60 +1,108 @@
...
+
+export const validationRequirements = {
+	twoMonthsMin: 'subscription was taken out more than 2 months ago',
+	notAlreadyUsed: 'this discount has not already been used',
+	noNegativePreviewItems: `next invoice has no negative items`,
+	isActive: 'subscription status is active',
+	zeroAccountBalance: 'account balance is zero',
+	atLeastCatalogPrice: 'next invoice must be at least the catalog price',
+};
```
johnduffell, line 108: this keeps all possible validation errors together in the code and lets the tests assert them by label rather than hard coded

handlers/discount-api/test/applyDiscountIntegration.test.ts
```diff
@@ -30,6 +30,7 @@ test('Supporter Plus subscriptions can have a discount and get an email', async
 		stage,
 		{ 'x-identity-id': validIdentityId },
 		subscriptionNumber,
+		paymentDate.add(2, 'months').add(1, 'day'),
```
johnduffell, line 33: we can only apply the discount if we pretend it's more than 2 months later than we did the subscribe call on zuora.

handlers/discount-api/test/eligibilityChecker.test.ts
```diff
@@ -2,80 +2,133 @@ import {
...
 
-	const actual = () => {
-		eligibilityCheckerCode.getNextBillingDateIfEligible(
-			billingPreview,
-			getDiscountableRatePlan(sub),
+	const ac2 = () =>
+		eligibilityChecker.assertEligibleForFreePeriod(
+			discount.discount.productRatePlanId,
+			sub,
+			dayjs('2024-08-04'),
```
johnduffell, line 89: not quite long enough to have had 2 months

handlers/discount-api/test/helpers.ts
```diff
@@ -70,7 +70,7 @@ async function subscribe(
 				}[];
 				Subscription: {
 					ContractEffectiveDate: string;
-					ContractAcceptanceDate: string;
+					ContractAcceptanceDate: string; // actually CustomerAcceptanceDate but zuora API has a typo
```
johnduffell, line 73: how bizarre that even zuora have got their own field name wrong!

modules/zuora/src/zuoraSchemas.ts
```diff
@@ -81,6 +81,10 @@ export const zuoraAccountBasicInfoSchema = z
 		identityId: obj.IdentityId__c,
 	}));
 
+export const metricsSchema = z.object({
+	totalInvoiceBalance: z.number(),
+	currency: z.string(),
```
johnduffell, line 86: we used to get currency from the individual invoice items, but it's an account level setting (and it makes the code slightly simpler to use it from here)

handlers/discount-api/src/productToDiscountMapping.ts
```diff
@@ -19,25 +19,28 @@ export const getDiscountFromSubscription = (
...
 	subscription: ZuoraSubscription,
 ) => {
-	const nonDiscountRatePlan = getDiscountableRatePlan(subscription);
+	const discountableProductRatePlanId =
+		getDiscountableRatePlan(subscription).productRatePlanId;
 	const discount =
-		ProductToDiscountMapping[stage][nonDiscountRatePlan.productRatePlanId];
+		ProductToDiscountMapping[stage][discountableProductRatePlanId];
 
 	if (discount == undefined) {
```
rBangay, line 27: If you want strict equality here maybe you should use: 

`discount === undefined`

If a falsey value is good enough maybe you could go with:

`!discount`

handlers/discount-api/test/eligibilityChecker.test.ts
rBangay, line 89: hmm, I 'm not sure if this is an issue or not ... I was wondering if it would be worth calculating this date (2 months onwards) from the start date declared in [full-price.json fixture](https://github.com/guardian/support-service-lambdas/tree/3e0b6b8751e8066b80ce362e5f361af5c7381397/handlers/discount-api/test/fixtures)/[supporter-plus](https://github.com/guardian/support-service-lambdas/tree/3e0b6b8751e8066b80ce362e5f361af5c7381397/handlers/discount-api/test/fixtures/supporter-plus) that way if those dates ever change this test still remain relevant.

Whatcha reckon?

rBangay (approved):
LGTM 👍

handlers/discount-api/src/productToDiscountMapping.ts
johnduffell: will go for the `===` as it could only be undefined

handlers/discount-api/test/eligibilityChecker.test.ts
johnduffell: yeah it's a good point, how much should I replicate the logic in the tests and how much I should hard code things!
I will have a play and see what it looks like

johnduffell: it came out alright so I pushed it, it does mean we're no longer testing the add(2 months) functionality but I think that should be pretty reliable
see the last commit https://github.com/guardian/support-service-lambdas/pull/2343/commits/243a32c209c9d49b6044fd3343af94793bb98703

johnduffell:
tested in CODE and looks good. Although it's hard to test thoroughly as the sub has to be 2 months old in order to get through the eligibility check!
I also noticed the digi sub flow claims to send a confirmation email but I've no idea how that would work and I didn't actually get one
image

PR #2344, opened 2024-06-20
zuora<-> salesforce link remover stack and fetch billing accounts lambda outline (logic to follow)

david-pepper:
## What does this change?
Kick off the project by creating the zuora-salesforce-link-remover stack and the getBillingAccounts lambda that will retrieve billing accounts from Salesforce. The lambda does not contain any logic yet. That will follow in a subsequent PR.

Screenshot 2024-06-24 at 16 49 02

## Notes
A couple of files unrelated to my changes have been formatted by prettier:

- alarms-handler/src/index.ts
- product-switch-api/src/supporterProductData.ts

The changes are minor so I'll keep them in the PR instead of reverting them.


package.json
```diff
@@ -9,8 +9,9 @@
 		"it-test": "pnpm --stream -r run it-test",
 		"new-lambda": "pnpm hygen new-lambda api-gateway && pnpm install && cd ./cdk && yarn lint --fix && yarn test -u && git add ./lib/__snapshots__/*.test.ts.snap",
 		"prepare": "husky",
-		"format": "prettier --write handlers/**/*.ts modules/**/*.ts",
-		"check-formatting": "pnpm -r run check-formatting"
+		"format": "prettier --write handlers/**/*.ts cdk/**/*.ts modules/**/*.ts",
+		"check-formatting": "pnpm -r run check-formatting",
+		"test-update": "jest -u"
```
andrewHEguardian, line 14: what is this for?

cdk/lib/zuora-salesforce-link-remover.ts
```diff
@@ -0,0 +1,25 @@
...
+			runtime: Runtime.NODEJS_20_X,
+			handler: 'getBillingAccounts.handler',
+			fileName: `${appName}.zip`,
+		});
+	}
+}
+
+// make a test file
+// make a snapshot file jest --u
+// run pnpm
```
charleycampbell, line 25: Really like having comments as to how to test! are there any supporting docs on how and where to create the test file and snapshot file?

package.json
david-pepper: I was trying to figure out how to run a script and generate a new snapshot. I've removed this now

handlers/zuora-salesforce-link-remover/src/getBillingAccounts.ts
```diff
@@ -0,0 +1,5 @@
+export function handler() {
+	console.log('hello david and andrew');
+
+	return 'abcdef';
```
charleycampbell, line 4: Is it ok to leave this as is or should we update the console log and return to something a bit more helpful?

charleycampbell: this might also be irrelevant as I haven't had much experience with lambda building!

cdk/lib/zuora-salesforce-link-remover.ts
david-pepper: a whoops, these were notes on what I needed to do. Removed them now.

I think the location of the test file is up to the developer, but there is a clear precedent in support service lambdas where cdk test files go alongside the files they are testing

Screenshot 2024-06-24 at 17 13 36


and the handler test files go in a separate test folder adjacent to the src folder for each handler

Screenshot 2024-06-24 at 17 15 02

Snapshot file gets created by running jest --u command, and the snapshot needed to get updated when new infratructure gets added to the cdk file, so rerunning this command updates the snapshot

handlers/zuora-salesforce-link-remover/src/getBillingAccounts.ts
david-pepper: The purpose of the lambda at the moment is just to output something and return something. It doesn't matter what - we just want the infrastructure in place so that when we add our logic next, we get a nice concise PR related only to that logic, and not to the lambda infrastructure in this PR.

cdk/lib/zuora-salesforce-link-remover.ts
```diff
@@ -0,0 +1,21 @@
...
+		super(scope, id, props);
+
+		const appName = 'zuora-salesforce-link-remover';
+
+		new GuLambdaFunction(this, 'get-billing-accounts-lambda', {
+			app: appName,
+			functionName: `${appName}-get-billing-accounts-${this.stage}`,
+			runtime: Runtime.NODEJS_20_X,
+			handler: 'getBillingAccounts.handler',
+			fileName: `${appName}.zip`,
```
andrewHEguardian, line 18: a nice tidbit from the [example TS lambda](https://github.com/guardian/example-typescript-lambda/blob/f683d601ec056ce9574a50396d903800f7020be3/packages/cdk/lib/example-typescript-lambda.ts#L46) is that apparently specifying Arm64 is more performant and cheaper!

andrewHEguardian (approved):
Looks good to me! Though you probably want a review from someone who is more familiar with this repo

PR #2346, opened 2024-06-25
zuora<-> salesforce link remover get secrets from secrets manager

david-pepper:
## What does this change?
Get credentials necessary to connect to Salesforce from Secrets Manager:

- Connected App client Id and client secret
- Salesforce Api Username, password and token

Screenshot 2024-06-25 at 12 23 49

## How to test
To verify the secrets have been retrieved successfully, output some of the values stored in the secret to the console.

For connected app, authUrl and secret name can be safely output. No need to output clientId and clientSecret.
For Salesforce user, output username, not password or token.

Screenshot 2024-06-25 at 12 19 53

### DIFF
--- a/cdk/lib/zuora-salesforce-link-remover.ts
+++ b/cdk/lib/zuora-salesforce-link-remover.ts
@@ -23,6 +23,7 @@ export class ZuoraSalesforceLinkRemover extends GuStack {
 					actions: ['secretsmanager:GetSecretValue'],
 					resources: [
 						`arn:aws:secretsmanager:${this.region}:${this.account}:secret:DEV/Salesforce/ConnectedApp/AwsConnectorSandbox-oO8Phf`,
+						`arn:aws:secretsmanager:${this.region}:${this.account}:secret:DEV/Salesforce/User/MembersDataAPI-yaO9GU`,
 					],
 				}),
 			],
--- a/cdk/lib/zuora-salesforce-link-remover.ts
+++ b/cdk/lib/zuora-salesforce-link-remover.ts
@@ -20,9 +20,7 @@ export class ZuoraSalesforceLinkRemover extends GuStack {
 			architecture: Architecture.ARM_64,
 			initialPolicy: [
 				new PolicyStatement({
-					actions: [
-						'secretsmanager:GetSecretValue',
-					],
+					actions: ['secretsmanager:GetSecretValue'],
 					resources: [
 						`arn:aws:secretsmanager:${this.region}:${this.account}:secret:DEV/Salesforce/ConnectedApp/AwsConnectorSandbox-oO8Phf`,
 					],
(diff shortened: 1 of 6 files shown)


handlers/zuora-salesforce-link-remover/src/secrets.ts
```diff
@@ -0,0 +1,38 @@
+import {
+	GetSecretValueCommand,
+	SecretsManagerClient,
+} from '@aws-sdk/client-secrets-manager';
+
+export async function getSecretValue<T>(secretName: string): Promise<T> {
```
tomrf1, line 6: Not that important right now, but much of this file is generic and useful beyond this lambda. I can see [two other lambdas](https://github.com/search?q=repo%3Aguardian%2Fsupport-service-lambdas+SecretsManagerClient+language%3ATypeScript&type=code&l=TypeScript) in this repo that implement something similar. Perhaps we should consider refactoring these into a shared file

tomrf1 (approved):
looks good

cdk/lib/zuora-salesforce-link-remover.ts
```diff
@@ -17,6 +18,15 @@ export class ZuoraSalesforceLinkRemover extends GuStack {
 			handler: 'getBillingAccounts.handler',
 			fileName: `${appName}.zip`,
 			architecture: Architecture.ARM_64,
+			initialPolicy: [
+				new PolicyStatement({
+					actions: ['secretsmanager:GetSecretValue'],
+					resources: [
+						`arn:aws:secretsmanager:${this.region}:${this.account}:secret:DEV/Salesforce/ConnectedApp/AwsConnectorSandbox-oO8Phf`,
+						`arn:aws:secretsmanager:${this.region}:${this.account}:secret:DEV/Salesforce/User/MembersDataAPI-yaO9GU`,
```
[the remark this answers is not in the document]
david-pepper: Our approach to security for integrations with SF is as follows:

PROD
We use a combination of credentials associated with a connected app (clientId and clientSecret) and API User (username, password and token) for each integration. So for an integration that is intended to remove billing accounts, we would have a dedicated billing account remover connected app and a dedicated billing account remover api user. This is a considered decision that has been taken to reduce our exposure should any credentials become compromised. The trade-off here is that it means a boatload of connected apps and users, so that means cost and time/effort for maintenance. We are happy to assume the cost for the security. When it comes to the respective connected apps and users for the sandboxes, we decided that we would only have one connected app and one api user to save on maintenance of these.

Now here we are deviating from the one dedicated user by selecting a pre-existing user. This was likely done for simplicity and never updated. Again, confusingly, I've actually currently updated it to a dedicatd user (billing account remover api user) for the time being, but soon I will delete this user, and update the running user in a sandbox to a generic user for integrations (I tried creating one today but ran into some permissions issues to reverted to billing account remover api user).

Hope that all makes sense. Any questions, do shout. Cheers

handlers/zuora-salesforce-link-remover/src/secrets.ts
david-pepper: I've raised a PR for this: #2362

PR #2347, opened 2024-06-25
Zuora salesforce link remover get billing accounts from salesforce

david-pepper:

### DIFF
--- a/handlers/zuora-salesforce-link-remover/src/secrets.ts
+++ b/handlers/zuora-salesforce-link-remover/src/secrets.ts
@@ -3,7 +3,7 @@ import {
 	SecretsManagerClient,
 } from '@aws-sdk/client-secrets-manager';
 
-export async function getSecretValue<T>(secretName: string): Promise<T>{
+export async function getSecretValue<T>(secretName: string): Promise<T> {
 	try {
 		const secretsManagerClient = new SecretsManagerClient({
 			region: process.env.region,
@@ -14,7 +14,7 @@ export async function getSecretValue<T>(secretName: string): Promise<T>{
 		});
 
 		const response = await secretsManagerClient.send(command);
-		
+
 		if (!response.SecretString) {
 			throw new Error(`No secret found with name:$secretName`);
 		}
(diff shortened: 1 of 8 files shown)


handlers/zuora-salesforce-link-remover/src/http.ts
```diff
@@ -0,0 +1,109 @@
...
+				'Content-Type': 'application/json',
+			},
+		},
+	);
+
+	if (!response.ok) {
+		throw new Error(`Failed to execute query: ${response.statusText}`);
+	}
+
+	return (await response.json()) as SalesforceQueryResponse;
```
andrewHEguardian, line 87: we should parse this response

handlers/zuora-salesforce-link-remover/src/secrets.ts
```diff
@@ -3,6 +3,30 @@ import {
 	SecretsManagerClient,
 } from '@aws-sdk/client-secrets-manager';
 
+export function getSecretNameDependingOnEnvironment(
+	stage: string,
```
andrewHEguardian, line 7: we could restrict stage to `CODE` | `PROD` and do away with the default switch case

andrewHEguardian (approved):
👍

handlers/zuora-salesforce-link-remover/src/http.ts
david-pepper: Happy to do so, but what's the thinking here?

andrewHEguardian: you're not actually using Zod for any purpose here, if you `safeParse` the response then we can be guaranteed the type will be `SalesforceQueryResponse` instead of casting an `any` to that type

david-pepper: Thanks, I've had a bash at this. Please can you have a look and let me know your thoughts?

handlers/zuora-salesforce-link-remover/src/secrets.ts
david-pepper: awesome, thanks for the suggestion

handlers/zuora-salesforce-link-remover/src/http.ts
andrewHEguardian: looks good to me!

PR #2333, opened 2024-06-19
Add T3 as a HolidayCreditProduct

rupertbates:
## Accessibility

-   [ ] [Tested with screen reader](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#screen-reader)
-   [ ] [Navigable with keyboard](https://github.com/guardian/accessibility/blob/main/people-and-technology/02-physical.md#keyboard)
-   [ ] [Colour contrast passed](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#contrast)
-   [ ] [The change doesn't use only colour to convey meaning](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#use-of-colour)

PR #2348, opened 2024-06-25
Revert "Make holiday stops work with new T3 product"

rupertbates:
Reverts guardian/support-service-lambdas#2333


ecguardian (approved):
👍

PR #2350, opened 2024-06-26
Revert "Revert "Make holiday stops work with new T3 product""

rupertbates:
Reverts guardian/support-service-lambdas#2348 which originally reverted guardian/support-service-lambdas/pull/2333

There was a bug with the original PR which  introduced the new Tier Three as a new supported product for holiday stops. 

The intention was that T3 should be treated as a Guardian Weekly  from the point of view of credits (see code below):

```scala
 def creditProduct(stage: Stage)(plan: RatePlan): Option[CreditProduct] = (stage, plan.productName) match {
      case (Stage.Prod, s"Guardian Weekly$_") => Some(HolidayCreditProduct.Prod.GuardianWeekly)
      // We need to match 'Tier Three' here because that is what the product is called in Zuora, 
      // however the delivery product we are doing a credit for is Guardian Weekly
      case (Stage.Prod, "Tier Three") => Some(HolidayCreditProduct.Prod.GuardianWeekly)
```

However each product type loads a fulfilment dates file from S3 based on the product name, this is incorrect for T3 which should use the Guardian Weekly fulfilment dates. This PR fixes that issue.


lib/fulfilment-dates/src/main/scala/com/gu/fulfilmentdates/FulfilmentDatesLocation.scala
```diff
@@ -2,19 +2,25 @@ package com.gu.fulfilmentdates
...
     val formattedDate = date.format(DateTimeFormatter.ofPattern("yyyy-MM-dd"))
-    val path = s"${zuoraProductType.name}/${formattedDate}_${zuoraProductType.name}.json"
+    val productTypeKey = getProductTypeKey(zuoraProductType)
+    val path = s"$productTypeKey/${formattedDate}_$productTypeKey.json"
     val bucket = fulfilmentDatesBucket(stage)
     S3Location(bucket, path)
   }
 
+  def getProductTypeKey(zuoraProductType: ZuoraProductType) = zuoraProductType match {
+    case ZuoraProductTypes.TierThree => "Guardian Weekly" // Tier Three is a bundle product containing Guardian Weekly
```
rupertbates, line 20: This is the only new change over and above the previous PR

johnduffell (approved):
better luck this time! 👍

lib/fulfilment-dates/src/main/scala/com/gu/fulfilmentdates/FulfilmentDatesLocation.scala
johnduffell: maybe not one for today but perhaps it should be using the product type on the charge instead of on the product?  same as the actual fulfilment has been changed to use

PR #2351, opened 2024-06-27
Tie off TierThree work and add extra properties to package.json

jamesgorrie:
- adds `name` and `version` to `package.json` to allow it being installed from other repos
- updates the `typeObject` as this hadn't been done when adding `TierThree`
- updates the Zuora catalog fixtures to include `TierThree` data

This would allow other repos to `import` elements from this repo [like this](https://github.com/guardian/support-frontend/pull/6135/files#diff-8cac668d346c16ad0a71fdc4fce40bc95c322e441e2edec8bc692980da8af2c0R93).

There are alternative, but all have significant overhead for what I think is a co-location problem i.e. the product catalog API and consumers should exist in the same repo.

Some alternatives:
- **publish as an npm module**: I can't see the benefit of this over what is implemented here, but with less code
- **publish API docs and corresponding types**: This would be using something like [the OpenAPI spec](https://swagger.io/specification/) or using tools like [typespec](https://typespec.io/). While I think this might make the API universally more useful, there is a lot of work to get their given that this is currently a JSON file published to S3.

### DIFF
--- a/modules/product-catalog/src/productCatalogSchema.ts
+++ b/modules/product-catalog/src/productCatalogSchema.ts
@@ -329,7 +329,8 @@ export const productCatalogSchema = z.object({
 					GBP: z.number(),
 				}),
 				charges: z.object({
-					Subscription: z.object({ id: z.string() }),
+					SupporterPlus: z.object({ id: z.string() }),
+					GuardianWeekly: z.object({ id: z.string() }),
 				}),
 				billingPeriod: z
 					.enum(typeObject.SupporterPlus.billingPeriods)
--- a/modules/product-catalog/src/productCatalogSchema.ts
+++ b/modules/product-catalog/src/productCatalogSchema.ts
@@ -320,6 +320,73 @@ export const productCatalogSchema = z.object({
 			}),
 		}),
 	}),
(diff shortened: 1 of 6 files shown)


modules/product-catalog/src/productCatalogSchema.ts
```diff
@@ -320,6 +320,74 @@ export const productCatalogSchema = z.object({
 			}),
 		}),
 	}),
+	TierThree: z.object({
```
andrewHEguardian, line 323: is there a reason this is so repetitive? Eg the billing period is the same schema for each of these

andrewHEguardian (approved):
👍

modules/product-catalog/src/productCatalogSchema.ts
jamesgorrie: It's a relatively simple and easy to read model, so abstracting it would mean having to lookup that abstraction to understand what it is. 

In this instance it's clear from the start what it is.

jamesgorrie: There's a lot of repetition through the tests etc, which I think could use some work in terms of devX though.

PR #2353, opened 2024-06-28
Add T3 into the DeliveryCreditProcessor

rupertbates:
## What does this change?
Similar to the holiday stop processor which was updated here: https://github.com/guardian/support-service-lambdas/pull/2350

the delivery credit processor also needs updating to take account of the new Tier Three product.

### DIFF
--- a/handlers/delivery-problem-credit-processor/src/main/scala/com/gu/deliveryproblemcreditprocessor/DeliveryCreditProcessor.scala
+++ b/handlers/delivery-problem-credit-processor/src/main/scala/com/gu/deliveryproblemcreditprocessor/DeliveryCreditProcessor.scala
@@ -12,7 +12,7 @@ import com.gu.salesforce.{RecordsWrapperCaseClass, SFAuthConfig}
 import com.gu.util.Logging
 import com.gu.util.config.ConfigReads.ConfigFailure
 import com.gu.util.config.{ConfigLocation, LoadConfigModule, Stage}
-import com.gu.zuora.ZuoraProductTypes.{GuardianWeekly, NewspaperHomeDelivery, NewspaperNationalDelivery, ZuoraProductType}
+import com.gu.zuora.ZuoraProductTypes.{GuardianWeekly, NewspaperHomeDelivery, NewspaperNationalDelivery, TierThree, ZuoraProductType}
 import com.gu.zuora.subscription._
 import com.gu.zuora.{AccessToken, HolidayStopProcessorZuoraConfig, Zuora}
 import io.circe.generic.auto._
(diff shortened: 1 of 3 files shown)


handlers/delivery-problem-credit-processor/src/main/scala/com/gu/deliveryproblemcreditprocessor/DeliveryCreditProcessor.scala
```diff
@@ -64,7 +64,7 @@ object DeliveryCreditProcessor extends Logging {
       }
 
   val processAllProducts: RIO[Clock, List[DeliveryCreditResult]] = {
-    val productTypes = List(NewspaperHomeDelivery, GuardianWeekly, NewspaperNationalDelivery)
+    val productTypes = List(NewspaperHomeDelivery, GuardianWeekly, NewspaperNationalDelivery, TierThree)
```
rupertbates, line 67: Add Tier Three into the list of products to process

handlers/delivery-problem-credit-processor/src/main/scala/com/gu/deliveryproblemcreditprocessor/DeliveryCreditProduct.scala
```diff
@@ -54,9 +54,11 @@ object DeliveryCreditProduct {
 
     def creditProduct(stage: Stage)(plan: RatePlan): Option[CreditProduct] = (stage, plan.productName) match {
       case (Stage.Prod, s"Guardian Weekly$_") => Some(DeliveryCreditProduct.Prod.GuardianWeekly)
+      case (Stage.Prod, s"Tier Three") => Some(DeliveryCreditProduct.Prod.GuardianWeekly)
```
rupertbates, line 57: Tier three should be treated as a Guardian Weekly for the purpose of delivery credits

handlers/delivery-problem-credit-processor/src/main/scala/com/gu/deliveryproblemcreditprocessor/Handler.scala
```diff
@@ -1,19 +1,35 @@
...
 import io.github.mkotsur.aws.handler.Lambda._
+import java.lang.{System => JavaSystem}
+
+import java.io.{OutputStream, PrintStream}
 
 object Handler extends Lambda[None.type, List[DeliveryCreditResult]] {
 
   private val runtime = zio.Runtime.default
 
+  val printStream = new PrintStream(new OutputStream() {
```
rupertbates, line 15: Same issue with ZIO logging, probably should break this out into a separate lib in support-service-lambdas but not in this PR

johnduffell: yeah.. we originally decided it was ok as a hack because product-move-api is on the way out, now it's getting everywhere!
I think the issue is that it still splits things up, just not as badly as on every newline.
I think the proper way was to add it as a logging handler (so it gets the actual log lines) rather than system out which just gives it as arbitrary chunks of bytes?

johnduffell (approved):
yep looks pretty uncontroversial, assume it's tested out ok as that would be the hardest part!

PR #2357, opened 2024-07-01
Zuora <-> salesforce link remover - update sf billing accounts

david-pepper:
## What does this change?
Create a new lambda that updates billing account in Salesforce

Screenshot 2024-07-04 at 12 20 44

- updates billing accounts via the composite api that allows for multiple records to be updated in one callout. The value for GDPR_Update_Attempt__c on billing account record in Salesforce will be incremented by 1. An alerting system already exists within Salesforce that fires upon 5 failed update attempts.
- lambda will take in a list of billing account objects generated by the state machine (yet to be built, but coming soon), so for now I have mocked this input.

## How to test
- Execute the lambda and verify the records in Salesforce are updated as expected


handlers/zuora-salesforce-link-remover/src/salesforceHttp.ts
```diff
@@ -144,3 +144,102 @@ const SalesforceQueryResponseSchema = z.object({
...
+		);
+	}
+
+	const sfUpdateResponse = (await response.json()) as SalesforceUpdateResponse;
+	const parseResponse =
+		SalesforceUpdateResponseArraySchema.safeParse(sfUpdateResponse);
+
+	if (!parseResponse.success) {
+		const parseError = `Error parsing response from Salesforce: ${JSON.stringify(parseResponse.error.format())}`;
+		console.error(parseError);
```
tomrf1, line 195: do we need to log errors here?

handlers/zuora-salesforce-link-remover/src/secrets.ts
```diff
@@ -17,6 +18,11 @@ export function getSalesforceSecretNames(stage: 'CODE' | 'PROD'): SecretNames {
 				connectedAppSecretName:
 					'PROD/Salesforce/ConnectedApp/AwsConnectorSandbox',
 			};
+		default:
```
tomrf1, line 21: when is this used?

david-pepper: ```import { stageFromEnvironment } from '@modules/stage';```

The Stage type, imported from modules, has a 3rd possibility - 'CSBX' but that is not in play for this solution

tomrf1 (approved):
👍

handlers/zuora-salesforce-link-remover/src/salesforceHttp.ts
david-pepper: Nope. I've removed these console logs

PR #2362, opened 2024-07-04
Add secrets shared module

david-pepper:
## What does this change?
add a common secrets-manager functionality that can be used across projects. This functionality has been lifted and shifted from the zuora-salesforce-link-remover solution, and was originally introduced in #2346


handlers/zuora-salesforce-link-remover/src/getBillingAccounts.ts
```diff
@@ -1,13 +1,19 @@
+import { getSecretValue } from '@modules/secrets-manager/src/getSecret';
```
david-pepper, line 1: I'm not sure why the ```src``` folder is explicitly referenced here but not in other imports, such as

```
import type { ZuoraSuccessResponse } from '@modules/zuora/zuoraSchemas';
```

modules/secrets-manager/package-lock.json
```diff
@@ -0,0 +1,1196 @@
+{
```
tomrf1, line 1: the other modules don't have a package-lock, is this needed?

tomrf1 (approved):
👍 great

PR #2368, opened 2024-07-10
Handle discounts in refund lambda

rupertbates:
## What does this change?
The Supporter Plus cancellation refund lambda in the product-move-api doesn't handle discounts correctly because it was written before discounts were supported on S+. 

This PR updates it so that it works correctly, see PR comments for further explanation.

### DIFF
--- a/handlers/product-move-api/src/main/scala/com/gu/productmove/refund/RefundSupporterPlus.scala
+++ b/handlers/product-move-api/src/main/scala/com/gu/productmove/refund/RefundSupporterPlus.scala
@@ -1,33 +1,16 @@
 package com.gu.productmove.refund
 
-import com.amazonaws.services.lambda.runtime.RequestHandler
-import com.amazonaws.services.lambda.runtime.events.SQSEvent
-import com.gu.productmove.{AwsCredentialsLive, AwsS3, AwsS3Live, GuStageLive, SQSLive, SttpClientLive}
+import com.gu.productmove.*
 import com.gu.productmove.GuStageLive.Stage
-import com.gu.productmove.endpoint.move.ProductMoveEndpointTypes.{ErrorResponse, OutputBody, Success, TransactionError}
-import zio.{Clock, RIO, Task, ZIO}
+import com.gu.productmove.endpoint.move.ProductMoveEndpointTypes.TransactionError
 import com.gu.productmove.invoicingapi.InvoicingApiRefund
-import com.gu.productmove.zuora.InvoiceItemWithTaxDetails
+import com.gu.productmove.zuora.*
 import com.gu.productmove.zuora.model.SubscriptionName
-import com.gu.productmove.zuora.rest.{ZuoraClientLive, ZuoraGetLive}
-import com.gu.productmove.zuora.{
-  CreditBalanceAdjustment,
-  GetAccountLive,
-  GetInvoice,
-  GetRefundInvoiceDetails,
-  GetSubscriptionLive,
-  InvoiceItemAdjustment,
-  RefundInvoiceDetails,
-  SubscribeLive,
-  ZuoraCancel,
-  ZuoraCancelLive,
-  ZuoraSetCancellationReason,
-}
-import sttp.capabilities.zio.ZioStreams
 import sttp.client3.SttpBackend
 import zio.json.{DeriveJsonDecoder, DeriveJsonEncoder, JsonDecoder, JsonEncoder}
+import zio.{Task, ZIO}
 
-import java.time.{LocalDate, LocalDateTime}
+import java.time.LocalDate
 
 case class RefundInput(subscriptionName: SubscriptionName)
 
@@ -117,12 +100,14 @@ object RefundSupporterPlus {
     // - one for the charge where the SourceType is "InvoiceDetail" and SourceId is the invoice item id
     // - one for the tax where the SourceType is "Tax" and SourceId is the taxation item id
     // https://www.zuora.com/developer/api-references/older-api/operation/Object_POSTInvoiceItemAdjustment/#!path=SourceType&t=request
-    invoiceItems.filter(_.amountWithTax != 0).flatMap { invoiceItem =>
+    invoiceItems.filter(item => item.amountWithTax != 0).flatMap { invoiceItem =>
       val chargeAdjustment =
         List(
           InvoiceItemAdjustment.PostBody(
             AdjustmentDate = invoiceItem.chargeDateAsDate,
-            Amount = invoiceItem.ChargeAmount.abs,
+            // ChargeAmount will be negative for charges and positive for discounts,
+            // we need to invert this for the adjustments
+            Amount = invoiceItem.ChargeAmount * -1,
             InvoiceId = invoiceItem.InvoiceId,
             SourceId = invoiceItem.Id,
             SourceType = "InvoiceDetail",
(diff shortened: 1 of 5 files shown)


handlers/product-move-api/src/main/scala/com/gu/productmove/refund/RefundSupporterPlus.scala
```diff
@@ -117,24 +106,49 @@ object RefundSupporterPlus {
...
             SourceType = "InvoiceDetail",
           ),
         )
-      val taxAdjustment = invoiceItem.TaxDetails match {
+      val taxAdjustment = chargeWithDiscount.charge.TaxDetails match {
         case Some(taxDetails) =>
           List(
             InvoiceItemAdjustment.PostBody(
-              AdjustmentDate = invoiceItem.chargeDateAsDate,
-              Amount = taxDetails.amount.abs,
```
rupertbates, line 136: This line caused a problem because we had a mix of positive and negative charges which were all being converted to positives and then summed as individual adjustments.

handlers/product-move-api/src/main/scala/com/gu/productmove/refund/RefundSupporterPlus.scala
```diff
@@ -117,24 +106,49 @@ object RefundSupporterPlus {
...
+      .map(charge => ChargeWithDiscount(charge, if discountedCharge.contains(charge) then maybeDiscount else None))
+
+    val chargeAdjustments = chargesWithDiscounts.flatMap { chargeWithDiscount =>
+      val discountChargeAmount = chargeWithDiscount.discount.map(_.ChargeAmount).getOrElse(BigDecimal(0))
+      val discountTaxAmount = chargeWithDiscount.discount.map(_.taxAmount).getOrElse(BigDecimal(0))
       val chargeAdjustment =
         List(
           InvoiceItemAdjustment.PostBody(
-            AdjustmentDate = invoiceItem.chargeDateAsDate,
-            Amount = invoiceItem.ChargeAmount.abs,
```
rupertbates, line 125: This line caused a problem because we had a mix of positive and negative charges which were all being converted to positives and made into individual adjustments which were then summed together by Zuora.

handlers/product-move-api/src/main/scala/com/gu/productmove/refund/RefundSupporterPlus.scala
```diff
@@ -117,24 +106,49 @@ object RefundSupporterPlus {
...
-            InvoiceId = invoiceItem.InvoiceId,
-            SourceId = invoiceItem.Id,
+            AdjustmentDate = chargeWithDiscount.charge.chargeDateAsDate,
+            // ChargeAmount will be negative for charges and positive for discounts,
+            // we need to invert this for the adjustments
+            Amount = chargeWithDiscount.charge.ChargeAmount.abs - discountChargeAmount,
```
rupertbates, line 137: The fix is that now we subtract the discount amount from the charge amount before creating the adjustment.

handlers/product-move-api/src/main/scala/com/gu/productmove/refund/RefundSupporterPlus.scala
```diff
@@ -117,24 +102,49 @@ object RefundSupporterPlus {
     // - one for the charge where the SourceType is "InvoiceDetail" and SourceId is the invoice item id
     // - one for the tax where the SourceType is "Tax" and SourceId is the taxation item id
     // https://www.zuora.com/developer/api-references/older-api/operation/Object_POSTInvoiceItemAdjustment/#!path=SourceType&t=request
-    invoiceItems.filter(_.amountWithTax != 0).flatMap { invoiceItem =>
+
+    // Charges will have a negative value because we are dealing with the negative invoice created by cancellation,
+    // discounts will have a positive value. Unfortunately we can't just create an adjustment for each of the charges
+    // and discounts as this takes the invoice to values which are not allowed by Zuora. Instead we need to subtract
```
johnduffell, line 108: what does "takes the invoice to values which are not allowed by Zuora" mean?

handlers/product-move-api/src/main/scala/com/gu/productmove/refund/RefundSupporterPlus.scala
```diff
@@ -117,24 +102,49 @@ object RefundSupporterPlus {
...
+    // the discount from the charge and create an adjustment for the difference.
```
johnduffell, line 109:
```suggestion
    // We are dealing with a refund invoice created by the back dated cancellation, so all of the charges are the opposite way around to the original invoice.
    // i.e. Charges will have a negative value and discounts will have a positive value.
    // Zuora doesn't allow the invoice amount go <negative? out of range? 2 adjustments in quick succession?>
    // This means we have to calculate the overall adjustment and then adjust the invoice in one go.
```

handlers/product-move-api/src/main/scala/com/gu/productmove/refund/RefundSupporterPlus.scala
```diff
@@ -117,24 +102,49 @@ object RefundSupporterPlus {
...
+    val (charges, discounts) = invoiceItems
+      .filter(_.amountWithTax != 0) // Ignore any items with a zero amount, they will be the contribution charge where the user is just paying the base price
+      .partition(_.amountWithTax < 0)
+    val maybeDiscount = discounts.headOption
+    val discountedCharge = maybeDiscount.flatMap(discount =>
+      charges.find(charge =>
+        charge.ChargeAmount.abs >= discount.ChargeAmount &&
+          charge.taxAmount.abs >= discount.taxAmount,
+      ),
+    )
```
johnduffell, line 119: In the normal case (especially for a newish and discounted S+) we expect there to be only 1 charge and one discount.
so we assume there's only one discount is there at the most, and then we find (any/the) charge that is bigger than the discount and subtract it from that.
Is that all about right?

handlers/product-move-api/src/main/scala/com/gu/productmove/refund/RefundSupporterPlus.scala
rupertbates: It means I can't remember the exact error message! Basically the adjustments are applied in sequence and trying to add discounts and charge adjustments took the balance out of the allowed parameters - for instance it's not allowed to go from negative to positive and there was one other constraint that I hit.

handlers/product-move-api/src/main/scala/com/gu/productmove/refund/RefundSupporterPlus.scala
johnduffell: we are using a query, and we are only getting Id, ChargeAmount, TaxAmount, ChargeDate, InvoiceId

https://github.com/guardian/support-service-lambdas/blob/45bd77421a85caa15c234efcec371b3e3d0b8c57/handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/GetRefundInvoiceDetails.scala#L30

however you could add an extra couple of fields maybe `RatePlanChargeId`  would match between the two, or failing that `AppliedToChargeNumber`  and `ChargeNumber` to correlate the discount and charge, see
https://knowledgecenter.zuora.com/Zuora_Central_Platform/API/G_SOAP_API/E1_SOAP_API_Object_Reference/InvoiceItem

johnduffell: this flatmap and the one below would probably be neater as for comprehensions
```suggestion
    val discountedCharge = for {
      discount <- maybeDiscount
    } yield charges.find(charge =>
      charge.ChargeAmount.abs >= discount.ChargeAmount &&
        charge.taxAmount.abs >= discount.taxAmount,
    )
```

rupertbates: There will be two charges, one for the subscription and one for the contribution but the contribution one will usually be zero. Then there may also be at most one discount, in which case we find any charge where the charge amount AND the tax amount are larger than the discount charge amount and tax amount.

johnduffell (approved):
good, much improved on the old broken one, but see comments for possible changes or followup work.
as you mentioned maybe the zuora Invoice Settlement feature will let us do this more easily!

PR #2380, opened 2024-07-16
zuora <-> salesforce link remover - enable access to prod secrets for access to salesforce

david-pepper:
## What does this change?
Enable access to prod secrets for salesforce api user and connected app. These credentials will allow the lambdas to authenticate with Salesforce and retrieve the billing accounts for processing.

## How to test
Run in production and verify records are being returned from the start query


tomrf1 (approved):
👍 
but could you also conditionally allow the correct permissions based on `this.stage`?

david-pepper:
> 👍 but could you also conditionally allow the correct permissions based on `this.stage`?

Interesting! so this would mean the prod lambda only has access to the prod credentials. This makes sense. I'll add this to the list of minor tidy-up items to be performed post-release. Thanks!

PR #2384, opened 2024-07-18
Zuora <-> Salesforce link remover - disengage from test mode

david-pepper:
## What does this change?
We have now verified that the state machine has executed and processed a subset of production records records as expected, so we can now remove the test query and use the production query instead.

I've also:

- updated the code in getBillingAccounts lambda to get the stage value from a shared module
- updated the maximum concurrent invocations of the updateZuoraBillingAccount lambda from 1 at a time to 10 at a time, which will reduce the execution time.


tomrf1 (approved):
👍

PR #2385, opened 2024-07-18
zuora <-> salesforce link remover - Run the state machine on a schedule

david-pepper:
## What does this change?
Run every hour in prod, once per year in CODE (because I think there needs to be some value for CODE rather than nothing)


charleycampbell (approved):
Looks good to go 🚀

PR #2392, opened 2024-07-22
Add T3 into new-product-api/catalog

rupertbates:
## What does this change?
This PR makes the new Tier Three product available to buy via the new product api.


graham228221 (approved):
Thanks for this - tested in CODE and all looks good.

PR #2393, opened 2024-07-23
Add salesforce common module

david-pepper:
## What does this change?
Add a Salesforce module that contains actions  common across typescript projects in the repo, such as:

- config e.g. current Salesforce api version
- authentication with Salesforce using connected app and api user credentials
- query for records in Salesforce using REST api
- update records in Salesforce using REST api

test modules have been added for each module

modules have been plugged into from the new zuora-salesforce-link-remover project

This module should be used or added to for Salesforce integrations in future, and retrofitted to existing solutions when it makes sense to do so


handlers/zuora-salesforce-link-remover/src/handlers/getBillingAccounts.ts
```diff
@@ -36,16 +35,35 @@ export async function handler() {
...
+
+export const BillingAccountRecordSchema = RecordSchema.extend({
+	GDPR_Removal_Attempts__c: z.number(),
+	Zuora__External_Id__c: z.string(),
+});
+export type BillingAccountRecord = z.infer<typeof BillingAccountRecordSchema>;
+
+//might need hoisting
+export const BillingAccountRecordsSchema = z.array(BillingAccountRecordSchema);
+export type BillingAccountRecords = z.infer<typeof BillingAccountRecordsSchema>;
```
tomrf1, line 61: is this type useful? I think I'd find `BillingAccountRecord[]` more readable

modules/salesforce/src/query.ts
```diff
@@ -0,0 +1,45 @@
...
+				},
+			},
+		);
+
+		if (!response.ok) {
+			throw new Error(`Failed to execute query: ${response.statusText}`);
+		}
+
+		const sfQueryResponse =
+			(await response.json()) as SalesforceQueryResponse<T>;
```
tomrf1, line 29: I think the cast here is unnecessary if you're parsing it immediately after

modules/salesforce/src/query.ts
```diff
@@ -0,0 +1,45 @@
...
+import type { SfAuthResponse } from './auth';
+import { sfApiVersion } from './config';
+import type { SalesforceQueryResponse } from './recordSchema';
+import { SalesforceQueryResponseSchema } from './recordSchema';
+
+export async function executeSalesforceQuery<T extends z.ZodTypeAny>(
+	sfAuthResponse: SfAuthResponse,
+	query: string,
+	schema: T,
+): Promise<SalesforceQueryResponse<z.infer<T>>> {
```
tomrf1, line 11: is `z.infer` necessary here?

tomrf1:
I found the docs [here](https://github.com/colinhacks/zod?tab=readme-ov-file#writing-generic-functions) useful for following this PR

modules/salesforce/src/query.ts
david-pepper: yep, you're right. I've removed the casting.

modules/salesforce/src/query.ts
david-pepper: Nope. Removed it

handlers/zuora-salesforce-link-remover/src/handlers/getBillingAccounts.ts
david-pepper: agreed, I've updated the code to use ```z.array(BillingAccountRecordSchema)``` instead

modules/salesforce/src/config.ts
```diff
@@ -0,0 +1,8 @@
+export const sfApiVersion = (): string => {
+	const sfApiVersion = process.env.SF_API_VERSION;
+
+	if (!sfApiVersion) {
+		return 'v54.0';
```
charleycampbell, line 5: Does this version change frequently? I wonder if a code comment explaining why v54.0 is used if sfApiVersion is false

charleycampbell (approved):
I really like your console logging through out your PR - it's really been considered where to put it and will help if we ever need to trawl through the logs!

modules/salesforce/src/config.ts
david-pepper: Good shout!

PR #2400, opened 2024-07-29
Improved logging & a test for T3

rupertbates:
## What does this change?
This PR justs adds some additional logging to the soft-opt-in-consent-setter lambda as currently it is very minimal. It also adds a test for the new Tier Three product although I'm not sure there's a massive amount of value to it as it doesn't use the actual json file where consents are defined (this is in S3).

I think it would an improvement to move the definition of consents from this json file into this repo where it can be properly tested and version controlled, but that will be a separate PR.

### DIFF
--- a/handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/IdentityConnector.scala
+++ b/handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/IdentityConnector.scala
@@ -1,12 +1,15 @@
 package com.gu.soft_opt_in_consent_setter
 
 import com.gu.soft_opt_in_consent_setter.models.{IdentityConfig, SoftOptInError}
+import com.typesafe.scalalogging.LazyLogging
 import scalaj.http.{Http, HttpResponse}
+
 import scala.util.Try
 
-class IdentityConnector(config: IdentityConfig) {
+class IdentityConnector(config: IdentityConfig) extends LazyLogging {
 
   def sendConsentsReq(identityId: String, body: String): Either[SoftOptInError, Unit] = {
+    logger.info(s"Sending consents request to Identity for $identityId with body $body")
     handleConsentsResp(
       sendReq(url = s"${config.identityUrl}/users/$identityId/consents", body),
       errorDesc = s"Identity request failed while processing $identityId with body $body",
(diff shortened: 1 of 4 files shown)


handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/IdentityConnector.scala
```diff
@@ -1,7 +1,9 @@
 package com.gu.soft_opt_in_consent_setter
 
 import com.gu.soft_opt_in_consent_setter.models.{IdentityConfig, SoftOptInError}
+import com.typesafe.scalalogging.LazyLogging
```
johnduffell, line 4: does this import get used?

rupertbates: Nope good spot

handlers/soft-opt-in-consent-setter/src/test/scala/com.gu.soft_opt_in_consent_setter/HandlerTests.scala
```diff
@@ -261,6 +261,53 @@ class HandlerTests extends AnyFunSuite with Matchers with MockFactory {
       mockSfConnector,
     )
 
+    result shouldBe Right(())
+  }
+  test(testName = "processAcquiredSub should handle a Tier Three acquisition event correctly") {
+    mockSendConsentsReq
+      .expects(
```
johnduffell, line 268: it's not quite clear when the assertion is actually made, but I assume it fails if the mockSendConsentsReq is not called?
I assume there's some tearDown logic called that will check all these at the end of the test.
I'm used to the mockito style where we do all the assertions afterwards that the mock is called.

rupertbates: Yes it is different to most of the tests we have but it seems to be the pattern in this project

handlers/soft-opt-in-consent-setter/src/test/scala/com.gu.soft_opt_in_consent_setter/testData/ConsentsCalculatorTestData.scala
```diff
@@ -20,6 +27,7 @@ object ConsentsCalculatorTestData {
     "guardianweekly" -> guWeeklyMapping,
     "testproduct" -> testProductMapping,
     "FeastInAppPurchase" -> testFeastMobileSubscriptionMapping,
+    "Tier Three" -> testTierThreeMapping,
```
johnduffell, line 30: I know it's test data but why not be more consistent with the existing product names?  or are the existing ones inconsistent with PROD?

johnduffell (approved):
looks good although see comments, nothing show stopping though.

handlers/soft-opt-in-consent-setter/src/test/scala/com.gu.soft_opt_in_consent_setter/testData/ConsentsCalculatorTestData.scala
rupertbates: The existing ones are inconsistent with what is in PROD, they should map to the product names in Salesforce.

PR #2401, opened 2024-07-30
Refactor: Move deprecated state machine definition to definitionProperties

michaelwmcnamara:
## What does this change?
Statemachine.definition has been deprecated and replaced with definitionBody. 
This change updates the zuora-salesforce-link-remover state-machine, which is still using 'definition' to use definitionBody instead. 

It is used almost exactly the same way.
`
const definition =  getSalesforceBillingAccountsFromLambdaTask.next(billingAccountsExistChoice,);
`
becomes:
`
const definitionBody = StateMachine.definitionBody.fromChainable(getSalesforceBillingAccountsFromLambdaTask
			.next(billingAccountsExistChoice)
  );
`
This work is to help position the repo to start using latest version of cdk. 

## How to test

No functional change and no change to the generated Cloudformation template.
Testing can be limited to deploying to CODE and confirming Cloudformation resources are as expected.
Confirmed no change to cloudformation template.
Confirmed deploy to CODE successful.
Confirmed zero drift in cloudformation stack after deploy.
Confirmed resources all look as expected.

This will remove deprecation warnings and be one less barrier to upgrading cdk and other dependencies.

The change populates the same parts of the State Machine in the same way. Confirming the Cloudformation template is unchanged means that this code change will not introduce any new risk.

### DIFF
--- a/cdk/lib/zuora-salesforce-link-remover.ts
+++ b/cdk/lib/zuora-salesforce-link-remover.ts
@@ -18,6 +18,7 @@ import { Topic } from 'aws-cdk-lib/aws-sns';
 import {
 	Choice,
 	Condition,
+	DefinitionBody,
 	JsonPath,
 	Map,
 	Pass,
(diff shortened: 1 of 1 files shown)


MaelGNM (approved):
Looks good!

cdk/lib/zuora-salesforce-link-remover.ts
```diff
@@ -177,15 +178,17 @@ export class ZuoraSalesforceLinkRemover extends GuStack {
...
+				billingAccountsExistChoice,
+			),
 		);
 
 		const stateMachine = new StateMachine(
 			this,
 			`zuora-salesforce-link-remover-state-machine-${this.stage}`,
 			{
-				definition,
+				definitionBody: definitionBody,
```
jamesgorrie, line 191: If you wanted to be a bit more succinct without losing meaning
```suggestion
				definitionBody,
```

PR #2403, opened 2024-08-01
prioritise holiday stop processing more appropriately

johnduffell:
We have noticed that when there's a lot of holiday stops, and it's processing a whole weekend, sometimes the stops aren't processed in time for fulfilment, especially national delivery.
National delivery fulfilment is being made later in this PR https://github.com/guardian/national-delivery-fulfilment/pull/47

This PR changes it to do national delivery first in the process, so that it's more likely to complete it.


graham228221 (approved):
Thanks John!

PR #2410, opened 2024-08-06
Move Soft opt-in settings into code

rupertbates:
## What does this change?
The soft-opt-in-consent-setter lambdas are responsible for setting the correct opt-in values for users when they subscribe to or cancel products.

Previously the mappings between products and consents were stored in a JSON file in S3 and loaded at run time by the lambda, this PR moves that mapping into the project for 3 reasons:
- so that it is version controlled
- so that it is checked by the compiler as opposed to a json file which could be malformed
- to remove a point of failure when the file is loaded from S3

### DIFF
--- a/handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/models/ConsentsMapping.scala
+++ b/handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/models/ConsentsMapping.scala
@@ -1,6 +1,6 @@
 package com.gu.soft_opt_in_consent_setter.models
 
-object ConsentMappings {
+object ConsentsMapping {
   val yourSupportOnboarding = "your_support_onboarding"
   val similarGuardianProducts = "similar_guardian_products"
   val supporterNewsletter = "supporter_newsletter"
--- a/handlers/soft-opt-in-consent-setter/src/test/scala/com.gu.soft_opt_in_consent_setter/ConsentsCalculatorTests.scala
+++ b/handlers/soft-opt-in-consent-setter/src/test/scala/com.gu.soft_opt_in_consent_setter/ConsentsCalculatorTests.scala
@@ -1,6 +1,6 @@
 package com.gu.soft_opt_in_consent_setter
 
-import com.gu.soft_opt_in_consent_setter.models.SoftOptInError
+import com.gu.soft_opt_in_consent_setter.models.{ConsentsMapping, SoftOptInError}
 import com.gu.soft_opt_in_consent_setter.testData.ConsentsCalculatorTestData.{
   contributionMapping,
   guWeeklyMapping,
@@ -15,11 +15,11 @@ import org.scalatest.matchers.should
 
 class ConsentsCalculatorTests extends AnyFlatSpec with should.Matchers with EitherValues {
 
-  val calculator = new ConsentsCalculator(testConsentMappings)
+  val calculator = new ConsentsCalculator(ConsentsMapping.consentsMapping)
 
   // getAcquisitionConsents success cases
   "getSoftOptInsByProduct" should "correctly return the mapping when a known product is passed" in {
-    calculator.getSoftOptInsByProduct("membership") shouldBe Right(membershipMapping)
+    calculator.getSoftOptInsByProduct("Membership") shouldBe Right(membershipMapping)
   }
 
   // getAcquisitionConsents failure cases
@@ -33,7 +33,7 @@ class ConsentsCalculatorTests extends AnyFlatSpec with should.Matchers with Eith
 
   // getSoftOptInsByProducts success cases
   "getSoftOptInsByProducts" should "correctly return both mappings when two products are passed in" in {
-    calculator.getSoftOptInsByProducts(Set("contributions", "supporterPlus")) shouldBe Right(
+    calculator.getSoftOptInsByProducts(Set("Contributor", "Supporter Plus")) shouldBe Right(
       contributionMapping ++ supporterPlusMapping,
     )
   }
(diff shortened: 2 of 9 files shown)


david-pepper (approved):

jamesgorrie (approved):
Great, way more clear.

[Should we empty the S3 bucket](https://github.com/guardian/recommendations/blob/654e08854c2c0d3e83616339f5558010830d8fbd/AWS.md?plain=1#L77-L80)? If anything. for tech debt, but also to find-by-breaking anything that still uses this?

[I couldn't find anything though in GitHub that was](https://github.com/search?q=org%3Aguardian+soft-opt-in-consent-setter&type=code).

handlers/soft-opt-in-consent-setter/src/test/scala/com.gu.soft_opt_in_consent_setter/ConsentsCalculatorTests.scala
```diff
@@ -33,7 +26,7 @@ class ConsentsCalculatorTests extends AnyFlatSpec with should.Matchers with Eith
 
   // getSoftOptInsByProducts success cases
   "getSoftOptInsByProducts" should "correctly return both mappings when two products are passed in" in {
-    calculator.getSoftOptInsByProducts(Set("contributions", "supporterPlus")) shouldBe Right(
+    calculator.getSoftOptInsByProducts(Set("Contributor", "Supporter Plus")) shouldBe Right(
```
tjmw, line 29: Just to make sure I understand - all these casing changes are necessary because previously we were using a test set of mappings, which had slightly different casing than the actual mappings?

tjmw (approved):
👏  Really nice improvement.

handlers/soft-opt-in-consent-setter/src/test/scala/com.gu.soft_opt_in_consent_setter/ConsentsCalculatorTests.scala
rupertbates: Yes exactly, in the production system these need to match with the product names stored in Salesforce.

rupertbates:
> Great, way more clear.
> 
> [Should we empty the S3 bucket](https://github.com/guardian/recommendations/blob/654e08854c2c0d3e83616339f5558010830d8fbd/AWS.md?plain=1#L77-L80)? To avoid tech debt, but also to find-by-breaking anything that still uses this?
> 
> [I couldn't find anything though in GitHub that was](https://github.com/search?q=org%3Aguardian+soft-opt-in-consent-setter&type=code).

Yes, will do.

handlers/soft-opt-in-consent-setter/src/test/scala/com.gu.soft_opt_in_consent_setter/ConsentsCalculatorTests.scala
```diff
@@ -49,33 +49,29 @@ class ConsentsCalculatorTests extends AnyFlatSpec with should.Matchers with Eith
 
   // getCancellationConsents success cases
   "getCancellationConsents" should "correctly return the mapping when a known product is passed and there are no owned products" in {
-    calculator.getCancellationConsents("membership", Set()) shouldBe Right(membershipMapping)
-  }
-
-  "getCancellationConsents" should "correctly return the mapping when a known product is passed and there are owned products but do not overlap" in {
```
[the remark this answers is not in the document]
johnduffell: if you feel it is a worthwhile test (e.g. if we need to handle the case where it's set up in that way in future) you could parameterise the mappings rather than totally hard coding them.

handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/models/ConsentsMapping.scala
```diff
@@ -0,0 +1,100 @@
...
+      yourSupportOnboarding,
+      similarGuardianProducts,
+      supporterNewsletter,
+    ),
+    "newspaper" -> Set(
+      yourSupportOnboarding,
+      similarGuardianProducts,
+      subscriberPreview,
+      supporterNewsletter,
+    ),
```
johnduffell, line 42: what is this one? I can see all the other types of paper elsewhere in the list

handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/models/ConsentsMapping.scala
```diff
@@ -0,0 +1,100 @@
...
+    "Recurring Contribution" -> Set(
+      yourSupportOnboarding,
+      similarGuardianProducts,
+      supporterNewsletter,
+    ),
+    "Contributor" -> Set(
+      yourSupportOnboarding,
+      similarGuardianProducts,
+      supporterNewsletter,
+    ),
```
johnduffell, line 36: do we really have 3 types? I assume one is single, one is recurring, and the other?

handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/models/ConsentsMapping.scala
```diff
@@ -0,0 +1,100 @@
...
+    "Membership" -> Set(
+      yourSupportOnboarding,
+      similarGuardianProducts,
+      supporterNewsletter,
+    ),
+    "Supporter" -> Set(
+      yourSupportOnboarding,
+      similarGuardianProducts,
+      supporterNewsletter,
+    ),
```
johnduffell, line 21: is supporter part of membership?

handlers/soft-opt-in-consent-setter/src/test/scala/com.gu.soft_opt_in_consent_setter/ConsentsCalculatorTests.scala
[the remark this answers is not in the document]
johnduffell: actually it's already parameterised so could be hard coded separately for this test or even amended from the hard coded data for more realism
```scala
  val calculator = new ConsentsCalculator(ConsentsMapping.consentsMapping.updated(
    "Newspaper - Digital Voucher" -> Set(
      supporterNewsletter,
    ),
    "Newspaper - Voucher Book" -> Set(
      yourSupportOnboarding,
    ),
))
```

johnduffell (approved):
great I totally approve of letting the compiler validate as much configuration as we can get away with.  The way we work it's much easier and more familiar to update code than config not to mention all the other benefits you mention!

handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/models/ConsentsMapping.scala
rupertbates: Yes, I'm not sure why it's in there but it's from the prod json file. These keys all map to product names in Salesforce so it's possible it's not used but I didn't want to risk it.

handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/models/ConsentsMapping.scala
rupertbates: Again, these are taken from the prod json file so may not be used but we'd need to do more investigation before removing them.

handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/models/ConsentsMapping.scala
rupertbates: Not sure, you would think so but as with the questions above, it's not really clear why the current prod data is the way it is as there is no version history for it.

rupertbates:
> > Great, way more clear.
> > [Should we empty the S3 bucket](https://github.com/guardian/recommendations/blob/654e08854c2c0d3e83616339f5558010830d8fbd/AWS.md?plain=1#L77-L80)? To avoid tech debt, but also to find-by-breaking anything that still uses this?
> > [I couldn't find anything though in GitHub that was](https://github.com/search?q=org%3Aguardian+soft-opt-in-consent-setter&type=code).
> 
> Yes, will do.

Done

PR #2413, opened 2024-08-09
New ticket-tailor-webhook lambda

tjmw:
## What does this change?

This PR introduces a new pair of cloudformation stacks (1 for CODE, 1 for prod) where we define the infrastructure for a webhook to handle Ticket Tailor [`order.create`](https://developers.tickettailor.com/#configuration) events.

The architecture looks like this:

`API Gateway --> SQS Queue --> Lambda`

At the moment, all the lambda does is log the event payload. In future we'll start doing stuff with the data received.

I used `pnpm new-lambda` to create the new lambda, and then modified the generated code (adding the SQS queue for example).

See also:

* guardian/support-reminders#204
* https://docs.aws.amazon.com/prescriptive-guidance/latest/patterns/integrate-amazon-api-gateway-with-amazon-sqs-to-handle-asynchronous-rest-apis.html

## How to test

We've tested by invoking the API Gateway using curl from the command line.

## Accessibility


handlers/ticket-tailor-webhook/test/firstIntegration.test.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * This is an integration test, the `@group integration` tag ensures that it will only be run by the `pnpm it-test`
+ * command and will not be run during continuous integration.
+ * This makes it useful for testing things that require credentials which are available locally but not on the CI server.
+ *
+ * @group integration
+ */
+test('my app', () => {
```
michaelwmcnamara, line 8: If these tests are just placeholders for future work, we should include a comment to that effect.

michaelwmcnamara: This also applies to https://github.com/guardian/support-service-lambdas/pull/2413/files#diff-da676424c2a997dc52fc9512ba4307925beb04a4f70b41cc80ede7a9c154b5be

michaelwmcnamara: Removed these auto-generated tests for now. They are easy to re-add in future when there is something to test.

Joemgru: it looks like removing the tests has inadvertently caused the build to fail.

Im happy with the work in the PR but I guess we need to either put the tests back in to please the gods of CI or, alternatively change the CI command to allow no tests ?

michaelwmcnamara: I've restored the tests for now and added a comment to explain they are placeholders. We can discuss the ecclesiastical wisdom of modifying the sacred ci texts tomorrow. All praise to the CI Gods!

Joemgru (approved):
lgtm

handlers/ticket-tailor-webhook/src/index.ts
```diff
@@ -0,0 +1,15 @@
+import type {
+	APIGatewayProxyEvent,
+	APIGatewayProxyResult,
+	Handler,
+} from 'aws-lambda';
+
+export const handler: Handler = async (
+	event: APIGatewayProxyEvent,
```
tjmw, line 8: Should this be an SQS event type?

cdk/lib/ticket-tailor-webhook.ts
```diff
@@ -0,0 +1,149 @@
...
+			path: `${this.account}/${queue.queueName}`,
+			integrationHttpMethod: 'POST',
+			options: {
+				credentialsRole: apiRole,
+				requestParameters: {
+					'integration.request.header.Content-Type':
+						"'application/x-www-form-urlencoded'",
+				},
+				requestTemplates: {
+					'application/json': 'Action=SendMessage&MessageBody=$input.body',
```
tjmw, line 83: @Joemgru @michaelwmcnamara not for this PR I don't think, but in the next stage when we want to verify that the signature in the header is correct I think we'll need to change this template to make sure that the header value is correctly made available. It looks like there's an example of this in the support-reminders project where the geo header is passed through: https://github.com/guardian/support-reminders/pull/204/files#diff-e48cd48cee203f62637ed3a72fc9ed12cbb9724a1eb3061cc1eff51cc0a415d6R105

handlers/ticket-tailor-webhook/src/index.ts
tjmw: @Joemgru and I chatted about this offline and agreed to leave this for the next PR (at the moment the type doesn't really matter as we're not trying to read any specific properties from it).

handlers/ticket-tailor-webhook/src/index.ts
```diff
@@ -0,0 +1,15 @@
...
+): Promise<APIGatewayProxyResult> => {
+	console.log(`Input is ${JSON.stringify(event)}`);
+	return await Promise.resolve({
+		body: 'Hello World',
+		statusCode: 200,
```
tjmw, line 13: In future we can change this - we're not responding to an HTTP request directly here so it doesn't make sense to return a body or status code.

handlers/ticket-tailor-webhook/src/index.ts
michaelwmcnamara: Good catch though.

PR #2419, opened 2024-08-13
Removed digitalSubscriberPreview consent as it's not needed

MaelGNM:
## What does this change?

This removes the `digitalSubscriberPreview` consent from the soft opt-in consent setter for all subscriptions as it's not used by MRR anymore.

### DIFF
--- a/handlers/soft-opt-in-consent-setter/src/test/scala/com.gu.soft_opt_in_consent_setter/HandlerTests.scala
+++ b/handlers/soft-opt-in-consent-setter/src/test/scala/com.gu.soft_opt_in_consent_setter/HandlerTests.scala
@@ -38,10 +38,6 @@ class HandlerTests extends AnyFunSuite with Matchers with MockFactory {
       .expects(
         identityId,
         """[
-          |  {
-          |    "id" : "digital_subscriber_preview",
-          |    "consented" : true
-          |  }
           |]""".stripMargin,
       )
       .returning(Right(()))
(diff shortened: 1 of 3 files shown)


handlers/soft-opt-in-consent-setter/src/test/scala/com.gu.soft_opt_in_consent_setter/HandlerTests.scala
```diff
@@ -291,6 +283,10 @@ class HandlerTests extends AnyFunSuite with Matchers with MockFactory {
           |  {
           |    "id" : "supporter_newsletter",
           |    "consented" : true
+          |  },
+          |  {
+          |    "id" : "guardian_weekly_newsletter",
+          |    "consented" : true
```
MaelGNM, line 289: I had to move this one around for the test to pass

rBangay: Does the test expect this list/array to be in a particular order? Do you know why you had to move this in order for the test to pass?

MaelGNM: I don't know the tests very much myself, but in the logs the only difference between the expected and actual results was the order. So I reordered it to match and then the tests all passed

rupertbates (approved):
Nice 👍

PR #2422, opened 2024-08-14
Standardise the node runtime used by TS lambdas

rupertbates:
## What does this change?
This PR extracts the node runtime used by Typescript lambdas into a const that can be imported into CDK lib files. this means that we only need to update one place when we want to upgrade the node version our lambdas are running on.

I have tested all the lambdas which have been upgraded to Node 20 as a result of this change on CODE.


johnduffell (approved):
good idea 👍

AndreaDiotallevi (approved):
More consistent - great!

cdk/lib/node-version.ts
```diff
@@ -0,0 +1,4 @@
+import { Runtime } from 'aws-cdk-lib/aws-lambda';
+
+// If you change this version make sure you update the version in the .nvmrc file as well
```
jamesgorrie, line 3: Nice comment - worth leaving one in the .nvmrc file too?

jamesgorrie (approved):
Nice and easy

cdk/lib/node-version.ts
rupertbates: I tried but the file format doesn't allow comments unfortunately.

PR #2425, opened 2024-08-19
Move invoicing-api to Platform

graham228221:
## What does this change?
Moves the invoicing-api alarms to the Platform team.

This repo was designed as a set of small, reusable functions that interact with Zuora. It's only used in a few limited places, notably by manage-frontend to [get invoices to display in MMA](https://github.com/guardian/manage-frontend/blob/1edf632dccddf5c0e7082119e7ee8d873d89896a/server/routes/api.ts#L304), [get a specific pdf for a user in MMA](https://github.com/guardian/manage-frontend/blob/1edf632dccddf5c0e7082119e7ee8d873d89896a/server/routes/api.ts#L304) and [for Supporter Plus cancellation refunds](https://github.com/guardian/support-service-lambdas/blob/main/handlers/product-move-api/src/main/scala/com/gu/productmove/invoicingapi/InvoicingApiRefund.scala).

### DIFF
--- a/handlers/alarms-handler/src/alarmMappings.ts
+++ b/handlers/alarms-handler/src/alarmMappings.ts
@@ -30,7 +30,6 @@ const teamToAppMappings: Record<Team, string[]> = {
 		'digital-voucher-api',
 		'dotcom-components',
 		'fulfilment-date-calculator',
-		'invoicing-api',
 		...sharedMobilePurchasesApps,
 		'new-product-api',
 		'price-migration-engine-state-machine',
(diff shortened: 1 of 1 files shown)


tomrf1 (approved):

graham228221:
@johnduffell Paul pointed out that invoicing-api has been earmarked to move to Value, what do you think? I'll go ahead and merge this to move the alarms out of Growth, we can easily raise another PR to move it again.

I think the arguments for Value: 
- invoicing-api is ONLY used by MMA currently, although it was intended to be reused more widely there's not been any work towards that recently. 
- There's been discussions around migrating it into support-service-lambdas, closer to the applications that actually use it

For Platform:
- as is, it's essentially a service layer on top of Zuora, and Zuora oversight is with Platform
- the most recent alarms (related to the pdf endpoint) were caused by (and subsequently fixed by) the Platform team 

cc @paulbrown1982

PR #2430, opened 2024-08-20
fix high severity vulnerabilities

johnduffell:
This PR fixes various high severity and medium vulnerabilities.
High severity were logback, netty and jackson

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -589,6 +589,8 @@ lazy val `zuora-datalake-export` = lambdaProject(
   "zuora-datalake-export",
   "Zuora to Datalake export using Stateful AQuA API which exports incremental changes",
   Seq(scalaLambda, scalajHttp, awsS3, enumeratum),
+).settings(
+  scalaLambdaCirceOverride
 )
 
 lazy val `batch-email-sender` = lambdaProject(
@@ -607,6 +609,8 @@ lazy val `holiday-stop-processor` = lambdaProject(
     `holiday-stops` % "compile->compile;test->test",
     effects,
   ),
+).settings(
+  scalaLambdaCirceOverride
 )
 
 lazy val `delivery-problem-credit-processor` = lambdaProject(
@@ -621,6 +625,8 @@ lazy val `delivery-problem-credit-processor` = lambdaProject(
     diffx,
   ),
   Seq(`credit-processor`, `salesforce-sttp-client`, effects),
+).settings(
+  scalaLambdaCirceOverride
 )
 
 lazy val `product-move-api` = lambdaProject(
@@ -710,6 +716,8 @@ lazy val `fulfilment-date-calculator` = lambdaProject(
   "Generate files in S3 bucket containing relevant fulfilment-related dates, for example, acquisitionsStartDate, holidayStopFirstAvailableDate, etc.",
   Seq(scalaLambda, scalajHttp, enumeratum),
   Seq(testDep, `fulfilment-dates`),
+).settings(
+  scalaLambdaCirceOverride
 )
 
 lazy val `delivery-records-api` = lambdaProject(
--- a/handlers/delivery-records-api/src/main/scala/com/gu/delivery_records_api/DeliveryRecordApiRoutes.scala
+++ b/handlers/delivery-records-api/src/main/scala/com/gu/delivery_records_api/DeliveryRecordApiRoutes.scala
@@ -5,7 +5,10 @@ import cats.effect.Effect
 import cats.syntax.all._
 import com.gu.delivery_records_api.service.createproblem.{CreateDeliveryProblem, CreateDeliveryProblemService}
 import com.gu.delivery_records_api.service.getrecords.{DeliveryRecordsApiResponse, GetDeliveryRecordsService}
-import com.gu.delivery_records_api.service.{DeliveryRecordServiceGenericError, DeliveryRecordServiceSubscriptionNotFound}
+import com.gu.delivery_records_api.service.{
+  DeliveryRecordServiceGenericError,
+  DeliveryRecordServiceSubscriptionNotFound,
+}
 import com.gu.salesforce.{Contact, SalesforceHandlerSupport}
 import io.circe.generic.auto._
 import org.http4s.circe.CirceEntityDecoder._
@@ -17,15 +20,18 @@ import java.time.LocalDate
 
 case class DeliveryRecordApiRoutesError(message: String)
 
-class DeliveryRecordApiRoutes[F[_]: Effect](createDeliveryProblemService: CreateDeliveryProblemService[F], getDeliveryRecordsService: GetDeliveryRecordsService[F]) {
+class DeliveryRecordApiRoutes[F[_]: Effect](
+    createDeliveryProblemService: CreateDeliveryProblemService[F],
+    getDeliveryRecordsService: GetDeliveryRecordsService[F],
+) {
 
   private object http4sDsl extends Http4sDsl[F]
   import http4sDsl._
 
   val routes: HttpRoutes[F] = HttpRoutes.of[F] {
 
-    case request @ GET -> Root / "delivery-records" / subscriptionNumber => (
-      for {
+    case request @ GET -> Root / "delivery-records" / subscriptionNumber =>
+      (for {
         contact <- getContactFromHeaders(request)
         startDate <- parseDateFromQueryString(request, "startDate")
         endDate <- parseDateFromQueryString(request, "endDate")
@@ -38,8 +44,8 @@ class DeliveryRecordApiRoutes[F[_]: Effect](createDeliveryProblemService: Create
         )
       } yield Ok(records)).merge.flatten
 
-    case request @ GET -> Root / "delivery-records" / subscriptionNumber / "cancel" => (
-      for {
+    case request @ GET -> Root / "delivery-records" / subscriptionNumber / "cancel" =>
+      (for {
         contact <- getContactFromHeaders(request)
         cancellationEffectiveDate <- parseDateFromQueryString(request, "effectiveCancellationDate")
         records <- getDeliveryRecords(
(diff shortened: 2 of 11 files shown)


build.sbt
```diff
@@ -589,6 +589,8 @@ lazy val `zuora-datalake-export` = lambdaProject(
   "zuora-datalake-export",
   "Zuora to Datalake export using Stateful AQuA API which exports incremental changes",
   Seq(scalaLambda, scalajHttp, awsS3, enumeratum),
+).settings(
+  scalaLambdaCirceOverride
```
johnduffell, line 593: scala lambda library uses an older version of circe, which is not officially binary compatible, but seems to compile fine against the later version in my PR https://github.com/mkotsur/aws-lambda-scala/pull/30 , so hopefully the binary incompatibility isn't going to hit us at run time.
This does the minimum to stop it being a fatal error at the `sbt update` stage.

handlers/delivery-records-api/src/main/scala/com/gu/delivery_records_api/DeliveryRecordApiRoutes.scala
```diff
@@ -5,7 +5,10 @@ import cats.effect.Effect
 import cats.syntax.all._
 import com.gu.delivery_records_api.service.createproblem.{CreateDeliveryProblem, CreateDeliveryProblemService}
 import com.gu.delivery_records_api.service.getrecords.{DeliveryRecordsApiResponse, GetDeliveryRecordsService}
-import com.gu.delivery_records_api.service.{DeliveryRecordServiceGenericError, DeliveryRecordServiceSubscriptionNotFound}
+import com.gu.delivery_records_api.service.{
+  DeliveryRecordServiceGenericError,
+  DeliveryRecordServiceSubscriptionNotFound,
+}
```
johnduffell, line 11: it seems that if I change a file, scalafmt starts flagging existing unchanged parts as not being formatted correctly

handlers/delivery-records-api/src/main/scala/com/gu/delivery_records_api/DeliveryRecordApiRoutes.scala
```diff
@@ -76,18 +82,18 @@ class DeliveryRecordApiRoutes[F[_]: Effect](createDeliveryProblemService: Create
   private def getContactFromHeaders(request: Request[F]): EitherT[F, F[Response[F]], Contact] = {
     SalesforceHandlerSupport
       .extractContactFromHeaders(
-        request.headers.toList.map(header => header.name.value -> header.value),
+        request.headers.headers.map(header => header.name.toString -> header.value),
```
johnduffell, line 85: this didn't compile due to the http4s update

handlers/delivery-records-api/src/main/scala/com/gu/delivery_records_api/DeliveryRecordsApiApp.scala
```diff
@@ -46,7 +46,7 @@ object DeliveryRecordsApiApp extends LazyLogging {
     Logger.httpRoutes(
       logHeaders = true,
       logBody = true,
-      redactHeadersWhen = { headerKey: CaseInsensitiveString => headerKey.value == "x-api-key" },
+      redactHeadersWhen = { headerKey: CIString => headerKey.toString == "x-api-key" },
```
johnduffell, line 49: http4s update

handlers/delivery-records-api/src/test/scala/com/gu/delivery_records_api/DeliveryRecordsApiEffectsTest.scala
```diff
@@ -79,13 +88,13 @@ class DeliveryRecordsApiEffectsTest extends AnyFlatSpec with DiffShouldMatcher w
 
     val request = Request[IO](
       method = Method.GET,
-      Uri(path = s"/delivery-records/$subscriptionNumber"),
-      headers = Headers.of(Header("x-identity-id", identityId)),
+      uri = Uri(path = Uri.Path.unsafeFromString(s"/delivery-records/$subscriptionNumber")),
+      headers = Headers(Header.Raw(CIString("x-identity-id"), identityId)),
```
johnduffell, line 92: http4s update (and repeated several times in this test)

lib/http4s-lambda-handler/src/main/scala/com/gu/http4s/Http4sLambdaHandler.scala
```diff
@@ -83,7 +82,7 @@ class Http4sLambdaHandler(service: HttpRoutes[IO]) {
       responseBody <- http4sResponse
         .attemptAs[String]
         .leftMap(decodingFailure => s"Failed to convert response body to string: ${decodingFailure}")
-      headers = http4sResponse.headers.iterator.map { header => (header.name.toString(), header.value) }.toMap
+      headers = http4sResponse.headers.headers.map { header => (header.name.toString, header.value) }.toMap
```
johnduffell, line 85: http4s update - toString is now a field, so it can't be called with parentheses

lib/http4s-lambda-handler/src/main/scala/com/gu/http4s/Http4sLambdaHandler.scala
```diff
@@ -110,14 +109,14 @@ class Http4sLambdaHandler(service: HttpRoutes[IO]) {
         }
         .toList
 
-    Uri(path = apiGateWayRequest.path, query = Query(queryStringValues: _*))
+    Uri(path = Uri.Path.unsafeFromString(apiGateWayRequest.path), query = Query(queryStringValues: _*))
```
johnduffell, line 112: http4s update - I used `unsafe...`, as it's come through from API gateway as a single opaque string so I don't see how we can build it more safely even if we should.

lib/http4s-lambda-handler/src/main/scala/com/gu/http4s/Http4sLambdaHandler.scala
```diff
@@ -110,14 +109,14 @@ class Http4sLambdaHandler(service: HttpRoutes[IO]) {
...
 
   private def extractHeaders(apiGateWayRequest: LambdaRequest): Headers = {
-    Headers.of(
+    Headers(
       apiGateWayRequest.multiValueHeaders
         .getOrElse(Nil)
-        .flatMap { case (key, multiValue) => multiValue.map(value => Header(key, value)) }
+        .flatMap { case (key, multiValue) => multiValue.map[Header.ToRaw](value => (key, value)) }
         .toList: _*,
     )
```
johnduffell, line 121: `Headers.apply`  seems to take the values as a Header.ToRaw which is the target of an implicit conversion from various other kinds of value, so I've leveraged that.  The other option (to avoid the implicit conversion) would be to build a `Header.Raw` myself, and pass in to `new Headers`
more like this
```suggestion
    new Headers(
      apiGateWayRequest.multiValueHeaders
        .getOrElse(Nil)
        .flatMap { case (key, multiValue) => multiValue.map(value => Header.Raw(CIString(key), value)) }
        .toList: _*,
    )
```

project/Dependencies.scala
```diff
@@ -1,18 +1,19 @@
...
 
 object Dependencies {
-  val awsSdkVersion = "2.21.10"
-  val circeVersion = "0.13.0"
-  val sttpVersion = "3.8.15"
-  val http4sVersion = "0.21.34"
+  val awsSdkVersion = "2.27.7"
+  val circeVersion = "0.14.9"
+  val sttpVersion = "3.9.7"
+  val http4sVersion = "0.22.15" // later versions pull in cats effect 3 which is not compatible
```
johnduffell, line 11: these are most of the version bumps, although we are ok for now, we are a bit stuck with http4s as ce2->ce3 is a big change with a new API.
Having said that, ce2 is likely to be maintained for a while, so we can always use dependency overrides if http4s doesn't maintain the old line.

project/Dependencies.scala
```diff
@@ -84,15 +93,15 @@ object Dependencies {
   val scalaMock = "org.scalamock" %% "scalamock" % "5.2.0" % Test
   val mockito = "org.mockito" % "mockito-core" % "5.3.1" % Test
   val nettyCodec = "io.netty" % "netty-codec" % "4.1.92.Final"
-  val jacksonVersion = "2.13.2"
-  val jacksonDatabindVersion = "2.13.2.2"
+  // play-json still uses an old version of jackson-core which has a vulnerability - https://security.snyk.io/vuln/SNYK-JAVA-COMFASTERXMLJACKSONCORE-7569538
+  val jacksonVersion = "2.17.2"
```
johnduffell, line 97: stolen from https://github.com/guardian/support-admin-console/pull/621 
 normally we can't bump minor versions of jackson, but if it works there, I'll do it here.
There are some upstream issues that have blocked a new version of play-json https://github.com/playframework/playframework/pull/12440

lib/http4s-lambda-handler/src/main/scala/com/gu/http4s/Http4sLambdaHandler.scala
```diff
@@ -110,14 +109,14 @@ class Http4sLambdaHandler(service: HttpRoutes[IO]) {
...
+    Uri(path = Uri.Path.unsafeFromString(apiGateWayRequest.path), query = Query(queryStringValues: _*))
   }
 
   private def extractHeaders(apiGateWayRequest: LambdaRequest): Headers = {
-    Headers.of(
+    Headers(
       apiGateWayRequest.multiValueHeaders
         .getOrElse(Nil)
-        .flatMap { case (key, multiValue) => multiValue.map(value => Header(key, value)) }
+        .flatMap { case (key, multiValue) => multiValue.map[Header.ToRaw](value => (key, value)) }
```
johnduffell, line 119:
```suggestion
        .flatMap { case (key, multiValue) => multiValue.map[Header.ToRaw](value => (key, value)) } // implicit conversion from pair to "ToRaw"
```

michaelwmcnamara (approved):
Discussed all questions in person.
Happy for this to merge.

PR #2431, opened 2024-08-20
fix: move metrics-push-api alarms to alarms-handler

jamesgorrie:
Alerts `alarms-handler` from `metric-push-api` and assigns those to Portfolio


andrewHEguardian (approved):
Not even sure what metric-push-api is doing but 👍

jamesgorrie:
> Not even sure what metric-push-api is doing but 👍

https://github.com/guardian/support-service-lambdas/blob/main/handlers/metric-push-api/README.md

Basically acting as an error proxy for our JS - a workaround to not being able to serve errors via HTTP, AKA the problem we've been having.

PR #2432, opened 2024-08-20
return a billing preview before discount when previewing a discount

johnduffell:
as per https://trello.com/c/yd1qCiC6/425-return-billing-preview-dates-and-amounts-before-discount-from-discount-api-endpoint

We want to display the two missed payments if someone has a price rise during their proposed discount/suspension.

This PR makes the discount preview endpoint return the "before" amounts for all the payments affected by the discount.

### DIFF
--- a/handlers/discount-api/src/discountEndpoint.ts
+++ b/handlers/discount-api/src/discountEndpoint.ts
@@ -6,9 +6,9 @@ import { addDiscount, previewDiscount } from '@modules/zuora/addDiscount';
 import {
 	billingPreviewToSimpleInvoiceItems,
 	getBillingPreview,
-	getOrderedInvoiceTotals,
 	getNextInvoiceItems,
 	getNextNonFreePaymentDate,
+	getOrderedInvoiceTotals,
 } from '@modules/zuora/billingPreview';
 import { zuoraDateFormat } from '@modules/zuora/common';
 import { getAccount } from '@modules/zuora/getAccount';
--- a/handlers/discount-api/src/discountEndpoint.ts
+++ b/handlers/discount-api/src/discountEndpoint.ts
@@ -6,6 +6,7 @@ import { addDiscount, previewDiscount } from '@modules/zuora/addDiscount';
 import {
 	billingPreviewToSimpleInvoiceItems,
 	getBillingPreview,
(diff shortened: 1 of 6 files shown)


handlers/discount-api/src/eligibilityChecker.ts
```diff
@@ -27,7 +27,9 @@ export class EligibilityChecker {
 			`${accountBalance}`,
 		);
 
-		console.log('Working out the date to apply the discount');
+		console.log(
+			'ensuring there are no refunds/discounts expected on the affected invoices',
+		);
```
johnduffell, line 32: this was just a copy and paste error previously

handlers/discount-api/src/responseSchema.ts
```diff
@@ -6,6 +6,9 @@ export const previewDiscountSchema = z.object({
 	upToPeriodsType: z.string(),
 	firstDiscountedPaymentDate: z.string(),
 	nextNonDiscountedPaymentDate: z.string(),
+	nonDiscountedPayments: z.array(
+		z.object({ amount: z.number(), date: z.string() }),
+	),
```
johnduffell, line 11: this is the new field that is returned with the extra data

johnduffell: see handlers/discount-api/test/previewDiscountIntegration.test.ts for examples of the data format

handlers/discount-api/src/discountEndpoint.ts
```diff
@@ -38,15 +39,11 @@ export const previewDiscountEndpoint = async (
...
-		account,
-		zuoraClient,
-		today,
-	);
+	const { discount, dateToApply, orderedInvoiceTotals } =
+		await getDiscountToApply(stage, subscription, account, zuoraClient, today);
 
 	console.log('Preview the new price once the discount has been applied');
+	// note that this only returns the next payment - payments are not guaranteed to be identical
 	const previewResponse = await previewDiscount(
```
johnduffell, line 47: I'd quite like to change previewDiscount to use a billing preview via amendment rather than previwing a single invoice, then we could safely preview non-free discounts where a price rise is in effect.

Joemgru (approved):

PR #2420, opened 2024-08-13
verify ticket tailor hmac signature

michaelwmcnamara:
## Accessibility

-   [ ] [Tested with screen reader](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#screen-reader)
-   [ ] [Navigable with keyboard](https://github.com/guardian/accessibility/blob/main/people-and-technology/02-physical.md#keyboard)
-   [ ] [Colour contrast passed](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#contrast)
-   [ ] [The change doesn't use only colour to convey meaning](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#use-of-colour)

PR #2434, opened 2024-08-21
Invalidate webhook if tickettailor-webhook-signature has a timestamp older than 5 minutes

michaelwmcnamara:
## What does this change?

This is a follow-on to https://github.com/guardian/support-service-lambdas/pull/2420 
It implements the second security recommendation in the [ticket-tailor webhook documentation](https://developers.tickettailor.com/#security): 
"We also recommend to invalidate webhooks received where the sent timestamp is older than 5 minutes"

We do this by checking the timestamp in tickettailor-webhook-signature (which is used in validating the signature) and ensuring that it is 300 seconds old or less.

This minimises the window within which a captured request (which will have a valid signature) could be re-used with bad intentions.

Trello card [here:](https://trello.com/c/RoGDmFz2/476-add-5-min-timeout-to-ticket-tailor-webhook-validation).

## How to test

Tests were added to confirm the validation would fail if a request with a 

## Accessibility


handlers/ticket-tailor-webhook/test/validateRequest.test.ts
```diff
@@ -0,0 +1,151 @@
...
+} from '../src/validateRequest';
+import { HmacKey } from '../src/hMacKey';
+
+/**
+ * This is a unit test, it can be run by the `pnpm test` command, and will be run by the CI/CD pipeline
+ * WARNING: Do not alter key or mock records in this file as it will stop the test from working.
+ */
+
+const mockKey: HmacKey = {
+	secret: '9dn189d53me1ania7d73a45d5de4674d',
```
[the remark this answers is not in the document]
michaelwmcnamara: These are dummy credentials used for unit testing they are not valid for use anywhere.

handlers/ticket-tailor-webhook/src/validateRequest.ts
```diff
@@ -0,0 +1,99 @@
...
+
+export interface Payload {
+	payload: {
+		buyer_details: {
+			email: string;
+		};
+	};
+}
+
+export const getTimestampAndSignature = (
```
michaelwmcnamara, line 13: Moved this code to its own method as it was needed for both the isWithinTimeWindow() and the hasMatchingSignature() method.

handlers/ticket-tailor-webhook/src/validateRequest.ts
```diff
@@ -0,0 +1,99 @@
...
+import type { HmacKey } from './hMacKey';
+
+export interface Payload {
+	payload: {
+		buyer_details: {
+			email: string;
+		};
+	};
+}
+
```
michaelwmcnamara, line 12: Moved the code to get timestamp and signature from the 'tickettailor-webhook-signature' messageAttribute to its own method. The information in this message attribute is needed for both the isWithinTimeWindow() and the hasMatchingSignature() method.

Joemgru:
Struggling with this - it doesn't seem to have the changes I made from the Mael's review of the previous work ?

handlers/ticket-tailor-webhook/src/validateRequest.ts
```diff
@@ -0,0 +1,99 @@
...
+	record: SQSRecord,
+): [string, string] => {
+	const signatureWithTs =
+		record.messageAttributes['tickettailor-webhook-signature']?.stringValue;
+	if (typeof signatureWithTs === 'string') {
+		const timestamp = signatureWithTs.split(',')[0]?.split('=')[1];
+		const signature = signatureWithTs.split(',')[1]?.split('=')[1];
+		if (timestamp && signature) {
+			if (!isNaN(Number(timestamp))) {
+				return [timestamp, signature];
```
michaelwmcnamara, line 23: returning [string, string] and throwing error if any problems means we don't need to check the type any more when using timestamp or  signature.

handlers/ticket-tailor-webhook/src/validateRequest.ts
```diff
@@ -0,0 +1,99 @@
...
+
+export const isWithinTimeWindow = (
+	allowedTimeWindowInSeconds: number,
+	timestamp: string,
+	currentDateTime: Date,
+): boolean => {
+	const timestampEpochSeconds = Number(timestamp);
+	const timeDiff = currentDateTime.getSeconds() - timestampEpochSeconds;
+	if (timeDiff < 0) {
+		throw new Error(
```
michaelwmcnamara, line 49: Erroring for this case here rather than returning false as it is a validation issue, not a timing issue.

handlers/ticket-tailor-webhook/src/validateRequest.ts
```diff
@@ -0,0 +1,99 @@
...
+export const getTimestampAndSignature = (
+	record: SQSRecord,
+): [string, string] => {
+	const signatureWithTs =
+		record.messageAttributes['tickettailor-webhook-signature']?.stringValue;
+	if (typeof signatureWithTs === 'string') {
+		const timestamp = signatureWithTs.split(',')[0]?.split('=')[1];
+		const signature = signatureWithTs.split(',')[1]?.split('=')[1];
+		if (timestamp && signature) {
+			if (!isNaN(Number(timestamp))) {
```
michaelwmcnamara, line 22: This confirms the timestamp is a numeric string. Timestamp should be in EpochSecond format.

handlers/ticket-tailor-webhook/src/validateRequest.ts
```diff
@@ -0,0 +1,99 @@
...
+		);
+	} else {
+		return timeDiff <= allowedTimeWindowInSeconds;
+	}
+};
+
+export const hasMatchingSignature = (
+	timestamp: string,
+	signature: string,
+	record: SQSRecord,
```
michaelwmcnamara, line 60: Passing in the timestamp and signature rather than extracting from the record means we only need to run the string handling extraction function once. Also it makes it easier to test variant scenarios for this method.

handlers/ticket-tailor-webhook/src/validateRequest.ts
```diff
@@ -0,0 +1,99 @@
...
+	const hash = createHmac('sha256', validationSecret.secret)
+		.update(timestamp.concat(record.body))
+		.digest('hex');
+	return timingSafeEqual(Buffer.from(hash), Buffer.from(signature));
+};
+
+export const validateRequest = (
+	record: SQSRecord,
+	validationSecret: HmacKey,
+): void => {
```
michaelwmcnamara, line 72: This method throws errors in place rather than returning a boolean value for success or failure. Reason for this is that there are too many possible ways to fail the request validation now. Simply passing a boolean back would lose too much information.

handlers/ticket-tailor-webhook/test/validateRequest.test.ts
```diff
@@ -0,0 +1,151 @@
...
+	const signatureCheckResult = hasMatchingSignature(
+		invalidTimestamp,
+		validSQSRecordSignature,
+		invalidTimestampSQSRecord,
+		mockKey,
+	);
+	expect(signatureCheckResult).toBe(false);
+});
+
+test('If a valid SQS event has a timestamp that is just on the allowed time window, isWithinTimeWindow() will return true', () => {
```
michaelwmcnamara, line 125: Timestamp is factored into the generation of the signature value, therefore a signature that otherwise matches the content of the request body should also fail if the timestamp is incorrect.

handlers/ticket-tailor-webhook/test/validateRequest.test.ts
```diff
@@ -0,0 +1,151 @@
...
+	const signatureCheckResult = hasMatchingSignature(
+		validSQSRecordTimestamp,
+		invalidSignature,
+		invalidSignatureSQSRecord,
+		mockKey,
+	);
+	expect(signatureCheckResult).toBe(false);
+});
+
+test('If the SQS event has a valid signature but an invalid value for timestamp, hasMatchingSignature() will return false', () => {
```
michaelwmcnamara, line 115: Timestamp is factored into the generation of the signature value, therefore a signature that otherwise matches the content of the request body should also fail if the timestamp is incorrect.

handlers/ticket-tailor-webhook/test/validateRequest.test.ts
michaelwmcnamara, line 125: Added tests for isWithinTimeWindow()

michaelwmcnamara:
> Struggling with this- it doesn't seem to have the changes I made from the Mael's review of the previous work ? maybe something gone wrong with the merge.
> 
> I think the main two things were the handler return type and the logic&comment for splitting out the ts & signature https://github.com/guardian/support-service-lambdas/blob/main/handlers/ticket-tailor-webhook/src/index.ts#L7
> 
> https://github.com/guardian/support-service-lambdas/blob/main/handlers/ticket-tailor-webhook/src/verifySignature.ts#L21-L23

Thats ok I think I need to merge main and fix any conflicts.

michaelwmcnamara:
> > Struggling with this- it doesn't seem to have the changes I made from the Mael's review of the previous work ? maybe something gone wrong with the merge.
> > I think the main two things were the handler return type and the logic&comment for splitting out the ts & signature https://github.com/guardian/support-service-lambdas/blob/main/handlers/ticket-tailor-webhook/src/index.ts#L7
> > https://github.com/guardian/support-service-lambdas/blob/main/handlers/ticket-tailor-webhook/src/verifySignature.ts#L21-L23
> 
> Thats ok I think I need to merge main and fix any conflicts.

Merged latest main and fixed conflicts. This should make more sense now.

michaelwmcnamara:
This has now been replaced by https://github.com/guardian/support-service-lambdas/pull/2438 - the reason for this is to allow smoother parallel development by building off of the changes in https://github.com/guardian/janus/pull/4306

PR #2435, opened 2024-08-22
Add Id API create user functionality

michaelwmcnamara:
## What does this change?
When a ticket tailor order is placed, this change processes the received request. 
It first queries the ID API to determine if there is an existing user associated with the email in the request. 
If there is no existing user for this email address it then sends a request to the ID API to create a guest user with this email address.

trello card [here](https://trello.com/c/WS4rFm7J/450-ticket-tailor-webhook-idapi-integration).

## How to test
Change was deployed to CODE and 2 Ticket Tailor orders were placed. The first using an email with an existing Identity account and the second with a new email.
It was confirmed in the logs that the account type for the email with the existing account was returned but no request was then made to create a guest account.
It was confirmed that for a request was made to create a guest account for the email address that did not have an existing account.
Integration tests added for local development, however they are not ideal as the IDAPI token we have does not have permissions to delete items, so cleanup must be done manually.
 
All orders from TicketTailor for users who do not have an existing Identity account, will have a guest account created for them.


handlers/ticket-tailor-webhook/src/idapiService.ts
```diff
@@ -0,0 +1,63 @@
+import type { Stage } from '@modules/stage';
+import { getIdApiSecret } from './getSecrets';
+
+export type UserTypeResponse = {
+	userType: string;
+};
+
+const stage = process.env.STAGE as Stage;
+
+const idapiToken = getIdApiSecret(stage);
```
tjmw, line 10: Not a big deal at all but I'd be tempted to avoid caching the `idapiToken` promise and move the call to `getIdApiSecret` down into the places where the token is used. It may result in more calls to secrets manager, but it avoids ever having a confusing situation where the lambda runtime hangs around and the value in secrets manager gets updated but the secret in the promise here is still the old one.

handlers/ticket-tailor-webhook/src/idapiService.ts
```diff
@@ -0,0 +1,63 @@
...
+				'Content-Type': 'application/json',
+				'X-GU-ID-Client-Access-Token': bearerToken,
+			},
+		},
+	).then((response) => {
+		if (!response.ok) {
+			throw new Error(response.statusText);
+		}
+		console.log(`User type response is ${response.type}`);
+		return response.json() as Promise<UserTypeResponse>;
```
tjmw, line 37: Perhaps in future as an optimisation we could validate the response here with [zod](https://github.com/colinhacks/zod) or similar, rather than casting? This would allow us to handle a scenario where the response isn't what we expect more explicitly.

handlers/ticket-tailor-webhook/src/index.ts
```diff
@@ -1,19 +1,26 @@
...
-export const handler: Handler = async (event: SQSEvent): Promise<boolean> => {
+const stage = process.env.STAGE as Stage;
+
+export const handler: Handler = async (event: SQSEvent) => {
 	const res = await event.Records.flatMap(async (record) => {
 		console.log(`Processing TT Webhook. Message id is: ${record.messageId}`);
-		const stage = process.env.STAGE as Stage;
 		const validationSecret = await getWebhookValidationSecret(stage);
 		const matches = hasMatchingSignature(record, validationSecret);
 		if (matches) {
```
tjmw, line 15: Totally subjective but now we're adding more complexity here (user creation) it might make the handler simpler to read if we return/error early if the signature doesn't match as this would remove some nesting/indentation. I.e. invert this logic to:

```js
if (!matches) {
    throw new Error(
        'Signatures do not match - check Ticket Tailor signing secret matches the one stored in AWS.',
    );
}

// ... continue to handling the request
```

What do you reckon?

handlers/ticket-tailor-webhook/test/idapiIntegrationTest.test.ts
```diff
@@ -0,0 +1,26 @@
+/**
+ * Fetches the userType from IDAPI for a given email.
+ * Asks IDAPI to create a guest account if there is no account associated with that email.
+ *
+ * @group integration
+ */
+
+import { createGuestAccount, fetchUserType } from '../src/idapiService';
+
+test('we are able to fetch the user type for a given user from identity', async () => {
```
tjmw, line 10: Are these tests actually hitting idapi? If so it might be nice to stub the HTTP requests using fetch-mock or similar so that we can control the responses.

tjmw (commented):
This is looking really good, nice work! I left a handful of comments and questions, see what you think. I'm really happy to pair if it'd be useful.

cdk/lib/ticket-tailor-webhook.ts
```diff
@@ -155,6 +155,8 @@ export class TicketTailorWebhook extends GuStack {
 						resources: [
 							`arn:aws:secretsmanager:${this.region}:${this.account}:secret:PROD/TicketTailor/Webhook-validation-ECS6P8`,
 							`arn:aws:secretsmanager:${this.region}:${this.account}:secret:CODE/TicketTailor/Webhook-validation-eEsTGW`,
+							`arn:aws:secretsmanager:${this.region}:${this.account}:secret:CODE/TicketTailor/IdApi-token-JBI5VI`,
+							`arn:aws:secretsmanager:${this.region}:${this.account}:secret:PROD/TicketTailor/IdApi-token-eu45YS`,
```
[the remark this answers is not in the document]
Joemgru: that's a cool use of globbing, I'd not considered that, good spot.

handlers/ticket-tailor-webhook/test/idapiIntegrationTest.test.ts
michaelwmcnamara: These are hitting the CODE IDAPI - I agree that iterating on these would be useful and mocking them is something I would like to do as currently they will write entries into the CODE Identity database and I don't believe we have the AWS access to clean these up. 
However, given these are set to be integration tests, run only on request by a local developer, and given the due date for this being live is the 29th August, I would prefer to put this in place for the short term, and improve the tests in a separate PR.

handlers/ticket-tailor-webhook/src/idapiService.ts
michaelwmcnamara: Agree this would be a good iteration on this work, perhaps in a future PR.

MaelGNM (approved):
It looks good to me for now as remaining comments to be addressed in further PRs

PR #2438, opened 2024-08-23
Merge after #2435 - Invalidate webhook if tickettailor-webhook-signature has a timestamp older than 5 minutes

michaelwmcnamara:
This PR replaces: https://github.com/guardian/support-service-lambdas/pull/2434 
The reason for this is to enable it to be built off the changes for https://github.com/guardian/janus/pull/4306 which has higher priority features. This allows for parallel development.

It is essentially the same code albeit with a few more tests and the new function now returns a boolean value to easily fit in with the current structure of index.ts.

Key things to note: 
While the validateRequest() function is currently set to return a boolean, all paths in which the value of that boolean would be false currently throw an error. 
Future work would be to make this a side-effect function (ie return nothing) or to have it return a bool | error and have more explicit handling in the index.ts file.
For now though I am trying to make this change fit in easily with https://github.com/guardian/support-service-lambdas/pull/2435 and have kept modifications of index.ts to a minimum 

## What does this change?

This is a follow-on to https://github.com/guardian/support-service-lambdas/pull/2420 
It implements the second security recommendation in the [ticket-tailor webhook documentation](https://developers.tickettailor.com/#security): 
"We also recommend to invalidate webhooks received where the sent timestamp is older than 5 minutes"

We do this by checking the timestamp in tickettailor-webhook-signature (which is used in validating the signature) and ensuring that it is 300 seconds old or less.

This minimises the window within which a captured request (which will have a valid signature) could be re-used with bad intentions.

Trello card [here:](https://trello.com/c/RoGDmFz2/476-add-5-min-timeout-to-ticket-tailor-webhook-validation).

## How to test

Tests were added to confirm the validation would fail if a request with a valid record and signature but who's timestamp was older than 5 minutes would not be processed.

The Ticket Tailor webhook is more secure and it is harder for a bad actor to spoof requests over a longer timeframe.

The timeout is looking at the timestamp associated with the record signature, this should not have any negative impact on retry attempts as these should be sent as new requests.


Joemgru (approved):

handlers/ticket-tailor-webhook/src/validateRequest.ts
```diff
@@ -0,0 +1,102 @@
...
+	if (!signatureMatches) {
+		throw Error(
+			'Signatures do not match - check Ticket Tailor signing secret matches the one stored in AWS.',
+		);
+	} else if (!withinTimeWindow) {
+		throw Error(
+			`Webhook Signature timestamp ${timestamp} is older than ${maxValidTimeWindowSeconds} seconds. Webhook will not be processed.`,
+		);
+	} else {
+		return true;
```
johnduffell, line 100: why return true when it can never return false?  why not just return void?

PR #2442, opened 2024-08-26
Fix gap in prettier checking/formatting

tjmw:
## What does this change?

In #2436, some files under handlers/zuora-salesforce-link-remover were reformatted by prettier. I was surprised that these formatting violations hadn't been caught by prettier as part of the build. It turns out there were a couple of issues which caused this, which this PR fixes:

* Add the prettier `check-formatting` command to the zuora-salesforce-link-remover `package` script
* Ensure that the prettier command(s) catch all .ts files under each workspace, no matter how they're nested. This wasn't happening on zuora-salesforce-link-remover previously. I went through various iterations and the only pattern I found which
matched _all_ .ts files nested under the workspace was **.ts. I've made this change for all workspaces, though zuora-salesforce-link-remover is the only one where this has a material impact (due to the way this workspace is organized, there's more nesting here than in other places).


johnduffell (approved):
now I can see the logic behind sbt writing the build as code - It's a bit like CDK vs CFN, you can reuse more easily.

PR #2445, opened 2024-08-27
bump simple configuration for snyk reasons

johnduffell:
We did some changes last week to clean up the high severity issues, but it seems like now we are still hitting a netty issue in the build's scan: https://security.snyk.io/vuln/SNYK-JAVA-IONETTY-5953332

I have done some more changes to get a later AWS sdk which will hopefully fix it.  I had mixed results when running snyk test locally.

All unit tests still pass locally.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,7 +1,7 @@
 import Dependencies.*
 
 val scala2Settings = Seq(
-  ThisBuild / scalaVersion := "2.13.12",
+  ThisBuild / scalaVersion := "2.13.14",
   version := "0.0.1",
   organization := "com.gu",
   scalacOptions ++= Seq(
@@ -25,7 +25,7 @@ val scala2Settings = Seq(
 )
 
 val scala3Settings = Seq(
-  scalaVersion := "3.3.1",
+  scalaVersion := "3.3.3",
   version := "0.0.1",
   organization := "com.gu",
   scalacOptions ++= Seq(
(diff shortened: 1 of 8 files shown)


build.sbt
```diff
@@ -1,7 +1,7 @@
 import Dependencies.*
 
 val scala2Settings = Seq(
-  ThisBuild / scalaVersion := "2.13.12",
+  ThisBuild / scalaVersion := "2.13.14",
```
johnduffell, line 4: some of the bumps are just opportunistic

build.sbt
```diff
@@ -247,10 +247,10 @@ lazy val `effects-lambda` = library(project in file("lib/effects-lambda"), Seq(t
 
 lazy val `config-core` = library(project in file("lib/config-core"), Seq(), scala3Settings)
 
-lazy val `config-cats` = library(project in file("lib/config-cats"))
+lazy val `config-cats` = library(project in file("lib/config-cats"), Seq(`effects-s3`))
   .settings(
     libraryDependencies ++= Seq(simpleConfig, catsEffect, circe, circeConfig),
-    dependencyOverrides ++= Seq(nettyCodec) ++ jacksonDependencies,
+    dependencyOverrides ++= jacksonDependencies,
```
johnduffell, line 253: we don't need to override this any more, we are actually getting a later version from AWS sdk for free.

handlers/digital-voucher-api/src/main/scala/com/gu/digital_voucher_api/Handler.scala
```diff
@@ -13,7 +14,10 @@ object Handler
       implicit val contextShift = IO.contextShift(ExecutionContext.global)
       AsyncHttpClientCatsBackend[IO]()
         .flatMap { sttpBackend =>
-          DigitalVoucherApiApp(AppIdentity.whoAmI(defaultAppName = "digital-voucher-api"), sttpBackend).value
+          DigitalVoucherApiApp(
+            AppIdentity.whoAmI(defaultAppName = "digital-voucher-api", aws.CredentialsProvider).get,
```
johnduffell, line 18: now we pass in the credentials provider instead of using the default.
I call .get, because in the code that's just relating to if the ASG is not working, but in our case it will come from environment variables as it's a lambda.

PR #2447, opened 2024-08-28
clean up the return types

Joemgru:
## Accessibility

### DIFF
--- a/handlers/ticket-tailor-webhook/src/idapiService.ts
+++ b/handlers/ticket-tailor-webhook/src/idapiService.ts
@@ -51,7 +51,7 @@ export const createGuestAccount = async (email: string) => {
 	);
 	const bearerToken = `Bearer ${idapiSecret.token}`;
 
-	return await fetch(idapiUrl.concat(guestEndpoint), {
+	 await fetch(idapiUrl.concat(guestEndpoint), {
 		method: 'POST',
 		headers: {
 			'Content-Type': 'application/json',
@@ -65,6 +65,5 @@ export const createGuestAccount = async (email: string) => {
 			throw new Error(response.statusText);
 		}
 		console.log(response);
-		return true;
 	});
 };
--- a/handlers/ticket-tailor-webhook/src/index.ts
+++ b/handlers/ticket-tailor-webhook/src/index.ts
@@ -11,8 +11,8 @@ export type HmacKey = {
 	secret: string;
 };
 
-export const handler: Handler = async (event: SQSEvent) => {
-	return await event.Records.flatMap(async (record) => {
+export const handler: Handler = (event: SQSEvent) => {
+	  event.Records.flatMap(async (record) => {
 		console.log(`Processing TT Webhook. Message id is: ${record.messageId}`);
 		const validationSecret = await getSecretValue<HmacKey>(
 			`${stage}/TicketTailor/Webhook-validation`,
@@ -32,13 +32,14 @@ export const handler: Handler = async (event: SQSEvent) => {
 			const email = payload.payload.buyer_details.email;
 			const userTypeResponse = await fetchUserType(email);
 			if (userTypeResponse.userType === 'new') {
-				return await createGuestAccount(email);
+				  createGuestAccount(email).catch(() => {throw new Error("Error creating guest account");});
+			
 			} else {
 				console.log(
 					`Skipping guest creation as user of type ${userTypeResponse.userType} exists already`,
 				);
-				return userTypeResponse;
+				 userTypeResponse;
 			}
 		}
-	}).at(0);
+	})
 };


handlers/ticket-tailor-webhook/src/index.ts
```diff
@@ -11,8 +11,8 @@ export type HmacKey = {
 	secret: string;
 };
 
-export const handler: Handler = async (event: SQSEvent) => {
-	return await event.Records.flatMap(async (record) => {
+export const handler: Handler = (event: SQSEvent) => {
```
Joemgru, line 14: does it matter if the handler doesn't return anything?

handlers/ticket-tailor-webhook/src/index.ts
```diff
@@ -32,13 +32,14 @@ export const handler: Handler = async (event: SQSEvent) => {
...
+					throw new Error('Error creating guest account');
+				});
 			} else {
 				console.log(
 					`Skipping guest creation as user of type ${userTypeResponse.userType} exists already`,
 				);
-				return userTypeResponse;
 			}
 		}
-	}).at(0);
```
Joemgru, line 43: in theory will only ever have one record but still feels better not to have hardcoded array access

handlers/ticket-tailor-webhook/src/index.ts
```diff
@@ -11,8 +11,8 @@ export type HmacKey = {
 	secret: string;
 };
 
-export const handler: Handler = async (event: SQSEvent) => {
-	return await event.Records.flatMap(async (record) => {
+export const handler: Handler = (event: SQSEvent) => {
+	event.Records.flatMap(async (record) => {
```
tjmw, line 15: Since there's async work going on downstream of this `flatMap` I _think_ we should be waiting for this to complete before returning from the handler. I don't believe that's happening right now since we're not returning a promise which represents completion of this work, but I think we should be. Happy to look together at how best to do that if it'd be helpful.

michaelwmcnamara: Resolved issues and questions in mobbing session.

handlers/ticket-tailor-webhook/src/idapiService.ts
```diff
@@ -34,37 +38,37 @@ export const fetchUserType = async (email: string) => {
...
-		if (!response.ok) {
-			throw new Error(response.statusText);
-		}
-		console.log(response);
-		return true;
 	});
+	console.log(`Create Guest Account response status: ${response.statusText}`);
+	if (!response.ok) {
+		throw new Error(
+			`Guest account creation for email: ${email} with status ${response.statusText}. Response body is: ${JSON.stringify(response.body)}`,
```
Joemgru, line 70: Do we want to log email ? i tend to try and avoid logging PII just to be on the safe side of data compliance laws

michaelwmcnamara: Normally I would avoid, however I think its essential for debugging in this case. We do have a very strict retention period on log groups in this account - to ensure GDPR compliance. Personally I think this is a necessary use of the data.

handlers/ticket-tailor-webhook/src/idapiService.ts
```diff
@@ -34,37 +36,39 @@ export const fetchUserType = async (email: string) => {
...
-		if (!response.ok) {
-			throw new Error(response.statusText);
-		}
-		console.log(response);
-		return true;
 	});
+	console.log(`Create Guest Account response status: ${response.statusText}`);
+	if (!response.ok) {
+		throw new Error(
+			`Guest account creation for email: ${email} failed with status ${response.statusText} }`,
```
MaelGNM, line 71: Do we get more information about the error in the body of the response that would be useful to include somewhere?

handlers/ticket-tailor-webhook/src/index.ts
```diff
@@ -24,21 +24,26 @@ export const handler: Handler = async (event: SQSEvent) => {
 			currentDateTime,
 		);
 		if (!validRequest) {
-			throw new Error(
-				'Signatures do not match - check Ticket Tailor signing secret matches the one stored in AWS.',
-			);
+			console.error('Request failed validation. Processing terminated.');
 		} else {
 			const payload = JSON.parse(record.body) as Payload;
```
johnduffell, line 29: it's quite hard to follow as we have a lot of very generic names that it's not clear what the difference is - `results` `payload` `Payload` `record`

perhaps `results` should be `promises`
`record` should be `apiGatewayRequest`?
`Payload` and `payload` should be ? `ticketTailorHttpBody`?

handlers/ticket-tailor-webhook/src/index.ts
```diff
@@ -24,21 +24,26 @@ export const handler: Handler = async (event: SQSEvent) => {
 			currentDateTime,
 		);
 		if (!validRequest) {
-			throw new Error(
-				'Signatures do not match - check Ticket Tailor signing secret matches the one stored in AWS.',
-			);
+			console.error('Request failed validation. Processing terminated.');
```
johnduffell, line 27: I'm not 100% sure on this log and continue style, it does feel more functional and presumably we deicded we want the item to still be removed from the queue?  Rather than building up in the DLQ?  I'm not sure I 100% agree, we would want to know asap if the signatures are not being validated generally?
If so pull out the "else" into a separate function to make it clear that everything after the else will run regardless.  Otherwise if someone later adds some logic afterwards, it would happen even on an invalid signature..
```suggestion
			console.error('Request failed validation. Skipping over with "success" to remove it from the queue.');
```

handlers/ticket-tailor-webhook/src/index.ts
```diff
@@ -24,21 +24,26 @@ export const handler: Handler = async (event: SQSEvent) => {
...
-					`Skipping guest creation as user of type ${userTypeResponse.userType} exists already`,
+					`Skipping guest creation as account of type ${userTypeResponse.userType} already exists for user.`,
 				);
-				return userTypeResponse;
 			}
 		}
-	}).at(0);
+	});
+
+	await Promise.all<void>(results);
```
johnduffell, line 48: presume this will throw if any of the promises failed, meaning the lambda will fail and not remove the item from the queue?  I assume if we return successfully it will delete the SQS event(s?) from the queue.

handlers/ticket-tailor-webhook/src/idapiService.ts
michaelwmcnamara: On reflection, we might be able to get all we need using timestamps. Lets remove email address for now and add it back if we find its too hard to debug without it.

handlers/ticket-tailor-webhook/src/index.ts
michaelwmcnamara: This is a good point. I think its worth renaming things to be more readable. I wont remember any of this in 6 months so even for us, clarity will be golden.

Joemgru: I've had a go at making it more readable, although record is an apiGatewayRequest at one point it get's transformed further upstream so I've lengthened it out to sqsRecord so there's less dependency on editors type checker when reading.

handlers/ticket-tailor-webhook/src/index.ts
michaelwmcnamara: Resolved this in person and implemented changes to this effect.

handlers/ticket-tailor-webhook/src/validateRequest.ts
```diff
@@ -1,41 +1,36 @@
...
-		throw new Error(
+
+	if (!(typeof signatureWithTs === 'string')) {
+		console.error(
 			'No valid value found for MessgeAttritbute: tickettailor-webhook-signature on incoming request.',
 		);
+		return;
 	}
+
+	const timestamp = signatureWithTs.split(',')[0]?.split('t=')[1];
```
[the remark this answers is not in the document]
johnduffell: good point, although I'm not generally in favour of regex because it's hard to read
I'd be tempted to split on `,` and then split each element on `=`, then put the results into an object, then we can just do `result.t` and `result.v1`
How to actually write that - I could do it in scala off hand but not so much in typescript!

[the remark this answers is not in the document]
johnduffell: something like this?
```
const signatureTsParsed = Object.fromEntries(signatureWithTs.split(',').map((r) => r.split('=')))
const timestamp = signatureTsParsed.t;
const signature = signatureTsParsed.v1;
```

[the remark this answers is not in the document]
michaelwmcnamara: Worth looking into in future PR - will need to read up on how to get around the error TS2531: Object is possibly 'null' error first though.

PR #2449, opened 2024-08-28
add RC discount and reduce data duplication

johnduffell:
This PR is part of the cancellation saves OKR.

At the moment, recurring contribtuiors who are cancelling, only get offered to change their amount.

This backend work allows them (once the frontend is done) to also get offered two months free/pause.

Tested in CODE and working fine with a recurring contribtion.

I wasn't sure how much it was worth adding unit (and especiially integration) tests given it's a simplified version of S+, but  opinons welcome.

### DIFF
--- a/handlers/discount-api/src/discountEndpoint.ts
+++ b/handlers/discount-api/src/discountEndpoint.ts
@@ -23,7 +23,7 @@ import type { APIGatewayProxyEventHeaders } from 'aws-lambda';
 import dayjs from 'dayjs';
 import { EligibilityChecker } from './eligibilityChecker';
 import { generateCancellationDiscountConfirmationEmail } from './generateCancellationDiscountConfirmationEmail';
-import { productToDiscountMapping } from './productToDiscountMapping';
+import { getDiscountFromSubscription } from './productToDiscountMapping';
 
 export const previewDiscountEndpoint = async (
 	stage: Stage,
@@ -138,14 +138,17 @@ export const applyDiscountEndpoint = async (
 	);
 
 	const emailPayload = discount.sendEmail
-		? generateCancellationDiscountConfirmationEmail({
-				firstDiscountedPaymentDate: dayjs(dateToApply),
-				nextNonDiscountedPaymentDate: dayjs(nextPaymentDate),
-				emailAddress: account.billToContact.workEmail,
-				firstName: account.billToContact.firstName,
-				lastName: account.billToContact.lastName,
-				identityId: account.basicInfo.identityId,
-			})
+		? generateCancellationDiscountConfirmationEmail(
+				{
+					firstDiscountedPaymentDate: dayjs(dateToApply),
+					nextNonDiscountedPaymentDate: dayjs(nextPaymentDate),
+					emailAddress: account.billToContact.workEmail,
+					firstName: account.billToContact.firstName,
+					lastName: account.billToContact.lastName,
+					identityId: account.basicInfo.identityId,
+				},
+				discount.sendEmail,
+			)
 		: undefined;
 
 	return {
@@ -204,7 +207,7 @@ async function getDiscountToApply(
 
 	console.log('Working out the appropriate discount for the subscription');
 	const { discount, discountableProductRatePlanId } =
-		productToDiscountMapping(stage).getDiscountFromSubscription(subscription);
+		getDiscountFromSubscription(stage, subscription);
 
 	console.log('Checking this subscription is eligible for the discount');
 	switch (discount.eligibilityCheckForRatePlan) {
--- a/handlers/discount-api/src/discountEndpoint.ts
+++ b/handlers/discount-api/src/discountEndpoint.ts
@@ -23,7 +23,7 @@ import type { APIGatewayProxyEventHeaders } from 'aws-lambda';
 import dayjs from 'dayjs';
 import { EligibilityChecker } from './eligibilityChecker';
 import { generateCancellationDiscountConfirmationEmail } from './generateCancellationDiscountConfirmationEmail';
-import { getDiscountFromSubscription } from './productToDiscountMapping';
+import { productToDiscountMapping } from './productToDiscountMapping';
 
 export const previewDiscountEndpoint = async (
 	stage: Stage,
@@ -204,7 +204,7 @@ async function getDiscountToApply(
 
 	console.log('Working out the appropriate discount for the subscription');
 	const { discount, discountableProductRatePlanId } =
-		getDiscountFromSubscription(stage, subscription);
+		productToDiscountMapping(stage).getDiscountFromSubscription(subscription);
 
 	console.log('Checking this subscription is eligible for the discount');
 	switch (discount.eligibilityCheckForRatePlan) {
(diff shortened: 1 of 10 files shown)


handlers/discount-api/src/discountEndpoint.ts
```diff
@@ -223,6 +226,8 @@ async function getDiscountToApply(
 				account.metrics.currency,
 			);
 			break;
+		case 'NoCheck':
+			break;
```
johnduffell, line 230: there are no special checks for RC, you can pause as many times as you like

handlers/discount-api/src/productToDiscountMapping.ts
```diff
@@ -63,83 +71,77 @@ export const catalog = {
...
-			upToPeriods: 12,
-			upToPeriodsType: 'Months',
-			sendEmail: false,
-			eligibilityCheckForRatePlan: 'AtCatalogPrice',
-		},
-		cancellationFree2Mo: {
-			productRatePlanId: '8ad081dd8fd3d9df018fe2b6a7bc379d',
+const Discounts: (stage: Stage) => { [K in string]: Discount } = (
+	stage: Stage,
+) => {
```
johnduffell, line 82: this now takes stage as a parameter rather than as the top level key.  This reduces repetition so means we can guarantee that the promo details are the same in CODE and PROD, but still vary the ID.

handlers/discount-api/src/productToDiscountMapping.ts
```diff
@@ -63,83 +71,77 @@ export const catalog = {
...
-		[catalog.PROD.supporterPlus.Month]: Discounts.PROD.cancellationFree2Mo,
-	},
-	PROD: {
-		[catalog.PROD.digiSub.Month]: Discounts.PROD.cancellation25pc3mo,
-		[catalog.PROD.digiSub.Quarter]: Discounts.PROD.cancellation25pc3mo,
-		[catalog.PROD.digiSub.Annual]: Discounts.PROD.cancellation25pc12mo,
-		[catalog.PROD.supporterPlus.Month]: Discounts.PROD.cancellationFree2Mo,
-	},
-};
+function ProductToDiscountMapping(stage: Stage) {
```
johnduffell, line 134: same here - stage is a parameter so we can be sure the mappings don't vary between CODE and PROD

modules/stage.ts
```diff
@@ -1,4 +1,4 @@
-export type Stage = 'CODE' | 'CSBX' | 'PROD';
+export type Stage = 'CODE' | 'PROD';
```
johnduffell, line 1: I don't think CSBX is used - Rupert mentioned previously that it was more of an experiment.

handlers/discount-api/src/generateCancellationDiscountConfirmationEmail.ts
```diff
@@ -11,14 +13,17 @@ export type EmailFields = {
...
+export const generateCancellationDiscountConfirmationEmail = (
+	{
+		firstDiscountedPaymentDate,
+		nextNonDiscountedPaymentDate,
+		emailAddress,
+		firstName,
+		lastName,
+		identityId,
+	}: EmailFields,
+	dataExtensionName: DataExtensionName,
```
johnduffell, line 25: I wasn't really sure the difference between having the EmailFields thing or just passing normal parameters, I guess the former is like a neat syntax for pattern matching a structure if we need to reuse.  However in this case this is literally the only place we use it!

handlers/discount-api/src/productToDiscountMapping.ts
```diff
@@ -1,17 +1,18 @@
...
 import { ValidationError } from '@modules/errors';
 import type { Stage } from '@modules/stage';
 import { isNotRemovedOrDiscount } from '@modules/zuora/rateplan';
 import type { ZuoraSubscription } from '@modules/zuora/zuoraSchemas';
 
-export function getDiscountableRatePlan(subscription: ZuoraSubscription) {
+function getDiscountableRatePlan(subscription: ZuoraSubscription) {
 	return getSingleOrThrow(
-		subscription.ratePlans,
+		subscription.ratePlans.filter(isNotRemovedOrDiscount),
```
johnduffell, line 11: it seemed more confusing to have a separate filter parameter for getSingleOrThrow than just to use the normal filter before hand so I simplified it

handlers/discount-api/src/generateCancellationDiscountConfirmationEmail.ts
Joemgru: it's a neat way of unpacking everything in a typesafe way but maybe easier to pass `emailObj: EmailFields` and then do dot access in the function ? - would seem more to idiomatic to me anyway!

Joemgru (approved):

modules/arrayFunctions.ts
```diff
@@ -30,19 +30,17 @@ export const sortBy = <T>(array: T[], fn: (item: T) => string): T[] => {
 export const getSingleOrThrow = <T>(
 	array: T[],
 	error: (msg: string) => Error,
-	filter?: (element: T) => boolean,
```
Joemgru, line 33: good spot

handlers/discount-api/src/generateCancellationDiscountConfirmationEmail.ts
tjmw: Yeah I think I'd slightly lean towards Joe's suggestion to split this up (my brain finds it hard to unpack _too_ much going on in single place) but I don't feel too strongly!

handlers/discount-api/src/productToDiscountMapping.ts
```diff
@@ -33,13 +34,17 @@ export const getDiscountFromSubscription = (
...
+	| 'AtCatalogPrice'
+	| 'NoCheck';
+
 export type Discount = {
 	productRatePlanId: string;
 	name: string;
 	upToPeriods: number;
 	upToPeriodsType: string;
-	sendEmail: boolean;
+	sendEmail?: DataExtensionName;
```
tjmw, line 47: I wonder if something like `emailIdentifier` would be clearer here? At a glance I think I'd expect `sendEmail` to be a boolean.

tjmw (approved):
Some nice refactorings along the way 👍

handlers/discount-api/src/generateCancellationDiscountConfirmationEmail.ts
johnduffell: will do this in another PR

handlers/discount-api/src/productToDiscountMapping.ts
johnduffell: will do this in another PR

handlers/discount-api/src/generateCancellationDiscountConfirmationEmail.ts
johnduffell: https://github.com/guardian/support-service-lambdas/pull/2476

handlers/discount-api/src/productToDiscountMapping.ts
johnduffell: https://github.com/guardian/support-service-lambdas/pull/2476

PR #2450, opened 2024-08-29
Clarify amount validation message for recurring contributions in new-product-api

graham228221:
## What does this change?
Within new-product-api, brings the amount validation of Recurring Contributions into line with the newer Supporter Plus product.

This is in response to a bug report, where a user was confused why the error they got advised them that a recurring contribution must be at least AU$1000. The error message was based on minor units, and this updates it to use major units along with stating the currency.

Also updates some of the minimum amounts to match the current product pricing grid from MRR. 

Before:
Screenshot 2024-08-29 at 15 42 25

After:
Screenshot 2024-08-29 at 16 58 31

I've also updated the way the min/max amounts are stored to be more in line with Supporter Plus (I do think it's clearer).

## How to test
On a Billing Account in Salesforce, click "New Subscription" and setup a new Recurring Contribution. Set the amount above or below the minimum prices defined in handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/validation/AmountLimits.scala

CSRs are less likely to be confused by this error message.

Have tested in CODE and we verify in production after deployment.

### DIFF
--- a/handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/validation/AmountLimits.scala
+++ b/handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/validation/AmountLimits.scala
@@ -15,33 +15,33 @@ object AmountLimits {
   def fromMinorToMajor(value: Int) = value / 100
 
   val gbp = ContributionLimits(
-    monthly = AmountLimits.limitsFromMajorToMinorUnits(min = 2, max = 166),
-    annual = AmountLimits.limitsFromMajorToMinorUnits(min = 10, max = 2000),
+    monthly = AmountLimits.limitsFromMajorToMinorUnits(min = 4, max = 166),
+    annual = AmountLimits.limitsFromMajorToMinorUnits(min = 50, max = 2000),
   )
 
   val aud = ContributionLimits(
     monthly = AmountLimits.limitsFromMajorToMinorUnits(min = 10, max = 200),
-    annual = AmountLimits.limitsFromMajorToMinorUnits(min = 10, max = 2000),
+    annual = AmountLimits.limitsFromMajorToMinorUnits(min = 80, max = 2000),
   )
 
   val usd = ContributionLimits(
-    monthly = AmountLimits.limitsFromMajorToMinorUnits(min = 2, max = 166),
-    annual = AmountLimits.limitsFromMajorToMinorUnits(min = 10, max = 2000),
+    monthly = AmountLimits.limitsFromMajorToMinorUnits(min = 3, max = 166),
+    annual = AmountLimits.limitsFromMajorToMinorUnits(min = 30, max = 2000),
   )
 
   val nzd = ContributionLimits(
     monthly = AmountLimits.limitsFromMajorToMinorUnits(min = 10, max = 200),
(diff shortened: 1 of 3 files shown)


graham228221:
I'm awaiting confirmation from MRR colleagues on whether the actual min/max amounts for recurring contributions need updating too.

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/validation/AmountLimits.scala
```diff
@@ -11,34 +11,37 @@ case class AmountLimits(min: Int, max: Int)
 
 object AmountLimits {
 
+  def limitsFromMajorToMinorUnits(min: Int, max: Int) = AmountLimits(min * 100, max * 100)
```
johnduffell, line 14:
```suggestion
  def fromMajorUnits(min: Int, max: Int) = AmountLimits(min * 100, max * 100)
```

johnduffell (approved):
sounds like you are bringing this in line with S+ so it makes sense for it to be consistent.  Plus the error message is much better!

JoeMitchellGuardian:
This PR also ties in slightly to this card on the platform board https://trello.com/c/jM31KSne/170-return-more-useful-error-messages-from-new-product-api - although on the card it's a slightly different error scenario!

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/validation/contribution/ContributionValidations.scala
```diff
@@ -25,7 +25,8 @@ object ContributionValidations {
       amount <- validatableFields.amountMinorUnits getOrFailWith s"amountMinorUnits is missing"
       _ <- isValidStartDate(validatableFields.startDate)
       limits = limitsFor(planId, currency)
-      _ <- (amount.value <= limits.max) orFailWith s"amount must not be more than ${limits.max}"
-      _ <- (amount.value >= limits.min) orFailWith s"amount must be at least ${limits.min}"
+      _ <- (amount.value <= limits.max) orFailWith s"amount must not be more than $currency ${AmountLimits
+          .fromMinorToMajor(limits.max)}"
+      _ <- (amount.value >= limits.min) orFailWith s"amount must be at least $currency ${AmountLimits.fromMinorToMajor(limits.min)}"
```
johnduffell, line 30: it looks like `PlanId` is human readable, so it wouldn't be a problem to also include that in the message also

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/validation/AmountLimits.scala
graham228221: Thanks John! https://github.com/guardian/support-service-lambdas/pull/2450/commits/f4c8b82da78ecae37654268a25eb39a240e00bd4

graham228221:
> This PR also ties in slightly to this card on the platform board https://trello.com/c/jM31KSne/170-return-more-useful-error-messages-from-new-product-api - although on the card it's a slightly different error scenario!

Yes wrong error message unfortunately, but maybe we can pick this up soon as it's still an issue for our CSRs

PR #2451, opened 2024-08-30
Use zod to validate response from identity-api for checking if account already exists

michaelwmcnamara:
## What does this change?

Adds zod schema validation to check that the IDAPI response for the Get User Type request contains the expected `userType: string` field and both logs a message and throws an error if the field is missing from this response.

## How to test

- [x] unit tests run correctly.
- [x] Tested on CODE that an order from an existing identity user does not create a guest account.
- [x] Tested on CODE that an order from a new user will result in a request to IDAPI to create a guest account.

If we get a bad response from IDAPI for the Get User Type request, we will now detect and handle it properly.

Minimal functional change. This change reduces existing risks.

### DIFF
--- a/handlers/ticket-tailor-webhook/package.json
+++ b/handlers/ticket-tailor-webhook/package.json
@@ -11,7 +11,8 @@
     "fix-formatting": "prettier --write **.ts"
   },
   "dependencies": {
-    "@aws-sdk/client-secrets-manager": "^3.629.0"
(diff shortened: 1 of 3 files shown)


tjmw (approved):
This LGTM! Might be worth getting another ✅ from someone else since we paired on it.

MaelGNM (approved):
Nice!

PR #2453, opened 2024-09-02
Experiment to parallelise part of the holiday stops processing

Joemgru:
## Accessibility

### DIFF
--- a/handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/Handler.scala
+++ b/handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/Handler.scala
@@ -56,7 +56,7 @@ object Handler extends Lambda[Option[ProductTypeAndStopDate], List[ZuoraHolidayC
             GetFromS3.fetchString,
           ),
         )
-        _ <- RIO.foreach_(results)(result => RIO.effect(ProcessResult.log(result)))
+        _ <- RIO.foreachPar_(results)(result => RIO.effect(ProcessResult.log(result)))
         zuoraWriteResults <- results.flatMap(_.overallFailure.toList) match {
           case Nil =>
             val (_, successfulZuoraResponses) = results.flatMap(_.creditResults).separate


handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/Handler.scala
```diff
@@ -56,7 +56,7 @@ object Handler extends Lambda[Option[ProductTypeAndStopDate], List[ZuoraHolidayC
             GetFromS3.fetchString,
           ),
         )
-        _ <- RIO.foreach_(results)(result => RIO.effect(ProcessResult.log(result)))
+        _ <- RIO.foreachPar_(results)(result => RIO.effect(ProcessResult.log(result)))
```
Joemgru, line 59: worth a shot @johnduffell ?

johnduffell: how many would that run in parallel? we have to be careful about zuora concurrent request limits, as they are shared between other processes as well.  I would assume we are doing 1 at the moment, so 2 would be a good improvement or maybe 3 or 4 if we feel like it.
https://developer.zuora.com/docs/guides/rate-limits/#concurrent-request-limits

It would also be good to check the logs and make sure it's clear what's going on (or at least no less clear than before - often our logging isn't great in the first place)

johnduffell: actually having looked again, that's just doing some logging in parallel which wouldn't really help 
https://github.com/guardian/support-service-lambdas/blob/81a678ef1c7e8e284f4aecf1b42dfdeb802949f6/lib/credit-processor/src/main/scala/com/gu/creditprocessor/ProcessResult.scala#L14-L23

Joemgru: Good spot

johnduffell:
running the 6 batches in parallel seems like a reasonable experiment 👍

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/HolidayStopCreditProcessor.scala
```diff
@@ -40,9 +40,9 @@ object HolidayStopCreditProcessor {
     val allProcessableProductTypes = List(
       NewspaperNationalDelivery,
       NewspaperHomeDelivery,
+      NewspaperVoucherBook,
```
johnduffell, line 43: why move voucher book above weekly?  If there is an issue with weekly it might affect fulfilment, however voucher book is only a credit, so it's less critical.

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/HolidayStopCreditProcessor.scala
```diff
@@ -60,41 +60,39 @@ object HolidayStopCreditProcessor {
...
-                Salesforce.holidayStopRequests(config.sfConfig),
-                fulfilmentDatesFetcher,
-                productTypeAndStopDateOverride.map(_.stopDate),
-                productType,
-                updateToApply,
-                ZuoraHolidayCreditAddResult.apply,
-                Salesforce.holidayStopUpdateResponse(config.sfConfig),
-                Zuora.accountGetResponse(config.zuoraConfig, zuoraAccessToken, backend),
-                NextInvoiceDate.getNextInvoiceDate,
+        productTypesToProcess.map { productType =>
```
johnduffell, line 63: could you flag the actual change here please as I'm struggling to see the difference?

handlers/holiday-stop-processor/src/test/scala/com/gu/holidaystopprocessor/ProcessorErrorHandlingSpec.scala
```diff
@@ -172,8 +172,6 @@ class ProcessorErrorHandlingSpec extends AnyFlatSpec with Matchers with OptionVa
     val (failedZuoraResponses, successfulZuoraResponses) = result.creditResults.separate
     failedZuoraResponses.size shouldBe 3
     successfulZuoraResponses.size shouldBe 0
-    (failedZuoraResponses.size + successfulZuoraResponses.size) shouldBe 3
-    result.overallFailure.value shouldBe OverallFailure("zuora boom 1")
```
johnduffell, line 176: why is this removed?

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/Salesforce.scala
```diff
@@ -12,16 +12,8 @@ import com.typesafe.scalalogging.LazyLogging
...
-   * Zuora takes 3-4s to add each credit to a sub so this should take ~ 10 mins,
-   * which is within the 15 min max that a lambda can run.
-   * As the lambda runs each hour, all should easily be processed within the
-   * 24 hour window available.
-   *
-   * Beware! This isn't a scalable solution.  If all the products between them
-   * have > 5000 credit requests to be processed in a 24 hour window some will be missed.
-   */
-  private val batchSize = 66
+  //TODO: Move the batching into a step function/state machine
```
johnduffell, line 15: Would be nice to update the comment with some justification for the batch size, I'm not sure what to put though.. I've made something up in the suggestion
```suggestion
  // we choose 800 because we run them in parallel across all products and any that are not completed would be picked up after timeout, we don't need to ensure we complete all within the lambda timeout any more.
```

lib/credit-processor/src/main/scala/com/gu/creditprocessor/ProcessResult.scala
```diff
@@ -13,10 +13,9 @@ case class ProcessResult[ResultType <: ZuoraCreditAddResult](
 object ProcessResult extends LazyLogging {
   def log[ResultType <: ZuoraCreditAddResult](processResult: ProcessResult[ResultType]): Unit = {
     import processResult._
-    logger.info(s"${creditsToApply.size} credits to apply:")
```
johnduffell, line 16: do we want to keep the logging but mention which product it is?

lib/credit-processor/src/main/scala/com/gu/creditprocessor/Processor.scala
```diff
@@ -106,17 +108,37 @@ object Processor {
...
-            updateToApply,
-            updateSubscription,
-            resultOfZuoraCreditAdd,
-            getNextInvoiceDate,
-          ),
-        )
+
+        val bySub =
+          creditRequests
+            .groupBy(_.subscriptionName)
```
johnduffell, line 114:
```suggestion
            .groupBy(_.subscriptionName) // we need to group by sub name to avoid concurrent locking exceptions
```

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/HolidayStopCreditProcessor.scala
Joemgru: I think it's a whitespace/newline thing 

try with this:

https://github.com/guardian/support-service-lambdas/pull/2453/files/df4d6bb7bda4ee82821b1aab237b39586636a71c?diff=split&w=1

lib/credit-processor/src/main/scala/com/gu/creditprocessor/Processor.scala
```diff
@@ -106,17 +108,37 @@ object Processor {
...
+              creditProduct,
+              getSubscription,
+              getAccount,
+              updateToApply,
+              updateSubscription,
+              resultOfZuoraCreditAdd,
+              getNextInvoiceDate,
+            ),
+          )
+        }.toList
```
johnduffell, line 138: not sure if we can change to for comprehension without making it do the contents of each batch in parallel?
also a few naming update suggestions
```suggestion
        val allZuoraCreditResponses = (for {
          batch <- creditRequestBatches
          request <- batch
        } yield addCreditToSubscription(
            creditProduct,
            getSubscription,
            getAccount,
            updateToApply,
            updateSubscription,
            resultOfZuoraCreditAdd,
            getNextInvoiceDate,
          )(request)
        ).toList
```

lib/credit-processor/src/main/scala/com/gu/creditprocessor/Processor.scala
```diff
@@ -106,17 +108,37 @@ object Processor {
...
-            getSubscription,
-            getAccount,
-            updateToApply,
-            updateSubscription,
-            resultOfZuoraCreditAdd,
-            getNextInvoiceDate,
-          ),
-        )
+
+        val bySub =
```
johnduffell, line 112:
```suggestion
        val creditRequestBatches =
```

lib/zuora-core/src/main/scala/com/gu/zuora/Zuora.scala
```diff
@@ -68,6 +69,7 @@ object Zuora {
     val errMsg = (reason: String) =>
       s"Failed to update subscription '${subscription.subscriptionNumber}' with $update. Reason: $reason"
     basicRequest
+      .readTimeout(2.minutes)
       .put(uri"${config.baseUrl}/subscriptions/${subscription.subscriptionNumber}")
```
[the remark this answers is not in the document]
johnduffell: what do you mean by in line with?  it looks like we are not yet using orders api

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/HolidayStopCreditProcessor.scala
Joemgru: I changed it to be no longer parallel by product. 
The re-ordering was based on volume, putting the larger ones first, but I will put it back to how it was, doing the more critical stuff first makes sense.

handlers/holiday-stop-processor/src/test/scala/com/gu/holidaystopprocessor/ProcessorErrorHandlingSpec.scala
Joemgru: line 175 is checking that line 173 and 174 add up...so is redundant.

line 176 is no longer deterministic a "overallFailure" takes the first failure value, but the zuora requests happen in parallel so it could be any of the three reasons - I'll update it to reflect that.

lib/credit-processor/src/main/scala/com/gu/creditprocessor/ProcessResult.scala
Joemgru: I found this log line misleading as it happens at the end of the batch and would print `0 credits to apply:` and then look like it was about to do something before finishing.

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/Salesforce.scala
```diff
@@ -12,8 +12,9 @@ import com.typesafe.scalalogging.LazyLogging
 
 object Salesforce extends LazyLogging {
 
-  //TODO: Move the batching into a step function/state machine
-  private val batchSize = 800
+  // TODO: Move the batching into a step function/state machine
```
johnduffell, line 15: I'm not sure if this line adds much.  It would be more useful to put a link to the doc that you wrote

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/Salesforce.scala
```diff
@@ -12,8 +12,9 @@ import com.typesafe.scalalogging.LazyLogging
...
+  // For now we seem to handle ~800 per invocation and keeping the batching a bit lower ensures some progress is saved
```
johnduffell, line 16: this batch size is per product, so this could be anywhere between 700 in total and  6*700=4200 in total.
With the batches being done in parallel rather than in sequence, all the writebacks to salesforce would be at the end rather than interleaved like they used to be.
If we are saying it can handle about 800 then shouldn't we make the batch size 100?  Or do you mean when the batch size is 800 we can handle the typical distribution of products?

lib/credit-processor/src/main/scala/com/gu/creditprocessor/Processor.scala
```diff
@@ -109,11 +109,15 @@ object Processor {
...
-        val bySub =
+        // we group the creditRequests by subscription to make the requests to zuora in parallel
+        // & avoid lock contention on the resource
+
+        val creditRequestBatches =
           creditRequests
             .groupBy(_.subscriptionName)
             .values
             .toList
+            .flatten
```
johnduffell, line 120: if we groupBy and then flatten then we're back where we started?

handlers/holiday-stop-processor/src/main/scala/com/gu/holidaystopprocessor/Salesforce.scala
Joemgru: it doesn't parallelise by product - when I changed it to parallelise by subscription I removed the parallelise by product as I didn't want the thread count to be distributed and multiplicative of (product threads x subscription threads) - I thought keeping both would obfuscate the amount of concurrent requests we send to zuora.

handlers/delivery-problem-credit-processor/src/main/scala/com/gu/deliveryproblemcreditprocessor/DeliveryCreditProcessor.scala
```diff
@@ -74,14 +82,18 @@ object DeliveryCreditProcessor extends Logging {
...
-        case None => Task.succeed(())
-        case Some(e) => Task.fail(new RuntimeException(e.reason))
+      _ <- Task.effect(ProcessResult.log(processResults))
+      _ <- Task.effect(processResults).map { pr =>
+        pr.map {
+          _.overallFailure match {
+            case None => Task.succeed(())
+            case Some(e) => Task.fail(new RuntimeException(e.reason))
+          }
+        }
```
johnduffell, line 94: I'm not sure this will do what you expect?
the type would be Task[List[Task[Unit]]] wouldn't it? but the inner Task might be failed, and previously we would have short circuited the for comp, but now we are discarding the value.
I think we might need something more akin to 
```suggestion
      pr <- Task.effect(processResults)
      _ <-  pr.traverse {
          _.overallFailure match {
            case None => Task.succeed(())
            case Some(e) => Task.fail(new RuntimeException(e.reason))
          }
```

handlers/delivery-problem-credit-processor/src/main/scala/com/gu/deliveryproblemcreditprocessor/DeliveryCreditProcessor.scala
```diff
@@ -234,7 +246,7 @@ object DeliveryCreditProcessor extends Logging {
             .leftMap { e =>
               SalesforceApiFailure(e.message)
             }
```
johnduffell, line 248: again by moving this into a map instead of flatMap I think the content (which we don't give any type when we return it) will be discarded.  If we are intending this to fail the overall operation, it won't actually do that as it will be discarded.

lib/credit-processor/src/main/scala/com/gu/creditprocessor/Processor.scala
```diff
@@ -100,39 +102,80 @@ object Processor {
...
+                  .leftMap { failure =>
+                    ProcessResult(
+                      creditRequests,
+                      List(zuoraApiResponse),
+                      List.empty,
+                      Some(OverallFailure(failure.reason)),
+                    )
+                  }
+                sfResult
+              }
```
johnduffell, line 168: could we pull this out into a separate function please
I guess it's combining the separate logic we had to call zuora and then salesforce, but I can't quite understand it when it's so deeply nested.

handlers/delivery-problem-credit-processor/src/main/scala/com/gu/deliveryproblemcreditprocessor/DeliveryCreditProcessor.scala
Joemgru: I think we'll have to pair on this i dont understand Zio enough to get that to work

johnduffell (approved):
good this is definitely performing a lot better, let's hope it stays that way! 👏 thanks for working through all that.

PR #2455, opened 2024-09-03
Bug Fix: Handle situation where missing return amount because at end of payment period

michaelwmcnamara:
This is a bug fix. Associated trello card is [here](https://trello.com/c/4WpazHGb/463-support-fix-unhandled-case-in-product-switch-api).

## What does this change?

This changes contributionsToSupporterPlus.ts and adds a check to handle the situation where the payment-switch is made on the same day of month as the term-start-date.
In this situation the previous payment term has completed but the payments for the next term have not yet been taken.
Consequently there is no payment to refund.

In this situation, the zuoraPreviewResponse invoice will not have an invoiceItem with the refund amount.
Currently this throws an error. It should instead return 0 as there is nothing to refund.

## The specifics of the change:
If there is no invoice item found for the recurring payments charge id, this change will check the term-start-date against the current date. If they have the same day of month, it will return 0 instead of throwing an error.

This change involved adding the zuoraSubscription as a parameter to the previewResponseFromZuoraResponse method, so that we can extract the term-start-date.

## How to test

- [x] New unit tests created to capture this scenario.
- [x] Unit tests pass.
- [x] Successfully deployed to CODE.

This particular scenario will not throw an error.
Other scenario's where the zuoraPreviewResponse invoice is lacking an invoiceItem with the refund amount will throw an error.

Only risk is that fix is incorrectly applied. Unit tests confirm this is not the case.

### DIFF
--- a/handlers/product-switch-api/src/contributionToSupporterPlus.ts
+++ b/handlers/product-switch-api/src/contributionToSupporterPlus.ts
@@ -7,7 +7,11 @@ import type {
 	PreviewOrderRequest,
 } from '@modules/zuora/orders';
 import type { ZuoraClient } from '@modules/zuora/zuoraClient';
-import type { ZuoraSubscription } from '@modules/zuora/zuoraSchemas';
+import type {
+	RatePlan,
+	RatePlanCharge,
+	ZuoraSubscription,
+} from '@modules/zuora/zuoraSchemas';
 import type { Dayjs } from 'dayjs';
 import dayjs from 'dayjs';
 import { removePendingUpdateAmendments } from './amendments';
@@ -15,7 +19,12 @@ import type { CatalogInformation } from './catalogInformation';
 import { takePaymentOrAdjustInvoice } from './payment';
 import { sendThankYouEmail } from './productSwitchEmail';
 import { sendSalesforceTracking } from './salesforceTracking';
-import type { ZuoraPreviewResponse, ZuoraSwitchResponse } from './schemas';
+import type {
+	ZuoraPreviewResponse,
+	ZuoraPreviewResponseInvoice,
+	ZuoraPreviewResponseInvoiceItem,
+	ZuoraSwitchResponse,
+} from './schemas';
 import {
 	zuoraPreviewResponseSchema,
 	zuoraSwitchResponseSchema,
@@ -56,30 +65,51 @@ export const switchToSupporterPlus = async (
 	};
 };
 
-const getContributionRefundAmount = (
-	zuoraResponse: ZuoraPreviewResponse,
+export const handleMissingRefundAmount = (
 	catalogInformation: CatalogInformation,
-	termStartDate: Date,
+	subscription: ZuoraSubscription,
+	currentDate: Date,
 ): number => {
-	const termStartDateDayOfMonth = termStartDate.getUTCDate();
-	const dayOfMonthToday = new Date().getUTCDate();
-
-	const invoice = getIfDefined(
-		zuoraResponse.previewResult?.invoices[0],
-		'No invoice found in the preview response',
+	const ratePlan = getIfDefined(
+		subscription.ratePlans.find(
+			(ratePlan: RatePlan) =>
+				ratePlan.productRatePlanId ===
+				catalogInformation.contribution.productRatePlanId,
+		),
+		'No matching RatePlan found in Subscription,',
 	);
 
-	const contributionRefundAmount = invoice.invoiceItems.find(
-		(invoiceItem) =>
+	const chargedThroughDate: Date = getIfDefined(
+		ratePlan.ratePlanCharges.find(
+			(ratePlanCharge: RatePlanCharge) =>
+				ratePlanCharge.productRatePlanChargeId ===
+				catalogInformation.contribution.chargeId,
+		)?.chargedThroughDate,
+		'No matching chargedThroughDate found in Subscription',
+	);
+
+	if (!(currentDate.toDateString() == chargedThroughDate.toDateString())) {
+		throw Error('No contribution refund amount found in the preview response');
+	}
+
+	return 0;
+};
+
+export const getContributionRefundAmount = (
+	zuoraPreviewInvoice: ZuoraPreviewResponseInvoice,
+	catalogInformation: CatalogInformation,
+	subscription: ZuoraSubscription,
+): number => {
+	const contributionRefundAmount = zuoraPreviewInvoice.invoiceItems.find(
+		(invoiceItem: ZuoraPreviewResponseInvoiceItem) =>
 			invoiceItem.productRatePlanChargeId ===
 			catalogInformation.contribution.chargeId,
 	)?.amountWithoutTax;
 
-	if (!contributionRefundAmount && termStartDateDayOfMonth != dayOfMonthToday) {
-		throw Error('No contribution refund amount found in the preview response');
-	}
-
-	return contributionRefundAmount ? contributionRefundAmount : 0;
+	return (
+		contributionRefundAmount ??
+		handleMissingRefundAmount(catalogInformation, subscription, new Date())
+	);
 };
 
 export const previewResponseFromZuoraResponse = (
@@ -87,25 +117,20 @@ export const previewResponseFromZuoraResponse = (
 	catalogInformation: CatalogInformation,
 	subscription: ZuoraSubscription,
 ): PreviewResponse => {
-	const termStartDate = getIfDefined(
-		subscription.termStartDate,
-		'Unable to find termStartDate for subscription',
-	);
-
-	const invoice = getIfDefined(
+	const invoice: ZuoraPreviewResponseInvoice = getIfDefined(
 		zuoraResponse.previewResult?.invoices[0],
 		'No invoice found in the preview response',
 	);
 
 	const contributionRefundAmount = getContributionRefundAmount(
-		zuoraResponse,
+		invoice,
 		catalogInformation,
-		termStartDate,
+		subscription,
 	);
 
 	const supporterPlusSubscriptionInvoiceItem = getIfDefined(
 		invoice.invoiceItems.find(
-			(invoiceItem) =>
+			(invoiceItem: ZuoraPreviewResponseInvoiceItem) =>
 				invoiceItem.productRatePlanChargeId ===
 				catalogInformation.supporterPlus.subscriptionChargeId,
 		),
@@ -132,6 +157,7 @@ export const previewResponseFromZuoraResponse = (
 		),
 	};
 };
+
 export const preview = async (
 	zuoraClient: ZuoraClient,
 	productSwitchInformation: SwitchInformation,
--- a/handlers/product-switch-api/src/contributionToSupporterPlus.ts
+++ b/handlers/product-switch-api/src/contributionToSupporterPlus.ts
@@ -7,6 +7,7 @@ import type {
 	PreviewOrderRequest,
 } from '@modules/zuora/orders';
 import type { ZuoraClient } from '@modules/zuora/zuoraClient';
+import type { ZuoraSubscription } from '@modules/zuora/zuoraSchemas';
 import type { Dayjs } from 'dayjs';
 import dayjs from 'dayjs';
 import { removePendingUpdateAmendments } from './amendments';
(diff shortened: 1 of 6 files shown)


handlers/product-switch-api/src/contributionToSupporterPlus.ts
```diff
@@ -55,21 +56,49 @@ export const switchToSupporterPlus = async (
 	};
 };
 
+const getContributionRefundAmount = (
+	zuoraResponse: ZuoraPreviewResponse,
+	catalogInformation: CatalogInformation,
+	termStartDate: Date,
+): number => {
+	const termStartDateDayOfMonth = termStartDate.getUTCDate();
+	const dayOfMonthToday = new Date().getUTCDate();
```
johnduffell, line 65: would help testability/repeatability if date is parameterised

handlers/product-switch-api/src/contributionToSupporterPlus.ts
```diff
@@ -55,21 +56,49 @@ export const switchToSupporterPlus = async (
...
+	return contributionRefundAmount ? contributionRefundAmount : 0;
+};
+
 export const previewResponseFromZuoraResponse = (
 	zuoraResponse: ZuoraPreviewResponse,
 	catalogInformation: CatalogInformation,
+	subscription: ZuoraSubscription,
 ): PreviewResponse => {
+	const termStartDate = getIfDefined(
+		subscription.termStartDate,
```
johnduffell, line 89: is this actually optional or would this already been checked by the zod already? 
I think it is always returned from zuora, and it looks like zod checks it also, meaning this extra check is redundant perhaps?
https://github.com/guardian/support-service-lambdas/blob/81a678ef1c7e8e284f4aecf1b42dfdeb802949f6/modules/zuora/src/zuoraSchemas.ts#L33C2-L33C15

handlers/product-switch-api/src/contributionToSupporterPlus.ts
```diff
@@ -55,21 +56,49 @@ export const switchToSupporterPlus = async (
...
 
+const getContributionRefundAmount = (
+	zuoraResponse: ZuoraPreviewResponse,
+	catalogInformation: CatalogInformation,
+	termStartDate: Date,
+): number => {
+	const termStartDateDayOfMonth = termStartDate.getUTCDate();
+	const dayOfMonthToday = new Date().getUTCDate();
+	const invoice = getIfDefined(
+		zuoraResponse.previewResult?.invoices[0],
```
johnduffell, line 67: it's not quite clear what the order preview returns, but I assume we know it either returns exactly one invoice, or the first one is always the one we need?

handlers/product-switch-api/src/contributionToSupporterPlus.ts
```diff
@@ -55,21 +56,49 @@ export const switchToSupporterPlus = async (
...
+		'No invoice found in the preview response',
+	);
+	const contributionRefundAmount = invoice.invoiceItems.find(
+		(invoiceItem) =>
+			invoiceItem.productRatePlanChargeId ===
+			catalogInformation.contribution.chargeId,
+	)?.amountWithoutTax;
+
+	if (!contributionRefundAmount && termStartDateDayOfMonth != dayOfMonthToday) {
```
johnduffell, line 76:
```suggestion
  const todayIsPaymentDay = termStartDateDayOfMonth != dayOfMonthToday;
	if (!contributionRefundAmount && todayIsPaymentDay) {
```

johnduffell: this would only be the case if the billing run has not happened yet.  If the billing run has already happened (which is very early in the morning) then we would expect a refund to be due I think

johnduffell: does the payment date always line up with the term start date? what happens if they take it out on the 31st?  or if the charges are misaligned with the term for some reason?

handlers/product-switch-api/test/contributionToSupporterPlus.test.ts
```diff
@@ -172,21 +174,191 @@ test('preview amounts are correct', () => {
...
 });
 
+/*
+This tests a scenario that occurs when the product switch occurs on the day that the payments would renew.
+ In this scenario there is nothing to refund, so the Invoice Item will not be created.
+ In such a situation, no error should be thrown and the refund amount returned output should be 0.
+ */
+test('Preview that does not contain invoice item for the given charge id, where the day-of-month of the target date equals that of the current date returns an amountWithoutTax value of 0.', () => {
+	const subscription = zuoraSubscriptionSchema.parse(alreadySwitchedJson);
+	jest.useFakeTimers().setSystemTime(new Date(subscription.termStartDate)); //Date works in Epoch milli
```
johnduffell, line 203: ooh didn't know you could set the date timer globally like this!  not sure I like it but it's a useful trick

handlers/product-switch-api/src/contributionToSupporterPlus.ts
```diff
@@ -55,21 +56,49 @@ export const switchToSupporterPlus = async (
...
+		throw Error('No contribution refund amount found in the preview response');
+	}
+
+	return contributionRefundAmount ? contributionRefundAmount : 0;
```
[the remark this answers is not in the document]
tjmw: I think they do very slightly different things (but I don't think the difference is important here, and I like John's suggestion as it removes a bit of repetition/is shorter!)

`contributionRefundAmount ? contributionRefundAmount : 0` will return `contributionRefundAmount` if it's truthy and the right side of the ternary if it's falsey (which includes `0`, `NaN`, `null`, `undefined`).

`contributionRefundAmount ?? 0` (the [nullish coalescing operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing)) will return `contributionRefundAmount` if it's anything other than `null` or `undefined`. If it is `null` or `undefined` the right side will be returned. I think this is _usually_ what you want.

handlers/product-switch-api/src/contributionToSupporterPlus.ts
michaelwmcnamara: Thats a good question. I don't know for certain. Possibly @rupertbates might know more. I will chase up with him.

handlers/product-switch-api/test/contributionToSupporterPlus.test.ts
michaelwmcnamara: I saw it in a few other tests. Its a neat trick! I think in this case I like the idea of passing in the current date as a parameter better. Also think I can narrow the scope of these tests to the getContributionRefundAmount() method rather than running all of previewResponseFromZuoraResponse().

handlers/product-switch-api/src/contributionToSupporterPlus.ts
michaelwmcnamara: Thats a good point.

handlers/product-switch-api/src/contributionToSupporterPlus.ts
michaelwmcnamara: Done in latest commit.

handlers/product-switch-api/src/contributionToSupporterPlus.ts
michaelwmcnamara: Removed testing for undefined.

handlers/product-switch-api/src/contributionToSupporterPlus.ts
michaelwmcnamara: Modified this to pass the invoice itself to getContributionRefundAmount().

handlers/product-switch-api/src/contributionToSupporterPlus.ts
michaelwmcnamara: For payment dates that are later than the size of the month - eg 29th or later in February or 31st in April - the payment will be taken on the 1st of the subsequent month - eg 1st of March or 1st of May for the examples above.
I have added a method called `isStartOfPaymentPeriod()` to capture this complexity and return true if the current date matches these conditions.

handlers/product-switch-api/src/contributionToSupporterPlus.ts
```diff
@@ -55,26 +61,86 @@ export const switchToSupporterPlus = async (
...
+				return [1, 29, 30, 31].includes(termStartDayOfMonth);
+			//Case for payments for 31st in Months with only 30 days.
+			case 4:
+			case 6:
+			case 9:
+			case 11:
+				return termStartDayOfMonth == 1 || termStartDayOfMonth == 31;
+			default:
+				return termStartDayOfMonth == 1;
+		}
```
johnduffell, line 90: I don't really like the look of this! I can imagine something breaking when we get to a leap year, and this should be a solved problem via date libraries.  We are also replicating what zuora knows (or should know).

I think the gold standard would be to ask zuora when the "charged through  date" (i.e. invoiced until) is (it should be in the getsubscription response) which should be midnight first thing on the day that the next payment will be taken.
(dates in zuora are always midnight first thing in the morning)
There may be other options like a billing preview, but I don't think we should need to do them.

If none of that works, I'd prefer to use a date library to add months, and let it decide which months have which days and which years are leap years.

rupertbates: I would use dayjs to check that logic, we use it in a number of other the projects.
You could build a string with the current year and month and the payment day you want to check, and then do this:
```typescript
dayjs('2022-02-31', 'YYYY-MM-DD', true).isValid();
```

michaelwmcnamara: After looking at this with @johnduffell we decided using the charge-through-date for the corresponding rate-plan-charge in the subscription was the most effective way to test this. It means doing a single comparison and is hopefully much easier to follow.

michaelwmcnamara:
> Hi Mike it's quite hard to understand the changes made since the previous reviews - although it can be useful to force push commits when you're just fixing a commit no-one has looked at e.g. a lint fix - please can you treat commits as a unit of review, therefore not rewrite any commits that have been reviewed. (I.e. always push new commits once someone has looked at the previous one), then it's easier to see which changes are new. Otherwise I have to look through the whole thing to try to work out what has changed. Let me know if that makes sense.

Hi @johnduffell understood - I needed to rebase with main as I had failing scala tests of all things when I pushed up my commit. I got a bit over-zealous when rebasing. 
Happy to run through the most recent changes with you if that helps?

handlers/product-switch-api/src/contributionToSupporterPlus.ts
```diff
@@ -56,56 +65,72 @@ export const switchToSupporterPlus = async (
...
+			(ratePlanCharge: RatePlanCharge) =>
+				ratePlanCharge.productRatePlanChargeId ===
+				catalogInformation.contribution.chargeId,
+		)?.chargedThroughDate,
+		'No matching chargedThroughDate found in Subscription',
 	);
 
-	const contributionRefundAmount = invoice.invoiceItems.find(
-		(invoiceItem) =>
+	if (!(currentDate.toDateString() == chargedThroughDate.toDateString())) {
```
johnduffell, line 91:
```suggestion
	if (currentDate.toDateString() !== chargedThroughDate.toDateString()) {
```
I'm not sure if we really need it to be a dateString either just to compare it?  but JS is not my area of expertise

handlers/product-switch-api/src/contributionToSupporterPlus.ts
```diff
@@ -56,56 +65,72 @@ export const switchToSupporterPlus = async (
...
-	const termStartDateDayOfMonth = termStartDate.getUTCDate();
-	const dayOfMonthToday = new Date().getUTCDate();
+	const ratePlan = getIfDefined(
+		subscription.ratePlans.find(
+			(ratePlan: RatePlan) =>
+				ratePlan.productRatePlanId ===
+				catalogInformation.contribution.productRatePlanId,
+		),
+		'No matching RatePlan found in Subscription,',
+	);
```
johnduffell, line 80: not one for this PR, but this is a common pattern and it needs so many lines of boilerplate.  It would be good to invent something where we just wrap a subscription and say something like
```suggestion
	const ratePlan = getRatePlanById(
		subscription,
		catalogInformation.contribution,
	);
```
or
```suggestion
	const ratePlan = getContributionRatePlan(subscription);
```
but I realise that's my scala brain at work

handlers/product-switch-api/test/contributionToSupporterPlus.test.ts
```diff
@@ -198,164 +203,63 @@ This tests a scenario that occurs when the product switch occurs on the day that
...
 	};
+	const subscription: ZuoraSubscription = zuoraSubscriptionSchema.parse(
+		zuoraSubscriptionWithMonthlyContribution,
+	);
+
+	const chargeThroughDate =
+		subscription.ratePlans[0]?.ratePlanCharges[0]?.chargedThroughDate;
+	if (!chargeThroughDate) {
+		throw Error(
+			'Problem with test data: zuoraSubscriptionWithMonthlyContribution should contain a charge-through-date',
```
johnduffell, line 227:
```suggestion
	const chargedThroughDate =
		subscription.ratePlans[0]?.ratePlanCharges[0]?.chargedThroughDate;
	if (!chargedThroughDate) {
		throw Error(
			'Problem with test data: zuoraSubscriptionWithMonthlyContribution should contain a charged-through-date',
```

handlers/product-switch-api/test/contributionToSupporterPlus.test.ts
```diff
@@ -198,164 +203,63 @@ This tests a scenario that occurs when the product switch occurs on the day that
...
-				contribution: {
-					productRatePlanId: 'not_used',
-					chargeId: '2c92c0f85e2d19af015e3896e84d092e',
-				},
-			},
-			subscription,
-		).contributionRefundAmount;
-	}).toThrow(
-		Error('No contribution refund amount found in the preview response'),
+	//Current value of charge-through-date is '2024-07-01'
```
johnduffell, line 254:
```suggestion
	//Current value of charged-through-date is '2024-07-01'
```

handlers/product-switch-api/src/contributionToSupporterPlus.ts
```diff
@@ -56,56 +65,72 @@ export const switchToSupporterPlus = async (
...
 
-	if (!contributionRefundAmount && termStartDateDayOfMonth != dayOfMonthToday) {
-		throw Error('No contribution refund amount found in the preview response');
-	}
-
-	return contributionRefundAmount ? contributionRefundAmount : 0;
+	return (
+		contributionRefundAmount ??
+		handleMissingRefundAmount(catalogInformation, subscription, new Date())
+	);
```
johnduffell, line 112: not keen on using the word "handle" because it makes it unclear what it's actually doing

I'd be tempted to deal with it as an assertion and then continue as normal if it passes.

The "refundExpected" function is the new name for handleMissingRefundAmount
```suggestion
  if (contributionRefundAmount == undefined && refundExpected(catalogInformation, subscription, new Date())) {
    throw Error('No contribution refund amount found in the preview response');
  }
  
	return (contributionRefundAmount ?? 0);
```

johnduffell (approved):
good work I think this is going to be nice and reliable and I've made a few style suggestions and comments if you want to tidy it up a bit more.

PR #2457, opened 2024-09-05
fix: revert single-contribution-salesforce-writes lambda to Java 11 as PATCH method not supported

AndreaDiotallevi:
## What does this change?

PATCH is (or was) not an official http method, so the built in JRE http client doesn't support it.
https://github.com/openjdk/jdk/blob/b389bb456726184e4691777b1bb02d4b8a8a3f97/src/java.base/share/classes/java/net/HttpURLConnection.java#L363

This reverts the lambda to use Java 11 to get it working again.

Longer term we should move off that http client and onto one of the others e.g. okhttp so that we can update.


johnduffell (approved):
thanks for that! I've just added some context to the description in case of people looking back in future 👍

PR #2458, opened 2024-09-05
prevent alarms for expired tokens for CSRs in holiday stops API

johnduffell:
CSRs use the holiday stops API from salesforce using their own token.  This means that if their password has expired during the day, we get a load of 5xx errors and an alarm.

Salesforce returns this
```
[
{
    "message": "Your password has expired. Please reset your password.",
    "errorCode": "INVALID_OPERATION_WITH_EXPIRED_PASSWORD"
}
]
```

This PR changes is so that if that happens, we return a 400 which would hopefully give more information at the CSRs end, and not set off our alarm.

As part of this PR I have removed the Lazy behaviour of SF auth (we used to only authenticate when we make an actual SF call).   This means we can deal with the auth failure once up front, using a nice type, rather than trying to shoehorn the error into all of the actual calls to SF.
I have also improved click through in Intellij by removing some "function" types and made them classes with named methods.

Unit test added
- `SalesforceAuthenticateTest should parse a password auth failure alright`
- `com.gu.holiday_stops.HandlerTest get should fail correctly if CSR password is expired`

Tested in CODE both success cases and failure cases.


handlers/holiday-stop-api/src/main/scala/com/gu/holiday_stops/Handler.scala
```diff
@@ -240,6 +241,19 @@ object Handler extends Logging {
     ApiGatewayResponse.messageResponse("500", error.message)
   }
 
+  implicit class SFToApiGatewayOp[A](theEither: Either[ClientFailure, A]) {
+    def sfTokenHandlingToApiGatewayOp(action: String): ApiGatewayOp[A] =
+      theEither match {
+        case Right(success) => ContinueProcessing(success)
+        case Left(error @ CustomError(message)) if message.startsWith("INVALID_OPERATION_WITH_EXPIRED_PASSWORD") =>
+          logger.error(s"SF token was not valid - returning 400 $action: $error")
+          ReturnWithResponse(badRequest("SF credentials could not be used"))
```
johnduffell, line 250: @graham228221 would this response body be displayed in the SF UI for CSRs? if so we can improve the wording and perhaps suggest logging out and in again?

tjmw (approved):
Looks great, and I think semantically a 400 makes more sense in this case 👍

handlers/holiday-stop-api/src/main/scala/com/gu/holiday_stops/Handler.scala
graham228221: Sorry I missed this! No it wouldn't be shown they'd just see a generic error, I think. But it would appear in our error logs on the SF side. 

But I will say that they do generally try logging out and in again as a first attempt to fix anyway if they see any errors with the Holiday Stop page in SF as this is the most common cause.

handlers/holiday-stop-api/src/main/scala/com/gu/holiday_stops/Handler.scala
```diff
@@ -240,6 +241,19 @@ object Handler extends Logging {
     ApiGatewayResponse.messageResponse("500", error.message)
   }
 
+  implicit class SFToApiGatewayOp[A](theEither: Either[ClientFailure, A]) {
+    def sfTokenHandlingToApiGatewayOp(action: String): ApiGatewayOp[A] =
+      theEither match {
+        case Right(success) => ContinueProcessing(success)
+        case Left(error @ CustomError(message)) if message.startsWith("INVALID_OPERATION_WITH_EXPIRED_PASSWORD") =>
```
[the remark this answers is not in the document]
johnduffell: it actually concatenates them together before we get hold of it (which is annoying, I'm tempted to make it also store them separately as I don't like looking inside strings if I don't have to)
https://github.com/guardian/support-service-lambdas/blob/b0194326780271679c71a63096cf4c863ae1aa52/lib/salesforce/client/src/main/scala/com/gu/salesforce/SalesforceClient.scala#L68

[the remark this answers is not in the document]
tjmw: Ahh I see, thanks for explaining.

handlers/holiday-stop-api/src/main/scala/com/gu/holiday_stops/Handler.scala
```diff
@@ -103,11 +103,15 @@ object Handler extends Logging {
...
-      ).value.toDisjunction.toApiGatewayOp("authenticate with SalesForce")
+      sfClient <- SalesforceClient.auth(response, config.sfConfig) match {
+        case Right(success) => ContinueProcessing(success)
+        case Left(errors) if errors.exists(_.errorCode == "INVALID_OPERATION_WITH_EXPIRED_PASSWORD") =>
+          logger.error(s"SF token was not valid - returning 400 authenticate with SalesForce: $errors")
+          ReturnWithResponse(badRequest("salesforce returned INVALID_OPERATION_WITH_EXPIRED_PASSWORD"))
+        case Left(error) =>
+          logger.error(s"Failed to authenticate with SalesForce: $error")
+          ReturnWithResponse(internalServerError("Failed to execute lambda - unable to authenticate with SalesForce"))
+      }
```
johnduffell, line 114: this lambda is the one that uses credentials from salesforce/request. 
So this is the only place we need to check if it's an expired password and return accordingly.

handlers/holiday-stop-api/src/main/scala/com/gu/holiday_stops/Handler.scala
```diff
@@ -235,9 +239,10 @@ object Handler extends Logging {
   private def exposeSfErrorMessageIn500ApiResponse[A <: Product](
       action: String,
       inputThatCausedError: Option[A] = None,
-  ): ClientFailure => ApiResponse = (error: ClientFailure) => {
+      error: String,
+  ): ApiResponse = {
     logger.error(s"Failed to $action using input $inputThatCausedError: $error")
-    ApiGatewayResponse.messageResponse("500", error.message)
+    ApiGatewayResponse.messageResponse("500", error)
   }
```
johnduffell, line 246: slight cleanup of the function signature

handlers/holiday-stop-api/src/main/scala/com/gu/holiday_stops/Handler.scala
```diff
@@ -461,7 +468,7 @@ object Handler extends Logging {
   )(req: ApiGatewayRequest, sfClient: SfClient): ApiResponse = {
     val sfClientGetWithParams = sfClient.wrapWith(JsonHttp.getWithParams)
     val lookupOpHolidayStopsOp = LookupByContactAndOptionalSubscriptionName(sfClientGetWithParams)
-    val updateRequestDetailOp = CancelHolidayStopRequestDetail(sfClient.wrapWith(JsonHttp.post))
+    val updateRequestDetailOp = CancelHolidayStopRequestDetail.doRequest(sfClient.wrapWith(JsonHttp.post)) _
```
johnduffell, line 471: rename `apply` to `doRequest` to make it clearer.  And the underscore shows that there's more to do to call the method.  Another option would be to make it even more explicit (eta expand it) with something like
```suggestion
    def updateRequestDetailOp(something: SomeType) = CancelHolidayStopRequestDetail.doRequest(sfClient.wrapWith(JsonHttp.post))(something)
```

handlers/holiday-stop-api/src/main/scala/com/gu/holiday_stops/Handler.scala
```diff
@@ -481,14 +488,23 @@ object Handler extends Logging {
...
           ),
         )
+      _ <- updateResult.toApiGatewayOp(e =>
+        exposeSfErrorMessageIn500ApiResponse(
+          s"cancel holiday stop request details: ${holidayStopRequestDetailToUpdate.map(_.Id.value).mkString(",")} " +
+            s"(contact $contact)",
+          Some(cancelBody),
+          e,
+        ),
+      )
```
johnduffell, line 507: there could be an HTTP error returned, or the salesforce 200 response could contain something that indicated an issue.
The http errors were handled with a (http) ClientFailure, and a subclass had been added called CustomError which would be used for any issues with what was returned from salesforce.
This primarily caused an issue for auth, so I removed the CustomError there so that we could return the original deserialised messages, to check for the password expired error.

I then removed the CustomError completely from the http library, as it didn't make sense to have a SF thing in the http library, but it did mean we ended up with http ClientFailure wrapping an Either to handle the errors at each level.
These then need to be unpacked in the for comprehension and in reality we deal with them the same way.

So I'm in two minds whether to do this change or just add the SF CustomError back into the http library and just use it in situations where just want to fail regardless.

handlers/identity-backfill/src/test/scala/com/gu/identityBackfill/EndToEndHandlerTest.scala
```diff
@@ -65,6 +59,12 @@ class EndToEndHandlerTest extends AnyFlatSpec with Matchers {
           """{"queryString":"SELECT Id FROM Contact where WorkEmail='email@address'"}""",
         ),
         BasicRequest("GET", "/user?emailAddress=email%40address", ""),
+        BasicRequest(
+          "POST",
+          "/services/oauth2/token",
+          "client_id=clientsfclient&client_secret=clientsecretsfsecret&username=usernamesf" +
+            "&password=passSFpasswordtokentokenSFtoken&grant_type=password",
+        ),
```
johnduffell, line 67: this shows that SF auth happens first now, rather than just before the SF call.

handlers/sf-datalake-export/src/main/scala/com/gu/sf_datalake_export/handlers/DownloadBatchHandler.scala
```diff
@@ -103,7 +103,7 @@ object DownloadBatchHandler extends LazyLogging {
     val loadConfig = LoadConfigModule(stage, getFromS3)
     for {
       sfConfig <- loadConfig.load[SFAuthConfig](SFExportAuthConfig.location, sfAuthConfigReads).toTry
-      sfClient <- SalesforceClient(getResponse, sfConfig).value.toTry
+      sfClient <- SalesforceClient.auth(getResponse, sfConfig).leftMap(bodies => new Throwable(bodies.toString)).toTry
```
johnduffell, line 106: to get it into a try we can't just convert the ClientFailableOp directly (as it's the deserialised error body) so we convert the Left to a throwable first.

lib/holiday-stops/src/main/scala/com/gu/salesforce/holiday_stops/SalesforceHolidayStopRequest.scala
```diff
@@ -243,34 +242,34 @@ object SalesforceHolidayStopRequest extends Logging {
...
+      Right(response)
+    } else {
+      logger.error(response.toString)
+      val failuresStr: String = failures
+        .flatMap(_.body.map(_.validate[List[SalesforceErrorResponseBody]].asOpt))
+        .flatten
+        .mkString(", ")
+      Left(s"MULTIPLE ERRORS : ${failuresStr.take(500)}${if (failuresStr.length > 500) "..." else ""}")
+    }
+  }
```
johnduffell, line 261: this used to be an "HttpOpWrapper" which did both pre- and post-request processing in one go.  However it doesn't look that nice.
I thought about converting to a class with 2 methods, but actually it seemed easier to just have "setupRequest" and "postRequest" separately.

lib/salesforce/client/src/main/scala/com/gu/salesforce/SalesforceClient.scala
```diff
@@ -1,32 +1,36 @@
...
+  def auth(
       getResponse: Request => Response,
       config: SFAuthConfig,
-      shouldExposeSalesforceErrorMessageInClientFailure: Boolean = false,
-  ): LazyClientFailableOp[HttpOp[StringHttpRequest, BodyAsString]] =
-    SalesforceAuthenticate(getResponse)(config).map { sfAuth: SalesforceAuth =>
-      HttpOp(getResponse)
-        .flatMap {
-          toClientFailableOp(
-            shouldExposeSalesforceErrorMessageInClientFailure.toOption(parseSalesforceErrorResponseAsCustomError _),
```
johnduffell, line 23: the auth failure used to be dealt with at this point alongside the error from the actual call we just made.
Now you can see the auth error is dealt with just below (`Json.parse(cf.body).as[List[SalesforceErrorResponseBody]]`) and this is purely the error for the actual SF call we will make.

johnduffell: shouldExposeSalesforceErrorMessageInClientFailure only used to be true for the "auth" call. so we don't need it any more.  The relevant calls will expose the entire SF response in their wrapped Either that's in the ClientFailure, so our code can decide for itself how they want to deal with it.

lib/salesforce/client/src/main/scala/com/gu/salesforce/SalesforceClient.scala
```diff
@@ -67,22 +71,7 @@ object SalesforceClient extends LazyLogging {
...
-        )
-      case _ =>
-        logger.warn("Salesforce error response didn't contain any detail")
-        genericError
-    }
-  } catch {
-    case _: Throwable =>
-      logger.warn("Couldn't parse the Salesforce error response body")
-      genericError
-  }
```
johnduffell, line 87: we no longer serialise the SF error into the CustomError, we just return it on the Left of an Either as-is

lib/salesforce/client/src/main/scala/com/gu/salesforce/SalesforceClient.scala
johnduffell: update - we also rely on this logic for passing the validation errors back from salesforce to the client https://github.com/guardian/support-service-lambdas/pull/466

johnduffell: although after some testing with `main` on CODE, it doesn't seem like it's actually working as the SF response doesn't match the schema (any more?)
image

handlers/holiday-stop-api/src/main/scala/com/gu/holiday_stops/Handler.scala
```diff
@@ -78,36 +76,40 @@ object Handler extends Logging {
 
   def operationForEffects(
       response: Request => Response,
-      stage: Stage,
       fetchString: StringFromS3,
       backend: SttpBackend[Identity, Any],
       idGenerator: => String,
+      now: () => ZonedDateTime = () => ZonedDateTime.now(),
```
johnduffell, line 82: we now pass `now` through from here, so that in the new test, the withdrawal date of the stop is deterministic

handlers/holiday-stop-api/src/main/scala/com/gu/holiday_stops/Handler.scala
```diff
@@ -232,14 +237,6 @@ object Handler extends Logging {
         ),
       )
 
-  private def exposeSfErrorMessageIn500ApiResponse[A <: Product](
-      action: String,
-      inputThatCausedError: Option[A] = None,
-  ): ClientFailure => ApiResponse = (error: ClientFailure) => {
-    logger.error(s"Failed to $action using input $inputThatCausedError: $error")
-    ApiGatewayResponse.messageResponse("500", error.message)
-  }
```
johnduffell, line 241: this is moved to the file SFErrorExposer, which has suitable logic to extract either failed HTTP responses or successful responses which wrap a list of failures.  Rather than it being previously shoehorned into a ClientFailure's message field.

handlers/holiday-stop-api/src/main/scala/com/gu/holiday_stops/Handler.scala
```diff
@@ -345,19 +342,23 @@ object Handler extends Logging {
   )(req: ApiGatewayRequest, sfClient: SfClient): ApiResponse = {
 
     val lookupOp =
-      SalesforceHolidayStopRequest.LookupByContactAndOptionalSubscriptionName(sfClient.wrapWith(JsonHttp.getWithParams))
+      new SalesforceHolidayStopRequest.LookupByContactAndOptionalSubscriptionName(
+        sfClient.wrapWith(JsonHttp.getWithParams),
+      )
```
johnduffell, line 347: all the old function types meant that cmd-click was unusable, so I have converted them all to classes with a method, so that can be clicked through normally

handlers/holiday-stop-api/src/main/scala/com/gu/holiday_stops/Handler.scala
```diff
@@ -345,19 +342,23 @@ object Handler extends Logging {
...
     (for {
       contact <- extractContactFromHeaders(req.headers)
       subName <- extractSubNameOp
-      usersHolidayStopRequests <- lookupOp(
-        contact,
-        Some(subName),
-        Some(MutableCalendar.today.minusMonths(6)),
-      ).toDisjunction
+      usersHolidayStopRequests <- lookupOp
+        .run(
```
johnduffell, line 356: if you cmd click `run` here it will go to the implementation 🎉

handlers/holiday-stop-api/src/main/scala/com/gu/holiday_stops/Handler.scala
```diff
@@ -441,12 +443,11 @@ object Handler extends Logging {
...
-      _ <- createOp(createBody).toDisjunction.toApiGatewayOp(
-        exposeSfErrorMessageIn500ApiResponse(
-          s"create new Holiday Stop Request for subscription ${requestBody.subscriptionName} (contact $contact)",
-          Some(createBody),
-        ),
+      sfErrorExposer = new SFErrorExposer(
+        s"create new Holiday Stop Request for subscription ${requestBody.subscriptionName} (contact $contact)",
+        Some(createBody),
       )
+      _ <- createOp.run(createBody).toDisjunction.toApiGatewayOp(sfErrorExposer.parseFailureTo500ApiResponse _)
```
johnduffell, line 450: now we use `parseFailureTo500ApiResponse` which turns a 4xx error into a message for the client, but only if it can parse the appropriate fields out.  Rather than assuming it's already in the `message` field based on `shouldExposeSalesforceErrorMessageInClientFailure = true` and some special case logic in our http client code.

handlers/holiday-stop-api/src/main/scala/com/gu/holiday_stops/Handler.scala
```diff
@@ -481,14 +484,16 @@ object Handler extends Logging {
...
+      sfErrorExposer = new SFErrorExposer(
+        s"cancel holiday stop request details: ${holidayStopRequestDetailToUpdate.map(_.Id.value).mkString(",")} " +
+          s"(contact $contact)",
+        Some(cancelBody),
+      )
+      updateResult <- updateRequestDetailOp
+        .doRequest(cancelBody)
+        .toDisjunction
+        .toApiGatewayOp(sfErrorExposer.parseFailureTo500ApiResponse _)
+      _ <- sfErrorExposer.compositeResponseErrorto500ApiResponse(updateResult)
```
johnduffell, line 496: for a composite operation, the overall operation might have succeeded, but the contained operations may have failed.  So this step goes through the results in the 200 response and checks them all appropriately.

handlers/holiday-stop-api/src/test/scala/com/gu/holiday_stops/HandlerTest.scala
```diff
@@ -186,13 +175,45 @@ class HandlerTest extends AnyFlatSpec with Matchers {
...
+              ),
+            ).response,
+            FakeFetchString.fetchString,
+            testBackend,
+            "test-generated-id",
+          ),
+        )
+          .map(_.steps(ApiGatewayRequest(None, None, None, None, None, None))),
+      ) { case ReturnWithResponse(response) =>
+        response.statusCode should equal("400")
```
johnduffell, line 202: if this test passes (400 error), then we shouldn't get a 5xx alarm if the csr password is expired.

handlers/holiday-stop-api/src/test/scala/com/gu/holiday_stops/HandlerTest.scala
```diff
@@ -353,6 +372,71 @@ class HandlerTest extends AnyFlatSpec with Matchers {
       }
     }
 
+  "DELETE /hsr/<<sub name>>/<<holidayStopRequestId>> endpoint" should
+    "not let you withdraw when it's been actioned" in {
```
johnduffell, line 376: this test is added mainly to check that the SF error message is exposed in the response.  The exposing itself was added in this PR https://github.com/guardian/support-service-lambdas/pull/466

handlers/holiday-stop-api/src/test/scala/com/gu/holiday_stops/HandlerTest.scala
```diff
@@ -353,6 +372,71 @@ class HandlerTest extends AnyFlatSpec with Matchers {
...
+          FakeFetchString.fetchString,
+          testBackend,
+          testId,
+          () => withdrawTime,
+        ),
+      ).map(_.steps(request))
+    ) { case ContinueProcessing(response) =>
+      response.statusCode should equal("500")
+      inside ((Json.parse(response.body) \ "message").get) {
+        case JsString(message) => message should be("FIELD_CUSTOM_VALIDATION_EXCEPTION : Holiday Stop Request cannot be withdrawn because some publications have been actioned (credited to customer)")
```
johnduffell, line 435: this test uses hard coded data, but it was recorded from an attempt which triggered this message 
https://github.com/guardian/salesforce/blob/446849259846202ea058923be046525a48e0211b/force-app/main/default/classes/HolidayStopUtilities.cls#L6

handlers/holiday-stop-api/src/test/scala/com/gu/holiday_stops/HandlerTest.scala
```diff
@@ -353,6 +372,71 @@ class HandlerTest extends AnyFlatSpec with Matchers {
...
+      id = ref,
+      subscriptionName = SubscriptionName(subscriptionName),
+      requestDetail = List(holidayStopRequestsDetail),
+    )
+
+    val request = ApiGatewayRequest(
+      httpMethod = Some("DELETE"),
+      queryStringParameters = None,
+      body = None,
+      headers = Some(Map("x-salesforce-contact-id" -> contactId)),
```
johnduffell, line 404: I noticed all the tests use the SF contact id rather than identity Id (used from MMA).  Not sure if we should improve the coverage there too.

handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/Handler.scala
```diff
@@ -123,7 +119,8 @@ object Handler {
       zuoraRestConfig <- loadConfig.load[ZuoraRestConfig].toApiGatewayOp("load zuora config")
       sfAuthConfig <- loadConfig.load[SFAuthConfig].toApiGatewayOp("load sfAuth config")
       identityConfig <- loadConfig.load[IdentityConfig].toApiGatewayOp("load identity config")
-      configuredOp = operation(zuoraRestConfig, sfAuthConfig, identityConfig)
+      sfAuth <- SalesforceClient.auth(response, sfAuthConfig).toApiGatewayOp("sf auth")
+      configuredOp = operation(zuoraRestConfig, sfAuth, identityConfig)
```
johnduffell, line 123: SF auth is eager now rather than lazy, so we need to handle the failures here rather than when we call an op.

lib/effects/src/test/scala/com/gu/effects/SFTestEffects.scala
```diff
@@ -19,20 +22,54 @@ object SFTestEffects {
...
+    ) ->
+      HTTPResponse(200, authSuccessResponseBody)
+
+  val authFailureResponseBody: String =
+    s"""
+       |[{
+       |  "message": "The users password has expired, you must call SetPassword before attempting any other API operations",
+       |  "errorCode": "INVALID_OPERATION_WITH_EXPIRED_PASSWORD"
+       | }]
+    """.stripMargin
```
johnduffell, line 39: this represents failed auth

lib/effects/src/test/scala/com/gu/effects/SFTestEffects.scala
```diff
@@ -19,20 +22,54 @@ object SFTestEffects {
...
+    ) ->
+      HTTPResponse(
+        400,
+        """[
+        |{
+        |    "message": "Holiday Stop Request cannot be withdrawn because some publications have been actioned (credited to customer)",
+        |    "errorCode": "FIELD_CUSTOM_VALIDATION_EXCEPTION",
+        |    "fields": []
+        |}
+        |]""".stripMargin,
```
johnduffell, line 72: this represents a non withdrawable hol stop.  I noticed that some errors come back wrapped in a much deeper object, which we don't handle, so I haven't added that facility.  I'm not sure how useful the errors are to CSRs.

lib/holiday-stops/src/main/scala/com/gu/salesforce/holiday_stops/SalesforceHolidayStopRequest.scala
```diff
@@ -91,16 +91,23 @@ object SalesforceHolidayStopRequest extends Logging {
 
   implicit val formatIds = Json.format[RecordsWrapperCaseClass[HolidayStopRequest]]
 
-  object LookupByContactAndOptionalSubscriptionName {
+  class LookupByContactAndOptionalSubscriptionName(sfGet: HttpOp[RestRequestMaker.GetRequestWithParams, JsValue]) {
```
johnduffell, line 94: all these objects have been changed to classes, to make the code more navigable, i.e. click through in Intellij works as not everything is a function.

lib/holiday-stops/src/main/scala/com/gu/salesforce/holiday_stops/SalesforceHolidayStopRequest.scala
```diff
@@ -91,16 +91,23 @@ object SalesforceHolidayStopRequest extends Logging {
...
+      optionalHistoricalCutOff: Option[LocalDate],
+    ): ClientFailableOp[List[HolidayStopRequest]] =
       sfGet
-        .setupRequestMultiArg(toRequest _)
         .parse[RecordsWrapperCaseClass[HolidayStopRequest]]
         .map(_.records)
-        .runRequestMultiArg
+        .runRequest(toRequest(contact, optionalSubscriptionName, optionalHistoricalCutOff))
+
+  }
```
johnduffell, line 108: this becomes a normal method rather than a method that returns a function.  This means the whole multiArg thing isn't needed.

lib/holiday-stops/src/main/scala/com/gu/salesforce/holiday_stops/SalesforceHolidayStopRequest.scala
```diff
@@ -243,34 +252,20 @@ object SalesforceHolidayStopRequest extends Logging {
...
-          ClientSuccess(response)
-        } else {
-          logger.error(response.toString)
-          val failuresStr: String = failures
-            .flatMap(_.body.map(_.validate[List[SalesforceErrorResponseBody]].asOpt))
-            .flatten
-            .mkString(", ")
-          CustomError(s"MULTIPLE ERRORS : ${failuresStr.take(500)}${if (failuresStr.length > 500) "..." else ""}")
-        }
-      },
```
johnduffell, line 262: this composite response parsing has moved to `sfErrorExposer.compositeResponseErrorto500ApiResponse`

lib/holiday-stops/src/main/scala/com/gu/salesforce/holiday_stops/SalesforceHolidayStopRequest.scala
```diff
@@ -389,18 +385,19 @@ object SalesforceHolidayStopRequest extends Logging {
   }
 
   object WithdrawHolidayStopRequest {
-    case class WithdrawnTimePatch(Withdrawn_Time__c: ZonedDateTime = ZonedDateTime.now())
+    case class WithdrawnTimePatch(Withdrawn_Time__c: ZonedDateTime)
```
johnduffell, line 388: the time is now passed in so that this code is deterministic, and a test is added.

lib/holiday-stops/src/main/scala/com/gu/salesforce/holiday_stops/SalesforceSFSubscription.scala
```diff
@@ -38,7 +38,7 @@ object SalesforceSFSubscription extends Logging {
         .map(_.records.headOption)
         .runRequestMultiArg
 
-    def toRequest(subscriptionName: SubscriptionName, contact: Contact) = {
+    private def toRequest(subscriptionName: SubscriptionName, contact: Contact) = {
```
johnduffell, line 41: there are too many accessible methods called toRequest and I was getting a clash somewhere else.

lib/restHttp/src/main/scala/com/gu/util/resthttp/HttpOp.scala
```diff
@@ -72,18 +69,6 @@ object HttpOp {
...
-case class LazyClientFailableOp[+VALUE](underlying: () => ClientFailableOp[VALUE]) {
-  lazy val value = underlying()
-
-  def flatMap[OUT](function: VALUE => LazyClientFailableOp[OUT]): LazyClientFailableOp[OUT] =
-    LazyClientFailableOp(() => value.flatMap(function(_).value))
-
-  def map[OUT](function: VALUE => OUT): LazyClientFailableOp[OUT] =
-    LazyClientFailableOp(() => value.map(function))
-
-}
```
johnduffell, line 85: IIRC this was added where we might not call salesforce at all in a lambda, so wanted to save latency. But  now all the places it was used, I could see that it would call salesforce regardless.
(The only doubtful case was identity-backfill, but that is a less used (if at all) manually called lambda., so it's not really an issue.)

lib/restHttp/src/main/scala/com/gu/util/resthttp/RestRequestMaker.scala
```diff
@@ -26,10 +23,11 @@ object RestRequestMaker extends LazyLogging {
       logger.error(
         s"HTTP request was unsuccessful, response status was ${response.code}, response body: \n $response\n$truncated",
       )
-      maybeErrorBodyParser match {
-        case Some(errorBodyParser) => errorBodyParser(body)
```
johnduffell, line 30: we used to try to parse any error body up here, and turn it into a "CustomError", but now we just fold the original body into the relevant ClientFailure object and let the caller do any parsing they like in their business logic (the SFErrorExposer in this case)

lib/restHttp/src/main/scala/com/gu/util/resthttp/Types.scala
```diff
@@ -10,15 +10,22 @@ object Types {
...
+  case class Unauthorised(message: String, body: String) extends Error4xx
 
-  case class CustomError(message: String) extends ClientFailure
+  // this case refers to deserialisation errors or issues with headers etc
+  case class GenericError(message: String, body: String = "") extends ClientFailure
 
-  case class PaymentError(message: String) extends ClientFailure
+  case class PaymentError(message: String) extends ClientFailure {
+    def body: String = ""
+  }
```
johnduffell, line 28: `PaymentError` is probably something that should be handled in the business logic rather than in the http client, but I didn't want to dig into it now.

lib/salesforce/client/src/main/scala/com/gu/salesforce/SalesforceAuthenticate.scala
```diff
@@ -35,6 +41,16 @@ object SalesforceAuthenticate extends Logging {
       }
       .parse[SalesforceAuth]
       .map(_.withLogging(s"salesforce auth"))
-      .runRequestLazy
+      .runRequest(config)
+      .toDisjunction
+      .left
+      .map { cf =>
+        Try(Json.parse(cf.body).as[List[SalesforceErrorResponseBody]]) match {
+          case Success(value) => value
```
johnduffell, line 49: if the SF auth call returns an error, parse it and return the error list as a Left, so we can identify password issues in the business logic.

lib/salesforce/client/src/main/scala/com/gu/salesforce/SalesforceClient.scala
johnduffell: just to round this out - some errors seem to be returned as a flat list (and were/are handled by this lambda) and some are returned in the above object and would just be logged and reported generically.
It probably should be fixed to report them all, but not in this PR.

handlers/holiday-stop-api/src/main/scala/com/gu/holiday_stops/Handler.scala
```diff
@@ -596,31 +610,34 @@ object Handler extends Logging {
...
+          s"lookup Holiday Stop Requests for contact $contact",
+        )
+      _ <- existingForUser
         .find(_.Id == pathParams.holidayStopRequestId)
         .toApiGatewayContinueProcessing(ApiGatewayResponse.forbidden("not your holiday stop"))
-      _ = hsr.Holiday_Stop_Request_Detail__r
-        .exists(_.records.exists(_.Is_Actioned__c))
-        .toApiGatewayContinueProcessing(
-          ApiGatewayResponse.forbidden("can't withdraw holiday stop with any actioned items"),
-        ) // FIXME: Does this do anyting?
```
johnduffell, line 618: this indeed didn't do anthing as it was an `=` instead of a `<-`, and the logic was backwards.
Since the SF call will do the same check and return a validation error, I just removed it here.

handlers/holiday-stop-api/src/main/scala/com/gu/holiday_stops/SFErrorExposer.scala
```diff
@@ -0,0 +1,69 @@
...
+      .filter(resp => !successStatusCodes.contains(resp.httpStatusCode))
+    if (failures.isEmpty) {
+      ContinueProcessing(())
+    } else {
+      logger.error(response.toString)
+      val failuresStr: String = failures
+        .flatMap(_.body.map(_.validate[List[SalesforceErrorResponseBody]].asOpt))
+        .flatten
+        .mkString(", ")
+      val sfErrorText = s"MULTIPLE ERRORS : ${failuresStr.take(500)}${if (failuresStr.length > 500) "..." else ""}"
```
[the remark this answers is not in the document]
johnduffell: will do - I hadn't looked at the content of this code as it is just moved from https://github.com/guardian/support-service-lambdas/pull/2458/files#diff-cfa6044d42c42cb2f880ba315b66e989a598a35853becb8826a04dd931ebf825L260

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/productcatalog/PricesFromZuoraCatalog.scala
```diff
@@ -5,12 +5,12 @@ import com.gu.newproduct.api.productcatalog.ZuoraIds.ProductRatePlanId
...
 import com.gu.util.config.ZuoraEnvironment
 import com.gu.effects.S3Location
-import com.gu.util.resthttp.Types.ClientFailableOp
+import com.gu.util.resthttp.Types.{ClientFailableOp, ClientSuccess, GenericError}
 import play.api.libs.json.{Json, Reads}
 import ZuoraCatalogWireModel._
 import com.gu.i18n.Currency
 
-import scala.util.Try
+import scala.util.{Failure, Success, Try}
```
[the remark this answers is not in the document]
johnduffell: funnily enough, neither, I'll clean up the imports here.
I never know whether to do a global organise imports on all edited files, and take the noise, or just try to be judicious about it.
I think some teams enforce no unused imports in the compiler, but we've never been strict on coding style at all until recently (added scalafmt).

handlers/holiday-stop-api/src/main/scala/com/gu/holiday_stops/SFErrorExposer.scala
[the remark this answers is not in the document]
johnduffell: intellij spotted that we also use it on line 36 above so I'll pull it out there too.

handlers/holiday-stop-api/src/main/scala/com/gu/holiday_stops/Handler.scala
johnduffell: this is not true, when a CSR uses this lambda, we do log in, but using the default credentials.  We then ignore the token that was returned, and use the CSR token instead.  This means that the whole point of this PR has not worked, and needs reworking to catch the error on the individual calls (or do a dummy auth call to check the token)

PR #2465, opened 2024-09-11
adding new consent for guardian_products_services

Joemgru:
## What does this change?

We are creating a new marketing permission called Guardian Products and Services to differentiate from Similar Guardian Products.

We’re currently using ‘Similar Guardian Products’ (SGP) for both the ‘negotiation of sale’ soft opt-in (account registrations), and the ‘sale’ soft opt-in (supporters). This means that there is no way to tell at source which type of marketing emails these readers should be receiving—acquisition or upsell.
see the 1 pager for more info
https://docs.google.com/document/d/1H4SitgsfONXFhu2mqzcdnXUHmYD75P_tf59c-lY2meo

We have a solution approved by Data Privacy: the creation of a new reader marketing permission called ‘Guardian Products and Services’ (GPS).

This change is adding the new consent setting `guardian_products_services` to the soft-opt in service.

~There is a corresponding change on the consent-autolapse lambda here:~
~https://github.com/guardian/consent-autolapse/pull/20~
update: we don't want to autolapse the consents

**Blocked** until corresponding changes in MMA made

### DIFF
--- a/handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/models/ConsentsMapping.scala
+++ b/handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/models/ConsentsMapping.scala
@@ -3,6 +3,7 @@ package com.gu.soft_opt_in_consent_setter.models
 object ConsentsMapping {
   val yourSupportOnboarding = "your_support_onboarding"
   val similarGuardianProducts = "similar_guardian_products"
+  val guardianProductsAndServices = "guardian_products_services"
   val supporterNewsletter = "supporter_newsletter"
   val subscriberPreview = "subscriber_preview"
   val guardianWeeklyNewsletter = "guardian_weekly_newsletter"
@@ -11,49 +12,58 @@ object ConsentsMapping {
     "Membership" -> Set(
       yourSupportOnboarding,
       similarGuardianProducts,
+      guardianProductsAndServices,
       supporterNewsletter,
     ),
     "Supporter" -> Set(
       yourSupportOnboarding,
       similarGuardianProducts,
+      guardianProductsAndServices,
       supporterNewsletter,
     ),
     "Contribution" -> Set(
       yourSupportOnboarding,
       similarGuardianProducts,
+      guardianProductsAndServices,
       supporterNewsletter,
     ),
     "Recurring Contribution" -> Set(
       yourSupportOnboarding,
       similarGuardianProducts,
+      guardianProductsAndServices,
       supporterNewsletter,
     ),
     "Contributor" -> Set(
       yourSupportOnboarding,
       similarGuardianProducts,
+      guardianProductsAndServices,
       supporterNewsletter,
     ),
     "newspaper" -> Set(
       yourSupportOnboarding,
       similarGuardianProducts,
+      guardianProductsAndServices,
       subscriberPreview,
       supporterNewsletter,
     ),
     "Newspaper - Home Delivery" -> Set(
       yourSupportOnboarding,
       similarGuardianProducts,
+      guardianProductsAndServices,
       subscriberPreview,
       supporterNewsletter,
     ),
     "Newspaper - Voucher Book" -> Set(
       yourSupportOnboarding,
       similarGuardianProducts,
+      guardianProductsAndServices,
       subscriberPreview,
       supporterNewsletter,
     ),
     "Newspaper - Digital Voucher" -> Set(
       yourSupportOnboarding,
       similarGuardianProducts,
+      guardianProductsAndServices,
       subscriberPreview,
       supporterNewsletter,
     ),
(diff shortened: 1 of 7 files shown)


handlers/soft-opt-in-consent-setter/src/test/scala/com.gu.soft_opt_in_consent_setter/ConsentsCalculatorTests.scala
```diff
@@ -131,14 +131,18 @@ class ConsentsCalculatorTests extends AnyFlatSpec with should.Matchers with Eith
...
+        |  {
         |    "id" : "supporter_newsletter",
         |    "consented" : false
         |  },
         |  {
         |    "id" : "guardian_weekly_newsletter",
         |    "consented" : true
         |  }
-        |]""".stripMargin)
+        |]""".stripMargin) // be careful this is whitespace sensitive
```
johnduffell, line 145: unless whitespace is actually important, just add a Json.parse or equivalent on both sides of the `shouldBe` equation.  I think usually the error message is as good if not better when it's been parsed, but worth confirming perhaps.

handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/models/ConsentsMapping.scala
```diff
@@ -64,33 +74,39 @@ object ConsentsMapping {
...
       yourSupportOnboarding,
       similarGuardianProducts,
+      guardianProductsAndServices,
       supporterNewsletter,
       subscriberPreview,
     ),
     "FeastInAppPurchase" -> Set(
       yourSupportOnboarding,
       similarGuardianProducts,
+      guardianProductsAndServices,
```
johnduffell, line 109: given that they all have these three consents, I feel like it's almost time to pull it out as a val rather than duplicating?

johnduffell (approved):
I assume this will both set and clear down the consents at the right time?  I am not sure if it's all handled here or if there's a separate lambda?

handlers/soft-opt-in-consent-setter/src/test/scala/com.gu.soft_opt_in_consent_setter/ConsentsCalculatorTests.scala
Joemgru: Good shout - on revisiting I noticed some of the tests already use a helper method called `removeWhiteSpace` so I have put those on the tests that didn't have it

handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/models/ConsentsMapping.scala
Joemgru: I don't understand the domain enough to come up with a name - is there a logical reason for grouping them ?

Joemgru:
> I assume this will both set and clear down the consents at the right time? I am not sure if it's all handled here or if there's a separate lambda?

Good shout - I have made a corresponding change in the autolapse lambda
https://github.com/guardian/consent-autolapse/pull/20

handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/models/ConsentsMapping.scala
johnduffell: just to reduce du[plication if they are bascially consents for all paid products.  Then it's easier to see the specific ones and less chance of a copy and paste error

johnduffell: could be just called `paidProductOptIns`

johnduffell:
just for reference, this new consent is not to be autolapsed, so it would only be turned off by identity.

handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/models/ConsentsMapping.scala
Joemgru:
```
  private val paidProductOptIns = Set(yourSupportOnboarding, similarGuardianProducts, guardianProductsAndServices)

  val consentsMapping: Map[String, Set[String]] = Map(
    "Membership" -> (paidProductOptIns union Set(
      supporterNewsletter,
    )),
    "Supporter" -> (paidProductOptIns union Set(
      supporterNewsletter,
    )), 
```

As they all (but one) have to union the new set - I think I prefer the duplication as it's clearer. Happy to change it if you are strongly in favour of the val?

johnduffell: fair enough if you've tried both and prefer the current then leave it as it is!

MaelGNM:
We are simplifying things and only keeping one permission. This is no longer needed.

PR #2477, opened 2024-09-27
Add some variation to the responseValidUntil date returned by the identity-retention lambda

graham228221:
#### Background
A key rotation [issue](https://mail.google.com/chat/u/0/#chat/space/AAAAIddGn5w/VlhvlPnNQZM/VlhvlPnNQZM) in March 2024 meant that a large number of accounts were not deleted as planned.  Once the issue was fixed, there was a special run done to catch up all the missed accounts.  The accounts are checked every 3 months from the last check, meaning that the re-check date for the caught up ones fell on the same day as each other in late June.

The Identity team have [recently observed](https://mail.google.com/chat/u/0/#chat/space/AAAAIddGn5w/twb0ZzTaY9g/Fu6889OyCnA) that their Identity account retention lambda did not delete any accounts for several days, raising their alarms. In the linked conversation, it was determined that this was likely due to the large number of lapsed relationship accounts that were impacted by the issue in late-March 2024 are now hitting their three month “check again” responseValidUntil date - as Identity's lambda can only process 7k or so accounts per day, it was not able to clear the backlog quickly enough and this led to it not deleting any accounts for several days.

Although we have scaled up to process them all, these have again got a date in 3 months time, and this will recur for up to 7 years if we do nothing.

####  Changes
We introduce some arbitrary variation in the responseValidUntil date, which should allow the backlog of lapsed relationship accounts to be somewhat spread out rather than appearing all within a few days. This should help avoid the backlog breaching the 7k per day limit (which is also being increased by allowing the Identity lambda to complete more runs each day).

## How to test
Updated IdentityRetentionApiResponsesTest to demonstrate the varied dates.

The next predicted spike in requests (late December?) this logic should kick in and help disperse future requests.

We should give the Identity team a heads up that we're making this change, and ask them to let us know if there's any odd behaviour observed from their side.

### DIFF
--- a/handlers/identity-retention/src/main/scala/com/gu/identityRetention/IdentityRetentionResponses.scala
+++ b/handlers/identity-retention/src/main/scala/com/gu/identityRetention/IdentityRetentionResponses.scala
@@ -1,6 +1,7 @@
 package com.gu.identityRetention
 
 import java.time.LocalDate
+import java.time.temporal.ChronoUnit
 import com.gu.util.apigateway.ResponseModels.ApiResponse
 import play.api.libs.json.{Json, Writes}
 
--- a/handlers/identity-retention/src/main/scala/com/gu/identityRetention/IdentityRetentionResponses.scala
+++ b/handlers/identity-retention/src/main/scala/com/gu/identityRetention/IdentityRetentionResponses.scala
@@ -34,12 +34,13 @@ object IdentityRetentionResponseModels {
       val responseValidUntil =
         if (effectiveDeletionDate isBefore today)
           today.plusMonths(3)
-        else
+        else {
           val proportionOfMaxLifetimes = today.until(effectiveDeletionDate, ChronoUnit.DAYS) / (365 * 7)
           List(
             effectiveDeletionDate, 
             today.plusDays(60 + (proportionOfMaxLifetimes * 60)) //adds some variation so that the validUntil date is between 60 and 120 days (or longer, if the deletion date is over 7 years in the future)
           ).min
+        }
       SuccessResponse(ongoingRelationship, relationshipEndDate, effectiveDeletionDate, responseValidUntil)
     }
   }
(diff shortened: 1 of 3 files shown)


handlers/identity-retention/src/main/scala/com/gu/identityRetention/IdentityRetentionResponses.scala
```diff
@@ -33,8 +34,13 @@ object IdentityRetentionResponseModels {
       val responseValidUntil =
         if (effectiveDeletionDate isBefore today)
           today.plusMonths(3)
-        else
-          List(effectiveDeletionDate, today.plusMonths(3)).min
+        else {
+          val proportionOfMaxLifetimes = today.until(effectiveDeletionDate, ChronoUnit.DAYS) / (365.25 * 7)
```
johnduffell, line 38: would be nice to put the max days in a const?  I assume the original 7 years const is not stored in this codebase.
```suggestion
          val maxLifeInDays = 365.25 * 7
          val proportionOfMaxLifetimes = today.until(effectiveDeletionDate, ChronoUnit.DAYS) / maxLifeInDays
```

handlers/identity-retention/src/main/scala/com/gu/identityRetention/IdentityRetentionResponses.scala
```diff
@@ -33,8 +34,13 @@ object IdentityRetentionResponseModels {
...
+          List(
+            effectiveDeletionDate, 
+            today.plusDays(60 + (proportionOfMaxLifetimes * 60).toInt) //adds some variation so that the validUntil date is between 60 and 120 days (or longer, if the deletion date is over 7 years in the future)
```
johnduffell, line 41: good comment

johnduffell (approved):
fab looks good just a couple of nice to have comments

PR #2485, opened 2024-10-08
don't get a billing preview for cancelled subscriptions

johnduffell:
This PR makes sure that the cancelled status is checked before we attempt to get a billing preview

This bug was introduced when the code was refactored to detect an existing discount based on the billing preview.

It was causing a lot of 500 errors `no invoice items in preview`

https://trello.com/c/i4CmGdkT/542-discount-api-mitigation-for-cancelled-subs


johnduffell:
after feedback from the team, it was decided that we should get the billing period lazily, rather than splitting the standard validation across two functions.
Closing this in favour of https://github.com/guardian/support-service-lambdas/pull/2490

PR #2487, opened 2024-10-09
add log links to alarm chat messages

johnduffell:
This PR adds log links to alarm chat messages.
It assumes that logs are /aws/lambda/<app>-<stage> by default, but it can be overridden if that is not the case.

Possible future improvement would be to link to other things like step function execution logs.


handlers/alarms-handler/src/alarmMappings.ts
```diff
@@ -17,7 +18,8 @@ const sharedMobilePurchasesApps = [
 	'mobile-purchases-google-update-subscriptions',
 ];
 
-const teamToAppMappings: Record<Team, string[]> = {
+type AppInfo = string | { app: string; logGroups: string[] };
```
johnduffell, line 21: now an app can have either just a name, or it can have a name and a list of associated log groups.
In future, rather than hard coding, this could maybe be generated by a call to AWS to find all log groups associated with lambdas/ec2 instances.

handlers/alarms-handler/src/alarmMappings.ts
```diff
@@ -107,33 +115,81 @@ const teamToAppMappings: Record<Team, string[]> = {
...
-	const mappings: Record<string, Team[]> = {};
-
-	for (const [team, apps] of Object.entries(teamToAppMappings)) {
-		for (const app of apps) {
-			const teams = mappings[app] ?? [];
-			teams.push(team as Team);
-
-			mappings[app] = teams;
-		}
+export class AlarmMappings {
```
johnduffell, line 118: I made this into a class so I could use simplified mappings in the tests more easily

handlers/alarms-handler/src/alarmMappings.ts
```diff
@@ -107,33 +115,81 @@ const teamToAppMappings: Record<Team, string[]> = {
...
-	if (appName && appToTeamMappings[appName]) {
-		return appToTeamMappings[appName] as Team[];
-	}
+	private buildAppToTeamMappings = (
+		theMappings: Record<Team, AppInfo[]>,
+	): Record<string, Team[]> => {
+		const entries: Array<[Team, AppInfo[]]> = Object.entries(theMappings) as Array<[
+			Team,
+			AppInfo[],
+		]>; // `as` - hmm?
```
johnduffell, line 130: for some reason calling `Object.entries` on a `Record<Team, AppInfo[]>` gives `Array<[string, AppInfo[]]>` so I had to give the type system the info back again.  Not sure if this is the safest way of doing it though?

handlers/alarms-handler/src/alarmMappings.ts
```diff
@@ -107,33 +115,81 @@ const teamToAppMappings: Record<Team, string[]> = {
...
+			Object.entries(groups).map(([app, info]) => [
+				app,
+				info.map(({ team }) => team),
+			]),
+		);
+
+		return mappings;
+	};
+
+	private buildAppToLogGroupOverrides = (
```
johnduffell, line 151: if an app is in multiple teams with different log groups, it will just pick one arbitrarily.
I could have got around it by having separate app->group mapping but I didn't think that would be as maintainable.  It does mean the type system doesn't help us out 100% though.

handlers/alarms-handler/src/alarmMappings.ts
```diff
@@ -107,33 +115,80 @@ const teamToAppMappings: Record<Team, string[]> = {
...
-export const getTeams = (appName?: string): Team[] => {
-	if (appName && appToTeamMappings[appName]) {
-		return appToTeamMappings[appName] as Team[];
-	}
+	private buildAppToTeamMappings = (
+		theMappings: Record<Team, AppInfo[]>,
+	): Record<string, Team[]> => {
+		const entries: Array<[Team, AppInfo[]]> = Object.entries(
+			theMappings,
+		) as Array<[Team, AppInfo[]]>; // `as` - hmm?
```
johnduffell, line 129: for some reason calling Object.entries on a Record<Team, AppInfo[]> gives Array<[string, AppInfo[]]> so I had to give the type system the info back again. Not sure if this is the safest way of doing it though?

handlers/alarms-handler/src/alarmMappings.ts
```diff
@@ -71,11 +73,17 @@ const teamToAppMappings: Record<Team, string[]> = {
...
+		{ app: 'frontend', logGroups: ['support-frontend'] },
 		'it-test-runner',
 		'stripe-intent',
-		'workers',
-		'payment-api',
+		{
+			app: 'workers',
+			logGroups: [
+				'/aws/lambda/CreatePaymentMethod',
+				'/aws/lambda/CreateZuoraSubscription', //etc
```
johnduffell, line 83: in future it might be useful to have a link to the state machine executions logs also, rather than just to the lambdas.

handlers/alarms-handler/src/alarmMappings.ts
```diff
@@ -107,33 +119,80 @@ const teamToAppMappings: Record<Team, string[]> = {
...
+				}),
+		);
+		const groups = groupBy(teamToApp, ({ app }) => app);
+
+		const mappings: Record<string, Team[]> = Object.fromEntries(
+			Object.entries(groups).map(([app, info]) => [
+				app,
+				info.map(({ team }) => team),
+			]),
+		);
```
johnduffell, line 149: tempted to make an object mapValues function - if there isn't already such a thing?
```suggestion
		const mappings: Record<string, Team[]> = 
			mapValues(
			  groups,
			  (info) => info.map(({ team }) => team),
			)
```

AndreaDiotallevi:
Interesting addition!

I have a comment regarding where to get the log group from.

I appreciate this is a prototype, but I am not sure placing the log group info in the alarm mapping (making it more complex) is the right place.

The alarm mapping is used for ownership and it should stay simple I believe.

If we talk about "change". If for instance that app log group changes, what is the reason of the change? It is not ownership, it is likely a change in application code. So it should be changed in the producer application code. Probably it should be stored in an alarm tag (e.g. `logGroupName`) (one to one mapping alarm <-> log group name) and retrieved at run time by the alarm handler.

johnduffell:
Thanks Andrea - after a bit of chat the solution is going to be to look for a `LogGroupNames` tag on the alarm and build the links from there.  Then it's completely opt-in and also keeps the info about the alarm with the alarm, rather than making this lambda know more than it should.

handlers/alarms-handler/src/alarmMappings.ts
AndreaDiotallevi: With Typescript you can use both OOP or functional style but at the Guardian we usually use a functional style with typescript.

So in this scenario I am not sure about the benefits or changing this part of the code to use a class since the PR does not change the behaviour of this part of the code (how mappings work).

handlers/alarms-handler/src/alarmMappings.ts
AndreaDiotallevi: Same comment as above - I am not sure why this needs changing.

handlers/alarms-handler/src/cloudwatch.ts
```diff
@@ -49,24 +49,25 @@ const buildCloudwatchClient = (awsAccountId: string): CloudWatchClient => {
...
-	alarmArn: string,
-	awsAccountId: string,
-): Promise<string | undefined> => {
-	const tags = await getTags(alarmArn, awsAccountId);
-	return tags.find((tag: Tag) => tag.Key === 'App')?.Value;
+	const tags = response.Tags ?? [];
+	const entries = tags.flatMap((tag: Tag) =>
+		tag.Key && tag.Value ? [[tag.Key, tag.Value]] : [],
+	);
+	return Object.fromEntries(entries) as Tags;
```
AndreaDiotallevi, line 72: Nice improvement!

handlers/alarms-handler/src/index.ts
```diff
@@ -10,30 +10,33 @@ const cloudWatchAlarmMessageSchema = z.object({
...
+		.optional(),
 });
 
 type CloudWatchAlarmMessage = z.infer<typeof cloudWatchAlarmMessageSchema>;
 
 export const handler = async (event: SQSEvent): Promise<void> => {
 	try {
+		const alarmMappings = new AlarmMappings();
 		for (const record of event.Records) {
-			console.log(record);
```
AndreaDiotallevi, line 20: What is the benefit of changing the way we handle CloudWatch alarms v SNS publish messages?

What I expected was just to just update the `handleCloudWatchAlarmMessage` function, by using the new tag and modifying the text message.

handlers/alarms-handler/src/index.ts
```diff
@@ -42,59 +45,127 @@ export const handler = async (event: SQSEvent): Promise<void> => {
...
+	diagnosticUrlTemplate: string,
+	Trigger:
+		| {
+				Period: number;
+				EvaluationPeriods: number;
+		  }
+		| undefined,
+	StateChangeTime: Date,
+) {
+	const assumedTimeForCompositeAlarms = 300;
```
AndreaDiotallevi, line 132: Interesting!

AndreaDiotallevi:
Thanks John for taking in the discussion and experimenting with the new tag.

I have just left a few more comments on coding paradigms between OOP and functional with Typescript and whether the refactoring should be done in a separate PR or even if we needed it at all.

See them as suggestions more than anything.

handlers/alarms-handler/src/index.ts
johnduffell: thanks for the review, so far I had a first pass at making the change we agreed (to use tags rather than hard coded), but I should have first made the PR into draft so it's clear it's back to in progress.  I will go through it again  myself and review it before it's ready.

Regarding your question, this change was to make it more testable as I could just check the return value rather than having to mock http calls, however I will have to confirm whether that test is still needed once I've made the updated.

handlers/alarms-handler/src/index.ts
johnduffell: I don't think it would be a help - just compostite alarms do not have a specific period reported from SNS.  So it would just choose a default period.  It might be possible to fetch the alarm details from AWS CLI and work something out, if it appears to cause confusion.

handlers/alarms-handler/src/alarmMappings.ts
johnduffell: ok yes thanks for that, I get super confused between all the different ways to write a function/method.  The idea behind making it parameterised rather than a singleton was so I could write a test that used mocked data rather than having to use the real data (which might get edited).
In scala I usually use classes because then you get better help from the compiler and IDE, but I will see how to boil it down to a function.

handlers/alarms-handler/src/alarmMappings.ts
johnduffell: this was because I changed it from a simple team->app mapping to a team->(app|{app, loglink[]}) mapping so I could extract the log links.
When I wrote the log link extractor I couldn't work out the equivalent to groupMap in typescript so I ended up with a mess!   I think the groupMap implementation is there in a later commit.  It made sense to update the existing mappings extractor to be consistent but now that it's not needed I will have a look to see if it's still adding value.

johnduffell:
> Thanks John for taking in the discussion and experimenting with the new tag.
> 
> I have just left a few more comments on coding paradigms between OOP and functional with Typescript and whether the refactoring should be done in a separate PR or even if we needed it at all.
> 
> See them as suggestions.

thanks for the comments, despite that it's still WIP I think this is helpful when I get it ready for another review.  I think there's probably some scala thinking coming through, for better or worse, but I will look through when I get more time and let you know when it's ready

johnduffell:
closed in favour of the above "referenced" PRs

PR #2490, opened 2024-10-10
lazy billing preview to avoid previewing cancelled subscriptions

johnduffell:
This PR replaces https://github.com/guardian/support-service-lambdas/pull/2485

There was an issue where we are getting a billing preview before checking if a subscription is cancelled.  Since there is no billing preview possible for a cancelled subscription, this was wasteful and caused an error.

In the above PR I split up the basic validation, but after review by the team it was decided to keep the logic together and make the billing preview called lazily.

### DIFF
--- a/handlers/discount-api/src/discountEndpoint.ts
+++ b/handlers/discount-api/src/discountEndpoint.ts
@@ -23,6 +23,7 @@ import type { APIGatewayProxyEventHeaders } from 'aws-lambda';
 import dayjs from 'dayjs';
 import { EligibilityChecker } from './eligibilityChecker';
 import { generateCancellationDiscountConfirmationEmail } from './generateCancellationDiscountConfirmationEmail';
+import { Lazy } from './lazy';
 import { getDiscountFromSubscription } from './productToDiscountMapping';
 
 export const previewDiscountEndpoint = async (
@@ -196,14 +197,25 @@ async function getDiscountToApply(
 		subscription.subscriptionNumber,
 	);
 
-	console.log('get billing preview for the subscription');
-	const billingPreview = billingPreviewToSimpleInvoiceItems(
-		await getBillingPreview(
-			zuoraClient,
-			today.add(13, 'months'),
-			subscription.accountNumber,
-		),
+	const lazyBillingPreview = new Lazy(
+		() =>
+			getBillingPreview(
+				zuoraClient,
+				today.add(13, 'months'),
+				subscription.accountNumber,
+			),
+		'get billing preview for the subscription',
+	).map(billingPreviewToSimpleInvoiceItems);
+
+	const lazyNextInvoiceItems = lazyBillingPreview.map(getNextInvoiceItems);
+
+	await eligibilityChecker.assertGenerallyEligible(
+		subscription,
+		account.metrics.totalInvoiceBalance,
+		lazyNextInvoiceItems.map(({ items }) => items),
 	);
+
+	const billingPreview = await lazyBillingPreview.get();
 
 	console.log('Working out the appropriate discount for the subscription');
 	const { discount, discountableProductRatePlanId } =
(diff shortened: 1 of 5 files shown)


handlers/discount-api/src/discountEndpoint.ts
```diff
@@ -196,15 +197,26 @@ async function getDiscountToApply(
...
+
+	const lazyNextInvoiceItems = lazyBillingPreview.map(getNextInvoiceItems);
+
+	await eligibilityChecker.assertGenerallyEligible(
+		subscription,
+		account.metrics.totalInvoiceBalance,
+		lazyNextInvoiceItems.map(({ items }) => items),
 	);
 
+	const billingPreview = await lazyBillingPreview.get();
```
johnduffell, line 218: now that `assertGenerallyEligible` is complete it's safe to get the billing preview out of the lazy promise.

handlers/discount-api/src/discountEndpoint.ts
```diff
@@ -196,15 +197,26 @@ async function getDiscountToApply(
...
-		),
+	const lazyBillingPreview = new Lazy(
+		() =>
+			getBillingPreview(
+				zuoraClient,
+				today.add(13, 'months'),
+				subscription.accountNumber,
+			),
+		'get billing preview for the subscription',
+	).map(billingPreviewToSimpleInvoiceItems);
```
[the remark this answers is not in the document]
johnduffell: as discussed we will rename `map` to `then` and also add a test for its behaviour

handlers/discount-api/src/discountEndpoint.ts
```diff
@@ -196,15 +198,26 @@ async function getDiscountToApply(
...
+				today.add(13, 'months'),
+				subscription.accountNumber,
+			),
+		'get billing preview for the subscription',
+	).then(billingPreviewToSimpleInvoiceItems);
+
+	await eligibilityChecker.assertGenerallyEligible(
+		subscription,
+		account.metrics.totalInvoiceBalance,
+		() => lazyBillingPreview.then(getNextInvoiceItems).get(),
```
johnduffell, line 215: @tjmw I tried out your suggestion of unwrapping the Lazy at this point, and it did make the code a bit nicer as it is a bit more transparent. 👏

handlers/discount-api/test/eligibilityChecker.test.ts
```diff
@@ -44,10 +48,10 @@ test('Eligibility check fails for a Supporter plus which has already had the off
 		eligibilityChecker.assertGenerallyEligible(
 			sub,
 			0,
-			getNextInvoiceItems(billingPreview).items,
+			asLazy(getNextInvoiceItems(billingPreview)),
```
johnduffell, line 51: note that I changed `getNextInvoiceItems` to only return the items array, not including the date, as most places only needed the items.
The places that (also) need the date, can call the new `getNextInvoice` function.

handlers/discount-api/src/discountEndpoint.ts
tjmw: Cool! I wonder now that we're getting laziness from passing an arrow function, if we need the `Lazy` construct here? We're now: wrapping the billing preview in a `Lazy`, and then unwrapping with `get()`, but the whole construct is lazy because of the extra level of function application.

Either way, happy to go with this and see how we like `Lazy` in the codebase.

handlers/discount-api/src/eligibilityChecker.ts
```diff
@@ -10,10 +10,10 @@ import dayjs from 'dayjs';
 export class EligibilityChecker {
 	constructor(private subscriptionNumber: string) {}
 
-	assertGenerallyEligible = (
+	assertGenerallyEligible = async (
 		subscription: ZuoraSubscription,
 		accountBalance: number,
-		nextInvoiceItems: SimpleInvoiceItem[],
+		lazyNextInvoiceItems: () => Promise<SimpleInvoiceItem[]>,
```
tjmw, line 16: I'd be tempted to call this something like `getNextInvoiceItems`, but obviously subjective!

tjmw (approved):
👍

handlers/discount-api/src/discountEndpoint.ts
johnduffell: yeah I did play with that a bit but the important thing is that the billing preview is called exactly once, (also the .then function gets rid of some messy boilerplate wrapping and unwrapping the function)
Having it self contained means that we can write some simple unit tests for it (once) and then know that the code won't be broken by a refactor.

handlers/discount-api/src/eligibilityChecker.ts
johnduffell: yep that's a good idea - done

PR #2502, opened 2024-10-15
add extra csr auth error

johnduffell:
We have noticed that holiday stop api sometimes fails with a 500 alarm for Invalid session id, as well as the expired password error handled in https://github.com/guardian/support-service-lambdas/pull/2458

This PR adds that error to the list of handled ones, to return a 400 instead of 500 error.

The CSRs know to log out and in if it happens.

### DIFF
--- a/handlers/holiday-stop-api/src/main/scala/com/gu/holiday_stops/Handler.scala
+++ b/handlers/holiday-stop-api/src/main/scala/com/gu/holiday_stops/Handler.scala
@@ -105,7 +105,7 @@ object Handler extends Logging {
     for {
       sfClient <- SalesforceClient.auth(response, config.sfConfig) match {
         case Right(success) => ContinueProcessing(success)
-        case Left(errors) if errors.exists(errors => csrAuthErrors.contains(errors.errorCode)) =>
+        case Left(errors) if errors.exists(error => csrAuthErrors.contains(error.errorCode)) =>
           logger.error(s"SF token was not valid - returning 400 authenticate with SalesForce: $errors")
           ReturnWithResponse(badRequest("salesforce returned auth error"))
         case Left(error) =>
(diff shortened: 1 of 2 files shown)


handlers/holiday-stop-api/src/main/scala/com/gu/holiday_stops/Handler.scala
```diff
@@ -103,9 +105,9 @@ object Handler extends Logging {
     for {
       sfClient <- SalesforceClient.auth(response, config.sfConfig) match {
         case Right(success) => ContinueProcessing(success)
-        case Left(errors) if errors.exists(_.errorCode == "INVALID_OPERATION_WITH_EXPIRED_PASSWORD") =>
+        case Left(errors) if errors.exists(errors => csrAuthErrors.contains(errors.errorCode)) =>
           logger.error(s"SF token was not valid - returning 400 authenticate with SalesForce: $errors")
-          ReturnWithResponse(badRequest("salesforce returned INVALID_OPERATION_WITH_EXPIRED_PASSWORD"))
+          ReturnWithResponse(badRequest("salesforce returned auth error"))
```
johnduffell, line 110: not sure if there's any value in keeping the actual error?

handlers/holiday-stop-api/src/main/scala/com/gu/holiday_stops/Handler.scala
```diff
@@ -103,9 +105,9 @@ object Handler extends Logging {
     for {
       sfClient <- SalesforceClient.auth(response, config.sfConfig) match {
         case Right(success) => ContinueProcessing(success)
-        case Left(errors) if errors.exists(_.errorCode == "INVALID_OPERATION_WITH_EXPIRED_PASSWORD") =>
+        case Left(errors) if errors.exists(errors => csrAuthErrors.contains(errors.errorCode)) =>
```
tjmw, line 108: Tiny nitpick, but in the predicate shouldn't the yielded value be singular?

```suggestion
        case Left(errors) if errors.exists(error => csrAuthErrors.contains(error.errorCode)) =>
```

tjmw (approved):
Nice!

PR #2503, opened 2024-10-15
Only start running digital-voucher-suspension-processor at 2am

graham228221:
## What does this change?
Changes the schedule for the digital-voucher-suspension-processor so that it starts running hourly from 2am UTC. 

This is due to a change on our partner's side where they have started preventing new suspensions for subscriptions that are currently under a suspension. A suspension is only lifted for a subscription by a cron job on their side running at 1am, so even though our suspension requests are not duplicates they are often failing if a subscription had a suspension the day before.

## How to test
View rule in Eventbridge and verify that the schedule looks as expected, starting from 2am.

We should stop seeing errors from our partner API caused by subscriptions being suspended 

This Lambda is typically processing 50-100 suspensions at a day at most, and can typically complete these in a single run, so starting a little later shouldn't be an issue.

### DIFF
--- a/handlers/digital-voucher-suspension-processor/cfn.yaml
+++ b/handlers/digital-voucher-suspension-processor/cfn.yaml
@@ -145,8 +145,8 @@ Resources:
     Type: AWS::Events::Rule
     DependsOn: SuspensionLambda
     Properties:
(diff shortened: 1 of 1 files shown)


AndreaDiotallevi (approved):
Nice investigation

PR #751, opened 2020-11-06
Replace custom billing preview with out-of-the-box one in holiday-stop-api

mario-galic:
## What does this change?

* Similar to https://github.com/guardian/support-service-lambdas/pull/743
* https://trello.com/c/i0a10FyJ/1542-replace-billing-date-calculation-in-support-service-lambdas-with-calls-to-billing-preview-endpoint
* Start replacing https://github.com/guardian/support-service-lambdas/pull/550 with out-of-the-box functionality
* Start wiring in https://github.com/guardian/invoicing-api/pull/23
* Note for now the new functionality will run alongside the old one so we can validate by comparing the two calculations. Once validated I will address the `FIXME`s.

## How to test

* Let it run in PROD for a while and see if `testInProdPreviewPublications` ever logs the following error

```
ERROR com.gu.holiday_stops.Handler$ - testInProdPreviewPublications failed A-S0000000?startDate=2020-11-13&endDate=2020-12-25 because List(PotentialHolidayStop(2020-11-29,Credit(-3.0,2020-12-29)), PotentialHolidayStop(2020-12-06,Credit(-3.0,2020-12-29)), PotentialHolidayStop(2020-12-13,Credit(-3.0,2020-12-29)), PotentialHolidayStop(2020-12-20,Credit(-3.0,2020-12-29))) =/= List(PotentialHolidayStop(2020-11-29,Credit(-3.0,2020-12-29)), PotentialHolidayStop(2020-12-06,Credit(-3.0,2020-12-29)), PotentialHolidayStop(2020-12-13,Credit(-3.0,2020-12-29)), PotentialHolidayStop(2020-12-20,Credit(-3.0,2020-12-29))) or 2020-12-29 =/= 2020-11-29
```

## How can we measure success?

After few days if no errors are logged we can begin to switch over to new functionality.

## Have we considered potential risks?

The risks have been mitigated by deploying in steps, that is, not immediately switching over to the new calculation and yet exercising the new paths at the same time.

PR #2504, opened 2024-10-15
remove invoicing api call from holiday stop processor

johnduffell:
We get a lot of errors from the holiday stop API, because it makes calls to the invoicing API.

The reason it makes the calls is that there was a migration in progress to make it more accurately determine the invoice date - the original (unfinished) work to switch over was in https://github.com/guardian/support-service-lambdas/pull/751
The issue still needs fixing but we have decided on a different route - see https://github.com/guardian/support-service-lambdas/pull/2243 for the WIP on that.

The reason this is affecting us now is that the API key needed was rotated/deleted at some point thus breaking the link and causing errors in the logs.  This is misleading and makes it hard to investigate.

The long term solution is to finish the above PR, but for now this PR just deletes the unused call to invoicing api.

Tested in CODE, all unit tests pass.


graham228221 (approved):
🙌  nice one

PR #2507, opened 2024-10-16
include sub id and product in all log lines

johnduffell:
## problem
at the moment it's hard to filter down the logs across executions (e.g. linking the preview and apply calls, or linking a failed attempt with an earlier or later successful attempt)

Now that we support S+, Digisub and RC in both monthly and annual, it also makes sense to be able to filter the logs by product

# this pr
This PR prefixes each log line with the subscription name (once we have parsed it from the request) and product (actually rate plan name once we have got the subscription JSON from zuora).

image

# other info
This is similar to the changes in MDAPI to show the identity ID on all log statements, although with a different way to implement it rather than implicits https://github.com/guardian/members-data-api/pull/1064

I'll admit it's not as typesafe and referentially transparent as the scala version, and it's harder to stop people hitting console.log by mistake, but it's better to have context most of the time than none.

# testing
Tested in CODE as above.


handlers/discount-api/src/discountEndpoint.ts
```diff
@@ -28,23 +30,32 @@ import { Lazy } from './lazy';
...
 	stage: Stage,
 	headers: APIGatewayProxyEventHeaders,
 	subscriptionNumber: string,
 	today: dayjs.Dayjs,
 ) => {
-	const zuoraClient = await ZuoraClient.create(stage);
+	const zuoraClient = await ZuoraClient.create(stage, logger);
 
 	const { subscription, account } = await getSubscriptionIfBelongsToIdentityId(
+		logger.log.bind(logger),
```
johnduffell, line 42: I didn't want to pass the whole Logger object around unnecessarily, however if you try to pass the method you get a TS error without the `.bind(logger)` https://typescript-eslint.io/rules/unbound-method/

handlers/discount-api/src/discountEndpoint.ts
```diff
@@ -187,16 +209,15 @@ async function getSubscriptionIfBelongsToIdentityId(
...
+	logger: Logger,
 	stage: Stage,
 	subscription: ZuoraSubscription,
 	account: ZuoraAccount,
 	zuoraClient: ZuoraClient,
 	today: dayjs.Dayjs,
 ) {
 	const catalog = () => getZuoraCatalog(stage);
-	const eligibilityChecker = new EligibilityChecker(
-		subscription.subscriptionNumber,
```
johnduffell, line 198: the subscription number was just used for logging context, but now we get that for free from the logger

handlers/discount-api/src/eligibilityChecker.ts
```diff
@@ -69,15 +70,12 @@ export class EligibilityChecker {
 		subscription: ZuoraSubscription,
 		now: Dayjs,
 	) => {
-		const eligibilityChecker = new EligibilityChecker(
-			subscription.subscriptionNumber,
-		);
```
johnduffell, line 74: must have been a copy and paste error, we didn't need to create an EligibilityChecker from one of its methods!

handlers/discount-api/src/index.ts
```diff
@@ -22,43 +23,48 @@ const stage = process.env.STAGE as Stage;
 export const handler: Handler = async (
 	event: APIGatewayProxyEvent,
 ): Promise<APIGatewayProxyResult> => {
-	console.log(`Input is ${JSON.stringify(event)}`);
-	const response = await routeRequest(event);
-	console.log(`Response is ${JSON.stringify(response)}`);
+	const logger = new Logger();
+	logger.log(`Input is ${JSON.stringify(event)}`);
```
johnduffell, line 27: we won't (yet) have any context for this log line

rBangay (approved):
👍

PR #2508, opened 2024-10-17
feat: always process all suspensions and throw error at the end if there are failures

AndreaDiotallevi:
## What does this change?

[Trello](https://trello.com/c/Nmfkwoxo/324-%F0%9F%9A%A8-alarm-urgent-9-5-prod-failed-to-suspend-digital-voucher-subscriptions-has-triggered)

At the moment, if there are 50 suspensions to process and the first fails, an error is thrown straight away.

This update allows to always process all suspensions and only throw the error at the end if there are failures.

Big thanks to @johnduffell for pairing on this!

## How to test

We have tested this on PROD (CODE is broken) and it worked nicely.


graham228221 (approved):
Thanks very much for this! 

In production today, we have 97 subscriptions that are due to be suspended.

Before this change only 2 subscriptions out of 97 were being suspended before it hit a failure and quit trying the rest. 

With your change, 94 out of 97 subscriptions are now successfully suspended with the other 3 failing due to a problem on the third party API side. Much better!

This will also help a lot in future as often subscriptions are blocked individually from being suspended for various reasons, which would block subsequent suspensions in the same daily batch from being attempted. But this change will ensure that we always suspend the maximum number of subscriptions possible 👍

rupertbates (approved):
Looks good to me

PR #2513, opened 2024-10-18
Press reader entitlements API

andrewHEguardian:
## What does this change?

Create a new API gateway + lambda to serve user entitlements to Press Reader app. We accept a user ID, which we will swap for an identity ID and query supporter product data to work out whether someone is entitled to use the app. The response is an XML doc.

[Trello card](https://trello.com/c/zK8tEETk/1134-editions-app-migration-to-pressreader-platform-api-build)
image

## How to test

Make a request to https://press-reader-entitlements-code.support.guardianapis.com/user-entitlements with a user ID query param and the api key as a header.


handlers/press-reader-entitlements/src/identity.ts
```diff
@@ -0,0 +1,53 @@
+import { GetParameterCommand, SSMClient } from '@aws-sdk/client-ssm';
+import { awsConfig } from '@modules/aws/config';
+import type { Stage } from '@modules/stage';
+import { getIdentityIdSchema } from './schemas';
+
+export async function getIdentityClientAccessToken() {
+	const ssmClient = new SSMClient(awsConfig);
```
andrewHEguardian, line 7: Should we put this outside the function, the same way we're doing with the `dynamoClient`?

rupertbates: I've put the fetching of the client access token into a `Lazy` value so that it will only be fetched once

andrewHEguardian: Nice! That's very cool. Should we do the same for loading the product catalog?

rupertbates: Yes good idea

rupertbates (approved):

PR #2524, opened 2024-10-24
Fix new-lambda script

rupertbates:
## What does this change?
Since we moved the cdk module to use pnpm, the new-lambda script has been broken as it explicitly uses Yarn. This PR fixes that.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
 		"clean": "rm -rf .parcel-cache && pnpm -r run clean",
 		"test": "pnpm --stream -r run test",
 		"it-test": "pnpm --stream -r run it-test",
-		"new-lambda": "pnpm hygen new-lambda api-gateway && pnpm install && cd ./cdk && yarn lint --fix && yarn test -u && git add ./lib/__snapshots__/*.test.ts.snap",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -9,7 +9,7 @@
 		"clean": "rm -rf .parcel-cache && pnpm -r run clean",
 		"test": "pnpm --stream -r run test",
 		"it-test": "pnpm --stream -r run it-test",
-		"new-lambda": "pnpm hygen new-lambda api-gateway && pnpm install && cd ./cdk && yarn lint --fix && yarn test -u && git add ./lib/__snapshots__/*.test.ts.snap",
+		"new-lambda": "pnpm hygen new-lambda api-gateway && pnpm install && pnpm --filter cdk lint --fix && pnpm fix-formatting && pnpm --filter cdk test -- -u && git add ./cdk/lib/__snapshots__/*.test.ts.snap",
```
tjmw, line 12: No harm either way, but do we want to explicitly fix formatting of only the cdk workspace?

```suggestion
		"new-lambda": "pnpm hygen new-lambda api-gateway && pnpm install && pnpm --filter cdk lint --fix && pnpm --filter cdk fix-formatting && pnpm --filter cdk test -- -u && git add ./cdk/lib/__snapshots__/*.test.ts.snap",
```

tjmw (approved):
Looks good 👍

package.json
rupertbates: We shouldn't need to do it for any of the files really, but for some reason the templates kept failing the formatting test in the cdk directory so I gave up and added this in!

PR #2535, opened 2024-10-31
more logging and better searchability for update supporter plus amount lamdba

johnduffell:
We tried to investigate someone who tried to change their S+ amount 4 times and it failed every time as it was trying to back date the amendment to when they switched from annual contribution to supporter plus.  See example at https://eu-west-1.console.aws.amazon.com/cloudwatch/home?region=eu-west-1#logsV2:log-groups/log-group/$252Faws$252Flambda$252Fupdate-supporter-plus-amount-PROD/log-events/2024$252F10$252F27$252F$255B$2524LATEST$255Dad4e0a97414b4fdf9ca2e72ca92a8711$3Fstart$3D1730065591579$26refEventId$3D38581751936778500848853083164490340067694877865427861506

We did notice that they are one of the "split" S+ (see https://docs.google.com/document/d/1kDwJ13984jhv4fwkqtHo43upMvYZEZWyfHNSkaYUJbk/edit ) however we don't think that's a particular issue.

We tried going through the log line by line and it was clear that it should have picked the right rate plan and the right charge and date.  We even pasted the data into a unit test and it still seemed to work ok!

As a last resort, this PR adds a load more logging so we can have an idea of where it went wrong.  This also brings in the logger context from this pr https://github.com/guardian/support-service-lambdas/pull/2507 which will only include subscription ID in this lambda (as product is always S+).

Tested in CODE and it seems ok with dummy data via postman.

### DIFF
--- a/handlers/update-supporter-plus-amount/src/updateSupporterPlusAmount.ts
+++ b/handlers/update-supporter-plus-amount/src/updateSupporterPlusAmount.ts
@@ -1,3 +1,4 @@
+import { getSingleOrThrow } from '@modules/arrayFunctions';
 import { ValidationError } from '@modules/errors';
 import { getIfDefined } from '@modules/nullAndUndefined';
 import { prettyPrint } from '@modules/prettyPrint';
@@ -8,8 +9,10 @@ import type {
 	ProductRatePlan,
 } from '@modules/product-catalog/productCatalog';
 import { isProductCurrency } from '@modules/product-catalog/productCatalog';
+import { zuoraDateFormat } from '@modules/zuora/common';
 import { getAccount } from '@modules/zuora/getAccount';
 import { getSubscription } from '@modules/zuora/getSubscription';
+import type { Logger } from '@modules/zuora/logger';
 import type { ZuoraClient } from '@modules/zuora/zuoraClient';
 import type {
 	RatePlan,
@@ -42,6 +45,7 @@ type ProductData = {
 };
 
 export const getSupporterPlusData = (
+	logger: Logger,
 	productCatalog: ProductCatalog,
 	ratePlans: RatePlan[],
 ): SupporterPlusData => {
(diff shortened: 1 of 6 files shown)


handlers/update-supporter-plus-amount/src/updateSupporterPlusAmount.ts
```diff
@@ -71,44 +74,48 @@ export const getSupporterPlusData = (
...
-		);
-	}
-	return updatableChargeData[0];
+	const activeRatePlans = ratePlans.filter(
+		(ratePlan) => ratePlan.lastChangeType !== 'Remove',
+	);
+	logger.log(
+		'active rate plans',
+		activeRatePlans.map((rp) => rp.id),
+	);
```
johnduffell, line 83: I changed this code from a map/filter chain to use intermediate consts that are more useful for logging. This means we can see in the logs how it's narrowing down the subscription data into the specific charges.

handlers/update-supporter-plus-amount/src/updateSupporterPlusAmount.ts
```diff
@@ -71,44 +74,48 @@ export const getSupporterPlusData = (
...
+	const relevantRatePlans = activeRatePlans.flatMap((ratePlan) => {
```
johnduffell, line 84: I wanted to use `collect` but that doesn't seem to be a thing in typescript.  Although simpler in concept, it would be a toss up between collect, flatMap and map then filter in terms of readability since typescript doesn't really have partial functions

handlers/update-supporter-plus-amount/src/updateSupporterPlusAmount.ts
```diff
@@ -71,44 +74,48 @@ export const getSupporterPlusData = (
...
+	});
+	logger.log(
+		'relevant rate plans',
+		relevantRatePlans.map((rp) => rp.ratePlan.id),
+	);
+
+	const relevantRatePlan = getSingleOrThrow(
+		relevantRatePlans,
+		(msg) => new Error(`wrong number of supporter plus rate plans: ${msg}`),
+	);
```
johnduffell, line 98: updated to use the existing `getSingleOrThrow` function rather than custom logic

handlers/update-supporter-plus-amount/src/updateSupporterPlusAmount.ts
```diff
@@ -190,25 +199,31 @@ export const updateSupporterPlusAmount = async (
 			? newPaymentAmount - supporterPlusData.productRatePlan.pricing[currency]
 			: newPaymentAmount;
 
+	const { chargeToUpdate } = supporterPlusData;
+
 	const applyFromDate = dayjs(
-		supporterPlusData.chargeToUpdate.chargedThroughDate ?? // If the currently active charge is the one that was invoiced last
-			supporterPlusData.chargeToUpdate.effectiveStartDate, // If there is a pending amendment
+		chargeToUpdate.chargedThroughDate ?? // If the currently active charge is the one that was invoiced last
+			chargeToUpdate.effectiveStartDate, // If there is a pending amendment
```
johnduffell, line 206: this is the logc that seemed to be malfunctioning, it seemed to be using the effectiveStartDate instead of the chargedThroughDate even though it was there in the json

ecguardian (approved):
👍  LGTM

PR #2546, opened 2024-11-13
stop running cdk tests and linting on every push

johnduffell:
I was working on another project recently and noticed the pushes were really quick.  I noticed they take around 20s on this project even with trivial changes.  After looking into why they are so slow, it seems like we are running CDK linting and tests on every push.

This PR changes it so it only runs when you change cdk/* so otherwise push finishes before I even click away (1-2 seconds).

Script courtesy of @AndreaDiotallevi via AI!

Original PR adding the check is here https://github.com/guardian/support-service-lambdas/pull/2199

### DIFF
--- a/.husky/pre-push
+++ b/.husky/pre-push
@@ -1,1 +1,0 @@
-pnpm --filter cdk lint && pnpm --filter cdk test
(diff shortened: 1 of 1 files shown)


AndreaDiotallevi:
For me it was the opposite.

I was working with CDK and I lost so much time waiting for the CI to fail in GitHub.

So I'd say, we can improve it by running CDK tests only if files inside the CDK folder have changed?

PR #2551, opened 2024-11-14
User correct stage for SSM parameter

rupertbates:
## What does this change?
#2513 incorrectly hard coded the stage in the SSM parameter name which it uses to load the identity client access token, this PR fixes that.

It has meant moving the code which fetches the token to the top level of the application so that it is aware of the stage at start up.


andrewHEguardian (approved):
Good catch 👍

PR #2553, opened 2024-11-15
Add patrons to product catalog

rupertbates:
## What does this change?
This PR adds some products to the product catalog
### The membership products:
- Supporter Membership
- Partner Membership
- Patron Membership
These are no longer sold but we still have active subscriptions to them. To reflect this a new `active` boolean has been added to the product object in the catalog which flags whether a product is still on sale or not.

### Guardian Patrons:
This is a product which does not exist in Zuora, it is billed through Stripe. This is a bit more of a complex change as it means integrating another billing system into the code which generates the product catalog. 
To keep track of which product is billed in which system there is a new flag `billingSystem` on the product object in the product catalog.

### DIFF
--- a/modules/product-catalog/src/stripeProducts.ts
+++ b/modules/product-catalog/src/stripeProducts.ts
@@ -1,0 +1,34 @@
+export const stripeProducts = {
+	GuardianPatron: {
+		billingSystem: 'stripe',
+		ratePlans: {
+			GuardianPatron: {
+				id: 'guardian_patron',
+				pricing: {
+					GBP: 0,
+					USD: 0,
+					NZD: 0,
+					EUR: 0,
+					AUD: 0,
+					CAD: 0,
+				},
+				charges: {
+					GuardianPatron: { id: 'guardian_patron' },
+				},
+				billingPeriod: 'Month',
+			},
+		},
+	},
+};
+
+export const stripeTypeObject = {
+	GuardianPatron: {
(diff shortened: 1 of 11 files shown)


modules/product-catalog/src/typeObject.ts
```diff
@@ -4,7 +4,7 @@ export const typeObject = {
 		billingPeriods: ['Month'],
 		productRatePlans: {
 			Monthly: {
-				GuardianLight: {},
+				Subscription: {},
```
rupertbates, line 7: Just a bit of standardisation, all the other products with only one charge name the charge `Subscription` in the product catalog (not in Zuora)

modules/product-catalog/src/stripeProducts.ts
```diff
@@ -0,0 +1,47 @@
...
+} from '@modules/product-catalog/productCatalog';
+
+export const stripeProducts: Partial<Record<ProductKey, Product<ProductKey>>> =
+	{
+		GuardianPatron: {
+			billingSystem: 'stripe',
+			active: true,
+			ratePlans: {
+				GuardianPatron: {
+					id: 'guardian_patron',
```
andrewHEguardian, line 16: Would this ratePlanId be recorded in SupporterProductData?

rupertbates: Yes it is.

andrewHEguardian (approved):

PR #2555, opened 2024-11-18
Press reader error handling

rupertbates:
## What does this change?
Currently we are returning a 500 if the press reader API is called with a user id which does not exist. This is a user error rather than an application error so we should return a 400 response.

We are also alarming on every 400 response, which is not necessary. This PR increases the threshold for triggering that alarm.

Finally it tidies up a few formatting issues in one of the hygen templates


tjmw (approved):
👍

PR #2561, opened 2024-11-18
Get all product currencies in type object

rupertbates:
The product-catalog module has a type object which allows us to generate a `ProductCurrency<P extends Product>` type, which in turn lets us validate that a given currency is valid for a particular product.

Previously the code which generates this type was only checking the currencies which were supported on the first product rate plan of any product, on the assumption that they would be the same for all rate plans, however this assumption was incorrect - for some products supported currencies can vary between rate plans.

This PR updates the `product.currencies` to contain all currencies supported by a product across all rate plans.


rupertbates:
Closed in favour of #2562 2562

PR #2562, opened 2024-11-19
Remove ProductCurrency type

rupertbates:
## What does this change?
Previously we had a type in the product-catalog Typescript module called `ProductCurrency<P>` which allowed us to check whether a given currency was valid for a particular product. 

However this was not actually accurate because there are some products where different rate plans allow different currencies, for instance T3. 

To make this accurate would mean making the type into `ProductRatePlanCurrency<P, PRP>` where PRP is the product rate plan type. This starts to become quite unwieldy and in fact wasn't possible to derive from the type object which we use to derive all the other types. 

Given that the check was only used in two places, I have decided to remove the product element of the currency, and just have a list of supported currencies instead.

------------------------------

## For future reference
The issue with deriving the product rate plan currency from a type object is this:
``` typescript
export const typeObject = {
	GuardianLight: {
		billingPeriods: ['Month'],
		productRatePlans: {
			Monthly: {
				currencies: { GBP: {} },
			},
		},
	},
} as const;

type TypeObject = typeof typeObject;

export type ProductKey = keyof TypeObject;

export type ProductRatePlanKey<P extends ProductKey> =
	keyof TypeObject[P]['productRatePlans']; // At the first level we are able to get the productRatePlans keys

export type ProductCurrency<
	P extends ProductKey,
	PRP extends ProductRatePlanKey<P>,
> = keyof TypeObject[P]['productRatePlans'][PRP]['currencies']; // At the next level down we cannot get the currencies keys 
// - error TS2536: Type '"currencies"' cannot be used to index type '{ readonly GuardianLight: { readonly billingPeriods: readonly ["Month"]; readonly productRatePlans: { readonly Monthly: { readonly currencies: { readonly GBP: {}; }; }; }; }; }[P]["productRatePlans"][PRP]'.
```

### Update - managed to resolve this issue with the following changes:
I've still decided to go ahead with this PR though as there were a couple of other type issues that were non-trivial to fix

``` Typescript
export const typeObject = {
	GuardianLight: {
		billingPeriods: ['Month'],
		productRatePlans: {
			Monthly: {
				currencies: { GBP: {}, USD: {} },
			},
		},
	},
} as const;

type TypeObject = typeof typeObject;

export type ProductKey = keyof TypeObject;

export type ProductRatePlanKey<P extends ProductKey> =
	keyof TypeObject[P]['productRatePlans'];

export type ProductRatePlanCurrency<
	P extends ProductKey,
	PRP extends ProductRatePlanKey<P>,
> = keyof (TypeObject[P]['productRatePlans'][PRP] & {
	currencies: Record<string, unknown>;
})['currencies'];

export const currency: ProductRatePlanCurrency<'GuardianLight', 'Monthly'> =
	'USD';
```

### DIFF
--- a/modules/internationalisation/src/currency.ts
+++ b/modules/internationalisation/src/currency.ts
@@ -1,0 +1,30 @@
+export const CurrencyValues = [
+	'GBP',
+	'EUR',
+	'AUD',
+	'USD',
+	'CAD',
+	'NZD',
+] as const;
+export type Currency = (typeof CurrencyValues)[number];
+
+export const isSupportedCurrency = (
+	maybeCurrency: string,
+): maybeCurrency is Currency => {
+	return (CurrencyValues as readonly string[]).includes(maybeCurrency);
+};
+
+export const getCurrencyGlyph = (currency: Currency) => {
+	switch (currency) {
+		case 'GBP':
+			return '£';
+		case 'EUR':
+			return '€';
+		case 'AUD':
+		case 'CAD':
+		case 'NZD':
+		case 'USD':
+			return '$';
+	}
+	throw new Error(`Unsupported currency ${currency}`);
+};
--- a/modules/internationalisation/src/currency.ts
+++ b/modules/internationalisation/src/currency.ts
@@ -14,17 +14,14 @@ export const isSupportedCurrency = (
 	return (CurrencyValues as readonly string[]).includes(maybeCurrency);
 };
 
-export const getCurrencyGlyph = (currency: Currency) => {
-	switch (currency) {
-		case 'GBP':
-			return '£';
-		case 'EUR':
-			return '€';
-		case 'AUD':
-		case 'CAD':
-		case 'NZD':
-		case 'USD':
-			return '$';
-	}
-	throw new Error(`Unsupported currency ${currency}`);
+const currencyToGlyphMapping: { [C in Currency]: string } = {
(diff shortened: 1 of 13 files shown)


modules/internationalisation/src/currency.ts
```diff
@@ -0,0 +1,30 @@
+export const CurrencyValues = [
```
tjmw, line 1: Not necessarily something for now, but I wonder if there's a slightly different implementation where we define a mapping of currency => symbol like this:

```js
const CurrencySymbolMapping = {
        'GBP' => '£',
	'EUR' => '€',
	'AUD' => '$',
	'USD' => '$',
	'CAD' => '$',
	'NZD' => '$',
}
```

And that's the single/canonical place where this information is defined. That way it'd be impossible to say add a country code but forget to add a symbol. TypeScript seems to know when the switch in `getCurrencyGlyph` is exhaustive (right now it tells me the line where we throw `Unsupported currency` is unreachable, but that goes away if I remove any of the `case`s) but it doesn't warn or error on it.

Just a thought, definitely not a blocker.

tjmw (approved):
Nice! One minor question/suggestion, but definitely not a blocker.

modules/internationalisation/src/currency.ts
rupertbates: Yup good idea

PR #2568, opened 2024-11-25
Script to synchronise supporter product data with Zuora

rupertbates:
## What does this change?
From time to time we find users where the synchronisation between Zuora and the SupporterProductData DynamoDB table which powers members-data-api has failed, meaning that there is no record in Dynamo for a valid subscription. 

This PR introduces a script which can be run with an identity id to fix that synchronisation for a given user. It will find all accounts that belong to the provided identity id, then all the subscriptions linked to those accounts, and put a message onto the supporter-product-data SQS queue where they will be picked up and inserted into Dynamo.

The script is idempotent, so can be run multiple times without causing any issues even if the data is already in sync.

## Usage

```bash
pnpm --filter sync-supporter-product-data sync-user [environment] [identityId]
```
Eg.
```bash
pnpm --filter sync-supporter-product-data sync-user CODE 200065441


graham228221 (approved):
This looks really useful - as a future development could we have this available as an API endpoint? User Help / CSRs sometimes need to "fix" subscriptions that are not linked to identity ids properly and this would be helpful for them to complete fixes entirely without dev intervention (if we could have, say, a button in Salesforce that would call this endpoint). 

Potentially we incorporate it into [this lambda](https://github.com/guardian/support-service-lambdas/tree/main/handlers/identity-backfill)? This "backfills" Identity IDs into Salesforce and Zuora, and was originally used for migrating Identity IDs into Zuora years back but can still be used for ad hoc fixes.

rupertbates:
> This looks really useful - as a future development could we have this available as an API endpoint? User Help / CSRs sometimes need to "fix" subscriptions that are not linked to identity ids properly and this would be helpful for them to complete fixes entirely without dev intervention (if we could have, say, a button in Salesforce that would call this endpoint).
> 
> Potentially we incorporate it into [this lambda](https://github.com/guardian/support-service-lambdas/tree/main/handlers/identity-backfill)? This "backfills" Identity IDs into Salesforce and Zuora, and was originally used for migrating Identity IDs into Zuora years back but can still be used for ad hoc fixes.

We can definitely do that, it should be pretty straightforward. If you want to pair on it sometime I'd be happy to do that.

PR #2570, opened 2024-11-26
Add Newspaper Voucher and Old GW products to product catalog

rupertbates:
## What does this change?
This PR adds various products and rate plans into the product catalog, namely:
- Newspaper Vouchers 
- The old zoned Guardian Weekly products (Zone A, B & C)
- The print + digital rate plans to all newspaper products which have them (all except national delivery)

None of these products or rate plans are available for sale any more, but we do have customers with active subscriptions to them. 

In the past the product catalog was mainly used for acquisitions so there was no need to include inactive products in it, but now it is increasingly being used to identify the products held by a customer, so it is important that it includes all products which we have active subscribers for.

This PR also gives editions access to newspaper voucher customers.


handlers/press-reader-entitlements/src/supporterProductData.ts
```diff
@@ -69,6 +69,7 @@ export function getLatestValidSubscription(
 		'HomeDelivery',
 		'NationalDelivery',
 		'SubscriptionCard',
+		'NewspaperVoucher',
```
rupertbates, line 72: Give editions access to voucher customers

modules/product-catalog/src/productCatalogSchema.ts
```diff
@@ -391,6 +458,268 @@ export const productCatalogSchema = z.object({
 			}),
 		}),
 	}),
+	SubscriptionCard: z.object({
```
rupertbates, line 461: I've reodered this schema to group products of a similar type together, this has made the diff look bigger than it really is.

modules/product-catalog/src/productCatalogSchema.ts
```diff
@@ -341,6 +341,73 @@ export const productCatalogSchema = z.object({
...
+				id: z.string(),
+				pricing: z.object({ GBP: z.number() }),
+				charges: z.object({
+					Saturday: z.object({ id: z.string() }),
+					Monday: z.object({ id: z.string() }),
+					Sunday: z.object({ id: z.string() }),
+					Friday: z.object({ id: z.string() }),
+					Wednesday: z.object({ id: z.string() }),
+					Tuesday: z.object({ id: z.string() }),
+					Thursday: z.object({ id: z.string() }),
```
tjmw, line 354: Total nitpick but what do you think about ordering the days Mon-Sun?

modules/product-catalog/src/productCatalogSchema.ts
```diff
@@ -341,6 +341,73 @@ export const productCatalogSchema = z.object({
 					.enum(typeObject.HomeDelivery.billingPeriods)
 					.optional(),
 			}),
+			'Everyday+': z.object({
```
tjmw, line 344: Tangential: I wonder if at some point it'd be useful to have an `active` attribute on rate plans to reflect their status in the same way we do for products?

tjmw (approved):
Looks great 👍

modules/product-catalog/src/productCatalogSchema.ts
rupertbates: Yes I think it would, I just wasn't 100% sure of the best way to do it so I've left it for now.

PR #2571, opened 2024-11-27
Separate supporterProductData.ts into its own module

rupertbates:
## What does this change?
#2511 introduced some code to access the supporter-product-data Dynamo table, but it was in the lambda module. This PR moves it out to a stand alone module so that it can be accessed by other lambdas.

### DIFF
--- a/modules/supporter-product-data/test/supporterProductDataIntegration.test.ts
+++ b/modules/supporter-product-data/test/supporterProductDataIntegration.test.ts
@@ -1,0 +1,6 @@
+import { getSupporterProductData } from '@modules/supporter-product-data/supporterProductData';
+
+test('Dynamo Integration', async () => {
+	const supporterData = await getSupporterProductData('CODE', '110001137');
+	expect(supporterData?.length).toEqual(4);
+});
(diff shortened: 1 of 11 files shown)


modules/supporter-product-data/test/supporterProductDataIntegration.test.ts
```diff
@@ -0,0 +1,6 @@
+import { getSupporterProductData } from '@modules/supporter-product-data/supporterProductData';
+
+test('Dynamo Integration', async () => {
+	const supporterData = await getSupporterProductData('CODE', '110001137');
+	expect(supporterData?.length).toEqual(4);
```
tjmw, line 5: Does this test actually talk to `CODE`? What's the significance of the `4` here?

tjmw (approved):
Looks good 👍 

One question, but not a blocker (especially since the code isn't new, it's just moved).

modules/supporter-product-data/test/supporterProductDataIntegration.test.ts
rupertbates: Yes it queries a real user in CODE and there are 4 items returned. As this is an integration test it is mostly used for local testing, it doesn't get run in CI.

PR #2572, opened 2024-11-27
User benefits API

rupertbates:
## What does this change?
This PR adds a new API Gateway triggered lambda to return information about the digital benefits or entitlements that a user has as a result of the subscriptions they have with us. Justification for this work is in [this document](https://docs.google.com/document/d/1k4JDClyNWLjXESxp31MUq43lfd_2b89Y0iTJkwCYiSQ/edit?tab=t.0#heading=h.mor0122zxwk6) 

The url of the new api will be `user-benefits.guardianapis.com/benefits/me`, it will be authenticated via an Okta JWT header using the `identity` module.


handlers/user-benefits/src/index.ts
```diff
@@ -0,0 +1,82 @@
...
+	if (!(event.path === '/benefits/me' && event.httpMethod === 'GET')) {
+		return {
+			body: 'Not Found',
+			statusCode: 404,
+		};
+	}
+	try {
+		const authenticatedEvent = await identityAuthenticator.authenticate(event);
+		if (authenticatedEvent.type === 'FailedAuthenticationResponse') {
+			return authenticatedEvent;
```
tjmw, line 55: It took me a moment to realise that `authenticatedEvent` has `statusCode` and `body` properties (I think possibly only when the type is `FailedAuthenticationResponse`?). I wonder if it'd be worth adding a comment here making that clear? Or what if the `statusCode` and `body` were nested in a property of `authenticatedEvent`, so you'd return something like `authenticatedEvent.httpUnauthorizedResponse` or similar?

modules/product-benefits/test/getUserBenefits.test.ts
```diff
@@ -0,0 +1,23 @@
...
+	expect(getBenefits(['GuardianLight'])).toEqual(['rejectTracking']);
+	expect(getBenefits(['SupporterPlus'])).toEqual(supporterPlusBenefits);
+	expect(getBenefits(['TierThree'])).toEqual(tierThreeBenefits);
+	expect(getBenefits(['GuardianWeeklyDomestic'])).toEqual(['fewerSupportAsks']);
+	expect(getBenefits([])).toEqual([]);
+});
+
+test('getUserBenefits returns distinct benefits', () => {
+	expect(getBenefits(['TierThree', 'DigitalSubscription'])).toEqual(
+		tierThreeBenefits,
```
tjmw, line 21: Is there another test which could be added two products which have overlapping, but also distinct, sets of benefits? Just thinking this one alone wouldn't catch an accidental regression where only the benefits of the first product are returned. What do you reckon?

tjmw (approved):
This looks really good! I left a few comments/suggestions - see what you think.

What's the traffic for this going to look like? I wonder if longer term we'll need to move off of a lambda (great use case for getting this into prod quickly though).

handlers/user-benefits/src/index.ts
rupertbates: Does adding the type explicitly make it clearer?
```typescript
const maybeAuthenticatedEvent:
	| AuthenticatedApiGatewayEvent
	| FailedAuthenticationResponse =
	await identityAuthenticator.authenticate(event);
```

PR #2573, opened 2024-11-29
Identity module

rupertbates:
## What does this change?
This PR adds a module to handle authenticating API Gateway requests using an Okta token set by Identity.

See readme for more information.


modules/identity/src/apiGateway.ts
```diff
@@ -0,0 +1,85 @@
...
+		event: APIGatewayProxyEvent,
+	): Promise<AuthenticatedApiGatewayEvent | FailedAuthenticationResponse> {
+		const authHeader = event.headers.Authorization;
+		if (!authHeader) {
+			console.log('No Authorization header provided in request', event);
+			return {
+				type: 'FailedAuthenticationResponse',
+				statusCode: 403,
+				body: JSON.stringify({ message: 'No Authorization header provided' }),
+			};
```
coldlink, line 36: No authorization header should be a 401

```suggestion
			return {
				type: 'FailedAuthenticationResponse',
				statusCode: 401,
				body: JSON.stringify({ message: 'No Authorization header provided' }),
			};
```

modules/identity/src/identity.ts
```diff
@@ -0,0 +1,109 @@
+import { ValidationError } from '@modules/errors';
+import type { Stage } from '@modules/stage';
+import type { JwtClaims } from '@okta/jwt-verifier';
+import OktaJwtVerifier from '@okta/jwt-verifier';
+
+// The claims object returned by Okta. It should include the identity_id
+interface JwtClaimsWithIdentityID extends JwtClaims {
+	legacy_identity_id?: string;
+	braze_uuid?: string;
```
coldlink, line 9: `braze_uuid` isn't always expected to be in the token.

I see it's optional here, but I don't see any checks/use for it.

modules/identity/src/identity.ts
```diff
@@ -0,0 +1,109 @@
...
+				this.config.audience,
+			);
+		} else {
+			throw new InvalidTokenError('Invalid Authorization header');
+		}
+	};
+	getIdentityId = async (authHeader: string): Promise<string> => {
+		try {
+			const jwt = await this.verifyAccessToken(authHeader);
+			console.log(`Verified access token: ${JSON.stringify(jwt)}`);
```
coldlink, line 78: Probably best not to log the entire jwt server side, if you need to log a specific value then I suggest you can log the `jti` claim on the token, which is a unique identifier for that particular token.

Unless this was for debugging purposes and can be removed?

modules/identity/test/identity.test.ts
```diff
@@ -0,0 +1,120 @@
...
...di10aGVndWFyZGlhbi5jb20vIiwic3ViIjoicnVwZXJ0LmJhdGVzK3QzQG9ic2VydmVyLmNvLnVrIiwiaWF0IjoxNzMyODAxNzE5LCJleHAiOjE3MzI4MjMzMTksImNpZCI6IjBvYTUzeDZrNXdHWVhPR3ptMHg3IiwidWlkIjoiMDB1ODFyenVjdG16QTRKeHcweDciLCJzY3AiOlsiZ3VhcmRpYW4uaWRlbnRpdHktYXBpLm5ld3NsZXR0ZXJzLnVwZGF0ZS5zZWxmIiwiZ3VhcmRpYW4uaWRlbnRpdHktYXBpLnVzZXIudXNlcm5hbWUuY3JlYXRlLnNlbGYuc2VjdXJlIiwiZ3VhcmRpYW4uZGlzY3Vzc2lvbi1hcGkucHJpdmF0ZS1wcm9maWxlLnJlYWQuc2VsZiIsImlkX3Rva2VuLnByb2ZpbGUudGhlZ3VhcmRpYW4iLCJvcGVuaWQiLCJwcm9maWxlIiwiZ3VhcmRpYW4uZGlzY3Vzc2lvbi1hcGkudXBkYXRlLnNlY3VyZSIsImd1YXJkaWFuLmlkZW50aXR5LWFwaS5uZXdzbGV0dGVycy5yZWFkLnNlbGYiLCJndWFyZGlhbi5tZW1iZXJzLWRhdGEtYXBpLnJlYWQuc2VsZiIsImVtYWlsIl0sImF1dGhfdGltZSI6MTczMjcyNjY4OCwiaWRlbnRpdHlfdXNlcm5hbWUiOiIiLCJlbWFpbF92YWxpZGF0ZWQiOnRydWUsImxlZ2FjeV9pZGVudGl0eV9pZCI6IjIwMDI2NDQwNCIsImVtYWlsIjoicnVwZXJ0LmJhdGVzK3QzQG9ic2VydmVyLmNvLnVrIn0.QkcH3_2psLJ1482K00ePQDbSEEhx2ObrJHSpsnpYOfRrBPd_OZpn6nBXm_nbxmx9YscIictqOmJw92Xz5kwPruA98BlHZ_C_dFui1DWkvZv6KbcDpUU4AM0oWi0pptTg8zqpMU-b0gAI_28hQG2S45Q8sVhyuUX_VruBsD-bpGkSmmgTis5uc2ZmrFekkeKwvnZxxsmmqgwJxkfeykIMWcEPCzPpKMZs4IikvlGyGBgW-E8ECMRt2WtStg32MRFF_kNmLqB5SqrXfZ7kcIm4-BYwNPZU-oXO7TMgnS0a4rLOxDKKU_ALQCupMF-JxoQ-pHbyoyMsY9-V-78sFZ61HA';
```
coldlink, line 27: Best not to include entire jwts in the codebase even if they are expired/testing.

Could load the expired one from an environment variable, and have the user manually inpout the valid one?

coldlink (changes requested):
Great implementation, this will be amazing for future reference.

Left a few minor comments!

modules/identity/test/identity.test.ts
coldlink: It's *probably* fine though.

rupertbates: I've added some config in to hold the test values which can be downloaded from S3

coldlink (approved):

PR #2583, opened 2024-12-10
Give Guardian staff access to the PressReader editions app

rupertbates:
## What does this change?
This PR updates the press-reader-entitlements api to grant access to guardian staff using the email address field which is now being returned from identity since [this PR](https://github.com/guardian/identity/pull/2654)

### DIFF
--- a/handlers/press-reader-entitlements/src/index.ts
+++ b/handlers/press-reader-entitlements/src/index.ts
@@ -1,8 +1,8 @@
 import { ValidationError } from '@modules/errors';
+import { Lazy } from '@modules/lazy';
 import { getIfDefined } from '@modules/nullAndUndefined';
 import { userHasGuardianEmail } from '@modules/product-benefits/userBenefits';
 import { getProductCatalogFromApi } from '@modules/product-catalog/api';
-import type { ProductCatalog } from '@modules/product-catalog/productCatalog';
 import type { Stage } from '@modules/stage';
 import type { SupporterRatePlanItem } from '@modules/supporter-product-data/supporterProductData';
 import type {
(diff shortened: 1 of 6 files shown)


handlers/press-reader-entitlements/src/index.ts
```diff
@@ -1,22 +1,29 @@
...
-import { getClientAccessToken, getIdentityId } from './identity';
+import { getClientAccessToken, getUserDetails } from './identity';
 import { getLatestSubscription } from './supporterProductData';
 import type { Member } from './xmlBuilder';
 import { buildXml } from './xmlBuilder';
 
 const stage = process.env.STAGE as Stage;
-let productCatalog: ProductCatalog | undefined = undefined;
-let identityClientAccessToken: string | undefined = undefined;
+const lazyProductCatalog = new Lazy(
```
rupertbates, line 19: I have changed this initialisation to use the Lazy construct because it allows us to use `await` at the top level which the project is otherwise not set up to do. Once we can configure support-service-lambdas to allow top level await then we can remove it.

handlers/press-reader-entitlements/test/pressReaderEntitlementsIntegration.test.ts
```diff
@@ -21,7 +21,7 @@ test('Entitlements check', async () => {
 test('getIdentityId', async () => {
 	const accessToken = await getClientAccessToken('CODE');
 	expect(accessToken).toBeDefined();
-	const identityId = await getIdentityId(
+	const identityId = await getUserDetails(
```
tjmw, line 24: Does this still pass? Isn't the return value a `Promise<IdentityUserDetails>` now? (Below we're asserting that it matches the given identity ID).

tjmw (approved):
Looks good, with one question about a test which I think might now be failing!

handlers/press-reader-entitlements/test/pressReaderEntitlementsIntegration.test.ts
rupertbates: Good spot 👍

PR #2585, opened 2024-12-11
don't offer a pause to contribitions with zero amount

johnduffell:
We noticed that errors were being thrown when a zero-contributor tried to cancel (and thus called the preview-discount endpoint)
This shouldn't have prevented them from cancelling, as manage would handle this, but it caused an error/alarm for us.

This PR changes it so that it notices a zero invoice, and doesn't offer a discount using a normal check.


handlers/discount-api/test/eligibilityChecker.test.ts
```diff
@@ -52,7 +53,9 @@ test('Eligibility check fails for a Supporter plus which has already had the off
 			asLazy(getNextInvoiceItems(billingPreview)),
 		);
 
-	expect(actual).rejects.toThrow(validationRequirements.noNegativePreviewItems);
+	await expect(actual).rejects.toThrow(
```
johnduffell, line 56: we noticed that if you miss the `await` call it failed the tests overall, but didn't tell you which one had the issue.  This makes it show the specific test as failing.

PR #2587, opened 2024-12-13
Add one time contributions into the product catalog

rupertbates:
## What does this change?
We want single or one-time contributions to be represented as a product in the product catalog so that we can identify users with this product in the supporter product data database.

This PR also adds one time contribution to the product-benefits code as contributors receive the 'fewerSupportAsks' benefit. This is a bit more complex for one time than it is for any other product however as the benefit is only given for three months after the contribution date. 

I have tried to add this additional validation in a way which is extensible for other products if they were to require additional validation (for instance in the past we had benefits which only applied if a product was purchased before a particular date).


modules/product-benefits/src/userBenefits.ts
```diff
@@ -25,15 +27,29 @@ export const getUserProducts = async (
...
+): ProductKey[] =>
+	supporterProductDataItems
+		.flatMap((item) => {
+			const product = productCatalogHelper.findProductDetails(
+				item.productRatePlanId,
+			)?.zuoraProduct;
+
+			if (
+				product !== undefined &&
+				validationFunctionForProduct(product)(item)
```
tjmw, line 45: Since we always immediately invoke the result of `validationFunctionForProduct(product)` I wonder if it'd be simpler to remove the extra layer of function application here? I.e. just `validationForProduct(product, item)`? (Where `validationForProduct` just calls the underlying validation function, rather than returning it).

modules/product-catalog/src/productCatalogSchema.ts
```diff
@@ -8,14 +8,7 @@ export const productCatalogSchema = z.object({
...
 				id: z.string(),
-				pricing: z.object({
-					GBP: z.number(),
-					USD: z.number(),
-					NZD: z.number(),
-					EUR: z.number(),
-					AUD: z.number(),
-					CAD: z.number(),
-				}),
+				pricing: z.object({}),
```
tjmw, line 11: Is this part of the one time contribution change?

tjmw (approved):
Looks good. I left a couple of minor questions but no blockers.

modules/product-benefits/src/userBenefits.ts
rupertbates: Yeah I did think about that, I think you may be right.

PR #2588, opened 2024-12-13
Add some tests for the user benefits handler

tjmw:
## What does this change?

I'm planning on making some changes to the user benefits handler, including to the routing, so add some tests. This involved a little bit of setup to make jest work in this workspace.

## How to test

```
$ pnpm --filter user-benefits test
```

## Accessibility

### DIFF
--- a/handlers/user-benefits/src/index.ts
+++ b/handlers/user-benefits/src/index.ts
@@ -1,5 +1,5 @@
 import { ValidationError } from '@modules/errors';
-import { IdentityApiGatewayAuthenticator } from '@modules/identity/apiGateway';
+import { buildAuthenticate } from '@modules/identity/apiGateway';
 import type { IdentityUserDetails } from '@modules/identity/identity';
 import { Lazy } from '@modules/lazy';
 import type { UserBenefitsResponse } from '@modules/product-benefits/schemas';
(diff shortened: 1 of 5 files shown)


modules/identity/src/apiGateway.ts
```diff
@@ -84,3 +84,17 @@ export class IdentityApiGatewayAuthenticator {
 		}
 	}
 }
+
+export const buildAuthenticate = (
```
tjmw, line 88: I've added this to simplify the interface from the user benefits handler perspected (it makes mocking a bit easier).

handlers/user-benefits/src/index.ts
```diff
@@ -44,7 +40,7 @@ const getUserBenefitsResponse = async (
 	};
 };
 
-export const handler: Handler = async (
```
tjmw, line 47: I've removed this so we can be slightly looser in the tests about the args specified (strictly speaking according to the `Handler` definition, we should invoke `handler` with two additional args which we don't actually need). This is consistent with other lambda definitions in this repo.

rupertbates (approved):

handlers/user-benefits/test/index.test.ts
```diff
@@ -0,0 +1,60 @@
...
+}));
+jest.mock('@modules/product-catalog/api', () => ({
+	getProductCatalogFromApi: () => ({}),
+}));
+jest.mock('@modules/product-benefits/userBenefits', () => ({
+	getUserBenefits: () => ['adFree'],
+}));
+
+describe('handler', () => {
+	it('returns a 404 for an unrecognized path or HTTP method', async () => {
```
rupertbates, line 31: Probably don't need this now, it's already tested in the router module

tjmw: Discussed offline: we're not migrated over to the router module yet, but that's the next step (and I'll remove this test then).

PR #2595, opened 2024-12-16
Ensure that 4xx response bubble up correctly from identity

tjmw:
## What does this change?

Previously API gateway wasn't happy with the `type` property which was added to the response resulting in a 502 to the client. This only existed so that internally we could tell what kind of response we'd got back. Now the response for a failed auth request is nested, so the type doesn't appear in the response we give back to API gateway.

## How to test

- [x] Run the integration tests
- [x] Check in CODE

## Accessibility

### DIFF
--- a/modules/identity/test/identity.test.ts
+++ b/modules/identity/test/identity.test.ts
@@ -81,46 +81,45 @@ test('we throw an error if the token is expired', async () => {
 const authenticator = new IdentityApiGatewayAuthenticator('CODE', []);
 
 test('ApiGateway event with a valid token returns an authenticated result', async () => {
-	const response = await authenticator.authenticate(
+	const result = await authenticator.authenticate(
 		buildProxyEvent(validAuthHeader),
 	);
-	console.log(response);
-	expect(response.type).toBe('AuthenticatedApiGatewayEvent');
-	if (response.type !== 'AuthenticatedApiGatewayEvent') {
+	expect(result.type).toBe('success');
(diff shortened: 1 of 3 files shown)


modules/identity/test/identity.test.ts
```diff
@@ -81,59 +81,58 @@ test('we throw an error if the token is expired', async () => {
...
 test('ApiGateway event with expired token returns a 401 response', async () => {
-	const response = await authenticator.authenticate(
+	const result = await authenticator.authenticate(
 		buildProxyEvent(expiredAuthHeader),
 	);
-	expect(response.type).toBe('FailedAuthenticationResponse');
-	if (response.type !== 'FailedAuthenticationResponse') {
-		throw new Error('Expected response to be a FailedAuthenticationResponse');
+	expect(result.type).toBe('failure');
+	if (result.type !== 'failure') {
```
tjmw, line 99: I'm not sure we'd reach this line if the assertion above failed. I'll have a go at removing once I've run the integration tests.

rupertbates: I think that's there to keep Typescript happy with accessing result.response.statusCode below.

rupertbates (approved):
Nice, I really like this solution.

modules/identity/test/identity.test.ts
tjmw: Ahh that makes sense, thanks 👍

PR #2597, opened 2024-12-16
Disable deploy comments

rupertbates:
## What does this change?
This PR updates the 2 CI workflows which build the various scala and typescript lambdas, to use a new flag available to the riff-raff action which prevents deploy comments being added to the PR. 

The reason for this is that due to the large number of projects in this repo, the deploy comments would get very noisy and it was easy to miss comments from contributors in amongst all the noise.


MaelGNM (approved):
Nice!

PR #2600, opened 2024-12-17
Use Router in discount-api

rupertbates:
## What does this change?
Use the new `Router` module in the discount-api project.


handlers/discount-api/src/index.ts
```diff
@@ -25,85 +25,80 @@ import {
...
-					statusCode: 200,
-				};
-			}
-			default:
-				return {
-					body: 'Not found',
-					statusCode: 404,
-				};
-		}
-	} catch (error) {
```
rupertbates, line 94: This error handling is done in the router now.

PR #2602, opened 2024-12-18
add sqs permission to product-switch lambda

johnduffell:
We have noticed that product switch records were not in the datalake due to them not ending up in salesforce.

The switch lambda is designed to add a record to a queue which is later written to salesforce.  When it was moved over to the product-switch-api the extra permission needed to write to that queue was missed, and for some reason, writing to a queue that you don't have access silently discards the message.

This means that we didn't get any data between 9th May and when this PR is merged.  There are small numbers of records (1 a day or so) that may be from CSR switches.

This PR adds the permission so hopefully the records will be sent from now on.


rBangay (approved):
👍

PR #2603, opened 2024-12-18
Linting Improvements

tjmw:
## What does this change?

* Lint modules and tests locally and in CI (and fix any previously uncaught linting errors)
* Linting without specifying a workspace (`pnpm lint`) streams to all workspaces
* Update the README to include section on linting.

## How to test

`pnpm lint`
`pnpm --filter <workspace> lint`

## Accessibility

### DIFF
--- a/_templates/new-lambda/api-gateway/package.ejs.t
+++ b/_templates/new-lambda/api-gateway/package.ejs.t
@@ -11,7 +11,7 @@ sh: git add handlers/<%=lambdaName%>/package.json
     "it-test": "jest --group=integration",
     "type-check": "tsc --noEmit",
     "build": "esbuild --bundle --platform=node --target=node18 --outfile=target/index.js src/index.ts",
-    "lint": "eslint src/**/*.ts",
+    "lint": "eslint src/**/*.ts test/**/*.ts",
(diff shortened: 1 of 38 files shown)


_templates/new-lambda/api-gateway/package.ejs.t
```diff
@@ -11,7 +11,7 @@ sh: git add handlers/<%=lambdaName%>/package.json
     "it-test": "jest --group=integration",
     "type-check": "tsc --noEmit",
     "build": "esbuild --bundle --platform=node --target=node18 --outfile=target/index.js src/index.ts",
-    "lint": "eslint src/**/*.ts",
+    "lint": "eslint src/**/*.ts test/**/*.ts --no-error-on-unmatched-pattern",
```
rupertbates, line 14: Is the flag so that we don't get an error if there are no tests?

rupertbates (approved):
Nice 👍

_templates/new-lambda/api-gateway/package.ejs.t
tjmw: Yeah that's right. I added it here (since we don't know if a new lambda will end up having tests) but not in existing lambdas, since we do know whether they already have tests. There were a couple of cases where I omitted the test dir because there are no tests. Maybe that's weird though, I could use this flag everywhere. Any thoughts?

PR #2606, opened 2024-12-18
Add a new endpoint `GET /benefits/<identity id>` to the user benefits API

tjmw:
## What does this change?

Add a new endpoint `/benefits/<identity id>` which is auth'd using an `x-api-key` header (enforced in API Gateway). Because the auth is performed differently, at the API Gateway, I've migrated to using the `GuApiGatewayWithLambdaByPath` with a lambda per endpoint.

This will be called from support-frontend to prevent a user who already has a the `adFree` benefit from purchasing Guardian Ad-Lite.

**Note: in the first implementation the endpoint does not know anything about internal users**

## How to test

I've deployed to CODE and tested.

**Note: this requires a dangerous deploy in Riff Raff because the old lambda is deleted.**

## Accessibility


handlers/user-benefits/test/benefitsMe.test.ts
```diff
@@ -27,34 +27,20 @@ jest.mock('@modules/product-benefits/userBenefits', () => ({
 	getUserBenefits: () => ['adFree'],
 }));
 
-describe('handler', () => {
-	it('returns a 404 for an unrecognized path or HTTP method', async () => {
```
tjmw, line 31: This diff is a bit messy but basically all that's happened here is the removal of the old 404 test, this now happens upstream.

cdk/lib/user-benefits.ts
```diff
@@ -39,38 +43,73 @@ export class UserBenefits extends GuStack {
...
 			memorySize: 1024,
 			timeout: Duration.seconds(300),
 			environment: commonEnvironmentVariables,
-			// Create an alarm
+		};
+		const userBenefitsMeLambda = new GuLambdaFunction(
+			this,
+			`user-benefits-me-lambda`,
+			{
+				description:
```
rupertbates, line 59: This description could be in the common props if it's going to be the same for both.

tjmw: Ah good spot, thanks. I think I'll keep it separate but put a bit more detail in here.

cdk/lib/user-benefits.ts
```diff
@@ -144,6 +183,6 @@ export class UserBenefits extends GuStack {
 			],
 		});
 
-		lambda.role?.attachInlinePolicy(s3InlinePolicy);
+		userBenefitsMeLambda.role?.attachInlinePolicy(s3InlinePolicy);
```
rupertbates, line 186: Do both lambdas need this policy? Or only one of them?

tjmw: Oh that's weird, I would have thought both of them. But things have been working fine without adding this for the new lambda 🤔

handlers/user-benefits/src/benefitsIdentityId.ts
```diff
@@ -0,0 +1,64 @@
...
+			statusCode: 400,
+			body: JSON.stringify({
+				message: 'Identity ID missing from request path',
+			}),
+		};
+	}
+
+	const userBenefitsResponse = await getUserBenefitsResponse(
+		stage,
+		new ProductCatalogHelper(await productCatalog.get()),
```
rupertbates, line 52: Maybe create a `Lazy<ProductCatalogHelper>` at the top level rather than a productCatalog and then creating a new helper every time?

handlers/user-benefits/src/benefitsIdentityId.ts
```diff
@@ -0,0 +1,64 @@
...
+		identityId,
+	);
+
+	return {
```
rupertbates, line 56: Would be good to extract this return object out into a function somewhere. Something like

`getNonCachedSuccessResponse(body: string)`

Or maybe just make the cache control headers a const somewhere and reuse them.

handlers/user-benefits/src/index.ts
```diff
@@ -1,82 +1,2 @@
...
-				statusCode: 400,
-			};
-		}
-		return {
-			body: 'Internal server error',
-			statusCode: 500,
-		};
-	}
-};
+export { benefitsMeHandler } from './benefitsMe';
```
rupertbates, line 1: Why do we need this file now?

cdk/lib/user-benefits.ts
rupertbates: Maybe the original one doesn't actually need it?

handlers/user-benefits/src/index.ts
tjmw: This acts as the single entry point for `esbuild` which gives access to both handlers (but means we can still keep the handlers in separate files).

cdk/lib/user-benefits.ts
tjmw: Discussed offline - we get this for free with gu cdk so I've removed.

rupertbates (approved):
Looks good 👍

PR #2607, opened 2024-12-19
Revert "fix pre-push to compare with main on first push"

shtukas:
Reverts guardian/support-service-lambdas#2605 , because it breaks pushing for me. 

```
support-service-lambdas % git push origin HEAD                 
fatal: bad revision 'origin/ph-20241219-1154-upgrade'
husky - pre-push script failed (code 128)
error: failed to push some refs to 'https://github.com/guardian/support-service-lambdas.git'
```

### DIFF
--- a/.husky/pre-push
+++ b/.husky/pre-push
@@ -1,27 +1,13 @@
+# Fetch the latest changes from the remote without merging
 git fetch origin
 
+# Get the name of the current branch
 CURRENT_BRANCH=$(git rev-parse --abbrev-ref HEAD)
 
-git diff --exit-code --quiet origin/$CURRENT_BRANCH -- cdk/
-RC=$?
-if [ $RC -eq 0 ]; then
-  # no changes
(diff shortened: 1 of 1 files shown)


MaelGNM (approved):

johnduffell:
sorry about that, it's strange, I thought I tested all four combinations carefully! Thanks for reverting it.  I also didn't realise that a nonzero exit code would kill the whole script without the -e flag.
According to git docs [hooks are just "programs"](https://git-scm.com/docs/githooks#:~:text=Hooks%20are%20programs%20you%20can%20place%20in%20a%20hooks%20directory%20to%20trigger%20actions%20at%20certain%20points%20in%20git%E2%80%99s%20execution.%20Hooks%20that%20don%E2%80%99t%20have%20the%20executable%20bit%20set%20are%20ignored.) and there's no #! line at the top of the file, so I wonder if the default shell could affect things 

@shtukas  Is your default shell set up to die on error by default?  Perhaps this script needs to be more explicit.

PR #2615, opened 2025-01-06
remove snyk workflow as we use dependabot now

johnduffell:
This is only working for the Scala, not the TS, so we get a red X.

The original plan was to fix it to work with TS, but we are using dependabot for the TS, and since recently also for Scala (There's scala stuff in here: https://github.com/guardian/support-service-lambdas/security/dependabot )

https://trello.com/c/n6Pxguan/471-add-snyk-for-typescript-in-support-service-lambdas

If we do NOT merge this, the scala will still be logged in snyk, however we will still have an annoying red X.
If we DO merge this, the scala will no longer be logged in snyk, and we will have a green tick.  So we need to make sure we use dependabot if we merge this.

Another option would be to filter this GHA down to scala only, to get the green tick while keeping the scala logged in snyk.


shtukas:
Amazing. Thanks @johnduffell ! Let's get clearance from DexEx :)

akash1810:
cc @guardian/devx-security

akash1810 (approved):
LGTM!

cc @guardian/devx-security for final(?) approval.

kelvin-chappell (approved):
👍 from @guardian/devx-security

PR #2616, opened 2025-01-08
Migrate catalog-service to alarms-handler

tomrf1:


handlers/catalog-service/cfn.yaml
```diff
@@ -194,7 +194,7 @@ Resources:
       Condition: CreateProdOnlyResources
       Properties:
         AlarmActions:
-        - !Sub arn:aws:sns:${AWS::Region}:${AWS::AccountId}:reader-revenue-dev
+          - Fn::Sub: arn:aws:sns:${AWS::Region}:${AWS::AccountId}:alarms-handler-topic-PROD
```
tomrf1, line 197: The cloudformation defines PROD and CODE resources together, which is unusual.
If it's important to know about issues with the CODE service then it should perhaps go to the PROD alarms-handler

tjmw: Yeah it's an unusual setup - a pair of lambdas in CODE (one talking to Zuora prod, one talking to Zuora code) and a pair in PROD (again, one talking to Zuora prod, one talking to Zuora code). I'm not sure of the history of this but I think given that the later two are production lambdas, we should route alarms from both to the prod alarms handler, as you've done 👍

tjmw (approved):
Thanks for picking this up!

PR #2619, opened 2025-01-09
Create discount expiry notifier state machine

david-pepper:
## What does this change?
Add a new state machine that will contain functions that orchestrate a notification to customers with a discount that is soon to expire. 

image

[Discounted Subs Notifications - Proposed Process flow](https://docs.google.com/drawings/d/1A_5kW-HSaN7JybFlpkiKbObyiPlpkckIpchMAiZJXQM/edit?usp=sharing)

### Notes
Empty lambda is present in these changes. Seems like this is the easy way to introduce a new state machine in support -service-lambdas. This lambda (getSubsWithExpiringDiscounts) will be the fleshed out in the next piece of work.


cdk/bin/cdk.ts
```diff
@@ -287,3 +288,11 @@ new UserBenefits(app, 'user-benefits-PROD', {
 	supporterProductDataTable:
 		'supporter-product-data-tables-PROD-SupporterProductDataTable',
 });
+new DiscountExpiryNotifier(app, 'discount-expiry-notifier-CODE', {
+	stack: 'membership',
```
AndreaDiotallevi, line 292: Super minor comment is that I just remembered that we'd prefer going forward to use `support` rather than `membership` as a stack name.

It would mean updating those values where they come up (cdk and riff raff) and redeploy.

This is not a "must do".

AndreaDiotallevi (approved):
Looking good - just a super minot comment on the stack name

cdk/bin/cdk.ts
david-pepper: Thanks

PR #2621, opened 2025-01-10
Prettier shouldn't care about cdk/dist

tjmw:
## What does this change?

The cdk workspace has a build command which creates compiled code in dist. On my local machine (though strangely not in CI), running:

`$ pnpm --filter cdk build`

followed by:

`$ pnpm --filter cdk check-formatting`

causes prettier to return errors relating to files in dist, which we don't want.

Also, in the package script instead of calling build (which generates dist files) just call type-check.

## Accessibility


cdk/package.json
```diff
@@ -4,13 +4,14 @@
   "private": true,
   "scripts": {
     "build": "tsc",
+    "type-check": "tsc --noEmit",
```
rupertbates, line 7: Good idea 👍

rupertbates (approved):

PR #2628, opened 2025-01-14
Type active and inactive products in the catalog distinctly

tjmw:
## What does this change?

In the product catalog a distinction is already made between active and inactive products, where active == products we're still selling and inactive == products we aren't, but which people may still have. However, this distinction isn't available at the type level.

In support-frontend, where we consume the product catalog data, we use the `ProductKey` type. Currently this includes all products and in cases where we have a `Record<ProductKey, A>` we have to provide an entry for all products, even inactive ones, which doesn't really make sense if we aren't selling them as the data won't be used. We could move to a `Partial<Record<ProductKey, A>>` but this means we have to handle undefined when looking up by a product (even if we know in practice this won't happen. I think a better approach would be to have separate types for active and inactive products. This would support `Record<ActiveProductKey, A>` meaning we'd only have to provide data about active products and not deprecated products.

So, this PR splits the single generated typeObject to two - one for active products and one for inactive. We then derived separate (and combined) types from these two objects.

This PR also adds GuardianAdLite, which is now in the type object (an active project) since it's now in the Zuora data.

## Accessibility

### DIFF
--- a/modules/product-catalog/package.json
+++ b/modules/product-catalog/package.json
@@ -2,7 +2,7 @@
   "name": "product-catalog",
   "version": "1.0.0",
   "scripts": {
-    "generateTypes": "ts-node src/generateTypeObjectCommand.ts",
+    "generateTypes": "ts-node src/generateTypeObjectCommand.ts && prettier --write src/typeObject.ts",
     "updateSnapshots": "jest -u --group=-integration",
     "test": "jest --group=-integration",
     "it-test": "jest --group=integration",
--- a/modules/product-catalog/src/generateTypeObject.ts
+++ b/modules/product-catalog/src/generateTypeObject.ts
@@ -7,6 +7,7 @@ import type {
 import { oneTimeContributionTypeObject } from '@modules/product-catalog/oneTimeContributionProduct';
 import { stripeTypeObject } from '@modules/product-catalog/stripeProducts';
 import {
+	activeProducts,
 	getProductRatePlanChargeKey,
 	getProductRatePlanKey,
 	getZuoraProductKey,
@@ -65,7 +66,7 @@ const getZuoraProduct = (productRatePlans: ZuoraProductRatePlan[]) => {
 		),
 	};
 };
-export const generateTypeObject = (catalog: ZuoraCatalog) => {
+export const generateTypeObjects = (catalog: ZuoraCatalog) => {
 	const supportedProducts = catalog.products.filter((product) =>
 		isSupportedProduct(product.name),
 	);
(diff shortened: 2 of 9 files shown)


modules/product-catalog/src/generateTypeObject.ts
```diff
@@ -79,9 +80,30 @@ export const generateTypeObject = (catalog: ZuoraCatalog) => {
 
 	const zuoraTypeObject = arrayToObject(arrayVersion);
 
+	const activeProductKeys = Object.keys(zuoraTypeObject).filter((k) =>
```
tjmw, line 83: Yes, we're iterating over `zouraTypeObject` twice here which feels slightly inefficient, but I'm not too worried since this is only run as a package script. This could be a good use case for [`Object.groupBy`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/groupBy) but that's only available in Node 21, and we're on Node 20.

modules/product-catalog/src/typeObject.ts
```diff
@@ -256,6 +241,103 @@ export const typeObject = {
...
+				DigitalPack: {},
+				Sunday: {},
+			},
+			'Saturday+': {
+				DigitalPack: {},
+				Saturday: {},
+			},
+		},
+	},
+	GuardianPatron: {
```
andrewHEguardian, line 308: Is there a problem with this abstraction for this example? Patron is definitely an active product, but not one supported by our checkout

tjmw: Yeah `GuardianPatron` is a product that falls slightly between the cracks. It is an active product (so I think it's right that it's in the active object) but it's true that we don't sell it on the support site. It feels like another abstraction (sold on support frontend) might be a bit too support site specific for the catalog, since "active" is very close to what we need. What do you reckon?

andrewHEguardian: Yep, if it reduces the exceptions we have to make in support-frontend from half a dozen to just one that's still a good improvement

modules/product-catalog/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "product-catalog",
   "version": "1.0.0",
   "scripts": {
-    "generateTypes": "ts-node src/generateTypeObjectCommand.ts",
+    "generateTypes": "ts-node src/generateTypeObjectCommand.ts && prettier --write src/typeObject.ts",
```
rupertbates, line 5: 👍

rupertbates (approved):
Looks good to me

PR #2629, opened 2025-01-14
Rename fewerSupportAsks to hideSupportMessaging

rupertbates:
## What does this change?
Dotcom uses the term 'hide support messaging' extensively to describe the benefit of not seeing banners and epics. Also the cookie is called `gu_hide_support_messaging` so I think it makes sense to use this terminology in the user benefits API.

### DIFF
--- a/modules/product-benefits/src/productBenefit.ts
+++ b/modules/product-benefits/src/productBenefit.ts
@@ -11,7 +11,7 @@ export const supporterPlusBenefits: ProductBenefit[] = [
 	'adFree',
 	'liveApp',
 	'feastApp',
-	'fewerSupportAsks',
+	'hideSupportMessaging',
 	'rejectTracking',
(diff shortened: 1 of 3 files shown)


tjmw (approved):
Looks good to me 👍

The difference in language is interesting. I think we use the "fewer support asks" type of language on the support site. I always assumed that was related to things like not being able to guarantee that a user wouldn't see support related thrashers, or perhaps Braze upsell messaging.

rupertbates:
> Looks good to me 👍
> 
> The difference in language is interesting. I think we use the "fewer support asks" type of language on the support site. I always assumed that was related to things like not being able to guarantee that a user wouldn't see support related thrashers, or perhaps Braze upsell messaging.

Yes I think fewer support asks is more accurate, but consistency probably wins out in this case.

PR #2632, opened 2025-01-14
fix pre-push to compare with main on first push (second try)

johnduffell:
following on from 
https://github.com/guardian/support-service-lambdas/pull/2605
which was reverted in
https://github.com/guardian/support-service-lambdas/pull/2607

I think this should stop git overriding the exit on error flag, and let it check the branches correctly.

### DIFF
--- a/.husky/pre-push
+++ b/.husky/pre-push
@@ -1,4 +1,6 @@
-#!/bin/sh +e
+#!/bin/sh
+
+set +e
 
 git fetch origin
 
--- a/.husky/pre-push
+++ b/.husky/pre-push
@@ -1,13 +1,29 @@
-# Fetch the latest changes from the remote without merging
+#!/bin/sh +e
+
 git fetch origin
 
-# Get the name of the current branch
 CURRENT_BRANCH=$(git rev-parse --abbrev-ref HEAD)
 
-# Check if there are changes in the 'cdk' folder between the local branch and its remote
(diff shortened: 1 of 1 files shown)


.husky/pre-push
```diff
@@ -1,13 +1,31 @@
-# Fetch the latest changes from the remote without merging
+#!/bin/sh
+
+set +e
```
johnduffell, line 3: @shtukas I think I fixed the undefined exit on error issue like this, but if you have a chance to test it locally before I merge that would be fab!

.husky/pre-push
```diff
@@ -1,13 +1,31 @@
...
+  RC=$?
+    if [ $RC -eq 0 ]; then
+      # no changes
+      echo "(main) No changes detected in the 'cdk' folder. Skipping lint and test."
+    elif [ $RC -eq 1 ]; then
+      echo "(main) Changes detected in the 'cdk' folder. Running lint and test..."
+      pnpm --filter cdk lint && pnpm --filter cdk test
+    else
+      echo "Error: could not compare with upstream"
+    fi
```
johnduffell, line 30: these changes are all the same as the previous PR

.husky/pre-push
```diff
@@ -1,13 +1,31 @@
-# Fetch the latest changes from the remote without merging
+#!/bin/sh
+
+set +e
+
```
johnduffell, line 4: the changes below here are all the same as the previus PR

shtukas (approved):

PR #2635, opened 2025-01-15
aid alarms-handler testability by making the main logic a more pure function

johnduffell:
A while ago I proposed a new feature in alarms handler to send relevant log links to chat as well.

I [made a PR](https://github.com/guardian/support-service-lambdas/pull/2487) but it got very cluttered with pre-refactoring so I will pull out individual changes to make it easier to review.

This is the first PR
Second PR is here: https://github.com/guardian/support-service-lambdas/pull/2636
Third PR is here: https://github.com/guardian/support-service-lambdas/pull/2641

Regarding this PR
1. the new tests I needed required parameterising the alarm mappings, so I pulled out the main lambda logic into a function and added that parameter
2. To make it easier to check what messages should be sent, I made the above return the message list rather than having to trap the HTTP fetch callouts.  i.e. pure-ish function, and then the effects can sit at the edge.
3. As part of that, some duplication naturally fell out between the code paths.

Tested in CODE and still sends alarms

### DIFF
--- a/handlers/alarms-handler/src/alarmMappings.ts
+++ b/handlers/alarms-handler/src/alarmMappings.ts
@@ -143,9 +143,9 @@ export type AlarmMappings = {
 	getTeamWebhookUrl: (team: Team) => string;
 };
 
-export const AlarmMappings: (
+export const buildAlarmMappings = (
 	mappings: Record<string, string[]>,
-) => AlarmMappings = (mappings: Record<string, string[]>) => {
+): AlarmMappings => {
 	const appToTeamMappings: Record<string, Team[]> =
 		buildAppToTeamMappings(mappings);
 
@@ -167,4 +167,4 @@ export const AlarmMappings: (
 	return { getTeams, getTeamWebhookUrl };
 };
 
-export const ProdAlarmMappings = AlarmMappings(teamToAppMappings);
+export const ProdAlarmMappings = buildAlarmMappings(teamToAppMappings);
--- a/handlers/alarms-handler/src/alarmMappings.ts
+++ b/handlers/alarms-handler/src/alarmMappings.ts
@@ -122,10 +122,12 @@ const teamToAppMappings: Record<Team, string[]> = {
 	],
 };
 
-const buildAppToTeamMappings = (): Record<string, Team[]> => {
+const buildAppToTeamMappings = (
+	theMappings: Record<Team, string[]>,
+): Record<string, Team[]> => {
 	const mappings: Record<string, Team[]> = {};
 
-	for (const [team, apps] of Object.entries(teamToAppMappings)) {
+	for (const [team, apps] of Object.entries(theMappings)) {
 		for (const app of apps) {
 			const teams = mappings[app] ?? [];
 			teams.push(team as Team);
@@ -136,19 +138,33 @@ const buildAppToTeamMappings = (): Record<string, Team[]> => {
 	return mappings;
 };
 
-const appToTeamMappings: Record<string, Team[]> = buildAppToTeamMappings();
-
-export const getTeams = (appName?: string): Team[] => {
-	if (appName && appToTeamMappings[appName]) {
-		return appToTeamMappings[appName];
-	}
-
-	return ['SRE'];
+export type AlarmMappings = {
+	getTeams: (appName?: string) => Team[];
+	getTeamWebhookUrl: (team: Team) => string;
 };
 
-export const getTeamWebhookUrl = (team: Team): string => {
-	return getIfDefined<string>(
-		process.env[`${team}_WEBHOOK`],
-		`${team}_WEBHOOK environment variable not set`,
-	);
+export const AlarmMappings: (
+	mappings: Record<string, string[]>,
+) => AlarmMappings = (mappings: Record<string, string[]>) => {
+	const appToTeamMappings: Record<string, Team[]> =
+		buildAppToTeamMappings(mappings);
+
+	const getTeams = (appName?: string): Team[] => {
+		if (appName && appToTeamMappings[appName]) {
+			return appToTeamMappings[appName];
+		}
+
+		return ['SRE'];
+	};
+
+	const getTeamWebhookUrl = (team: Team): string => {
+		return getIfDefined<string>(
+			process.env[`${team}_WEBHOOK`],
+			`${team}_WEBHOOK environment variable not set`,
+		);
+	};
+
+	return { getTeams, getTeamWebhookUrl };
 };
+
+export const ProdAlarmMappings = AlarmMappings(teamToAppMappings);
--- a/handlers/alarms-handler/src/alarmMappings.ts
+++ b/handlers/alarms-handler/src/alarmMappings.ts
@@ -167,4 +167,4 @@ export const buildAlarmMappings = (
 	return { getTeams, getTeamWebhookUrl };
 };
 
-export const ProdAlarmMappings = buildAlarmMappings(teamToAppMappings);
+export const prodAlarmMappings = buildAlarmMappings(teamToAppMappings);
--- a/handlers/alarms-handler/src/index.ts
+++ b/handlers/alarms-handler/src/index.ts
@@ -44,7 +44,7 @@ export const handler = async (event: SQSEvent): Promise<void> => {
 export async function getChatMessages(
 	record: SQSRecord,
 	alarmMappings: AlarmMappings,
-) {
+): Promise<{ webhookUrls: string[]; text: string } | undefined> {
 	console.log('sqsRecord', record);
 
 	const { Message, MessageAttributes } = JSON.parse(
--- a/handlers/alarms-handler/src/index.ts
+++ b/handlers/alarms-handler/src/index.ts
@@ -1,7 +1,7 @@
 import type { SNSEventRecord, SQSEvent, SQSRecord } from 'aws-lambda';
 import { z } from 'zod';
 import type { AlarmMappings } from './alarmMappings';
-import { ProdAlarmMappings } from './alarmMappings';
+import { prodAlarmMappings } from './alarmMappings';
 import { getAppNameTag } from './cloudwatch';
 
 const cloudWatchAlarmMessageSchema = z.object({
@@ -20,7 +20,7 @@ export const handler = async (event: SQSEvent): Promise<void> => {
 		for (const record of event.Records) {
 			const maybeChatMessages = await getChatMessages(
 				record,
-				ProdAlarmMappings,
+				prodAlarmMappings,
 			);
 
 			if (maybeChatMessages) {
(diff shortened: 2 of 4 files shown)


handlers/alarms-handler/src/alarmMappings.ts
```diff
@@ -136,19 +138,33 @@ const buildAppToTeamMappings = (): Record<string, Team[]> => {
 	return mappings;
 };
 
-const appToTeamMappings: Record<string, Team[]> = buildAppToTeamMappings();
+export type AlarmMappings = {
+	getTeams: (appName?: string) => Team[];
+	getTeamWebhookUrl: (team: Team) => string;
+};
```
johnduffell, line 144: in the previous PR [I did a class](https://github.com/guardian/support-service-lambdas/pull/2487/files#r1794094593) as that gives you the type for free, but I wasn't sure how to do similar as function based (assuming this is even what was meant) - perhaps there's a typeof trick I should be using rather than replicating the function signatures here?

handlers/alarms-handler/src/alarmMappings.ts
johnduffell, line 1: the changes in this file are simply to make the mappings parameterised rather than being a singleton.

handlers/alarms-handler/test/index.test.ts
```diff
@@ -62,6 +63,25 @@ describe('Handler', () => {
 		expect(getAppNameTag).toHaveBeenCalledWith('mock-arn', '111111');
 		expect(fetch).toHaveBeenCalledWith(mockEnv.SRE_WEBHOOK, expect.any(Object));
 	});
+	it('should handle captured CloudWatch alarm message', async () => {
+		(getAppNameTag as jest.Mock).mockResolvedValueOnce('mock-app');
```
johnduffell, line 67: we still have one "mocked" function but it's more minor functionality and doesn't lend itself to parameterising as easily as the alarm mappings.

handlers/alarms-handler/src/alarmMappings.ts
```diff
@@ -136,19 +138,33 @@ const buildAppToTeamMappings = (): Record<string, Team[]> => {
...
 
-export const getTeams = (appName?: string): Team[] => {
-	if (appName && appToTeamMappings[appName]) {
-		return appToTeamMappings[appName];
-	}
+export const AlarmMappings: (
```
tomrf1, line 146: I think more idiomatic would be:
`export const buildAlarmMappings = (mappings: Record<string, string[]>): AlarmMappings => {...`

handlers/alarms-handler/src/alarmMappings.ts
tomrf1: I think defining the type to be returned by the function here is good.
Technically this [could be an interface](https://blog.logrocket.com/types-vs-interfaces-typescript/) though.

Regarding "replicating the function signatures" - I think what you have here is fine. It is possible to reference the type of a field as follows (though I don't think it's necessary here):
`const getTeams: AlarmMappings['getTeams'] = (appName) => {...}`

handlers/alarms-handler/src/index.ts
```diff
@@ -42,6 +41,44 @@ export const handler = async (event: SQSEvent): Promise<void> => {
 	}
 };
 
+export async function getChatMessages(
+	record: SQSRecord,
+	alarmMappings: AlarmMappings,
+) {
```
tomrf1, line 47: return type?

johnduffell: it's `Promise<{webhookUrls: string[], text: string} | undefined>`

Last year someone in the team told me in typescript you are supposed to avoid specifying return types as it can break your code at run time.  Someone sent me some videos about it which I watched, but I forgot the exact reasoning, and can't find them any more.  Perhaps some of these? https://marcoheine.com/today-i-learned/dont-use-return-types-in-typescript

handlers/alarms-handler/src/alarmMappings.ts
johnduffell: ok thanks, done that 👍

handlers/alarms-handler/src/alarmMappings.ts
johnduffell: thanks Tom for the link and I will leave it as it is then, I agree on the referencing thing.

handlers/alarms-handler/src/index.ts
tomrf1: I've not heard that before...
In general if the return type isn't obvious then it's helpful to other humans to specify it. In this case it wasn't obvious to me

johnduffell: great in that case I've added it in!

handlers/alarms-handler/src/alarmMappings.ts
johnduffell: aha here was the original note suggesting to avoid them which I clearly took some time to digest! https://github.com/guardian/support-service-lambdas/pull/2317#discussion_r1634485906

handlers/alarms-handler/src/index.ts
johnduffell: aha here was the original note suggesting to avoid them which I clearly took some time to digest! https://github.com/guardian/support-service-lambdas/pull/2317#discussion_r1634485906

handlers/alarms-handler/src/alarmMappings.ts
```diff
@@ -136,19 +138,33 @@ const buildAppToTeamMappings = (): Record<string, Team[]> => {
...
+		return getIfDefined<string>(
+			process.env[`${team}_WEBHOOK`],
+			`${team}_WEBHOOK environment variable not set`,
+		);
+	};
+
+	return { getTeams, getTeamWebhookUrl };
 };
+
+export const ProdAlarmMappings = buildAlarmMappings(teamToAppMappings);
```
tjmw, line 170: Maybe just me, but I feel like it's _slightly_ unconventional to have this be camel case with the first letter in caps (which makes me think it's a class name or even a type). I'd probably go for `prodAlarmMappings`. Obviously a very minor thing so no big deal either way!

handlers/alarms-handler/src/index.ts
```diff
@@ -42,6 +41,44 @@ export const handler = async (event: SQSEvent): Promise<void> => {
...
+	}
+
+	if (message) {
+		const teams = alarmMappings.getTeams(message.app);
+		console.log('sending message to teams', teams);
+		const webhookUrls = teams.map(alarmMappings.getTeamWebhookUrl);
+		return { webhookUrls, text: message.text };
+	} else {
+		return undefined;
+	}
```
tjmw, line 79: Totally subjective, but we don't need this else block I don't think. With no explicit return `undefined` will be returned implicitly.

handlers/alarms-handler/src/index.ts
```diff
@@ -42,6 +41,44 @@ export const handler = async (event: SQSEvent): Promise<void> => {
 	}
 };
 
+export async function getChatMessages(
+	record: SQSRecord,
+	alarmMappings: AlarmMappings,
+): Promise<{ webhookUrls: string[]; text: string } | undefined> {
+	console.log('sqsRecord', record);
```
tjmw, line 48: Should these `console.log`s be removed before merging?

handlers/alarms-handler/src/index.ts
```diff
@@ -42,6 +41,44 @@ export const handler = async (event: SQSEvent): Promise<void> => {
...
+
+	const parsedMessage = attemptToParseMessageString({
+		messageString: Message,
+	});
+
+	console.log('parsedMessage', parsedMessage);
+
+	let message;
+	if (parsedMessage) {
+		message = await handleCloudWatchAlarmMessage({ message: parsedMessage });
```
tjmw, line 64: I wonder if `handle` is the right prefix for this method now? To me it kind of implies actually doing something with the message. What about `buildCloudwatchAlarmMessage`?

handlers/alarms-handler/src/index.ts
```diff
@@ -42,6 +41,44 @@ export const handler = async (event: SQSEvent): Promise<void> => {
...
+	} else {
+		message = handleSnsPublishMessage({
+			message: Message,
+			messageAttributes: MessageAttributes,
+		});
+	}
+
+	if (message) {
+		const teams = alarmMappings.getTeams(message.app);
+		console.log('sending message to teams', teams);
```
tjmw, line 74: If we keep this log line, what about `mapping message to teams` since we aren't actually sending it here?

```suggestion
		console.log('mapping message to teams', teams);
```

handlers/alarms-handler/src/index.ts
```diff
@@ -109,21 +135,10 @@ const handleSnsPublishMessage = async ({
...
+	console.log(`SNS publish message from ${app}`);
 
-			return fetch(webhookUrl, {
-				method: 'POST',
-				headers: { 'Content-Type': 'application/json' },
-				body: JSON.stringify({ text }),
-			});
-		}),
-	);
+	return { app, text };
```
tjmw, line 143: Totally subjective, and very minor, but rather than assigning to a new const I'd probably just map from one name to the other in the return:

```suggestion
	return { app, text: message };
```

tjmw (approved):
This looks good! I left a bunch of pretty minor comments/suggestions to take or leave.

handlers/alarms-handler/src/index.ts
johnduffell: I intentionally left them as I think this is the correct way to log things in cloudwatch in a JS lambda, if that's right?

handlers/alarms-handler/src/index.ts
johnduffell: got it, so it's like of like mapping an Option in scala

johnduffell: I tried that but it complains, not sure if I need to change something else?
image

handlers/alarms-handler/src/index.ts
johnduffell: good spot, same for `buildSnsPublishMessage`

handlers/alarms-handler/src/index.ts
johnduffell: thanks, went for
`console.log(`app ${message.app} is assigned to teams ${teams.join(', ')}`);`
 in the end

johnduffell:
> This looks good! I left a bunch of pretty minor comments/suggestions to take or leave.

Thanks this is exactly the kind of thing that's useful, I've made all your suggestions apart from the undefined one as that broke the type checker/linter.  In retrospect I think it's kind of nice to be explicit, in scala I'd use an Option but I don't think that's usual TS.

handlers/alarms-handler/src/index.ts
tjmw: Oh weird sorry about that. Maybe it's the fact that the return value is wrapped in a promise? Best to leave it as is (which I think you've done!)

tjmw: Actually I'm just wrong here - even without the promise wrapping I see the same thing in the TS playground!

Screenshot 2025-01-23 at 11 53 32

PR #2636, opened 2025-01-15
get all tags in alarm handler rather than just App

johnduffell:
This PR follows on from https://github.com/guardian/support-service-lambdas/pull/2635 and is the next part split out from the old PR https://github.com/guardian/support-service-lambdas/pull/2487

In this PR I refactor the code that retrieves the tags on the alarm.  Rather than just selecting the `App` tag, this returns an object containing all defined tags.

This means that in the next PR we can look at a new tag to build a log URL.

PR #2637, opened 2025-01-15
use alarm tags to add diagnostic urls to chat messages

johnduffell:
EDIT: superseded by https://github.com/guardian/support-service-lambdas/pull/2641 which I think is a much better solution.  However I will keep this here in case people prefer it.

--

This PR follows on from https://github.com/guardian/support-service-lambdas/pull/2636 which follows https://github.com/guardian/support-service-lambdas/pull/2635 and is the next part split out from the old PR https://github.com/guardian/support-service-lambdas/pull/2487

In this PR I add the logic to read the DiagnosticsUrls tag from the alarm, and template in the start/end times when we are building a log link.

Question: I have done it as templating a url rather than telling the lambda that it's a cloudwatch log, which is quite flexible but also you end up with very verbose tag values.  I am wondering whether it would be better to just say "it's a lambda log and this is its name", then it can calculate the url in the code?

### DIFF
--- a/handlers/alarms-handler/src/index.ts
+++ b/handlers/alarms-handler/src/index.ts
@@ -11,6 +11,13 @@ const cloudWatchAlarmMessageSchema = z.object({
 	NewStateReason: z.string(),
 	NewStateValue: z.string(),
 	AWSAccountId: z.string(),
+	StateChangeTime: z.coerce.date(),
+	Trigger: z
+		.object({
+			Period: z.number(),
+			EvaluationPeriods: z.number(),
+		})
+		.optional(),
 });
 
 type CloudWatchAlarmMessage = z.infer<typeof cloudWatchAlarmMessageSchema>;
@@ -61,7 +68,7 @@ export async function getChatMessages(
 
 	let message;
 	if (parsedMessage) {
-		message = await handleCloudWatchAlarmMessage({ message: parsedMessage });
+		message = await handleCloudWatchAlarmMessage(parsedMessage);
 	} else {
 		message = handleSnsPublishMessage({
 			message: Message,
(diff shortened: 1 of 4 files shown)


handlers/alarms-handler/src/index.ts
```diff
@@ -61,7 +68,7 @@ export async function getChatMessages(
 
 	let message;
 	if (parsedMessage) {
-		message = await handleCloudWatchAlarmMessage({ message: parsedMessage });
+		message = await handleCloudWatchAlarmMessage(parsedMessage);
```
johnduffell, line 71: not sure why the value was being wrapped in an object, unwrapping it seemed to work the same but simpler though.

handlers/alarms-handler/src/index.ts
```diff
@@ -93,34 +100,67 @@ const attemptToParseMessageString = ({
...
-	message: CloudWatchAlarmMessage;
-}) => {
-	const {
-		AlarmArn,
-		AlarmName,
-		NewStateReason,
-		NewStateValue,
-		AlarmDescription,
-		AWSAccountId,
-	} = message;
```
johnduffell, line 107: not sure why the value was being wrapped in an object, unwrapping it seemed to work the same but simpler though.

handlers/alarms-handler/src/index.ts
```diff
@@ -93,34 +100,67 @@ const attemptToParseMessageString = ({
 };
 
 const handleCloudWatchAlarmMessage = async ({
```
johnduffell, line 102: not sure why the parameter was being wrapped in an object, unwrapping it seemed to work the same but simpler though.

handlers/alarms-handler/src/index.ts
```diff
@@ -93,34 +100,67 @@ const attemptToParseMessageString = ({
...
+	const alarmCoveredTimeSeconds = Trigger
+		? Trigger.EvaluationPeriods * Trigger.Period
+		: assumedTimeForCompositeAlarms;
+	const alarmEndTimeMillis = StateChangeTime.getTime();
+	const alarmStartTimeMillis =
+		alarmEndTimeMillis - 1000 * alarmCoveredTimeSeconds;
+
+	const url = diagnosticUrlTemplate
+		.replaceAll('$startMillis', `${alarmStartTimeMillis}`)
+		.replaceAll('$endMillis', `${alarmEndTimeMillis}`);
```
johnduffell, line 159: I wasn't sure whether to go for START_MILLIS, $startMillis, %%start_millis%%, ${startMillis} or something else!  It won't be easy to change once it's set in stone though.

johnduffell:
closed in favour of https://github.com/guardian/support-service-lambdas/pull/2641

PR #2640, opened 2025-01-16
Discount expiry notifier - Add empty lambdas

david-pepper:
## What does this change?
Add empty lambdas to state machine. Logic for these lambdas will follow in subsequent PRs. Add map iterator to process each discount individually.

[Design](https://docs.google.com/drawings/d/1A_5kW-HSaN7JybFlpkiKbObyiPlpkckIpchMAiZJXQM/edit?usp=sharing)

**_State machine_**
image


graham228221 (approved):
👍 

I think as we are using the Data Lake as the source, we'll need to make another quick check in Zuora directly to make sure that the subscription is still active:
Screenshot 2025-01-20 at 09 54 29

The Data Lake does not contain real time data, it's only extracted from Zuora once a day so there's a change that a sub could be cancelled in the time between the Data Lake being updated and this step function picking it up - the chances are fairly slim, but not zero and it would be quite confusing for a customer when it happened.

PR #2641, opened 2025-01-16
use alarm tags to add log urls to chat messages

johnduffell:
This PR follows on from https://github.com/guardian/support-service-lambdas/pull/2636 which follows https://github.com/guardian/support-service-lambdas/pull/2635 and is the next part split out from the old PR https://github.com/guardian/support-service-lambdas/pull/2487

In this PR I add the logic to read the DiagnosticsLinks tag from the alarm, and expand it to build a log link (so far only cloudwatch lambda logs are supported)

This mechanism of building links reduces verbosity so the tagging PRs will be transparently easy to review, and also means if we change the url for lambda logs in future e.g. put them in kibana/bigquery, or make them go via Janus to log you in, it's a one line change rather than retagging everything.

I have also added a change to product-move-api so the alarms have log links.  See screenshot
image

I noticed that we need to update our version of CDK/gucdk to get it tagged across all our CDK based stacks, this will be a later PR.  We are on a version before this was released: https://github.com/aws/aws-cdk/issues/29759#issuecomment-2081754930


shtukas (approved):

rupertbates (approved):
👍

handlers/product-move-api/cfn.yaml
```diff
@@ -1,4 +1,3 @@
----
```
johnduffell, line 1: not sure what this was for

handlers/product-move-api/cfn.yaml
```diff
@@ -32,7 +31,8 @@ Resources:
     Properties:
       AlarmActions:
         - Fn::Sub: arn:aws:sns:${AWS::Region}:${AWS::AccountId}:alarms-handler-topic-${Stage}
-      AlarmName: An error in the Product Move lambda. Please check the logs to diagnose
+      AlarmName:
+        Fn::Sub: ${Stage} An error in the Product Move lambda. Please check the logs to diagnose
```
johnduffell, line 35: I needed to change this so I could create the alarm (temporarily) in CODE.  We should probably have the stage on all alarms so I left it in.

handlers/product-move-api/cfn.yaml
```diff
@@ -371,8 +378,8 @@ Resources:
       AlarmActions:
         - Fn::Sub: arn:aws:sns:${AWS::Region}:${AWS::AccountId}:alarms-handler-topic-${Stage}
       AlarmName:
-        Fn::Sub: The product-move-api returned a 500 response
-      AlarmDescription: Check the logs for details - https://eu-west-1.console.aws.amazon.com/cloudwatch/home?region=eu-west-1#logsV2:log-groups/log-group/$252Faws$252Flambda$252Fmove-product-PROD
```
johnduffell, line 375: don't need the generic hard coded PROD link any more

johnduffell:
I think we need the time periods to be a bit more fuzzy as it doesn't always get the period when the actual log lines were written.
e.g.
`2025-01-29T16:13:45.169` the lamba finished 
`2025-01-29 16:14:00` the metric was positive
`2025-01-29T16:15:29.420` the alarm state changed (on a 1 minute period)

I think maybe the API gateway reports its metrics once a minute, the metric covers a 1 minute period, and the alarm state is checked once a minute, but none of them are guaranteed to line up

PR #2644, opened 2025-01-21
Discount expiry notifier - Add active sub check logic

david-pepper:
## What does this change?
Add logic to get sub from Zuora and return the status 

Screenshot 2025-01-21 at 12 21 13

[State Machine Design](https://docs.google.com/drawings/d/1A_5kW-HSaN7JybFlpkiKbObyiPlpkckIpchMAiZJXQM/edit?usp=sharing)

## How to test
- get a cancelled sub and an active sub from zuora
- add each subname to the lambdas environment variable
- execute the lambda verify that the status returned by the lambda is correct
image

### Notes
- For now lambda is sub name via environment variable, but will be piped in as a parameter as development progresses


cdk/lib/discount-expiry-notifier.ts
```diff
@@ -56,6 +56,15 @@ export class DiscountExpiryNotifier extends GuStack {
 				handler: 'subIsActive.handler',
 				fileName: `${appName}.zip`,
 				architecture: Architecture.ARM_64,
+				initialPolicy: [
+					new PolicyStatement({
+						actions: ['secretsmanager:GetSecretValue'],
+						resources: [
+							`arn:aws:secretsmanager:${this.region}:${this.account}:secret:CODE/Zuora-OAuth/SupportServiceLambdas-S8QM4l`,
```
AndreaDiotallevi, line 63: This works but it will stop working if that secret gets deleted and recreated, since the id will change.

A safer approach would be this:
`arn:aws:secretsmanager:${this.region}:${this.account}:secret:${this.stage}/Zuora-OAuth/SupportServiceLambdas-*`

handlers/discount-expiry-notifier/src/secrets.ts
```diff
@@ -0,0 +1,14 @@
+import type { Stage } from '@modules/stage';
+
+export function getZuoraSecretName(stage: Stage): string {
```
AndreaDiotallevi, line 3: You can simplify this by:
`${stage}/Zuora-OAuth/SupportServiceLambdas`

handlers/discount-expiry-notifier/src/zuoraHttp.ts
```diff
@@ -0,0 +1,18 @@
+import { stageFromEnvironment } from '@modules/stage';
+import { ZuoraClient } from '@modules/zuora/zuoraClient';
+import { zuoraSubscriptionSchema } from '@modules/zuora/zuoraSchemas';
+import type { ZuoraSubscription } from '@modules/zuora/zuoraSchemas';
+
+export async function getSub(subName: string): Promise<ZuoraSubscription> {
```
AndreaDiotallevi, line 6: I think there is already a `getSubscription` function inside the Zuora modules.
Search for `getSubscription.ts`

cdk/lib/discount-expiry-notifier.ts
david-pepper: Thanks, done

handlers/discount-expiry-notifier/src/secrets.ts
david-pepper: yep agreed, done

handlers/discount-expiry-notifier/src/zuoraHttp.ts
david-pepper: Good shout, done

AndreaDiotallevi (approved):
Thanks for looking at the comments!

PR #2651, opened 2025-01-27
Discount expiry notifier - save operations status to s3

david-pepper:
## What does this change?
- add logic to write the state machine JSON object to an s3 bucket. This object will contain the inputs and outputs from each lambda, which will contain the successes and failures of attempts to send details to membership workflow to initiate email send. We can refine what gets saved to S3 when each lambda is properly integrated to the execution of the state machine and we have a better idea of what the (input/output) object definitions will look like for each lambda.

Screenshot 2025-01-27 at 16 23 18

## How to test
- execute the state machine and verify that the input to the saveResults lambda gets saved to the s3 bucket with file name runtime date:

_file in s3 bucket_
Screenshot 2025-01-27 at 16 21 35

_file contents_
image

 
### To do in subsequent PRs
- handle reruns of the state machine in case of error. The file in the s3 bucket should not be overwritten, but instead it should either be appended to, or a separate file should be created for subsequent executions.
- the contents of the file will be refined to just the information we are interested in: query results, successes and failures. At the moment we are just dumping the entirety of the JSON object from the output of the state machine.
- I've noticed that the query is taking a long time to run when the resulted are limited with LIMIT 2 - over 30 seconds which results in a timeout error. Specifying two subs in the where clause allows it to complete quicker and without error. We will likely need to look to improve the performance of the query for production.


handlers/discount-expiry-notifier/src/bigquery.ts
```diff
@@ -72,7 +72,7 @@ export const runQuery = async (
 				AND sub.is_latest_version = TRUE 
 				AND sub.status = 'Active' 
 				AND DATE_ADD(charge.effective_start_date, INTERVAL charge.up_to_periods MONTH) = DATE_ADD(CURRENT_DATE(), INTERVAL 32 DAY) 
-			LIMIT 2
+				AND sub.name IN ('A-S02282302', 'A-S02282308')
```
david-pepper, line 75: when query is not limit to 2 sub names for testing, it is taking a long time to run - more than 30 seconds, and there is a time out error occurring. Restricting query to these two subs, but will need to find a way to optimise the query for production.

handlers/discount-expiry-notifier/src/handlers/saveResults.ts
```diff
@@ -1 +1,42 @@
...
+};
+
+type ExpiringDiscountProcessingAttempt = {
+	status: string;
+};
+
+type LambdaInput = {
+	discountsToProcess: ExpiringDiscountToProcess[];
+	discountProcessingAttempts: ExpiringDiscountProcessingAttempt[];
+};
```
david-pepper: Will refine these types in a subsequent PR when the lambdas are running together in the state machine

handlers/discount-expiry-notifier/src/handlers/saveResults.ts
```diff
@@ -1 +1,44 @@
...
+	status: string;
+};
+
+type LambdaInput = {
+	expiringDiscountsToProcess: ExpiringDiscountToProcess[];
+	expiringDiscountProcessingAttempts: ExpiringDiscountProcessingAttempt[];
+};
+
+export const handler = async (event: LambdaInput) => {
+	const bucketName = `discount-expiry-notifier-${stageFromEnvironment()}`;
```
AndreaDiotallevi, line 22: An idea - since you create the bucket in CDK, you could pass down an environment variable BUCKET_NAME, so you don't need to couple the string logic here.

handlers/discount-expiry-notifier/src/handlers/saveResults.ts
```diff
@@ -1 +1,44 @@
...
+
+	const getCurrentDateString = (): string => {
+		const now = new Date();
+		const year = now.getFullYear();
+		const month = String(now.getMonth() + 1).padStart(2, '0'); // Months are 0-based
+		const day = String(now.getDate()).padStart(2, '0');
+		return `${year}-${month}-${day}`;
```
AndreaDiotallevi, line 29: I am not sure about the scale and how many files we are processing, but it could be useful to use folders for year, month and days. e.g. year/month/day.

handlers/discount-expiry-notifier/src/s3.ts
```diff
@@ -0,0 +1,61 @@
+import type {
```
AndreaDiotallevi, line 1: I see this is a copy of "handlers/salesforce-disaster-recovery/src/services/s3.ts", so we might want to create a module?

AndreaDiotallevi (approved):

handlers/discount-expiry-notifier/src/handlers/saveResults.ts
david-pepper: at the moment it'll be one file per execution, so I think it'll be fine as is. If a second run is required because of an error, we can create another file with an appropriate suffix i.e 2024-01-28 (1)

handlers/discount-expiry-notifier/src/s3.ts
david-pepper: Yes for sure, I'll do that post-release

PR #2654, opened 2025-01-28
Delete revenue recogniser job 

rupertbates:
## What does this change?
The revenue recogniser job lambda was only needed for Digital Subscription gifts which we have not sold since 11/01/2023 and which we have removed from the codebase here: https://github.com/guardian/support-frontend/pull/6736

I will manually delete the AWS infrastructure once this PR is merged.

## Description of what it did, copied from README:

The revenue recognition job is needed because some subscriptions (Digital Subscription gifts)
have a manual revenue recognition rule set up.  This is so that revenue can be recognised
on redemption rather than on purchase.

Unfortunately this means that if a subscription is never redeemed, the revenue is never recognised.
Also, if a refund is made, the refund is not distributed.

The job will search for all subscriptions that need to be recognised, and distribute the schedules
as appropriate.


johnduffell (approved):
thanks for doing this 👍 
I added a couple more commits because it was still building an artifact that only contained the dependencies, and was pushing it to riffraff https://github.com/guardian/support-service-lambdas/actions/runs/13009635579/job/36284225194#step:6:563

off the top of my head as you mention we would have to
- delete the CD from riffraff
- (no scheduled deploy to delete - this has bitten us before!!)
- delete the CODE and PROD stack from AWS
- (no config or zuora users to delete, I think it just uses the standard ZuoraRest.json)

rupertbates:
CODE and PROD stacks are deleted now

johnduffell:
thanks I have deleted the CD for PROD from riffraff (almost exactly 4 years since I added it)

PR #2656, opened 2025-01-28
Discount expiry notifier - send subscriber details to membership workflow

david-pepper:


graham228221 (approved):
Looks good to me, and I can see the requests reaching membership-workflow as expected:

Screenshot 2025-01-30 at 10 23 16

PR #2658, opened 2025-01-28
Add benefits/list endpoint to user-benefits API

rupertbates:
## What does this change?
This PR adds a new endpoint to the user-benefits api to return a list of all benefits by product.

The response of the endpoint can be in HTML or JSON depending on the accept header of the request, if the accept header includes 'text/html' the API will return HTML, otherwise it will return JSON.

### HTML response
Screenshot 2025-01-29 at 10 52 53

### Partial JSON response
Screenshot 2025-01-29 at 10 53 51

### DIFF
--- a/cdk/lib/user-benefits.ts
+++ b/cdk/lib/user-benefits.ts
@@ -102,7 +102,7 @@ export class UserBenefits extends GuStack {
 					apiKeyRequired: true,
 				},
 				{
-					path: '/benefits/list/{format+}',
+					path: '/benefits/list',
 					httpMethod: 'GET',
 					lambda: userBenefitsListLambda,
 				},
--- a/cdk/lib/user-benefits.ts
+++ b/cdk/lib/user-benefits.ts
@@ -102,7 +102,7 @@ export class UserBenefits extends GuStack {
 					apiKeyRequired: true,
 				},
 				{
-					path: '/benefits/list',
+					path: '/benefits/list/{format+}',
 					httpMethod: 'GET',
 					lambda: userBenefitsListLambda,
 				},
--- a/cdk/lib/user-benefits.ts
+++ b/cdk/lib/user-benefits.ts
@@ -80,7 +80,7 @@ export class UserBenefits extends GuStack {
 			`user-benefits-list-lambda`,
 			{
 				description:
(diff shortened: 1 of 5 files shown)


cdk/lib/user-benefits.ts
```diff
@@ -75,6 +75,17 @@ export class UserBenefits extends GuStack {
 				...commonLambdaProps,
 			},
 		);
+		const userBenefitsListLambda = new GuLambdaFunction(
+			this,
+			`user-benefits-list-lambda`,
+			{
+				description:
+					'An API Gateway triggered lambda to get the benefits of the user identified in the request path',
```
andrewHEguardian, line 83: Re-used description?

andrewHEguardian (approved):
Looks good 👍 What is the use case and how will you share it?

handlers/user-benefits/src/benefitsList.ts
```diff
@@ -0,0 +1,71 @@
...
+		return getHttpResponse(getHtmlBody(), 'text/html');
+	}
+	return getHttpResponse(JSON.stringify(productBenefitMapping));
+};
+
+const getHttpResponse = (
+	body: string,
+	contentType?: string,
+): Promise<APIGatewayProxyResult> => {
+	return Promise.resolve({
```
tjmw, line 20: Does this need to be a promise? Could the return value just be `APIGatewayProxyResult`?

tjmw (approved):
👍

rupertbates:
> Looks good 👍 What is the use case and how will you share it?

The use case is that currently various people have spreadsheets which they maintain to work out which benefits are associated with which products, and sometimes these are wrong or out of date. 

Providing a url that anyone can go to in a browser to see the real mapping which is used by all our systems, should fix that.

handlers/user-benefits/src/benefitsList.ts
rupertbates: That doesn't seem to work, I get a 500 response. I think it needs to be a promise.

tjmw: Oh weird, I would have thought it would be enough to make sure the handler itself returned a promise (which it will because it's marked as async) but that it wouldn't be necessary at this level. Sorry for the diversion!

PR #2659, opened 2025-01-29
Discount expiry notifier - enable optional input for execution date

david-pepper:
## What does this change?
Enable the state machine to take in an input date parameter so that we can rerun it for a given date if an error occurs.

## How to test
- run the state machine with an input parameter and verify this is the date used in the query
image

- run the state machine without an input parameter and verify today's date is used in the query


AndreaDiotallevi (approved):
Good idea 👍

PR #2662, opened 2025-01-30
supporter plus update amount fix

rBangay:
## What does this change?

Subtract actual base price rather than catalog base price when updating supporter plus amount.

This solves an issue where customers that haven't had had a price rise applied yet try to update their supporter plus amount which appears to work but is actually less than the amount they chose.


johnduffell (approved):
nice one, good pairing this morning 👍

PR #2666, opened 2025-02-03
Use latest corepack

tjmw:
## What does this change?

Uses a more recent version of corepack in CI.

We started seeing issues with pnpm signatures, which is apparently fixed with a later corepack release.

See pnpm/pnpm#9014

## How to test

Ensure the build passes.

### DIFF
--- a/.github/workflows/ci-scala.yml
+++ b/.github/workflows/ci-scala.yml
@@ -50,7 +50,7 @@ jobs:
 
       - name: Run tests and build subproject JAR
         run: sbt "project ${{ matrix.subproject }}" test assembly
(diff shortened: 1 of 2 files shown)


andrewHEguardian (approved):

rupertbates (approved):

johnduffell (approved):
looks good, the fact that we do '--global' does that mean it will store it in the GHA cache for next time to improve speed?

tjmw:
> looks good, the fact that we do '--global' does that mean it will store it in the GHA cache for next time to improve speed?

I'm actually not sure about this. I think I'll merge to unblock things and investigate as a possible follow up optimisation.

PR #2670, opened 2025-02-03
Discount expiry notifier - handle multiple executions for given day

david-pepper:
## Description
- update the filepath (key) so that multiple executions result in one file per execution (and do not overwrite previous files)

Screenshot 2025-02-03 at 17 31 18

### Changes
- add logic to save file in folder named as discountExpiryDate in format YYYY-MM-DD, and as file named as executionTime in format YYYY-MM-DDTHH:mm:ss.SSSZ e.g. 2024-02-03/2025-02-03T16:55:39.457Z. This ensure unique filenames regardless of how many executions are run for the same discountExpiryDate.
- some tidyup on variable names and addition of helper methods in getSubsWithExpiringDiscounts.ts
- shifting of discountExpiryDate determination from query to ts logic
- adding '32' as an environment variable to represent time window between execution time and discount expiry date

## Testing
Execute the state machine and verify the object in the jsonpath get saved at filepath following the convention outlined above. Execute again and ensure the second file does not overwrite the first.

Screenshot 2025-02-03 at 17 41 34

### DIFF
--- a/cdk/lib/__snapshots__/discount-expiry-notifier.test.ts.snap
+++ b/cdk/lib/__snapshots__/discount-expiry-notifier.test.ts.snap
@@ -424,6 +424,7 @@ exports[`The discount-expiry-notifier stack matches the snapshot 1`] = `
         "Environment": {
           "Variables": {
             "APP": "discount-expiry-notifier",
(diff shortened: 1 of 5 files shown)


AndreaDiotallevi (approved):
Nice

PR #2679, opened 2025-02-06
Move corepack enable to after Node setup

tjmw:
It feels a bit weird that we were running corepack before the Node setup. Try moving it after (and see if that removes the need to install a specific later version, after all Node ships with corepack by default).

## Accessibility


tjmw:
Reading actions/setup-node#480 and some of the examples in there, it looks like it's expected that `corepack enable` would be run ahead of `setup-node`.

PR #2680, opened 2025-02-06
Discount expiry notifier - various improvements and fixes

david-pepper:
## What does this change?

Some low-level fixes/improvements:

- Fix campaign id for CODE
- add some record counts to the object at the JSON path to help with logging (total subs count, filtered subs count)
- remove the product filtering as per updated requirement to only filter by region (anod not by product)
- update a variable name to match the query response schema (subName -> zuoraSubName)
- fix braze merge fields for email campaign
- update logic to handle lowercase country names
- handle error on callout to zuora to get sub status
- throw error on error when grabbing records from bigquery

### Notes
- we are using an older version of the email campaign at the moment for testing (the GW-specific version). The most up to date, product-agnostic, version is currently having its copy reviewed.


AndreaDiotallevi (approved):
👍

PR #2681, opened 2025-02-07
Fix Braces Vulnerabilities 

ecguardian:
## What does this change?

This PR increased the version of Braces to 3.0.3. It is indirectly being used by Jest, which can not be upgraded to a higher version of Braces. So have implement an override on the micromatch, to make sure that all instances of braces used are 3.0.3.

## How to test

This has been tested by running pnpm install and looking to see if the correct version of Braces is being used using "npm ls braces"

## Accessibility

-   [ ] [Tested with screen reader](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#screen-reader)
-   [ ] [Navigable with keyboard](https://github.com/guardian/accessibility/blob/main/people-and-technology/02-physical.md#keyboard)
-   [ ] [Colour contrast passed](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#contrast)
-   [ ] [The change doesn't use only colour to convey meaning](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#use-of-colour)

PR #2682, opened 2025-02-07
Fix cross spawn vulnerability

ecguardian:
## What does this change?

This PR increased the version of cross-spawn to 7.0.6. It is indirectly being used by other packages, which can not be upgraded to a higher version. So have implement an override, to make sure that all instances of braces used 7.0.6.

## Accessibility

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,5 +31,10 @@
 		"ts-jest": "^29.2.5",
 		"typescript": "^5.6.3"
 	},
+	"pnpm": {
+		"overrides": {
+			"micromatch>braces": "^3.0.3"
+		}
+	},
 	"packageManager": "pnpm@9.1.1"
 }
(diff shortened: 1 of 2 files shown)


johnduffell (approved):
I like the idea, although I would defer to the tests - if they pass, then it must be ok.

package.json
```diff
@@ -31,5 +31,11 @@
 		"ts-jest": "^29.2.5",
 		"typescript": "^5.6.3"
 	},
+	"pnpm": {
+		"overrides": {
+			"micromatch>braces": "^3.0.3",
```
johnduffell, line 36: this line is also in https://github.com/guardian/support-service-lambdas/pull/2681 as I guess you based this branch off that one?
might be worth editing the target of this PR to that branch, rather than `main`, so that the changes are clearer in each PR (that's what I usually do).  It also avoids me accidentally merging the "second" pr first when there are outstanding changes on the "first" one, github doesn't prevent it.

PR #2683, opened 2025-02-07
Metric Push API Lambda

tjmw:
## What does this change?

This PR introduces a lambda to handle metric-push-api requests. We now check the referer in the lambda and if it matches a pre-defined list (the support site) then emit a custom metric. Alarm on this metric. This gives us much more control over filtering requests which we think are legit and actually want to emit the metric for.

## How to test

Deploy to CODE. Requests to the API with a valid referer should emit the metric, which we should be able to see on the alarm graph. Requests with an invalid referer should not emit the metric.

More meaningful client side error alarms.

## Accessibility


rupertbates (approved):
Great 👍

PR #2688, opened 2025-02-10
Discount expiry notifier - add alarms

david-pepper:
## What does this change?
For the following lambdas, if an error is encountered, throw an fatal error and send an alert:

- getSubsWithExpiringDiscounts,
- filterSubs,
- alarmsOnFailures,

Screenshot 2025-02-10 at 17 29 38

For the following lambdas, if an error is encountered, catch it and add its details to the JSON path object. Then check JSONpath for errors in new alarmsOnFailures lambda, and if found, throw an error and send an alert:

- subIsActive
- initiateEmailSend

Screenshot 2025-02-10 at 17 30 03

## How to test
- simulate errors in each of the lambdas and verify that an alert is sent to the platform chat channel:
Screenshot 2025-02-10 at 17 31 00

### Notes
Input types for lambdas are getting a bit unwieldy. Will tidy these up by adding named types in the next PR


handlers/discount-expiry-notifier/package.json
```diff
@@ -9,11 +9,14 @@
...
-		"fix-formatting": "prettier --write **.ts"
+		"fix-formatting": "prettier --write **.ts",
+    	"watch": "chokidar 'src/**/*.ts' -c 'pnpm fix-formatting'"
+
 	},
 	"devDependencies": {
 		"@types/aws-lambda": "^8.10.147",
-		"@types/aws-sdk": "^2.7.4"
+		"@types/aws-sdk": "^2.7.4",
+		"chokidar-cli": "^3.0.0"
```
AndreaDiotallevi, line 19: What is chokidar cli?

AndreaDiotallevi (approved):

handlers/discount-expiry-notifier/package.json
david-pepper: Using it to format my code on save

AndreaDiotallevi: You can use VSCode and specify prettier as the formatter for TypeScript so that it uses the prettier config present in the repo. Usually you don't want to add something you use locally for everybody to install.

david-pepper: Great, makes sense. I've removed it

PR #2690, opened 2025-02-11
import cas common into digi sub expiry lambda

johnduffell:
We have a [dependabot vulnerability](https://github.com/advisories/GHSA-78wr-2p64-hpwj) in commons.io via digital-subscription-expiry lambda which needed fixing.

It was coming via https://github.com/guardian/content-authorisation-common

That project was archived, and only used from the above lambda.

This PR copies in the code from the above repo and updates the dependency versions appropriately.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -400,7 +400,7 @@ lazy val `identity-backfill` = lambdaProject(
 lazy val `digital-subscription-expiry` = lambdaProject(
   "digital-subscription-expiry",
   "check digital subscription expiration for authorisation purposes",
-  Seq(contentAuthCommon),
+  Seq(jodaTime, commonsIO, typesafeConfig),
   Seq(zuora, handler, effectsDepIncludingTestFolder, testDep),
 )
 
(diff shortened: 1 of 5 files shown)


build.sbt
```diff
@@ -400,7 +400,7 @@ lazy val `identity-backfill` = lambdaProject(
 lazy val `digital-subscription-expiry` = lambdaProject(
   "digital-subscription-expiry",
   "check digital subscription expiration for authorisation purposes",
-  Seq(contentAuthCommon),
+  Seq(jodaTime, commonsIO, typesafeConfig),
```
johnduffell, line 403: needed by cas common

https://github.com/guardian/content-authorisation-common/blob/07196ed1e36d319bb3003c26f9e3fad9376ed51f/build.sbt#L14-L16

handlers/digital-subscription-expiry/src/main/scala/com/gu/cas/Config.scala
```diff
@@ -0,0 +1,5 @@
+package com.gu.cas
```
johnduffell, line 1: these files are from https://github.com/guardian/content-authorisation-common/tree/master/src/main/scala/com/gu/cas

tomrf1 (approved):

johnduffell:
I don't think emergency tokens are a thing any more, since subs frontend was decomissioned.  We can add logging to see if they are actually used.  The lambda itself is still getting a good number of requests per hour, but maybe not for emergency codes.

PR #2692, opened 2025-02-11
Discount expiry notifier add types

david-pepper:
## What does this change?
- add named types for lambda inputs

## Testing
execute state machine and verify it executes without error

## Notes
Still to do to enhance type checking is to add zod type definitions. I will do this in a PR soon but want to focus on handling null workemail in query results first


AndreaDiotallevi (approved):
Nice

PR #2699, opened 2025-02-11
Add soft opt-in mapping for Ad-Lite

MaelGNM:
## What does this change?

This adds mapping to the soft opt-in consent setter for `Guardian Ad-Lite` product


rBangay (approved):
👍

johnduffell (approved):

PR #2700, opened 2025-02-12
Discount expiry notifier - rename some lambdas

david-pepper:
## What does this change?
Rename some lambdas for clarity/simplicity


graham228221 (approved):
all seems sensible!

PR #2703, opened 2025-02-12
Don't run cdk package for every TypeScript lambda

tjmw:
## What does this change?

I noticed that the cdk package step is by far the slowest part of each TypeScript lambda build. The package script does type checking, formatting checking, linting, runs the tests and synths the Cloudformation. We don't need to do all of this for every lambda, so instead run all the checks in the common job and only synth the Cloudformation in the invidual lambda jobs (which we need to upload to Riff Raff).

This doesn't really improve the overall speed for a TypeScript build to complete as it just moves the cdk checks to the `common` job and the main TS builds still have to wait for `common` to complete. It does mean the combined overall time for runners is lower and provides slightly faster feedback in the case of a cdk issue (tests failing for example). It also means in the case of a cdk issue only one job (`common`) will fail, instead of repeating the same failure in the individual TS lambda jobs.

**Update 14/2:** I've now split the cdk checks and module checks (previously running sequentially in common) to separate jobs which run in parallel (there are no dependencies between them). Then when they're both successfully completed the individual lambda builds run. This _does_ have an overall speedup effect, and maintains the benefit of a cdk failure happening in only one place.

## How to test

Picking on the ticket-tailor-webhook build as an example.

Before:
`CI-Typescript / common`: 1m23s
`CI-Typescript / gu-cdk-build (ticket-tailor-webhook)`: 1m27s

After:
`CI-Typescript / common`: 2m10s
`CI-Typescript / gu-cdk-build (ticket-tailor-webhook)`: 40s

Latest:
In parallel:
`CI-Typescript / modules`: 1m
`CI-Typescript / cdk-checks`: 1m
Then:
`CI-Typescript / gu-cdk-build (ticket-tailor-webhook)`: 37s

(There's some variability in build run times, so take these with a grain of salt. But there does seem to be an overall improvement).

## Accessibility

### DIFF
--- a/.github/workflows/ci-typescript.yml
+++ b/.github/workflows/ci-typescript.yml
@@ -33,6 +33,7 @@ jobs:
       - run: pnpm --filter "./modules/**" check-formatting
       - run: pnpm --filter "./modules/**" lint
       - run: pnpm --filter "./modules/**" test
+      - run: pnpm --filter cdk verify
 
   gu-cdk-build:
     needs: common
@@ -67,7 +68,7 @@ jobs:
           node-version-file: '.nvmrc'
           cache: 'pnpm'
       - run: pnpm install
-      - run: pnpm --filter cdk package
+      - run: pnpm --filter cdk synth
       - run: pnpm --filter ${{ matrix.subproject }} package
 
       - name: Configure AWS credentials
--- a/.github/workflows/ci-typescript.yml
+++ b/.github/workflows/ci-typescript.yml
@@ -14,7 +14,7 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  common:
+  modules:
     runs-on: ubuntu-latest
     permissions:
       id-token: write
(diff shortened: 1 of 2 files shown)


johnduffell:
@tjmw  oh interesting, I noticed this issue last week and was working on it on my 10% time.  As this repo scales up, things seem to be creaking a bit.

My solution was slightly different, I was making a build definition generator (a bit like SBT does out of the box) to spit out a bunch of package.json etc on every git pull.  Then we could move the CDK stuff back into the individual handler directories.

It didn't get as far as a PR but I will try to pick it up again next time on 10% time.

I don't think pnpm is very well suited to mono repos without some kind of build generator, as we end up having to duplicate a lot of things otherwise.  Let me know if you want to pair next week or some other time?

johnduffell:
@tjmw I guess this doesn't make so much sense now since we have gone the opposite way (moving things out of common) in https://github.com/guardian/support-service-lambdas/pull/3552
It is still the case that the CDK part of the build is disproportionately slow, so I'm sure there's still something to be done to improve things.

tjmw:
Closing this as it's quite stale at this point.

Issue #2456, opened 2024-09-04
Breaking terms of the Apache license?


pjfanning:
The Apache License requires that the LICENSE and NOTICE are preserved when source code or compiled code are copied from another source with an Apache License.

Please see the Redistribution section of:
https://www.apache.org/licenses/LICENSE-2.0

This is an example that looks non-compliant. There may be more instances.

https://github.com/guardian/support-service-lambdas/blob/a0cc7445aec072d86727f8e90edb563e383a719e/project/Dependencies.scala#L130-L132

tjmw:
Hi @pjfanning thanks for reporting this. I've added a card for the team who own this project to investigate.

johnduffell:
Hi @pjfanning thanks for your interest in our code, and for reporting this, and sorry it's taken us so long to get around to it.

I have just merged a PR to fix it, after testing it locally, and all seems to be in order.

Feel free to raise any other issues you find!

https://github.com/guardian/support-service-lambdas/pull/2704

PR #2704, opened 2025-02-13
keep all copies of licences from jackson-core

johnduffell:
We are using jackson-core both directly, and via the shaded copy in AWS SDK for Java v2.
This means that we get multiple copies of the licence files with the same name (and similar content)

During the `assembly` task to produce a Jar file for lambda, we need to explicitly resolve any file conflicts.
```
[error] Deduplicate found different file contents in the following:
[error]   Jar name = jackson-core-2.17.2.jar, jar org = com.fasterxml.jackson.core, entry target = META-INF/FastDoubleParser-NOTICE
[error]   Jar name = third-party-jackson-core-2.27.24.jar, jar org = software.amazon.awssdk, entry target = META-INF/FastDoubleParser-NOTICE
[error] Total time: 2 s, completed 13 Feb 2025, 14:30:09
```

Previously we handled this by discarding the duplicates as it didn't cause any functionality issues, and the jar is only for internal use, however it has been pointed out in https://github.com/guardian/support-service-lambdas/issues/2456 that this is a breach of the licence agreement.

This PR updates the merge process so it will keep both by concatenation.  This will slightly push up the jar file size, but it shouldn't cause much slowdown in the grand scheme of things.

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -130,8 +130,8 @@ object Dependencies {
      * So it's redundant in a binary artefact.
      */
     case PathList("codegen-resources", _*) => MergeStrategy.discard
-    case PathList("META-INF", "FastDoubleParser-LICENSE") => MergeStrategy.discard
-    case PathList("META-INF", "FastDoubleParser-NOTICE") => MergeStrategy.discard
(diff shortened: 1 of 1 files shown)


rBangay (approved):
👍

MaelGNM (approved):

PR #2719, opened 2025-02-26
Add ssm to shared modules

david-pepper:
## What does this change?
Extract ssm from discount-expiry-notifier solution and add to common typescript modules

## How to test
- automated tests have been added to validate behaviour of getSSMParam() function
- execute discount-expiry-notifier state machine and verify parameter is returned from in getExpiringDiscounts lambda

### Notes
- also removed file secrets.ts as this is not being used.


AndreaDiotallevi (approved):
Nice to see this!

PR #2720, opened 2025-02-26
Add s3 to shared modules

david-pepper:
## What does this change?
Extract s3 from discount-expiry-notifier solution and add to common typescript modules

## How to test
- automated tests have been added to validate behaviour of uploadFileToS3() and getFileFromS3() functions
- execute discount-expiry-notifier state machine and verify results are saved to s3 bucket in saveResults lambda

### Notes
- took s3.test.ts directly from salesforce-disaster-recovery solution. Will update references from that solution in a subsequent PR.


AndreaDiotallevi (approved):
Thank you!

PR #2723, opened 2025-02-27
let people cancel supporter plus even if they switched earlier in the day

johnduffell:
We had some alarms in produce-move-api where a customer had subscribed to recurring contribution, then switched to supporter plus, and then tried to cancel, all in the same day.

The cancellation was throwing an error saying there were two rateplans on the subscription, and refusing to cancel.

After investigation, we discovered that zuora returns charges that ended the same day, even though the docs say it shouldn't [1]

The call was added in this PR [2] in order to make sure we use the currently active charges rather than future active charges

This PR filters out the erroneous charges in the lambda code.

We discussed whether to add a test and decided not to in the end, as it's such an edge case that the cost of the test would be more than the chance of regression.  
Testing in CODE worked fine by following the steps the customer did.

[1] https://developer.zuora.com/v1-api-reference/api/operation/GET_SubscriptionsByKey/#:~:text=current%2Dsegment%3A%20The%20segmented%20charge%20that%20is%20active%20on%20today%E2%80%99s%20date%20(effectiveStartDate%20%3C%3D%20today%E2%80%99s%20date%20%3C%20effectiveEndDate)

[2] https://github.com/guardian/support-service-lambdas/pull/2127

### DIFF
--- a/handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/cancel/SubscriptionCancelEndpointSteps.scala
+++ b/handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/cancel/SubscriptionCancelEndpointSteps.scala
@@ -33,6 +33,7 @@ class SubscriptionCancelEndpointSteps(
     sqs: SQS,
     stage: Stage,
     zuoraSetCancellationReason: ZuoraSetCancellationReason,
+    today: LocalDate,
 ) {
   private[productmove] def subscriptionCancel(
       subscriptionName: SubscriptionName,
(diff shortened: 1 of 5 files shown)


handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/cancel/SubscriptionCancelEndpointSteps.scala
```diff
@@ -63,9 +64,13 @@ class SubscriptionCancelEndpointSteps(
...
       // lastChangeType of 'Remove')
       ratePlan <- asSingle(
-        subscription.ratePlans.filter(ratePlan =>
-          ratePlan.productName != "Discounts" && ratePlan.ratePlanCharges.nonEmpty,
-        ),
+        subscription.ratePlans.filter { ratePlan =>
+          val isDiscount = ratePlan.productName == "Discounts"
+          // we found that contrary to the docs, if effectiveStartDate and effectiveEndDate are today,
+          // zuora still returns it as a current-segment, so we need to ignore it
+          val hasCharges = ratePlan.ratePlanCharges.exists(_.effectiveEndDate != today)
```
johnduffell, line 71: We decided to only add the single test here, rather than replicating the full logic from zuora.  We could in future switch to fetching all charges, and applying all the filtering here if we want.

rBangay: test reply

MaelGNM (approved):
Looks good to me!

PR #2725, opened 2025-03-03
Discount expiry notifier - improve definition and segmentation of test data for functional and automated testing

david-pepper:
## What does this change?
Configure the test data that is used for both functional and automated testing

## Background
We do not have a bigquery sandbox, so we have been mocking the data for testing in the CODE environment. Now that work has started on writing automated tests, we need to segment this test data into automated test and functional test versions.

### Notes
The functional test data is currently reliant on subscriptions in the dev environment, so that callouts for sub status and payment amount can be achieved. An improvement on this would be to mock these responses if the solution is running in the CODE environment. It's possible that we'll pick this up in a future piece of work if time allows.


graham228221 (approved):
All looks good to me, there a few console.logs that seem to be duplicated but you might be keeping those for testing.

PR #2735, opened 2025-03-05
Discount expiry notifier - try combination of query and billing preview to get old amount if target date is today

david-pepper:
## What does this change?

For obtaining the invoice items in order to derive the old payment amount, implement a decision on whether to use the query or billing-preview depending  on whether the invoice target date is in the past (query) or future (billing-preview).

Handle the scenario where the target date is on the execution date, where the appropriate mechanism (query vs billing preview) is based on whether the bill run has taken place yet, which we can't know at the point of execution, so try the billing-preview first, and if there are no results, use the query.

## Testing
Testing will take place in production by inspecting the values returned for old payment amount

### Notes
- the sendemail() functionality is removed whilst we perform this testing.
- This functionality would likely be useful to extract into a zuora common module that provides a billing summary. Would like to do this in future PR if time allows


handlers/discount-expiry-notifier/src/handlers/getOldPaymentAmount.ts
```diff
@@ -54,15 +62,167 @@ export const handler = async (event: GetOldPaymentAmountInput) => {
...
+
+	return {
+		...parsedEvent,
+		lastPaymentDateBeforeDiscountExpiry,
+		oldPaymentAmount,
+	};
+};
+
+//when target date is today, whether we use the query or billing preview depends on whether a bill run
+//has been run today. If a bill run has been run today, we use the query, otherwise we use the billing preview
```
graham228221, line 111: this comment could probably be made more accurate?

```suggestion
//when target date is today, whether we use the query or billing preview depends on whether a bill run
//has been run today. We try the billing-preview first for the target date, then use a query if nothing is return from the billing-preview
```

graham228221 (approved):

handlers/discount-expiry-notifier/src/handlers/getOldPaymentAmount.ts
david-pepper: Yes, thanks for pointing out. I've struggled a bit with the automated tests for this, specifically with mocking certain functions, and it was holding me up quite a bit, so I'm going to focus on functional testing for now and come back to it

handlers/discount-expiry-notifier/test/handlers/getOldPaymentAmount.test.ts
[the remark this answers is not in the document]
david-pepper: Yes, thanks for pointing out. I've struggled a bit with the automated tests for this, specifically with mocking certain functions, and it was holding me up quite a bit, so I'm going to focus on functional testing for now and come back to it

handlers/discount-expiry-notifier/src/handlers/getOldPaymentAmount.ts
david-pepper: Good shout

PR #2744, opened 2025-03-18
Switch api discount

rBangay:
## What does this change?

This pr adds the ability to apply a discount to the switch journey from an annual contribution to a supporter plus.

We have also refactored some of the integration tests (contributionToSupporterPlusIntegration.test.ts) so that they now dynamically create the subscriptions that the tests then run off instead of relying on subscriptions having to be manually created each time the test suite is run.

This work was done in conjunction with the following work on manage-frontend (https://github.com/guardian/manage-frontend/pull/1504) as part of a new cancellation offer to annual contributors, offering them a switch to supporter plus at half price for 1 year.

### DIFF
--- a/handlers/product-switch-api/src/contributionToSupporterPlus.ts
+++ b/handlers/product-switch-api/src/contributionToSupporterPlus.ts
@@ -154,8 +154,6 @@ export const previewResponseFromZuoraResponse = (
 		'No supporter plus invoice item found in the preview response',
 	);
 
-	// const possibleDiscount: SwitchDiscountResponse | null = {};
-
 	const response: PreviewResponse = {
 		amountPayableToday: invoice.amount,
 		contributionRefundAmount,
@@ -166,22 +164,24 @@ export const previewResponseFromZuoraResponse = (
 			dayjs(supporterPlusSubscriptionInvoiceItem.serviceEndDate).add(1, 'days'),
 		),
 	};
-	// if (possibleDiscount) {
-	// 	response.discount = possibleDiscount;
-	// }
+
+	const possibleDiscount = invoice.invoiceItems.find(
+		(invoiceItem) =>
+			invoiceItem.processingType === 'Discount' &&
+			invoiceItem.chargeName === '50% off for 1 Year',
+	);
+	if (possibleDiscount) {
+		const discountPercentage = possibleDiscount.unitPrice;
+		response.discount = {
+			discountedPrice:
+				supporterPlusSubscriptionInvoiceItem.unitPrice +
+				possibleDiscount.amountWithoutTax,
+			discountPercentage: discountPercentage,
+			upToPeriods: 1,
+			upToPeriodsType: 'Years',
+		};
+	}
 	return response;
-	/*
-    amountPayableToday: number;
-	contributionRefundAmount: number;
-	supporterPlusPurchaseAmount: number;
-	nextPaymentDate: string;
-    discount?: {
-        discountedPrice:number;
-        upToPeriods: number; // 3,
-        upToPeriodsType: 'Months' | 'Years';
-        discountPercentage: number; // 25,
-    };
-    */
 };
 
 export const preview = async (
@@ -233,7 +233,10 @@ export const doSwitch = async (
 	return zuoraResponse;
 };
 
-const buildAddDiscountOrderAction = (discount: Discount, orderDate: Dayjs): OrderAction[] => {
+const buildAddDiscountOrderAction = (
+	discount: Discount,
+	orderDate: Dayjs,
+): OrderAction[] => {
 	return [
 		{
 			type: 'AddProduct',
@@ -406,6 +409,6 @@ export const buildSwitchRequestBody = (
 					...newTermOrderActions,
 				],
 			},
-		],	
+		],
 	};
 };
--- a/handlers/product-switch-api/src/contributionToSupporterPlus.ts
+++ b/handlers/product-switch-api/src/contributionToSupporterPlus.ts
@@ -16,6 +16,7 @@ import type { Dayjs } from 'dayjs';
 import dayjs from 'dayjs';
 import { removePendingUpdateAmendments } from './amendments';
 import type { CatalogInformation } from './catalogInformation';
+import type { Discount } from './discounts';
 import { takePaymentOrAdjustInvoice } from './payment';
 import { sendThankYouEmail } from './productSwitchEmail';
 import { sendSalesforceTracking } from './salesforceTracking';
@@ -32,11 +33,19 @@ import {
 import { sendToSupporterProductData } from './supporterProductData';
 import type { SwitchInformation } from './switchInformation';
 
+export interface SwitchDiscountResponse {
+	discountedPrice: number;
+	upToPeriods: number; // 3,
+	upToPeriodsType: 'Months' | 'Years';
+	discountPercentage: number; // 25,
+}
+
 export type PreviewResponse = {
 	amountPayableToday: number;
 	contributionRefundAmount: number;
 	supporterPlusPurchaseAmount: number;
 	nextPaymentDate: string;
+	discount?: SwitchDiscountResponse;
 };
 
 export type SwitchResponse = { message: string };
@@ -145,7 +154,9 @@ export const previewResponseFromZuoraResponse = (
 		'No supporter plus invoice item found in the preview response',
 	);
 
-	return {
+	// const possibleDiscount: SwitchDiscountResponse | null = {};
+
+	const response: PreviewResponse = {
 		amountPayableToday: invoice.amount,
 		contributionRefundAmount,
 		supporterPlusPurchaseAmount:
@@ -155,6 +166,22 @@ export const previewResponseFromZuoraResponse = (
 			dayjs(supporterPlusSubscriptionInvoiceItem.serviceEndDate).add(1, 'days'),
 		),
 	};
+	// if (possibleDiscount) {
+	// 	response.discount = possibleDiscount;
+	// }
+	return response;
+	/*
+    amountPayableToday: number;
+	contributionRefundAmount: number;
+	supporterPlusPurchaseAmount: number;
+	nextPaymentDate: string;
+    discount?: {
+        discountedPrice:number;
+        upToPeriods: number; // 3,
+        upToPeriodsType: 'Months' | 'Years';
+        discountPercentage: number; // 25,
+    };
+    */
 };
 
 export const preview = async (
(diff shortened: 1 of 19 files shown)


handlers/product-switch-api/src/discounts.ts
```diff
@@ -0,0 +1,53 @@
...
+		stage === 'PROD'
+			? '8a12994695aa4f680195ae2dc9d221d8'
+			: '71a1383e2b395842e6f58a2754ad00c1',
+	productRatePlanChargeId:
+		stage === 'PROD'
+			? '8a1280be95aa24270195ae3039934db7'
+			: '71a116628da95844dde58a2a18ef0059',
+	name: 'Cancellation Save - Annual Contribution to Supporter Plus Switch 50% off 1 year',
+	upToPeriods: 1,
+	upToPeriodsType: 'Years',
```
johnduffell, line 28: IIRC normally we input discounts as being for 12 Months in zuora, I'm not sure whether there's a good reason for it, or whether we should stick with the convention, but I'd hate to regret later

handlers/product-switch-api/src/switchInformation.ts
```diff
@@ -188,6 +191,16 @@ export const getSwitchInformationWithOwnerCheck = (
 	const startOfToday = today.startOf('day');
 	const startNewTerm = termStartDate.isBefore(startOfToday);
 
+	const maybeDiscount = getDiscount(
+		!!input.applyDiscountIfAvailable,
+		input.price,
```
johnduffell, line 196: should we be using their current amount from zuora rather than getting from the input?

handlers/product-switch-api/src/discounts.ts
```diff
@@ -0,0 +1,53 @@
...
+	contributionAmount: number,
+	supporterPlusPrice: number,
+	billingPeriod: BillingPeriod,
+	subscriptionStatus: string,
+	invoiceBalance: number,
+	stage: Stage,
+): Discount | undefined => {
+	const isEligibleForDiscount =
+		clientWantsADiscount &&
+		contributionAmount <= supporterPlusPrice &&
```
johnduffell, line 43: this should be the discounted price I think, not the full price?

handlers/product-switch-api/test/contributionToSupporterPlusIntegration.test.ts
```diff
@@ -74,6 +74,76 @@ describe('product-switching behaviour', () => {
 
 		expect(result).toMatchObject(expectedResult);
 	});
+	it('can preview an annual recurring contribution switch with 50% discount', async () => {
```
johnduffell, line 77: worth adding a gap to make it easier to pick out the different tests
```suggestion

	it('can preview an annual recurring contribution switch with 50% discount', async () => {
```

handlers/product-switch-api/src/discounts.ts
```diff
@@ -0,0 +1,53 @@
+import type { BillingPeriod } from '@modules/billingPeriod';
+import type { Stage } from '@modules/stage';
+
+export type Discount = {
+	// the following fields match the charge in the zuora catalog
+	// https://knowledgecenter.zuora.com/Zuora_Platform/API/G_SOAP_API/E1_SOAP_API_Object_Reference/ProductRatePlanCharge
```
johnduffell, line 6: SOAP api often comes up via google but I'm not sure it's that relevant nowadays is it?

handlers/product-switch-api/src/contributionToSupporterPlus.ts
```diff
@@ -32,11 +33,19 @@ import {
...
+	upToPeriodsType: 'Months' | 'Years';
+	discountPercentage: number; // 25,
+}
+
 export type PreviewResponse = {
 	amountPayableToday: number;
 	contributionRefundAmount: number;
 	supporterPlusPurchaseAmount: number;
 	nextPaymentDate: string;
+	discount?: SwitchDiscountResponse;
```
johnduffell, line 48: not sure why but I feel like the whole thing is the Response, subparts of it would just have a name without "response" on the end

I'm not sure if that's just my mind though, and it's not a major thing.
I remember many years ago we used "...Wire" to make it clear what gets goes over the wire but I don't think that's something we do much now.

handlers/product-switch-api/src/contributionToSupporterPlus.ts
```diff
@@ -155,6 +165,27 @@ export const previewResponseFromZuoraResponse = (
...
 	};
+
+	if (possibleDiscount) {
+		const discountInvoiceItem = invoice.invoiceItems.find(
+			(invoiceItem) =>
+				invoiceItem.productRatePlanChargeId ===
+				possibleDiscount.productRatePlanChargeId,
+		);
+		if (discountInvoiceItem) {
+			response.discount = {
```
johnduffell, line 176: took me a minute to work out what's happening here - I'm not sure if it's just my scala brain but I wouldn't use mutability in the logic, I would get a value back from this code and then explicitly insert it into the response object.

johnduffell:
might be worth adding more detail to both the description and the title of this PR, so that people in years to come will understand what and why it's happening

handlers/product-switch-api/src/discounts.ts
rBangay: If we could leave it like this then we don't have to do any date/time period translation in the client

handlers/product-switch-api/src/switchInformation.ts
rBangay: It would be nice if the client didn't have to pass through this value at all if possible

handlers/product-switch-api/src/discounts.ts
johnduffell: I can tell you've never experienced the wrath of a zuora malfunction!  I can't think of any obvious problem as long as the billing period of the main charge and discount match, there's not currently any way to switch product from supporter plus, but that would cause issues whether we have it monthly or annually.

handlers/product-switch-api/src/discounts.ts
```diff
@@ -0,0 +1,53 @@
...
+	subscriptionStatus: string,
+	invoiceBalance: number,
+	stage: Stage,
+): Discount | undefined => {
+	const isEligibleForDiscount =
+		clientWantsADiscount &&
+		contributionAmount <= supporterPlusPrice &&
+		billingPeriod === 'Annual' &&
+		subscriptionStatus === 'Active' &&
+		invoiceBalance === 0;
```
johnduffell, line 46: As future proofing it might be worth checking that there are no current discounts in force like we do on discount api.  We basically do an invoice preview and check the next invoice only has positive items.  This applies across all products.
Alternatively we could add a comment into the code to prompt the next person to do that (as I think it's low risk for annual RC)

handlers/product-switch-api/src/contributionToSupporterPlus.ts
```diff
@@ -32,11 +33,19 @@ import {
 import { sendToSupporterProductData } from './supporterProductData';
 import type { SwitchInformation } from './switchInformation';
 
+export interface SwitchDiscountResponse {
+	discountedPrice: number;
+	upToPeriods: number; // 3,
+	upToPeriodsType: 'Months' | 'Years';
+	discountPercentage: number; // 25,
```
johnduffell, line 40:
```suggestion
	upToPeriods: number;
	upToPeriodsType: 'Months' | 'Years';
	discountPercentage: number;
```

handlers/product-switch-api/src/contributionToSupporterPlus.ts
rBangay: I guess with my js/typscript hat on, const does not allow me to reassign the variable so if I had:

```
const myObj = {'dobedo': 'lalala'}

// I can't do this: myObj = {'dobedo': 'hi there'}
```

But there is nothing stopping me from changing the inner content of the object:

```
const myObj = {'dobedo': 'lalala'}
myObj.dobedo = 'hi there' // this is fine
```

If you want an object to be purely immutable in js then you can use the Object.freeze method (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/freeze).

I guess its like if I had a box that was a const I couldn't change the box itself, but I can play around with the contents as much as I like.

Const in this context doesn't mean the variable is imutable it just means that it cannot be reassigned.

As for whether or not this is good practice or not I'm open to suggestions, but it is valid to do this in js/ts.

handlers/product-switch-api/src/contributionToSupporterPlus.ts
johnduffell: probably not important

handlers/product-switch-api/test/contributionToSupporterPlusIntegration.test.ts
johnduffell: probably not important

handlers/product-switch-api/src/contributionToSupporterPlus.ts
johnduffell: I put the alternative in here as it was too hard to explain in a comment.
https://github.com/guardian/support-service-lambdas/pull/2811/files
It's a PR into this branch, to make it easier to compare.
Feel free to pull it into your PR if you like it, or leave it if you don't.

handlers/product-switch-api/src/contributionToSupporterPlus.ts
rBangay: SwitchDiscountResponsePart?

andresilva-guardian (approved):
Nice job @rBangay 💪  , thank you for your effort! 🚀

handlers/product-switch-api/test/contributionToSupporterPlusIntegration.test.ts
```diff
@@ -74,6 +74,76 @@ describe('product-switching behaviour', () => {
...
+				discountedPrice: 60,
+				discountPercentage: 50,
+				upToPeriods: 1,
+				upToPeriodsType: 'Years',
+			},
+		};
+
+		expect(result).toEqual(expectedResult);
+	});
+	it('can preview an annual recurring contribution (non UK - German) switch with 50% discount', async () => {
```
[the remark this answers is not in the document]
johnduffell: this was added in this file `"preview of annual recurring contribution switch with 50% discount fails validation check"`

handlers/product-switch-api/src/supporterProductData.ts
```diff
@@ -20,16 +20,6 @@ export type ContributionAmount = { amount: number; currency: string };
...
 ): SupporterRatePlanItem => {
-	const contributionAmount =
-		switchInformation.contributionAmount > 0
-			? {
-					contributionAmount: {
-						amount: switchInformation.contributionAmount,
-						currency: 'GBP',
-					},
-				}
-			: {};
```
johnduffell, line 31: we couldn't work out what to put in as the amount when there's a time limited discount.
After investigation, nothing uses the contribution amount for supporter plus any more, and it should be removed.

handlers/product-switch-api/src/switchInformation.ts
```diff
@@ -171,8 +177,30 @@ export const getSwitchInformationWithOwnerCheck = (
...
+
+	const contributionAmount = Math.max(0, previousAmount - actualBasePrice);
+
+	/*
+	 * existing situation: the price is passed in when someone switches.
+	 * a min of the supporterplus base price is passed through, more if the user already contributes more than the min
+	 *
+	 * propsoal: continue to pass through the price from the client, however also calculate this figure on the backend and use the 2 as a comparison.
+	 *
+	 */
```
johnduffell, line 203: this is a WIP comment and should be either deleted or reworded to clarify

PR #2745, opened 2025-03-18
update AWS sdk v2 version to latest

johnduffell:
we need the latest version of netty to avoid https://github.com/advisories/GHSA-4g8c-wm8x-jfhw


rBangay (approved):
👍

PR #2749, opened 2025-03-19
add workflow queues to value team

johnduffell:
The queues are in a different stack from the app,so adding the queues as well.


michaelbjacobson (approved):
LGTM!

PR #2753, opened 2025-03-21
infra: setup application observer-data-export

AndreaDiotallevi:
## What does this change?

[Trello](https://trello.com/c/Hv8pCgD3/489-observer-sale-s3-bucket-for-sharing-subscriber-data)

Setup application in charge of creating, updating & managing the export of Observer-only data.

## Details

```ts
const app = 'observer-data-export';

const bucket = new Bucket(this, 'Bucket', {
	bucketName: `${app}-${this.stage.toLowerCase()}`,
	lifecycleRules: [{ expiration: Duration.days(28) }],
});

const unifidaUser = new User(this, 'UnifidaUser', {
	userName: `unifida-${this.stage.toLowerCase()}`,
});

bucket.grantRead(unifidaUser);
```

- Create S3 bucket with name: `observer-data-export-code` & `observer-data-export-prod`
- Create IAM [user](https://us-east-1.console.aws.amazon.com/iam/home?region=eu-west-1#/users) with name: `unifida-code` & `unifida-prod`
- Grant user read-only permission to the full bucket

## Next

Generate access keys for both IAM users and send them to Unifida's team (I don't have permissions to create keys).


JoeMitchellGuardian (approved):

cdk/lib/observer-data-export.ts
```diff
@@ -0,0 +1,47 @@
...
+			fileName: `${app}.zip`,
+			runtime: nodeVersion,
+			timeout: Duration.seconds(300),
+			environment: { Stage: this.stage },
+		};
+
+		new GuLambdaFunction(this, 'PlaceholderLambda', {
+			...lambdaDefaultConfig,
+			handler: 'placeholder.handler',
+			functionName: `placeholder-${this.stage}`,
```
graham228221, line 44: Do we need a Lambda as part of the stack setup or is it purely a placeholder for convenience later? 

Wondering if it would be better to call it something more specific.

graham228221: `Wondering if it would be better to call it something more specific.` like observer-data-placeholder or similar

AndreaDiotallevi: Unfortunately the CI has been setup so that it expects a lambda to be present.
I have updated the name to something specific to the application.

PR #2756, opened 2025-03-24
Discount expiry notifier - exclude delivery and holiday credits from new payment amount

david-pepper:
## What does this change?
Exclude delivery and holiday credits from new payment amount


handlers/discount-expiry-notifier/src/handlers/getOldPaymentAmount.ts
```diff
@@ -206,7 +206,7 @@ export const getPastInvoiceItems = async (
 };
 
 const query = (subName: string, serviceStartDate: string): string =>
-	`SELECT chargeAmount, taxAmount, serviceStartDate, subscriptionNumber FROM InvoiceItem WHERE subscriptionNumber = '${subName}' AND ServiceStartDate = '${serviceStartDate}' AND ChargeName!='Delivery-problem credit' AND ChargeName!='Holiday Credit'`;
+	`SELECT chargeName, chargeAmount, taxAmount, serviceStartDate, subscriptionNumber FROM InvoiceItem WHERE subscriptionNumber = '${subName}' AND ServiceStartDate = '${serviceStartDate}' AND ChargeName!='Delivery-problem credit' AND ChargeName!='Holiday Credit'`;
```
graham228221, line 209: Do you defintely want to filter them out here AND exclude them in the `filterRecords` function?

AndreaDiotallevi (approved):

graham228221 (approved):
Something makes me uncomfortable relying on charge names, but we probably do this in a number of places already and there's I can't think of an alternative straightforward and safe way to exclude them (join to the rate plan charge and exclude one-off charges?)

handlers/discount-expiry-notifier/src/handlers/getOldPaymentAmount.ts
david-pepper: the query is for past invoice items (old payment amount), and the filterRecords function is for future invoice items (new payment amount), so yes, need both

PR #2768, opened 2025-04-02
`soft-opt-in-consent-setter` CDK Migration (1)

michaelbjacobson:
This is the initial step in migrating the existing `soft-opt-in-consent-setter` CloudFormation template to use the GuCDK instead. This passes the existing vanilla CFN YAML file into the CDK, which generates a CDK-style template from it.

To ensure this new old template deploys correctly the relevant CI and Riff-Raff configs have been adjusted.

### DIFF
--- a/handlers/soft-opt-in-consent-setter/riff-raff.yaml
+++ b/handlers/soft-opt-in-consent-setter/riff-raff.yaml
@@ -6,19 +6,21 @@ allowedStages:
   - CODE
   - PROD
 deployments:
-  cfn:
+  soft-opt-in-consent-setter-cloudformation:
     type: cloud-formation
     app: soft-opt-in-consent-setter
     parameters:
-      templatePath: cfn.yaml
+      templateStagePaths:
+        CODE: soft-opt-in-consent-setter-CODE.template.json
+        PROD: soft-opt-in-consent-setter-PROD.template.json
 
   soft-opt-in-consent-setter:
     type: aws-lambda
     parameters:
       fileName: soft-opt-in-consent-setter.jar
-      bucket: support-service-lambdas-dist
+      bucketSsmLookup: true
       prefixStack: false
       functionNames:
         - soft-opt-in-consent-setter-
         - soft-opt-in-consent-setter-IAP-
-    dependencies: [ cfn ]
+    dependencies: [soft-opt-in-consent-setter-cloudformation]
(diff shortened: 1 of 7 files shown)


handlers/soft-opt-in-consent-setter/riff-raff.yaml
```diff
@@ -6,19 +6,21 @@ allowedStages:
...
+      templateStagePaths:
+        CODE: soft-opt-in-consent-setter-CODE.template.json
+        PROD: soft-opt-in-consent-setter-PROD.template.json
 
   soft-opt-in-consent-setter:
     type: aws-lambda
     parameters:
       fileName: soft-opt-in-consent-setter.jar
-      bucket: support-service-lambdas-dist
+      bucketSsmLookup: true
```
johnduffell, line 21: strange that (existing) the lambda function is read from `support-service-lambdas-dist` but gets an IAM permission to read from membership-dist?

johnduffell: It's not really consistent any more
we used to upload to support-service-lambdas-dist
and now we are uploading to membership-dist 
The actual CDK/CFN is pointing at support-service-lambdas-dist but the deploy is overriding that back to membership-dist after it's cloudformed?

Presumably if we are making this change, we should also change the CDK to point to the one in membership-dist, and delete the one in support-service-lambdas-dist so it doesn't confuse things.

johnduffell:
looking at the "old" yaml it looks like there are a load of changes to it, including linting and moving away from the serverless language.  Is this just an automated transformation or should I be reviewing it more carefully?

michaelbjacobson:
> looking at the "old" yaml it looks like there are a load of changes to it, including linting and moving away from the serverless language. Is this just an automated transformation or should I be reviewing it more carefully?

Thanks, @johnduffell! Sorry if this was unclear.

Basically I went to the AWS console for this CFN stack, and in the Template section I viewed the version of it post-transformation. The AWS console always gives this in JSON, so I converted it to YAML to minimise changes to the original.

I might be wrong, or out of date, but I believe this is recommended when passing a CFN template into the `CfnInclude` construct: https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.cloudformation_include.CfnInclude.html

BUT after making this change I also manually began an update of the `CODE` stack, using the new YAML template version, and there were no changes whatsoever in the preview, so I'm confident that you don't need to pay much attention to these changes.

johnduffell (approved):

PR #2204, opened 2024-03-28
Fix soft opt in consent setter idapi call

tjmw:
## What does this change?

Revert the runtime of the soft-opt-in-consent-setter lambdas to Java 11. This is to fix an error we've been seeing which is causing the `PATCH` request to IDAPI to set the consents to fail:

```
IdentityConnector: Identity request failed: java.lang.reflect.InaccessibleObjectException: Unable to make field private final sun.net.www.protocol.https.DelegateHttpsURLConnection sun.net.www.protocol.https.HttpsURLConnectionImpl.delegate accessible: module java.base does not "opens sun.net.www.protocol.https" to unnamed module @6cc4c815
```

## Accessibility

-   [ ] [Tested with screen reader](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#screen-reader)
-   [ ] [Navigable with keyboard](https://github.com/guardian/accessibility/blob/main/people-and-technology/02-physical.md#keyboard)
-   [ ] [Colour contrast passed](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#contrast)
-   [ ] [The change doesn't use only colour to convey meaning](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#use-of-colour)

PR #2769, opened 2025-04-02
`soft-opt-in-consent-setter` CDK Migration (2)

michaelbjacobson:
This PR recreates the existing `soft-opt-in-consent-setter` CloudFormation infrastructure in Typescript using the AWS CDK, with GuCDK constructs where practical.

Stateful resources, such as queues (which might have events in-flight) and DynamoDB tables, have had their logical IDs overridden to prevent the current instances from being deleted instead of updated. [See this doc for more info](https://github.com/guardian/cdk/blob/main/docs/stateful-resources.md).

Here's the CFN preview changeset for these changes. I've colour coded rows to make it easier to see which new construct replaces which old one. The stateful resources are modified, as opposed to deleted and recreated. Everything else is a straight swap with a new unique Logical ID, except for the two new Lambda policies which (I think) are generated automatically by the GuCDK.

Screenshot 2025-04-03 at 10 53 03

[You can view the part one PR here.](https://github.com/guardian/support-service-lambdas/pull/2768)


handlers/soft-opt-in-consent-setter/cfn.yaml
```diff
@@ -274,7 +274,7 @@ Resources:
           sfApiVersion: v46.0
       Tags:
         - Key: lambda:createdBy
-          Value: SAM
+          Value: SAMl
```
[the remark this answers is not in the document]
michaelbjacobson: I think that's a typo where I accidentally pressed the 'L' key with my cursor there and didn't notice

cdk/lib/soft-opt-in-consent-setter.ts
```diff
@@ -1,19 +1,392 @@
...
+				noMonitoring: true,
+			},
+			rules: [
+				{
+					schedule: Schedule.expression(props.schedule),
+					description: 'Runs Soft Opt-In Consent Setter',
+				},
+			],
+			app: '',
+			fileName: '',
```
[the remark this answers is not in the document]
michaelbjacobson: Yes! Good catch! I meant to add TODOs there!

cdk/lib/soft-opt-in-consent-setter.ts
```diff
@@ -1,19 +1,392 @@
...
+			role: lambdaFunctionRole,
+			functionName: `soft-opt-in-consent-setter-${this.stage}`,
+			runtime: Runtime.JAVA_11,
```
johnduffell, line 185: if you feel like bumping to them to java 21 that would give an efficiency boost (fine if you want to avoid changes at this moment)

michaelbjacobson: I'd sooner keep to basically only the necessary changes in the migration, but deffo worth bumping afterwards as part of the overall `soft-opt-in-consent-setter` work.

cdk/lib/soft-opt-in-consent-setter.ts
```diff
@@ -1,19 +1,392 @@
...
+			metricName: 'failed_salesforce_update',
+			namespace: 'soft-opt-in-consent-setter',
+			period: 3600,
+			statistic: 'Sum',
+			threshold: 1,
+			treatMissingData: 'notBreaching',
+		});
+		failedUpdateAlarm.cfnOptions.condition = isProd;
+
+		const failedDynamoUpdateAlarm = new CfnAlarm(
```
johnduffell, line 340: might be worth testing these alarms by setting the state to ALARM once it's shipped.  I think the alarm handler uses the App tag, and that can get affected by what's in the riff-raff.yaml I think?
(If not, they will go to the SRE channel.  I noticed that was the case with another one recently, hence why it's on my mine.)

cdk/lib/soft-opt-in-consent-setter.ts
graham228221: @johnduffell we had to revert that last time we tried it 😁  https://github.com/guardian/support-service-lambdas/pull/2204

cdk/lib/soft-opt-in-consent-setter.ts
```diff
@@ -0,0 +1,392 @@
...
+			}),
+		);
+		lambdaFunctionIAPRole.addToPolicy(
+			new PolicyStatement({
+				actions: [
+					'secretsmanager:DescribeSecret',
+					'secretsmanager:GetSecretValue',
+				],
+				resources: [
+					'arn:aws:secretsmanager:eu-west-1:865473395570:secret:CODE/Salesforce/ConnectedApp/AwsConnectorSandbox-jaCgRl',
```
JuliaBrigitte, line 158: I asked @jorgeazevedo if there is a way to not hard code the account id and was pointed here: https://github.com/guardian/platform/blob/7af80e5e9458a0773e0d1de01f0b340709510a9e/cdk/lib/admin.ts#L53

cdk/lib/soft-opt-in-consent-setter.ts
```diff
@@ -0,0 +1,394 @@
...
+				),
+			}),
+		);
+
+		// Lambda Functions
+		const lambdaFunction = new GuScheduledLambda(this, 'LambdaFunction', {
+			app: 'soft-opt-in-consent-setter',
+			fileName: 'soft-opt-in-consent-setter.jar',
+			monitoringConfiguration: {
+				noMonitoring: true,
```
jorgeazevedo, line 181: Have you considered leveraging the monitoring configuration option available in the pattern to avoid defining the alarms for both lambdas independently a few lines below? You can pass the property from `bin/cdk` with different values for PROD and CODE, so the code ends up looking like this https://github.com/guardian/devx-backup-monitoring/blob/main/cdk/bin/cdk.ts#L20

I get the sense that the cdk code is meant to mimic the yaml quite closely by design, right? I think that's totally valid, just wanted to put a few of these alternative ways of doing it in your radar, maybe for a second pass.

michaelbjacobson: I'm inclined to keep thing as they are for now, mirroring the old YAML config as much as possible. But yeah, absolutely something I'd like to implement in a follow-up.

cdk/lib/soft-opt-in-consent-setter.ts
```diff
@@ -186,7 +186,7 @@ export class SoftOptInConsentSetter extends GuStack {
 					description: 'Runs Soft Opt-In Consent Setter',
 				},
 			],
-			functionName: `soft-opt-in-consent-setter-${this.stage}-CDK`,
+			functionName: `soft-opt-in-consent-setter-${this.stage}`,
 			runtime: Runtime.JAVA_11,
```
johnduffell, line 190:
```suggestion
			runtime: Runtime.JAVA_11, // keep on 11 for now due to http PATCH issue
```

cdk/lib/soft-opt-in-consent-setter.ts
```diff
@@ -201,7 +201,7 @@ export class SoftOptInConsentSetter extends GuStack {
 			app: 'soft-opt-in-consent-setter',
 			fileName: 'soft-opt-in-consent-setter.jar',
 			role: lambdaFunctionIAPRole,
-			functionName: `soft-opt-in-consent-setter-IAP-${this.stage}-CDK`,
+			functionName: `soft-opt-in-consent-setter-IAP-${this.stage}`,
 			runtime: Runtime.JAVA_11,
```
johnduffell, line 205:
```suggestion
			runtime: Runtime.JAVA_11, // keep on 11 for now due to http PATCH issue
```

johnduffell (approved):
nice work! once it's all deployable in CODE I think it's good to go, if something goes wrong at runtime, the items should stay on the queue/in salesforce until it's deployed back to the old version.

cdk/lib/__snapshots__/soft-opt-in-consent-setter.test.ts.snap
```diff
@@ -421,7 +421,7 @@ exports[`The SoftOptInConsentSetter stack matches the snapshot 1`] = `
             ],
           },
         ],
-        "RoleName": "SoftOptInConsentSetter-LambdaFunctioIAPnRole-CDK",
+        "RoleName": "SoftOptInConsentSetter-LambdaFunctioIAPnRole",
```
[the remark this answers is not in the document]
michaelbjacobson: Yes! Thanks!

cdk/lib/soft-opt-in-consent-setter.ts
michaelbjacobson: Will do 👍🏼

PR #2774, opened 2025-04-03
Remove Sunday rate plan from new-product-api

graham228221:
## What does this change?
Removes the Sunday (Observer) rate plan from the product catalog returned by new-product-api. The Sunday+ option can still be offered to customers.

This is a temporary change - we need to take this off sale during the handover period. Not 100% sure when it will be bought back as there is some additional work required to guard against Sunday-only subscriptions being taken out into the wrong payment gateway - it will definitely be after 22 April 2025.

I'm removing on server side because it's much simpler than doing it in Salesforce.

## How to test
Call the `/product-catalog` endpoint or load up the New Subscription page in Salesforce:

Screenshot 2025-04-03 at 18 21 30

Disrupting CSR acquisitions would be a costly problem, so we should be ready to roll this change back if we spot or hear about any problems.


graham228221:
Adding `DO NOT MERGE` until we get the green light to make this change.

JoeMitchellGuardian (approved):
Tested in Salesforce - looks good to me 👍

paulbrown1982 (approved):
👍 LGTM. We should raise a revert PR as soon as this is merged, so we can more easily roll it back when required.

JoeMitchellGuardian:
https://github.com/guardian/support-service-lambdas/pull/2778 revert PR if needed @paulbrown1982

PR #2775, opened 2025-04-03
ensure consistent role names by avoiding referencing stack

michaelbjacobson:
There was an issue where the mobile account coudln't assume the role to write to our SQS queue any more.  It turned out that the role had changed from 
membership-PROD-soft-opt-in-consent-setter-QueueCrossAccountRole
to
soft-opt-in-consent-setter-PROD-QueueCrossAccountRole
as the old one was using the AWS stack name at deploy time as a prefix, but the new one was using the CDK's local stack name parameter.

This is a small PR to adjust the names of AWS IAM Roles within the CDK definition. It seems that `this.stackName` differs from `AWS::StackName`. We are only referencing the stage dynamically, everything else is hard coded in the interest of clarity and predictability.


johnduffell (approved):
fingers crossed

PR #2776, opened 2025-04-03
Add rule to Acquisitions event bus to pass events to `soft-opt-in-consent-setter-queue`

michaelbjacobson:
This updates the `soft-opt-in-consent-setter` CDK template to apply a new rule to the [Acquisitions event bus](https://github.com/guardian/support-frontend/blob/b67bc8fbd61716935bba0ac6085cef4e38a12254/cdk/lib/bigquery-acquisitions-publisher.ts#L27) to send events from Support Workers to the existing soft opt in SQS queue.

This is essentially a repurposing of the initial [proposed implementation in the Support Workers repo](https://github.com/guardian/support-frontend/pull/6893), which was abandoned in favour of making the change here.

NB: I've checked the CDK pnpm lockfile into version control for this PR. We didn't have it in the repo previously, but I think it's considered best practice. We're checking it into version control in [these repos](https://github.com/search?q=org%3Aguardian+cdk%2Fpnpm-lock.yaml&type=code) and, although we're using pnpm now, [the Yarn docs explicitly advise including the Yarn lockfile in version control](https://classic.yarnpkg.com/en/docs/migrating-from-npm).


cdk/lib/soft-opt-in-consent-setter.ts
```diff
@@ -411,5 +414,38 @@ export class SoftOptInConsentSetter extends GuStack {
...
+				new SqsQueue(softOptInsQueue, {
+					message: aws_events.RuleTargetInput.fromObject({
+						subscriptionId: aws_events.EventField.fromPath(
+							'$.detail.zuoraSubscriptionNumber',
+						),
+						identityId: aws_events.EventField.fromPath('$.detail.identityId'),
+						eventType: 'Acquisition',
+						productName: aws_events.EventField.fromPath(
+							'$.detail.ZuoraProductName',
+						),
```
johnduffell, line 444: I think we need to add this field here: https://github.com/guardian/support-frontend/blob/4906e5f65befeff42259862558bf4f54488cf62d/support-workers/src/main/scala/com/gu/acquisitions/AcquisitionDataRowBuilder.scala#L61-L93
Also, I notice that everywhere else in that structure is `thisWay` rather than `ThisWay` so I think we should keep that convention when adding new fields.
```suggestion
						productName: aws_events.EventField.fromPath(
							'$.detail.zuoraProductName',
						),
```

johnduffell, line 444:
```suggestion
						productName: aws_events.EventField.fromPath(
							'$.detail.product',
						),
						printOptions: aws_events.EventField.fromPath(
							'$.detail.printOptions',
						),
```

johnduffell (approved):
nice one, looks good

MaelGNM (approved):

PR #2779, opened 2025-04-04
`soft-opt-in-consent-setter` Lambda Permissions Fix

michaelbjacobson:
This fixes a mistake whereby the `lambdaFunctionRole` wasn't attached to its Lambda. It also ensures that all necessary inline policies are included in this role.


MaelGNM (approved):
LGTM!

PR #2785, opened 2025-04-09
move test subscription creation into shared directory

rBangay:
## What does this change?

Moves the creation of the digital subscription and supporter plus test products into a shared directory so that they can be used in integration tests where we currently rely on a manually created subscription. This should enable a refactor of the (product-switch) integration tests so that they can be run in CI with dynically created test subs.

### DIFF
--- a/handlers/discount-api/test/helpers.ts
+++ b/handlers/discount-api/test/helpers.ts
@@ -1,116 +1,13 @@
-import { getIfDefined } from '@modules/nullAndUndefined';
 import type { ZuoraClient } from '@modules/zuora/zuoraClient';
 import type {
-	ZuoraSubscribeResponse,
 	ZuoraSubscription,
 	ZuoraSuccessResponse,
 } from '@modules/zuora/zuoraSchemas';
 import {
-	zuoraSubscribeResponseSchema,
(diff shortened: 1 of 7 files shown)


handlers/discount-api/test/helpers.ts
```diff
@@ -1,117 +1,14 @@
...
-	zuoraSubscribeResponseSchema,
 	zuoraSuccessResponseSchema,
 } from '@modules/zuora/zuoraSchemas';
 import type { Dayjs } from 'dayjs';
-import dayjs from 'dayjs';
-import { digiSubSubscribeBody } from './fixtures/request-bodies/digitalSub-subscribe-body-old-price';
-import { supporterPlusSubscribeBody } from './fixtures/request-bodies/supporterplus-subscribe-body-tier2';
 import { updateSubscriptionBody } from './fixtures/request-bodies/update-subscription-body';
 
-export const createDigitalSubscription = async (
```
johnduffell, line 18: these functions are moved to `modules/zuora/test/`

johnduffell (approved):
looks good, (although I did pair on it so maybe I'm not the most independent reviewer)

need to run `pnpm fix-formatting`  I think

PR #2793, opened 2025-04-17
Allow dynamic creation of test contributions for use in integration tests

rBangay:
## What does this change?

Allows contributions to be created dynamically so that they can be used in the integration tests (contributionToSupporterPlusIntegration.test.ts for now) instead of having to manually create the contributions.

### DIFF
--- a/handlers/product-switch-api/src/contributionToSupporterPlus.ts
+++ b/handlers/product-switch-api/src/contributionToSupporterPlus.ts
@@ -192,7 +192,7 @@ export const doSwitch = async (
 
 	const requestBody = buildSwitchRequestBody(dayjs(), productSwitchInformation);
 	const zuoraResponse: ZuoraSwitchResponse = await zuoraClient.post(
-		'v1/orders',
+		'v1/orders?returnIds=true',
 		JSON.stringify(requestBody),
 		zuoraSwitchResponseSchema,
 	);
--- a/handlers/product-switch-api/src/payment.ts
+++ b/handlers/product-switch-api/src/payment.ts
@@ -45,14 +45,13 @@ export const takePaymentOrAdjustInvoice = async (
 	accountId: string,
 	paymentMethodId: string,
 ): Promise<number> => {
-	const invoiceNumber = getIfDefined(
-		switchResponse.invoiceNumbers?.at(0),
+	const invoiceId = getIfDefined(
+		switchResponse.invoiceIds?.at(0),
 		'No invoice number found in the switch response',
 	);
 
-	const invoice = await getInvoice(zuoraClient, invoiceNumber);
+	const invoice = await getInvoice(zuoraClient, invoiceId);
 	const amountPayableToday = invoice.amount;
-	const invoiceId = invoice.id;
 
 	if (amountPayableToday === 0) {
 		// Nothing to do, we don't need to take a payment and the account balance will be correct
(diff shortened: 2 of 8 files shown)


handlers/product-switch-api/src/contributionToSupporterPlus.ts
```diff
@@ -192,7 +192,7 @@ export const doSwitch = async (
 
 	const requestBody = buildSwitchRequestBody(dayjs(), productSwitchInformation);
 	const zuoraResponse: ZuoraSwitchResponse = await zuoraClient.post(
-		'v1/orders',
+		'v1/orders?returnIds=true',
```
johnduffell, line 195: this is needed to get the invoiceIds field 
image
(from https://developer.zuora.com/v1-api-reference/api/operation/POST_Order/ )

handlers/product-switch-api/src/payment.ts
```diff
@@ -45,14 +45,13 @@ export const takePaymentOrAdjustInvoice = async (
...
 ): Promise<number> => {
-	const invoiceNumber = getIfDefined(
-		switchResponse.invoiceNumbers?.at(0),
+	const invoiceId = getIfDefined(
+		switchResponse.invoiceIds?.at(0),
 		'No invoice number found in the switch response',
 	);
 
-	const invoice = await getInvoice(zuoraClient, invoiceNumber);
+	const invoice = await getInvoice(zuoraClient, invoiceId);
```
johnduffell, line 53: according to the zuora docs, getInvoice can take an ID or number equally well

handlers/product-switch-api/test/contributionToSupporterPlusIntegration.test.ts
```diff
@@ -126,15 +129,20 @@ describe('product-switching behaviour', () => {
...
+			);
+
+			const switchResponse = await doSwitch(zuoraClient, switchInformation);
 
 			const response = await adjustNonCollectedInvoice(
 				zuoraClient,
-				invoiceId,
-				1.33,
-				'8ad08cbd8586721c01858804e3715378',
+				switchResponse.invoiceIds?.[0] ?? '',
```
johnduffell, line 143: I'd rather not go ahead with the call if the value isn't present
```suggestion
      const invoiceId = getIfDefined(
        switchResponse.invoiceIds?.[0],
        'invoice id was undefined in response from Zuora',
	    );
      
			const response = await adjustNonCollectedInvoice(
				zuoraClient,
				invoiceId,
```

modules/zuora/test/fixtures/request-bodies/contribution-subscribe-body.ts
```diff
@@ -0,0 +1,78 @@
+import type { Dayjs } from 'dayjs';
+import { zuoraDateFormat } from '@modules/zuora/common';
+import { catalog } from '../../../../../handlers/discount-api/src/productToDiscountMapping';
```
johnduffell, line 3: something we were slightly unsure about was referencing handlers from modules.  I would avoid that in scala by moving the relevant code to an existing or new module, but since TS seems to be more file based, maybe it's ok?
We briefly considered moving/merging `catalog` to the zuora-catalog module but didn't delve into it that deeply.

johnduffell (approved):
nice looks good to me! I think it's great to get these tests more repeatable and I do hope we can automate any that are suitable in future.

PR #2801, opened 2025-04-24
Implement user-selected opt-ins and outs functionality

michaelbjacobson:
This PR updates the consent-setting behaviour for Similar Guardian Products by conditionally overriding it if the user has made a preference at checkout.

This PR updates the behaviour as follows:
- All products for which users have previously been opted-in to SGP, as the default behaviour, will continue to do so...
- ... Unless the user explicitly opts-out

This also:
- Updates relevant existing unit tests to work with refactors
- Adds unit tests for new functionality

PR #2802, opened 2025-04-24
make consent setter clear SGP consent down when requested

johnduffell:
I wanted to make some complex suggestions, which is not possible to do clearly via "PR comments" so I made them in a branch instead with this PR as a diff against the actual one https://github.com/guardian/support-service-lambdas/pull/2801

Of course they are only suggestions just like PR comments, so feel free to copy what makes sense, and leave what doesn't.  If you want to accept them wholesale as a starting point, feel free to merge this branch into yours.

Hopefully it's helpful even though I didn't get a chance to pair on this.

See inline comments for my explanations


handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/HandlerIAP.scala
```diff
@@ -34,7 +34,7 @@ object HandlerIAP extends LazyLogging with RequestHandler[SQSEvent, Unit] {
   }
 
   case class UserConsentsOverrides(
-      similarGuardianProducts: Option[Boolean]
+      similarGuardianProducts: Option[Boolean],
```
johnduffell, line 37: scalafmt added trailing commas

handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/HandlerIAP.scala
```diff
@@ -150,20 +150,16 @@ object HandlerIAP extends LazyLogging with RequestHandler[SQSEvent, Unit] {
...
-      consentsBody = consentsCalculator.buildConsentsBody(amendedConsents, state = true)
-
+      softConsentKeys <- consentsCalculator.getSoftOptInsByProduct(message.productName)
+      softConsents = softConsentKeys.map(_ -> true).toMap
+      explicitConsents =
+        for {
+          userConsentsOverrides <- message.userConsentsOverrides
+          consented <- userConsentsOverrides.similarGuardianProducts
+        } yield similarGuardianProducts -> consented
+      consentsBody = consentsCalculator.buildConsentsBody(softConsents ++ explicitConsents)
```
johnduffell, line 160: the existing code would have set it if we said `true` and left it as is for undefined and `false`.  We need to clear it down if it's `false`.

handlers/soft-opt-in-consent-setter/src/test/scala/com.gu.soft_opt_in_consent_setter/ConsentsCalculatorTests.scala
```diff
@@ -126,10 +126,6 @@ class ConsentsCalculatorTests extends AnyFlatSpec with should.Matchers with Eith
       Set("Guardian Weekly"),
       calculator,
     ) shouldBe Right("""[
-        |  {
-        |    "id" : "similar_guardian_products",
-        |    "consented" : false
-        |  },
```
johnduffell, line 132: I don't think this test was correct in the first place, we would not have wanted to clear SGP when switching to GW!  Although that would have been a pretty rare case so almost no impact.

handlers/soft-opt-in-consent-setter/src/test/scala/com.gu.soft_opt_in_consent_setter/HandlerTests.scala
```diff
@@ -121,7 +121,7 @@ class HandlerTests extends AnyFunSuite with Matchers with MockFactory {
         """[
           |]""".stripMargin,
       )
-      .never
+      .never()
```
johnduffell, line 124: that was a warning

handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/HandlerIAP.scala
johnduffell: we would need to add tests for this logic, I would say extract this chunk into a new method in consentsCalculator and add tests to just make sure SGP appears as true or false as applicable.

handlers/soft-opt-in-consent-setter/src/test/scala/com.gu.soft_opt_in_consent_setter/testData/SFSubscriptionTestData.scala
```diff
@@ -63,12 +63,12 @@ object SFSubscriptionTestData {
   val overlappingAssociatedSub2 = SFAssociatedSubRecord("contribution", identityId)
   val nonOverlappingAssociatedSub2 = SFAssociatedSubRecord("contribution", s"identityId${0}")
 
-  val associatedSubsWithOverlap2: Seq[SFAssociatedSubRecord] = Seq(
+  lazy val associatedSubsWithOverlap2: Seq[SFAssociatedSubRecord] = Seq(
```
johnduffell, line 66: these were forward references and the compiler was warning about them so I just made them lazy as a quick fix.  I'm not sure why nothing was breaking as they would have been null.

johnduffell:
this PR was just to produce suggestions for another PR which has now been completed, so closing

PR #2805, opened 2025-05-01
make consent setter understand and map event bus style queue messages

johnduffell:
This is a PR on top of related PR https://github.com/guardian/support-service-lambdas/pull/2801

To support soft opt in + explicit consent at the same time, we need to trigger soft opt in from the event bus rather than from the SF query.

We have noticed that the product name is represented as a more capitalised version in the event bus https://github.com/guardian/support-frontend/blob/beef97734c1ca1549bc1cb5f1ea5b4501d24fc46/support-modules/acquisition-events/src/main/scala/com/gu/support/acquisitions/models/AcquisitionDataRow.scala#L99-L117
and also some of the necessary information is in a separate field `printOptions` to distinguish GW and newspaper.
https://github.com/guardian/support-frontend/blob/beef97734c1ca1549bc1cb5f1ea5b4501d24fc46/support-modules/acquisition-events/src/main/scala/com/gu/support/acquisitions/models/AcquisitionDataRow.scala#L147-L193

We can't change the names in the mappings, as cancellation and switching are going to continue to be handled via the SF query.

After some back and forth, this PR makes the existing consent setter queue contain the extra information if it comes from the event bus, and have a mapping from old to new format in the file.

### DIFF
--- a/handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/models/ConsentsMapping.scala
+++ b/handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/models/ConsentsMapping.scala
@@ -19,7 +19,6 @@ object ConsentsMapping {
       case "PRINT_SUBSCRIPTION" if !printOptions.contains("GUARDIAN_WEEKLY") => "newspaper"
       case "PRINT_SUBSCRIPTION" if printOptions.contains("GUARDIAN_WEEKLY") => "Guardian Weekly"
       case "GUARDIAN_AD_LITE" => "Guardian Ad-Lite"
-      case s"Newspaper - $_" => "newspaper"
       case other => other
     }
   }
@@ -46,6 +45,26 @@ object ConsentsMapping {
       supporterNewsletter,
     ),
     "newspaper" -> Set(
+      yourSupportOnboarding,
+      subscriberPreview,
+      supporterNewsletter,
+    ),
+    "Newspaper - Home Delivery" -> Set(
+      yourSupportOnboarding,
+      subscriberPreview,
+      supporterNewsletter,
+    ),
+    "Newspaper - Voucher Book" -> Set(
+      yourSupportOnboarding,
+      subscriberPreview,
+      supporterNewsletter,
+    ),
+    "Newspaper - Digital Voucher" -> Set(
+      yourSupportOnboarding,
+      subscriberPreview,
+      supporterNewsletter,
+    ),
+    "Newspaper - National Delivery" -> Set(
       yourSupportOnboarding,
       subscriberPreview,
       supporterNewsletter,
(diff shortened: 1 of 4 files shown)


handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/models/ConsentsMapping.scala
```diff
@@ -47,6 +64,11 @@ object ConsentsMapping {
       subscriberPreview,
       supporterNewsletter,
     ),
+    "Newspaper - National Delivery" -> Set(
+      yourSupportOnboarding,
+      subscriberPreview,
+      supporterNewsletter,
+    ),
```
johnduffell, line 71: we took the moment to move the newspapers all together and in the same order

handlers/soft-opt-in-consent-setter/src/test/scala/com.gu.soft_opt_in_consent_setter/ConsentsMappingTest.scala
```diff
@@ -0,0 +1,27 @@
+package com.gu.soft_opt_in_consent_setter
+
+import com.gu.soft_opt_in_consent_setter.models.ConsentsMapping._
+import org.scalatest.flatspec.AnyFlatSpec
+import org.scalatest.matchers.should.Matchers
+
+class ConsentsMappingTest extends AnyFlatSpec with Matchers {
+
+  "productMappings" should "map ophan style name to the zuora style name" in {
```
rupertbates, line 9: Minor nitpick but I don't really recognise the second style as Zuora style. In Zuora newspaper is four separate products and GW is two. Really it is just a specific style for this lambda.

rupertbates (approved):
One question and a minor nitpick but this looks good to me 👍

handlers/soft-opt-in-consent-setter/src/test/scala/com.gu.soft_opt_in_consent_setter/ConsentsMappingTest.scala
johnduffell: yeah it's a good point that in the general case, the output of this function isn't related to zuora names.

However in this exact subtest, it really must be the Zuora product name, as for cancellations it comes from Zuora to SF and then into the lambda.  And Salesforce uses the producttype_c value from the zuora catalog for this field.

I can see that the test names can be improved though so I will give them a refresh.

johnduffell:
@michaelbjacobson now that this is approved, feel free to merge it into your branch so that it's ready to go when everything else is shipped

handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/models/ConsentsMapping.scala
```diff
@@ -6,6 +6,23 @@ object ConsentsMapping {
...
+  This function is needed because when events come from the acquisition event bus, they have an ophan style product name.
+   */
+  def productMappings(productName: String, printOptions: Option[String]): String = {
+    // the values on the left come from https://github.com/guardian/support-frontend/blob/beef97734c1ca1549bc1cb5f1ea5b4501d24fc46/support-modules/acquisition-events/src/main/scala/com/gu/support/acquisitions/models/AcquisitionDataRow.scala#L97
+    productName match {
+      case "RECURRING_CONTRIBUTION" => "Contribution"
+      case "SUPPORTER_PLUS" => "Supporter Plus"
+      case "TIER_THREE" => "Tier Three"
+      case "DIGITAL_SUBSCRIPTION" => "Digital Pack"
+      case "PRINT_SUBSCRIPTION" if !printOptions.contains("GUARDIAN_WEEKLY") => "newspaper"
```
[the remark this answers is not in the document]
rupertbates: I think it's fine

PR #2809, opened 2025-05-06
fix: update bucket lifecycle rules to specify a rule for each prefix

AndreaDiotallevi:
## What does this change?

The observer-data-export S3 bucket has currently a retention period of 28 days. However, it also contains Unifida's public key which should not expire (`s3://observer-data-export-prod/Public_keys/unifida_public_rsa_key.pem`).

Therefore, this PR updates the S3 bucket lifecycle rules to specify which folders should have a retention of 28 days, rather than including the whole bucket.

### DIFF
--- a/cdk/lib/observer-data-export.ts
+++ b/cdk/lib/observer-data-export.ts
@@ -34,7 +34,10 @@ export class ObserverDataExport extends GuStack {
 
 		const sharedBucket = new Bucket(this, 'Bucket', {
 			bucketName: `${app}-${this.stage.toLowerCase()}`,
-			lifecycleRules: [{ expiration: Duration.days(28) }],
+			lifecycleRules: [
+				{ expiration: Duration.days(28) },
+				{ prefix: 'Public_keys/' }, // No expiration
+			],
 		});
 
 		const md5FingerprintsBucket = new Bucket(this, 'Md5FingerprintsBucket', {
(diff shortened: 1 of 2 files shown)


cdk/lib/observer-data-export.ts
```diff
@@ -34,7 +34,10 @@ export class ObserverDataExport extends GuStack {
 
 		const sharedBucket = new Bucket(this, 'Bucket', {
 			bucketName: `${app}-${this.stage.toLowerCase()}`,
-			lifecycleRules: [{ expiration: Duration.days(28) }],
+			lifecycleRules: [
+				{ expiration: Duration.days(28) },
+				{ prefix: 'Public_keys/' }, // No expiration
```
david-pepper, line 39: I see a comment saying 'no expiration' but I also see a key value pair saying ``` expiration: Duration.days(28) ``` which appears to be contradictory

david-pepper:
In order to review this I would need a little more detail in the Description, including some context, and the issue that is being addressed

cdk/lib/observer-data-export.ts
AndreaDiotallevi: Thank you for the comment.

I have improved the code.

AndreaDiotallevi:
> In order to review this I would need a little more detail in the Description, including some context, and the issue that is being addressed

Good point. I have improved the description now.

david-pepper (approved):

PR #2810, opened 2025-05-06
Fixed price change + Added price validation when switching to recurring contribution

andresilva-guardian:
## What does this change?

This PR adds price validation to the product move API when switching to a recurring contribution. Previously, we were accepting any price input from the frontend when processing this switch. Now, we:

1. Added a new validateOldMembershipPrice method that checks if the provided price matches the expected price for the given currency and billing period
2. Fixed an issue where we were using the previousAmount instead of the new price when creating rate plans (link [here](https://trello.com/c/4dqTTGKs/288-severity-2-medium-bug-report-74-system-impacted-manage-my-account-mma-product-switch-payment-amount-not-reflected-in-mma)).
3. The validation ensures prices match our expected values for different currency and billing period combinations

The main benefits are:

- Prevents potential security issues by validating prices server-side
- Ensures price consistency between frontend and backend
- Makes the backend the source of truth for valid prices

Some context:

- **Somewhere in 2023 there was a Membership Price Rise.**
- **Users that still have the "Supporter Membership" subscription are allowed (we offer them) to move their product to a Recurring Subscription keeping the current price (old price, before 2023).**

Compare old and new prices.

Determine if a user is eligible for a "save" journey (e.g., offering discounts or alternative plans to retain the user).

Display the correct pricing based on the user's subscription status (e.g., whether they are on the old or new pricing).

## How to test

1. Try switching a subscription to a recurring contribution with valid prices:

- GBP: £5/month or £49/year
- USD: $6.99/month or $69/year
- EUR: €4.99/month or €49/year
- AUD: $10/month or $100/year
- CAD: $6.99/month or $69/year

2. Try switching with an invalid price (should fail with appropriate error message):

- Example: GBP £8/month (not in valid price list)
- Example: USD $10/month (not in valid price list)

3. Verify error messages are clear and helpful

1. No failed switches due to price mismatches between frontend and backend
2. Reduction in support tickets related to price inconsistencies
3. Successful blocking of any attempts to use invalid prices
4. No performance degradation in the switching process

- **Limited flexibility**: By hardcoding accepted prices, we'll need to update both frontend and backend if prices change. This is an acceptable tradeoff given the temporary nature of this feature.
- **Error handling**: We need to ensure errors are appropriately surfaced to users with clear messages about what went wrong.
- **Future maintenance**: The hardcoded price table will need to be kept in sync with any price changes, but as known, this is a transitional solution until all users are migrated.
- **Performance impact**: The validation adds minimal overhead and doesn't require any API calls, so performance impact should be negligible.

## Images

N/A - This is a backend change with no UI modifications.

## Accessibility

N/A - This change doesn't affect the UI or accessibility features.

### DIFF
--- a/handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/move/switchtype/ToRecurringContribution.scala
+++ b/handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/move/switchtype/ToRecurringContribution.scala
@@ -1,10 +1,10 @@
 package com.gu.productmove.endpoint.move.switchtype
 
+import com.gu.i18n.Currency
 import com.gu.newproduct.api.productcatalog.ZuoraIds.zuoraIdsForStage
 import com.gu.newproduct.api.productcatalog.{Annual, BillingPeriod, Monthly}
 import com.gu.productmove.*
 import com.gu.productmove.GuStageLive.Stage
-import com.gu.productmove.endpoint.available.Currency
 import com.gu.productmove.endpoint.move.ProductMoveEndpoint.SwitchType
 import com.gu.productmove.endpoint.move.ProductMoveEndpointTypes.*
 import com.gu.productmove.endpoint.move.stringFor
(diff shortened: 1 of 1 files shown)


ecguardian (approved):
👍

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/move/switchtype/ToRecurringContribution.scala
```diff
@@ -71,9 +72,12 @@ class ToRecurringContributionImpl(
         .fromOption(activeRatePlanCharge.billingPeriod)
         .orElseFail(new Throwable(s"billingPeriod is null for rate plan charge $activeRatePlanCharge"))
 
+      // Make sure that price is valid and acceptable
+      _ <- validateOldMembershipPrice(currency, billingPeriod, price)
+
       updateRequestBody <- getRatePlans(
         billingPeriod,
-        previousAmount,
+        price,
```
johnduffell, line 80: this is the important change as it makes sure that we are using the list price that the reader saw on screen, rather than the (higher) price that they were paying for the membership

johnduffell (approved):
looks good, one thing that surprised me was that we are switching people who are paying the current membership amount, to the old one.  I'm not sure how much we would lose for that, but it doesn't seem right.
Also, I believe this endpoint only handles supporter membership to contribution, but worth confirming, as partner/patron membership would be more expensive.

PR #2814, opened 2025-05-06
add extra call to delete the zuora accounts

johnduffell:
The cleaner was originally added to save space in salesforce.
However now there seems to be a call to delete the whole account.

This PR changes it so it also deletes the account after cancelling it.

It might be possible to remove the existing code so it just goes straight for the delete.  But I haven't had a chance to test that option.
Also, I haven't done a mop up because @rupertbates has recently done that manually.

Tested locally and it deleted an account that was identified.

### DIFF
--- a/handlers/dev-env-cleaner/src/main/scala/com/gu/cleaner/Handler.scala
+++ b/handlers/dev-env-cleaner/src/main/scala/com/gu/cleaner/Handler.scala
@@ -174,20 +174,18 @@ class Steps(log: String => Unit) {
         .map { case id :: termEndDate :: Nil => cancelSub.run(id, dateToCancel(LocalDate.parse(termEndDate), today())) }
         .toList
         .sequence
-      accountsToCancel = queryResults(accounts_to_cancel).toList
-      _ <- accountsToCancel
+      _ <- queryResults(accounts_to_cancel)
         .map { case id :: creditBalance :: Nil =>
-          creditBalance.toDouble match {
-            case 0 => cancelAccount.run(id)
-            case _ =>
-              removeAccountCrm.run(id) // can't cancel an account with a credit balance, so just remove the CRMId
-          }
+          for {
+            _ <- creditBalance.toDouble match {
+              case 0 => cancelAccount.run(id)
+              case _ =>
+                removeAccountCrm.run(id) // can't cancel an account with a credit balance, so just remove the CRMId
+            }
+            _ <- deleteAccount.run(id)
+          } yield ()
         }
-        .sequence
-      _ <- accountsToCancel
-        .map { case id :: _ =>
-          deleteAccount.run(id)
-        }
+        .toList
         .sequence
     } yield ()
     zRes.toDisjunction.leftMap(failure =>


handlers/dev-env-cleaner/src/main/scala/com/gu/cleaner/Handler.scala
```diff
@@ -279,3 +287,11 @@ case class RemoveAccountCrm(log: String => Unit, restRequestMaker: RestRequestMa
   }
 
 }
+
+case class DeleteAccount(log: String => Unit, restRequestMaker: RestRequestMaker.Requests) {
+  def run(accountId: String): ClientFailableOp[Unit] = {
+    log(s"Delete account: $accountId")
+    restRequestMaker.delete[Unit](s"accounts/$accountId")
```
rupertbates, line 294: Does the /v1/ part of the url get prepended?

handlers/dev-env-cleaner/src/main/scala/com/gu/cleaner/Handler.scala
```diff
@@ -171,15 +174,20 @@ class Steps(log: String => Unit) {
...
             case 0 => cancelAccount.run(id)
             case _ =>
               removeAccountCrm.run(id) // can't cancel an account with a credit balance, so just remove the CRMId
           }
         }
-        .toList
+        .sequence
+      _ <- accountsToCancel
+        .map { case id :: _ =>
+          deleteAccount.run(id)
```
rupertbates, line 189: Why run this here rather than after the cancellation on line 181?

johnduffell: actually I'm going to have a play with removing the sub/acc cancellation, as I've also realised it will get rid of one of the AQUA queries which is a big bonus for keeping them in sync
Then it will be a moot point about combining them.

johnduffell:
> To me it seems simpler to carry out both at the same time but I don't feel particularly strongly

OK I have made the change as suggested and will ship like that.  Then if I have time I will look at removing the calls that may not be needed

handlers/dev-env-cleaner/src/main/scala/com/gu/cleaner/Handler.scala
```diff
@@ -173,11 +176,14 @@ class Steps(log: String => Unit) {
...
-            case _ =>
-              removeAccountCrm.run(id) // can't cancel an account with a credit balance, so just remove the CRMId
-          }
+          for {
+            _ <- creditBalance.toDouble match {
+              case 0 => cancelAccount.run(id)
+              case _ =>
+                removeAccountCrm.run(id) // can't cancel an account with a credit balance, so just remove the CRMId
+            }
+            _ <- deleteAccount.run(id)
```
rupertbates, line 185: Will this work for an account with a credit balance?

johnduffell: sorry missed your message but yes it seems to.  The ones with a credit balance were just building up, getting returned by the query every time, and getting skipped, so this is doubly good that they can be deleted.

PR #2815, opened 2025-05-06
don't use salesforce query to set SOI consents on acquisition

johnduffell:
As part of switching over to triggering acquisition consents via the queue for zuora products https://github.com/guardian/support-service-lambdas/pull/2776 , we need to disable the old salesforce based consent mechanism.

This PR is a simple way by making acquisiton consent setting a no-op.

In the longer term we can make it not appear in the SF query, but this means we can quickly reenable it if there is an issue by reverting this PR.

This PR should be merged once we are happy that the queue based consent is coming through alright.  Until that happens, there will be a race condition where both lambdas will write the consents, but this lambda will usually come in last, making it safer.


michaelbjacobson (approved):
🚀

michaelbjacobson (approved):

MaelGNM (approved):

PR #2820, opened 2025-05-09
ignore removed plans when calculating a switch

johnduffell:
We noticed that when switching, the code was removing all previously removed rateplans again.  This was causing a lot of unnecessary amendments, and it was causing errors from zuora where we exceeded 10 amendments in one call (when testing - users wouldn't hit this in PROD)

This PR makes it ignore all previously removed plans.


andresilva-guardian (approved):
Nice work @johnduffell , thanks 🙏

PR #2822, opened 2025-05-09
Rename Paper+ rate plans to PaperPlus to avoid problems with url encoding

rupertbates:
## What does this change?
We are going to start selling paper plus digital plans again, however the current naming of the rate plans in the product catalog (Weekend+, Everyday+ etc.) causes a problem when trying to purchase one of those plans through the checkout because + is a special character in URLs. 

This means that we either have to URL encode the rate plan name in the checkout ratePlan query parameter, eg. https://support.thegulocal.com/uk/checkout?product=NationalDelivery&ratePlan=Weekend%2b or rename the plans in the catalog and we have decided that renaming is a better approach.

# Notes
- Various lambdas use the schema defined in the product-catalog module when loading the product catalog

### DIFF
--- a/modules/product-catalog/src/productCatalogSchema.ts
+++ b/modules/product-catalog/src/productCatalogSchema.ts
@@ -644,73 +644,83 @@ export const productCatalogSchema = z.object({
 					.enum(typeObject.NewspaperVoucher.billingPeriods)
 					.optional(),
 			}),
-			EverydayPlus: z.object({
-				id: z.string(),
-				pricing: z.object({ GBP: z.number() }),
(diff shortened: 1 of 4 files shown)


johnduffell (approved):
great idea.  but anything that relies on this will fail in between the time that things are deployed and the next execution of zuora catalog fetcher.
If it's a concern, since we don't have versioning of the catalog, it might be worth duplicating the plans with both names temporarily, then waiting for everything to use the new name, and then deleting the old names.

rupertbates:
Replaced this PR with #2842

PR #2829, opened 2025-05-13
send date only to salesforce tracking queue

johnduffell:
Last july some product switches were moved over to the product-switch-api which was a clean TS rewrite.

Unfortunately we missed that the events weren't ending up in salesforce, culminating in a fix in late december to the permissions of the switch lambda so it could write to SQS. https://github.com/guardian/support-service-lambdas/pull/2602

However after that, we again failed to check it was working, and since then it's been constantly failing with 
```
Error '.requestedDate(2025-05-13T13:00:50.183Z is not a valid ISO-8601 format, illegal local date at index 10)' when decoding JSON to SalesforceRecordInput with body: 
```
https://865473395570-3z35t5l4.eu-west-1.console.aws.amazon.com/cloudwatch/home?region=eu-west-1#logsV2:log-groups/log-group/$252Faws$252Flambda$252Fproduct-switch-salesforce-tracking-PROD/log-events$3Fstart$3D-1800000

This is because the JS date format is a complete offsetdatetime, however the lambda that reads is requires it to only be a localdate, with no time or timezone appended.

This PR fixes that and also adds a test.

Tested in CODE which produced this data, which was then ingested successfully by lambda 
product-switch-salesforce-tracking-CODE and passed to salesforce DEV
```
{
    "subscriptionName": "A-S00985326",
    "previousAmount": 8,
    "newAmount": 12,
    "previousProductName": "Contributor",
    "previousRatePlanName": "Monthly Contribution",
    "newRatePlanName": "Supporter Plus",
    "requestedDate": "2025-05-14",
    "effectiveDate": "2025-05-14",
    "paidAmount": 4
}
```

TODO in a later PR, make sure there's an alarm on that queue product-switch-salesforce-tracking-PROD

### DIFF
--- a/handlers/product-switch-api/src/salesforceTracking.ts
+++ b/handlers/product-switch-api/src/salesforceTracking.ts
@@ -10,18 +10,18 @@ export type SalesforceTrackingInput = {
 	previousProductName: string;
 	previousRatePlanName: string;
 	newRatePlanName: string;
-	requestedDate: Date;
-	effectiveDate: Date;
+	requestedDate: string;
+	effectiveDate: string;
 	paidAmount: number;
 	csrUserId?: string;
 	caseId?: string;
 };
-export const sendSalesforceTracking = async (
-	paidAmount: number,
+
+export function createSQSMessageBody(
 	switchInformation: SwitchInformation,
-) => {
-	const queueName = `product-switch-salesforce-tracking-${switchInformation.stage}`;
-	const client = new SQSClient(awsConfig);
+	paidAmount: number,
+	now: Date,
+) {
 	const {
 		subscriptionNumber,
 		previousProductName,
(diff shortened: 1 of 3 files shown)


handlers/product-switch-api/test/salesforceTracking.test.ts
```diff
@@ -0,0 +1,63 @@
...
+				price: 0,
+				productRatePlanId: '',
+				subscriptionChargeId: '',
+				contributionChargeId: '',
+			},
+			contribution: {
+				productRatePlanId: '',
+				chargeId: '',
+			},
+		},
```
johnduffell, line 43: I think most of this info isn't needed but I wasn't sure a qucik way to avoid it

modules/prettyPrint.ts
```diff
@@ -1,3 +1,3 @@
-export const prettyPrint = (object: object) => {
+export const prettyPrint = (object: any) => {
```
johnduffell, line 1: the types weren't lining up between parse/stringify and prettyPrint, so this corrects that assuming there wasn't some other intention behind it

handlers/product-switch-api/src/salesforceTracking.ts
```diff
@@ -10,18 +10,18 @@ export type SalesforceTrackingInput = {
 	previousProductName: string;
 	previousRatePlanName: string;
 	newRatePlanName: string;
-	requestedDate: Date;
-	effectiveDate: Date;
```
johnduffell, line 14: Date serialises as a full OffsetDateTime, so we have to manually serialise it to a string

handlers/product-switch-api/src/salesforceTracking.ts
```diff
@@ -10,18 +10,18 @@ export type SalesforceTrackingInput = {
...
+	requestedDate: string;
+	effectiveDate: string;
 	paidAmount: number;
 	csrUserId?: string;
 	caseId?: string;
 };
-export const sendSalesforceTracking = async (
-	paidAmount: number,
+
+export function createSQSMessageBody(
```
johnduffell, line 20: extracted into a separate function for testability

handlers/product-switch-api/src/salesforceTracking.ts
```diff
@@ -37,20 +37,35 @@ export const sendSalesforceTracking = async (
...
+		paidAmount,
+		new Date(),
+	);
+
+	const client = new SQSClient(awsConfig);
+	const queueName = `product-switch-salesforce-tracking-${switchInformation.stage}`;
 	console.log(
 		`Sending Salesforce tracking message ${prettyPrint(
-			salesforceTrackingInput,
+			JSON.parse(messageBody),
```
johnduffell, line 63: we have to parse it again so we can pretty print it.  Alternatively I could have returned the object, and stringified it later, even though that would test slightly less of the codepath?

handlers/product-switch-api/src/salesforceTracking.ts
```diff
@@ -10,18 +10,18 @@ export type SalesforceTrackingInput = {
 	previousProductName: string;
 	previousRatePlanName: string;
 	newRatePlanName: string;
-	requestedDate: Date;
-	effectiveDate: Date;
+	requestedDate: string;
+	effectiveDate: string;
 	paidAmount: number;
 	csrUserId?: string;
 	caseId?: string;
```
johnduffell, line 17: https://github.com/guardian/support-service-lambdas/blob/e384bf4c25484cff5079f52fdd5fa1983a98f122/handlers/product-move-api/src/main/scala/com/gu/productmove/salesforce/Salesforce.scala#L12-L24

rBangay (approved):
👍

PR #2830, opened 2025-05-14
add alarms for the product move switch tracking lambda

johnduffell:
We have not had tracking of product switches in salesforce since June last year.

We didn't notice because there were no alarms, and JSON serialisation errors didn't make the lambda Error

Traffic to the lambda was very low until december, (fixed in https://github.com/guardian/support-service-lambdas/pull/2602 ) and after that point it was throwing JSON deserialisation errors on every request. (Fixed in https://github.com/guardian/support-service-lambdas/pull/2829 )

This PR
1. adds both a low traffic alarm and an errors alarm
2. changes the lambda so it will attempt to execute all records in the queue, and Error if any one of them doesn't work.  This does mean that if there are multiple records, some records might be retried even if they have worked, however I can see that the BatchSize is set to 1 anyway.

Thresholds are arbitrary and can be tweaked later if needed.

Tested in AWS console and it shows that the alarms will be created:

image

and

image

Tested in CODE
one valid JSON and one invalid worked correctly:
image

### DIFF
--- a/handlers/product-move-api/src/main/scala/com/gu/productmove/salesforce/SalesforceHandler.scala
+++ b/handlers/product-move-api/src/main/scala/com/gu/productmove/salesforce/SalesforceHandler.scala
@@ -1,12 +1,10 @@
 package com.gu.productmove.salesforce
 
 import com.amazonaws.services.lambda.runtime.events.SQSEvent
-import com.amazonaws.services.lambda.runtime.events.SQSEvent.SQSMessage
-import com.gu.productmove.GuStageLive.Stage
 import com.amazonaws.services.lambda.runtime.{Context, RequestHandler}
-import com.gu.productmove.salesforce.CreateRecord.CreateRecordRequest
+import com.gu.productmove.GuStageLive.Stage
 import com.gu.productmove.salesforce.Salesforce.{SalesforceRecordInput, createSfRecord}
-import com.gu.productmove.{AwsCredentialsLive, AwsS3, AwsS3Live, GuStageLive, SecretsLive, SttpClientLive}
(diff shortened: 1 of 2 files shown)


handlers/product-move-api/src/main/scala/com/gu/productmove/salesforce/SalesforceHandler.scala
```diff
@@ -17,18 +15,27 @@ class SalesforceHandler extends RequestHandler[SQSEvent, Unit] {
...
-      }
+    val results = records.map { record =>
+      val result = for {
+        salesforceRecordInput <- record.getBody
+          .fromJson[SalesforceRecordInput]
+          .left
+          .map(msg => new RuntimeException("failed to deserialise input: " + msg))
+        _ = context.getLogger.log(s"Processing salesforceRecordInput with body: ${record.getBody}")
+        _ <- runZio(salesforceRecordInput, context)
+      } yield ()
```
johnduffell, line 26: made this into a for comprehension so we can gather up any errors in one place

handlers/product-move-api/src/main/scala/com/gu/productmove/salesforce/SalesforceHandler.scala
```diff
@@ -17,18 +15,27 @@ class SalesforceHandler extends RequestHandler[SQSEvent, Unit] {
...
+      result.left.foreach(ex => context.getLogger.log(ex.toString))
+      result
     }
+    val numErrors =
+      results.collect { case Left(ex) =>
+        ex
+      }.length
+    if (numErrors > 0)
+      throw new RuntimeException("Lambda failed due to previously logged errors")
```
johnduffell, line 35: now we count up the errors and fail the lambda if there are any

PR #2831, opened 2025-05-14
Script to generate a Zod schema for the product catalog

rupertbates:
## What does this change?
The product-catalog module defines a set of types which allow us to work with Guardian products in an easy type safe way. It also contains a Zod schema which parses the product-catalog API response into these types.

Previously the types themselves were derived from an object in the typeObject.ts file, and this was generated from the Zuora catalog by a script (generateTypes). The Zod schema was created manually and had to be kept in sync with the generated types or there would be compile, test or runtime failures.

This PR changes this setup so that we now generate the Zod schema from the Zuora catalog, and then infer the types from that. This means that there is now no code that needs to manually be kept in sync and should make it easier to publish changes to the product catalog.

## Changes
- Create a new script `generateSchema` which generates `productCatalogSchema.ts` from the Zuora catalog. This replaces the old `generateTypes` script
- Add a new file `productBillingPeriods.ts`. The previous mechanism created the `ProductBillingPeriod<ProductKey>` type which is used in the update-supporter-plus-amounts lambda but which does not exist in the product catalog API or schema so it has been extracted to this file. This is also generated by the `generatedSchema` script.
- Delete the old type object as it is no longer needed.


modules/product-catalog/src/generateProductBillingPeriods.ts
```diff
@@ -0,0 +1,77 @@
...
+import type { StripeProductKey } from './stripeProducts';
+import { stripeProducts } from './stripeProducts';
+
+const getBillingPeriodsForProduct = (
+	productRatePlans: ZuoraProductRatePlan[],
+) =>
+	distinct(
+		productRatePlans
+			.flatMap((productRatePlan) =>
+				productRatePlan.productRatePlanCharges.flatMap(
```
tjmw, line 20: Is the flatMap needed here? There won't be nested arrays at this level will there?

modules/product-catalog/src/generateProductBillingPeriods.ts
```diff
@@ -0,0 +1,77 @@
...
+				(billingPeriod) =>
+					billingPeriod !== null && billingPeriod != 'Specific_Weeks',
+			) as string[],
+	);
+
+const getBillingPeriodsForStripeProduct = (
+	productKey: StripeProductKey,
+	product: Product<StripeProductKey>,
+): Record<string, string[]> => {
+	const billingPeriods = Object.entries(product.ratePlans).map(
```
tjmw, line 34: If we only care about the values could we use `Object.values` here?

modules/product-catalog/src/generateSchemaCommand.ts
```diff
@@ -0,0 +1,16 @@
...
+const writeSchemaToFile = async () => {
+	const prodCatalog = await getZuoraCatalogFromS3('PROD');
+	const generatedSchema = generateSchema(prodCatalog);
+	fs.writeFileSync('./src/productCatalogSchema.ts', generatedSchema);
+	const productBillingPeriods = generateProductBillingPeriods(prodCatalog);
+	fs.writeFileSync('./src/productBillingPeriods.ts', productBillingPeriods);
+};
+
+void (async function () {
+	await writeSchemaToFile();
```
tjmw, line 15: No top level await here?

modules/product-catalog/src/productBillingPeriods.ts
```diff
@@ -0,0 +1,39 @@
+import type { ProductKey } from '@modules/product-catalog/productCatalog';
+
```
tjmw, line 2: When generating this would it be worth adding a comment about this file being auto generated so that developers aren't tempted to modify it directly in future? (I know it's hinted at below)

modules/product-catalog/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "product-catalog",
   "version": "1.0.0",
   "scripts": {
-    "generateTypes": "ts-node src/generateTypeObjectCommand.ts && prettier --write src/typeObject.ts",
+    "generateSchema": "ts-node src/generateSchemaCommand.ts && prettier --write src/productCatalogSchema.ts && prettier --write src/productBillingPeriods.ts",
```
tjmw, line 5: Would it be worth attempting to compile the generated file with `tsc` at the end of all this to get instant feedback if the generated code is somehow invalid?

modules/product-catalog/src/generateSchemaCommand.ts
rupertbates: I don't think we need it?

tjmw (approved):
Great!

PR #2832, opened 2025-05-15
increase no-switches alarm to 4 hours

johnduffell:
Yesterday I did a PR to add a no-switches alarm. https://github.com/guardian/support-service-lambdas/pull/2830

I gave it a low threshold so we could see whether it worked overnight.

Overnight it went off several times correctly.

This PR changes the threshold to something more reasonable now that the alarm is working well.

We can still tweak it further if needed.


MaelGNM (approved):
Nice!

PR #2835, opened 2025-05-16
add new fields to product switch confirmation email for next payment

johnduffell:
We noticed that the switch confirmation email mentions the date and amount of the immediate payment, and the ongoing amount, but it doesn't mention the exact date of the next payment.

This PR adds two new fields to the email
next_payment_amount
date_of_next_payment

After this is shipped, the email template can be updated in braze to use it.

Before: 
image

After (tested in CODE)
```
{
    "To": {
        "Address": "aaa.aaa+aaaa@guardian.co.uk",
        "ContactAttributes": {
            "SubscriberAttributes": {
                "first_name": "aaaa",
                "last_name": "aaaaaa",
                "currency": "£",
                "price": "60.00",
                "first_payment_amount": "10.00",
                "date_of_first_payment": "19 May 2025",
                "next_payment_amount": "60.00",
                "date_of_next_payment": "19 May 2026",
                "payment_frequency": "Annually",
                "subscription_id": "A-S001234"
            }
        }
    },
    "DataExtensionName": "SV_RCtoSP_Switch",
    "IdentityUserId": "1234"
}
```

### DIFF
--- a/handlers/product-switch-api/src/productSwitchEmail.ts
+++ b/handlers/product-switch-api/src/productSwitchEmail.ts
@@ -6,29 +6,20 @@ import { getCurrencyGlyph } from '@modules/internationalisation/currency';
 import dayjs from 'dayjs';
 import type { SwitchInformation } from './switchInformation';
 
-export const buildEmailMessage = ({
-	dateOfFirstPayment,
-	emailAddress,
-	firstName,
-	lastName,
-	currency,
-	productPrice,
-	firstPaymentAmount,
-	billingPeriod,
-	subscriptionNumber,
-	identityId,
-}: {
-	dateOfFirstPayment: dayjs.Dayjs;
-	emailAddress: string;
-	firstName: string;
-	lastName: string;
-	currency: Currency;
-	productPrice: number;
-	firstPaymentAmount: number;
-	billingPeriod: BillingPeriod;
-	subscriptionNumber: string;
-	identityId: string;
-}) => {
+type Payment = { date: dayjs.Dayjs; amount: number };
+
+export const buildEmailMessage = (
+	payments: { first: Payment; next: Payment },
+	emailAddress: string,
+	firstName: string,
+	lastName: string,
+	currency: Currency,
+	productPrice: number,
+	billingPeriod: BillingPeriod,
+	subscriptionNumber: string,
+	identityId: string,
+) => {
+	const { first, next } = payments;
 	return {
 		To: {
 			Address: emailAddress,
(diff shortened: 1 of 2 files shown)


handlers/product-switch-api/src/productSwitchEmail.ts
```diff
@@ -60,18 +53,35 @@ export const sendThankYouEmail = async (
...
-		identityId: identityId,
-	});
+	const billingPeriodMonths: number =
+		switchInformation.subscription.billingPeriod == 'Month'
+			? 1
+			: switchInformation.subscription.billingPeriod == 'Annual'
+				? 12
+				: switchInformation.subscription.billingPeriod == 'Quarter'
+					? 3
+					: 1;
```
johnduffell, line 63: the type can only be Month/Annual/Quarter, but typescript didn't work out that it was impossible if I put undefined as the last option.
I wasn't sure if a switch statement was a normal alternative to scala's pattern matching, I couldn't immediately get it to work as it seems to be a statement rather than an expression.

handlers/product-switch-api/src/productSwitchEmail.ts
```diff
@@ -60,18 +53,35 @@ export const sendThankYouEmail = async (
...
-		billingPeriod: billingPeriod,
-		subscriptionNumber: subscriptionNumber,
-		identityId: identityId,
-	});
+	const billingPeriodMonths: number =
+		switchInformation.subscription.billingPeriod == 'Month'
+			? 1
+			: switchInformation.subscription.billingPeriod == 'Annual'
+				? 12
+				: switchInformation.subscription.billingPeriod == 'Quarter'
```
johnduffell, line 61: I couldn't work out what this means
`61:7  error  Unnecessary conditional, both sides of the expression are literal values  @typescript-eslint/no-unnecessary-condition`
https://github.com/guardian/support-service-lambdas/actions/runs/15070189029/job/42364415016?pr=2835#step:8:19
I tried putting round brackets in, but it kept removing them as soon as I clicked away

rBangay: It probably doesn't like the final or case as it can never reach it ... I wonder whether a lookup object might be better/easier to read here, something like:

```
const billingPeriodMonthsLookup:Record<'Month' | 'Annual' | 'Quarter', number> = {
    'Month': 1,
    "Annual": 12,
    "Quarter": 3
}
```
(you might have to let prettier do it's thing on that!)

Then you can reference it like:

```
const billingPeriodMonths = billingPeriodMonthsLookup[switchInformation.subscription.billingPeriod]
```

johnduffell: nice one thanks Richard!  works a treat and so much neater. 👏

PR #2839, opened 2025-05-19
fix product-switch-api tests and make them run in the CI

johnduffell:
Richard and I noticed that we are not running the product-switch-api tests in the CI, and several tests have been broken recently.

This PR makes them run in CI again, and also fixes the failing tests.

I checked the other package.jsons and it looks like the test command is there everywhere else.  Looking back in the history for this file, it was never there.  It must have been a typo or oversight.

### DIFF
--- a/handlers/product-switch-api/src/discounts.ts
+++ b/handlers/product-switch-api/src/discounts.ts
@@ -1,11 +1,7 @@
 import type { BillingPeriod } from '@modules/billingPeriod';
+import type { Lazy } from '@modules/lazy';
 import type { Stage } from '@modules/stage';
-import {
-	billingPreviewToSimpleInvoiceItems,
-	getBillingPreview,
-} from '@modules/zuora/billingPreview';
-import type { ZuoraClient } from '@modules/zuora/zuoraClient';
-import dayjs from 'dayjs';
+import type { SimpleInvoiceItem } from '@modules/zuora/billingPreview';
 
 export type Discount = {
 	// Zuora discount API reference:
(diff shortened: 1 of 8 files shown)


handlers/product-switch-api/src/discounts.ts
```diff
@@ -56,15 +51,8 @@ export const getDiscount = async (
 	const subIsActive = subscriptionStatus === 'Active';
 
 	if (subIsActive) {
-		const getBillingPreviewResponse = await getBillingPreview(
-			zuoraClient,
-			dayjs().add(13, 'months'),
-			accountNumber,
-		);
+		const nextInvoiceItems = await lazyBillingPreview.get();
```
johnduffell, line 54: using a Lazy means we can pass in dummy values in the tests.  As it was, the CI tests were trying to call out to sandbox zuora.

handlers/product-switch-api/test/contributionToSupporterPlus.test.ts
```diff
@@ -113,20 +111,19 @@ test("owner check doesn't allow incorrect owner", async () => {
...
 			productCatalog,
-			'12345', // incorrect identity id
-			zuoraClient,
+			'12345',
+			new Lazy(() => Promise.resolve([]), 'test'),
 			today,
-		);
-	expect(switchInformation).toThrow(ValidationError);
+		),
+	).rejects.toThrow(ValidationError);
```
johnduffell, line 126: before the promise was rejecting, but we were just checking that there was no exception thrown when creating the promise.
The key missing bit was the `.rejects`
I've tested it as either passing or failing, and it works both ways.

handlers/product-switch-api/test/supporterRatePlanItem.test.ts
```diff
@@ -71,9 +71,5 @@ test('supporterRatePlanItemFromSwitchInformation works with a contribution eleme
 		productRatePlanName: 'Supporter Plus V2 - Monthly',
 		termEndDate: zuoraDateFormat(dayjs().add(1, 'year')),
 		contractEffectiveDate: zuoraDateFormat(dayjs()),
-		contributionAmount: {
-			amount: 10,
-			currency: 'GBP',
-		},
```
johnduffell, line 77: this was removed recently as we don't need the amount of supporter pluses as they are not really treated as a contribution any more.

rBangay (approved):
nice, thanks for getting the tests to run 👍

PR #2840, opened 2025-05-19
Prevent 404 errors from deleted accounts causing DLQ entries in soft-opt-in-consent-setter

andresilva-guardian:
## What does this change?
In response to the ticket ["Ignore 404 from IDAPI in soft opt in consent setter IAP"](https://trello.com/c/9uMcrsiz/797-ignore-404-from-idapi-in-soft-opt-in-consent-setter-iap), this PR modifies the soft-opt-in-consent-setter-iap lambda to handle 404 errors gracefully when attempting to remove consents for deleted accounts. Previously, when processing cancellations for accounts that had been deleted from Identity, the 404 responses from IDAPI would cause the lambda to fail and put the message on the dead letter queue (DLQ) for retries.

The change:

1. Extends the SoftOptInError class to capture HTTP status codes
2. Updates the IdentityConnector to include status codes in error responses
3. Modifies the cancellation flow in HandlerIAP to specifically handle 404 responses by logging a warning and treating them as successful operations rather than failures

This prevents unnecessary DLQ entries and retries for situations that can never succeed (deleted accounts).

## How to test

1. Trigger the lambda with an SQS message simulating a cancellation event for a non-existent identity ID
2. Verify that the lambda logs a warning about the 404 but completes successfully without sending the message to DLQ
3. Confirm that legitimate errors (non-404) are still properly propagated and will cause the lambda to fail

- Reduction in the number of error messages in the DLQ related to 404 responses from IDAPI
- Decrease in CloudWatch error metrics for the lambda
- No change in behavior for valid identity accounts (consents should still be updated properly)

- Low risk: This change only affects handling of accounts that no longer exist in Identity
- There's a small risk that we might mask legitimate 404 errors that should be retried, but in our analysis all 404s were due to deleted accounts which can never be recovered
- We're still propagating all other errors, so any other issues will continue to trigger retries via DLQ

## Images

No visual changes as this is a backend lambda modification.

## Accessibility

Not applicable as this is a backend change with no user interface components.


handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/HandlerIAP.scala
```diff
@@ -242,7 +242,16 @@ object HandlerIAP extends LazyLogging with RequestHandler[SQSEvent, Unit] {
...
+            sendConsentsReq(identityId, consentsBody) match {
+              case Left(notFoundErr) if notFoundErr.statusCode.contains(404) =>
+                logger.warn(s"(cancellation) Consents request for $identityId failed with 404 Not Found")
+                Right(())
+              case Left(otherError) =>
+                // Propagate all other errors
+                Left(otherError)
+              case Right(_) =>
+                Right(())
+            }
```
johnduffell, line 254: This whole function is a bit lumpy even before your change, perhaps this would be more concise end result?
```suggestion
    def sendCancellationConsents(identityId: String, consents: Set[String]): Either[SoftOptInError, Unit] = {
      val maybeError: Option[SoftOptInError] =
        for {
          con <- Some(consents).filter(_.nonEmpty)
          consentsBody = consentsCalculator.buildConsentsBody(con.map(_ -> false).toMap)
          _ = logger.info(
            s"(cancellation) Sending consents request for identityId $identityId with payload: $consentsBody",
          )
          err <- sendConsentsReq(identityId, consentsBody).swap.toOption
          if !err.statusCode.contains(404)
        } yield err
      maybeError.toLeft(())
    }
```

johnduffell (approved):
good, I've made a suggestion to make it more scalaish but overall I think it will work well

I assume this is intentional to only apply for cancellations, rather than all, which makes sense because you are unlikely to delete your account very quickly after switching or acquisition.

MaelGNM (approved):
I'm happy with this solution. I think it's fine to only do it for cancellations as that should be the only possible case to see this happen, so other cases would alarm.

handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/HandlerIAP.scala
andresilva-guardian: Thank you @johnduffell for taking the time to refactor this function 🙏

PR #2841, opened 2025-05-20
Sort the keys of generated objects to make them easier to diff

rupertbates:
## What does this change?
We generate two objects from the Zuora product catalog
- A Zod product catalog schema object
- A type object from which we derive a type which tells us which products support which billing periods

There is an issue with this which is that the keys of the generated objects are ordered in a 'last updated' fashion by Zuora and this makes diffs when we update these objects hard to understand. This PR sorts the keys of the generated objects using ESLint so that diffs remain stable and changes are easy to understand.

## Changes
- Add `eslint-plugin-sort-keys-fix` plugin to handle sorting the keys of the generated objects
- Add an `.eslintrc.json` file to ensure that the new rule only runs on the generated files
- Regenerate productCatalogSchema.ts & productBillingPeriods.ts with the new sorted keys
- Add new json test fixtures because regenerating the schema has introduced new rate plans (National Delivery paper+)


modules/product-catalog/package.json
```diff
@@ -3,20 +3,21 @@
   "version": "1.0.0",
   "scripts": {
     "generateFiles": "ts-node src/generateSchemaCommand.ts",
-    "validateSchema": "prettier --write src/productCatalogSchema.ts && eslint src/productCatalogSchema.ts",
-    "validateBillingPeriods": "prettier --write src/productBillingPeriods.ts",
+    "validateSchema": "prettier --write src/productCatalogSchema.ts && for i in {1..3}; do eslint --fix src/productCatalogSchema.ts; done",
```
rupertbates, line 6: Unfortunately `eslint --fix` needs multiple passes - one for each level of nesting in the target object. We handle this by running it in a loop.

tjmw: That's annoying! It'd be good to have a comment on this for future developers. We can't add comments directly to the package.json, maybe a named script in here which we call out to? What do you reckon?

modules/product-catalog/test/__snapshots__/generateProductCatalog.test.ts.snap
```diff
@@ -701,6 +701,39 @@ exports[`code Generated product catalog matches snapshot 1`] = `
           "GBP": 83.99,
         },
       },
+      "Everyday+": {
```
tjmw, line 704: Just checking that it's expected that the new rateplans have been added here too?

tjmw (approved):
👍

modules/product-catalog/test/__snapshots__/generateProductCatalog.test.ts.snap
rupertbates: Yes because they're in the API now.

PR #2842, opened 2025-05-20
Change paper+ rate plan names to paperPlus

rupertbates:
## What does this change?

We are going to start selling paper plus digital plans again, however the current naming of the rate plans in the product catalog (Weekend+, Everyday+ etc.) causes a problem when trying to purchase one of those plans through the checkout because + is a special character in URLs.

This means that we either have to URL encode the rate plan name in the checkout ratePlan query parameter, eg. https://support.thegulocal.com/uk/checkout?product=NationalDelivery&ratePlan=Weekend%2b or rename the plans in the catalog and we have decided that renaming is a better approach.

I have left the old rate plan names in the schema and marked them as `.optional()`. The new plans are also optional so that there will be no schema validation issues caused by race conditions. Once this PR has deployed successfully and the product catalog API contains the new plans, the old ones can be removed as well as the optional attribute.

### DIFF
--- a/modules/product-catalog/src/productCatalogSchema.ts
+++ b/modules/product-catalog/src/productCatalogSchema.ts
@@ -449,37 +449,72 @@ export const productCatalogSchema = z.object({
 				id: z.string(),
 				pricing: z.object({ GBP: z.number() }),
 			}),
-			'Everyday+': z.object({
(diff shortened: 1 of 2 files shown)


tjmw (approved):
Looks good, and the deployment plan sounds sensible 👍

PR #2845, opened 2025-05-22
Increase lambda memory to recommended amounts

rupertbates:
## What does this change?
AWS Trusted Advisor has a report on lambdas which are under-provisioned for memory size: https://us-east-1.console.aws.amazon.com/trustedadvisor/home?region=eu-west-1#/category/performance

This PR makes the updates recommended in that report


paulbrown1982 (approved):
👍

maloky (approved):
there was still one in the AWS report `support-bandit-get-bandit-tests-*` but i could not find it anywhere in the codebase. Does it lives in another repo?

rupertbates:
> there was still one in the AWS report `support-bandit-get-bandit-tests-*` but i could not find it anywhere in the codebase. Does it lives in another repo?

Yes it's definitely another repo, I need to track down which one.

PR #2849, opened 2025-05-27
return generic error string on 500 response for the switch-api endpoint

rBangay:
## What does this change?

Stop returning the entire error object back to the client if the switch api throws a 500 error, instead return a generic error message string. This is to avoid leaking any sensitive information.


johnduffell (approved):
looks good, this was added in this pr https://github.com/guardian/support-service-lambdas/pull/2201/files#diff-a7cd5eae7d430d539df85a12ae98b4d905d39b919787f1b15e1063ddf52430acR54

We also checked the other lambdas in this repo and they don't have the issue, so it was likely a one off oversight.

PR #2851, opened 2025-05-28
Remove `similar-guardian-products` soft opt-in for Live App IAPs

michaelbjacobson:
### What does this PR do?
This PR updates the soft opt-in `ConsentsMapping` to prevent Live App IAPs being opted-in to `similar-guardian-products`. 

### Why are we doing this?
We are removing `similar_guardian_products` but keeping `your_support_onboarding`, so we only need to update the consents mapping here [support-service-lambdas/handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/models/ConsentsMapping.scala at main · guardian/support-service-lambdas](https://github.com/guardian/support-service-lambdas/blob/main/handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/models/ConsentsMapping.scala)

https://docs.google.com/document/d/19RvjreFZezqe_sgXp496y4Ha5W58TAtiggQVxKPSOEk/edit?tab=t.0Connect your Google account

We need to make a change to how we capture Soft Opt-in on the Feast app. We currently capture Similar Guardian Products at the point of sign-in. However we also capture it at the point of IAP purchase. We need to remove this consent capture at the point of IAP purchase and only rely on the sign-in journey to capture SGP consent.

We will need to sync up with the Feast team so they can adapt the copy on the Thank You screen post purchase to reflect the consent mechanism change.

### DIFF
--- a/handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/Handler.scala
+++ b/handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/Handler.scala
@@ -19,6 +19,7 @@ object Handler extends LazyLogging {
   val readyToProcessAcquisitionStatus = "Ready to process acquisition"
   val readyToProcessCancellationStatus = "Ready to process cancellation"
   val readyProcessSwitchStatus = "Ready to process switch"
+  val similarGuardianProducts = "similar_guardian_products"
 
   def main(args: Array[String]): Unit = {
     handleRequest()
@@ -119,6 +120,7 @@ object Handler extends LazyLogging {
       toRemove = oldProductSoftOptIns.diff(currentProductSoftOptIns).map(ConsentsObject(_, false))
       toAdd = newProductSoftOptIns
         .filter(option => !oldProductSoftOptIns.contains(option) && !allOtherProductSoftOptIns.contains(option))
+        .filterNot(_ == similarGuardianProducts)
         .map(ConsentsObject(_, true))
 
       consentsBody = (toRemove ++ toAdd).asJson.toString()
--- a/handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/Handler.scala
+++ b/handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/Handler.scala
@@ -1,15 +1,7 @@
 package com.gu.soft_opt_in_consent_setter
 
-import com.gu.soft_opt_in_consent_setter.models.{
-  ConsentsMapping,
-  EnhancedSub,
-  SFAssociatedSubResponse,
-  SFSubRecord,
-  SFSubRecordUpdate,
-  SFSubRecordUpdateRequest,
-  SoftOptInConfig,
-  SoftOptInError,
-}
+import com.gu.soft_opt_in_consent_setter.models.ConsentsMapping.similarGuardianProducts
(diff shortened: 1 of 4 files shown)


handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/IAPMessageProcessor.scala
```diff
@@ -132,6 +132,7 @@ object IAPMessageProcessor extends StrictLogging {
       toRemove = oldProductSoftOptIns.diff(currentProductSoftOptIns).map(ConsentsObject(_, false))
       toAdd = newProductSoftOptIns
         .filter(option => !oldProductSoftOptIns.contains(option) && !allOtherProductSoftOptIns.contains(option))
+        .filterNot(_ == similarGuardianProducts)
```
MaelGNM, line 135: I'm not familiar with the switch side of things, why do you need to filter our similarGuardianProducts here?

andresilva-guardian (approved):
Nice work @michaelbjacobson , thanks 🙏

handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/IAPMessageProcessor.scala
MaelGNM: Ah I see, yes that's good, I think it's preferable not to set it.

handlers/soft-opt-in-consent-setter/src/test/scala/com/gu/soft_opt_in_consent_setter/ConsentsCalculatorTests.scala
```diff
@@ -213,24 +213,6 @@ class ConsentsCalculatorTests extends AnyFlatSpec with should.Matchers with Eith
...
-        |  {
-        |    "id" : "guardian_weekly_newsletter",
-        |    "consented" : false
-        |  },
-        |  {
-        |    "id" : "subscriber_preview",
-        |    "consented" : true
-        |  }
-        |]""".stripMargin)
-  }
```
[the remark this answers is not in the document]
michaelbjacobson: @johnduffell, @MaelGNM thanks for this. I think I got myself confused on which tests were failing and why while I was working on the logic. I've re-added this test and made the necessary changes elsewhere.

handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/Handler.scala
```diff
@@ -19,6 +19,7 @@ object Handler extends LazyLogging {
   val readyToProcessAcquisitionStatus = "Ready to process acquisition"
   val readyToProcessCancellationStatus = "Ready to process cancellation"
   val readyProcessSwitchStatus = "Ready to process switch"
+  val similarGuardianProducts = "similar_guardian_products"
```
MaelGNM, line 22: It might be better to import it rather than declaring it again. Same as in [IAPMessageProcessor](https://github.com/guardian/support-service-lambdas/blob/f18600acca1cfff3bf0c98b3d489d5ec237fd401/handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/IAPMessageProcessor.scala#L4)

MaelGNM: That's not a blocking thing, more nitpicking

MaelGNM (approved):
It looks good to me 👏

johnduffell (approved):
nice one thanks 👏

PR #2852, opened 2025-05-28
update to scala 3.7.0

johnduffell:
updated to 3.7.0 and recompiled with `-rewrite -source 3.7-migration` to squash all the warnings that can be auto fixed since 3.3
https://www.scala-lang.org/news/3.7.0/#:~:text=Scala%203.7%20provides%20an%20automated%20migration%20path%20for%20existing%20codebases%20through%20the%20compiler%20flags%20%2Drewrite%20%2Dsource%3A3.7%2Dmigration

https://github.com/guardian/recommendations/blob/main/scala.md#:~:text=latest%20for%20production%20code.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -25,7 +25,7 @@ val scala2Settings = Seq(
 )
 
 val scala3Settings = Seq(
-  scalaVersion := "3.3.4",
+  scalaVersion := "3.7.0",
   version := "0.0.1",
   organization := "com.gu",
   scalacOptions ++= Seq(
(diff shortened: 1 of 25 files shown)


build.sbt
```diff
@@ -25,7 +25,7 @@ val scala2Settings = Seq(
 )
 
 val scala3Settings = Seq(
-  scalaVersion := "3.3.4",
+  scalaVersion := "3.7.0",
```
johnduffell, line 28: this is the only actual change, the rest are automatic

rBangay (approved):
👍

PR #2854, opened 2025-05-28
First Draft of the mParticle API function

andresilva-guardian:
## Accessibility

### DIFF
--- a/handlers/mparticle-api/README.md
+++ b/handlers/mparticle-api/README.md
@@ -1,0 +1,32 @@
+# mParticle Api
+
+## Background
+This project provides a TypeScript AWS Lambda for integrating with the [mParticle Data Subject Request (DSR) API](https://docs.mparticle.com/developers/apis/dsr-api/v3/). It enables the submission and tracking of data subject requests (such as access or erasure requests) in compliance with privacy regulations.
+
+The lambda is designed to be robust, maintainable, and easy to contribute to, following the conventions used across the support-service-lambdas monorepo.
+
+## Features
+
+- Submit Data Subject Requests (DSR) to mParticle in OpenDSR format.
+- Query the status of existing DSRs.
+- (Planned) Receive and process status callbacks from mParticle.
+- Written in TypeScript for maintainability and ease of contribution.
+
+## Development
+
+This project uses TypeScript. See the [root README](../../README.md#getting-started---typescript) for more information on TypeScript conventions in this repo.
+
+To work with this project:
+
+- Ensure [pnpm](https://pnpm.io/) is installed.
+- Run `pnpm install` from the root of the repo to install dependencies.
+- Run the unit tests with `pnpm test` (all tests in the monorepo) or `pnpm --filter mparticle-api test` (just the tests in this project).
+
+## Testing
+
+Integration with the mParticle API can be tested by invoking the lambda with appropriate payloads. Ensure you have valid credentials and configuration for the mParticle API.
+
+## Future Improvements
+
+- Implement an endpoint to receive and process status callbacks from mParticle.
+- Integrate with downstream systems (e.g., BigQuery erasure app) as required by privacy workflows.
--- a/handlers/mparticle-api/interfaces/data-subject-request-state.ts
+++ b/handlers/mparticle-api/interfaces/data-subject-request-state.ts
@@ -1,0 +1,41 @@
+export enum DataSubjectRequestStatus {
+    Pending = 'pending',
+    InProgress = 'in-progress',
+    Completed = 'completed',
+    Cancelled = 'cancelled',
+}
+
+/**
+ * Get the status of an OpenDSR request
+ * https://docs.mparticle.com/developers/apis/dsr-api/v3/#example-response-body
+ */
+export interface DataSubjectRequestState {
+    /**
+     * The estimated time by which the request will be fulfilled, in UTC.
+     */
+    expectedCompletionTime: Date;
+
+    /**
+     * The controller-provided identifier of the request in a GUID v4 format.
+     */
+    subjectRequestId: string;
+
+    /**
+     * A unique ID representing the data controller. mParticles sets this to the workspace ID.
+     */
+    controllerId: string;
+
+    /**
+     * The status of the request.
+     */
+    requestStatus: DataSubjectRequestStatus;
+
+    /**
+     * For Access/Portability requests, a download link to the request results data.
+     * This field contains null unless the request is complete. After a request completes,
+     * the resultsUrl is valid for 7 days. After that time, attempting to access this URL
+     * results in a 410 Gone HTTP response. If no records can be found matching the identities
+     * in the request, a request returns a 404 error.
+     */
+    resultsUrl: string | null;
+}
--- a/handlers/mparticle-api/src/index.ts
+++ b/handlers/mparticle-api/src/index.ts
@@ -1,0 +1,192 @@
+import type {
+	APIGatewayProxyEvent,
+	APIGatewayProxyResult,
+	Handler,
+} from 'aws-lambda';
+import axios from 'axios';
+import type { AxiosInstance } from 'axios';
+import type { DataSubjectRequestForm } from '../interfaces/data-subject-request-form';
+import type { DataSubjectRequestState } from '../interfaces/data-subject-request-state';
+import { DataSubjectRequestStatus } from '../interfaces/data-subject-request-state';
+import type { DataSubjectRequestSubmission } from '../interfaces/data-subject-request-submission';
+
+// #region API Clients
+
+let _dataSubjectRequestApiClient: AxiosInstance | undefined = undefined;
+const getDataSubjectRequestApiClient = (): AxiosInstance => {
+	if (!_dataSubjectRequestApiClient) {
+		_dataSubjectRequestApiClient = axios.create({
+			baseURL: `https://opendsr.mparticle.com/v3`,
+			headers: {
+				'Content-Type': 'application/json',
+				/**
+				 * Authentication
+				 * The DSR API is secured via basic authentication. Credentials are issued at the level of an mParticle workspace.
+				 * You can obtain credentials for your workspace from the Workspace Settings screen. Note that this authentication
+				 * is for a single workspace and scopes the DSR to this workspace only.
+				 * https://docs.mparticle.com/developers/apis/dsr-api/v3/#authentication
+				 */
+				'Authorization': `Basic ${Buffer.from(`${process.env.apiKey}:${process.env.apiSecret}`).toString('base64')}`,
+			},
+		});
+	}
+	return _dataSubjectRequestApiClient;
+}
+
+// #endregion
+
+// #region Tools
+
+const parseDataSubjectRequestStatus = (status: 'pending' | 'in_progress' | 'completed' | 'cancelled'): DataSubjectRequestStatus => {
+	switch (status) {
+		case 'pending':
+			return DataSubjectRequestStatus.Pending;
+		case 'in_progress':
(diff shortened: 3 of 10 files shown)


handlers/mparticle-api/src/index.ts
```diff
@@ -0,0 +1,192 @@
...
+			headers: {
+				'Content-Type': 'application/json',
+				/**
+				 * Authentication
+				 * The DSR API is secured via basic authentication. Credentials are issued at the level of an mParticle workspace.
+				 * You can obtain credentials for your workspace from the Workspace Settings screen. Note that this authentication
+				 * is for a single workspace and scopes the DSR to this workspace only.
+				 * https://docs.mparticle.com/developers/apis/dsr-api/v3/#authentication
+				 */
+				'Authorization': `Basic ${Buffer.from(`${process.env.apiKey}:${process.env.apiSecret}`).toString('base64')}`,
```
johnduffell, line 29: although opinions may vary, usually we store secrets in AWS (SSM parameter store) because it makes it easier to run code locally e.g. integration tests in CODE env.  It does add a bit of latency but generally it's not an issue.
It looks like there's a client https://github.com/guardian/support-service-lambdas/blob/eaffa5cc5182485e11e2c91e1cb3bf8b141ac386/modules/aws/src/ssm.ts

I've also asked on chat as there seems to be quite a variety already, I think everyone's already done whatever came to mind.

handlers/mparticle-api/src/index.ts
```diff
@@ -0,0 +1,192 @@
...
+import type { AxiosInstance } from 'axios';
+import type { DataSubjectRequestForm } from '../interfaces/data-subject-request-form';
+import type { DataSubjectRequestState } from '../interfaces/data-subject-request-state';
+import { DataSubjectRequestStatus } from '../interfaces/data-subject-request-state';
+import type { DataSubjectRequestSubmission } from '../interfaces/data-subject-request-submission';
+
+// #region API Clients
+
+let _dataSubjectRequestApiClient: AxiosInstance | undefined = undefined;
+const getDataSubjectRequestApiClient = (): AxiosInstance => {
```
johnduffell, line 16: I assume this is to make it lazy initialised?  I wouldn't object to just creating it on cold start as we'll surely need it most of the time

handlers/mparticle-api/src/index.ts
```diff
@@ -0,0 +1,192 @@
...
+
+// #endregion
+
+// #region API Interface
+
+/**
+ * Submit a Data Subject Request (DSR)
+ * A request in the OpenDSR format communicates a Data Subject’s wish to access or erase their data.
+ * The OpenDSR Request takes a JSON request body and requires a Content-Type: application/json header.
+ * https://docs.mparticle.com/developers/apis/dsr-api/v3/#submit-a-data-subject-request-dsr
```
johnduffell, line 61: nice to put docs link in here 👏

handlers/mparticle-api/src/index.ts
```diff
@@ -0,0 +1,192 @@
...
+			skip_waiting_period: true,
+			subject_identities: form.subjectIdentities.map(identity => ({
+				identity_type: identity.identityType,
+				value: identity.value,
+				encoding: identity.encoding,
+			})),
+			api_version: "3.0",
+			"status_callback_urls": [
+				"https://exampleurl.com/opendsr/callbacks" // <- We should create an endpoint on this app to receive this status callback and propagate its state. In alternative we can call the BigQuery erasure app.
+			],
```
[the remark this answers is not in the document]
johnduffell: agree with AI - exampleurl.com could be hoovering up everyone's PII!

But to answer your question, I'm not sure, I'm not familiar with baton's patterns.  If we need to report back to baton when the request is complete then it's nicer to use a callback than to poll, assuming it's reliable and secure.

handlers/mparticle-api/src/index.ts
```diff
@@ -0,0 +1,192 @@
...
+			"group_id": "my-group", // <- Let's maybe use the User Unique Id to group all requests related to that user (max 150 requests per group)
```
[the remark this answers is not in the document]
andresilva-guardian: I agree Copilot, no worries, we will group it by user, I hope so. At least that's what I'm looking for.

handlers/mparticle-api/README.md
```diff
@@ -0,0 +1,32 @@
+# mParticle Api
```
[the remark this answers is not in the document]
andresilva-guardian: Don't be like that Copilot, specially because "mParticle" is a brand ;)

handlers/mparticle-api/src/index.ts
```diff
@@ -0,0 +1,192 @@
...
+import type { DataSubjectRequestForm } from '../interfaces/data-subject-request-form';
+import type { DataSubjectRequestState } from '../interfaces/data-subject-request-state';
+import { DataSubjectRequestStatus } from '../interfaces/data-subject-request-state';
+import type { DataSubjectRequestSubmission } from '../interfaces/data-subject-request-submission';
+
+// #region API Clients
+
+let _dataSubjectRequestApiClient: AxiosInstance | undefined = undefined;
+const getDataSubjectRequestApiClient = (): AxiosInstance => {
+	if (!_dataSubjectRequestApiClient) {
```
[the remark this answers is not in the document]
andresilva-guardian: I tend to do this for 2 main reasons:
- It improves memory usage of the lambda function, specially when the use case is not used a lot (low traffic)
- It improves warmup times when AWS has to spin-off new instances

handlers/mparticle-api/src/index.ts
```diff
@@ -0,0 +1,192 @@
...
+		});
+		return {
+			expectedCompletionTime: new Date(response.data.expected_completion_time),
+			receivedTime: new Date(response.data.received_time),
+			subjectRequestId: response.data.subject_request_id,
+			controllerId: response.data.controller_id,
+		};
+	} catch (err) {
+		console.error('Error submitting Data Subject Request:', err);
+		throw err;
```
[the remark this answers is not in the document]
andresilva-guardian: How should I setup an alarm for this @johnduffell ?

handlers/mparticle-api/interfaces/data-subject-request-state.ts
```diff
@@ -0,0 +1,41 @@
+export enum DataSubjectRequestStatus {
+    Pending = 'pending',
+    InProgress = 'in-progress',
```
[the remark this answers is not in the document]
andresilva-guardian: Don't agree, I'm trying to create an abstraction keeping our own convention.

ecguardian (approved):
👍

modules/routing/src/router.ts
```diff
@@ -14,26 +15,93 @@ type Route = {
...
+							};
+						}
+						throw error;
+					}
+
+					// Validate request body
+					try {
+						if (route.validation?.body) {
+							const parsedBody: unknown = eventWithParams.body ? JSON.parse(eventWithParams.body) : undefined;
+							route.validation.body.parse(parsedBody);
```
johnduffell, line 88: now that we have parsed it successfully shouldn't we do something with it? e.g. pass it to the route.handler?

andresilva-guardian: I thought about it, but I wasn't sure whether to extend the event specification or not. Do you have any suggestion @johnduffell ?

modules/routing/src/router.ts
```diff
@@ -14,26 +15,93 @@ type Route = {
...
+		body?: z.Schema;
+	};
 };
 
 export const NotFoundResponse = {
 	body: 'Not Found',
 	statusCode: 404,
 };
 
+function matchPath(routePath: string, eventPath: string): { matched: boolean; params: Record<string, string> } {
```
johnduffell, line 29: would it make more sense to return a sum type (ie make it optional).
Then we can make it clear from the types that matched=false and params have a value is not possible
```suggestion
function matchPath(routePath: string, eventPath: string): params: Record<string, string> | undefined {
```

modules/routing/src/router.ts
```diff
@@ -14,26 +15,93 @@ type Route = {
...
-			);
-			if (route) {
-				return await route.handler(event);
+			for (const route of this.routes) {
+				const { matched, params } = matchPath(route.path, event.path);
+				if (matched && route.httpMethod === event.httpMethod.toUpperCase()) {
+					// Attach pathParameters to event
+					const eventWithParams = {
+						...event,
+						pathParameters: { ...(event.pathParameters ?? {}), ...params },
```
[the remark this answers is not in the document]
andresilva-guardian: "event.pathParameters" does not include the parameters. Not sure why. I believe that you mention the correct reason, the url path params are not mapped "automatically" it seems.

modules/routing/src/router.ts
```diff
@@ -14,26 +15,93 @@ type Route = {
...
+	for (let i = 0; i < routeParts.length; i++) {
+		const routePart = routeParts[i];
+		const eventPart = eventParts[i];
+		if (routePart?.startsWith('{') && routePart.endsWith('}')) {
+			const paramName = routePart.slice(1, -1);
+			params[paramName] = eventPart as string;
+		} else if (routePart !== eventPart) {
+			return { matched: false, params: {} };
+		}
+	}
```
johnduffell, line 47: my scala brain would find something like this more logical, due to being able to understand it step by step rather than the whole loop at once (suggestion assumes the change suggested above is done)

1. pair up the expected/actual parts
2. partition them into matchers and literals
3. if any literal doesn't match, we don't match the route
4. return the matchers as the parameters

```suggestion
  const routeEventPairs = zip(routeParts, eventParts)
  const [matchers, literals] = mapPartition(routeEventPairs, ([routePart, eventPart]) => {
    const maybeParamName = routePart.match(/^\{.*\}$/)[0];
    maybeParamName ? [maybeParamName, eventPart] : undefined;
  });
  if (literals.some(([routePart, eventPart]) => routePart !== eventPart) {
    return undefined; // or if we prefer the existing types: { matched: false, params: {} }
  }
  return Object.fromEntries(matchers); // or if we prefer the existing types: { matched: true, params: Object.fromEntries(matchers) }
```

with something like this in arrayFunctions.ts
```
const zip = <T, U>(ts: T[], us: U[]) => ts.map((value, index) => [value, us[index]])

// like map, but if it returns undefined, it puts it in the second list
const mapPartition = <T, U>(array: T[], f: T => U | undefined) => 
  array.reduce<[T[], U[]]>(
		(acc, val) => {
      const res = fn(val);
			if (res) acc[0].push(res);
			else acc[1].push(val);
			return acc;
		},
		[[], []],
	);
```

I would normally use `collect` (rather than making `mapPartition`) in scala, but I don't think typescript has anything similar

modules/routing/src/router.ts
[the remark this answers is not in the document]
andresilva-guardian: New pull request created just for the changes on router module -> https://github.com/guardian/support-service-lambdas/pull/2886

modules/routing/src/router.ts
andresilva-guardian: New pull request created just for the changes on router module -> https://github.com/guardian/support-service-lambdas/pull/2886

modules/routing/src/router.ts
andresilva-guardian: New pull request created just for the changes on router module -> https://github.com/guardian/support-service-lambdas/pull/2886

johnduffell:
the new lambda code needed to be copied to S3 before the CDK could be deployed.  I did that and now it's all passed https://riffraff.gutools.co.uk/deployment/view/feffc250-fd05-49a7-a7aa-0a338e18aa59

PR #2826, opened 2025-05-12
generate invoice async when backdated supporter plus cancellation is requested

johnduffell:
This needed a few bits of refactoring so this is branched off these PRs which should be approved/merged first:

https://github.com/guardian/support-service-lambdas/pull/2852 - MERGED
https://github.com/guardian/support-service-lambdas/pull/2853
https://github.com/guardian/support-service-lambdas/pull/2855
https://github.com/guardian/support-service-lambdas/pull/2856

We are getting a lot of 500 errors from supporter plus cancellation endpoint.

When you cancel within 14 days, the cancellation is backdated and you get a refund.
When the cancellation is done, we call it with runBilling = true, in order to generate the relevant (negative) invoice.
We then put a record on the queue to actually do the refund.

Unfortunately zuora seems to go really slow (like a minute) when you do that and it generates an invoice, thus API gateway gives up and the reader gets a bad experience (although the subscription is correctly dealt with as the lambda continues in the background)

This PR makes it just do the cancellation synchronously, and moves the invoice generation to happen after the queue.

Tested in CODE and working ok
https://apisandbox.zuora.com/apps/CustomerAccount.do?method=view&id=8ad083f0970b057201971b8cb9ce67c5
image

https://865473395570-3z35t5l4.eu-west-1.console.aws.amazon.com/cloudwatch/home?region=eu-west-1#logsV2:log-groups/log-group/$252Faws$252Flambda$252Fproduct-switch-refund-CODE/log-events/2025$252F05$252F29$252F$255B$2524LATEST$255D9753a4643bde4160b3adc447f0ae552b

PR #2856, opened 2025-05-29
separate zuora query code from invoice grouping plus tidyup

johnduffell:
This is yet more pre-refactoring for https://github.com/guardian/support-service-lambdas/pull/2826

The main change is to split out the code that runs queries against zuora from the code that groups and zips the results.  The processing code has been simplified at the same time.  This will make it easier to identify and return the invoice id of the negative invoice in the next PR.

There are also a few changes to logging, imports etc.  See inline comments.


handlers/product-move-api/src/main/scala/com/gu/productmove/refund/GetRefundInvoiceDetails.scala
```diff
@@ -0,0 +1,107 @@
+package com.gu.productmove.refund
```
johnduffell, line 1: this file is the business logic parts from the old handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/GetRefundInvoiceDetails.scala
(The query client itself has been moved to InvoiceItemQuery.scala)

handlers/product-move-api/src/main/scala/com/gu/productmove/refund/GetRefundInvoiceDetails.scala
```diff
@@ -0,0 +1,107 @@
...
+    } yield ZIO.succeed(InvoiceGroups(latestId, latestItems, previousItems))
+    invoiceItemGroups.getOrElse {
+      ZIO.fail(
+        new Throwable(
+          s"There was only one invoice item in response $items this is an error " +
+            s"as we need the cancellation invoice items to carry out a refund",
+        ),
+      )
+    }
+  }
```
johnduffell, line 46: the logic has changed to group them once and then return the values in one go, rather than regrouping them repeatedly as per the old logic

handlers/product-move-api/src/main/scala/com/gu/productmove/refund/RefundHandler.scala
```diff
@@ -1,29 +1,48 @@
...
+    // try to avoid the logs getting split up too much
+    System.setOut(printStream)
+    System.setErr(printStream)
+
     val records: List[SQSEvent.SQSMessage] = input.getRecords.asScala.toList
 
     records.foreach { record =>
-      val maybeRefundInput = record.getBody.fromJson[RefundInput]
+      val body = record.getBody
+      context.getLogger.log("Processing SQS record:\n" + body)
```
johnduffell, line 39: it's handy to have the incoming body logged

andresilva-guardian (approved):
Nice work @johnduffell it seems like a very solid architectural improvement 💪

PR #2857, opened 2025-05-29
filter out unrelated invoices from billing preview

johnduffell:
Someone had a monthly contribution and a voucher
MC was billed on the 6th and voucher on the 18th
They requested a cancellation for the contribution on the 28th

the discount api checked if they could have a pause, which includes a billing preview and an amemdment preview.

The billing preview returns all invoices for the accunt, and the amendment preview is applied as of the next invoice date, even if it's for the wrong subsubscription.

In this case it was the wrong subscription, and the amendment preview returned no records, which then caused a deserialisation error.

This PR changes it so that discount api and product switch api both filter out invoices that are unrelated to the current subscription.


modules/zuora/src/billingPreview.ts
```diff
@@ -22,6 +22,12 @@ export const getBillingPreview = async (
 	return zuoraClient.post(path, body, billingPreviewSchema);
 };
 
+export const itemsForSubscription =
+	(subscriptionNumber: string) => (billingPreview: BillingPreview) =>
+		billingPreview.invoiceItems.filter(
+			(item) => item.subscriptionNumber === subscriptionNumber,
+		);
```
johnduffell, line 29: this is the important function

PR #2859, opened 2025-06-02
Fix zod version in package.json

david-pepper:
## What does this change?
I noticed the zod version is not standard for specifying a package version, so updated to most recent version.


david-pepper:
I think I understand what's happening here. There is a configuration that uses the 'catalog:' identifier to point to the zod version:

Screenshot 2025-06-02 at 12 56 58

so will abandon these changes as they are not necessary

PR #2864, opened 2025-06-02
Negative invoices get records

david-pepper:

### DIFF
--- a/handlers/negative-invoices-processor/package.json
+++ b/handlers/negative-invoices-processor/package.json
@@ -21,6 +21,6 @@
 		"aws-sdk": "^2.1692.0",
 		"dayjs": "^1.11.13",
 		"google-auth-library": "^9.15.0",
-		"zod": "^3.23.8"
+		"zod": "catalog:"
 	}
 }
--- a/handlers/negative-invoices-processor/package.json
+++ b/handlers/negative-invoices-processor/package.json
@@ -1,0 +1,26 @@
+{
+	"name": "negative-invoices-processor",
+	"description": "A state machine that identifies and balances negative invoices.",
+	"scripts": {
+		"test": "jest --group=-integration",
+		"it-test": "jest --group=integration",
+		"build": "esbuild --bundle --platform=node --target=node20 --outdir=target src/handlers/*.ts",
+		"lint": "eslint src/**/*.ts test/**/*.ts",
+		"package": "pnpm type-check && pnpm lint && pnpm check-formatting && pnpm test && pnpm build && cd target && zip -qr negative-invoices-processor.zip ./*.js",
+		"type-check": "tsc --noEmit",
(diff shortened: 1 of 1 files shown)


handlers/negative-invoices-processor/src/handlers/getInvoices.ts
```diff
@@ -0,0 +1,59 @@
...
+        inv.id,
+        STRING_AGG(distinct inv.account_id, ',') as account_id,
+    FROM 
+        datatech-fivetran.zuora.invoice inv
+    WHERE 
+        inv.amount < 0 AND inv.balance != 0
+    GROUP BY 
+        1
+    ORDER BY 
+        invoice_date
```
graham228221, line 58:
```suggestion
    SELECT
        inv.id,
        inv.account_id
    FROM 
        datatech-fivetran.zuora.invoice inv
    WHERE 
        inv.amount < 0 AND inv.balance != 0
    ORDER BY 
        invoice_date
```

With this query you don't need `GROUP BY`, as your only getting a single rows from a single table (and you're not aggregating anything).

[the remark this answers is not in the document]
AndreaDiotallevi: Yes this makes sense. By filtering only active subs, we make sure there is no overlap with the other system.
Any invoice related to a non-active sub will be picked up by the other system.

handlers/negative-invoices-processor/src/handlers/getInvoices.ts
```diff
@@ -47,12 +47,17 @@ const query = (): string =>
 	`
     SELECT
```
graham228221, line 48: You'll want to bring the `GROUP BY` back if you're joining a table like subscriptions, sorry 😁  you might return multiple subscriptions.

Joining subscriptions from invoices I think would best to do via the invoice items, not the account. You might pick up subscriptions not related to the invoice we're interested in.

Something like this should work:

```
SELECT
        inv.id,
        STRING_AGG(distinct inv.account_id, ',') AS account_id,
        MAX(inv.balance) AS invoice_balance
    FROM 
        datatech-fivetran.zuora.invoice inv
    INNER JOIN 
        datatech-fivetran.zuora.invoice_item item 
        ON item.invoice_id = inv.id
    INNER JOIN
        datatech-fivetran.zuora.subscription sub
        ON item.subscription_id = sub.id AND sub.is_latest_version
    WHERE 
        inv.amount < 0
        AND inv.balance != 0
        AND sub.status = 'Active'
    GROUP BY 
        inv.id
```

david-pepper: righto, I've updated the query

graham228221 (approved):
Happy with the query 👍  state machine looks good to me in AWS, but no problem if you want to get a more qualified approval for the lambda

PR #2865, opened 2025-06-03
make sure Lazy only initialises once even when mapped

johnduffell:
discount api alarm was going off sometimes due to it taking nearly 30 seconds to execute, due to a zuora slowdown. (billing preview taking 10-18 seconds to complete)

When investigating, I noticed that the billing preview call was being made twice, unnecessarily.

I found that due to a bug in the lazy class, it was initialising it twice.

This PR fixes lazy to only initialise once, even when it's mapped.


MaelGNM (approved):
Good spot!

PR #2874, opened 2025-06-10
new client to read config from SSM in a standard way

johnduffell:
In the guardian for app config, the "paved path" is to use SSM and `/stage/stack/app/*` as the keys. The relevant permissions are provided by CDK for free : https://github.com/guardian/cdk/blob/main/src/constructs/iam/policies/parameter-store-read.ts

In scala, we have simple-configuration-ssm that handles reading them recursively https://github.com/guardian/simple-configuration/blob/main/ssm/src/main/scala/com/gu/conf/SSMConfigurationLocation.scala

However in our TS lambdas we seem to use a mixture of SSM, secrets manager and env variables.

This PR adds a library to use SSM with a zod schema according to the above "paved path".
It also updates the alarms handler to use it.

The idea is that all lambdas will use this so that we are consistent by default.

Tested in CODE ok.


handlers/alarms-handler/src/alarmMappings.ts
```diff
@@ -145,33 +143,18 @@ const buildAppToTeamMappings = (
...
-
-		return ['SRE'];
-	};
-
-	const getTeamWebhookUrl = (team: Team): string => {
-		return getIfDefined<string>(
-			process.env[`${team}_WEBHOOK`],
-			`${team}_WEBHOOK environment variable not set`,
-		);
-	};
```
johnduffell, line 172: the webhook urls for each team come from the config rather than the mappings, so it's been moved accordingly

handlers/alarms-handler/src/cloudwatch.ts
```diff
@@ -1,65 +1,74 @@
+import type { MetricAlarm, Tag } from '@aws-sdk/client-cloudwatch';
```
johnduffell, line 1: in this file, the essential change is that it takes in the aws config as an object and builds a set of cloudwatch clients, rather than building them internally from env vars.  So a bit more "functional" rather than "side effectful"

handlers/alarms-handler/test/index.test.ts
```diff
@@ -1,16 +1,17 @@
...
-		VALUE_WEBHOOK: 'value-webhook-url',
-		GROWTH_WEBHOOK: 'growth-webhook-url',
-		PP_WEBHOOK: 'pp-webhook-url',
-		SRE_WEBHOOK: 'sre-webhook-url',
+	const dummyWebhookUrls: WebhookUrls = {
+		VALUE: 'value-webhook-url',
+		GROWTH: 'growth-webhook-url',
+		SRE: 'sre-webhook-url',
+		PORTFOLIO: 'portfolio-webhook-url',
+		PLATFORM: 'platform-webhook-url',
```
johnduffell, line 14: now we pass these in as parameters instead of putting them in the env.

modules/aws/package.json
```diff
@@ -9,9 +9,11 @@
 	"dependencies": {
 		"@aws-sdk/client-cloudwatch": "3.777.0",
 		"@aws-sdk/client-s3": "3.777.0",
+		"@aws-sdk/client-ssm": "^3.777.0",
```
johnduffell, line 12: use SSM v3 rather than the existing v2

modules/aws/src/appConfig.ts
```diff
@@ -0,0 +1,120 @@
...
+/**
+ * App config uses the guardian-standard SSM keys to load config.  The GU CDK lambda
+ * provides permissions by default, meaning that you get the policies for free.
+ *
+ * @param stage
+ * @param stack
+ * @param app
+ * @param schema this schema will validate that all your config is present
+ */
+export const loadConfig = async <O>(
```
johnduffell, line 14: in other places e.g. zuoraClient we have a complex type signature here, but it seems like this also works.  (It's very hard to understand the types as it comes out as something very complicated.)

modules/aws/src/ssm.ts
```diff
@@ -1,22 +1,19 @@
...
-	const ssm = new AWS.SSM({ region: 'eu-west-1' });
-	return ssm
-		.getParameter({
-			Name: name,
-			WithDecryption: true,
-		})
-		.promise()
-		.then((result) => {
-			const value = result.Parameter?.Value;
+export const getSSMParam = async (name: string): Promise<string> => {
```
johnduffell, line 4: this code is not directly related to this change, but I converted to use AWS v3 instead of v2 for completeness.

modules/aws/runManual/runLoadConfig.ts
```diff
@@ -0,0 +1,9 @@
+import { z } from 'zod';
+import { loadConfig } from '@modules/aws/appConfig';
+
+// by making it "any", it will not strip out any unknown keys, making it useful to have a look at some config for an app
+export const schema = z.any();
+
+export type SchemaType = z.infer<typeof schema>;
+
+loadConfig('CODE', 'support', 'alarms-handler', schema).then(console.log);
```
johnduffell, line 9: this will display the config on the screen.  Config for individual keys can be easily edited in SSM via the console.

It might be nice to have a little script to download/upload the config in bulk, just to make it easier to manage?  But not in this PR.  This would primarily be useful when setting up a lambda in the first place, but I think it's a valid use case to want to run lambdas without credentials, e.g. for testing.  This would mean that the loadConfig method might need to fall back to local file (in /etc/gu) if there are no credentials.

modules/aws/package.json
```diff
@@ -9,9 +9,11 @@
...
 		"@aws-sdk/credential-provider-node": "3.782.0",
 		"@types/aws-sdk": "^2.7.4",
 		"aws-sdk": "^2.1692.0",
-		"aws-sdk-client-mock": "4.1.0"
+		"aws-sdk-client-mock": "4.1.0",
+		"zod": "catalog:"
```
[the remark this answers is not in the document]
tomrf1: what is `catalog:`?

rupertbates (approved):
Looks good to me, I'm not too familiar with the alarms-handler so I was slightly struggling to work that out at the same time, but I think it all makes sense 👍

modules/aws/src/appConfig.ts
```diff
@@ -0,0 +1,120 @@
...
+				JSON.stringify(configTree, null, 2) +
+				'\nDue to error: ' +
+				parseResult.error,
+		); // ZodError instance
+	} else {
+		return parseResult.data;
+	}
+}
+
+export type ConfigFlat = Record<string, string>[];
```
tomrf1, line 81: we could make it clearer that this is the model we get from SSM

tomrf1: also, it's nice to define the types before their use

modules/aws/src/appConfig.ts
```diff
@@ -0,0 +1,120 @@
...
+		([name, value]) =>
+			({
+				path: name
+					.replace(configRoot, '')
+					.replace(/^\//, '')
+					.split('/')
+					.filter((k) => k.length > 0)
+					.reverse(),
+				value,
+			}) as PathWithValue,
```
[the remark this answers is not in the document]
johnduffell: done

PR #2878, opened 2025-06-12
chore: fix conflicts

AndreaDiotallevi:
## What does this change?

Fix merge conflicts.


andresilva-guardian (approved):
Thanks! I've tested it by running `pnpm install` and it seems to be working as expected. Thank you for fixing it 💪

PR #2883, opened 2025-06-16
extract order trigger dates duplication

johnduffell:
I noticed that we keep repeating the same trigger block, and the formatting rules make it take a huge amount of space.

This PR extracts the trigger dates into a shared function.

This makes it much more readable of the logic we are actually interested in


MaelGNM (approved):
Nice little refactor!

PR #2887, opened 2025-06-17
New Salesforce Event Bus and Contact Update Event rule

graham228221:
## What does this change?
This adds a new Event Bus for Salesforce platform event handling, utilising Eventbridge and [Salesforce Event Relay](https://developer.salesforce.com/docs/platform/event-relay/overview). Event Relay streams events to AWS natively and securely using Eventbridge Partner APIs, removing the need to manage/share credentials and maintain middleware for each integration.

The actual Event Bus (and Partner Source) is generated automatically by Salesforce Event Relay, so it is not defined in this PR (we just reference it via its ARN).

This PR adds the following resources:

- A placeholder Lambda so that it can be incorporated into the support-service-lambdas repo
- A dead letter queue (DLQ) that can be used by the Event Bus rules if they fail to send events to the targets successfully
- An alarm that notifies us if the DLQ has one or more items
- A rule that sends `Contact_Update__e` platform events from Salesforce to the salesforce-outbound-messages-[STAGE] queue that is consumed by membership-workflow

This new rule is intended to completely replace [salesforce-message-handler](https://github.com/guardian/salesforce-message-handler). The rule uses an input transformer to convert the incoming event to the required message for the destination queue `salesforce-outbound-messages-<STAGE>` (and its consumer, membership-workflow).

Screenshot 2025-06-25 at 11 41 48

An upcoming change in Salesforce will switch from sending outbound messages from contact record updates to publishing `Contact_Update__e` platform events that will be picked up by this new Event Bus - it is fine to merge this change immediately, the change in Salesforce is the "switch" the moves between the old outbound message mechanism to the new Platform Events. Once the switch is made in Salesforce, salesforce-message-handler can be archived.

## How to test
Update a Contact record in Salesforce, and verify that the item appears on the salesforce-outbound-messages-CODE queue. This can be verified by checking in the membership-workflow logs and looking for the updated contact's Record ID.

We can retire salesforce-message-handler, easily and quickly set up new Salesforce-to-AWS integrations and also look at migrating other event-driven integrations across to the new Event Bus (e.g. [batch-email-sender](https://github.com/guardian/support-service-lambdas/tree/main/handlers/batch-email-sender)).

We should closely monitor performance once the new system is live, including the functionality update in Salesforce. It should be straightforward to revert the changes, simply by switching back to the old workflow rule that sends the outbound message.

### DIFF
--- a/cdk/bin/cdk.ts
+++ b/cdk/bin/cdk.ts
@@ -360,6 +360,8 @@ new SalesforceEventBus(app, 'salesforce-event-bus-CODE', {
 });
 new SalesforceEventBus(app, 'salesforce-event-bus-PROD', {
 	stack: 'support',
+	stage: 'PROD',
+});
 new MParticleApi(app, 'mparticle-api-CODE', {
 	stack: 'membership',
 	stage: 'CODE',


cdk/lib/salesforce-event-bus.ts
```diff
@@ -0,0 +1,117 @@
...
+			functionName: `${app}-placeholder-${this.stage}`,
+		});
+
+		const salesforceBus = EventBus.fromEventBusArn(
+			this,
+			'SalesforceBus',
+			this.stage === 'PROD'
+				? `arn:aws:events:${this.region}::event-source/aws.partner/salesforce.com/00D20000000nq5gEAA/0YLQv00000000zJOAQ`
+				: `arn:aws:events:${this.region}::event-source/aws.partner/salesforce.com/00D9E0000004jvhUAA/0YLUD00000008Ll4AI`,
+		);
```
graham228221, line 43: The Event Bus is created automagically from Salesforce via Eventbridge Partner APIs, we can't create it ourselves in CDK or in the console.

cdk/lib/salesforce-event-bus.ts
```diff
@@ -0,0 +1,117 @@
...
+
+		const deadLetterQueue = new Queue(this, `dead-letters-${app}-queue`, {
+			queueName: `dead-letters-${app}-queue-${props.stage}`,
+			retentionPeriod: Duration.days(14),
+		});
```
graham228221, line 48: This is a dead letter queue for the rule itself - the target (in membership-workflow) has its own dead letter queue. If the rule fails to send the message to the target, it will get put into this DLQ.

Handily, the attributes on the message in the DLQ includes the error code and reason:
Screenshot 2025-06-23 at 17 28 27

cdk/lib/salesforce-event-bus.ts
```diff
@@ -0,0 +1,117 @@
...
+
+		new GuLambdaFunction(this, 'SalesforceEventBusPlaceholder', {
+			app,
+			memorySize: 1024,
+			fileName: `${app}.zip`,
+			runtime: nodeVersion,
+			timeout: Duration.seconds(300),
+			handler: 'salesforceEventBusPlaceholder.handler',
+			functionName: `${app}-placeholder-${this.stage}`,
+		});
```
graham228221, line 35: Is it possible to have this in support-service-lambdas without a lambda?

cdk/lib/salesforce-event-bus.ts
```diff
@@ -0,0 +1,117 @@
...
+		contactUpdateToSqsRule.addTarget(
+			new SqsQueue(sfOutboundMessageQueue, {
+				deadLetterQueue: deadLetterQueue,
+				maxEventAge: Duration.hours(2),
+				retryAttempts: 2,
+				message: RuleTargetInput.fromObject({
+					//   "InputPathsMap": { "detail-payload-Contact_ID__c": "$.detail.payload.Contact_ID__c" }
+					//   "InputTemplate": "{\"contactId\": <detail-payload-Contact_ID__c>}"
+					contactId: EventField.fromPath('$.detail.payload.Contact_ID__c'),
+				}),
```
graham228221, line 91: Screenshot 2025-06-25 at 11 41 48

cdk/lib/salesforce-event-bus.ts
johnduffell: sorry the page was cached so I didn't see Tom's comment for some reason until mine sent!  But I think it's a good point that this whole thing could be in the membership-workflow repo (although that's no actually CDKed so it might be more trouble than it's worth.)

cdk/lib/salesforce-event-bus.ts
johnduffell: it's hard to check these IDs are correct! I'm not sure if it's possible/desirable to look up by tags or something though.

cdk/lib/salesforce-event-bus.ts
```diff
@@ -0,0 +1,117 @@
...
+		const salesforceBus = EventBus.fromEventBusArn(
+			this,
+			'SalesforceBus',
+			this.stage === 'PROD'
+				? `arn:aws:events:${this.region}::event-source/aws.partner/salesforce.com/00D20000000nq5gEAA/0YLQv00000000zJOAQ`
+				: `arn:aws:events:${this.region}::event-source/aws.partner/salesforce.com/00D9E0000004jvhUAA/0YLUD00000008Ll4AI`,
+		);
+
+		const deadLetterQueue = new Queue(this, `dead-letters-${app}-queue`, {
+			queueName: `dead-letters-${app}-queue-${props.stage}`,
```
johnduffell, line 46: A minor point and I notice it's a mixed bag already, but normally I prefer to suffix rather than prefix with -dead-letter or similar, because the AWS console search only searches on prefix, so you can see all the related queues at the same time, rather than all the dead letter queues at the same time.
However I realise there are a lot with the prefix already, so I'm not sure if that's a new AWS standard or something?

cdk/lib/salesforce-event-bus.ts
```diff
@@ -0,0 +1,117 @@
...
+			retentionPeriod: Duration.days(14),
+		});
+
+		new GuAlarm(this, 'DeadLetterQueueAlarm', {
+			app: app,
+			alarmName: `An event for ${props.stage} ${app} was not processed`,
+			alarmDescription:
+				`There is one or more event in the ${app} dead letter queue (DLQ). ` +
+				'Check that the rule corresponding to the failed message is correctly configured.\n' +
+				`DLQ: https://eu-west-1.console.aws.amazon.com/sqs/v2/home?region=eu-west-1#/queues/https%3A%2F%2Fsqs.eu-west-1.amazonaws.com%2F865473395570%2F${deadLetterQueue.queueName}`,
```
johnduffell, line 56: nice touch to add the link.  It might also be worth mentioning that any errors would appear on the properties of the message, as it's not really obvious at first.

johnduffell: also as Tom mentioned elsewhere, it's recommended not to put the AWS account ids in the repo.

handlers/alarms-handler/src/alarmMappings.ts
```diff
@@ -103,6 +103,7 @@ const teamToAppMappings: Record<Team, string[]> = {
 		'salesforce-disaster-recovery-health-check',
 		'single-contribution-salesforce-writes',
 		'sf-datalake-export',
+		'salesforce-event-bus',
```
johnduffell, line 106: good spot to add this! 
Not necessarily one for this PR unless it's easy, but it would be good if we could just import line 25 of cdk/lib/salesforce-event-bus.ts rather than replicating it here.

tjmw:
> Out of interest it does make me wonder what extra value the SQS queue is adding if we're adding eventbridge? Eventbridge is a kind of multi queue with filtering, SQS is a single one.

On the apps-metering-events project I put an SQS queue between an EventBridge rule and a lambda because of the nice integration between SQS and lambda. Things like queue visibility, retry strategy and dead letter functionality which work out the box.

cdk/lib/salesforce-event-bus.ts
```diff
@@ -0,0 +1,117 @@
...
+				detailType: ['Contact_Update__e'],
+			},
+			eventBus: salesforceBus,
+		});
+
+		contactUpdateToSqsRule.addTarget(
+			new SqsQueue(sfOutboundMessageQueue, {
+				deadLetterQueue: deadLetterQueue,
+				maxEventAge: Duration.hours(2),
+				retryAttempts: 2,
```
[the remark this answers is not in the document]
graham228221: Ha, no just taking what was in [the docs](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_events_targets-readme.html#:~:text=dead%20letter%20queue-,maxEventAge%3A%20Duration.hours(2)%2C,-//%20Optional%3A%20set%20the) without thinking to much about it. 

It looks like we normally just leave out these optional values? I'll take them out as well (but please let me know if we should have them).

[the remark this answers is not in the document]
johnduffell: I don't know, I guess it could mean that if SQS was down or inaccessible it would try to queue it again. But I think the default backoff means that 2 retries would only take a couple of seconds.
If SQS was really down, then it wouldn't be able to put it in the DLQ anyway!

cdk/lib/salesforce-event-bus.ts
graham228221: > also as Tom mentioned elsewhere, it's recommended not to put the AWS account ids in the repo.

Thanks, removed here: https://github.com/guardian/support-service-lambdas/pull/2887/commits/cf77f12ac6878e92e2cdf5036a213b3d2827241e#diff-0d9875bc484ab7f61746c9d5a94172bf6b5569fe2e6429980fb74ca044233606R56

> It might also be worth mentioning that any errors would appear on the properties of the message, as it's not really obvious at first.

Good shout: https://github.com/guardian/support-service-lambdas/pull/2887/commits/d8d6b98728a621d5a00a9c168deac9b8bf9a06f9

cdk/lib/salesforce-event-bus.ts
```diff
@@ -0,0 +1,117 @@
...
+		);
+
+		const deadLetterQueue = new Queue(this, `dead-letters-${app}-queue`, {
+			queueName: `dead-letters-${app}-queue-${props.stage}`,
+			retentionPeriod: Duration.days(14),
+		});
+
+		new GuAlarm(this, 'DeadLetterQueueAlarm', {
+			app: app,
+			alarmName: `An event for ${props.stage} ${app} was not processed`,
```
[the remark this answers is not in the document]
graham228221: To be honest, no idea 😆 I think I copy/pasted from somewhere else.

Updated to use `this` for consistency: https://github.com/guardian/support-service-lambdas/pull/2887/commits/4eae1c6562eb1abe4290cc47e4559c2a47e33d6a

cdk/lib/salesforce-event-bus.ts
graham228221: Thanks for this @johnduffell @tjmw, I've removed the Lambda now (very painless!): https://github.com/guardian/support-service-lambdas/pull/2887/commits/f7de2d104b471047379531572b9c6ae38a506253

johnduffell (approved):
nice one, not too much in the PR now that the lambda is gone.  Looks good 👏 not sure how to get it into PROD and take over from the existing lambda - it might be worth running it into another queue for a while and comparing the data?  If the target is idempodent we can run both targeting the same queue for a while also.

PR #2660, opened 2025-01-29
update to latest GuCDK/CDK

johnduffell:
We are on quite an old GuCDK.

This PR updates to the latest.

This is needed so that I can add tags to alarms, which has only been supported since https://github.com/aws/aws-cdk/issues/29759#issuecomment-2081754930

I have commented all the types of change that are in the files, I don't think anything should change behaviour.

PR #2890, opened 2025-06-18
add DiagnosticLinks tag to most CDKed alarms

johnduffell:
## Note: when deploying to PROD, we need to manually add the alarmshandlerbackupemailaddress to alarms-handler stack.

following on from cdk update https://github.com/guardian/support-service-lambdas/pull/2660
and DiagnosticLinks in alarms handler https://github.com/guardian/support-service-lambdas/pull/2641

This PR adds a new SrLambdaAlarm construct ( cdk/lib/cdk/sr-lambda-alarm.ts ) and uses it where possible in our stacks.

I've also make some other tweaks so the alarms are enabled only in PROD by default, and they use the appropriate alarms handler topic.  This reduces duplication in our stacks.

I also noticed that alarms-handler alarms went to alarms-handler-topic.  This means if alarms handler broke, it would be silently looping.  I've added provision for those alarms to be emailed to us as I wasn't sure what best to do.

It has not been added everywhere (yet) - some places we use CfnAlarm rather than GuAlarm, which is apparently a low level construct.  And also places we use GuLambda's monitoringConfig rather than a separate alarm.  This can be resolved in future.

Also in future this may be exportable to support-frontend repo if it's useful there.

### DIFF
--- a/cdk/lib/alarms-handler.ts
+++ b/cdk/lib/alarms-handler.ts
@@ -16,7 +16,10 @@ import {
 import { LoggingFormat } from 'aws-cdk-lib/aws-lambda';
 import { SqsEventSource } from 'aws-cdk-lib/aws-lambda-event-sources';
 import { Topic } from 'aws-cdk-lib/aws-sns';
-import { SqsSubscription } from 'aws-cdk-lib/aws-sns-subscriptions';
+import {
+	EmailSubscription,
+	SqsSubscription,
+} from 'aws-cdk-lib/aws-sns-subscriptions';
 import { Queue } from 'aws-cdk-lib/aws-sqs';
 import { nodeVersion } from './node-version';
 
@@ -67,6 +70,15 @@ export class AlarmsHandler extends GuStack {
 			{
 				description:
 					'ARN of role in the targeting account which allows cloudwatch:ListTagsForResource',
+			},
+		);
+
+		const backupEmailAddress = new GuStringParameter(
+			this,
+			`${app}-backup-email-address`,
+			{
+				description:
+					'Alarm email address to use if the alarms handler itself fails',
 			},
 		);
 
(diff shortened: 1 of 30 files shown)


cdk/lib/batch-email-sender.ts
```diff
@@ -96,14 +94,13 @@ export class BatchEmailSender extends GuStack {
...
 		// ---- Alarms ---- //
-		new GuAlarm(this, 'FailedEmailApiAlarm', {
+		new SrLambdaAlarm(this, 'FailedEmailApiAlarm', {
 			app,
 			alarmName: `URGENT 9-5 - ${this.stage}: Failed to send email triggered by Salesforce - 5XXError (CDK)`,
 			alarmDescription: `API responded with 5xx to Salesforce meaning some emails failed to send. Logs at /aws/lambda/batch-email-sender-${this.stage} repo at https://github.com/guardian/support-service-lambdas/blob/main/handlers/batch-email-sender/`,
 			evaluationPeriods: 1,
 			threshold: 1,
-			actionsEnabled: isProd,
-			snsTopicName: alarmTopic,
```
johnduffell, line 106: SrLambdaAlarm fills these two props with sensible values.  But we do need to let it know which lambda function it is, so it can produce a log link in the message.

cdk/lib/cdk/sr-lambda-alarm.ts
```diff
@@ -0,0 +1,32 @@
+import type { GuAlarmProps } from '@guardian/cdk/lib/constructs/cloudwatch';
+import { GuAlarm } from '@guardian/cdk/lib/constructs/cloudwatch';
+import type { GuStack } from '@guardian/cdk/lib/constructs/core';
+import { Tags } from 'aws-cdk-lib';
+
+interface SRLambdaAlarmProps
+	extends Omit<GuAlarmProps, 'snsTopicName' | 'actionsEnabled'> {
+	lambdaFunctionNames: string | string[];
+	actionsEnabled?: boolean; // defaults to PROD only
```
johnduffell, line 9: this is in case people want to enable it in CODE for testing purposes

cdk/lib/alarms-handler.ts
```diff
@@ -160,7 +180,7 @@ export class AlarmsHandler extends GuStack {
 				runtime: nodeVersion,
 				timeout: Duration.seconds(15),
 				handler: 'indexScheduled.handler',
-				functionName: `${app}-scheduled-${this.stage}`,
```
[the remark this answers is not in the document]
johnduffell: oops, finger trouble, good spot!

cdk/lib/new-product-api.ts
```diff
@@ -97,14 +95,13 @@ export class NewProductApi extends GuStack {
...
-		new GuAlarm(this, 'ApiGateway4XXAlarm', {
+		new SrLambdaAlarm(this, 'ApiGateway4XXAlarm', {
 			app,
 			alarmName: `new-product-api-${this.stage} API gateway 4XX response`,
 			alarmDescription: 'New Product API received an invalid request',
 			evaluationPeriods: 1,
 			threshold: 6,
-			actionsEnabled: isProd,
-			snsTopicName: alarmTopic,
+			lambdaFunctionNames: addSubscriptionLambda.functionName,
```
MaelGNM, line 104: Should this include both function names?
```suggestion
			lambdaFunctionNames: [
				addSubscriptionLambda.functionName,
				productCatalogLambda.functionName,
			],
```

PR #2899, opened 2025-06-23
Negative invoices processor - create case in salesforce

david-pepper:


johnduffell:

PR #2907, opened 2025-06-26
fail build if TS build files are not consistent (for alarms-handler and discount-api)

johnduffell:
## Context

We have a lot of (supposedly) similar build files across the repo, including package.json, riff-raff.yaml and more.

There's no easy way to make sure they keep in step or encourage recommended dependencies, especially AWS sdk versions.  The pnpm catalog can help but there's nothing to encourage people to look at it and things still get in without it e.g.
https://github.com/guardian/support-service-lambdas/blob/87716c73ac9ecb0c3776c9d30f94f6bc219b314b/handlers/mparticle-api/package.json#L21
There's also difficulty in reliably creating and reviewing the boilerplate for new lambdas.

Having said all that, despite some strange issues with AWS sdk versions, and some modules not running their tests, we are getting by reasonably well as things are.

I couldn't find any material on how you could use pnpm in an sbt like way on a mono repo like this.

## This PR

This PR implements a kind of "CDK for build files" tool called buildcheck, covering only the two lambdas converted to split-cdk in the above PR:
- to check (in CI) for drift in the managed build files
- to provide a command to refresh the managed build files from the build definition

### CI checks
If there is drift, it would appear in the build log for that lambda as follows:
```
  ● file on disk (+) contains the expected content (-) › handlers/alarms-handler/jest.config.js (template: buildcheck/data/handlerTemplate/jest.config.js.ts)

    expect(received).toBe(expected) // Object.is equality

    - Expected  - 1
    + Received  + 0

    @@ -6,8 +6,7 @@
      	runner: 'groups',
      	moduleNameMapper: {
      		'@modules/(.*)/(.*)$': '<rootDir>/../../modules/$1/src/$2',
      		'@modules/(.*)$': '<rootDir>/../../modules/$1',
      	},
    - 	testPathIgnorePatterns: ['<rootDir>/cdk/'],
      };
```
(see build https://github.com/guardian/support-service-lambdas/actions/runs/17913821387/job/50931392992?pr=2907 )

### Managed files:

 | here are the managed files marked in red (note, they are all committed as normal to github) | here is the same view showing only the non-managed files | and here are the central templates and build/dependencies.ts |
| -- | -- | -- |
| image | image | image |

I have deployed the two lambdas to CODE successfully

## Pros and cons of buildcheck
The benefits and drawbacks of buildcheck are similar to CDK or SBT:
- prevent inconsistencies between lambdas/modules including dependency versions and pnpm scripts
- have a central (auto completable) list of recommended dependencies
- files can be generated where necessary, or buildcheck can be pulled in as a module to reuse definitions
- allows more fine grained modules to be manageable (i.e. separate CDK per lambda) improving build times and organisation
- easier to add and review new lambda boilerplate as it's guaranteed to be standard
- all the usual DRY benefits

The disadvantages are mainly around tooling:
- automated PRs to bump dependencies will fail unless the dependencies.ts file is updated accordingly
- harder to add dependencies as you need to add them to build.ts and then run update-build.
- non standard, could surprise new people
- extra level of abstraction, could slow down regular developers

## Alternative options
- no change (also see rollback plan below)
- bazel/pants/Nx/Turborepo/more advanced build system
  - of those, Nx looks most promising, certainly as a longer term target as it handles incremental build and better dependency between modules.  however it would require a larger migration effort.

## Rollback plan
At any later point if we decide the benefits are not worth it, we can easily remove buildcheck without disruption:
1. remove buildcheck from the GHA workflow
1. amend buildcheck to not add the "this is a generated file" notices
1. rerun `pnpm snapshot:update` to refresh the files for a final time
1. delete the buildcheck folder
1. push the changes

## Future work (in later PRs)
- include `modules/*` rather than just `handlers/*` in the checker logic
- migrate more lambdas over time (focus on the value teams' handlers first)
- possibly share code between buildcheck and lambdas/cdk to reduce duplication

### Feedback from the mob review
- enforce use of the pnpm catalog (this is a separate change from this PR but likely a good idea, although there's no dependabot and poor IDE support for pnpm catalog)
  - option 1 - pnpm-workspace is primary https://github.com/guardian/support-service-lambdas/pull/3096
  - option 2 - build definition is primary https://github.com/guardian/support-service-lambdas/pull/3098
- can you have all the scripts "streamed" at the root package.json? (I don't know what this means and there doesn't seem to be any kind of facility like this)


.github/workflows/ci-typescript.yml
```diff
@@ -72,7 +72,10 @@ jobs:
           node-version-file: '.nvmrc'
           cache: 'pnpm'
       - run: pnpm install
-      - run: pnpm --filter cdk package
+      - run: |
+          cdk/generator/generate.sh "${{ matrix.subproject }}"
```
johnduffell, line 76: this will unnecessarily generate a few files for projects that haven't been moved across, but firstly that will be super quick, and secondly if this is approved, we should move the rest across very soon

cdk/bin/cdk.ts
```diff
@@ -218,14 +217,6 @@ new ProductSwitchApi(app, 'product-switch-api-PROD', {
...
 });
 
-new AlarmsHandler(app, 'alarms-handler-CODE', {
-	stack: 'support',
-	stage: 'CODE',
-});
-new AlarmsHandler(app, 'alarms-handler-PROD', {
-	stack: 'support',
-	stage: 'PROD',
-});
```
johnduffell, line 228: this is now generated only for the alarms handler build, not for all CDK builds.

cdk/module/node-version.ts
johnduffell, line 1: this is a copy of cdk/lib/node-version.ts but it's exported for use by individual handlers/*/cdk/ modules.
I didn't want to import the main CDK lib as there's a lot there to untangle, and I didn't want to change all the other cdk files to reference this new one as it would add a lot of clutter to the PR.
In time the other will become redundant, although ideally we should do the change to make everything use this one, maybe in a separate PR for clarity.

handlers/alarms-handler/.eslintignore
johnduffell, line 1: I had trouble with the handler package picking up stuff from the nested cdk package.
I did consider various solutions e.g. putting the handler and cdk alongside each other, making them both cooperate enough to have them as one build, or even making them one package with multiple scripts.
However I think all those are beyond the scope of what I can do for now, so for now I have manually added exclusions of "cdk/" that were needed.

handlers/alarms-handler/cdk/bin/cdk.ts
johnduffell, line 1: this is just the existing snippet from above but pulled out to this file

.husky/post-merge
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh -e
```
johnduffell, line 1: this file is run when you do a git pull after you have run pnpm install the first time at the root.
We may need something in the readme or a script when people are doing initial setup so they don't miss out on running this the first time.

cdk/module/package.json
```diff
@@ -0,0 +1,7 @@
+{
```
johnduffell, line 1: I had to add this before anything could import the above node-version file, it works, however it was auto generated by copilot so needs checking

handlers/alarms-handler/.eslintignore
```diff
@@ -0,0 +1 @@
+cdk/
```
johnduffell, line 1: I had trouble with the handler package picking up stuff from the nested cdk package.
I did consider various solutions e.g. putting the handler and cdk alongside each other, making them both cooperate enough to have them as one build, or even making them one package with multiple scripts.
However I think all those are beyond the scope of what I can do for now, so for now I have manually added exclusions of "cdk/" that were needed.

.github/workflows/ci-typescript.yml
```diff
@@ -71,8 +71,22 @@ jobs:
...
+        run: |
+          if [ -d "./handlers/${{ matrix.subproject }}/cdk" ]; then
+            echo "cdk-path=./handlers/${{ matrix.subproject }}/cdk" >> $GITHUB_OUTPUT
+            echo "lambda-path=./handlers/${{ matrix.subproject }}/lambda" >> $GITHUB_OUTPUT
+            echo "cdk-module=${{ matrix.subproject }}-cdk" >> $GITHUB_OUTPUT
+          else
+            echo "cdk-path=./cdk" >> $GITHUB_OUTPUT
+            echo "lambda-path=./handlers/${{ matrix.subproject }}" >> $GITHUB_OUTPUT
+            echo "cdk-module=cdk" >> $GITHUB_OUTPUT
+          fi
```
johnduffell, line 86: if it's got a cdk folder then we use it and assume the lambda folder is alongside, otherwise the traditional layout

buildgen/package.json
```diff
@@ -0,0 +1,22 @@
+{
+  "name": "buildgen",
+  "version": "1.0.0",
+  "private": true,
+  "description": "Build file generator for handlers",
+  "scripts": {
+    "generate": "bash scripts/generate-template-list.sh && ts-node src/generate.ts \"$(pwd | sed 's|/buildgen.*||')\"",
+    "check": "bash scripts/generate-template-list.sh && tsc --noEmit",
+    "clean": "rm -rf src/data/generated.ts dist",
```
johnduffell, line 9: clean only cleans buildgen internally, there's not yet a clean for the generated files.

buildgen/jest.config.js
```diff
@@ -0,0 +1,6 @@
+/** @type {import('ts-jest').JestConfigWithTsJest} */
+module.exports = {
+	preset: 'ts-jest',
+	testEnvironment: 'node',
+	runner: 'groups',
+};
```
johnduffell, line 6: ironically the buildgen package needs to have all the generic files otherwise we'd have a chicken and eg situation

buildgen/scripts/generate-template-list.sh
```diff
@@ -0,0 +1,50 @@
+#!/bin/bash -e
+# generates a typescript file with all the template files imported
```
johnduffell, line 2: this script is a bit of a nice to have.

It's good to have the template folder with exactly the structure the files will be created in the lambda.
However it's tricky to iterate the folder in typescript, especially in a type-safe way.

The solution is to have a single file that imports each template file, and returns records containing the file name and the file content, then the compiler will type check it all.

This script generates the above mentioned file, so that we don't have to keep it up to date.

The single file looks like this when generated:
```typescript
import { Template } from '../util/templater';

import cdk__prettierignore from './template/cdk/.prettierignore';
import cdk_cdk_json from './template/cdk/cdk.json';
import cdk_jest_setup_js from './template/cdk/jest.setup.js';
import cdk_package_json from './template/cdk/package.json';
import cdk_tsconfig_json from './template/cdk/tsconfig.json';
import lambda_jest_config_js from './template/lambda/jest.config.js';
import lambda_package_json from './template/lambda/package.json';
import lambda_tsconfig_json from './template/lambda/tsconfig.json';
import riff_raff_yaml from './template/riff-raff.yaml';

/*
 * This file is gitignored and edits will be overwritten.
 * Generated by scripts/generate-template-list.sh
 * Generated on Fri Jul 18 22:48:24 BST 2025
 */

const templates: Template[] = [
  { name: "cdk/.prettierignore", template: cdk__prettierignore },
  { name: "cdk/cdk.json", template: cdk_cdk_json },
  { name: "cdk/jest.setup.js", template: cdk_jest_setup_js },
  { name: "cdk/package.json", template: cdk_package_json },
  { name: "cdk/tsconfig.json", template: cdk_tsconfig_json },
  { name: "lambda/jest.config.js", template: lambda_jest_config_js },
  { name: "lambda/package.json", template: lambda_package_json },
  { name: "lambda/tsconfig.json", template: lambda_tsconfig_json },
  { name: "riff-raff.yaml", template: riff_raff_yaml },
];

export default templates;
```

buildgen/scripts/generate-template-list.sh
```diff
@@ -0,0 +1,50 @@
...
+script_name=$(basename "$0")
+
+echo "$script_name: START generating template list..."
+
+script_dir=$(readlink -f "$(dirname "${BASH_SOURCE[0]}")")
+src_dir=$(readlink -f "$script_dir/../src")
```
johnduffell, line 8: readlink resolves the ".." from the filenames which makes it easier to debug

buildgen/src/data/config.ts
```diff
@@ -0,0 +1,17 @@
...
+}
+
+export const generatorConfig: GeneratorConfig = {
+	packages: [
+		{
+			name: 'alarms-handler',
+			functionNames: ['alarms-handler-', 'alarms-handler-scheduled-'],
+		},
+	],
+};
```
johnduffell, line 17: this is a list of supported handlers, and any data to be templated into the files

buildgen/src/data/template/cdk/.prettierignore.ts
```diff
@@ -0,0 +1 @@
+export default 'dist/\n';
```
johnduffell, line 1: these files in src/data/template/ reflect the structure that will be created in the handler

buildgen/src/data/template/cdk/package.json.ts
```diff
@@ -0,0 +1,40 @@
+import { HandlerConfig } from '../../config';
+
+export default (config: HandlerConfig) => ({
+	name: `${config.name}-cdk`,
```
johnduffell, line 4: this is an example of one that exports a function to take the config and use it for the content

buildgen/src/data/template/riff-raff.yaml.ts
```diff
@@ -0,0 +1,29 @@
...
+				},
+			},
+		},
+		[pkg.name]: {
+			type: 'aws-lambda',
+			parameters: {
+				fileName: `${pkg.name}.zip`,
+				bucketSsmLookup: true,
+				prefixStack: false,
+				functionNames: pkg.functionNames,
```
johnduffell, line 24: it's easier to have dynamic function names if the yaml is written as an object and serialised later.  However on the flip side it's less copy and pasteable.

buildgen/jest.config.js
```diff
@@ -3,8 +3,4 @@ module.exports = {
 	preset: 'ts-jest',
 	testEnvironment: 'node',
 	runner: 'groups',
-	moduleNameMapper: {
-		'@modules/(.*)/(.*)$': '<rootDir>/../../modules/$1/src/$2',
-		'@modules/(.*)$': '<rootDir>/../../modules/$1',
-	},
```
johnduffell, line 9: git thinks this file was renamed and edited, but actually it's a totally new file with similar name and content.

package.json
```diff
@@ -12,7 +12,9 @@
 		"new-lambda": "pnpm hygen new-lambda api-gateway && pnpm install && pnpm --filter cdk lint --fix && pnpm fix-formatting && pnpm --filter cdk test -- -u && git add ./cdk/lib/__snapshots__/*.test.ts.snap",
 		"prepare": "husky",
 		"fix-formatting": "pnpm -r run fix-formatting",
-		"check-formatting": "pnpm -r run check-formatting"
+		"check-formatting": "pnpm -r run check-formatting",
+        "preinstall": "if [ ! -d buildgen/node_modules ]; then cd buildgen && pnpm install --frozen-lockfile --ignore-scripts; cd ..; fi && pnpm run buildgen && pnpm install --ignore-scripts",
```
johnduffell, line 16: preinstall actually happens *after* install has completed, so if we generate package.jsons then it's too late.
So we do a trick of running buildgen, then running pnpm install again with --ignore-scripts, which will pick up the missed package.jsons without ending up in a recursive loop.
Because pnpm install doesn't redo work that's already done, there seems to be little extra performance impact.

tsconfig.json
```diff
@@ -41,6 +41,7 @@
       "@modules/identity/*": ["./modules/identity/src/*"],
       "@modules/routing/*": ["./modules/routing/src/*"],
       "@modules/secrets-manager/*": ["./modules/secrets-manager/src/*"],
+      "@modules/cdk/*": ["./cdk/module/*"],
```
johnduffell, line 44: allows access the shared CDK code (i.e. node-version) from the individual handlers' cdk.

buildgen/README.md
```diff
@@ -0,0 +1,55 @@
...
+pnpm run generate
+
+# Generate a specific handler only
+pnpm run generate <handler-name>
+```
+- TODO add a global clean script
+- TODO prettier/lint 
+- TODO write a script or pnpm target to do both migrate and diff
+- TODO snapshot tests
+
```
johnduffell, line 37: it's probably not worth doing all these TODOs before the concept is reviewed, because this PR is more of a working proof of concept.  If the concept is good then we can easily implement all these other things as per our usual project standards.

buildgen/package.json
```diff
@@ -0,0 +1,26 @@
...
+  "version": "1.0.0",
+  "private": true,
+  "description": "Build file generator for handlers",
+  "scripts": {
+    "buildgen": "pnpm snapshot:update",
+    "snapshot:update": "bash scripts/generate-template-list.sh && REPO_ROOT=\"$(pwd | sed 's|/buildgen.*||')\" && ts-node src/cli.ts --generate \"$REPO_ROOT\"",
+    "snapshot:clean": "bash scripts/generate-template-list.sh && REPO_ROOT=\"$(pwd | sed 's|/buildgen.*||')\" && ts-node src/cli.ts --clean \"$REPO_ROOT\"",
+    "snapshot:assert": "bash scripts/generate-template-list.sh && REPO_ROOT=\"$(pwd | sed 's|/buildgen.*||')\" && bash -c 'jest --testNamePattern=\"checkActual $*\"' -- \"$@\"",
+    "migrate-handler": "REPO_ROOT=\"$(pwd | sed 's|/buildgen.*||')\" && bash scripts/migrate-handler.sh \"$REPO_ROOT\"",
+    "check": "bash scripts/generate-template-list.sh && tsc --noEmit",
```
johnduffell, line 12: TODO need to add linting and prettier to this module (buildgen)

buildgen/scripts/migrate-handler.sh
```diff
@@ -0,0 +1,75 @@
+#!/bin/bash -e
```
johnduffell, line 1: this does most of the mechanical work of moving from the old to the new handler structure.  There's still a bit of manual work to do after running this, see the README.

buildgen/src/cli.ts
```diff
@@ -0,0 +1,33 @@
...
+
+	const previouslyGeneratedFiles = loadGeneratedFileNames(
+		repoRoot,
+		warningFileName,
+	);
+	console.log('previouslyGeneratedFiles to delete', previouslyGeneratedFiles);
+
+	switch (mode) {
+		case 'generate':
+			deleteRepoFiles(repoRoot, previouslyGeneratedFiles);
```
johnduffell, line 22: we always effectively do a clean build, to make sure that files don't hang around if they're no longer managed

buildgen/src/data/build.ts
```diff
@@ -0,0 +1,43 @@
+import { dep } from './dependencies';
```
johnduffell, line 1: this file here is the main build definition, it has a record for each handler with any handler specific information and dependencies.  The information is templated into the various riffraff, package.json etc. and can be extended when we find more differences between handlers.

In future, there's no fundamental reason why the handlers' cdk and lambda modules couldn't reference shared values from here.

buildgen/src/data/dependencies.ts
```diff
@@ -0,0 +1,17 @@
+const awsClientVersion = '^3.848.0';
+
+const awsSdk = (name: string) => ({
```
johnduffell, line 3: we could probably change `string` to a big union type to get better auto complete

buildgen/src/data/handlerTemplate/cdk/.prettierignore.ts
```diff
@@ -0,0 +1 @@
+export default 'dist/\n';
```
johnduffell, line 1: these files are the handler templates, whatever the ts file exports is what ends up in the file with the according name.  I.e. this is
`buildgen/src/data/handlerTemplate/cdk/.prettierignore.ts`
so the resulting file will be
`handler/<myHandler>/cdk/.prettierignore`

.github/workflows/ci-typescript.yml
```diff
@@ -73,8 +73,28 @@ jobs:
...
+
       - run: pnpm install
-      - run: pnpm --filter cdk package
+      - run: |
+          if [ -f "./handlers/${{ matrix.subproject }}/BUILDCHECK.md" ]; then
+            pnpm --filter "./buildcheck" check-formatting
+            pnpm --filter "./buildcheck" lint
+            pnpm --filter "./buildcheck" snapshot:assert handlers/${{ matrix.subproject }}
+          fi
```
johnduffell, line 96: skip buildcheck completely if it's not a managed handler, and filter the check down to files that are in this handler

buildcheck/data/handlerTemplate/cdk/cdk.json.ts
```diff
@@ -0,0 +1,7 @@
+export default {
```
johnduffell, line 1: for the purposes of this PR, I have little opinion on the contents of the resulting files, I just tried to generate roughly what was already there.

cdk/bin/cdk.ts
```diff
@@ -30,14 +28,16 @@ import { UpdateSupporterPlusAmount } from '../lib/update-supporter-plus-amount';
...
+	supportHostedZoneId,
+} from '../module/constants';
 
 const app = new App();
-const membershipHostedZoneId = 'Z1E4V12LQGXFEC';
-const membershipCertificateId = 'c1efc564-9ff8-4a03-be48-d1990a3d79d2';
-const membershipApisDomain = 'membership.guardianapis.com';
-export const supportHostedZoneId = 'Z3KO35ELNWZMSX';
-export const supportCertificateId = 'b384a6a0-2f54-4874-b99b-96eeff96c009';
-export const supportApisDomain = 'support.guardianapis.com';
```
johnduffell, line 40: moved to a shared module so that individual cdk modules can also use them

buildcheck/data/build.ts
```diff
@@ -0,0 +1,45 @@
...
+	name: 'discount-api',
+	dependencies: {
+		...dep.dayjs,
+		...dep['source-map-support'],
+		...dep.zod,
+	},
+	devDependencies: {
+		...dep['@types/aws-lambda'],
+	},
+};
```
johnduffell, line 43: each handler will have an object defined here with data to be included in the templates

buildcheck/src/util/templater.ts
```diff
@@ -0,0 +1,55 @@
...
+		return content;
+	}
+
+	const extension = path.extname(relativePath);
+	if (extension === '.yaml') {
+		return yaml.dump(content, { indent: 2 });
+	}
+
+	if (extension === '.json') {
+		return JSON.stringify(content, null, 2);
```
johnduffell, line 49: should we add a trailing newline here?

.github/workflows/ci-typescript.yml
```diff
@@ -30,6 +30,10 @@ jobs:
           node-version-file: '.nvmrc'
           cache: 'pnpm'
       - run: pnpm install
+      - run: pnpm --filter "./buildcheck" type-check
+      - run: pnpm --filter "./buildcheck" check-formatting
+      - run: pnpm --filter "./buildcheck" lint
+      - run: pnpm --filter "./buildcheck" snapshot:assert ROOT
```
johnduffell, line 36: this line checks that the root level managed files (outside individual handlers and modules) are correct, and the per handler one below just checks the files for that handler.

buildcheck/data/dependencies.ts
```diff
@@ -0,0 +1,18 @@
+import {
```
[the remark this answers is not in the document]
johnduffell: as per feedback, there is a follow on PR if we want to move over to reading the catalog as the source of truth instead of this dependencies.ts file
https://github.com/guardian/support-service-lambdas/pull/3096
or putting the pnpm catalog as a managed file
https://github.com/guardian/support-service-lambdas/pull/3098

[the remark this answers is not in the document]
johnduffell: as per mob feedback, as future improvement we can make package.json use `"catalog:"` for all of these
To do that we can either make pnpm catalog generated from this file https://github.com/guardian/support-service-lambdas/pull/3098 , or
generate this file from pnpm catalog https://github.com/guardian/support-service-lambdas/pull/3096

I didn't put them in this PR because 1. I wasn't sure which approach will be preferred, and 2. this PR is already big enough.

.github/workflows/ci-typescript.yml
```diff
@@ -74,6 +78,7 @@ jobs:
           node-version-file: '.nvmrc'
           cache: 'pnpm'
       - run: pnpm install
+      - run: pnpm --filter "./buildcheck" snapshot:assert handlers/${{ matrix.subproject }}
```
johnduffell, line 81: this checks just the files for the current handler (if any)

buildcheck/jest.config.js
```diff
@@ -0,0 +1,6 @@
+/** @type {import('ts-jest').JestConfigWithTsJest} */
```
johnduffell, line 1: the irony is that buildcheck has its own set of build files that can't be managed by itself

rupertbates (approved):
As discussed, let's try it out and see how we get on!

PR #2912, opened 2025-06-27
drop SOI messages without identity ids

johnduffell:
Following on from pre-refactor https://github.com/guardian/support-service-lambdas/pull/2923

We have had some alarms because sometimes identity id is not set for records in the SOI queue, however this lambda expects identity id to always be set.

---

It's possible that identity id could be empty if payment API couldn't create an identity account as part of the post payment tasks
https://github.com/guardian/support-frontend/blob/d414d71e4d9df0fb2b144fa0bbd49021c0d3701d/support-payment-api/src/main/scala/backend/StripeBackend.scala#L280-L281

Until 15th May, we used to explicitly drop those https://github.com/guardian/support-frontend/blob/de28ba3e5a965956bb6c2586feaf9c54063f3e64/support-payment-api/src/main/scala/services/SoftOptInsService.scala#L27-L29
but the above logic was removed on 15th May https://github.com/guardian/support-frontend/pull/7026

See further chat on here: https://chat.google.com/room/AAAA6HaLTYQ/h_R2Fl-oTC8/h_R2Fl-oTC8?cls=10

It seems like the change in logic was not when the queue started getting full, so there may be another issue (outside of payment api) that was hidden by the fact that we didn't get notification of alarms stuck in alarm state.
image
(ignore the large surge in Apr/may which was fixed in https://github.com/guardian/support-service-lambdas/pull/2817 )

---

Although we could try to filter out the records in CDK as part of the Rule that pulls into this queue, this is hard to develop and test.  Therefore the easiest thing is to put the check in this lambda.

### This PR
This PR makes SOI queue reader drop single contribution SOI records without an identity ID.

In more low level detail:
currently we have either identity id is set, or the whole lambda fails.  There are unit tests for the "success" case.

I needed to add a "filter" state depending on the type - either

- success - as before
- filter - type is single contribution and identity id is not present
- fail - any other type and identity id is not present (or any other parsing issue).

I could have
- made a custom circe decoder to do the validation (I decided "normal" business logic is easier to understand, also it's easier to unit test when I don't have to deal with exceptions)
- made a custom type (this would have been more code, but better naming)
- model the 3 states as Try[Option[A]].  The try indicates whether it's fail state or not, and if it isn't fail, then the Option indicates whether it should be filtered.

I went for the last option in the end.

### Testing
tested locally - unit tests added


handlers/soft-opt-in-consent-setter/src/test/scala/com/gu/soft_opt_in_consent_setter/JsonCodecSpec.scala
```diff
@@ -76,9 +75,40 @@ class JsonCodecSpec extends AnyFlatSpec with should.Matchers with Inside {
...
+        |        "similarGuardianProducts": true
+        |    }
+        |}""".stripMargin
+    val expected = "1234"
+    inside(HandlerIAP.parseMessages(List(testDataNoIdentityId, testDataWithIdentityId))) { case List(actual) =>
+      inside(actual) { case messageBody: MessageBody =>
+        messageBody.identityId should be(expected)
+      }
+    }
+  }
```
johnduffell, line 112: new test to check that no identity id is filtered, but not with identity id

handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/HandlerIAP.scala
```diff
@@ -62,37 +78,52 @@ object HandlerIAP extends LazyLogging with RequestHandler[SQSEvent, Unit] {
...
-              SoftOptInError(s"Unknown error when decoding JSON to MessageBody with body: ${message.getBody}", ex)
-            handleError(exception)
-          case Right(result) =>
-            logger.info(s"Decoded message body: $result")
-            result
-        },
-      )
+    val messages: List[MessageBody] = parseMessages(input.getRecords.asScala.toList.map(_.getBody))
+      .collect {
+        case Failure(exception) => rethrowError(exception)
```
johnduffell, line 124: there's only ever one record as it's super low traffic and we set the max messages to process to 1 in CDK, so eagerly throwing is not a big issue.

MaelGNM (approved):
👍

PR #2913, opened 2025-06-30
Refactor submitDataSubjectRequest to remove lambdaDomainUrl parameter…

andresilva-guardian:
This pull request simplifies the handling of API gateway URLs in the `mparticle-api` service by removing the concept of generating the lambda domain URL based on the request and replacing it with a secret-based approach. 

### Simplification of API Gateway URL Handling:

* [`handlers/mparticle-api/src/apis/data-subject-requests.ts`](diffhunk://#diff-576d76677bb1f209c474d33ac5b8abe5d0924e3ff5547e3dc046a5a8e7efd118L80-R82): Removed the `lambdaDomainUrl` parameter from the `submitDataSubjectRequest` function. The callback URL is now constructed using a secret value retrieved from the environment instead of relying on the dynamic lambda domain URL. [[1]](diffhunk://#diff-576d76677bb1f209c474d33ac5b8abe5d0924e3ff5547e3dc046a5a8e7efd118L80-R82) [[2]](diffhunk://#diff-576d76677bb1f209c474d33ac5b8abe5d0924e3ff5547e3dc046a5a8e7efd118L106-R108)

### Code Cleanup:

* [`handlers/mparticle-api/src/index.ts`](diffhunk://#diff-5337205cb740a327a40dd884da9e0ebb8b16e47175c40ba890d8e477436e8501L58-R60): Removed the logic for dynamically constructing the `lambdaDomainUrl` from request headers and simplified the call to `submitDataSubjectRequest` by eliminating the `lambdaDomainUrl` argument.

### DIFF
--- a/handlers/mparticle-api/src/apis/data-subject-requests.ts
+++ b/handlers/mparticle-api/src/apis/data-subject-requests.ts
@@ -77,10 +77,9 @@ async function requestDataSubjectApi<T>(url: string, options: {
  * The OpenDSR Request takes a JSON request body and requires a Content-Type: application/json header.
  * https://docs.mparticle.com/developers/apis/dsr-api/v3/#submit-a-data-subject-request-dsr
  * @param {DataSubjectRequestForm} form - The form containing the data subject request details.
- * @param {string} lambdaDomainUrl - Current running lambda domain url
  * @returns https://docs.mparticle.com/developers/apis/dsr-api/v3/#example-success-response-body
  */
-export const submitDataSubjectRequest = async (form: DataSubjectRequestForm, lambdaDomainUrl: string): Promise<DataSubjectRequestSubmission> => {
+export const submitDataSubjectRequest = async (form: DataSubjectRequestForm): Promise<DataSubjectRequestSubmission> => {
     const response = await requestDataSubjectApi<{
         expected_completion_time: Date;
         received_time: Date;
@@ -103,7 +102,10 @@ export const submitDataSubjectRequest = async (form: DataSubjectRequestForm, lam
             },
             api_version: "3.0",
             status_callback_urls: [
-                `${lambdaDomainUrl}/data-subject-requests/${form.requestId}/callback`
+                `${getSecretValue(
+                    `/mparticle-api/${stageFromEnvironment()}/api-gateway-url`,
+                    'MPARTICLE_API_GATEWAY_URL'
+                )}/data-subject-requests/${form.requestId}/callback`
             ],
             group_id: form.userId, // Let's group by User Unique Id to group all requests related to that user (max 150 requests per group)
         }
--- a/handlers/mparticle-api/src/index.ts
+++ b/handlers/mparticle-api/src/index.ts
@@ -55,15 +55,9 @@ const router = new Router([
                 console.warn("It was not possible to set the User Attribute to remove user from audiences or from event forwarding during the waiting period.", error)
             }
 
-            // Request for Erasure
-            const domain = event.headers['host'] ?? 'code.dev-theguardian.com'; // e.g., "abc123.lambda-url.region.on.aws" or API Gateway domain
-            const protocol = event.headers['x-forwarded-proto'] ?? 'https';
-            const lambdaDomainUrl = `${protocol}://${domain}`;
-            const requestForErasureResult = await submitDataSubjectRequest(parsed.body, lambdaDomainUrl)
-
             return {
                 statusCode: 201,
-                body: JSON.stringify(requestForErasureResult),
+                body: JSON.stringify(await submitDataSubjectRequest(parsed.body)),
             };
         }),
         parser: {


handlers/mparticle-api/src/apis/data-subject-requests.ts
```diff
@@ -103,7 +102,10 @@ export const submitDataSubjectRequest = async (form: DataSubjectRequestForm, lam
             },
             api_version: "3.0",
             status_callback_urls: [
-                `${lambdaDomainUrl}/data-subject-requests/${form.requestId}/callback`
+                `${getSecretValue(
```
johnduffell, line 105: is it secret? It's going to be something like mparticle-api.support.guardianapis.com or mparticle-api.support.code.dev-guardianapis.com, so we could just use an if statement based on the stage?  Then it's more easy to review, and we get a better change history and searcahbility in github.

andresilva-guardian: I made it a parameter because at this specific moment in time I do not know which value will be here. So, to unblock i'm assuming that I can deploy/close this now and then fill the parameter accordingly for now.

johnduffell: ok yes I think fill in accordingly with a TODO for now and once we know we can do another PR.

handlers/mparticle-api/src/apis/data-subject-requests.ts
```diff
@@ -103,7 +102,10 @@ export const submitDataSubjectRequest = async (form: DataSubjectRequestForm, lam
...
+                    `/mparticle-api/${stageFromEnvironment()}/api-gateway-url`,
+                    'MPARTICLE_API_GATEWAY_URL'
+                )}/data-subject-requests/${form.requestId}/callback`
```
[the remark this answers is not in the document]
andresilva-guardian: It's done @johnduffell please take another look at it, thank you.

cdk/lib/mparticle-api.ts
```diff
@@ -98,6 +99,13 @@ export class MParticleApi extends GuStack {
 			}),
 		);
 
+		new SrLambdaDomain(this, {
+			subdomain: 'mparticle-api',
+			stage: this.stage,
+			restApi: apiGateway.api,
+			apiDomain: 'membership',
+		});
```
johnduffell, line 107: nice I think this is great and simple

cdk/lib/cdk/sr-lambda-domain.ts
```diff
@@ -0,0 +1,65 @@
...
+				apiDomain = 'support.guardianapis.com';
+				certificateId = 'b384a6a0-2f54-4874-b99b-96eeff96c009';
+				break;
+			case 'membership':
+				apiDomain = 'membership.guardianapis.com';
+				certificateId = 'c1efc564-9ff8-4a03-be48-d1990a3d79d2';
+				break;
+		}
+
+		const domainNameString = `${props.subdomain}-${props.stage.toLowerCase()}.${apiDomain}`;
```
johnduffell, line 41: if the stage is CODE then I think it woud need to use the domain as code.support.dev-guardianapis.com

johnduffell: once we get an answer on devs channel we can update it.  this won't block the PR as it's only for CODE and we can change it later https://chat.google.com/room/AAAA6HaLTYQ/8RMVN9cSgXs/8RMVN9cSgXs?cls=10

johnduffell: The outcome is that we need to use this style
my-api-code.support.guardianapis.com`

handlers/mparticle-api/src/apis/data-subject-requests.ts
```diff
@@ -103,9 +74,14 @@ export const submitDataSubjectRequest = async (form: DataSubjectRequestForm, lam
             },
             api_version: "3.0",
             status_callback_urls: [
-                `${lambdaDomainUrl}/data-subject-requests/${form.requestId}/callback`
+                `https://mparticle-api-${getEnv('STAGE').toLowerCase()}.membership.guardianapis.com/data-subject-requests/${form.requestId}/callback`
```
johnduffell, line 77: support.guardianapis.com is the one we use for new things.  Also we only include `-stage` in the url if it's not PROD

cdk/lib/cdk/sr-lambda-domain.ts
```diff
@@ -0,0 +1,65 @@
+import type { GuStack } from '@guardian/cdk/lib/constructs/core';
+import type { IRestApi } from 'aws-cdk-lib/aws-apigateway';
+import { CfnBasePathMapping, CfnDomainName } from 'aws-cdk-lib/aws-apigateway';
+import { CfnRecordSet } from 'aws-cdk-lib/aws-route53';
+import { Construct } from 'constructs';
+
+interface GuDomainProps {
+	subdomain: string; // e.g. "metric-push-api", "mparticle-api"
+	stage: string;
```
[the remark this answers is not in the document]
andresilva-guardian: Done, thank you @johnduffell

cdk/lib/cdk/sr-lambda-domain.ts
andresilva-guardian: Done, thank you @johnduffell

handlers/mparticle-api/src/apis/data-subject-requests.ts
```diff
@@ -202,20 +178,18 @@ export const processDataSubjectRequestCallback = async (requestId: string, paylo
...
     });
     console.debug(
-        `Sending message ${JSON.stringify(message)} to queue '${queueName}'`,
+        `Sending message ${JSON.stringify(message)} to Ophan queue`,
     );
+
+    const appConfig = await getAppConfig();
     const command = new SendMessageCommand({
-        QueueUrl: queueName,
+        QueueUrl: appConfig.ophanErasureQueueUrl,
```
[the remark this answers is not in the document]
andresilva-guardian: The are the URLS:
- https://sqs.eu-west-1.amazonaws.com/021353022223/ophan-data-lake-CODE-erasure-Queue-GRLOB6EAD0O9.fifo
-https://sqs.eu-west-1.amazonaws.com/021353022223/ophan-data-lake-PROD-erasure-Queue-1H020S409D2OY.fifo

Because of the "GRLOB6EAD0O9" and the "1H020S409D2OY" I don't feel safe trying to generate the url locally.

handlers/mparticle-api/src/apis/data-subject-requests.ts
andresilva-guardian: Done, thank you @johnduffell

handlers/mparticle-api/src/config.ts
```diff
@@ -0,0 +1,31 @@
...
+        secret: z.string(),
+    }),
+    inputPlatform: z.object({
+        key: z.string(),
+        secret: z.string(),
+    }),
+    ophanErasureQueueUrl: z.string(),
+    pod: z.string(),
+});
+export type AppConfig = z.infer<typeof appConfigSchema>;
```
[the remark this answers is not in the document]
andresilva-guardian: Done, thank you @johnduffell

handlers/mparticle-api/src/apis/data-subject-requests.ts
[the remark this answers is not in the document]
johnduffell: oh I see. In fact you can actually just put in the queue name as the queue url, and it does work, although it logs an exception.  Your way of using the proper URL makes more sense rather than relying on that or doing an API call to look it up.

To be honest if that's the two URLs I'd just do 
```
const funnyNumberForStage = stage === "PROD" ? "1H020S409D2OY" : "GRLOB6EAD0O9"
const queueUrl = `https://sqs.eu-west-1.amazonaws.com/021353022223/ophan-data-lake-${stage}-erasure-Queue-${funnyNumberForStage}.fifo`
```

cdk/lib/mparticle-api.ts
```diff
@@ -98,16 +117,15 @@ export class MParticleApi extends GuStack {
...
-		// 		],
-		// 	}),
-		// );
+		new SrLambdaDomain(this, {
+			subdomain: 'mparticle-api',
+			restApi: apiGateway.api,
+		});
+
+		new CfnOutput(this, 'MParticleLambdaRoleArn', {
+			value: mparticleLambdaRole.roleArn,
```
johnduffell, line 126: ah now I see why you made an explicit role - but the default should be available like this
```suggestion
			value: lambda.role.roleArn,
```

cdk/lib/mparticle-api.ts
```diff
@@ -98,16 +117,15 @@ export class MParticleApi extends GuStack {
...
+			description: 'ARN of the mParticle Lambda execution role',
+			exportName: `${app}-${this.stage}-lambda-role-arn`,
```
johnduffell, line 128: unless the exports are global you can export it without the extra info
```suggestion
			exportName: `lambda-role-arn`,
```

cdk/lib/mparticle-api.ts
```diff
@@ -98,16 +117,15 @@ export class MParticleApi extends GuStack {
...
+		});
```
johnduffell, line 129: could do with a comment here (ideally in the CDK) explaining why this needs to be exported.  I guess something is going to come along and add a separate permission so that the lambda can write to a queue?

handlers/mparticle-api/src/apis/data-subject-requests.ts
```diff
@@ -94,18 +66,24 @@ export const submitDataSubjectRequest = async (form: DataSubjectRequestForm, lam
...
             subject_identities: {
                 "controller_customer_id": {
                     value: form.userId,
                     encoding: 'raw',
                 }
             },
             api_version: "3.0",
-            status_callback_urls: [
-                `${lambdaDomainUrl}/data-subject-requests/${form.requestId}/callback`
+            status_callback_urls: getEnv('STAGE').toLowerCase() === "code" ? [
```
johnduffell, line 76: nit pick: this is set by CDK and will always be upper case, no need for toLowerCase

important: I might have suggested this the wrong way around, but we need to check PROD vs non-PROD rather than CODE vs non-CODE
This is because if we run it as TEST or DEV, it won't affect PROD

andresilva-guardian: Done @johnduffell thank you.

handlers/mparticle-api/runManual/runLoadConfig.ts
```diff
@@ -0,0 +1,8 @@
+import { loadConfig } from '@modules/aws/appConfig';
+import { ConfigSchema } from '../src/config';
+
+// run this to check the CODE and PROD config are readable and look right
+
+loadConfig('CODE', 'membership', 'mparticle-api', ConfigSchema).then(console.log);
+
+loadConfig('PROD', 'membership', 'mparticle-api', ConfigSchema).then(console.log);
```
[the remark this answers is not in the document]
andresilva-guardian: Done @johnduffell . I had to change the work done by @michaelbjacobson and move the lambda and it's params config to the "support" stack, but it is done and completed. Thank you.

cdk/lib/cdk/sr-lambda-domain.ts
```diff
@@ -0,0 +1,58 @@
...
+export class SrLambdaDomain extends Construct {
+	constructor(scope: GuStack, props: GuDomainProps) {
+		const safeId = generateSafeId(props.subdomain);
+		super(scope, `${safeId}Domain`);
+
+		const apiDomain = 'support.guardianapis.com';
+		const domainNameString =
+			scope.stage.toLowerCase() === 'code'
+				? `${props.subdomain}-code.${apiDomain}`
+				: `${props.subdomain}.${apiDomain}`;
```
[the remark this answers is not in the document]
andresilva-guardian: Done @johnduffell , thank you.

cdk/lib/mparticle-api.ts
andresilva-guardian: Changed @johnduffell , thank you.

cdk/lib/mparticle-api.ts
andresilva-guardian: @johnduffell I could be mistaken but I believe that there is a problem with your suggestion of using a generic export name. CloudFormation export names must be globally unique within an AWS region. If we use just lambda-role-arn, we'll run into issues:
- Naming conflicts: If we have multiple stacks or deploy to multiple stages, they'll all try to create the same export name
- Deployment failures: CloudFormation will fail with "Export already exists" errors
- Cross-stage conflicts: Your PROD and CODE stages would conflict with each other

Should we still try this?

cdk/lib/mparticle-api.ts
andresilva-guardian: Done @johnduffell , thank you.

cdk/lib/mparticle-api.ts
johnduffell: ok that fails my condition "unless the exports are global" which I did not check.

handlers/mparticle-api/src/index.ts
```diff
@@ -96,7 +90,8 @@ const router = new Router([
         httpMethod: 'POST',
         path: '/data-subject-requests/{requestId}/callback',
         handler: routerHandler(async (event, parsed) => {
-            const callbackValidationResult = await validateDataSubjectRequestCallback(event.headers['x-opendsr-processor-domain'], event.headers['x-opendsr-signature'], event.body);
+            const getHeader = (key: string) => Object.entries(event.headers).find(([k]) => k.toLowerCase() === key.toLowerCase())?.[1];
+            const callbackValidationResult = await validateDataSubjectRequestCallback(getHeader('x-opendsr-processor-domain'), getHeader('x-opendsr-signature'), event.body);
```
johnduffell, line 94: nit pick/optional:
I guess this change is because header names are not case sensitive - It is a bit harder to follow the `k` vs `key` and the `?.[1]`, I think it'll be slightly clearer to preprocess them and keep the later logic the same, what do you think?
```suggestion
            const headers = mapValues(event.headers, v => v.toLowerCase());
            const callbackValidationResult = await validateDataSubjectRequestCallback(headers['x-opendsr-processor-domain'], headers['x-opendsr-signature'], event.body);
```
and you can import mapValues from arrayFunctions.ts

johnduffell (approved):
looks good to me, thanks for bearing with my lack of understanding of this area! 👍

PR #2920, opened 2025-07-04
Redesign mParticle API as a Baton Right to Erasure Data Provider

andresilva-guardian:
After reaching the final stage of code review of the Baton Integration with mParticle plan, some concerns were raised by the team that some of the code being worked on might not be in an up-to-date state of maintenance. So the strategy initially accepted was rethought with the least possible impact to divert the implementation directly to the Baton application. This way, there will be a new button in the Baton application that will allow the user's information to be deleted from the CDP (mParticle).

The goal of this pull request is to redesign the `mparticle-api` to work as a Baton Right to Erasure Data Provider. The most important changes include splitting the single Lambda into dual handlers (HTTP and Baton), implementing Baton RER endpoints, updating infrastructure with dedicated IAM roles, removing unused code and dependencies, and adding comprehensive test coverage for the new architecture.

The context for this change can be found on this [Trello card](https://trello.com/c/ytan5gvJ/863-redesign-mparticle-api-as-a-baton-rer-provider).

### Architecture Overview

The new architecture implements a **dual handler approach** within a single Lambda deployment, where:

- **HTTP Handler** (`handlerHttp`): Continues to handle mParticle callbacks and HTTP API endpoints
- **Baton Handler** (`handlerBaton`): Handles Baton Right to Erasure (RER) requests using a direct invocation pattern

This design enables Baton to invoke the Lambda directly for RER operations while maintaining all existing HTTP functionality for mParticle integrations.

```mermaid
sequenceDiagram
    participant B as Baton App
    participant BL as mParticle Baton Lambda
    participant MP as mParticle API
    participant HL as mParticle HTTP Lambda
    
    Note over B,HL: Initiate Erasure Request
    B->>BL: Invoke Lambda (RER initiate)
    BL->>MP: Submit DSR (erasure)
    BL->>MP: Set user attributes (opt-out)
    MP-->>BL: Return request ID & completion time
    BL-->>B: Return initiation reference
    
    Note over B,HL: Status Check
    B->>BL: Invoke Lambda (RER status)
    BL->>MP: Query DSR status
    MP-->>BL: Return current status
    BL-->>B: Return status (pending/completed/failed)
    
    Note over B,HL: Callback Processing (unchanged)
    MP->>HL: HTTP callback
    HL->>HL: Process callback
    HL-->>MP: HTTP response
```

### What does this change?

This PR redesigns the `mparticle-api` Lambda to act as a Baton-compatible Right to Erasure (RER) Data Provider, implementing a dual-handler architecture within a single deployment. The main changes are:

#### **🔄 Dual Handler Architecture**
- **Splits functionality** into two handlers within index.ts:
  - `handlerHttp`: Processes API Gateway events for mParticle callbacks and HTTP endpoints
  - `handlerBaton`: Processes direct Lambda invocations from Baton for RER operations
- **Introduces router separation**:
  - http.ts: Handles HTTP routes (moved from original index.ts)
  - baton.ts: **New** - Handles Baton RER requests with `initiate` and `status` actions

#### **🎯 Baton RER Integration**
Implements the [Baton Data Provider pattern](https://github.com/guardian/baton?tab=readme-ov-file#adding-data-sources-to-baton) with:

- **Initiate Endpoint**: Creates new erasure requests in mParticle
  ```typescript
  {
    requestType: 'RER',
    action: 'initiate',
    subjectId: string,
    subjectEmail?: string,
    dataProvider: 'mparticlerer'
  }
  ```
- **Status Endpoint**: Queries existing request status
  ```typescript
  {
    requestType: 'RER', 
    action: 'status',
    initiationReference: GUID
  }
  ```
- **Strong Validation**: Uses [Zod](https://zod.dev/) schemas for request/response validation
- **User Opt-out**: Automatically sets mParticle user attributes to [remove users from audiences during the waiting period](https://docs.mparticle.com/guides/data-subject-requests/#erasure-request-waiting-period)

#### **🏗️ Infrastructure Updates**
Updates CDK infrastructure in mparticle-api.ts:

- **Separate Lambda Functions**:
  - `mparticle-api-http-{STAGE}`: Handles HTTP/API Gateway traffic  
  - `mparticle-api-baton-{STAGE}`: Handles Baton RER invocations
- **New IAM Role**: Creates `baton-mparticle-lambda-role-{STAGE}` for cross-account access
  - Replaces previous Lambda role ARN export approach
  - Grants `lambda:InvokeFunction` permission only to the Baton Lambda
  - Follows [Baton integration best practices](https://github.com/guardian/baton-lambda-template/blob/61ebdec91bd53e218d5f5a2aa90494db69adfa0a/cfn.yaml#L44-L46)
- **Enhanced Monitoring**: Adds separate CloudWatch alarms for each Lambda function

#### **🧹 Code Organization & Cleanup**
- **File Restructuring**:
  - `src/config.ts` → `src/utils/config.ts`
  - http.ts → `src/utils/make-http-request.ts`  
  - `src/run.ts` → `src/utils/invoke-http-handler.ts`
- **New Utilities**:
  - invoke-baton-rer-handler.ts: Test utility for Baton handler invocation
- **Removed Files**:
  - `run-local.ts`: Local development script (replaced by manual test utilities)
  - Large portions of original index.ts (refactored into routers)

#### **📦 Dependency Management**
Updates package.json:
- **Removed**: Unused dependencies identified during cleanup
- **Maintained**: Core dependencies (`@peculiar/x509`, `zod`)
- **Enhanced**: Development dependencies for improved testing (`@faker-js/faker`)

#### **🧪 Testing Improvements**
- **New Test Architecture**:
  - baton.spec.ts: **New** - Comprehensive Baton RER functionality tests
  - `test/http.spec.ts`: **New** - HTTP endpoint tests (extracted from `api.spec.ts`)
  - Removed `test/api.spec.ts`: Replaced by modular test files
- **Enhanced Test Coverage**:
  - Baton request validation and error handling
  - mParticle API integration mocking
  - Status mapping and response validation

#### **Unit Tests**
```bash
# Run all tests for the mparticle-api
pnpm --filter mparticle-api test

# Run only Baton-specific tests  
pnpm --filter mparticle-api test -- test/baton.spec.ts

# Run only HTTP endpoint tests
pnpm --filter mparticle-api test -- test/http.spec.ts
```

#### **Manual Testing**
1. **Baton Integration Testing**:
   - Deploy the Lambda functions to a test environment
   - Configure Baton to use the new `baton-mparticle-lambda-role-{STAGE}` 
   - Test initiation and status checking through the Baton UI
   - Verify user opt-out attributes are set in mParticle

2. **HTTP Endpoint Testing**:
   - Test mParticle callback processing via API Gateway
   - Verify existing HTTP endpoints continue to function
   - Confirm CloudWatch alarms trigger appropriately

#### **Integration Testing**
- **Baton → Lambda**: Verify cross-account role assumption and function invocation
- **Lambda → mParticle**: Confirm DSR submission and status querying
- **mParticle → Lambda**: Test callback processing through API Gateway

- ✅ **Baton Integration**: Users can successfully initiate and track erasure requests in mParticle via the new Baton UI button ([Guardian Baton PR #391](https://github.com/guardian/baton/pull/391))
- ✅ **Backward Compatibility**: All existing mParticle callbacks and HTTP endpoints continue to work without changes
- ✅ **Monitoring**: CloudWatch alarms properly detect errors in both HTTP and Baton handlers  
- ✅ **Test Coverage**: All tests pass with comprehensive coverage of new Baton functionality
- ✅ **Clean Architecture**: Code is well-organized, maintainable, and free of unused dependencies
- ✅ **Security**: IAM roles follow least-privilege principles with appropriate cross-account access

#### **🔒 Security & Access Control**
- **IAM Role Configuration**: The new `baton-mparticle-lambda-role-{STAGE}` must be correctly configured with proper trust relationships and minimal permissions
- **Cross-Account Access**: Ensure Baton account can assume the role but no other unauthorized access is possible
- **Function Separation**: Only the Baton Lambda should be invokable via the cross-account role, not the HTTP Lambda

#### **🔄 Dual Handler Complexity** 
- **Event Routing**: The dual handler approach must correctly route Baton RER events vs API Gateway events
- **Error Handling**: Different error handling strategies for direct invocation (Baton) vs HTTP responses (API Gateway)
- **Monitoring**: Separate monitoring for each handler to avoid confusion during incident response

#### **📊 mParticle Integration**
- **API Compatibility**: Changes to mParticle DSR API structure or behavior could affect both handlers
- **Rate Limiting**: Increased traffic from Baton requests may impact mParticle API quotas
- **User Attribute Setting**: Failure to set opt-out attributes could result in continued data processing during erasure waiting period

#### **🚀 Deployment & Rollback**
- **Infrastructure Changes**: CDK changes include new Lambda functions and IAM roles that need careful deployment sequencing
- **Baton Dependency**: Baton must be updated to use the new role before old infrastructure can be safely removed
- **Rollback Strategy**: Plan for reverting to single-handler architecture if critical issues arise

#### **🧪 Testing Coverage**
- **Manual Testing Burden**: Dual architecture requires testing both HTTP and direct invocation patterns
- **Integration Testing**: Cross-account testing requires coordination between Baton and support-service-lambdas teams
- **Regression Testing**: Ensure no existing functionality is broken during the transition

#### **Related Documentation**
- [mParticle Data Subject Requests Guide](https://docs.mparticle.com/guides/data-subject-requests/) - Understanding erasure request waiting periods and user opt-out strategies
- [mParticle DSR API v3](https://docs.mparticle.com/developers/apis/dsr-api/v3/) - Technical API documentation for DSR operations
- [OpenDSR Framework](https://github.com/opengdpr/OpenDSR) - Standard format for data subject requests
- [Baton Lambda Integration Template](https://github.com/guardian/baton-lambda-template) - Reference implementation for Baton data providers

#### **Key Code Examples**

**Baton RER Request Format:**
```typescript
// Initiate erasure request
{
  requestType: 'RER',
  action: 'initiate', 
  subjectId: 'user@example.com',
  dataProvider: 'mparticlerer'
}

// Check status
{
  requestType: 'RER',
  action: 'status',
  initiationReference: '550e8400-e29b-41d4-a716-446655440000'
}
```

**Infrastructure Role Definition:**
```typescript
const batonInvokeRole = new Role(this, 'BatonInvokeRole', {
  roleName: `baton-mparticle-lambda-role-${this.stage}`,
  assumedBy: new AccountPrincipal(batonAccountId),
});
batonInvokeRole.attachInlinePolicy(
  new Policy(this, 'BatonRunLambdaPolicy', {
    statements: [
      new PolicyStatement({
        effect: Effect.ALLOW,
        actions: ['lambda:InvokeFunction'],
        resources: [batonLambda.functionArn], // Only Baton Lambda
      }),
    ],
  }),
);
```

### Images

N/A – This is a backend/API change only. The user-facing changes will be visible in the Baton application UI as a new erasure request button.

### Accessibility  

N/A – No user-facing UI is affected by this change. The Baton application changes are handled separately in [Guardian Baton PR #391](https://github.com/guardian/baton/pull/391).


handlers/mparticle-api/src/index.ts
```diff
@@ -1,164 +1,84 @@
...
+	} else if (isBatonRerEvent(event)) {
+		console.debug('Processing as Baton RER event');
+		return handleBatonRerEvent(event, context);
+	} else {
+		// Handle other event types or unknown events
+		console.error('Unknown event type:', event);
+
+		// If it might be an HTTP event without proper structure, try to handle gracefully
+		if (event.httpMethod || event.path) {
+			console.log('Attempting to process as malformed API Gateway event');
```
MaelGNM, line 75: I think this log line is a little misleading, we're not actually processing the event but returning a 400. It might be better to log the event for debugging purposes.

MaelGNM: I spent too long looking at these lines and forgot the event is already logged just above

handlers/mparticle-api/test/api.spec.ts
```diff
@@ -1,237 +1,214 @@
...
+
+		const result = await invokeHttpHandler({
+			httpMethod: 'POST',
+			path: `/data-subject-requests/${requestId}/callback`,
+			// Do not fake it to match the header signature
+			body: JSON.stringify({
+				controller_id: '1402',
+				expected_completion_time: '2025-06-09T00:00:00Z',
+				status_callback_url:
+					'https://webhook.site/6dfd0447-e1f9-4a98-a391-25481898763b',
```
MaelGNM, line 190: Is this URL meaningful here?

handlers/mparticle-api/src/apis/data-subject-requests.ts
```diff
@@ -148,63 +162,23 @@ export const getStatusOfDataSubjectRequest = async (requestId: string): Promise<
...
+	requestId: string,
+	payload: DataSubjectRequestCallback,
+): {
+	message: string;
+	timestamp: Date;
+} => {
+	// Just log this information so we can have track of it on Cloud Watch
+	console.info('Process Data Subject Request Callback from mParticle', {
+		requestId,
+		form: payload,
```
MaelGNM, line 175: Just a question: do we think it would be better not to log the whole payload so as to not include the user ID? I'm thinking that once the callback comes saying the data is deleted, there would technically still be references to the user ID in the logs here.

handlers/mparticle-api/src/index.ts
andresilva-guardian: Thanks @MaelGNM I changed the log message a bit to improve clarity :)

johnduffell:
> Baton users can successfully initiate and track erasure requests in mParticle via the new UI button.

is there a linked PR as I am sure this one can't add that button?

cdk/lib/mparticle-api.ts
```diff
@@ -105,15 +111,21 @@ export class MParticleApi extends GuStack {
...
-		/**
-		 * Export Lambda role ARN for cross-account queue access.
-		 * The SQS queue policy in account "Ophan" imports this ARN
-		 * to grant this Lambda sqs:SendMessage permissions to the erasure queue
-		 */
-		new CfnOutput(this, 'MParticleLambdaRoleArn', {
-			value: lambda.role!.roleArn,
-			description: 'ARN of the mParticle Lambda execution role',
-			exportName: `${app}-${this.stage}-lambda-role-arn`,
+		// Add Baton invoke role
```
johnduffell, line 114: this comment just repeats what the code says, it's worth adding a link to where it's used or suitable docs.  Or delete it if appropriate

handlers/mparticle-api/package.json
```diff
@@ -1,25 +1,23 @@
...
+    "test": "jest --group=-integration --silent",
     "it-test": "jest --group=integration",
     "type-check": "tsc --noEmit",
     "build": "esbuild --bundle --platform=node --target=node18 --outfile=target/index.js src/index.ts",
     "lint": "eslint src/**/*.ts test/**/*.ts",
     "package": "pnpm type-check && pnpm lint && pnpm check-formatting && pnpm test && pnpm build && cd target && zip -qr mparticle-api.zip ./*.js",
     "check-formatting": "prettier --check **.ts",
     "fix-formatting": "prettier --write **.ts",
-    "run-local": "esbuild --bundle --platform=node --target=node18 --outfile=target/run-local.js run-local.ts && node target/run-local.js"
+    "check-config": "ts-node runManual/runLoadConfig.ts"
```
johnduffell, line 12: nice idea to put it here, I didn't do that in mine (partly because I wasn't sure how to do it)

handlers/mparticle-api/src/apis/data-subject-requests.ts
```diff
@@ -1,47 +1,51 @@
-import { SendMessageCommand, SQSClient } from '@aws-sdk/client-sqs';
 import type { DataSubjectRequestCallback } from '../../interfaces/data-subject-request-callback';
```
johnduffell, line 1: it's hard to pick out the formatting changes from the code changes - worth to comment on any lines that have had changes to make it easier to review.
If it was my PR I'd be strongly tempted to extract a first PR to fix the formatting everywhere, and then branch this one off that one, to make it easier for people to review the real changes.

handlers/mparticle-api/src/index.ts
```diff
@@ -1,164 +1,84 @@
...
+		console.error('Unknown event type:', event);
+
+		// If it might be an HTTP event without proper structure, try to handle gracefully
+		if (event.httpMethod || event.path) {
+			console.log('Attempted to process as malformed API Gateway event');
+			return {
+				statusCode: 400,
+				body: JSON.stringify({ error: 'Malformed API Gateway event' }),
+			};
+		}
```
johnduffell, line 80: make sure it returns a 500 or throws an exception as we want to receive an alarm in this case either from failed lambda or 5xx error.  I would personally just remove this block completely and let it throw an exception if there's any issue with the request.

handlers/mparticle-api/src/index.ts
```diff
@@ -1,164 +1,84 @@
...
-            if (!callbackValidationResult) {
-                return {
-                    statusCode: 401,
-                    body: 'Data Subject Request Callback validation failed.',
-                };
-            }
+// HTTP handler function
+async function handleHttpRequest(
+	event: APIGatewayProxyEvent,
+	context: Context,
```
johnduffell, line 33: nitpick: context I think it's unused

handlers/mparticle-api/src/routers/baton.ts
```diff
@@ -0,0 +1,134 @@
...
+import { DataSubjectRequestStatus } from '../../interfaces/data-subject-request-state';
+import type { DataSubjectRequestSubmission } from '../../interfaces/data-subject-request-submission';
+import {
+	getStatusOfDataSubjectRequest,
+	submitDataSubjectRequest,
+} from '../apis/data-subject-requests';
+import { setUserAttributesForRightToErasureRequest } from '../apis/events';
+import { getEnv } from '../utils/config';
+
+// https://github.com/guardian/baton/blob/1037c63c9bd782aed514bf6aaa38a54dabf699eb/README.md
```
johnduffell, line 12: very handy comment.  Could be useful to reference the correct section.  And mention what they will see . as follows
```suggestion
// baton data structures are described here
// https://github.com/guardian/baton/blob/1037c63c9bd782aed514bf6aaa38a54dabf699eb/README.md#general-request--response-fields
```

handlers/mparticle-api/src/routers/baton.ts
```diff
@@ -0,0 +1,134 @@
...
+interface BatonRerEventRequestBase {
+	requestType: 'RER';
+}
+
+interface BatonRerEventInitiateRequest extends BatonRerEventRequestBase {
```
johnduffell, line 17: optional: easy to look at the SAR instead of the RER one (like I just did!) so another link might help
```suggestion
// data structure https://github.com/guardian/baton/blob/1037c63c9bd782aed514bf6aaa38a54dabf699eb/README.md#rer-initiate-call
interface BatonRerEventInitiateRequest extends BatonRerEventRequestBase {
```

andresilva-guardian:
> > Baton users can successfully initiate and track erasure requests in mParticle via the new UI button.
> 
> is there a linked PR as I am sure this one can't add that button?

You were right @johnduffell thanks for catching it, the pull request is this [one](https://github.com/guardian/baton/pull/391)

handlers/mparticle-api/src/routers/baton.ts
```diff
@@ -0,0 +1,134 @@
...
+		await setUserAttributesForRightToErasureRequest(
+			environment,
+			request.subjectId,
+			submittedTime,
+		);
+	} catch (error) {
+		console.warn(
+			'It was not possible to set the User Attribute to remove user from audiences or from event forwarding during the waiting period.',
+			error,
+		);
```
johnduffell, line 74: I assume we don't want to fail the lambda in this case.
recommend: if this would be a problem if it was happening on every request, we need to push a custom metric and set up an alarm on this situation

handlers/mparticle-api/package.json
```diff
@@ -1,25 +1,23 @@
 {
   "name": "mparticle-api",
   "scripts": {
-    "test": "jest --group=-integration",
+    "test": "jest --group=-integration --silent",
```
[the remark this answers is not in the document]
andresilva-guardian: Removes "console.logs" from the test execution @johnduffell. I would say that, we only really expect tests to fail during test driven development, but "99%" of the times that these tests will run they will pass. For that reason I kept this arg here to make it simpler and easier to read. But at the same time, it is different from other apps, so I shall remove it. What do you think?

handlers/mparticle-api/src/routers/baton.ts
```diff
@@ -0,0 +1,134 @@
...
+			requestType: 'erasure',
+			submittedTime,
+			userId: request.subjectId,
+			environment,
+		});
+	return {
+		requestType: 'RER',
+		action: 'initiate',
+		status: 'pending',
+		initiationReference: dataSubjectRequestSubmission.requestId,
```
johnduffell, line 91: we generate a uuid and send to both mparticle and baton in order to tie the request together later.
It looks like the ref has to be valid json
https://github.com/guardian/baton/blob/1037c63c9bd782aed514bf6aaa38a54dabf699eb/README.md#rer-initiate-call:~:text=%22initiationReference%22%3A%20%22%7BAny_Valid_Json%7D%22%2C
so when passing back to baton I think it woudl need wrapping in double quotes
```suggestion
		initiationReference: '"' + dataSubjectRequestSubmission.requestId + '"',
```

handlers/mparticle-api/interfaces/data-subject-request-callback.ts
```diff
@@ -4,74 +4,77 @@
  * NOTE: This syntax is defined by mParticle convention since it's a payload from a callback request (webhook)
  */
 export interface DataSubjectRequestCallback {
```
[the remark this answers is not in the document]
andresilva-guardian: Now formatting seems to be working as expected @johnduffell . It was not working on the first PR (not sure why) but currently everything seems correctly defined between my machine and the CI config.

cdk/lib/mparticle-api.ts
```diff
@@ -105,15 +111,21 @@ export class MParticleApi extends GuStack {
...
+		const batonInvokeRole = new Role(this, 'BatonInvokeRole', {
+			roleName: `baton-mparticle-lambda-role-${this.stage}`,
+			assumedBy: new AccountPrincipal('029312801662'),
```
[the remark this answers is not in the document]
andresilva-guardian: Agreed @johnduffell , removed.

cdk/lib/mparticle-api.ts
andresilva-guardian: Agreed @johnduffell , removed.

handlers/mparticle-api/src/routers/baton.ts
```diff
@@ -0,0 +1,134 @@
...
+		);
+	} catch (error) {
+		console.warn(
+			'It was not possible to set the User Attribute to remove user from audiences or from event forwarding during the waiting period.',
+			error,
+		);
+	}
+
+	// Submit
+	const dataSubjectRequestSubmission: DataSubjectRequestSubmission =
```
johnduffell, line 78: this is actually the mparticle response I think?  "Submission" makes it sound like the request.  It may be that mparticle uses confusing terminology, but if so it would be worth `DataSubjectRequestSubmissionResponse` at the minimum

cdk/lib/mparticle-api.ts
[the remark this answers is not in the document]
andresilva-guardian: @johnduffell This account is the "Baton" aws profile. This role was created to comply with those same instructions that you linked and you can see them on this other [PR](https://github.com/guardian/baton/pull/391/files#diff-510aa7616a9bb7ec060760b1a8297e58cfa324d546eb6f1451ce154b0c324c60). Thanks!

handlers/mparticle-api/src/index.ts
```diff
@@ -1,164 +1,84 @@
...
-                    },
-                    environment: parsed.body.environment
-                });
-            } catch (error) {
-                console.warn("It was not possible to set the User Attribute to remove user from audiences or from event forwarding during the waiting period.", error)
-            }
+// Type guard to check if event is Baton RER
+function isBatonRerEvent(event: any): event is BatonRerEventRequest {
+	return (
+		event && event.requestType && event.requestType.toUpperCase() === 'RER'
```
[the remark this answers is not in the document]
andresilva-guardian: Thanks @johnduffell I usually do these types of things a lot to minimize impact of possible future changes. Imagine that, for some reason, someone starts using "rer" instead of "RER" - with this, no breaking change would happen. But sure, it is not needed right now, so I removed it. 💪

handlers/mparticle-api/src/index.ts
```diff
@@ -1,164 +1,84 @@
...
-        handler: routerHandler(async (event, parsed) => {
-            const getHeader = (key: string) => Object.entries(event.headers).find(([k]) => k.toLowerCase() === key.toLowerCase())?.[1];
-            const callbackValidationResult = await validateDataSubjectRequestCallback(getHeader('x-opendsr-processor-domain'), getHeader('x-opendsr-signature'), event.body);
-            if (!callbackValidationResult) {
-                return {
-                    statusCode: 401,
-                    body: 'Data Subject Request Callback validation failed.',
-                };
-            }
+// HTTP handler function
```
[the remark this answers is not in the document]
andresilva-guardian: Removed unnecessary comments and improve code clarity in event handlers and tests 👍

handlers/mparticle-api/src/routers/baton.ts
andresilva-guardian: Improved @johnduffell , thanks 💪

handlers/mparticle-api/src/routers/baton.ts
andresilva-guardian: Improved @johnduffell , thank you for the suggestion 👍

handlers/mparticle-api/src/routers/baton.ts
```diff
@@ -0,0 +1,134 @@
...
+// https://github.com/guardian/baton/blob/1037c63c9bd782aed514bf6aaa38a54dabf699eb/README.md
+interface BatonRerEventRequestBase {
+	requestType: 'RER';
+}
+
+interface BatonRerEventInitiateRequest extends BatonRerEventRequestBase {
+	action: 'initiate';
+	subjectId: string;
+	subjectEmail?: string;
+	dataProvider: 'mparticlerer';
```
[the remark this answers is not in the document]
andresilva-guardian: @johnduffell I believe this has to match what is defined on Baton [here](https://github.com/guardian/baton/pull/391/files#diff-1b1880a182550cba01d1388ad07895494d3361104ce0a6c61a715685152b129f). Still I added a comment mentioning this.

handlers/mparticle-api/src/routers/baton.ts
```diff
@@ -0,0 +1,134 @@
...
+interface BatonRerEventResponseBase {
+	requestType: 'RER';
+	status: 'pending' | 'completed' | 'failed';
+	message?: string;
+}
+
+export interface BatonRerEventInitiateResponse
+	extends BatonRerEventResponseBase {
+	action: 'initiate';
+	initiationReference: string;
```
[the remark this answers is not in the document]
andresilva-guardian: @johnduffell Thanks for the question! I went with the branded type approach.
Since initiationReference will be a GUID, I created:
`typescripttype GUID = string & { readonly __brand: unique symbol };
type InitiationReference = GUID;`

The main reason I chose this over plain string is that we'll likely have other GUID fields in the future (user IDs, session IDs, etc.), and this prevents accidentally mixing them up. The branded type ensures you can't just pass any string - you have to be intentional about creating GUID values.

At runtime it's still just a string, so no performance impact. But at compile time, TypeScript will catch if we accidentally pass the wrong type of identifier.

You're right that it adds some complexity, but for something as critical as reference IDs, the extra type safety seems worth it. Plus it makes the code more self-documenting.

What do you think? If it feels like overkill for our use case, we can always simplify it back to string later.

handlers/mparticle-api/src/routers/baton.ts
```diff
@@ -0,0 +1,134 @@
...
+			submittedTime,
+		);
+	} catch (error) {
+		console.warn(
+			'It was not possible to set the User Attribute to remove user from audiences or from event forwarding during the waiting period.',
+			error,
+		);
+	}
+
+	// Submit
```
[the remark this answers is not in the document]
andresilva-guardian: Done 👍

handlers/mparticle-api/interfaces/data-subject-request-callback.ts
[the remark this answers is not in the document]
johnduffell: nice ok in that case let's assume it was a glitch!

handlers/mparticle-api/package.json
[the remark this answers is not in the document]
johnduffell: oh I see, well I don't have a strong opinion but I would say it would likely surprise people in future if their console.logs aren't appearing. So I would say keep it consistent if you can.  I think if someone runs `pnpm test --silent` it might pass that param through anyway? so it's more like opt in rather than out

handlers/mparticle-api/src/apis/data-subject-requests.ts
johnduffell: no worries actually I meant you can do it retrospectively if you like, I very often end up doing it and this is my technique if it's useful:

### extract the refactoring PR
1. make sure main is merged into my current branch
2. switch back to main, and run the formatter/whatever the main changes are
3. bring up a diff of main with my branch
4. check that the diff only shows the important changes for the second PR, pull in anything else uninteresting
5. push local changes to a new branch for the refactoring PR

### rebase the original PR onto the new PR
7. switch back to the original PR branch
8. merge the refactoring branch into the original PR branch (or rebase if you prefer)
9. raise a PR on the newly created branch
10. change the original PR's base from main to the refactoring branch - this will make the refactoring changes disappear from the diff
11. make a note at the top of the original PR description to merge the refactoring first

### get the changes into PROD (order is important)
12. merge the refactoring branch to main once reviewed (this will change the base of the original PR automatically) I avoid force pushing this PR or squashing as the second PR may end up based off different commit hashes.
13. merge the original PR to main.

It does leave a cluttered history on the second PR, and I should probably do a squash merge on the second PR, but I don't usually remember.

current example is here:
pre-refactoring PR: https://github.com/guardian/support-service-lambdas/pull/2923
actual changes PR: https://github.com/guardian/support-service-lambdas/pull/2912

handlers/mparticle-api/src/routers/baton.ts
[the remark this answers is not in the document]
johnduffell: got it, that's their convention.  I don't like it expecically given how small a visual difference between "sar" and "rer" but I agree we should stick with it. 

image
I am not complaining about your use of const type rather than string, I think it's good to be as strict as possible, but I'll defer to you as typescript is not my forte at all.

handlers/mparticle-api/src/routers/baton.ts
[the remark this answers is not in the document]
johnduffell: yeah that sounds great, it's conceptually almost exactly what I'd do in scala and it's nice to learn that trick from you
(if you're interested this is how it looks in Scala 2 https://docs.scala-lang.org/overviews/core/value-classes.html )

handlers/mparticle-api/src/routers/baton.ts
johnduffell: interesting, yes you are right that it seems to use plain uuid https://github.com/guardian/support-service-lambdas/blob/03a0ce20746748759fb2adbdaec6db6dfb82e8bd/handlers/zuora-rer/src/main/scala/com/gu/zuora/rer/ZuoraRerHandler.scala#L31
In that case it would be worth making the baton docs clearer for the next person if you have a chance.

johnduffell: and to confirm your interpretation - there is a specific test to ensure that it works with a plain string

https://github.com/guardian/baton/blob/9108c899c40ffb82e5ce8eb1492d71bedf1b18d3/test/models/JsonFormatsSpec.scala#L592-L604

handlers/mparticle-api/src/routers/baton.ts
```diff
@@ -9,21 +9,65 @@ import {
...
+ * A GUID that uniquely identifies an initiation reference.
+ *
+ * This branded type ensures that only proper GUID values can be assigned
+ * to initiation references, preventing mix-ups with other string values.
+ */
+export type InitiationReference = GUID;
+
+/**
+ * Baton data structures are described here:
+ * https://github.com/guardian/baton/blob/1037c63c9bd782aed514bf6aaa38a54dabf699eb/README.md
```
johnduffell, line 51:
```suggestion
 * https://github.com/guardian/baton/blob/1037c63c9bd782aed514bf6aaa38a54dabf699eb/README.md#general-request--response-fields
```

handlers/mparticle-api/src/routers/baton.ts
```diff
@@ -0,0 +1,134 @@
...
+			userId: request.subjectId,
+			environment,
+		});
+	return {
+		requestType: 'RER',
+		action: 'initiate',
+		status: 'pending',
+		initiationReference: dataSubjectRequestSubmission.requestId,
+		message: `Expected completion time: ${dataSubjectRequestSubmission.expectedCompletionTime.toISOString()}`,
+	};
```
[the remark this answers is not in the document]
andresilva-guardian: @johnduffell I have redesigned everything to infer interfaces from Zod to make sure that data that enters and leaves this function respects the defined typings during runtime. What do you think?

cdk/lib/mparticle-api.ts
```diff
@@ -105,15 +111,21 @@ export class MParticleApi extends GuStack {
...
-		/**
-		 * Export Lambda role ARN for cross-account queue access.
-		 * The SQS queue policy in account "Ophan" imports this ARN
-		 * to grant this Lambda sqs:SendMessage permissions to the erasure queue
-		 */
-		new CfnOutput(this, 'MParticleLambdaRoleArn', {
-			value: lambda.role!.roleArn,
-			description: 'ARN of the mParticle Lambda execution role',
-			exportName: `${app}-${this.stage}-lambda-role-arn`,
+		const batonAccountId = '029312801662';
```
AshCorr, line 114: We usually try to avoid revealing our AWS account IDs in repositories, could this be moved to a Parameter Store value or a CFN parameter?

handlers/mparticle-api/package.json
```diff
@@ -7,19 +7,17 @@
...
   "devDependencies": {
     "@faker-js/faker": "^9.8.0",
     "@types/aws-lambda": "^8.10.147"
   },
   "dependencies": {
-    "@aws-sdk/client-sqs": "3.787.0",
-    "@aws-sdk/credential-provider-node": "3.782.0",
     "@peculiar/x509": "^1.12.3",
-    "dotenv": "^16.5.0",
+    "uuid": "^11.1.0",
```
AshCorr, line 20: From what I could see we're only using uuid V4 in this codebase. Might it be worth using [Node's built in UUID generator](https://nodejs.org/api/all.html#all_crypto_cryptorandomuuidoptions) if we only need V4 UUIDs instead of relying on a third party library?

handlers/mparticle-api/package-lock.json
```diff
@@ -0,0 +1,132 @@
+{
```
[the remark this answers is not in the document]
andresilva-guardian: You are right @AshCorr , thanks, removed this file.

cdk/lib/mparticle-api.ts
andresilva-guardian: @AshCorr Thanks! I have cloned the same approach as soft-opt-in-consent-setter for the "mobile" account. What do you think?

AshCorr: Looks great, thanks!

cdk/lib/mparticle-api.ts
```diff
@@ -2,19 +2,36 @@ import { GuApiGatewayWithLambdaByPath } from '@guardian/cdk';
...
-	constructor(scope: App, id: string, props: GuStackProps) {
+	constructor(scope: App, id: string, props: MParticleApiProps) {
 		super(scope, id, props);
 
 		const app = 'mparticle-api';
 
+		const batonAccountId = StringParameter.fromStringParameterName(
+			this,
+			'BatonAccountId',
+			props.batonAccountIdSSMParam,
```
johnduffell, line 32: that's a good idea to have it in the root of SSM, I didn't realise that was a standard thing.  In alarms-handler it has it's own copy which will be unnecessary.

optional: If it's not already there it could be worth having it as `/accountIds/baton`

Also does it need to be parameterised e.g. for testing or can we just inline it?
```suggestion
			'batonAccountId',
```

handlers/mparticle-api/src/apis/data-subject-requests.ts
```diff
@@ -163,65 +162,18 @@ export const getStatusOfDataSubjectRequest = async (
...
 	requestId: string,
 	payload: DataSubjectRequestCallback,
-): Promise<{
+): {
 	message: string;
 	timestamp: Date;
-}> => {
-	console.debug('Process Data Subject Request Callback from mParticle', {
+} => {
+	// Just log this information so we can have track of it on Cloud Watch
```
johnduffell, line 172: is that just for general debugging? or any specific scenario where we would need it?

handlers/mparticle-api/src/index.ts
```diff
@@ -1,208 +1,62 @@
...
-			throw err;
-		}
-	};
+function isAPIGatewayEvent(event: any): event is APIGatewayProxyEvent {
+	return (
+		event &&
+		typeof event.httpMethod === 'string' &&
+		typeof event.path === 'string'
+	);
+}
```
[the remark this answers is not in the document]
johnduffell: To get a high level idea what I mean, AI suggests something like this to replace the handler function below.
```typescript
const APIEventSchema = z.object({
  httpMethod: z.string(),
  path: z.string(),
  headers: z.record(z.string(), z.string().or(z.array(z.string()))),
  body: z.string().optional(), // refine as needed
  // add other APIGatewayProxyEvent properties you rely on
});

const BatonEventSchema = z.object({
  batonId: z.string(),
  action: z.string(),
  // include whatever BatonRerEventRequest requires
});

export const handler = async (
  event: unknown
  // Try parsing as API Gateway event
  const apiParse = APIEventSchema.safeParse(event);
  if (apiParse.success) {
    const apiEvent = apiParse.data as APIGatewayProxyEvent;
    return handleHttpRequest(apiEvent);
  }

  // Else, try Baton Rer Event
  const batonParse = BatonEventSchema.safeParse(event);
  if (batonParse.success) {
    const batonEvent = batonParse.data as BatonRerEventRequest;
    return handleBatonRerEvent(batonEvent);
  }

  // Neither matched → invalid contract
  console.error('Invalid event shape', batonParse.error, apiParse.error);
  return {
    statusCode: 400,
    body: JSON.stringify({
      error: 'Invalid event',
      apiErrors: apiParse.error?.format(),
      batonErrors: batonParse.error?.format(),
    }),
  };
};
```

handlers/mparticle-api/src/routers/baton.ts
```diff
@@ -0,0 +1,209 @@
...
+		public readonly errors: z.ZodError,
+	) {
+		super(message);
+		this.name = 'ValidationError';
+	}
+}
+
+// Helper function to validate and parse requests
+function validateRequest(data: BatonRerEventRequest): BatonRerEventRequest {
+	const result = BatonRerEventRequestSchema.safeParse(data);
```
johnduffell, line 103: if we just do "parse" instead of "safeParse" will that just throw the exception for us?

johnduffell:
> Dual Router Architecture:
Introduces a dual router approach:
> 
> The HTTP router continues to handle mParticle callbacks and legacy/future HTTP endpoints.
The Baton router handles Baton RER requests, making the Lambda compatible with the Baton workflow.

I'm was not clear why we need dual router rather than just defining one router and importing both sets of endpoints, but I think the key information is that the built in router only handles API gateway, but baton calls our lambda directly with a totally different payload type?

So the idea is to first detect whether it's API gateway or Baton payload incoming, and then pass it to the appropriate child router in order to decide what to do.

I think I'd have made two separate lambdas in the CDK and different entry points in the same codebase (like [alarms handler](https://github.com/guardian/support-service-lambdas/blob/f65dd11371693b55c6c1fd0d907085ec38b783f5/handlers/alarms-handler/package.json#L7)), but I think your approach is fair enough if it's easy to make it clear.

But now it does make a lot more sense what's going on in the top level handler file and why it's not clear where the best place to parse is.

handlers/mparticle-api/src/index.ts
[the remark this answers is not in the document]
johnduffell: having realised that this is just a parent router that decides which actual router to use, I've realised why it doesn't make sense to parse a full API gateway request up front.
If it's too much trouble at this stage to deploy the code as two separate lambdas with separate entry points, then I think it's going to be good enough to check a few fields for this pre-routing.

cdk/lib/mparticle-api.ts
andresilva-guardian: Done @johnduffell , thanks!

handlers/mparticle-api/src/apis/data-subject-requests.ts
andresilva-guardian: I'm looking to have this here so we can have a recent log of status tracking of each process, in case that mParticle platform is down or something else fails. At least, for the upcoming weeks, it will help us know that everything is working as expected.

cdk/lib/mparticle-api.ts
```diff
@@ -97,25 +113,49 @@ export class MParticleApi extends GuStack {
...
-				lambdaFunctionNames: lambda.functionName,
+				lambdaFunctionNames: httpLambda.functionName,
+			});
+
+			// Baton Lambda error alarm
+			new SrLambdaAlarm(this, 'MParticleBatonLambdaErrorAlarm', {
+				app,
+				alarmName: 'An error occurred in the mParticle Baton Lambda',
+				alarmDescription:
+					'mParticle Baton Lambda failed, please check the logs to diagnose the issue.',
```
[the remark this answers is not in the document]
johnduffell: also the same would apply to the other alarms for the http side,  I guess it might be more complex if a callback is missed

handlers/mparticle-api/src/routers/baton.ts
```diff
@@ -0,0 +1,209 @@
...
+	const dataSubjectRequestState: DataSubjectRequestState =
+		await getStatusOfDataSubjectRequest(request.initiationReference);
+
+	const response: BatonRerEventStatusResponse = {
+		requestType: 'RER' as const,
+		action: 'status' as const,
+		status: mapStatus(dataSubjectRequestState.requestStatus),
+	};
+
+	return validateResponse(response);
```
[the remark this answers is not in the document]
andresilva-guardian: Correct, response validation removed ✅

cdk/lib/mparticle-api.ts
```diff
@@ -80,40 +134,47 @@ export class MParticleApi extends GuStack {
...
 
 		new SrLambdaDomain(this, {
 			subdomain: 'mparticle-api',
 			restApi: apiGateway.api,
 		});
 
-		/**
-		 * Export Lambda role ARN for cross-account queue access.
-		 * The SQS queue policy in account "Ophan" imports this ARN
-		 * to grant this Lambda sqs:SendMessage permissions to the erasure queue
```
johnduffell, line 111: I think we should keep this comment only updated for the baton route.
Maybe just "We grant the permission so baton can call the lambdas directly as per https://github.com/guardian/baton?tab=readme-ov-file#adding-data-sources-to-baton "
Also this link can either stay in a PR comment or be added as a comment in the CDK here, as it shows where the instructions came from to do this (although they are CFN rather than CDK)
https://github.com/guardian/baton-lambda-template/blob/61ebdec91bd53e218d5f5a2aa90494db69adfa0a/src/main/g8/cfn.yaml#L44-L46

cdk/lib/mparticle-api.ts
```diff
@@ -97,25 +113,49 @@ export class MParticleApi extends GuStack {
...
-						resources: ['*'],
-					}),
-				],
-			}),
-		);
+		// CloudWatch permissions for both lambdas
+		const cloudWatchPolicy = new Policy(this, `${app}-cloudwatch-policy`, {
+			statements: [
+				new PolicyStatement({
+					actions: ['cloudwatch:ListTagsForResource'],
```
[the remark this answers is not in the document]
andresilva-guardian: It was already there, I believe added from @ecguardian . "alarms-handler" also has this, so I believe that maybe it was a clone version from it.
We could replace it by:
```
const cloudWatchPolicy = new Policy(this, `${app}-cloudwatch-policy`, {
			statements: [
				new PolicyStatement({
					actions: [
						'logs:CreateLogGroup',
						'logs:CreateLogStream',
						'logs:PutLogEvents',
					],
					resources: ['*'],
				}),
			],
		});
```

But I have a feeling that we can remove this completely. I will test it and see if it works, let you know after!

cdk/lib/mparticle-api.ts
andresilva-guardian: Documentation added, thanks for the links @johnduffell ✅

andresilva-guardian (commented):
Comment done.

cdk/lib/mparticle-api.ts
```diff
@@ -15,17 +22,40 @@ export class MParticleApi extends GuStack {
...
 			runtime: nodeVersion,
 			timeout: Duration.seconds(15),
-			handler: 'index.handler',
-			functionName: `${app}-${this.stage}`,
+			handler: 'index.handlerHttp',
+			functionName: `${app}-http-${this.stage}`,
 			events: [],
 			environment: {
-				APP: app,
+				APP: `${app}-http`,
```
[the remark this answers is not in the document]
andresilva-guardian: @johnduffell They are now different lambdas, with different purposes and different infraestruture. I believe that is not a good ideia to use the same configs, as change the configs for one can be damaging to the other, ruining the isolation that we just provided 👍

cdk/lib/mparticle-api.ts
```diff
@@ -80,25 +112,36 @@ export class MParticleApi extends GuStack {
...
-					}),
-				],
-			}),
-		);
+			// Baton Lambda error alarm
+			new SrLambdaAlarm(this, 'MParticleBatonLambdaErrorAlarm', {
+				app,
+				alarmName: 'An error occurred in the mParticle Baton Lambda',
+				alarmDescription:
+					'mParticle Baton Lambda failed, please check the logs to diagnose the issue. Impact: a user may not be deleted from mParticle+Braze after an erasure request, and Baton would display an error.',
```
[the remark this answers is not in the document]
johnduffell: ok thanks I made a good guess then 🎉

cdk/lib/mparticle-api.ts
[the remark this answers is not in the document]
johnduffell: Firstly are you sure the override is working? It looks like the APP env is overridden to match the declared app of the lambda tags 
https://github.com/guardian/cdk/blob/df676ec08665bca7f6f689d399aada27981c12e2/src/constructs/lambda/lambda.ts#L147-L148


and more generally, do we aim for them to be isolated for config also?  my understanding was the isolation was only from the AWS/caller's side and to avoid extra routing logic.

I think this is revealing a bit of a grey area I'm having around the config here so I have a few questions!

Regarding config they both seem to use the same config schema do they? so they need all the same values.  Or does the Http lambda not use config at all?
If someone tries to rotate a credential when would they want it to apply to one lambda and not the other?  What happens if they get out of step with each other and is that something we would expect?

Looking at the overview above and the schema
https://github.com/guardian/support-service-lambdas/blob/513a6f34212e13f102232d62aa4babb805a5a4be/handlers/mparticle-api/src/utils/config.ts#L5-L15
From the names it's not clear whether workspace and inputPlatform and pod are all relating to connecting to mParticle API?
But it seems like only the baton lambda hits mparticle, it's unclear whether the http lambda needs any config at all?  It's hard to follow through to tell what lambdas ultimately call requestEventsApi
https://github.com/guardian/support-service-lambdas/blob/513a6f34212e13f102232d62aa4babb805a5a4be/handlers/mparticle-api/src/apis/events.ts#L13
and requestDataSubjectApi
https://github.com/guardian/support-service-lambdas/blob/79a114ecf32ac3d748ec722e721bd00afe6080f2/handlers/mparticle-api/src/apis/data-subject-requests.ts#L32
And I can't see where `pod` property is used at all.

I also notice that the config is loaded deep inside the code.  Actually the intention was the config would be loaded up front and passed in where it's needed, then it's clear which lambda needs which config (more like dependency injection approach)
Since here config is loaded at runtime as a side effect, if someone reuses the code in another app, the compiler wouldn't confirm that the config really exists.

handlers/mparticle-api/src/routers/http.ts
```diff
@@ -0,0 +1,192 @@
...
+		try {
+			return await fn(event, parsed);
+		} catch (err) {
+			if (err instanceof HttpError) {
+				console.error(
+					`Http Error: ${err.statusCode} ${err.statusText}`,
+					err.body,
+				);
+				return {
+					statusCode: err.statusCode,
```
johnduffell, line 39: It looks like this is designed so we can throw a generic HttpError in the codebase and have a custom response code defined at that point, rather than throwing a business level exception (e.g. MparticleUserNotFound) and converting it to a 404 explicitly.
This does hurt reusability of the code slightly, but it means for less code in http serialiation layer, if we know exactly what HTTP response we want to return.

In this case, what does status code 0 mean?
https://github.com/guardian/support-service-lambdas/blob/36d26ce89fe152ff5136c5ededa859772cadb0d7/handlers/mparticle-api/src/utils/make-http-request.ts#L84
It looks like we also forward on the status code and body from downstream services, 
https://github.com/guardian/support-service-lambdas/blob/36d26ce89fe152ff5136c5ededa859772cadb0d7/handlers/mparticle-api/src/utils/make-http-request.ts#L63

However generally speaking this is a [security risk](https://devguide.owasp.org/en/12-appendices/01-implementation-dos-donts/06-exception-error-handling/), because we are potentially passing on trusted information which could help an attacker.  Also if we are getting back a 4xx error due to a bug in this lambda, we should report that as a 5xx so that we get an alarm and can investigate.  i.e. we are the client and the server, so if it's our error, then upstream will return 4xx and we need to return 5xx.

Maybe this is less likely because mparticle is intended to be the client, however there's nothing stopping attackers probing it for weaknesses, and also this code may be extended or reused as a good example.

My suggestion would be to allow the Router to catch and log the error, and it will return a 500 error that doesn't tell the end user any further clue about the failure or where it occurred
 https://github.com/guardian/support-service-lambdas/blob/856dba0ec0cfe366020f6b0772afa99b50c08f6e/modules/routing/src/router.ts#L115-L118

handlers/mparticle-api/README.md
```diff
@@ -46,9 +46,6 @@ This project implements an AWS Lambda function (written in TypeScript) that inte
 - **Build:**  
   `pnpm --filter mparticle-api build`
 
-- **Run locally:**
```
johnduffell, line 49: could be in a separate PR but the readme could do with a once over just to make it clear that it's a "baton -> this lambda 1 -> mparticle -> this lambda 2" situation now.  Although it's obviously written so that other than baton could call it, it's pretty useful to briefly know that's how it's called in practice.
Also I think the "API Endpoints" section, they are changed now?  Maybe the first two are baton format direct call, and the third is still http/api gateway.  And the last one I'm not sure?

handlers/mparticle-api/src/routers/baton.ts
```diff
@@ -0,0 +1,200 @@
...
+ * A branded string type that represents a GUID (Globally Unique Identifier).
+ */
+export type GUID = string & { readonly __brand: unique symbol };
+
+/**
+ * A GUID that uniquely identifies an initiation reference.
+ */
+export type InitiationReference = GUID;
+
+// Zod schemas for validation
```
johnduffell, line 23:
```suggestion
// this prevents us mixing up UUIDs and normal strings at compile time
export type GUID = string & { readonly __brand: unique symbol };

export type InitiationReference = GUID;

```

handlers/mparticle-api/src/routers/baton.ts
```diff
@@ -0,0 +1,200 @@
...
+	constructor(
+		message: string,
+		public readonly errors: z.ZodError,
+	) {
+		super(message);
+		this.name = 'ValidationError';
+	}
+}
+
+// Helper function to validate and parse requests
```
johnduffell, line 101:
```suggestion
```

cdk/lib/mparticle-api.ts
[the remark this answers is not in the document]
andresilva-guardian: @johnduffell `pod` is used [here](https://github.com/guardian/support-service-lambdas/blob/afs/mparticle-baton-rer/handlers/mparticle-api/src/apis/events.ts#L16).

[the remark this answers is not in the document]
andresilva-guardian: @johnduffell Since there are default configurations, the ones on mParticle-api CDK were removed.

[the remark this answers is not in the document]
andresilva-guardian: @johnduffell I'm sorry but I'm not able to meet your expectations here. Changing config loading to a dependency injection approach would mean changing the routing module again so we could pass the config to each endpoint of the http lambda. Can you please code your suggestion and open a pull request to this branch? Thanks

handlers/mparticle-api/src/routers/http.ts
```diff
@@ -0,0 +1,192 @@
...
+					'It was not possible to set the User Attribute to remove user from audiences or from event forwarding during the waiting period.',
+					error,
+				);
+			}
+
+			return {
+				statusCode: 201,
+				body: JSON.stringify(await submitDataSubjectRequest(parsed.body)),
+			};
+		}),
```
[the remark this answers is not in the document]
andresilva-guardian: Done @johnduffell thank you for the suggestion.

handlers/mparticle-api/src/routers/baton.ts
andresilva-guardian: Done @johnduffell 💪 Thanks

handlers/mparticle-api/README.md
andresilva-guardian: @johnduffell I used AI within The Guardian's [policies](https://docs.google.com/document/d/1gUdmHNVl0q95pdp3p0dPtdiYlxc78dlIis8cWLWYlGA/edit?tab=t.bhh84qotdb4v) for AI usage to help me generate a couple of new documents about this API. This is something experimental, it is the first time that I'm guiding a copilot to generate proper documentation based on my inputs, code examples and instructions. If you please could help me by taking a look at it I would appreciate it very much. 🙏

handlers/mparticle-api/src/utils/config.ts
```diff
@@ -18,13 +18,9 @@ export type AppConfig = z.infer<typeof ConfigSchema>;
...
-		const stage = getEnv('STAGE');
-		const stack = getEnv('STACK');
-		const app = getEnv('APP');
-		appConfig = await loadConfig(stage, stack, app, ConfigSchema);
-	}
-	return appConfig;
+	const stage = getEnv('STAGE');
+	const stack = getEnv('STACK');
+	const app = getEnv('APP');
+	return loadConfig(stage, stack, app, ConfigSchema);
```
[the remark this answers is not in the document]
andresilva-guardian: I agree, but I believe the goal is so make sure that a reload of the variable will make it work without a instance reboot.

handlers/mparticle-api/README.md
```diff
@@ -1,75 +1,505 @@
...
+- 📈 **Analytics Forwarding**: Route event data to mParticle and downstream systems
+- 🛡️ **Enterprise Security**: Certificate validation and signature verification on callbacks from mParticle
 
-- **Track DSR Status:**  
-  Provides endpoints to query the status of existing DSRs by request ID.
+### Compliance & Standards
+- **GDPR Article 17**: Right to Erasure implementation
+- **CCPA Section 1798.105**: Consumer data deletion rights
+- **X.509 Certificate Validation**: Cryptographic security for callbacks
+- **AWS Security Best Practices**: Encrypted data storage and Parameter Store configuration
```
johnduffell, line 20: I'd probably say a lot of this is information about baton rather than this lambda.  Should we consider moving the general information to the Baton documentation?
Then we could shorten this to something like 
"This lambda integrates the [Baton](https://github.com/guardian/baton)  tool with mParticle API to allow both erasure and subject access requests to be raised and tracked in the standard way.  As part of this, it handles mParticle callbacks to track the status of requests in real time."
Then the extra information if it's useful could be added to baton readme.

handlers/mparticle-api/README.md
```diff
@@ -1,75 +1,505 @@
...
-## Key Features
+The mParticle API Lambda enables Guardian to fulfill GDPR and CCPA compliance requirements by providing a secure, scalable interface to mParticle's Data Subject Request (DSR) API. It processes privacy requests, tracks their status, and coordinates with Guardian's broader privacy ecosystem.
 
-- **Submit Data Subject Requests (DSR):**  
-  Accepts requests for access, portability, or erasure and submits them to mParticle in OpenDSR format.
+### Key Capabilities
+- 🔒 **Privacy Rights Fulfillment**: Process data deletion and export requests
+- 📊 **Real-time Status Tracking**: Monitor DSR progress with callback integration  
+- 🤖 **Automated Workflows**: Integrate with Baton for orchestrated privacy operations
+- 📈 **Analytics Forwarding**: Route event data to mParticle and downstream systems
```
johnduffell, line 13: is this a separate future function? or is event data something to do with GDPR?

handlers/mparticle-api/README.md
```diff
@@ -1,75 +1,505 @@
...
+- **Data Orchestration**: Real-time data routing to 300+ marketing, analytics, and data warehouse integrations
+- **Identity Resolution**: Cross-platform customer identity management and unification
+- **Audience Management**: Dynamic audience creation and real-time segmentation
+- **Privacy Controls**: Built-in data governance and privacy compliance tools
 
-## API Endpoints
+### Official Resources
+- [mParticle Documentation](https://docs.mparticle.com/)
+- [Data Subject Request API](https://docs.mparticle.com/developers/dsr-api/)
+- [Privacy Controls Overview](https://docs.mparticle.com/guides/data-privacy-controls/)
```
johnduffell, line 38: this is a great starting point with good jumping off points for more info.  Also, feel free to link to internal documents e.g. google docs (external people won't be able to access them, but that's fine)

handlers/mparticle-api/README.md
```diff
@@ -1,75 +1,505 @@
...
+- **Personalization Engine**: AI-powered content personalization and optimization
+- **Journey Orchestration**: Automated, cross-channel customer journey campaigns
+- **Real-Time Analytics**: Campaign performance tracking and user behavior analytics
+- **A/B Testing**: Comprehensive testing framework for message optimization
 
-## Development
+### Official Resources
+- [Braze Documentation](https://www.braze.com/docs/)
+- [Data Privacy & Protection](https://www.braze.com/docs/developer_guide/disclosures/security_qualifications/)
+- [GDPR Compliance Guide](https://www.braze.com/docs/help/gdpr_compliance/)
```
johnduffell, line 56: again great although it might be worth having a guardian centric description rather than the generic feature set - i.e. it's mainly used by CRM team for email campaigns driven from bigquery data, as well as triggered service emails.

handlers/mparticle-api/README.md
```diff
@@ -1,75 +1,505 @@
...
 
-- **Install dependencies:**  
-  Run `pnpm install` from the root of the repo.
+---
 
-- **Build:**  
-  `pnpm --filter mparticle-api build`
+## 🔗 mParticle and Braze Integration
```
johnduffell, line 60: very useful to have this section, but is this represented anywhere else? I'm thinking if there's an exsting repo or document then we can link there instead of replicating it?

handlers/mparticle-api/README.md
```diff
@@ -1,75 +1,505 @@
...
+    participant B as Baton
+    participant MP as mParticle API
+    participant M as mParticle
+    participant BR as Braze
+
+    Note over R,BR: Right to Erasure Request (RER) Flow
+    
+    R->>B: Request data erasure
+    B->>MP: Initiate RER (Baton Handler)
+    MP->>MP: Set user attributes<br/>(remove from audiences)
```
johnduffell, line 176: is this not to M?
```suggestion
    MP->>M: Set user attributes<br/>(remove from audiences)
```

handlers/mparticle-api/README.md
```diff
@@ -1,75 +1,505 @@
...
+```
+
+## Service Integration Flow
+
+```mermaid
+sequenceDiagram
+    participant R as Reader
+    participant B as Baton
+    participant MP as mParticle API
+    participant M as mParticle
```
johnduffell, line 169: should these be LAMBDA and MP or MPAPI and MP?

handlers/mparticle-api/README.md
```diff
@@ -1,75 +1,505 @@
...
+### Installation
+```bash
+# Install dependencies
+pnpm install
+
+# Build the project  
+pnpm build
+
+# Run tests
+pnpm test
```
johnduffell, line 203:
```suggestion
cd handlers/mparticle-api
pnpm install
pnpm build
pnpm test
```

handlers/mparticle-api/README.md
```diff
@@ -1,75 +1,505 @@
...
+  - *mParticle Reference*: [HTTP APIs Authentication](https://docs.mparticle.com/developers/apis/http/#authentication)
+
+#### Environment Configuration
+- **`/{stage}/support/mparticle-api/pod`**
+  - *Description*: mParticle pod/cluster identifier (e.g., "us1", "us2", "eu1") for regional API endpoints
+  - *mParticle Reference*: [Data Hosting Locations](https://docs.mparticle.com/developers/guides/data-localization/)
+
+---
+
+## 🌐 HTTP API Endpoints
```
johnduffell, line 234: I feel like this is a good summary of the code but if I wanted to know this I'd go straight to the router as it's more likely to be up to date and it's in context.  Would it be ok to link to the http router code instead and add any important missing context as comments?

handlers/mparticle-api/README.md
```diff
@@ -1,75 +1,505 @@
...
+│   │   ├── httpRouter.test.ts       # HTTP endpoint tests
+│   │   └── batonRouter.test.ts      # Baton integration tests
+│   ├── services/
+│   │   ├── mparticle.test.ts        # mParticle API client tests
+│   │   └── validation.test.ts       # Certificate validation tests
+│   └── schemas/
+│       └── validation.test.ts       # Zod schema validation tests
+├── fixtures/                        # Test data and mock responses
+└── helpers/                         # Test utility functions
+```
```
johnduffell, line 385: I think this isn't needed.  Although consider whether the file names could be more expressive in the light of the comments on the right

handlers/mparticle-api/README.md
```diff
@@ -1,75 +1,505 @@
...
+pnpm test
+
+# Run with coverage
+pnpm test --coverage
+
+# Run specific tests
+pnpm test httpRouter.test.ts
+
+# Development mode
+pnpm test --watch
```
johnduffell, line 399: interesting - I've never done coverage or watch so I must try them sometime!

handlers/mparticle-api/README.md
```diff
@@ -1,75 +1,505 @@
...
+- 🎯 **Subscription Data**: Payment information, subscription preferences
+- 📧 **Marketing Analytics**: Email engagement, campaign performance
+- 📱 **Cross-Platform Behavior**: Mobile app usage, web interactions
+
+This data flows through **mParticle** for analytics and audience segmentation, then forwards to **Braze** for marketing automation.
+
+### Risk Mitigation
+- **Automated processing** reduces manual errors
+- **Centralized orchestration** via Baton ensures no system is missed
+- **Real-time monitoring** detects and alerts on failures immediately
```
johnduffell, line 418: This is the more guardian centric angle I was asking for earlier in the file!  So perhaps combine this into the above section (or into an existing doc and link to it if we have one)
I think Risk mitigation can be deleted as it's mostly our standard practices for using baton and alerting.

However similar to the quick start section, if you feel there are any particular alarms or failure modes that might flummox people, it could be worth adding a section/separate file (and linking to it from the alarm description)

handlers/mparticle-api/README.md
```diff
@@ -1,75 +1,505 @@
...
+
+---
+
+## 🔧 Development & Configuration
+
+### Environment Variables
+- `STAGE`: Deployment environment (CODE/PROD)
+- `AWS_REGION`: AWS region for Parameter Store access
```
johnduffell, line 426: we always use eu-west-1 so to keep setup consistent between lambdas you can reuse the hard coded region:
https://github.com/guardian/support-service-lambdas/blob/c4d2e7d3770503ebdec39fc34d476a7a1e7629cb/modules/aws/src/config.ts#L8

handlers/mparticle-api/README.md
```diff
@@ -1,75 +1,505 @@
...
+pnpm fix-formatting     # Format code with Prettier
+pnpm check-config       # Check configuration files
+```
+
+### Local Development
+```bash
+# Set environment
+export STAGE=CODE
+export AWS_REGION=eu-west-1
+```
```
johnduffell, line 456: This doesn't actually say what to run.
I would incorporate this into the quick start.
I would say use a "runManual" script and hard code CODE in the file (or whatever env you prefer) and tell people to run that.  (intellij let's you right click and run but maybe there's a more universal way)

handlers/mparticle-api/README.md
```diff
@@ -1,75 +1,505 @@
...
+
+### Deployment
+Managed through AWS CDK with environment-specific configurations.
```
johnduffell, line 459: it's not it's actually the standard GHA->riffraff route and there's no need to mention it as it's not special.

handlers/mparticle-api/README.md
```diff
@@ -1,75 +1,505 @@
...
+
+---
+
+## 📞 Support & Monitoring
+
+### Team Information
+- **Primary Owner**: Value Team
+- **Slack Channel**: [@P&E/Value](https://chat.google.com/room/AAAAuotUxTg?cls=7)
```
johnduffell, line 467: nice idea but I think remove this as the [spreadsheet](https://docs.google.com/spreadsheets/d/1bb8WB-6ZFRdUwERHMOVIolN8WU8zJMcsyi-WJuwp07Q/edit?gid=0#gid=0) is the source of truth there and changes from time to time.

I think there are standard company ways of determining ownership for people outside the team based on the repo "admin" role (which is value team)

I think you are pickup up a few things here that should be at the repo level rather than the handler level.  I wouldn't object to a total refresh of the top level README!

handlers/mparticle-api/README.md
```diff
@@ -1,75 +1,505 @@
...
+
+### External Dependencies
+- **mParticle DSR API**: Data subject request processing
+- **AWS Parameter Store**: Configuration management
+- **CloudWatch**: Logging and monitoring
+- **Baton**: Privacy workflow orchestration
```
johnduffell, line 473: I think delete this it's just too much detail

handlers/mparticle-api/README.md
```diff
@@ -1,75 +1,505 @@
...
+The Lambda integrates with Guardian's Baton system for automated privacy workflows:
+- Cross-account Lambda invocation
+- Correlation tracking for multi-service requests
+- Status synchronization across privacy processors
+
+### mParticle Data Subject Requests
+Direct integration with mParticle's DSR API:
+- Automated request submission and tracking
+- Secure callback handling with certificate validation
+- Status polling and update processing
```
johnduffell, line 489: also delete as above, add any comments if important enough to the code

handlers/mparticle-api/README.md
```diff
@@ -1,75 +1,505 @@
...
+
+---
+
+## 📋 Quick Reference
+
+### AWS Resources
+- **Lambda Functions**: `mparticle-api-http-{stage}`, `mparticle-api-baton-{stage}`
+- **API Gateway**: `mparticle-api-{stage}`
+- **Parameter Store**: `/{stage}/support/mparticle-api/*`
+- **Cross-Account Role**: `baton-mparticle-lambda-role-{stage}`
```
johnduffell, line 499: you've named things so consistently and used standards where approprate, that I would say this isn't needed.
If I'm struggling I find the stack in AWS and navigate from the Resources tab.
Not sure if finding things in AWS is an issue throughout the team though - does seem like a knowledge sharing thing.

handlers/mparticle-api/src/apis/events.ts
```diff
@@ -67,4 +67,22 @@ export const uploadAnEventBatch = async (
 	return {};
 };
 
+export const setUserAttributesForRightToErasureRequest = async (
+	environment: 'production' | 'development',
+	userId: string,
+	submittedTime: string,
+): Promise<object> => {
```
johnduffell, line 74: Suggested post-merge refactors:
1:
object doesn't provide any useful information.  In fact looking at uploadAnEventBatch it seems always an empty object which would be better represented with `void`.

2 (not introduced in this PR):
The underlying response is not validated by Zod which we should do in all places
https://github.com/guardian/support-service-lambdas/blob/36d26ce89fe152ff5136c5ededa859772cadb0d7/handlers/mparticle-api/src/utils/make-http-request.ts#L76

See zuoraclient for an existing example of doing an authenticated http request with zod parsing:
https://github.com/guardian/support-service-lambdas/blob/db568862b899ab115fa1a5ddf16d6abdaf99d5b0/modules/zuora/src/zuoraClient.ts#L64-L63

handlers/mparticle-api/src/routers/baton.ts
```diff
@@ -0,0 +1,200 @@
...
+function validateRequest(data: BatonRerEventRequest): BatonRerEventRequest {
+	const result = BatonRerEventRequestSchema.safeParse(data);
+	if (!result.success) {
+		console.error('Request validation failed:', result.error);
+		throw new ValidationError('Invalid request format', result.error);
```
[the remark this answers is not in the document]
johnduffell: I think it will log the error in this lambda logs, and baton will see some kind of generic error, but as you mention the baton team can contact our team for more information?

handlers/mparticle-api/src/routers/http.ts
```diff
@@ -0,0 +1,47 @@
...
+		path: '/data-subject-requests',
+		handler: submitDataSubjectRequestHandler(),
+		parser: dataSubjectRequestFormParser,
+	}),
+	createRoute<{ requestId: string }, unknown>({
+		httpMethod: 'GET',
+		path: '/data-subject-requests/{requestId}',
+		handler: getDataSubjectRequestStatusHandler(),
+		parser: requestIdPathParser,
+	}),
```
johnduffell, line 34: there are four routes here but it looks like only one can be called
https://github.com/guardian/support-service-lambdas/blob/1f84ff25a69dfe78c179e15b1a1c51442a4c15b8/cdk/lib/mparticle-api.ts#L56-L62
It seems like we have a mismatch, as the lines in the CDK show that API gateway is doing the routing, but we are using http router here.
https://github.com/guardian/cdk/blob/9a90a1615ced38349cc5bbd49fb211a10bac7814/src/patterns/api-lambda.ts#L34-L37

Unless I'm missing something it looks like we should switch to `GuApiLambda` and remove the unused routes?

handlers/mparticle-api/src/routers/http.ts
```diff
@@ -0,0 +1,47 @@
...
+	createRoute<{ requestId: string }, DataSubjectRequestCallback>({
+		httpMethod: 'POST',
+		path: '/data-subject-requests/{requestId}/callback',
+		handler: dataSubjectRequestCallbackHandler(),
+		parser: dataSubjectRequestCallbackParser,
```
johnduffell, line 39: mparticle say "You must validate the authenticity of the request before parsing the request body."
https://docs.mparticle.com/developers/apis/dsr-api/v3/#callbacks:~:text=You%20must%20validate%20the%20authenticity%20of%20the%20request%20before%20parsing%20the%20request%20body.

I'm not sure exactly why as they don't mention the reason in the docs, do you know?

cdk/lib/mparticle-api.ts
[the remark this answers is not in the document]
johnduffell: good idea to suggest opening a PR to make sure I think through my suggestion in more detail. It has really helped my understanding of this PR to look through in more detail
The result is here, take a look: https://github.com/guardian/support-service-lambdas/pull/2960

johnduffell (approved):
I have had a good look through locally especially thanks to your suggestion to make a branch to demo my changes.  I have added a lot of outstanding questions/nitpicks/suggestions in the comments of this PR.

The PR with my additional suggestions is here also: https://github.com/guardian/support-service-lambdas/pull/2960

Since I'm off next week I will approve pre-emptively, as this ideally needs to be ticked off on the project plan before I get back.  But there will be time allocated to do a post PR refactor according to any outstanding comments.

In that light, I think there are no blockers that I could see in terms of functionality or security, and you have tested it and got a good result.  So I think the best course might be to merge and get the tick, and make a fresh PR for the comments and suggestions.

handlers/mparticle-api/src/index.ts
```diff
@@ -1,208 +1,40 @@
...
+export const handlerBaton: Handler<
+	BatonRerEventRequest,
+	BatonRerEventResponse
+> = async (event: BatonRerEventRequest): Promise<BatonRerEventResponse> => {
+	try {
+		console.debug('Processing Baton RER event');
+		return batonRerRouter.routeRequest(event);
+	} catch (error) {
+		console.error('Baton RER handler error:', error);
+		throw error; // Re-throw to trigger Lambda retry mechanism
```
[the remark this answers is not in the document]
andresilva-guardian: A erasure request is a serious subject within a company. Legally we have 28 days to comply with or face huge fines. The team members using Baton will see the error state and they can manually retry it on the UI (I did it myself) so no worries here as this as to be follow up closely everyday (every 12 hours I would suggest).

PR #2923, opened 2025-07-07
use semi-auto json in SOI setter

johnduffell:
pre-refactor to make it easier to review https://github.com/guardian/support-service-lambdas/pull/2912

This PR changes soft-opt-in-consent-setter to use [semiauto](https://circe.github.io/circe/codecs/semiauto-derivation.html) instead of [auto](https://circe.github.io/circe/codecs/auto-derivation.html) decoders.

Although this means there's more boiler plate as we can't use `derives` syntax in scala 2, it means that

- we won't accidentally conjoure up a decoder for something that shouldn't.
- It's more typical in our code to do it this way, and
- in the context of the following PR, it means we're less likely to accidentally mix up WireMessageBody and MessageBody either in the code or the tests.

### DIFF
--- a/handlers/soft-opt-in-consent-setter/src/test/scala/com/gu/soft_opt_in_consent_setter/JsonCodecSpec.scala
+++ b/handlers/soft-opt-in-consent-setter/src/test/scala/com/gu/soft_opt_in_consent_setter/JsonCodecSpec.scala
@@ -5,7 +5,6 @@ import com.gu.soft_opt_in_consent_setter.models.SFSubRecordResponse
 import com.gu.soft_opt_in_consent_setter.testData.SFSubscriptionTestData.{subRecord2, subRecord3}
 import org.scalatest.flatspec.AnyFlatSpec
 import org.scalatest.matchers.should
-import io.circe.generic.auto._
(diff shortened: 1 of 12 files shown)


handlers/soft-opt-in-consent-setter/src/test/scala/com/gu/soft_opt_in_consent_setter/JsonCodecSpec.scala
```diff
@@ -5,7 +5,6 @@ import com.gu.soft_opt_in_consent_setter.models.SFSubRecordResponse
 import com.gu.soft_opt_in_consent_setter.testData.SFSubscriptionTestData.{subRecord2, subRecord3}
 import org.scalatest.flatspec.AnyFlatSpec
 import org.scalatest.matchers.should
-import io.circe.generic.auto._
```
johnduffell, line 8: the tests were actually separately conjuring up codecs rather than using the ones the main business logic uses.  So it could have actually tested the wrong thing if there was a clash.  Removing the auto import means if it compiles it must be using the shared ones defined in the companion objects.

PR #2925, opened 2025-07-08
Negative invoices processor - fire alarm on failures

david-pepper:
## What does this change?
Fire an alarm to the Platform team channel when an error is thrown from one or more of the lambdas during processing of negative invoices

Screenshot 2025-07-08 at 12 18 00

## How to test
- simulate an error in an interaction with Zuora (e.g. try to apply credit to an account balance referencing an invoice without a negative balance). An error should get thrown, and an alarm should be fired and received in the Platform team chat channel:

_(note, CODE alarms go to SRE channel by default. PROD alarm will go to Platform Team)_
image

### Notes
- Ensure alarms-handler gets deployed in riff-raff for mappings of alarm->Chat channel


graham228221 (approved):
I think your tests for the `detectFailures` lambda need to be updated, but otherwise looks good 👍

PR #2926, opened 2025-07-09
Negative invoices processor - stop processing invoice when zuora callout fails

david-pepper:
## What does this change?

- Restrict the flow of execution so that if either of the Zuora callouts for checking for active subs or active payment methods fail, processing of that specific invoice will cease and the state machine map will move onto the next invoice.
 - Achieved by adding Choice states that inspect zuora callout 'Success' value 

Screenshot 2025-07-10 at 19 03 21

**Discovered and spot-fixed the following bugs whilst testing these changes**
- In zuoraClient module fetch function a check is performed on the response.ok attribute of zuora response, and if false an error is thrown. 
Screenshot 2025-07-10 at 19 15 45

This works for /v1/object/* endpoints, but not for /v1/accounts/* endpoint, which we use for payment method checks.  To fix this, we need to use the 'success' attribute of the response, so updated return object of getPaymentMethods.ts handler

Screenshot 2025-07-10 at 19 13 15

This will be fine for now, but we may want to think about improving in future.

- In detectFailures.ts, refundAttempt was not being checked for failures, so I've added this and updated automated tests accordingly.

## How to test

Also used these changes to initiate end-to-end testing with this [test script](https://docs.google.com/spreadsheets/d/1Kn8Ih-eUorJ48Lx4uWZuNFOdd4_-3FzJ5EhZudAeIzU/edit?usp=sharing)

All tests are now passing


andrade0 (approved):
Well done ;)

PR #2928, opened 2025-07-10
Fix Product Move API: Add Term Renewal to Prevent Cancellation Date Errors

andresilva-guardian:
# Fix Product Move API: Add Term Renewal to Prevent Cancellation Date Errors

## Problem

The product-move-api was failing in production with the error:
```
"Cancellation Date (03/17/2026) cannot be later than the subscription's current term end date (02/20/2026)."
```

This error occurs when the API tries to remove rate plans from a subscription using a `chargedThroughDate` that extends beyond the subscription's current term end date. Zuora rejects such operations because you cannot modify a subscription beyond its current term.

## Root Cause Analysis

The issue was in the `ToRecurringContribution` switch type implementation. When switching from a Supporter Plus subscription to a Recurring Contribution, the API:

1. Calculates a `removeDate` based on the `chargedThroughDate` 
2. Attempts to remove existing rate plans using this date
3. Fails if the remove date is beyond the subscription's current term end date

```scala
// In getRatePlans method - this was the problematic code
removeRatePlans = ratePlanAmendments
  .filterNot(_.lastChangeType.contains("Remove"))
  .map { ratePlan =>
    val effectiveStartDate = ratePlan.ratePlanCharges.headOption.map(_.effectiveStartDate)
    val removeDate = effectiveStartDate.filter(_.isAfter(chargedThroughDate)).getOrElse(chargedThroughDate)
    RemoveRatePlan(removeDate, ratePlan.id) // ❌ This fails if removeDate > term end date
  }
```

## Solution

Added term renewal logic to the `ToRecurringContribution` implementation, similar to what already exists in `RecurringContributionToSupporterPlus`. Additionally, enhanced the `GetSubscriptionResponse` model to include `termEndDate` for more precise term renewal decisions.

### Key Changes

1. **Added `termEndDate` to `GetSubscriptionResponse`**:
```scala
case class GetSubscriptionResponse(
    // ...existing fields...
    termEndDate: LocalDate, // ✅ Added this field
    // ...existing fields...
)
```

2. **Added TermRenewal dependency** to `ToRecurringContributionImpl`:
```scala
class ToRecurringContributionImpl(
    subscriptionUpdate: SubscriptionUpdate,
    termRenewal: TermRenewal, // ✅ Added this dependency
    sqs: SQS,
    stage: Stage,
) extends ToRecurringContribution
```

3. **Added conditional term renewal logic** before subscription updates:
```scala
// Check if we need term renewal to avoid "cancellation date cannot be later than term end date" error
// We need to perform term renewal if the chargedThroughDate (which will be used as the remove date)
// extends beyond the current subscription term end date.
needsTermRenewal = activeRatePlanCharge.chargedThroughDate.exists(_.isAfter(subscription.termEndDate))
_ <- ZIO.when(needsTermRenewal) {
  ZIO.log(
    s"Performing term renewal because chargedThroughDate $chargedThroughDate is after termEndDate ${subscription.termEndDate}",
  ) *>
  termRenewal.renewSubscription(subscriptionName, runBilling = false) *>
  ZIO.log(s"Term renewal completed for subscription $subscriptionName")
}
```

4. **Updated dependency injection** in `ProductMoveEndpoint`:
```scala
toRecurringContribution = new ToRecurringContributionImpl(
  subscriptionUpdate,
  termRenewal, // ✅ Added this parameter
  sqs,
  stage,
)
```

5. **Updated all test fixtures** to include `termEndDate` in the `GetSubscriptionResponse` model.

## Architecture Context

This fix follows the same pattern already established in the codebase:

```mermaid
graph TD
    A[Product Move API Request] --> B{Switch Type}
    B -->|ToRecurringContribution| C[ToRecurringContributionImpl]
    B -->|RecurringContributionToSupporterPlus| D[RecurringContributionToSupporterPlusImpl]
    
    C --> E[Conditional Term Renewal ✅ ADDED]
    C --> F[Subscription Update]
    
    D --> G[Term Renewal ✅ ALREADY EXISTS]
    D --> H[Subscription Update]
    
    E --> I{chargedThroughDate > termEndDate?}
    I -->|Yes| J[Extend subscription term]
    I -->|No| K[Skip term renewal]
    J --> L[Ensures removeDate ≤ termEndDate]
    K --> L
```

## Why This Approach

1. **Consistency**: The `RecurringContributionToSupporterPlus` switch type already had this logic, so we're applying the same pattern to `ToRecurringContribution`

2. **Optimal efficiency**: Instead of always performing term renewal, we only do it when `chargedThroughDate > termEndDate`, making the operation more efficient

3. **Data-driven decisions**: By adding `termEndDate` to the `GetSubscriptionResponse` model, we can make precise decisions based on actual subscription data from Zuora

4. **Minimal risk**: Term renewal with `runBilling = false` extends the subscription term without charging the customer, ensuring the operation can proceed safely

## Evidence that `termEndDate` is available

The Zuora API already returns `termEndDate` - other parts of the codebase use it:
- `digital-subscription-expiry` handler parses `termEndDate` from Zuora responses
- zuora-core has a `Subscription` model that includes `termEndDate`
- Multiple test fixtures in the codebase reference `termEndDate`

The issue was that the `product-move-api` was using its own simplified model that didn't include this field.

## Files Modified

- GetSubscription.scala
- ToRecurringContribution.scala
- ProductMoveEndpoint.scala
- Stubs.scala
- `handlers/product-move-api/src/test/scala/com/gu/productmove/endpoint/move/switchtype/ToRecurringContributionSpec.scala`
- `handlers/product-move-api/src/test/scala/com/gu/productmove/endpoint/move/switchtype/ManualTestToRecurringContribution.scala`

## Testing

- Updated unit tests to include the new `TermRenewal` dependency and `termEndDate` field
- Updated all test stubs to include realistic `termEndDate` values
- Added test scenarios for both cases: when term renewal is needed and when it's not
- Manual testing can be performed following the existing README instructions
- The fix addresses the specific production error scenario where cancellation date exceeds term end date

## Related

- This follows the same pattern as the existing `RecurringContributionToSupporterPlus` implementation
- Uses the same `TermRenewal` utility that's already proven to work in production
- Maintains compatibility with existing Zuora subscription management flows
- Aligns the `product-move-api` model with the data structure used by other handlers in the codebase

### DIFF
--- a/handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/move/switchtype/ToRecurringContribution.scala
+++ b/handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/move/switchtype/ToRecurringContribution.scala
@@ -90,10 +90,6 @@ class ToRecurringContributionImpl(
         ) *>
         termRenewal.renewSubscription(subscriptionName, runBilling = false) *>
         ZIO.log(s"Term renewal completed for subscription $subscriptionName")
-      } else {
-        ZIO.log(
-          s"No term renewal needed: chargedThroughDate $chargedThroughDate is within termEndDate ${subscription.termEndDate}",
-        )
       }
 
       updateRequestBody <- getRatePlans(
--- a/handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/move/switchtype/ToRecurringContribution.scala
+++ b/handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/move/switchtype/ToRecurringContribution.scala
@@ -82,14 +82,19 @@ class ToRecurringContributionImpl(
 
       // Check if we need term renewal to avoid "cancellation date cannot be later than term end date" error
       // We need to perform term renewal if the chargedThroughDate (which will be used as the remove date)
-      // extends beyond the current subscription term. Since we don't have access to termEndDate in the
-      // GetSubscriptionResponse, we'll conservatively always do a term renewal for this operation
-      // to ensure the subscription term covers the required dates.
-      _ <- ZIO.log(
-        s"Performing term renewal to ensure subscription term covers chargedThroughDate: $chargedThroughDate",
-      )
-      _ <- termRenewal.renewSubscription(subscriptionName, runBilling = false)
-      _ <- ZIO.log(s"Term renewal completed for subscription $subscriptionName")
+      // extends beyond the current subscription term end date.
+      needsTermRenewal = activeRatePlanCharge.chargedThroughDate.exists(_.isAfter(subscription.termEndDate))
+      _ <- if (needsTermRenewal) {
+        ZIO.log(
+          s"Performing term renewal because chargedThroughDate $chargedThroughDate is after termEndDate ${subscription.termEndDate}",
+        ) *>
+        termRenewal.renewSubscription(subscriptionName, runBilling = false) *>
+        ZIO.log(s"Term renewal completed for subscription $subscriptionName")
+      } else {
+        ZIO.log(
+          s"No term renewal needed: chargedThroughDate $chargedThroughDate is within termEndDate ${subscription.termEndDate}",
+        )
+      }
 
       updateRequestBody <- getRatePlans(
         billingPeriod,
--- a/handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/move/switchtype/ToRecurringContribution.scala
+++ b/handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/move/switchtype/ToRecurringContribution.scala
@@ -11,7 +11,7 @@ import com.gu.productmove.endpoint.move.stringFor
 import com.gu.productmove.salesforce.Salesforce.SalesforceRecordInput
 import com.gu.productmove.zuora.GetAccount.GetAccountResponse
 import com.gu.productmove.zuora.GetSubscription.{GetSubscriptionResponse, RatePlanCharge}
-import com.gu.productmove.zuora.model.SubscriptionName
+import com.gu.productmove.zuora.model.{InvoiceId, SubscriptionName}
 import com.gu.productmove.zuora.*
 import com.gu.util.config
 import sttp.tapir.*
@@ -19,6 +19,7 @@ import zio.{Clock, Task, ZIO}
 
 import java.time.LocalDate
 import java.time.format.DateTimeFormatter
+import java.time.temporal.ChronoUnit
 
 trait ToRecurringContribution {
   def run(
@@ -31,6 +32,7 @@ trait ToRecurringContribution {
 
 class ToRecurringContributionImpl(
     subscriptionUpdate: SubscriptionUpdate,
+    termRenewal: TermRenewal,
     sqs: SQS,
     stage: Stage,
 ) extends ToRecurringContribution {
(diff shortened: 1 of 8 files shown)


handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/move/switchtype/ToRecurringContribution.scala
```diff
@@ -75,11 +77,24 @@ class ToRecurringContributionImpl(
...
+      // We need to perform term renewal if the chargedThroughDate (which will be used as the remove date)
+      // extends beyond the current subscription term end date.
+      chargedThroughDate = activeRatePlanCharge.chargedThroughDate.get
+      needsTermRenewal = activeRatePlanCharge.chargedThroughDate.exists(_.isAfter(subscription.termEndDate))
+      _ <- ZIO.when(needsTermRenewal) {
+        ZIO.log(
+          s"Performing term renewal because chargedThroughDate $chargedThroughDate is after termEndDate ${subscription.termEndDate}",
+        ) *>
+          termRenewal.renewSubscription(subscriptionName, runBilling = false) *>
+          ZIO.log(s"Term renewal completed for subscription $subscriptionName")
```
johnduffell, line 90: what is `*>` for, is it a way to avoid a for comprehension and discard the result?

It does look concise and ZIO encourages it, but in the past we've shied away from ungooglable symbols where possible.

optional: Happy either way, but here is what I would have written, the main benefit is for compehension is very standard
```suggestion
        for {
          _ <- ZIO.log(
          s"Performing term renewal because chargedThroughDate $chargedThroughDate is after termEndDate ${subscription.termEndDate}",
        )
          renewal <- termRenewal.renewSubscription(subscriptionName, runBilling = false) *>
          _ <- ZIO.log(s"Term renewal completed for subscription $subscriptionName")
        } yield renewal
```

handlers/product-move-api/src/main/scala/com/gu/productmove/endpoint/move/switchtype/ToRecurringContribution.scala
```diff
@@ -75,11 +77,24 @@ class ToRecurringContributionImpl(
       // Make sure that price is valid and acceptable
       _ <- validateOldMembershipPrice(currency, billingPeriod, price)
 
+      // Check if we need term renewal to avoid "cancellation date cannot be later than term end date" error
+      // We need to perform term renewal if the chargedThroughDate (which will be used as the remove date)
+      // extends beyond the current subscription term end date.
```
johnduffell, line 82: good 👍  this comment will help people in future, as this is always a very knotty area with zuora

handlers/product-move-api/src/test/scala/com/gu/productmove/Stubs.scala
```diff
@@ -709,3 +713,36 @@ val getInvoiceItemsResponse = GetInvoiceItemsResponse(
 // Stubs for GetInvoice service
 //-----------------------------------------------------
 val getInvoiceResponse = GetInvoiceResponse(balance = 5)
+
+// Test stub for the case where term renewal is needed
+// chargedThroughDate (2024-11-15) is after termEndDate (2024-10-2)
+val getSubscriptionResponseNeedingTermRenewal = GetSubscriptionResponse(
```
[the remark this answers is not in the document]
andresilva-guardian: You are right @johnduffell , thanks for noticing it. [Here](handlers/product-move-api/src/test/scala/com/gu/productmove/HandlerSpec.scala) is the test being used now.

PR #2929, opened 2025-07-10
Fix prettier formatting commands in mparticle-api

andresilva-guardian:
**Problem:** The prettier formatting scripts were using incorrect glob patterns (`**.ts`) that weren't properly matching TypeScript files in subdirectories, causing formatting checks to pass when files actually needed formatting.

**Solution:** Updated glob patterns to use correct syntax (`**/*.ts`) and added proper quoting to ensure all TypeScript files are checked and formatted consistently.

**Changes:**
- Fixed `check-formatting` script to properly detect formatting issues
- Fixed `fix-formatting` script to format all TypeScript files
- No functional code changes, formatting-only fix

**Testing:** Verified commands now correctly identify and format files that need attention.

### DIFF
--- a/handlers/mparticle-api/package.json
+++ b/handlers/mparticle-api/package.json
@@ -7,8 +7,8 @@
     "build": "esbuild --bundle --platform=node --target=node18 --outfile=target/index.js src/index.ts",
     "lint": "eslint src/**/*.ts test/**/*.ts",
     "package": "pnpm type-check && pnpm lint && pnpm check-formatting && pnpm test && pnpm build && cd target && zip -qr mparticle-api.zip ./*.js",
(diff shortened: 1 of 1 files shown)


michaelbjacobson (approved):

johnduffell (commented):
nice one good spot that being wrong there.
As an aside (don't need to add it now): it's worth knowing about the [git ignore revs file](https://medium.com/@ramunarasinga/git-blame-ignore-revs-to-ignore-bulk-formatting-changes-f20ac23e6155) that we have when doing this kind of change.
https://github.com/guardian/support-service-lambdas/blob/f65dd11371693b55c6c1fd0d907085ec38b783f5/.git-blame-ignore-revs#L1-L14

PR #2931, opened 2025-07-11
get queue url before making request to queue via queue name

ecguardian:
## What does this change?

**Problem**
TypeScript Lambda functions in the `product-switch-api` were using queue names instead of queue URLs when sending messages to SQS. While this functionally works (messages are sent successfully), it causes the AWS SDK to log `TypeError: Invalid URL` errors for every message sent to the Queue which is very noisy. This PR sets to fix this issue by providing the correct Queue URL.

**What happens:**
- AWS SDK detects invalid URL format
- Logs TypeError: Invalid URL to CloudWatch
- SDK automatically calls GetQueueUrl behind the scenes
- Message sends successfully with resolved URL
- Logs are polluted with false error messages

**Log Example**
Here is an example of the error in the [logs](https://eu-west-1.console.aws.amazon.com/cloudwatch/home?region=eu-west-1#logsV2:log-groups/log-group/$252Faws$252Flambda$252Fproduct-switch-api-PROD/log-events$3FfilterPattern$3DTypeError$26start$3D-86400000) use membership credentials in Janus to see the logs.

**Solution**
Add an explicit `GetQueueUrl` API call to resolve queue names to proper SQS URLs before message sending, preventing the AWS SDK from performing this resolution internally and logging errors. The Lambda already has the right permissions to make a call to get the Queue URL. 

Changes:
- Added an SQS module
- Created integration test for sqs module
- Update all @aws-sdk/* packages to use consistent versions
- Fix TypeScript errors in S3 client mocking tests
- Ensures test suite runs without type errors

**Trello Ticket**
`Update support-service-lambdas to use queue url instead of name` - [link](https://trello.com/c/va3eIrKx)

**Testing**
- ✅ All tests in the AWS module are passing
- ✅ Error handling tests correctly catch and validate error cases
- ✅ S3 client mocking tests now work without TypeScript errors


handlers/product-switch-api/src/supporterProductData.ts
```diff
@@ -48,11 +52,18 @@ export const sendToSupporterProductData = async (
 		`Sending supporter product data message ${messageBody} to queue ${queueName}`,
 	);
 
+	const getQueueUrlCommand = new GetQueueUrlCommand({
+		QueueName: queueName,
+	});
+	const { QueueUrl } = await client.send(getQueueUrlCommand);
```
ecguardian, line 58: the lambda has permission to get the Queue URL in the CDK setup file, so we can use the Queue Name to get the Queue URL using GetQueueUrlCommand

modules/aws/package.json
```diff
@@ -10,11 +10,11 @@
 		"@aws-sdk/client-cloudwatch": "3.777.0",
 		"@aws-sdk/client-lambda": "^3.826.0",
 		"@aws-sdk/client-s3": "3.777.0",
+		"@aws-sdk/client-sqs": "3.787.0",
 		"@aws-sdk/client-ssm": "^3.777.0",
```
johnduffell, line 14: could this cause any issue if they are on different versions?

modules/aws/src/sqs.ts
```diff
@@ -0,0 +1,35 @@
...
+		if (!QueueUrl) throw new Error('Queue URL not found');
+
+		const command = new SendMessageCommand({
+			QueueUrl,
+			MessageBody: messageBody,
+		});
+		const response = await client.send(command);
+		return response;
+	} catch (error) {
+		console.error(error);
```
johnduffell, line 32: do you think we need to catch and then log/rethrow it? I would normally assume whatever ultimately catches it would log it already.

modules/aws/test/s3.test.ts
```diff
@@ -1,12 +1,17 @@
-import {
-	GetObjectCommand,
-	PutObjectCommand,
-	S3Client,
-} from '@aws-sdk/client-s3';
-import { mockClient } from 'aws-sdk-client-mock';
+import { S3Client } from '@aws-sdk/client-s3';
```
johnduffell, line 1: this file is the S3 tests - I'm not sure if you meant to change these perhaps as part of making them consistent with the new tests?

PR #2934, opened 2025-07-14
Negative invoices processor - Fix query field alias 

david-pepper:
## What does this change?
Fix query field alias to resolve this error in production:

image


andrade0 (approved):
all good ;)

PR #2944, opened 2025-07-17
Add Supporter Plus Student Plan to Product Catalog

rupertbates:
## What does this change?
This PR adds the new Supporter Plus one year fixed plan for Students into the code which generates the product catalog and also regenerates the Zod schema to include it.


maloky (approved):
LGTM 👍

PR #2945, opened 2025-07-17
Enhance Zuora HTTP response handling and streamline type system architecture

david-pepper:
## What does this change?

This PR introduces two key improvements to the typescript global zuora module :
- **Enhanced Zuora HTTP response handling** - Consolidates error handling patterns and improves response validation
- **Streamlined type system architecture** - Establishes a cleaner hierarchy from shared schemas to handler-specific types

**Enhanced Zuora HTTP response handling**
- Enhance ZuoraClient module: 
  - Improve success detection by enabling fetch function to handle multiple failure responses by adding checks for httpSuccess (callout status) and logicalSuccess (authentication, zuora business logic e.g. trying to apply credit to an account from an invoice that doesn't exist)
  - enable fetch function to inspect the response body for a variety of error responses and return an appropriate error message
  - extract ZuoraError class and put it into its own module
  - add a ZuoraErrorHandler module to generate errors appropriate to the response from Zuora

**Streamlined type system architecture**

- Create Generic Schema Factory Functions e.g. for query response schema
```
export const createQueryResponseSchema = <T extends z.ZodRawShape>(
    recordShape: T,
) => {
    return z.object({
        done: z.boolean(),
        size: z.number(),
        records: z.array(z.object(recordShape)).optional(),
    }).strict();
};
```
- Use Type-Safe Generic Functions with Defaults
```
export const getPaymentMethods = async <
    T extends z.ZodType = typeof DefaultPaymentMethodResponseSchema,
>(
    zuoraClient: ZuoraClient,
    accountId: string,
    schema?: T,
): Promise<z.infer<T>> => { ... }
```

- Introduced Hierarchical Schema Composition e.g. use a base schema (zuoraResponseSchema) and extend for custom use
```
export const PaymentMethodResponseSchema = zuoraResponseSchema.extend({
    creditcard: z.array(PaymentMethodSchema).optional(),
    // ... other payment method arrays
});
```
- Introduced Schema Inheritance and Reuse e.g. creating BasePaymentMethodSchema and then selecting subset fields for specific usage
```
export const PaymentMethodSchema = z.object({
    ...BasePaymentMethodSchema.pick({
        id: true,
        status: true,
        type: true,
        isDefault: true,
    }).shape,
});
```
- Centralized Type Definitions by consolidating all Zuora response types in httpResponse.ts

This architecture ensures type safety flows from schemas → runtime validation → compile-time types, eliminating duplication and maintaining consistency across the codebase.

### Testing
- Changes have been verified by executing the negative invoices processor and verifying expected behaviour for invoice processing. The negative invoices processor uses all of the functionality that has been enhanced in this PR, as can be seen in the relevant changes to its modules.
- addition of automated testing for zuoraClient.test.ts

### Notes
- There are a lot of changed files (and lines of code) in this PR, more than I would usually include. It started out with an objective to improve how Zuora error responses were being handled, because zuora has several (inconsistent) error response structures and codes, and then there was a need to begin updating the types that were being used at the zuora client common module, which propagated into changes to the negative invoices processor (which is how I've tested the changes as working correctly).
- Updated negative invoices processor handler return types to focus on **business logic results** rather than mixing HTTP request details with domain-specific data at the same object level

**Next steps**
In this PR I have address improvements that can primarily be validated by running the negative invoices processor. There is more similar work to to across the whole of the zuora module, which I will pick up next.


johnduffell:

PR #2960, opened 2025-07-25
illustration of dependency injection of side effects plus tweaks

johnduffell:
TODO:
1. resolve/address all outstanding comments on https://github.com/guardian/support-service-lambdas/pull/2920 that could not be completed before merging
2. decide whether also to implement dependency injection as illustrated in this PR

---

This PR is a demo on top of https://github.com/guardian/support-service-lambdas/pull/2920

specifically this comment https://github.com/guardian/support-service-lambdas/pull/2920#discussion_r2213738806

The main change is to show how we can inject the external dependencies as parameters similarly how we do with zuora client 
https://github.com/guardian/support-service-lambdas/blob/db568862b899ab115fa1a5ddf16d6abdaf99d5b0/modules/zuora/src/zuoraClient.ts
This is partly to make more pure functions which helps testability, but also readability as it makes it clearer which functions are going to do API calls/side effects and which are not.

For zuora API you may notice we also tend to split each API call into a separate file also, this is because zuora has a LOT of APIs available, but I didn't go to those lengths in this PR as we only call a few APIs.

See inline comments for more.  I didn't have a chance to check if it compiles or the tests work, but I hope it's useful enough to get an idea.

### DIFF
--- a/handlers/mparticle-api/src/apis/data-subject-requests.ts
+++ b/handlers/mparticle-api/src/apis/data-subject-requests.ts
@@ -3,9 +3,11 @@ import type { DataSubjectRequestForm } from '../../interfaces/data-subject-reque
 import type { DataSubjectRequestState } from '../../interfaces/data-subject-request-state';
 import { DataSubjectRequestStatus } from '../../interfaces/data-subject-request-state';
 import type { DataSubjectRequestSubmission } from '../../interfaces/data-subject-request-submission';
-import { getAppConfig, getEnv } from '../utils/config';
+import { AppConfig } from '../utils/config';
 import { makeHttpRequest } from '../utils/make-http-request';
 import type { HttpResponse } from '../utils/make-http-request';
+import { z } from 'zod';
+import { callbackPath, placeholder } from '../routers/http';
 
 function parseDataSubjectRequestStatus(
 	status: 'pending' | 'in_progress' | 'completed' | 'cancelled',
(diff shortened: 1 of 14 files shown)


handlers/mparticle-api/src/apis/data-subject-requests.ts
```diff
@@ -22,87 +24,118 @@ function parseDataSubjectRequestStatus(
...
-			 * The DSR API is secured via basic authentication. Credentials are issued at the level of an mParticle workspace.
-			 * You can obtain credentials for your workspace from the Workspace Settings screen. Note that this authentication
-			 * is for a single workspace and scopes the DSR to this workspace only.
-			 * https://docs.mparticle.com/developers/apis/dsr-api/v3/#authentication
-			 */
-			Authorization: `Basic ${Buffer.from(`${appConfig.workspace.key}:${appConfig.workspace.secret}`).toString('base64')}`,
-		},
-		body: options.body,
-	});
+export class MParticleDataSubjectClient {
```
johnduffell, line 27: I based this and the Events client on the [zuora client](https://github.com/guardian/support-service-lambdas/blob/db568862b899ab115fa1a5ddf16d6abdaf99d5b0/modules/zuora/src/zuoraClient.ts#L17-L16), each class has a  static create method and only handles making generic GET /POST etc with authentication and zod support.

Then the individual calls provide the relative url and the schema on top of that.

handlers/mparticle-api/src/apis/data-subject-requests.ts
```diff
@@ -22,87 +24,118 @@ function parseDataSubjectRequestStatus(
...
-						],
-			group_id: form.userId, // Let's group by User Unique Id to group all requests related to that user (max 150 requests per group)
-			extensions: {
-				'opendsr.mparticle.com': {
-					skip_waiting_period: true,
-				},
+		},
+		api_version: '3.0',
+		status_callback_urls: [
+			httpRouterBaseUrl + callbackPath.replace(placeholder, form.requestId),
```
johnduffell, line 105: rather than hard coding the url, I represented the dependency in the code, so the compiler will keep them in step (or stop it compiling)

handlers/mparticle-api/src/index.ts
```diff
@@ -3,20 +3,31 @@ import type {
...
 	APIGatewayProxyResult
 > = async (event: APIGatewayProxyEvent): Promise<APIGatewayProxyResult> => {
 	try {
+		const { mParticleDataSubjectClient, mParticleEventsClient, stage } =
+			await services();
+		const router = httpRouter(
+			mParticleDataSubjectClient,
+			mParticleEventsClient,
+			baseUrlForStage(stage),
+		);
```
johnduffell, line 27: now we hide all the side effecting code away in `services()` and pass in the relevant constructions into the router.
So we know that code elsewhere can't have any side effects other than what we passed in.

handlers/mparticle-api/src/routers/http.ts
```diff
@@ -18,30 +18,49 @@ import {
...
-	}),
-	createRoute<unknown, EventBatch>({
-		httpMethod: 'POST',
-		path: '/events',
-		handler: uploadEventBatchHandler(),
-		parser: eventBatchParser,
-	}),
-]);
+export const placeholder = '{requestId}';
+export const callbackPath = `/data-subject-requests/${placeholder}/callback`;
```
johnduffell, line 25: by extracting these, we can do basic reverse routing (i.e. build the path for a known route) in a type safe way

handlers/mparticle-api/src/utils/make-http-request.ts
```diff
@@ -68,7 +71,7 @@ export async function makeHttpRequest<T>(
 		}
 
 		try {
-			const data = (await response.json()) as T;
+			const data = options.schema.parse(await response.json());
```
johnduffell, line 74: now we parse using zod instead of a cast.

PR #2975, opened 2025-07-29
Refactor: Consolidate typescript account functions into a single module

david-pepper:
This PR is one in a series of PRs dedicated to consolidating object-specific functions into a single object-specific module within the zuora typescript common module

## Overall Objective
Combine all object-related functions into a single module so that:

- **Improved Developer Experience**
Developers working with subscriptions only need to import from one location instead of remembering which specific file contains each function

- **Logical Cohesion**
Functions are conceptually related

- **Reduced Import Complexity**
Instead of 3+ import statements for subscription-related functionality, there'll be a single import

- **Better Discoverability**
When looking for subscription functionality, there's one obvious place to look rather than searching across multiple files

- **Easier Maintenance**
Related functions are co-located, making it easier to understand dependencies and make coordinated changes

### Changes in this PR
Combines all account-related functions into a single file and updates all references accordingly.

### Testing
Automated build will fail if any of the references have been broken.

### DIFF
--- a/handlers/discount-api/src/discountEndpoint.ts
+++ b/handlers/discount-api/src/discountEndpoint.ts
@@ -3,7 +3,6 @@ import { ValidationError } from '@modules/errors';
 import { Lazy } from '@modules/lazy';
 import { getIfDefined } from '@modules/nullAndUndefined';
 import type { Stage } from '@modules/stage';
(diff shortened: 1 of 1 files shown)


andrade0 (approved):
LGTM

PR #2966, opened 2025-07-28
write-off-unpaid-invoices leaving behind credit balance on accounts

andrade0:
## What does this change?

Enhances the write-off invoices handler to apply existing account credit balances to unpaid invoices before using invoice item adjustments. This resolves cases where accounts had both credit balances and unpaid invoices, leading to persistent autocancellation alerts.

The handler now:

Checks if the account has available credit balance
Applies credit to positive invoice balances using applyCreditToAccountBalance
Falls back to existing invoice item adjustment logic for any remaining balance

## How to test

On PROD: Find accounts with both unpaid invoices and credit balances - observe credit balances remain unused after write-off.

On this branch: Same accounts should have credit balances properly applied to invoices, with remaining balances handled via item adjustments.

## How can we measure success?

Reduction in accounts with persistent credit balances
Fewer daily autocancellation alerts
Better account balance reconciliation in Zuora

## Have we considered potential risks?

Added comprehensive error handling for "Cannot find entity" scenarios and environment mismatches. If credit application fails, the process falls back to existing logic. Enhanced logging provides visibility into all steps.

## Accessibility

-   [ ] [Tested with screen reader](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#screen-reader)
-   [ ] [Navigable with keyboard](https://github.com/guardian/accessibility/blob/main/people-and-technology/02-physical.md#keyboard)
-   [ ] [Colour contrast passed](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#contrast)
-   [ ] [The change doesn't use only colour to convey meaning](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#use-of-colour)

PR #2985, opened 2025-07-29
Refactor invoice item adjustment types

david-pepper:


graham228221 (approved):
👍 

might cause some conflicts in https://github.com/guardian/support-service-lambdas/pull/2966 - but they can be mitigated there

PR #3000, opened 2025-08-04
Dp - zuora client refactor

david-pepper:


johnduffell:
I guess this was an experiement, either way it is old, has conflicts, and probably isn't going to be merged without lots of work

PR #3005, opened 2025-08-04
Use new zuora client schema - Negative invoices processor - doRefund

david-pepper:
### What does this change?
Updates the Negative Invoices Processor's doRefund functionality to use the new Zuora client schema:

- update structure of lambda return value
- update error detection to check for presence of error rather than success property

### Testing
Changes validated by simulating an error in refund path:
Screenshot 2025-08-04 at 17 17 04


handlers/negative-invoices-processor/src/types/shared/refund.ts
```diff
@@ -1,15 +1,14 @@
...
-export const RefundResponseSchema = zuoraResponseSchema.extend({
+export const RefundResponseSchema = z.object({
 	Id: z.string().optional(),
 });
 
 export type RefundResponse = z.infer<typeof RefundResponseSchema>;
 
 export const RefundResultSchema = RefundResponseSchema.extend({
-	refundAttempt: RefundResponseSchema,
+	refundAttempt: RefundResponseSchema.optional(),
```
rupertbates, line 11: Is this needed at all now or could it go?

david-pepper: This now contains only the Id of the successful refund. Technically, yes it could go since we don't need it for anything. I'll get rid, and if for some reason we need it in future we can bring it back.

handlers/negative-invoices-processor/src/handlers/doCreditBalanceRefund.ts
```diff
@@ -37,29 +34,28 @@ export const handler = async (
...
 			...parsedEvent,
 			refundResult: {
-				refundAttempt: response,
 				paymentMethod: paymentMethodToRefundTo,
 				refundAmount,
 			},
 		};
 	} catch (error) {
-		return {
+		const abc = {
```
rupertbates, line 47: Could this have a more meaningful name?

david-pepper: just a temp const for console.logging

modules/zuora/src/refund.ts
```diff
@@ -1,15 +1,13 @@
 import { z } from 'zod';
-import { zuoraResponseSchema } from './types';
+import { zuoraSuccessSchema } from './types';
 import type { ZuoraClient } from './zuoraClient';
 
-export const doRefund = async <
-	T extends z.ZodType = typeof zuoraResponseSchema,
->(
+export const doRefund = async <T extends z.ZodType = typeof zuoraSuccessSchema>(
```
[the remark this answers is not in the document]
david-pepper: Yes I think we are ok to remove this

rupertbates (approved):
Looks good 👍

handlers/negative-invoices-processor/src/types/shared/refund.ts
david-pepper: I've reconsidered this. I think it makes sense to keep a record of the successful refund Id, so will re-add it, both for logging purposes as well as an explicit success indicator

rupertbates: Why both the refundAttempt and the the Id field optional though, won't they always be there?

handlers/negative-invoices-processor/src/types/shared/refund.ts
```diff
@@ -2,13 +2,13 @@ import { z } from 'zod';
...
 export const RefundResponseSchema = z.object({
-	Id: z.string().optional(),
+	Id: z.string(),
 });
 
 export type RefundResponse = z.infer<typeof RefundResponseSchema>;
 
 export const RefundResultSchema = z.object({
-	refundAttempt: RefundResponseSchema.optional(),
+	refundId: z.string().optional(),
```
rupertbates, line 11: Can also be non-optional I think?

johnduffell:
this is old, has conflicts, and probably isn't going to be merged without more work, feel free to reopen if it's useful still

PR #3006, opened 2025-08-05
Use new zuora client schema   negative invoices processor - apply credit to account balance

david-pepper:


johnduffell:

PR #3009, opened 2025-08-06
make baton SAR handler copy file into baton S3 bucket on completion

johnduffell:
As part of the wider mparticle CDP work, mparticle acts as a data store under GDPR.  This means that we need to support erasure and subject access requests.

Functionality was added to the mparticle-api lambda to support SAR via mparticle's API in this PR https://github.com/guardian/support-service-lambdas/pull/2993

mparticle returns a results url https://docs.mparticle.com/developers/apis/dsr-api/v3/#example-response-body:~:text=a%20download%20link%20to%20the%20request%20results%20data
this is passed back to baton directly.

However at present the URL provided by mparticle is not understood by baton, which requires an S3 url https://github.com/guardian/baton/?tab=readme-ov-file#:~:text=a%20list%20of%20S3%20paths%20to%20the%20result%20file(s)

This PR updates the lambda to fetch the results from mparticle's URL, and writes them to the baton bucket, before letting baton know the file name.

Note: as it stands, baton zips up the files provided.  However since mparticle provides a zip file, this will end up with a doubly zipped file.  There is a draft PR on baton to make it notice if there's a single zip file, and pass it through without double zipping it.  https://github.com/guardian/baton/pull/393


cdk/lib/mparticle-api.ts
```diff
@@ -28,6 +28,15 @@ export class MParticleApi extends GuStack {
 			'/accountIds/baton',
 		).stringValue;
 
+		const sarResultsBucket = StringParameter.fromStringParameterName(
+			this,
+			'SarResultsBucket',
+			`/${this.stage}/${this.stack}/${this.app}/sarResultsBucket`,
```
johnduffell, line 34: TODO need to add this parameter to SSM.  It is used here to set the permissions, and in the lambda code to write to the bucket

cdk/lib/mparticle-api.ts
```diff
@@ -28,6 +28,15 @@ export class MParticleApi extends GuStack {
 			'/accountIds/baton',
 		).stringValue;
 
+		const sarResultsBucket = StringParameter.fromStringParameterName(
+			this,
+			'SarResultsBucket',
+			`/${this.stage}/${this.stack}/${app}/sarResultsBucket`,
```
johnduffell, line 34: added this parameter to SSM as it appears to be private.  It is used here to set the permissions, and in the lambda code to write to the bucket
It is the same as used in the zuoraSar lambda, see https://gu-reader-revenue-private.s3.eu-west-1.amazonaws.com/membership/support-service-lambdas/PROD/zuoraSar-PROD.v1.json
https://github.com/guardian/support-service-lambdas/blob/2219d969f121229515dbf5bab64350fd74ea965f/handlers/zuora-sar/src/main/scala/com/gu/zuora/sar/ConfigLoader.scala#L6-L10

values are here: 
https://eu-west-1.console.aws.amazon.com/systems-manager/parameters/?region=eu-west-1&tab=Table#list_parameter_filters=Name:Contains:%2Fsupport%2Fmparticle-api%2FsarResultsBucket

handlers/mparticle-api/src/routers/baton/handle-sar-status.ts
```diff
@@ -24,18 +28,39 @@ export async function handleSarStatus(
...
+	public handleSarStatus = async (request: BatonSarEventStatusRequest) => {
+		const dataSubjectRequestState: DataSubjectRequestState =
+			await getStatusOfDataSubjectRequest(request.initiationReference);
+
+		const resultLocations = dataSubjectRequestState.resultsUrl
+			? [
+					await this.fetchToS3(
+						dataSubjectRequestState.resultsUrl,
+						request.initiationReference,
+					),
```
johnduffell, line 52: this will fetch it to a (new) file in S3 every single time a status request is done with a completed state.
This may not be ideal, TODO check in the baton code whether it will keep requesting the status once it has returned "completed" once.
Alternative option is to give this lambda List access to the bucket, then it can check the file every time and only copy if it doesn't exist.

cdk/lib/mparticle-api.ts
```diff
@@ -28,27 +29,45 @@ export class MParticleApi extends GuStack {
...
+		const sarS3BaseKey = '/mparticle-results';
+
 		// HTTP API Lambda
 		const httpLambda = new GuLambdaFunction(this, `${app}-http-lambda`, {
 			app,
 			memorySize: 1024,
 			fileName: `${app}.zip`,
 			runtime: nodeVersion,
 			timeout: Duration.seconds(15),
+			loggingFormat: LoggingFormat.TEXT,
```
johnduffell, line 48: this is the standard in this repo.  Maybe we should extract an SrLambda in the same way we have for the SrAlarm so we don't have to duplicate as much.
https://github.com/guardian/support-service-lambdas/pull/2660#discussion_r1934039965

handlers/mparticle-api/package.json
```diff
@@ -13,7 +13,8 @@
   },
   "devDependencies": {
     "@faker-js/faker": "^9.8.0",
-    "@types/aws-lambda": "^8.10.147"
+    "@types/aws-lambda": "^8.10.147",
+    "@aws-sdk/client-s3": "3.859.0"
```
johnduffell, line 17: this is so the integration test can check that the (fake) SAR file was downloaded into the right bucket.

handlers/mparticle-api/src/routers/baton.ts
```diff
@@ -20,25 +23,30 @@ export function validateRequest(data: BatonEventRequest): BatonEventRequest {
...
+					case 'initiate':
+						return handleSarInitiate(validatedEvent);
+					case 'status':
+						return new HandleSarStatus(
+							sarResultsBucket,
+							sarS3BaseKey,
+						).handleSarStatus(validatedEvent);
+				}
+				break; // unreachable - only needed for no-fallthrough rule
+			case 'RER':
```
johnduffell, line 43: I wanted to make it clearer that the "else" is actually `RER`.  I was avoiding a comment as it isn't checked by the compiler, but this way eslint threw a spurious complaint about fallthrough.  intellij correctly realises the `break` is unreachable code, but linting is a build failure...

handlers/mparticle-api/src/apis/mparticleClient.ts
```diff
@@ -0,0 +1,110 @@
...
+	post<REQ, RESP>(
+		path: string,
+		body: REQ,
+		schema: z.ZodType<RESP, z.ZodTypeDef, unknown>,
+	): Promise<HttpResponse<RESP>>;
+
+	getStream(path: string): Promise<ReadableStream>;
+}
+
+export const mparticleDataSubjectBaseURL = 'https://opendsr.mparticle.com/v3';
```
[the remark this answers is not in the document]
johnduffell: actually I didn't need to use regex at all, it was just a bad trick to make it less cumbersome.  Changed to use normal string manupulation.

handlers/mparticle-api/interfaces/data-subject-request-submission.ts
```diff
@@ -17,9 +17,4 @@ export interface DataSubjectRequestSubmission {
 	 * The controller-provided identifier of the request in a GUID v4 format.
 	 */
 	requestId: string;
-
-	/**
-	 * A unique ID representing the data controller. mParticles sets this to the workspace ID.
-	 */
-	controllerId: string;
```
johnduffell, line 24: this seemed to be unused so I trimmed it out to save clutter in the tests

handlers/mparticle-api/src/apis/data-subject-requests.ts
```diff
@@ -22,94 +21,76 @@ function parseDataSubjectRequestStatus(
...
+	isProd: boolean,
 	form: DataSubjectRequestForm,
 ): Promise<DataSubjectRequestSubmission> => {
-	const response = await requestDataSubjectApi<{
-		expected_completion_time: Date;
-		received_time: Date;
-		encoded_request: string;
-		subject_request_id: string;
-		controller_id: string;
-	}>(`/requests`, {
```
johnduffell, line 68: this type has been moved to be a zod schema on line 65, which is then used to parse the response by the mparticle client

handlers/mparticle-api/src/apis/data-subject-requests.ts
```diff
@@ -22,94 +21,76 @@ function parseDataSubjectRequestStatus(
...
  * A request in the OpenDSR format communicates a Data Subject’s wish to access or erase their data.
  * The OpenDSR Request takes a JSON request body and requires a Content-Type: application/json header.
  * https://docs.mparticle.com/developers/apis/dsr-api/v3/#submit-a-data-subject-request-dsr
+ * @param mParticleDataSubjectClient
  * @param {DataSubjectRequestForm} form - The form containing the data subject request details.
  * @returns https://docs.mparticle.com/developers/apis/dsr-api/v3/#example-success-response-body
  */
 export const submitDataSubjectRequest = async (
+	mParticleDataSubjectClient: MParticleClient<DataSubjectAPI>,
+	isProd: boolean,
```
johnduffell, line 35: selected dependencies are passed in to make it easier to type safely run parts of the system in tests or scripts

handlers/mparticle-api/src/apis/events.ts
```diff
@@ -58,7 +32,8 @@ export const uploadAnEventBatch = async (
 			context: batch.context,
 			ip: batch.ip,
 		},
-	});
+		z.string(),
```
johnduffell, line 35: this is a bit of  a hack - a 202 response has no response body, so fetch returns an empty string.  Since `post` anticipates a body, we pass in a string schema which would not do much.
Perhaps we should make schema optional, and it would use a schema that expects an empty string only and returns undefined in that case?

handlers/mparticle-api/src/apis/mparticleClient.ts
```diff
@@ -0,0 +1,110 @@
...
+		schema: z.ZodType<RESP, z.ZodTypeDef, unknown>,
+	): Promise<HttpResponse<RESP>>;
+
+	post<REQ, RESP>(
+		path: string,
+		body: REQ,
+		schema: z.ZodType<RESP, z.ZodTypeDef, unknown>,
+	): Promise<HttpResponse<RESP>>;
+
+	getStream(path: string): Promise<ReadableStream>;
```
johnduffell, line 31: this is the new method, using it we can get a binary stream from mparticle api using the relevant credentials.

handlers/mparticle-api/src/routers/baton/handle-sar-status.ts
```diff
@@ -2,40 +2,75 @@ import type { DataSubjectRequestState } from '../../../interfaces/data-subject-r
...
+		// https://docs.mparticle.com/developers/apis/dsr-api/v3/#example-response-body:~:text=is%203.0.-,results_url,-string
+
+		// The trusted base url is automatically added in the mparticleClient.
+		// This means there's no chance of accidentally sending our credentials to an untrusted URL.
+		// Since we have a full url here, we need to strip off the base.
+		const path = stripBaseUrl(dataSubjectRequestState.resultsUrl);
+		const stream = await mParticleDataSubjectClient.getStream(path);
+		const s3Url = await srs3Client.write(initiationReference, stream);
+		resultLocations = [s3Url];
+	}
```
johnduffell, line 55: this is the new logic needed for the download/upload

handlers/mparticle-api/src/index.ts
```diff
@@ -9,32 +9,76 @@ import {
...
-	} catch (error) {
-		console.error('HTTP handler error:', error);
-		return {
-			statusCode: 500,
-			body: JSON.stringify({ error: 'Internal server error' }),
-		};
-	}
-};
+export const handlerHttp: Handler<APIGatewayProxyEvent, APIGatewayProxyResult> =
+	withLogging(
```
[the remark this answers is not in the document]
johnduffell: exit tracing
```
2025-08-14T11:18:33.949Z	123-1345-12312-123123-123231	INFO	TRACE handlerBaton EXIT SHORT_ARGS {
  event: {
    initiationReference: '6fd4c21e-a661-464b-8388-b09bf81604fc',
    requestType: 'SAR',
    action: 'status'
  }
} RESULT {
  requestType: 'SAR',
  action: 'status',
  status: 'completed',
  message: 'mParticle Request Id: "6fd4c21e-a661-464b-8388-b09bf81604fc". Expected completion time: 2025-08-14T00:00:00.000Z.',
  resultLocations: [
    's3://my-bucket-name/mparticle-results/2025-08-14T11:18:33.840Z-6fd4c21e-a661-464b-8388-b09bf81604fc.zip'
  ]
}
```

handlers/mparticle-api/src/index.ts
```diff
@@ -9,32 +9,76 @@ import {
...
-	} catch (error) {
-		console.error('Baton handler error:', error);
-		throw error; // Re-throw to trigger Lambda retry mechanism
-	}
-};
+// this must be the same base key as we have permissions set in the CDK
+const sarS3BaseKey = 'mparticle-results/';
+
+export const handlerBaton: Handler<BatonEventRequest, BatonEventResponse> =
+	withLogging(async (event: BatonEventRequest): Promise<BatonEventResponse> => {
```
johnduffell, line 44: entry tracing:
```
2025-08-14T11:18:27.791Z	1231231-1345-12312-123-1231231	INFO	TRACE handlerBaton ENTRY ARGS {
  event: {
    initiationReference: '6fd4c21e-a661-464b-8388-b09bf81604fc',
    requestType: 'SAR',
    action: 'status'
  },
  arg1: {
    callbackWaitsForEmptyEventLoop: [Getter/Setter],
    succeed: [Function (anonymous)],
    fail: [Function (anonymous)],
    done: [Function (anonymous)],
    functionVersion: '$LATEST',
    functionName: 'mparticle-api-baton-CODE',
    memoryLimitInMB: '1024',
    logGroupName: '/aws/lambda/mparticle-api-baton-CODE',
    logStreamName: '2025/08/14/[$LATEST]123123123123',
    clientContext: undefined,
    identity: undefined,
    invokedFunctionArn: 'arn:aws:lambda:eu-west-1:12345:function:mparticle-api-baton-CODE',
    awsRequestId: '123123-1345-2312-323-123',
    tenantId: undefined,
    getRemainingTimeInMillis: [Function: getRemainingTimeInMillis]
  },
  arg2: [Function (anonymous)]
}
```

johnduffell: it's not perfect, but hopefully it's better than nothing, and we can iterate on it

handlers/mparticle-api/src/index.ts
```diff
@@ -9,32 +9,76 @@ import {
...
+			config.pod,
+		),
+		srs3Client: new SRS3ClientImpl(
+			config.sarResultsBucket,
+			sarS3BaseKey,
+			() => new Date(),
+		),
+		isProd: stage === 'PROD',
+	};
+}
```
johnduffell, line 84: by creating the default http/aws clients in the top level handler and using dependency injection, we can easily pass in alternatives in the tests in a type safe and discoverable way.  However there is a certain amount of clutter having to pass things around.

handlers/mparticle-api/src/routers/baton/handle-sar-status.ts
```diff
@@ -2,40 +2,74 @@ import type { DataSubjectRequestState } from '../../../interfaces/data-subject-r
...
+		// The trusted base url is automatically added in the mparticleClient.
+		// This means there's no chance of accidentally sending our credentials to an untrusted URL.
+		// Since we have a full url here, we need to strip off the base.
+		const path = stripBaseUrl(
+			dataSubjectRequestState.resultsUrl,
+			mParticleDataSubjectClient.baseURL,
+		);
+		const stream = await mParticleDataSubjectClient.getStream(path);
+		const s3Url = await srs3Client.write(initiationReference, stream);
+		resultLocations = [s3Url];
```
johnduffell, line 53: this is the important additional code that pipes the file from mparticle to AWS.

handlers/mparticle-api/src/routers/http.ts
```diff
@@ -19,29 +24,38 @@ import {
...
+			path: '/data-subject-requests/{requestId}/callback',
+			handler: dataSubjectRequestCallbackHandler(mParticleDataSubjectClient),
+			parser: dataSubjectRequestCallbackParser,
+		}),
+		createRoute<unknown, EventBatch>({
+			httpMethod: 'POST',
+			path: '/events',
+			handler: uploadEventBatchHandler(mParticleEventsAPIClient),
+			parser: eventBatchParser,
+		}),
```
johnduffell, line 60: I couldn't work out where most of these were used from so it was harder to test.  Are they redundant code from previous iterations?  Also, the callback handler is currently a no-op, I'm not sure if that's a placeholder for future work, or also not needed?

handlers/mparticle-api/src/routers/http/submit-data-subject-request.ts
```diff
@@ -26,7 +35,9 @@ export function submitDataSubjectRequestHandler() {
 		 * https://docs.mparticle.com/guides/data-subject-requests/#erasure-request-waiting-period
 		 */
 		try {
+			// FIXME only set for erasure (not SAR?)
```
johnduffell, line 38: I notied that we set the attribute unconditionally, but this endpoint can be used for SAR (portability and access) requests.  So I think that might be a mistake.  However I am not sure where this endpoint is called from, if anywhere.

handlers/mparticle-api/src/utils/config.ts
```diff
@@ -12,6 +12,7 @@ export const ConfigSchema = z.object({
 		secret: z.string(),
 	}),
 	pod: z.string(),
+	sarResultsBucket: z.string(),
```
johnduffell, line 15: https://eu-west-1.console.aws.amazon.com/systems-manager/parameters/?region=eu-west-1&tab=Table#list_parameter_filters=Name:Contains:%2Fsupport%2Fmparticle-api%2FsarResultsBucket

handlers/mparticle-api/src/utils/make-http-request.ts
```diff
@@ -25,66 +29,103 @@ export type HttpResponse<T> =
...
-					errorBody = await response.text();
-				} catch {
-					/* empty */
+export class RestRequestMaker {
+	constructor(
+		public baseURL: string,
+		private headers: Record<string, string>,
+	) {}
+
+	makeRESTRequest = withLogging(
```
johnduffell, line 45: withLogging http call
```
2025-08-14T11:18:27.932Z	1111-22222-3333-4444	INFO	TRACE HTTP https://opendsr.mparticle.com/v3 ENTRY ARGS {
  method: 'GET',
  path: '/requests/6fd4c21e-a661-464b-8388-b09bf81604fc',
  schema: '(ZodObject not expanded)'
}
```

johnduffell: response
```
2025-08-14T11:18:28.204Z	11111-222222-333333-44444	INFO	TRACE HTTP https://opendsr.mparticle.com/v3 EXIT SHORT_ARGS {
  method: 'GET',
  path: '/requests/6fd4c21e-a661-464b-8388-b09bf81604fc'
} RESULT {
  success: true,
  data: {
    controller_id: '1402',
    expected_completion_time: 2025-08-14T00:00:00.000Z,
    subject_request_id: '6fd4c21e-a661-464b-8388-b09bf81604fc',
    request_status: 'completed',
    results_url: 'https://opendsr.mparticle.com/v3/access_result/6fd4c21e-a661-464b-8388-b09bf81604fc'
  }
}
```

handlers/mparticle-api/test/handle-sar-statusIntegration.test.ts
```diff
@@ -0,0 +1,137 @@
+/**
+ * check that we can copy from an http url to an s3 bucket on completion of an mparticle SAR
```
johnduffell, line 2: this can only run with membership credentials, but it's useful if you want to test the download/upload without mparticle or baton account credentials, as it does the normal functionality but reads from a dummy url and writes to an alternative bucket.

modules/aws/package.json
```diff
@@ -12,6 +12,7 @@
 		"@aws-sdk/client-s3": "^3.848.0",
 		"@aws-sdk/client-ssm": "^3.848.0",
 		"@aws-sdk/credential-provider-node": "^3.848.0",
+		"@aws-sdk/lib-storage": "^3.848.0",
```
johnduffell, line 15: needed for S3 binary streaming https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/Package/-aws-sdk-lib-storage/

handlers/mparticle-api/src/apis/srs3Client.ts
```diff
@@ -0,0 +1,31 @@
...
+		private readonly sarS3BaseKey: string,
+		private readonly now: () => Date,
+	) {}
+
+	write = async (
+		reference: string,
+		stream: ReadableStream,
+	): Promise<string> => {
+		// include date for uniqueness and reference to aid debugging.
+		const fileName = this.now().toISOString() + '-' + reference + '.zip';
```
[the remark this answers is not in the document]
johnduffell: that's ok, this Client is a specific zip file uploader for this purpose.  It's purpose is to wrap the bucket/key and current time into one object for easier dependency injection and mocking.  Maybe the naming could be improved though.

johnduffell:
tested all four request types in CODE, all seemed to work ok.  See the logs attached below:
The fifth request in the log is a SAR status request, however that request was completed more than 7 days ago, so it got a "410 Gone" when it tried to access the file.  This is expected - baton would need to download the file before that point.

https://eu-west-1.console.aws.amazon.com/cloudwatch/home?region=eu-west-1#logsV2:log-groups/log-group/$252Faws$252Flambda$252Fmparticle-api-baton-CODE/log-events/2025$252F08$252F14$252F$255B$2524LATEST$255D1f9d09f24c33462b823d5079986df7c0

If there are issues with the file disappearing before baton checks the status, we can make some changes to the mechanism to avoid this:
- on the callback from mparticle we can copy the file immediately
- when the "status" endpoint is called, we can populate the file with where it is expected to be in S3. (we would have to remove the timestamp from the name)

handlers/mparticle-api/runManual/invokeHandlerSARStatus.ts
```diff
@@ -0,0 +1,62 @@
...
+
+	const mockDate = () => new Date(0);
+	const batonS3Writer = new BatonS3WriterImpl(
+		sarResultsBucket,
+		sarS3BaseKey,
+		mockDate,
+	);
+
+	batonRerRouter(
+		//TODO call the child not the router
```
johnduffell, line 50: https://github.com/guardian/support-service-lambdas/pull/3023

handlers/mparticle-api/src/routers/baton.ts
andresilva-guardian: I agree, the more explicit, the better 💪

handlers/mparticle-api/src/apis/srs3Client.ts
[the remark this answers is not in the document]
andresilva-guardian: @johnduffell What happens if the file is named "file.ZIP" instead of "file.zip"?

handlers/mparticle-api/src/routers/baton/handle-sar-status.ts
andresilva-guardian: @johnduffell Implemented [here](https://github.com/guardian/support-service-lambdas/pull/3041/files#diff-4f7a5962523bcdb4905e27174cfa83732ff9dad6649de18e8b5e1d28ad977f82R68). Now we check for the file first. Also we are downloading the file right on the callback moment so we no longer incur on the risk of expiration after the first 30 days. What do you think?

PR #3015, opened 2025-08-11
raise cancellation refund invoices today rather than backdated

johnduffell:
When someone cancels their S+ within 14 days, we refund them automatically.

When they have set up a direct debit it can take up to a week to actually make the refund due to limitations of the DD system (`base can't be refunded because it is pending_submission`)

We handle this by retrying the refund lambda for up to a week.

We have moved the invoice generation to the async lambda for perf reasons in https://github.com/guardian/support-service-lambdas/pull/2826

It always uses the date of cancellation as both the target date and invoice date of the date of cancellation when generating the invoice.

Note that the invoice date must always be in the currently open accounting period.  Otherwise an error occurs `ZuoraError(ACCOUNTING_PERIOD_CLOSED,Invoices cannot be generated in a closed Accounting Period. Please enter another invoice date.)`

The problem at the moment is that since things may retry for a week, the billing call may fail due to the accounting period being closed.

The invoice date is only to show the date on the invoice, rather than to decide what the charges should be.  This means we should use the current date as the invoice date, to ensure it has not been closed.

See here for a definion of the different dates: https://knowledgecenter.zuora.com/Zuora_Billing/Bill_your_customers/AB_Automate_billing_document_generation/Bill_runs/AB_Create_bill_runs#:~:text=Run%20Dates%20section.-,Invoice%20Date%3A%20The%20invoice%20date%20displayed%20on%20the%20invoices.,default%2C%20the%20target%20date%20is%20equal%20to%20the%20bill%20run%20date.,-Creating%20a%20bill

This PR changes the run billing operation to use today's date for the invoice, rather than backdating it to the date of cancellation.  This will usually be the same, but it will avoid the `ACCOUNTING_PERIOD_CLOSED` error where it's on a retry since the period was closed.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -25,7 +25,7 @@ val scala2Settings = Seq(
 )
 
 val scala3Settings = Seq(
-  scalaVersion := "3.7.1",
+  scalaVersion := "3.7.2",
   version := "0.0.1",
   organization := "com.gu",
   scalacOptions ++= Seq(
--- a/handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/RunBilling.scala
+++ b/handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/RunBilling.scala
@@ -17,12 +17,15 @@ private class RunBillingLive(zuoraGet: ZuoraGet) extends RunBilling {
   import RunBilling.*
 
   // https://developer.zuora.com/v1-api-reference/older-api/operation/Action_POSTgenerate/
-  override def run(accountId: ZuoraAccountId, today: LocalDate): Task[Option[InvoiceId]] =
+  override def run(accountId: ZuoraAccountId, today: LocalDate, targetDate: LocalDate): Task[Option[InvoiceId]] = {
+    // make the invoice date today, in case there's been a delay and the accounting period has closed
+    val accountToGenerate = AccountToGenerate(accountId, today, targetDate)
+    val runBillingRequest = RunBillingRequest(List(accountToGenerate), "Invoice")
     for {
       errorOrRunBillingResponse <- zuoraGet
         .post[RunBillingRequest, RunBillingResponse](
           uri"action/generate",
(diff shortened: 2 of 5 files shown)


build.sbt
```diff
@@ -25,7 +25,7 @@ val scala2Settings = Seq(
 )
 
 val scala3Settings = Seq(
-  scalaVersion := "3.7.1",
+  scalaVersion := "3.7.2",
```
johnduffell, line 28: opprtunistic bump

handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/RunBilling.scala
```diff
@@ -17,12 +17,15 @@ private class RunBillingLive(zuoraGet: ZuoraGet) extends RunBilling {
   import RunBilling.*
 
   // https://developer.zuora.com/v1-api-reference/older-api/operation/Action_POSTgenerate/
-  override def run(accountId: ZuoraAccountId, today: LocalDate): Task[Option[InvoiceId]] =
+  override def run(accountId: ZuoraAccountId, today: LocalDate, targetDate: LocalDate): Task[Option[InvoiceId]] = {
+    // make the invoice date today, in case there's been a delay and the accounting period has closed
+    val accountToGenerate = AccountToGenerate(accountId, today, targetDate)
```
johnduffell, line 22: this is the important bit, rather than using `today` (which was actually the date it was added to the queue) for both values, it uses the actual today when it's invoked for the InvoiceDate
https://github.com/guardian/support-service-lambdas/blob/dbd400fe1827a9d8d2044ad58931de7da6db5052/handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/RunBilling.scala#L41

johnduffell:
once merged - I redrove the queue `product-switch-refund-dead-letter-PROD`
and the two messages were processed immediately and the refunds appeared correctly on both zuora subscriptions.

PR #3019, opened 2025-08-13
Add isDeliveryProduct to product catalog

rupertbates:
## What does this change?
- Add a deliveryProducts array and associated type which definitively records which products have a delivery element. 
- Add a productKeys array to match the `ProductKey` type to make runtime checks easier.
- Add a field `isDeliveryProduct` to the product catalog so that consumers of the API can tell which products have a delivery element.

### DIFF
--- a/modules/product-catalog/src/generateSchema.ts
+++ b/modules/product-catalog/src/generateSchema.ts
@@ -6,7 +6,7 @@ import type {
 	ZuoraProductRatePlan,
 	ZuoraProductRatePlanCharge,
 } from '@modules/zuora-catalog/zuoraCatalogSchema';
-import { deliveryProducts } from '@modules/product-catalog/productCatalog';
+import { isDeliveryProduct } from '@modules/product-catalog/productCatalog';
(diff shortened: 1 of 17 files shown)


modules/product-catalog/src/generateSchema.ts
```diff
@@ -47,10 +55,12 @@ const generateZuoraProductSchema = (product: CatalogProduct) => {
 	const ratePlanSchema = supportedRatePlans.map((productRatePlan) =>
 		generateProductRatePlanSchema(productRatePlan),
 	);
+	const isDeliveryProduct = deliveryProducts.includes(productName);
```
paul-daniel-dempsey, line 58: I got a type error here, perhaps this would work?
```suggestion
	const isDeliveryProduct = deliveryProducts.includes(productName);
```

paul-daniel-dempsey (approved):
Just one comment probably needs looking at. Have not looked at this code-base much on my own, mainly due to it being Scala, being Typescript now gives me more confidence to start approaching it again, thankyou. Understand the changes, would'nt mind a pointer in PR how to test it though (looking through readme now tbh)

PR #3021, opened 2025-08-14
Add mandatory product specific fields to the product purchase schema

rupertbates:
## What does this change?
The product-catalog defines a schema and type for valid product and rateplan combinations which is used by support-workers to ensure type safety when dealing with subscription purchases.

This PR extends that schema to also include fields which are mandatory for a particular type of purchase, for instance any delivery product requires a first delivery date and a delivery contact (address + name).


maloky (approved):
LGTM 👍

PR #3024, opened 2025-08-14
mparticle api - move schemas and types to where they are used

johnduffell:
Follows on from the mobbing review on https://github.com/guardian/support-service-lambdas/pull/3009

Previously the API client functions for the mparticle API were all together, and the schemas were all in a separate file.

This made it hard to pick out which schemas related to which calls, and make the API calls file quite long and unweidly.

There was also (and still is) some business logic in with the API client code.

The zuoraClient elsewhere in this repo already uses a layout where the API calls are self contained, which makes it easier to understand each REST call in isolation.

This PR makes a folder in place of each API client file, and puts each individual API call in its own file within.  The relevant schemas are moved into their associated file.  (Note that I have only moved whole functions or schemas, I haven't extracted or changed any logic - if I also changed logic, it would make it almost impossible to review those changes.)

See inline comments for relevant details.

The code is more cohesive if the size of each file is minimised, and the number of imports in each file (and the amount of `..` in each import) are also minimised.

I have also followed a similar principle for the baton router (but not the http router at this stage)


handlers/mparticle-api/src/apis/dataSubjectRequests/getStatus.ts
```diff
@@ -0,0 +1,62 @@
+import { z } from 'zod';
```
johnduffell, line 1: each of the things under dataSubjectRequests is one of the functions from the "deleted" file above, together with its associated schemas etc.

handlers/mparticle-api/src/routers/baton/access/schema.ts
```diff
@@ -0,0 +1,10 @@
+import { z } from 'zod';
```
johnduffell, line 1: this file contains the schemas relating to access requests in general, but not to either one of initiate or status

handlers/mparticle-api/src/routers/shared/addErasureExclusionAttributes.ts
```diff
@@ -0,0 +1,27 @@
+import type {
```
johnduffell, line 1: this was more like business logic that didn't belong to either the http or baton routers, so I made a shared home for it here.  I wasn't sure the exact naming that made sense so I went for "shared" for now i.e. shared between the routers.

handlers/mparticle-api/src/apis/dataSubjectRequests/getStatus.ts
```diff
@@ -0,0 +1,100 @@
...
+export enum DataSubjectRequestStatus {
+	Pending = 'pending',
+	InProgress = 'in-progress',
+	Completed = 'completed',
+	Cancelled = 'cancelled',
+}
+
+/**
+ * Get the status of an OpenDSR request
+ * https://docs.mparticle.com/developers/apis/dsr-api/v3/#example-response-body
```
johnduffell, line 16: this links to a page where the keys are_snake_case rather than this camelCase.

andresilva-guardian (approved):

MaelGNM (approved):
Sensible changes 👍

PR #3028, opened 2025-08-20
feat: Add a new customer facing name field to the Product Catalog JSON

paulbrown1982:
## What does this change?
Adds a new field to the Product Catalog JSON containing the user-facing name for a product. This can then become a source of truth, like the benefits list is in user-benefits API, and used in various channel applications such as:

- User Benefits API
- Salesforce
- Support Frontend
- Manage My Account

It was inspired after members of the Platform team visited the contact centre and realised CSRs were constantly translating between the various Zuora/legacy names and their current ones. This change is a first stepping stone to help resolve that by displaying the same exact names in the checkout, manage my account, help centre and Salesforce. 

Eventually the names will be stored in a custom field in Zuora for full CMS-ability, but for now they are hardcoded.

## How to test

All covered by extensive unit tests, of which only the snap file was changed in this PR.

Subsequent PRs are raised in the applications above.

Some apps might not be able to parse the new JSON. Testing of the above 4 apps in CODE after this is deployed in CODE is required.

## Trello

https://trello.com/c/F5qXyVPK/603-add-customer-facing-name-to-product-catalog-service

### DIFF
--- a/modules/product-catalog/src/generateProductCatalog.ts
+++ b/modules/product-catalog/src/generateProductCatalog.ts
@@ -6,7 +6,10 @@ import type {
 	ZuoraProductRatePlanCharge,
 } from '@modules/zuora-catalog/zuoraCatalogSchema';
 import type { ProductCatalog } from '@modules/product-catalog/productCatalog';
-import { isDeliveryProduct } from '@modules/product-catalog/productCatalog';
+import {
+	getCustomerFacingName,
+	isDeliveryProduct,
+} from '@modules/product-catalog/productCatalog';
 import { stripeProducts } from '@modules/product-catalog/stripeProducts';
 import {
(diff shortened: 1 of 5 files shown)


modules/product-catalog/src/productCatalog.ts
```diff
@@ -41,6 +41,33 @@ export function isDeliveryProduct(
 	return deliveryProducts.includes(productKey as DeliveryProductKey);
 }
 
+// Eventually all but OneTimeContribution will come from a custom field in Zuora's Product Catalog
```
paulbrown1982, line 44: The snap file is the only other place the customer facing names are repeated in the codebase.

rupertbates (approved):
Great 👍

PR #3041, opened 2025-08-28
Add streaming functionality to S3 tests and enhanced SAR download experience

andresilva-guardian:
## What does this change?

This PR introduces several improvements to the mParticle API Lambda to enhance performance, reduce AWS costs, and improve code maintainability. All of these changes are based on the suggestions from @johnduffell made on the https://github.com/guardian/support-service-lambdas/pull/3009.

### 1. File Naming Strategy Migration for SAR Results
- **Problem**: The current timestamp-based file naming (`{sarS3BaseKey}{timestamp}-{reference}.zip`) causes redundant downloads, storage waste, and complex lookups for every status check
- **Solution**: Migrated to deterministic file naming (`{sarS3BaseKey}{reference}.zip`) with direct file existence checking using S3 HeadObject API
- **Impact**: We only download the file once now.

### 2. Introduction of SrLambda Standard Construct
- **Problem**: Repetitive Lambda configuration code across the repository with potential for inconsistency
- **Solution**: Created a standardized `SrLambda` CDK construct that provides sensible defaults while allowing customization
- **Benefits**: Consistent configuration, reduced boilerplate, centralized maintenance of Lambda standards. This could be reused for other support service lambdas.

### 3. Enhanced S3 File Handling
- Added file existence checks before downloading SAR results
- Improved S3 permissions to include `ListBucket` action for better file management
- Integrated `BatonS3Writer` for more efficient file handling

### 4. Code Architecture Improvements
- Better separation of concerns in file handling logic
- Enhanced unit test coverage for S3 operations
- Improved error handling and logging

## How to test

1. **File naming migration**: 
   - On this branch, trigger a data subject request and verify that SAR files are named deterministically without timestamps
   - Check subsequent status requests don't re-download existing files

2. **SrLambda construct**:
   - Deploy the changes and verify Lambda functions are created with expected default configurations
   - Test custom overrides (timeout, memory) still work as expected

3. **Performance improvements**:
   - Compare Lambda execution times for repeated status checks between main and this branch
   - Monitor S3 storage growth patterns to confirm stabilization

- **Reduced S3 operations**: Monitor CloudWatch metrics for S3 PutObject operation count (should decrease by ~50% for repeated status checks)
- **Faster status responses**: Lambda execution time should improve for repeated status requests on the same data subject request
- **Lower costs**: Reduced S3 transfer and storage costs due to eliminating redundant file operations
- **Improved maintainability**: Future Lambda configurations will be more consistent and require less boilerplate code

- **Backwards compatibility**: The file naming change is backwards compatible as existing files continue to work
- **Rollback plan**: Can easily revert to original `BatonS3WriterImpl` if issues arise, while keeping beneficial CDK permissions
- **Storage considerations**: Deterministic naming means files are overwritten rather than versioned, but this is acceptable since content should be identical for the same request
- **Performance monitoring**: Will monitor key metrics post-deployment to ensure expected improvements are realized

## Images

Not applicable - this is a backend performance and architecture improvement.

## Accessibility

Not applicable - this change does not affect user-facing interfaces.


ecguardian (approved):
Nice 👍

johnduffell:
I've just reviewed the description so far.

> File Naming Strategy Migration for SAR Results

Sounds good your proposal, hopefully it was clear from my original PRs that the timestamp was only there for convenience, and the existing zuorasar lambda actually does the logic that you have now implemented.
I did check the Baton code and it seemed like the code was not going to call our lambda multiple times, hence we would get away with my implementation.  However on reflection I think the benefits of your stated approach will be better.

cdk/lib/cdk/sr-lambda.ts
```diff
@@ -0,0 +1,49 @@
+import type { GuStack } from '@guardian/cdk/lib/constructs/core';
+import { GuLambdaFunction } from '@guardian/cdk/lib/constructs/lambda';
+import { Duration } from 'aws-cdk-lib';
+import { LoggingFormat, type Runtime } from 'aws-cdk-lib/aws-lambda';
+import { nodeVersion } from '../node-version';
+
+// Define the props interface based on actual usage patterns in the codebase
```
johnduffell, line 7: I think that comment is more useful as a PR comment than in the codebase
```suggestion
```

handlers/mparticle-api/src/services/batonS3Writer.ts
```diff
@@ -1,24 +1,60 @@
...
 		readonly sarResultsBucket: string,
 		private readonly sarS3BaseKey: string,
-		private readonly now: () => Date,
 	) {}
 
+	/**
+	 * Generate a deterministic S3 key based on reference.
+	 * Format: {baseKey}{reference}.zip
+	 * This avoids timestamp-based naming which can cause lookup issues.
+	 */
```
johnduffell, line 20: I think the implementation is clear enough
```suggestion
```

handlers/mparticle-api/src/services/batonS3Writer.ts
```diff
@@ -1,24 +1,60 @@
...
+	private generateS3Key(reference: string): string {
+		return `${this.sarS3BaseKey}${reference}.zip`;
+	}
+
+	/**
+	 * Check if a file already exists for the given reference.
+	 * Returns the S3 URL if found, null if not found.
+	 * Uses deterministic naming for direct lookup.
+	 */
+	checkIfFileExists = async (reference: string): Promise<string | null> => {
```
johnduffell, line 30:
```suggestion
	getUrlIfExists = async (reference: string): Promise<string | null> => {
```

handlers/mparticle-api/test/services/batonS3Writer.test.ts
```diff
@@ -0,0 +1,104 @@
...
+	const bucketName = 'test-bucket';
+	const s3BaseKey = 'sar-results/';
+
+	let batonS3Writer: BatonS3WriterImpl;
+
+	beforeEach(() => {
+		jest.resetAllMocks();
+		batonS3Writer = new BatonS3WriterImpl(bucketName, s3BaseKey);
+		console.log = jest.fn();
+		console.warn = jest.fn();
```
johnduffell, line 29: is this to prevent logging in the tests?  on the face of it I would have expected it to be useful to keep it

modules/aws/src/s3FileExists.ts
```diff
@@ -0,0 +1,37 @@
...
+}): Promise<boolean> => {
+	try {
+		const command = new HeadObjectCommand({
+			Bucket: bucketName,
+			Key: filePath,
+		});
+		await s3Client.send(command);
+		return true;
+	} catch (error: any) {
+		// If the error is NoSuchKey or NotFound, the file doesn't exist
```
johnduffell, line 25: it's clear from the code
```suggestion
```

johnduffell (commented):
Thanks for the PR, I have made a few comments and questions on the code itself.  I think the SRLambda construct will be helpful and the naming changes are good.

See the individual comments, but I think before merging we should clarify whether the added complexity of the HEAD checks is worth it, or whether we should make it more determinstic by blindly downloading it on either the callback or status check.

handlers/mparticle-api/src/routers/http/dataSubjectRequestCallback/data-subject-request-callback.ts
```diff
@@ -82,21 +89,89 @@ export function dataSubjectRequestCallbackHandler(
...
-} => {
+	downloadResult?: string;
+}> => {
 	// Just log this information so we can have track of it on Cloud Watch
 	console.info('Process Data Subject Request Callback from mParticle', {
 		requestId,
 		form: payload,
 	});
 
+	// If the request is completed and has a results_url, download the file
```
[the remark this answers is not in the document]
andresilva-guardian: My concern is about link expiration @johnduffell as I believe that mparticle mentions that the link provided by them expires in 30 days. So if something fails on Baton side, the link will stop work and the user has to make another SAR request and wait again up to 14 days.

modules/aws/test/s3FileExists.test.ts
```diff
@@ -0,0 +1,73 @@
+import { S3Client } from '@aws-sdk/client-s3';
```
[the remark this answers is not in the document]
andresilva-guardian: I believe that having more tests does not hurt and that we also have the integration tests defined on the mparticle lambda so we have all covered.

> I realise the equation is skewed a bit by the fact that the integration tests are not run automatically (yet! - that's something we need to fix).

Let's open a new pull request to handle it?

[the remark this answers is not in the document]
johnduffell: I don't disagree that more tests have a benefit but there's definitely a cost as well - the time to build and run them not to mention maintenance.  Some people do go for 100% code coverage by some metric which is a valid strategy, and they accept that the benefit of knowing nothing is missed outweighs the extra cost.  However that can even distract from more appropriate forms of testing.  Ultimately testing is about mitigating risk and providiing confidence for developers to make changes.

Regarding automating the integration tests it's indeed a separate PR, I've only ever managed to automate one set of integration tests here which was the scala support-workers one, and although it works well it is a bit hacky.  Perhaps there is a better way now with github actions and restricted credentials?

handlers/mparticle-api/src/routers/http/dataSubjectRequestCallback/data-subject-request-callback.ts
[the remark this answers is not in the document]
johnduffell: yes in fact I think it's only 7 days before the link expires.  With the current implementation if baton doesn't call the status endpoint within that time, then the file will be lost and the request would have to be tried again as you mention.
Personally I still feel like simpler is better but it sounds like you have thought a lot more about the failure modes of baton and especially having the work complete already, let's stick with your changes.

johnduffell (approved):
thanks for following up on these things, the reminaing comments are for you to decide how to use them, and potentially also simplify the PR description. 👏

PR #3043, opened 2025-09-01
Fix the schema generation script

rupertbates:
## What does this change?
https://github.com/guardian/support-service-lambdas/pull/3028 added a new field to the product catalog and schema, but I think it manually edited the schema rather than using the generation script to do it. 

When I next ran the script it outputs a schema which doesn't compile because it includes literals without quotation marks. I actually think for this field we don't need a literal anyway, we can just use a string as we don't need strict type checking for what is just a display value.


paulbrown1982 (approved):
👍 Thank you! I went to-and-fro many times. around whether to make that field a literal or string.

PR #3044, opened 2025-09-01
Ensure that generated product catalog is parseable by the current Zod schema

rupertbates:
## Why are you making this change?
The generate-product-catalog lambda is triggered by changes to the raw Zuora catalog export file in S3. When it runs it takes that file and builds the json file which is returned by the product catalog api here: 
https://product-catalog.guardianapis.com/product-catalog.json 
or here: 
https://product-catalog.code.dev-guardianapis.com/product-catalog.json

We also have a Zod schema which is used by a number of projects including the user-benefits api and the product-switch api to parse the product catalog and give us a strongly typed representation of our product catalog.

There have been a number of production incidents over the last few months when the Zuora catalog has changed in a way that makes the generated product catalog incompatible with the schema used by these apps; for instance the removal of a charge from the Supporter Plus `OneYearStudent` rate plan caused a Zod parsing error:
```typescript
_ZodError: [{
  "code": "invalid_type",
  "expected": "object",
  "received": "undefined",
  "path": [
    "SupporterPlus",
    "ratePlans",
    "OneYearStudent",
    "charges",
    "Contribution"
  ],
  "message": "Required"
}]
```
When these issues occurred it meant that the apps which use the Zod schema are unable to function affecting end users.

## What this PR does
This PR adds a check within the generate-product-catalog lambda that the catalog it has generated is parseable by the current version of the Zod schema. If not it will discard it and trigger an alarm which will allow us to investigate what has gone wrong. 

While this means that the catalog may be out of date, it is far better than breaking production systems which depend on being able to parse the product catalog.

## How to test
- Deploy generate-product-catalog lambda from this branch to code
- Enable alarm actions for the alarm named 'The generate-product-catalog-CODE lambda generated a catalog which cannot be parsed by the current schema' (these will be disabled by default in CODE)
- In the sandbox version of Zuora deactivate the USD pricing for the 'Supporter Plus - One Year Student' rate plan 
Screenshot 2025-09-02 at 14 39 04
- Watch the logs for the generate-product-catalog lambda, you should see the error message 'ERROR The generated product catalog did not pass validation in the current Zod schema' appear, this should also trigger an alarm notification (it will come to the P&E/SR/SRE channel, I _think_ because it's a CODE alarm)
- Check the [product catalog in code](https://product-catalog.code.dev-guardianapis.com/product-catalog.json), the USD pricing should still be present in the OneYearStudent rate plan because the previous product catalog has not been overwritten
- Reactivate the USD pricing in Zuora and the alarm should go back to normal

## Alarm
The alarm is called 'The generate-product-catalog-{STAGE} lambda generated a catalog which cannot be parsed by the current schema' and notifications will go the the Portfolio Alarms chat channel

### DIFF
--- a/cdk/lib/__snapshots__/generate-product-catalog.test.ts.snap
+++ b/cdk/lib/__snapshots__/generate-product-catalog.test.ts.snap
@@ -432,7 +432,7 @@ Check the logs at https://eu-west-1.console.aws.amazon.com/cloudwatch/home?regio
         "EvaluationPeriods": 1,
         "MetricName": "failed-schema-validation",
         "Namespace": "support-service-lambdas",
-        "Period": 3600,
+        "Period": 60,
         "Statistic": "Sum",
         "Tags": [
           {
@@ -1178,7 +1178,7 @@ Check the logs at https://eu-west-1.console.aws.amazon.com/cloudwatch/home?regio
         "EvaluationPeriods": 1,
         "MetricName": "failed-schema-validation",
         "Namespace": "support-service-lambdas",
-        "Period": 3600,
+        "Period": 60,
         "Statistic": "Sum",
         "Tags": [
           {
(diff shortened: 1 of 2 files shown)


tjmw (approved):
👏  This is great

PR #3047, opened 2025-09-02
Stop holiday-stop-api sending duplicated authorisation headers in Salesforce requests

graham228221:
### Problem background
In the Salesforce sandbox we were seeing an issue where the holiday-stop-api would get a 400 error from Salesforce for some callouts. The 400 error was quite unusual, it was returning a html page as if it was a UI error:

Screenshot 2025-09-02 at 15 55 07

On further investigation, this seems to be related to the authentication headers added to the requests. Requests from holiday-stop-api have two headers:

1. `Authorization` - used for a bearer token e.g. `Bearer abcdef.123456789`
2. `X-SFDC-Session` - used for a session ID passed in from Salesforce to allow for the holiday-stop-api to actually "masquerade" as the CSR using the function so that additions/edits appear to made by them instead of an integration user.

The session ID is not always provided, only when the service is called from the Salesforce UI - i.e. from MMA no session ID is provided.

But holiday-stop-api actually sends BOTH headers. They are both populated by the session ID if provided, if not they are both populated with an oauth access token. Salesforce seems to be happy to work out what it should actually use.

We found that holiday-stop-api, when receiving the session id in the initial request, is actually sending two versions of each header - one version with the session id and one version with the oauth access token 😵 here's a Cloudwatch log line with some extra logging added to show the headers sent:

Screenshot 2025-09-02 at 15 59 08

I have been able to duplicate this unusual 400 error by using the same duplicate headers in postman.

If we forced Salesforce to not send the session ID, the errors would stop - because only one version of each header is then sent.

> [!WARNING]
> What I cannot explain is how this is working in production. It also must have been working in the sandbox until very recently - I was actually testing another change up until Thursday last week and I'm sure it was working! Unless this is all a weird dream. Maybe Salesforce made some sort of change??
>
> It's possible this is somehow related to the `INVALID_SESSION_ID` errors that we've long experienced with holiday-stop-api but never quite figured out?

### Solution
We always make sure to only send one version of the authorisation headers on each request.

## How to test
In Salesforce, use the Holiday Stop edit UI to make an amendment. Previously this would fail to load because holiday-stop-api needs to query SF. Withdrawals were also broken.

The Holiday Stop UI should be functional again in the sandbox AND we should avoid this problem appearing in production.

This should not impact holiday stops made through MMA which are not provided with the `X-Ephemeral-Salesforce-Access-Token` header.

In Salesforce, we should be able to verify that this hasn't completely broken the UI quickly. We can be ready to roll back if there any problems.

### DIFF
--- a/lib/effects/src/main/scala/com/gu/effects/Http.scala
+++ b/lib/effects/src/main/scala/com/gu/effects/Http.scala
@@ -27,11 +27,10 @@ object Http extends Logging {
 
     { request: Request =>
       val maybeBodySummary = Option(request.body).map(bodySummary)
-      val headersSummary = request.headers.asScala.map { h => s"${h.getFirst}: ${h.getSecond}" }.mkString(", ")
       logger.info(
-        s"HTTP request: ${request.method} ${request.url}" +
-          maybeBodySummary.map(summary => s", body:  $summary").getOrElse("") +
-          s", headers: {$headersSummary}" // TODO:delete comment - Logging request headers
+        s"HTTP request: ${request.method} ${request.url}" + maybeBodySummary
+          .map(summary => s", body:  $summary")
+          .getOrElse(""),
       )
       val response = restClient.newCall(request).execute
       logger.info(s"HTTP response: ${response.code}")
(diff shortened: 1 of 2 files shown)


johnduffell:
good spot, that's a bit random, well done for getting to the bottom of it!

PR #3049, opened 2025-09-03
Sort any arrays in product-catalog file generation

rupertbates:
## What does this change?

We now have three generated files within the product-catalog module:
- a Zod schema for the product catalog
- a Zod schema for all valid purchase options
- a file defining all valid billing periods for each product 

Up until now the ordering of elements in arrays within these generated files was undefined, which meant that diffs could be noisy when new versions were generated. This PR explicitly sorts any generated arrays so that they will retain the same ordering between versions.


modules/product-catalog/src/generateProductBillingPeriods.ts
```diff
@@ -25,7 +25,7 @@ const getBillingPeriodsForProduct = (
 				(billingPeriod) =>
 					billingPeriod !== null && billingPeriod != 'Specific_Weeks',
 			) as string[],
-	);
+	).sort((a, b) => a.localeCompare(b));
```
tjmw, line 28: Was there a reason to use `localeCompare` here instead of the default `sort` behaviour?

tjmw (approved):
Great!

modules/product-catalog/src/generateProductBillingPeriods.ts
rupertbates: No actually, it's not needed here 👍

PR #3055, opened 2025-09-05
Remove the contribution charge from S+ student plan

rupertbates:
## What does this change?
This PR removes the `Contribution` charge from the `OneYearStudent` rate plan of the Supporter Plus product in the product catalog schema.

This charge is not needed because we will never be adding a contribution payment to the student plan, it was added by accident when creating the new plan.

This change only removes the charge from the Zod schema used by client applications to parse the product catalog, this is necessary before removing the charge from Zuora to avoid parsing errors. Once it is deployed then we can remove the charge from Zuora.


modules/product-catalog/src/productCatalogSchema.ts
```diff
@@ -1604,9 +1604,6 @@ export const productCatalogSchema = z.object({
 			OneYearStudent: z.object({
 				billingPeriod: z.literal('Annual'),
 				charges: z.object({
-					Contribution: z.object({
-						id: z.string(),
-					}),
```
johnduffell, line 1609: so this is a "writes" schema and we parse the original catalog somehow and then only write the fields that existing in this schema?

johnduffell (approved):
don't quite understand how the schema relates to the processing of the catalog but I can't see that there's any issue with the change 👍

PR #3061, opened 2025-09-10
Add last 4, card type and expiration date to Stripe payment method

rupertbates:
## What does this change?
https://github.com/guardian/support-frontend/pull/7159 removed some fields from the `CreditCardReferenceTransaction` payment method object which is used to create a Zuora payment method from a Stripe token. They were
- last 4 credit card digits
- credit card type
- expiration month
- expiration year

This was done because those fields do not exist in the Zuora documentation for a `CreditCardReferenceTransaction` in either the old [subscribe and amend API](https://developer.zuora.com/v1-api-reference/older-api/operation/Action_POSTsubscribe/) or the newer [orders API](https://developer.zuora.com/v1-api-reference/api/operation/POST_Order/), so it was assumed that they were unused. 

However those fields do exist in the `CreditCard` payment method (the payment method used when creating a payment method with raw credit card information). It seems that setting these fields on a CreditCardReferenceTransaction does still store them in Zuora and in fact we rely on these being there for some functionality on manage my account.

While relying on undocumented behaviour in an API doesn't seem like best practice, this behaviour has been in place for a number of years now and is consistent between the two APIs so it seems safe to add back in which this PR does.

## How to test
See [this support-frontend PR](https://github.com/guardian/support-frontend/pull/7272)

### DIFF
--- a/modules/zuora/src/createSubscription/createSubscription.ts
+++ b/modules/zuora/src/createSubscription/createSubscription.ts
@@ -120,7 +120,7 @@ function buildCreateSubscriptionRequest<T extends PaymentMethod>(
 	return {
 		newAccount: newAccount,
 		orderDate: zuoraDateFormat(dayjs()),
-		description: `Created in execution ${createdRequestId} by support-workers, using createSubscription.ts in support-service-lambdas`,
+		description: `Created by createSubscription.ts in support-service-lambdas`,
 		subscriptions: [
 			{
 				orderActions: [createSubscriptionOrderAction],


modules/zuora/src/createSubscription/createSubscription.ts
```diff
@@ -120,7 +120,7 @@ function buildCreateSubscriptionRequest<T extends PaymentMethod>(
 	return {
 		newAccount: newAccount,
 		orderDate: zuoraDateFormat(dayjs()),
-		description: `Created by createSubscription.ts in support-service-lambdas`,
+		description: `Created in execution ${createdRequestId} by support-workers, using createSubscription.ts in support-service-lambdas`,
```
johnduffell, line 123: good idea although is this not also/intended to be called by things outside of support-workers?

modules/zuora/src/orders/paymentMethods.ts
```diff
@@ -2,6 +2,10 @@ export type CreditCardReferenceTransaction = {
 	type: 'CreditCardReferenceTransaction';
 	tokenId: string;
 	secondTokenId: string;
+	cardNumber: string;
+	cardType?: string;
```
johnduffell, line 6: (this is optional because stripe might have a card type that zuora doesn't know about)

rupertbates: That's right, it definitely does have.

PR #3062, opened 2025-09-12
add context logging to product-switch-api and improve loggers

johnduffell:
we had a lot of alarms over the last couple of days on product-switch-api, due to a known bug trying to switch with a pending amount update.

When looking at the logs, I could search for the error and find 72 instances, but it was hard to tell if they were for the same subscription repeatedly or not.

In discount-api I added a context logger in https://github.com/guardian/support-service-lambdas/pull/2507

This PR
- extends the logger to product-switch-api.
- makes the logger a global const with mutable context (should be ok for single threaded lambdas and saves typing)
- uses the standard logger.wrapFn to log the inputs and outputs of the Router and zuoraClient.fetch
- updates the logger to provide line number and function as an auto prefix (see below)

image

discount-api https://eu-west-1.console.aws.amazon.com/cloudwatch/home?region=eu-west-1#logsV2:log-groups/log-group/$252Faws$252Flambda$252Fdiscount-api-CODE/log-events/2025$252F09$252F15$252F$255B$2524LATEST$255D0a464ffd3ad9446bb699236579952f6d
product-switch-api https://eu-west-1.console.aws.amazon.com/cloudwatch/home?region=eu-west-1#logsV2:log-groups/log-group/$252Faws$252Flambda$252Fproduct-switch-api-CODE/log-events/2025$252F09$252F15$252F$255B$2524LATEST$255Dead863767117405eae18683334fb3eae
mparticle-api https://eu-west-1.console.aws.amazon.com/cloudwatch/home?region=eu-west-1#logsV2:log-groups/log-group/$252Faws$252Flambda$252Fmparticle-api-baton-CODE/log-events/2025$252F09$252F15$252F$255B$2524LATEST$255D426a4206d0e74ebfbf7b022dcf80a8b8

Tested in CODE and looks good for both discount and product-switch APIs as well as mparticle-api (handlers/mparticle-api/runManual/invokeHandlerRERStatusCODE.ts)

Potential future work (other PRs)
- convert more lambdas (use source maps and logger instead of console)
- look into making zuoraClient and mparticle client both use the make-http-request client (with logging)


handlers/metric-push-api/src/index.ts
```diff
@@ -14,13 +15,11 @@ const buildResponse = (statusCode: number) => ({
 	headers: { 'Cache-Control': 'private, no-store' },
 });
 
-const router = new Router([
+export const handler = Router([
 	{
 		httpMethod: 'GET',
 		path: '/metric-push-api',
 		handler: async (event: APIGatewayProxyEvent) => {
-			console.log(`Input is ${JSON.stringify(event)}`);
```
johnduffell, line 22: standard entry/exit logging for the lambda happens in the router now

handlers/metric-push-api/src/index.ts
```diff
@@ -14,13 +15,11 @@ const buildResponse = (statusCode: number) => ({
 	headers: { 'Cache-Control': 'private, no-store' },
 });
 
-const router = new Router([
+export const handler = Router([
```
johnduffell, line 18: `Router` is now a function that returns a function, so we don't need a separate handler function that just calls routeRequest.

modules/product-catalog/tsconfig.json
```diff
@@ -5,14 +5,4 @@
...
-  "compilerOptions": {
-    "paths": {
-      "@modules/aws/*": ["../aws/src/*"],
-      "@modules/zuora/*": ["../zuora/src/*"],
-      "@modules/zuora-catalog/*": ["../zuora-catalog/src/*"],
-      "@modules/product-catalog/*": ["../product-catalog/src/*"],
-      "@modules/internationalisation/*": ["../internationalisation/src/*"],
-      "@modules/*": ["../*"]
-    }
-  },
```
johnduffell, line 17: this was causing modules other than the ones listed to need /src/ in their import.  I deleted this so it falls back to the one at the root, which seems to work ok.

handlers/discount-api/src/index.ts
```diff
@@ -1,6 +1,7 @@
+import 'source-map-support/register';
```
johnduffell, line 1: this line is needed to enable source maps when running the AWS lambda.

modules/routing/src/logger.ts
```diff
@@ -0,0 +1,188 @@
...
+		console.log('logger: resetting context', this.prefix);
+		this.prefix = [];
+	}
+
+	public mutableAddContext(value: string): void {
+		console.log('logger: adding context', value);
+		this.prefix.push(value);
+	}
+
+	// look at the stack to work out the caller's details
```
johnduffell, line 16: this function uses some horrible regex on stack traces to look through the node stack traces and find the calling file name and function.  Apparently it's not the most performant thing to do, but my feeling is that the amount of time wasted looking through the code to find things is a lot higher than the amount of time to execute it!

modules/routing/src/logger.ts
```diff
@@ -0,0 +1,189 @@
...
+
+			return filename + (functionName ? '::' + functionName : '');
+		}
+		return undefined;
+	}
+
+	/* eslint-disable @typescript-eslint/no-explicit-any -- this has to match console.log */
+	/* eslint-disable @typescript-eslint/no-unsafe-argument -- this has to match console.log */
+	public log(message: any, ...optionalParams: any[]): void {
+		this.logWithCallerInfo(message, this.getCallerInfo(), optionalParams);
```
j-ruda-guardian, line 48: Do you want to spread the optional params here so the array is not passed as a single arg? Just double checking

modules/routing/src/logger.ts
```diff
@@ -0,0 +1,189 @@
...
+			const prettyArgs: Array<[string, unknown]> = args.map((arg, index) => {
+				const paramName = paramNames[index] ?? `arg${index}`;
+				const value =
+					arg instanceof ZodObject ? '(ZodObject not expanded)' : arg;
+				return [paramName, value];
+			});
+			const shortPrettyArgs =
+				shortArgsNum === 0
+					? undefined
+					: Object.fromEntries(prettyArgs.slice(0, shortArgsNum ?? 1));
```
j-ruda-guardian, line 117: So if there are no args, shortPrettyArgs is undefined, but later `shortPrettyArgs !== undefined ? '' : `SHORT_ARGS`` meaning it will log SHORT_ARGS when shortPrettyArgs is undefined. I think the logic is inverted here?

johnduffell: great spot yes you are exactly right, will fix
As a side note, I wasn't sure how far to go with unit testing these logging things, I did add a couple (actually updated the existing ones I previously added) but now I'm thinking I will add a few more.

modules/routing/src/logger.ts
johnduffell: yes copilot was very obliging with the regex but it does like to reinvent the wheel, and I agree it's fragile.

Looking at that link you sent, it's not clear if it supports node and it says for node to use https://www.npmjs.com/package/stack-trace
which seems not to be maintained.
There do seem to be other options though, e.g. https://www.npmjs.com/package/stacktrace-parser
Some options don't seem to support source maps so we are a little limited.

I'll try out the one you suggest and see what happens.

johnduffell: Having looked into a few libraries, generally the maintenance status seems doubtful, and given it's a controlled environment, I think sticking with regex will be ok for now.
If we bump the AWS node version in combination with the one we use for CI, the tests should pick up any incompatible changes.  If anything is missed, it's not critical to the functionality, just the debugging.
If someone decides a library is better in future, we can make another PR.

johnduffell:
> I left a couple of suggestions that we can look at in support. Just one extra comment here, is there a level of redaction needed or missing?

It's a good question on redaction, my opinion on that has always been that logs are protected by AWS access, and anyone who has access to the logs also has access to all the credentials, as well as their own access to the systems.
If we want to exclude certain things from casual viewing in the logs, we could certainly add a redaction list to the logger (similar to the context) and it could redact them automatically.  Happy for this to be added in a later PR.

johnduffell:
final test in CODE ok
product-switch-api
https://eu-west-1.console.aws.amazon.com/cloudwatch/home?region=eu-west-1#logsV2:log-groups/log-group/$252Faws$252Flambda$252Fproduct-switch-api-CODE/log-events/2025$252F09$252F16$252F$255B$2524LATEST$255D4997e048d9914e68adfa2d4ef863b672

discount-api
https://eu-west-1.console.aws.amazon.com/cloudwatch/home?region=eu-west-1#logsV2:log-groups/log-group/$252Faws$252Flambda$252Fdiscount-api-CODE/log-events/2025$252F09$252F16$252F$255B$2524LATEST$255D55b49da957d74af39ba60d06091b574c

mparticle-api
https://eu-west-1.console.aws.amazon.com/cloudwatch/home?region=eu-west-1#logsV2:log-groups/log-group/$252Faws$252Flambda$252Fmparticle-api-baton-CODE/log-events/2025$252F09$252F16$252F$255B$2524LATEST$255D8d5e7a69c36c49018065456e220605c7

PR #3063, opened 2025-09-15
Move zuora-datalake-export alarms to Platform

graham228221:
## What does this change?

Adds `zuora-datalake-export` to the Platform Team alarm mapping so that we can receive alerts about errors in this Lambda.


andrade0 (approved):
LGTM

PR #2875, opened 2025-06-10
add LastPlanAddedDate__c to subscriptions on switch

johnduffell:
Same purpose as https://github.com/guardian/support-frontend/pull/7040

This PR makes it so when switching/moving products, the LastPlanAddedDate__c is updated accordingly.

Tested in CODE
- supporter membership -> contribution
- contribution -> supporter plus
Both updated the existing value of the field to the current date.

PR #3067, opened 2025-09-16
don't preview term renewals in product-switch-api

johnduffell:
Recently I made some changes to product-switch-api to add a  new field on product switching.
At the same time, I also reused the same order object in both preview and actual.

The only concern I had was whether previewing the term renewal was ok:
https://github.com/guardian/support-service-lambdas/pull/2875/files#r2138006152

After testing, it seemed to be the case, however in PROD it's been failing for those with pending contribution updates on the zuora preview call:
```
[
    {
      code: '58730030',
      message: 'The shrink operation can only applied after the charge last change.'
    }
  ]
```

This PR changes it back to not do the term renewal on preview, while retaining the reuse.

Tested to preview in CODE both with and without a pending amendment.


handlers/product-switch-api/src/contributionToSupporterPlus.ts
```diff
@@ -313,7 +322,11 @@ export const buildSwitchRequestBody = (
 			runBilling: true,
 			collectPayment: false, // We will take payment separately because we don't want to charge the user if the amount payable is less than 50 pence/cents
 		},
-		...buildSwitchRequestWithoutOptions(productSwitchInformation, orderDate),
+		...buildSwitchRequestWithoutOptions(
+			productSwitchInformation,
+			orderDate,
+			false,
```
johnduffell, line 328: false means that it's not in preview mode, so the build function will include the renewal parts of the order

handlers/product-switch-api/test/fixtures/pendingAmendments.json
```diff
@@ -0,0 +1,252 @@
+{
```
johnduffell, line 1: this is an example contribution with a pending amount update (as of 16th sep 2025)

handlers/product-switch-api/src/contributionToSupporterPlus.ts
```diff
@@ -9,6 +9,7 @@ import type {
 	OrderRequest,
 	PreviewOrderRequest,
 } from '@modules/zuora/orders/orderRequests';
+import { previewOrderRequest } from '@modules/zuora/orders/orderRequests';
```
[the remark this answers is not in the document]
johnduffell: actually as far as I know the opposite is true, that's why that method exists.

MaelGNM (approved):
Looks good to me!

PR #3074, opened 2025-09-17
Set LastPlanAddedDate__c custom field

rupertbates:
## What does this change?
We need to set the LastPlanAddedDate__c custom field when creating new subscriptions to support various MMA functions, see https://github.com/guardian/support-frontend/pull/7040 for more information.

### DIFF
--- a/modules/zuora/src/createSubscription/createSubscription.ts
+++ b/modules/zuora/src/createSubscription/createSubscription.ts
@@ -116,10 +116,11 @@ function buildCreateSubscriptionRequest<T extends PaymentMethod>(
 	const customFields = {
 		DeliveryAgent__c: deliveryAgent.toString(),
 		ReaderType__c: readerType,
+		LastPlanAddedDate__c: zuoraDateFormat(contractEffectiveDate),
 	};
 	return {
 		newAccount: newAccount,
-		orderDate: zuoraDateFormat(dayjs()),
+		orderDate: zuoraDateFormat(contractEffectiveDate),
 		description: `Created by createSubscription.ts in support-service-lambdas`,
 		subscriptions: [
 			{
--- a/modules/zuora/src/orders/orderRequests.ts
+++ b/modules/zuora/src/orders/orderRequests.ts
@@ -30,7 +30,7 @@ export type OrderRequest = AccountOrderRequest & {
 		subscriptionNumber?: string;
 		orderActions: OrderAction[];
 		customFields?: {
-			LastPlanAddedDate__c?: string;
+			LastPlanAddedDate__c: string;
 			DeliveryAgent__c?: string;
 			ReaderType__c?: string;
 		};


modules/zuora/src/orders/orderRequests.ts
```diff
@@ -30,7 +30,7 @@ export type OrderRequest = AccountOrderRequest & {
 		subscriptionNumber?: string;
 		orderActions: OrderAction[];
 		customFields?: {
-			LastPlanAddedDate__c?: string;
+			LastPlanAddedDate__c: string;
```
rupertbates, line 33: No reason for this to be optional I don't think

tjmw (approved):

modules/zuora/src/createSubscription/createSubscription.ts
```diff
@@ -116,10 +116,11 @@ function buildCreateSubscriptionRequest<T extends PaymentMethod>(
 	const customFields = {
 		DeliveryAgent__c: deliveryAgent.toString(),
 		ReaderType__c: readerType,
+		LastPlanAddedDate__c: zuoraDateFormat(contractEffectiveDate),
 	};
 	return {
 		newAccount: newAccount,
-		orderDate: zuoraDateFormat(dayjs()),
+		orderDate: zuoraDateFormat(contractEffectiveDate),
```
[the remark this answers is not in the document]
johnduffell: dates in zuora really confuse me.
For subscribe and amend, there was no separate date on the initial subscription, but for orders, you have an order date separate from the contract effective date.  But we always make it today, and the customeracceptance date is the first issue or end of trial.
So you have changed this so that the order date always matches the date we made the agreement, even if the system is retrying overnight ?

modules/zuora/src/orders/orderRequests.ts
```diff
@@ -30,7 +30,7 @@ export type OrderRequest = AccountOrderRequest & {
 		subscriptionNumber?: string;
 		orderActions: OrderAction[];
 		customFields?: {
```
johnduffell, line 32: as per below comment
```suggestion
		customFields: {
```

modules/zuora/src/orders/orderRequests.ts
johnduffell: I'm not sure we should be requiring it for all orders, just the "subscribe" type ones - there must be some orders that we don't want to update the LastPlanAddedDate__c, what if we're just doing a holiday stop or something?
However at the moment it is the case that it's only product switching and support-workers that use this type, so it wil be ok for the time being.

johnduffell (approved):
looks good thanks for the quick PR although see comments below.

modules/zuora/src/orders/orderRequests.ts
rupertbates: Yeah the requests in that file only apply to subscribe type orders at the moment, we'll have to refactor if/when that stops being the case.

rupertbates: customFields is optional because it's not needed for preview requests

PR #3080, opened 2025-09-19
move CDK of discount-api and alarms-handler CDK into their handler folders

johnduffell:
Note: this takes account of the feedback from yesterday's meeting - DevX have validated the current approach - see "alternative options" link below

## The problem

- Each lambda is a separate pnpm package.
- There is a mono-CDK package for all handlers.

This structure artificially separates a handler's CDK and code, making it harder to find and develop both in conjunction.

In terms of the GHA for each handler:
- they wait for all common tests to finish before starting.
- each one then builds and tests all CDK (for all handlers), then
- builds the relevant handler.

The second step is slow and not scalable as we add more handlers.

It would be good if the build can be made more scalable, and for related code to be stored closer together.

## This PR

This PR switches two lambdas initially (alarms handler and discount api) over to a proposed "split CDK" structure:
- moving the associated CDK for both the alarms handler and discount api into their handler folder (see image below)

| What's in github before (lambda code directly in handler, combined CDK): | what's in github after (lambda code under lambda, and relevant CDK in cdk):  |
| -- | -- |
| image image | image |

In terms of convenience it's now easy to find and run the relevant CDK when developing the associated lambda.
Improvement in build times from the restructure is modest at the moment, but over time it will scale more as it's only building its own CDK. (note that most of the CI improvements could alternatively be gained by filtering which tests and builds to run directly in the central CDK package)
image
(Note that the "common" build has a similar problem and takes 2min 30) - that is, for every new bit of code in common, we build and test it all even if a lambda doesn't use it.  However it does save on GHA minutes by only doing it once.)

I have deployed the two lambdas to CODE successfully

## Disadvantages

More duplication as the CDK package.json and other files need to be duplicated inside multiple handlers (although see https://github.com/guardian/support-service-lambdas/pull/2907 for a potential solution)
Extra directory level between the handler directory and the lambda code.
Harder to do CDK-wide changes across all handlers without some clever scoping.

## Alternative options
- no change (default)
- no restructure, but have a "common" CDK build which somehow exports the artifacts to the individual builds (uses fewer GHA minutes overall, but would be at least as slow as the current way)
- keep the CDK all together as present but in the GHA build, filter the tests and build down to only run the current handler 
- keep a central CDK project but move the individual .ts files to the handlers.  In the GHA build, filter the tests and build down to only run the current handler. (this option was posed to DevX and they have recommended to stick with this PR method https://chat.google.com/room/AAAAag0I08g/f-UIOBvwmhs/f-UIOBvwmhs?cls=10 )

## Future work (in later PRs)
- migrate all TS lambdas (mechanical task - should be completed within days/weeks to ensure consistency, I have a script in another branch: https://github.com/guardian/support-service-lambdas/blob/48ed4c93c3eccfa4e72cfe00d277f046e575afdd/buildcheck/scripts/migrate-handler.sh )
- keep the extra build files in step using something like https://github.com/guardian/support-service-lambdas/pull/2907

### DIFF
--- a/handlers/alarms-handler/cdk/tsconfig.json
+++ b/handlers/alarms-handler/cdk/tsconfig.json
@@ -1,6 +1,9 @@
 {
   "extends": "../../../cdk/tsconfigBase.json",
-  "include": ["lib/**/*", "bin/**/*"],
+  "include": [
+    "lib/**/*",
+    "bin/**/*"
+  ],
   "exclude": [
     "node_modules",
     "cdk.out",
--- a/handlers/discount-api/cdk/tsconfig.json
+++ b/handlers/discount-api/cdk/tsconfig.json
@@ -1,6 +1,9 @@
 {
   "extends": "../../../cdk/tsconfigBase.json",
-  "include": ["lib/**/*", "bin/**/*"],
+  "include": [
+    "lib/**/*",
+    "bin/**/*"
+  ],
   "exclude": [
     "node_modules",
     "cdk.out",


handlers/alarms-handler/cdk/.prettierignore
```diff
@@ -0,0 +1 @@
+dist/
```
johnduffell, line 1: all these files are pretty much based on what the setup for the main cdk project is at the root.
We can simplify them in due course e.g. import a central one in the tsconfig.json file rather than duplicating things

handlers/alarms-handler/cdk/package.json
```diff
@@ -0,0 +1,44 @@
...
+      "<rootDir>/lib/**/*.test.ts"
+    ],
+    "transform": {
+      "^.+\\.tsx?$": "ts-jest"
+    },
+    "setupFilesAfterEnv": [
+      "./jest.setup.js"
+    ],
+    "moduleNameMapper": {
+      "@cdk/module/(.*)$": "<rootDir>/../../../cdk/module/$1"
```
johnduffell, line 41: this file is mostly a copy of the cdk/package.json, however this extra line is needed to get access to the shared constants.

handlers/alarms-handler/cdk/tsconfig.json
```diff
@@ -0,0 +1,48 @@
...
+    "inlineSources": true,
+    "experimentalDecorators": true,
+    "strictPropertyInitialization": false,
+    "typeRoots": [
+      "./node_modules/@types"
+    ],
+    "outDir": "dist",
+    "paths": {
+      "@cdk/module/*": [
+        "../../../cdk/module/*"
```
johnduffell, line 35: as per the package.json, this extra line is needed to get access to the shared constants.

handlers/alarms-handler/lambda/tsconfig.json
```diff
@@ -0,0 +1,3 @@
+{
```
johnduffell, line 1: also simply moved (with an extra `../` added)

handlers/discount-api/cdk/.prettierignore
```diff
@@ -0,0 +1 @@
+dist/
```
johnduffell, line 1: this is the start of the discount-api changes - pretty similar to the alarms handler

cdk/bin/cdk.ts
```diff
@@ -30,14 +28,16 @@ import { UpdateSupporterPlusAmount } from '../lib/update-supporter-plus-amount';
...
+	supportHostedZoneId,
+} from '../module/constants';
 
 const app = new App();
-const membershipHostedZoneId = 'Z1E4V12LQGXFEC';
-const membershipCertificateId = 'c1efc564-9ff8-4a03-be48-d1990a3d79d2';
-const membershipApisDomain = 'membership.guardianapis.com';
-export const supportHostedZoneId = 'Z3KO35ELNWZMSX';
-export const supportCertificateId = 'b384a6a0-2f54-4874-b99b-96eeff96c009';
-export const supportApisDomain = 'support.guardianapis.com';
```
johnduffell, line 40: these are all moved to a shared module so that the individiaul handlers' CDK can access them

cdk/lib/batch-email-sender.ts
```diff
@@ -16,7 +16,7 @@ import { ComparisonOperator, Metric } from 'aws-cdk-lib/aws-cloudwatch';
 import { Effect, Policy, PolicyStatement } from 'aws-cdk-lib/aws-iam';
 import { LoggingFormat, Runtime } from 'aws-cdk-lib/aws-lambda';
 import { CfnRecordSet } from 'aws-cdk-lib/aws-route53';
-import { SrLambdaAlarm } from './cdk/sr-lambda-alarm';
+import { SrLambdaAlarm } from '../module/cdk/sr-lambda-alarm';
```
johnduffell, line 19: just need to adjust the imports as some of the shared stuff is moved to the shared module.

handlers/discount-api/lambda/test/addDiscountIntegration.test.ts
```diff
@@ -9,7 +9,7 @@ import { addDiscount } from '@modules/zuora/discount';
 import { getSubscription } from '@modules/zuora/subscription';
 import { ZuoraClient } from '@modules/zuora/zuoraClient';
 import dayjs from 'dayjs';
-import { createDigitalSubscription } from '../../../modules/zuora/test/it-helpers/createGuardianSubscription';
+import { createDigitalSubscription } from '../../../../modules/zuora/test/it-helpers/createGuardianSubscription';
```
johnduffell, line 12: not sure why these are importing using a relative path, but this isn't something to solve in this PR

handlers/discount-api/lambda/package.json
```diff
@@ -4,7 +4,7 @@
     "test": "jest --group=-integration",
     "it-test": "jest --group=integration",
     "type-check": "tsc --noEmit",
-    "build": "esbuild --bundle --platform=node --target=node18 --outfile=target/index.js src/index.ts --sourcemap",
+    "build": "esbuild --bundle --platform=node --target=node18 --outdir=target/ src/index.ts --sourcemap",
```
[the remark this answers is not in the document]
johnduffell: this was intentional to keep it consistent between the two lambdas, and it doesn't generate anytthing it shouldn't
image

handlers/discount-api/cdk/tsconfig.json
```diff
@@ -0,0 +1,9 @@
+{
+  "extends": "../../../cdk/tsconfigBase.json",
+  "include": ["lib/**/*", "bin/**/*"],
+  "exclude": [
+    "node_modules",
+    "cdk.out",
+    "lib/**/__snapshots__/**"
+  ]
```
johnduffell, line 8: trying to inherit these values didn't work because it seemed to resolve the path relative to the file it's in, rather than this file that imported it originally

PR #3085, opened 2025-09-19
Fix/add lambda stripe secrets permissions

andrade0:
## What does this change?

  This PR adds the necessary IAM permissions for the Lambda producer to access Stripe webhook secrets stored in AWS Secrets Manager. This is required for
  webhook signature verification to ensure requests are legitimately from Stripe.

  Changes:
  - Added IAM policy for Lambda producer to read Stripe secrets
  - Scoped permissions to specific secret ARN pattern
  - Maintains existing Salesforce secret access permissions

### DIFF
--- a/cdk/lib/stripe-disputes.ts
+++ b/cdk/lib/stripe-disputes.ts
@@ -175,6 +175,20 @@ export class StripeDisputes extends GuStack {
 		);
 
 		this.createPolicyAndAttachToLambdas(
+			[{ lambda: lambdaProducer, name: 'Lambda Producer' }],
+			[
+				new PolicyStatement({
+					effect: Effect.ALLOW,
+					actions: ['secretsmanager:GetSecretValue'],
+					resources: [
+						`arn:aws:secretsmanager:${this.region}:${this.account}:secret:${this.stage}/Stripe/ConnectedApp/StripeDisputeWebhooks-*`,
+					],
+				}),
+			],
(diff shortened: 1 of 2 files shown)


cdk/lib/stripe-disputes.ts
```diff
@@ -174,6 +174,20 @@ export class StripeDisputes extends GuStack {
...
 		);
 
+		this.createPolicyAndAttachToLambdas(
+			[{ lambda: lambdaProducer, name: 'Lambda Producer' }],
+			[
+				new PolicyStatement({
+					effect: Effect.ALLOW,
+					actions: ['secretsmanager:GetSecretValue'],
+					resources: [
+						`arn:aws:secretsmanager:${this.region}:${this.account}:secret:${this.stage}/Stripe/ConnectedApp/StripeDisputeWebhooks-*`,
```
graham228221, line 184: It's not a "ConnectedApp" in use here, that's a Salesforce specific thing. But this is fine, it's not confusing as it's the same sort of thing as a ConnectedApp (a restricted access API Key).

graham228221: We don't have a format for this, but I'd probably suggest something like this, but what you've got is fine.

```suggestion
						`arn:aws:secretsmanager:${this.region}:${this.account}:secret:${this.stage}/Stripe/StripeDisputeWebhooks-*`,
```

PR #3095, opened 2025-09-24
when building CDK, don't run tests for unrelated handlers

johnduffell:
At present when running the build, we run all CDK tests.

This takes over a minute locally and over half a minute in the build
https://github.com/guardian/support-service-lambdas/actions/runs/17973058408/job/51120281086#step:7:2153

This PR changes it so 
1. it will only run the test for the relevant handler in the build (assuming it's named correctly)
1. each lambda gets a `pnpm cdk:test` and `pnpm cdk:test-update` command for convenience

Advantages
- the `test` step takes e.g. 11 instead of 34 seconds
- this should not increase as new handlers are added in future.
- when running locally, the test(s) take ~5s instead of >60s

Disadvantges
- tests will be ignored if incorrectly named
- unnecessary complexity in package.json (suggestions please)

Comparison with https://github.com/guardian/support-service-lambdas/pull/3080 :
- Partial scalability and speed improvements: CDK CI saving ~24s rather than ~39s at this stage
- No affect on daily workflow (no change to structure). 
- If we merge this, we can still implement the alternative PR later if we decide the extra benefits are worth it.


handlers/discount-api/package.json
```diff
@@ -1,6 +1,8 @@
 {
   "name": "discount-api",
   "scripts": {
+    "cdk:test": "pnpm --filter cdk test discount-api",
+    "cdk:test-update": "pnpm --filter cdk test-update discount-api",
```
johnduffell, line 5: two handlers as an example, but we can add it to them all if we agree on the approach

cdk/package.json
```diff
@@ -5,13 +5,13 @@
   "scripts": {
     "build": "tsc",
     "type-check": "tsc --noEmit",
-    "test": "jest",
-    "test-update": "jest -u",
+    "test": "sh -c 'jest ${1:+--testPathPattern=\"$1\"}' --",
+    "test-update": "sh -c 'jest -u ${1:+--testPathPattern=\"$1\"}' --",
```
johnduffell, line 9: these will run all tests containing "alarms-handler" etc in the name

shtukas (approved):

PR #3101, opened 2025-09-30
switch discount api and product switch api to cdk configured source maps

johnduffell:
TODO


cdk/bin/cdk.ts
```diff
@@ -32,12 +40,6 @@ import { WriteOffUnpaidInvoices } from '../lib/write-off-unpaid-invoices';
 import { ZuoraSalesforceLinkRemover } from '../lib/zuora-salesforce-link-remover';
 
 const app = new App();
-const membershipHostedZoneId = 'Z1E4V12LQGXFEC';
-const membershipCertificateId = 'c1efc564-9ff8-4a03-be48-d1990a3d79d2';
-const membershipApisDomain = 'membership.guardianapis.com';
-export const supportHostedZoneId = 'Z3KO35ELNWZMSX';
-export const supportCertificateId = 'b384a6a0-2f54-4874-b99b-96eeff96c009';
-export const supportApisDomain = 'support.guardianapis.com';
```
johnduffell, line 40: moved to constants.ts, so that they can be referenced from other files e.g. tests without importing this file therefore constructing every single stack

cdk/lib/__snapshots__/discount-api.test.ts.snap
```diff
@@ -178,7 +106,7 @@ exports[`The Discount API stack matches the snapshot 1`] = `
           },
           {
             "Key": "Stack",
-            "Value": "membership",
+            "Value": "support",
```
johnduffell, line 109: We no longer pass these in in the tests now so they are correctly inferred

cdk/lib/__snapshots__/discount-api.test.ts.snap
```diff
@@ -188,17 +116,35 @@ exports[`The Discount API stack matches the snapshot 1`] = `
...
+        },
+        "PolicyName": "S3catalogreadpolicyE18911B4",
+        "Roles": [
+          {
+            "Ref": "discountapilambdaServiceRole4CB5B1F4",
+          },
+        ],
+      },
+      "Type": "AWS::IAM::Policy",
+    },
```
johnduffell, line 139: this has been split from the below policy to allow reusibility

cdk/lib/__snapshots__/discount-api.test.ts.snap
```diff
@@ -294,16 +243,17 @@ exports[`The Discount API stack matches the snapshot 1`] = `
...
-          "S3Key": "membership/CODE/discount-api/discount-api.zip",
+          "S3Key": "support/CODE/discount-api/discount-api.zip",
         },
         "Description": "A lambda that enables the addition of discounts to existing subscriptions",
         "Environment": {
           "Variables": {
             "APP": "discount-api",
             "App": "discount-api",
-            "STACK": "membership",
+            "NODE_OPTIONS": "--enable-source-maps",
```
johnduffell, line 253: this is a new line intended to make source maps work automatically in logs/stack traces

cdk/lib/__snapshots__/discount-api.test.ts.snap
```diff
@@ -188,17 +116,35 @@ exports[`The Discount API stack matches the snapshot 1`] = `
...
     "S3inlinepolicy3B07399A": {
       "Properties": {
         "PolicyDocument": {
           "Statement": [
             {
               "Action": "s3:GetObject",
               "Effect": "Allow",
-              "Resource": [
-                "arn:aws:s3::*:membership-dist/membership/CODE/discount-api/",
-                "arn:aws:s3::*:gu-zuora-catalog/PROD/Zuora-CODE/*",
```
johnduffell, line 200: as above, the zuora-code catalog is now in the above policy

cdk/lib/__snapshots__/discount-api.test.ts.snap
```diff
@@ -1190,17 +1139,35 @@ exports[`The Discount API stack matches the snapshot 2`] = `
       },
       "Type": "AWS::ApiGateway::DomainName",
     },
+    "S3catalogreadpolicyE18911B4": {
```
johnduffell, line 1142: the rest of this snap file is just the same differences on the PROD stack as the CODE one

cdk/lib/__snapshots__/update-supporter-plus-amount.test.ts.snap
```diff
@@ -1015,8 +1014,8 @@ exports[`The Update supporter plus amount stack matches the snapshot 2`] = `
             ],
           },
         ],
-        "AlarmDescription": "Impact - Update supporter plus amount api returned a 5XX response. This means that a user who was trying to update the contribution amount of their supporter plus subscription has received an error. Please check the logs to diagnose the issue. Follow the process in https://docs.google.com/document/d/1_3El3cly9d7u_jPgTcRjLxmdG2e919zCLvmcFCLOYAk/edit",
-        "AlarmName": "update-supporter-plus-amount-PROD Update supporter plus amount - API gateway 5XX response",
+        "AlarmDescription": "Impact - Update supporter plus amount api returned a 5XX response. This means that a user who was trying to update the contribution amount of their supporter plus subscription has received an error. Search the log link below for "error". Follow the process in https://docs.google.com/document/d/1_3El3cly9d7u_jPgTcRjLxmdG2e919zCLvmcFCLOYAk/edit",
```
johnduffell, line 1017: the impact statement of this alarm was actually quite helpful so it's been retained

cdk/lib/cdk/policies.ts
```diff
@@ -0,0 +1,46 @@
+import { Effect, Policy, PolicyStatement } from 'aws-cdk-lib/aws-iam';
+import type { SrStack } from './sr-stack';
+
```
johnduffell, line 3: this file defines some useful standard policies needed for common things.  I wasn't sure whether using a function like this is best, or a class is standard?
Also it would be good to get auto complete, in scala you could wrap them all in an `object SrPolicies` and then type "SrPolicies." in order to see the options, not sure if I can/should do similar here?

cdk/lib/cdk/sr-lambda.ts
```diff
@@ -1,44 +1,70 @@
...
-}
+	runtime: Runtime;
+	loggingFormat: LoggingFormat;
+	memorySize: number;
+	timeout: Duration;
+	environment: Record<string, string>;
+};
+
+type SrLambdaProps = Omit<GuFunctionProps, keyof SrDefaultProps> &
+	Partial<SrDefaultProps>;
```
johnduffell, line 21: this makes all of the defaulted properties optional (so you can still override them)

cdk/lib/cdk/sr-rest-api.ts
```diff
@@ -0,0 +1,143 @@
...
+	CfnDomainName,
+} from 'aws-cdk-lib/aws-apigateway';
+import { ComparisonOperator, Metric } from 'aws-cdk-lib/aws-cloudwatch';
+import { Effect, Policy, PolicyStatement } from 'aws-cdk-lib/aws-iam';
+import { CfnRecordSet } from 'aws-cdk-lib/aws-route53';
+import { certForStack } from '../constants';
+import { SrApiLambda } from './sr-lambda';
+import { SrLambdaAlarm } from './sr-lambda-alarm';
+import type { SrStack } from './sr-stack';
+
```
johnduffell, line 14: the SrRestApi construct lets us have an API gateway with our standard domain setup etc in only a few lines of code.  There may be an existing similar thing, in which case we can (later) inherit some of the behaviour from that.

cdk/lib/cdk/sr-rest-api.ts
```diff
@@ -0,0 +1,143 @@
...
+					period: Duration.seconds(300),
+					dimensionsMap: {
+						ApiName: nameWithStage,
+					},
+				}),
+				lambdaFunctionNames: lambda.functionName,
+			});
+		}
+
+		const cert = certForStack[scope.stack];
```
johnduffell, line 116: now we look up the cert/domain info based on the stack, which means it's consistent in the tests

cdk/lib/cdk/sr-rest-api.ts
```diff
@@ -0,0 +1,143 @@
...
+					effect: Effect.ALLOW,
+					actions: ['s3:GetObject'],
+					resources: [
+						`arn:aws:s3::*:membership-dist/${scope.stack}/${scope.stage}/${app}/`,
+					],
+				}),
+			],
+		});
+
+		lambda.role?.attachInlinePolicy(s3InlinePolicy);
```
johnduffell, line 85: here we only attach the policy that lets it get the artifact zip

It gets policies to read the config here:
https://github.com/guardian/cdk/blob/5569c749211b518001666cffb558fe403ff0539c/src/constructs/lambda/lambda.ts#L184-L189

And any other lambda specific policies needed are added in the individual handler's CDK.

cdk/lib/constants.ts
```diff
@@ -0,0 +1,18 @@
+export const membershipHostedZoneId = 'Z1E4V12LQGXFEC';
```
johnduffell, line 1: these are all moved to a separate file, as otherwise every time they were referenced in a test it was loading the main file and constructing all the stacks in memory.

cdk/lib/discount-api.ts
```diff
@@ -1,186 +1,28 @@
...
-export class DiscountApi extends GuStack {
-	constructor(scope: App, id: string, props: DiscountApiProps) {
+	getSqsPolicy,
+	s3CatalogReadPolicy,
+	zuoraOAuthSecretsPolicy,
+} from './cdk/policies';
+import { SrRestApi } from './cdk/sr-rest-api';
+import type { SrStackProps } from './cdk/sr-stack';
+import { SrStack } from './cdk/sr-stack';
+
```
johnduffell, line 10: these files are where the real improvement is, you can see that now we only need a few lines to get the whole Rest API setup.

handlers/alarms-handler/package.json
```diff
@@ -4,7 +4,7 @@
 		"test": "jest --group=-integration",
 		"it-test": "jest --group=integration",
 		"type-check": "tsc --noEmit",
-		"build": "esbuild --bundle --platform=node --target=node18 --outdir=target/ src/index.ts src/indexScheduled.ts",
+		"build": "esbuild --bundle --platform=node --target=node20 --outdir=target/ src/index.ts src/indexScheduled.ts --sourcemap",
```
johnduffell, line 7: for some reason a lot of these were on node18, but the actual runtime is on 20 across the board
 https://github.com/guardian/support-service-lambdas/blob/73e90f172e9d71566db9621f26d7e94c52b0510e/cdk/lib/node-version.ts#L4

I have added the --sourcemap parameter to make sure it's always available, although only 3 lambdas use it so far

handlers/discount-api/src/index.ts
```diff
@@ -1,4 +1,3 @@
-import 'source-map-support/register';
```
johnduffell, line 1: no longer need to register it in the handler, it will happen from the environment variable defined in cdk

cdk/lib/__snapshots__/discount-api.test.ts.snap
```diff
@@ -692,14 +642,14 @@ exports[`The Discount API stack matches the snapshot 1`] = `
...
-    "discountapilambdadiscountapiCODEDeployment15948B741b868ab49ec282f04008c03e43e51ea1": {
+    "discountapilambdadiscountapiCODEDeployment15948B74c4abca5598ce0c19a9190665995d06f7": {
       "DependsOn": [
         "discountapilambdadiscountapiCODEproxyANY3023C5CD",
         "discountapilambdadiscountapiCODEproxyDA26EBF1",
         "discountapilambdadiscountapiCODEANY827A81C0",
       ],
       "Properties": {
-        "Description": "API Gateway created by CDK",
+        "Description": "API Gateway endpoint for the discount-api-CODE lambda",
```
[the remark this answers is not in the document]
johnduffell: I parameterised it in the end so we could keep the old deployment, just to be on the safe side

cdk/lib/__snapshots__/discount-api.test.ts.snap
```diff
@@ -168,6 +97,10 @@ exports[`The Discount API stack matches the snapshot 1`] = `
           ],
         },
         "Tags": [
+          {
+            "Key": "App",
+            "Value": "discount-api",
+          },
```
johnduffell, line 103: previously we weren't applying the App tag to anything in the stacks due to the app value not being passed up to the GuStack

cdk/lib/__snapshots__/product-switch-api.test.ts.snap
```diff
@@ -5,7 +5,7 @@ exports[`The Product switch api stack matches the snapshot 1`] = `
   "Metadata": {
```
johnduffell, line 5: this file has similar changes to discount-api snapshot, I'll comment on any other differences

cdk/lib/__snapshots__/discount-api.test.ts.snap
johnduffell: AWS say "We recommend that you break up policies by resource type."[1] however one of these is provided by GuCDK and the other is an unrelated one for this lambda, so I'm not too upset about separating them.
Theoretically we could get SrRestApi to group them, but I think it's fine as it is especially as they get readable names.

[1]
https://docs.aws.amazon.com/IAM/latest/UserGuide/access_policies.html#:~:text=We%20recommend%20that%20you%20break%20up%20policies%20by%20resource%20type.

cdk/lib/__snapshots__/discount-api.test.ts.snap
```diff
@@ -188,22 +121,30 @@ exports[`The Discount API stack matches the snapshot 1`] = `
...
-              ],
+              "Resource": {
+                "Fn::Join": [
+                  "",
+                  [
+                    "arn:aws:s3:::",
+                    {
+                      "Ref": "DistributionBucketName",
+                    },
+                    "/support/CODE/discount-api/*",
```
johnduffell, line 139: if I preview it comes out looking like this:
`arn:aws:s3:::CORRECT_bucket_name_here/support/PROD/discount-api/*`

johnduffell:
> Would it be possible to break this down into smaller PRs? There are so many things happening it's really hard to get a full understanding of them all.

can do, if it's going to be more likely to get reviewed, watch this space

andresilva-guardian (approved):
Thank you @johnduffell for your big effort here!

I really like the reusable constructs (SrStack, SrRestApi, SrApiLambda). Also, this environment variable approach is cleaner than per-file imports! Strong typing throughout with proper enums and interfaces!

Only suggestion: I would consider adding comments to the new construct classes explaining their purpose.

johnduffell:
pretty sure this has been done in another PR across all lambdas
https://github.com/search?q=repo%3Aguardian%2Fsupport-service-lambdas%20%22--enable-source-maps%22&type=code

PR #3106, opened 2025-10-01
create reusable sr domain construct

johnduffell:
building on https://github.com/guardian/support-service-lambdas/pull/3105

We reuse the code to make the *.support.guardianapis.com connection to API gateway.

Further more we actually use the wrong values in the snapshot tests, making them less useful.

This PR changes it so the code is in a single construct and bases the values off the app stage and stack, meaning they are guaranteed correct consistent and accurate with minimal boilerplate/repetition.

Tested in CODE:
- https://eu-west-1.console.aws.amazon.com/cloudwatch/home?region=eu-west-1#logsV2:log-groups/log-group/$252Faws$252Flambda$252Fupdate-supporter-plus-amount-CODE/log-events/2025$252F10$252F01$252F$255B$2524LATEST$255D63026190fad14739800b3a7578af91d5
- https://eu-west-1.console.aws.amazon.com/cloudwatch/home?region=eu-west-1#logsV2:log-groups/log-group/$252Faws$252Flambda$252Fdiscount-api-CODE/log-events/2025$252F10$252F01$252F$255B$2524LATEST$255D10cbb0ef47ba47e286f0ff3ec2b5641c
- https://eu-west-1.console.aws.amazon.com/cloudwatch/home?region=eu-west-1#logsV2:log-groups/log-group/$252Faws$252Flambda$252Fproduct-switch-api-CODE/log-events/2025$252F10$252F01$252F$255B$2524LATEST$255D0e10597dc593402da3133e22a5c4a9f6


cdk/lib/__snapshots__/discount-api.test.ts.snap
```diff
@@ -106,7 +110,7 @@ exports[`The Discount API stack matches the snapshot 1`] = `
           },
           {
             "Key": "Stack",
-            "Value": "membership",
+            "Value": "support",
```
johnduffell, line 113: this was in as membership in the tests, which was wrong

cdk/lib/__snapshots__/discount-api.test.ts.snap
```diff
@@ -82,6 +82,10 @@ exports[`The Discount API stack matches the snapshot 1`] = `
         "Period": 300,
         "Statistic": "Sum",
         "Tags": [
+          {
+            "Key": "App",
+            "Value": "discount-api",
+          },
```
johnduffell, line 88: GuCDK adds the app tag to all the resources if defined in the GuStack, we weren't doing that before.

cdk/lib/cdk/sr-stack.ts
```diff
@@ -0,0 +1,24 @@
+import type {
+	AppIdentity,
+	GuStackProps,
+} from '@guardian/cdk/lib/constructs/core';
+import { GuStack } from '@guardian/cdk/lib/constructs/core';
+import type { App } from 'aws-cdk-lib';
+
+export type SrStackNames = 'support' | 'membership';
+export type SrStageNames = 'CODE' | 'PROD';
```
johnduffell, line 9: we need to narrow down the type so that we can have a lookup on the certificates etc.  Also handy to stop typos etc.

cdk/lib/constants.ts
```diff
@@ -0,0 +1,18 @@
...
+		certificateId: supportCertificateId,
+		domainName: supportApisDomain,
+		hostedZoneId: supportHostedZoneId,
+	},
+	membership: {
+		certificateId: membershipCertificateId,
+		domainName: membershipApisDomain,
+		hostedZoneId: membershipHostedZoneId,
+	},
+} as const;
```
johnduffell, line 18: this is the cert lookup, so we don't have to reference the right values in both stacks and the tests too

cdk/lib/cdk/sr-stack.ts
rupertbates: You could import this from https://github.com/guardian/support-service-lambdas/blob/fc72cd039745262f0d80b8d4ac85ab9747c7a4e8/modules/stage.ts#L1

PR #3107, opened 2025-10-01
move more lambdas over to Sr lambda construct

johnduffell:
Following on from recent work to add a reusable SR domain construct https://github.com/guardian/support-service-lambdas/pull/3106
and a reusable SR lambda construct

It appears that using them can reduce duplication and make our stacks more readable and consistent.

It also means that changes e.g. adding en environment variable to tell node to use source maps, would be a one liner.

This PR moves a bunch more stacks over to use the above constructs.


cdk/bin/cdk.ts
```diff
@@ -138,21 +141,6 @@ stacks.forEach((Constructor) => {
...
-	hostedZoneId: supportHostedZoneId,
-	certificateId: supportCertificateId,
-});
-new StripeDisputes(app, 'stripe-disputes-PROD', {
-	stack: 'support',
-	stage: 'PROD',
-	domainName: `stripe-disputes.${supportApisDomain}`,
-	hostedZoneId: supportHostedZoneId,
-	certificateId: supportCertificateId,
-});
```
johnduffell, line 154: it's better not to repeart the parameters every time here and in the tests, as they're more likely to get mixed up, now the stacks are constructed with a standard incantation above

cdk/bin/cdk.ts
```diff
@@ -310,16 +282,10 @@ new DiscountExpiryNotifier(app, 'discount-expiry-notifier-PROD', {
...
-	cloudFormationStackName: 'membership-CODE-metric-push-api',
-});
-new MetricPushApi(app, 'metric-push-api-PROD', {
-	stack: 'membership',
-	stage: 'PROD',
-	cloudFormationStackName: 'membership-PROD-metric-push-api',
-});
+
+new MetricPushApi(app, 'CODE', 'membership-CODE-metric-push-api');
+new MetricPushApi(app, 'PROD', 'membership-PROD-metric-push-api');
```
johnduffell, line 287: not sure why this needs a specific stack name, but I kept it just in case

cdk/bin/cdk.ts
```diff
@@ -352,11 +318,3 @@ new SalesforceEventBus(app, 'salesforce-event-bus-PROD', {
...
 	stage: 'PROD',
 });
-new MParticleApi(app, 'mparticle-api-CODE', {
-	stack: 'support',
-	stage: 'CODE',
-});
-new MParticleApi(app, 'mparticle-api-PROD', {
-	stack: 'support',
-	stage: 'PROD',
-});
```
johnduffell, line 362: don't forget to load the "large diff"s below and review them all

cdk/lib/__snapshots__/alarms-handler.test.ts.snap
```diff
@@ -102,7 +105,7 @@ exports[`The alarms-handler stack matches the snapshot 1`] = `
           },
           {
             "Key": "Stack",
-            "Value": "membership",
+            "Value": "support",
```
johnduffell, line 108: as with most of the tests, for some reason it was passing in membership.  Now it's not passed in at all so it is definitely consistent.

cdk/lib/__snapshots__/metric-push-api.test.ts.snap
```diff
@@ -202,29 +269,14 @@ exports[`The MetricPushApi stack matches the snapshot 1`] = `
...
-        },
-        "Stage": {
-          "Ref": "metricpushapilambdametricpushapiCODEDeploymentStageCODED2BD8FA5",
-        },
-      },
-      "Type": "AWS::ApiGateway::BasePathMapping",
-    },
     "MetricPushDNSRecord": {
       "Properties": {
-        "Comment": "CNAME for metric-push-api API CODE",
```
johnduffell, line 221: I don't think this comment goes anywhere, I was looking in AWS console and I couldn't see it

cdk/lib/__snapshots__/metric-push-api.test.ts.snap
```diff
@@ -218,9 +226,8 @@ exports[`The MetricPushApi stack matches the snapshot 1`] = `
     },
     "MetricPushDNSRecord": {
       "Properties": {
-        "Comment": "CNAME for metric-push-api API CODE",
-        "HostedZoneName": "support.guardianapis.com.",
-        "Name": "metric-push-api-code.support.guardianapis.com",
+        "HostedZoneId": "Z1E4V12LQGXFEC",
```
johnduffell, line 229: some places use HostedZoneName and some HostedZoneId, better to be consistent

cdk/lib/__snapshots__/metric-push-api.test.ts.snap
```diff
@@ -246,19 +253,19 @@ exports[`The MetricPushApi stack matches the snapshot 1`] = `
...
-              {
-                "Ref": "AWS::Region",
-              },
-              ":",
+              "arn:aws:acm:eu-west-1:",
               {
                 "Ref": "AWS::AccountId",
               },
-              ":certificate/b384a6a0-2f54-4874-b99b-96eeff96c009",
+              ":certificate/c1efc564-9ff8-4a03-be48-d1990a3d79d2",
```
johnduffell, line 260: this was actually wrong in the snapshot
supportCertificateId instead of membershipCertificateId

PR #3108, opened 2025-10-02
Add SQS queue and SNS subscription for mParticle API CDK for deletion requests

ecguardian:
This enables the mParticle API to receive user deletion requests via SNS topic

## What does this change?

This is part of work to make sure that when a User requests to delete their account in MMA, that this is simplement in both mParticle and Braze. Since mParticle has been incorporated into The Guardian setup, this changes the flow of data from the datalake to Braze. Before users could be deleted from the datalake and with the Diff Publisher, the users would be deleted from Braze as well. However the way that data is synced with mParticle, to guaranty that a users is deleted,  into  user's accounts are deleted when a users requests to delete their account in MMA

- Add SQS queue for deletion requests with DLQ
- Add SNS topic subscription to route deletion requests to SQS
- Add new deletion Lambda with SQS event source mapping
- Add proper IAM permissions for SNS->SQS communication
- Add CloudWatch alarms for deletion processing errors and DLQ monitoring
- Configure queue policy to allow SNS to send messages

## Accessibility

### DIFF
--- a/cdk/lib/mparticle-api.ts
+++ b/cdk/lib/mparticle-api.ts
@@ -1,6 +1,7 @@
 import { GuApiGatewayWithLambdaByPath } from '@guardian/cdk';
 import type { GuStackProps } from '@guardian/cdk/lib/constructs/core';
 import { GuStack } from '@guardian/cdk/lib/constructs/core';
+import { GuLambdaFunction } from '@guardian/cdk/lib/constructs/lambda';
 import { type App, Duration } from 'aws-cdk-lib';
 import { ComparisonOperator, Metric } from 'aws-cdk-lib/aws-cloudwatch';
 import {
@@ -9,11 +10,18 @@ import {
 	Policy,
 	PolicyStatement,
 	Role,
+	ServicePrincipal,
 } from 'aws-cdk-lib/aws-iam';
+import { LoggingFormat } from 'aws-cdk-lib/aws-lambda';
+import { SqsEventSource } from 'aws-cdk-lib/aws-lambda-event-sources';
+import { Topic } from 'aws-cdk-lib/aws-sns';
+import { SqsSubscription } from 'aws-cdk-lib/aws-sns-subscriptions';
+import { Queue } from 'aws-cdk-lib/aws-sqs';
 import { StringParameter } from 'aws-cdk-lib/aws-ssm';
 import { SrLambda } from './cdk/sr-lambda';
 import { SrLambdaAlarm } from './cdk/sr-lambda-alarm';
 import { SrLambdaDomain } from './cdk/sr-lambda-domain';
+import { nodeVersion } from './node-version';
 
 export class MParticleApi extends GuStack {
 	constructor(scope: App, id: string, props: GuStackProps) {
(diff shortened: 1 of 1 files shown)


johnduffell:
> When a user makes a request to delete their account in MMA, an SQS Queue will subscribe to this SNS topic and receive user Delete events requests.

Nitpick but the sqs queue will subscribe to the topic all the time (cross account subscription) and when the request is made, a message will appear on the queue.

cdk/lib/mparticle-api.ts
```diff
@@ -1,6 +1,7 @@
 import { GuApiGatewayWithLambdaByPath } from '@guardian/cdk';
```
johnduffell, line 1: coincidentally I've also been changing these files! So we might have to coordinate/pair to deal with any conflicts - see https://github.com/guardian/support-service-lambdas/pull/3109

cdk/lib/mparticle-api.ts
```diff
@@ -33,9 +41,28 @@ export class MParticleApi extends GuStack {
...
+
 		const sarS3BaseKey = 'mparticle-results/';
 
+		const deletionDlq = new Queue(this, `${app}-deletion-dlq`, {
+			queueName: `${app}-deletion-dlq-${this.stage}`,
+			retentionPeriod: Duration.days(14),
+		});
+
+		const deletionRequestsQueue = new Queue(this, `${app}-deletion-queue`, {
+			queueName: `${app}-deletion-queue-${this.stage}`,
```
johnduffell, line 58: nitpick: although we don't have a standard, I think appending dlq on the end is fine, then we can reuse the name.  (note that you can only search queues by prefix)
https://eu-west-1.console.aws.amazon.com/sqs/v3/home?region=eu-west-1#/queues
```suggestion
    const queueName = `${app}-user-deletions-${this.stage}`;
		const deletionDlq = new Queue(this, `${queueName}-dlq`, {
			queueName: `${app}-${queueNameSuffix}-dlq-${this.stage}`,
			retentionPeriod: Duration.days(14),
		});

		const deletionRequestsQueue = new Queue(this, `${app}-deletion-queue`, {
			queueName,
```

cdk/lib/mparticle-api.ts
```diff
@@ -150,6 +213,42 @@ export class MParticleApi extends SrStack {
 				evaluationPeriods: 1,
 				lambdaFunctionNames: batonLambda.functionName,
 			});
+
+			new SrLambdaAlarm(this, 'MParticleMmaUserDeletionLambdaErrorAlarm', {
```
johnduffell: do we need this alarm? if the queue will retry 3 times, then we don't mind too much if it fails the first time

cdk/lib/mparticle-api.ts
```diff
@@ -36,8 +39,36 @@ export class MParticleApi extends SrStack {
...
+			`${app}-mma-user-deletion-dlq`,
+			{
+				queueName: `${app}-mma-user-deletion-requests-dlq-${this.stage}`,
+				retentionPeriod: Duration.days(14),
+			},
+		);
+
+		const mmaUserDeletionRequestsQueue = new Queue(
+			this,
+			`${app}-mma-user-deletion-requests-queue`,
```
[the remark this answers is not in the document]
ecguardian: yes, great point, i have switched to match the style of the rest of the file 👍

cdk/lib/mparticle-api.ts
```diff
@@ -36,8 +39,36 @@ export class MParticleApi extends SrStack {
...
 		).stringValue;
 
-		// this must be the same as used in the code
-		const sarS3BaseKey = 'mparticle-results/';
+		const identityMmaSnsDeletionRequestTopicArn =
+			StringParameter.fromStringParameterName(
+				this,
+				'IdentityMmaSnsDeletionRequestTopicArn',
+				`/${this.stage}/${this.stack}/${app}/IdentityMmaSnsDeletionRequestTopicArn`,
+			).stringValue;
```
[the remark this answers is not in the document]
ecguardian: i agree with this too, i have switched to use SrAppConfigKey instead 👍

PR #3121, opened 2025-10-06
enable source maps in CDK and add reusable SrApiLambda construct

johnduffell:
(supersedes/part of https://github.com/guardian/support-service-lambdas/pull/3101 )

I did some work to make two TS lambdas use source maps in https://github.com/guardian/support-service-lambdas/pull/3062
It is very useful to have correct line numbers on the log statements and stack traces
image
and there doesn't seem to be any obious performance hit.

I decided to roll it out to more lambdas, and it turns out we can enable it using an environment variable
NODE_OPTIONS set to [--enable-source-maps](https://nodejs.org/api/cli.html#--enable-source-maps)
passed to the lambda rather than having to add an import in the index.ts of each lambda.

It would be better to set it in CDK as it can apply across the board rather than relying on it being added in each place.

As part of the CDK deduplication I have added a SrApiLambda construct.  This is just to stop us having to duplicate the same ~15-20 lines in every cdk file and make it easier to maintain.

As part of this change I have made source maps switched on in node automatically in order to get better stack traces and logging tags.

### DIFF
--- a/cdk/lib/cdk/sr-lambda.ts
+++ b/cdk/lib/cdk/sr-lambda.ts
@@ -10,17 +10,21 @@ export type SrLambdaProps = {
 	nameSuffix?: string; // when you have multiple lambdas in an app
 };
 
-type GuLambdaOverrides = Omit<
-	GuFunctionProps,
-	keyof ReturnType<typeof defaultProps>
-> &
-	Partial<ReturnType<typeof defaultProps>>;
+type DefaultProps = ReturnType<typeof getLambdaDefaultProps>;
+type GuLambdaOverrides = Omit<GuFunctionProps, keyof DefaultProps> &
+	Partial<DefaultProps>;
 
-function getNameWithStage(identity: Identity, nameSuffix: string | undefined) {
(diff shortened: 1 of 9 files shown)


cdk/lib/cdk/sr-lambda.ts
```diff
@@ -30,6 +34,9 @@ export function defaultProps(scope: Identity, nameSuffix: string | undefined) {
 		loggingFormat: LoggingFormat.TEXT,
 		memorySize: 1024,
 		timeout: Duration.seconds(15),
+		environment: {
+			NODE_OPTIONS: '--enable-source-maps',
```
johnduffell, line 38: this is the important new line 
https://nodejs.org/api/cli.html#--enable-source-maps

handlers/discount-api/src/index.ts
```diff
@@ -1,4 +1,3 @@
-import 'source-map-support/register';
```
johnduffell, line 1: no longer need this import as it done via `--enable-source-maps`

PR #3123, opened 2025-10-06
move metric-push-api to support stack and to SrCDK

johnduffell:
(Following various SrCDK PRs such as https://github.com/guardian/support-service-lambdas/pull/3121 )

I noticed that metric-push-api stack is membership but its domain is support.  

This PR updates metric-push-api to use SrCDK
It also makes it wholly a support rather than part membership/part support stacks.

This will also gain better logging/stack traces due to the source maps being enabled by default in SrCDK

## Note 
 because the stack name changes to `support-CODE-metric-push-api` it's necessary to drop the old stack and then deploy.  I had an issue dropping the CODE stack so I had to delete the APIGateway stage manually, with "Failed retrieving identifier ownership" error from CFN.

## Tested in CODE:

I think the alarm will still work fine as it goes off the app tag which is present
https://github.com/guardian/support-service-lambdas/blob/cb9507cfb7f6f7fb65996e46adba10ec59046154/handlers/alarms-handler/src/alarmMappings.ts#L81

image

Looks good in the logs (although I didn't send a referrer header so it didn't log the metric)
https://eu-west-1.console.aws.amazon.com/cloudwatch/home?region=eu-west-1#logsV2:log-groups/log-group/$252Faws$252Flambda$252Fmetric-push-api-CODE/log-events/2025$252F10$252F06$252F$255B$2524LATEST$255D2bdbbb3ba7cd4c36b7a3895f0cb07967

### DIFF
--- a/cdk/lib/__snapshots__/metric-push-api.test.ts.snap
+++ b/cdk/lib/__snapshots__/metric-push-api.test.ts.snap
@@ -5,7 +5,6 @@ exports[`The MetricPushApi stack matches the snapshot 1`] = `
   "Metadata": {
     "gu:cdk:constructs": [
       "GuDistributionBucketParameter",
-      "GuApiLambda",
       "GuPutCloudwatchMetricsPolicy",
       "GuAlarm",
     ],
@@ -83,6 +82,10 @@ exports[`The MetricPushApi stack matches the snapshot 1`] = `
         "Statistic": "Sum",
         "Tags": [
           {
+            "Key": "App",
+            "Value": "metric-push-api",
+          },
+          {
             "Key": "DiagnosticLinks",
             "Value": {
               "Fn::Join": [
@@ -106,7 +109,7 @@ exports[`The MetricPushApi stack matches the snapshot 1`] = `
           },
           {
             "Key": "Stack",
-            "Value": "membership",
+            "Value": "support",
           },
           {
             "Key": "Stage",
@@ -181,6 +184,10 @@ exports[`The MetricPushApi stack matches the snapshot 1`] = `
         "Statistic": "Sum",
         "Tags": [
           {
+            "Key": "App",
+            "Value": "metric-push-api",
+          },
+          {
             "Key": "gu:cdk:version",
             "Value": "TEST",
           },
@@ -190,7 +197,7 @@ exports[`The MetricPushApi stack matches the snapshot 1`] = `
           },
           {
             "Key": "Stack",
-            "Value": "membership",
+            "Value": "support",
           },
           {
             "Key": "Stage",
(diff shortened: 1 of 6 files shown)


cdk/lib/__snapshots__/metric-push-api.test.ts.snap
```diff
@@ -218,8 +225,7 @@ exports[`The MetricPushApi stack matches the snapshot 1`] = `
     },
     "MetricPushDNSRecord": {
       "Properties": {
-        "Comment": "CNAME for metric-push-api API CODE",
```
johnduffell, line 221: doesn't seem to appear anywhere

cdk/lib/__snapshots__/metric-push-api.test.ts.snap
```diff
@@ -289,16 +295,17 @@ exports[`The MetricPushApi stack matches the snapshot 1`] = `
...
-          "S3Key": "membership/CODE/metric-push-api/metric-push-api.zip",
+          "S3Key": "support/CODE/metric-push-api/metric-push-api.zip",
         },
         "Description": "API triggered lambda to push a metric to cloudwatch so we can alarm on errors",
         "Environment": {
           "Variables": {
             "APP": "metric-push-api",
             "App": "metric-push-api",
-            "STACK": "membership",
+            "NODE_OPTIONS": "--enable-source-maps",
```
johnduffell, line 305: this gives us better stack traces for a small perf cost

cdk/lib/__snapshots__/metric-push-api.test.ts.snap
```diff
@@ -454,7 +461,7 @@ exports[`The MetricPushApi stack matches the snapshot 1`] = `
                     {
                       "Ref": "AWS::AccountId",
                     },
-                    ":parameter/CODE/membership/metric-push-api",
+                    ":parameter/CODE/support/metric-push-api",
```
johnduffell, line 464: nothing in SSM 
image

handlers/metric-push-api/riff-raff.yaml
```diff
@@ -1,5 +1,5 @@
 stacks:
-  - membership
+  - support
```
johnduffell, line 2: now riffraff will put the zip in the expected place for the updated stack (but need to also update the CD config to know the new stack name)

cdk/lib/cdk/sr-api-lambda.ts
```diff
@@ -21,18 +21,25 @@ function getApiLambdaDefaultProps(scope: Identity, props: SrApiLambdaProps) {
...
-				apiKeyRequired: true,
-			},
+			...(props.isPublic
+				? {}
+				: {
+						apiKeySourceType: ApiKeySourceType.HEADER,
+						defaultMethodOptions: {
+							apiKeyRequired: true,
+						},
+					}),
```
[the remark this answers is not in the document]
johnduffell: it's private by default (undefined), I want people to actively opt into public

PR #3125, opened 2025-10-06
Use a simple error message with Zuora errors, not a concatenated one

rupertbates:
## What does this change?

It is useful to be able to access the raw error message returned by Zuora as easily as possible, but since some refactoring work was carried out on the Zuora client, we have been concatenating the status code and various other things onto the message which makes it hard to work with.

This PR removes the concatenation and adds some tests.

I have checked the codebase for any code which relies on the old structure and there is none. The only code I know of which checks the exact error message is in support-workers and that is currently broken, this change will fix it.

### DIFF
--- a/modules/zuora/src/zuoraClient.ts
+++ b/modules/zuora/src/zuoraClient.ts
@@ -104,7 +104,9 @@ export class ZuoraClient {
 		} else {
 			// When Zuora returns a 429 status, the response headers typically contain important rate limiting information
 			if (response.status === 429) {
-				logger.log(response.headers);
+				logger.log(
+					`Received a 429 rate limit response with response headers ${response.headers}`,
+				);
 			}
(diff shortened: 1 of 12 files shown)


modules/zuora/src/zuoraClient.ts
```diff
@@ -104,7 +104,9 @@ export class ZuoraClient {
 		} else {
 			// When Zuora returns a 429 status, the response headers typically contain important rate limiting information
 			if (response.status === 429) {
-				logger.log(response.headers);
+				logger.log(
+					`Received a 429 rate limit response with response headers ${response.headers}`,
+				);
```
rupertbates, line 109: If we're going to log this we should make it easy to see what it is in the logs.

tjmw (approved):

PR #3170, opened 2025-10-27
Mobile purchases to supporter product data - full sync

rupertbates:
## What does this change?
This PR adds a script to run a full synchronisation between the mobile purchase user subscription table and the SupporterProductData dynamo DB table. This is so that signed in users with subscriptions purchasd on the Apple app store or Google play store can be recognised on the website and we can suppress supporter revenue messages for them.

### Changes
- make it easier to provide defaults for config settings
- create a script to run the synchronisation
- add a README describing how to run a synchronisation


tjmw (approved):

modules/supporter-product-data/package.json
```diff
@@ -3,7 +3,8 @@
   "version": "1.0.0",
   "scripts": {
     "build": "tsc --noEmit",
-    "lint": "eslint src/**/*.ts test/**/*.ts"
+    "lint": "eslint src/**/*.ts test/**/*.ts",
+    "test": "NODE_OPTIONS=\"$NODE_OPTIONS --experimental-vm-modules\" jest --group=-integration"
```
johnduffell, line 7: I don't think `test` was needed - this package has no tests at all anyway.  Maybe it was meant to be called `it-test`

PR #3269, opened 2025-12-05
Implement Shared Eligibility Checker

andresilva-guardian:
## What does this change?

This pull request is based on [this](https://github.com/guardian/support-service-lambdas/pull/3176#discussion_r2510122273) suggestion from @johnduffell .

Since this is a big change, this approach is now being displayed as a pull request to keep focus on #3176


tsconfig.json
```diff
@@ -38,6 +38,7 @@
       "@modules/product-catalog/*": ["./modules/product-catalog/src/*"],
       "@modules/promotions/*": ["./modules/promotions/src/*"],
       "@modules/email/*": ["./modules/email/src/*"],
+      "@modules/eligibility/*": ["./modules/eligibility/src/*"],
```
johnduffell: well done for remembering to put it in here, it's easy to forget

pnpm-workspace.yaml
```diff
@@ -11,6 +11,7 @@ packages:
 catalog:
   zod: ^3.23.8
   typescript: ^5.6.3
+  dayjs: ^1.11.13
```
johnduffell: interesting idea to put it here - I'm not 100% on the benefits of this but it does sound like a good idea.

does that mean we should do a PR to move all the other uses of dayjs over to `catalog:` as well?

handlers/discount-api/src/eligibilityChecker.ts
```diff
@@ -1,140 +1,6 @@
...
-	twoMonthsMin: 'subscription was taken out more than 2 months ago',
-	notAlreadyUsed: 'this discount has not already been used',
-	noNegativePreviewItems: `next invoice has no negative items`,
-	isActive: 'subscription status is active',
-	zeroAccountBalance: 'account balance is zero',
-	atLeastCatalogPrice: 'next invoice must be at least the catalog price',
-	nextInvoiceGreaterThanZero: 'next invoice total must be greater than zero',
-	mustHaveDiscountDefined: 'subscription must have a discount defined',
-};
+// Re-exporting from shared module for backward compatibility
```
johnduffell: good idea to keep the diff clean.  It would be nice to have a direct reference in the longer term though.

modules/eligibility/src/eligibilityChecker.ts
```diff
@@ -0,0 +1,140 @@
+import { ValidationError } from '@modules/errors';
```
johnduffell: I assume this is just a direct copy/move of the existing file?  if not please could you flag any notable changes

johnduffell (approved):
great idea, and thanks for splitting it into a separate PR.

I think this PR could probably target main (and be merged first) rather than your branch, which would take some of the pressure off your other branch.

I've approved with a few minor comments/questions (nothing blocking) for you to check

johnduffell:
would be good to get this implemented, but given the amount of conflicts, I think it's going to be a start from scratch job

PR #3346, opened 2026-01-14
Use dayjs dates for date fields in SupporterRatePlanItem type

rupertbates:
## What does this change?
Up until now the `SupporterRatePlanItem` type in support-service-lambdas has had two fields `termEndDate` and `contractEffectiveDate` which are actually dates, but were represented in the model as strings. This is confusing to work with and risks introducing bugs due to the reduced type safety.

This PR changes the type of these fields to be a dayjs date.


tjmw (approved):

handlers/mobile-purchases-to-supporter-product-data/src/index.ts
```diff
@@ -84,8 +84,8 @@ export const fetchSubscriptionAndDoUpdate = async (
 		subscriptionName: subscription.subscriptionId,
 		productRatePlanId: inAppPurchaseProductRatePlanId,
 		productRatePlanName: subscription.productId,
-		termEndDate: subscription.to.toISOString(),
-		contractEffectiveDate: subscription.from.toISOString(),
+		termEndDate: subscription.to,
+		contractEffectiveDate: subscription.from,
```
johnduffell, line 88: this is a good fix (together with the custom replacer) because I think before it was putting something like `2026-04-22T05:43:33.333Z` in the table, which breaks MDAPI because it expects a date only, not a date-time with offset.
It seems like there wasn't a backfill done, but I think gradually the bad values will be overwritten.

PR #3440, opened 2026-02-27
write all lambda interactions into S3 for regression testing (alarms-handler)

johnduffell:
this is the next part extracted from the prototype/10% time PR https://github.com/guardian/support-service-lambdas/pull/3404

The overall project is to let us run regression tests using real data, but without any chance of disruption for users and our production systems.  There have been a series of standardisations to the architecture and logging to get the code in a state where it's possible as a proof of concept.

I'm using alarms-handler as a prototype, so far in preceding PRs I have made it log
- lambda input/ouput
- http requests via restClient
- aws requests wrapped in the request logger
- cold start requests/input

This is now close to being able to run regression tests, as the code should have access to all external I/O that the lambda does.

This PR adds a RequestLogger that stores up all the above in a data structure , then writes it to S3 for each invocation.

It is implemented for alarms-handler triggered lambda only, for a few reasons
- it's a small and non customer facing lambda
- if the regression test accidentally hits the real endpoint, it won't cause any customer issue
- it doesn't process PII thus lowering the risk
- it makes one request to AWS and one restClient request.
- it doesn't use the Router function (which does a "cold start" in terms of zuora auth etc every time)
- it's low traffic

an example invocation in CODE is here:
S3: https://eu-west-1.console.aws.amazon.com/s3/object/gu-reader-revenue-logs?region=eu-west-1&prefix=CODE/alarms-handler/1772233064893.json

cloudwatch: https://eu-west-1.console.aws.amazon.com/cloudwatch/home?region=eu-west-1#logsV2:log-groups/log-group/$252Faws$252Flambda$252Falarms-handler-CODE/log-events/2026$252F02$252F27$252F$255B$2524LATEST$255Dd3eb15d5812a4cba9ce7dc77cf804b1f

Next steps (next PR) would be to write a regression test that catalogues that bucket and runs through each file, running it as a sandboxed test.


cdk/lib/cdk/policies.ts
```diff
@@ -92,3 +93,12 @@ export class AllowSupporterProductDataQueryPolicy extends GuAllowPolicy {
 		});
 	}
 }
+
+export class AllowS3RegressionLogsWritePolicy extends GuPutS3ObjectsPolicy {
+	constructor(scope: GuStack) {
+		super(scope, 'S3 Regression Logs Write Policy', {
+			bucketName: 'gu-reader-revenue-logs',
```
johnduffell: the logs are written to s3 bucket `gu-reader-revenue-logs` under `/CODE/alarms-handler/TIMESTAMP.json`
the bucket has an expiry of 14 days

PR #3468, opened 2026-03-12
Replace new-product-api with new-subscription-api

rupertbates:
## What does this change?
This PR is the second part of piece of work to replace the new-product-api. It follows on from https://github.com/guardian/support-service-lambdas/pull/3503 and wraps the account creation code in that PR in an API Gateway API so that it can be called from Salesforce.

## Recap of the goals of this work
The aim of this work is to replace the new-product-api - the API which Salesforce uses to enable Customer Service Representatives (CSRs) to subscribe to new products on behalf of customers.  It will have the following benefits:
- To use a single piece of code when creating new subscriptions whether the request comes from Salesforce or support.theguardian.com - doing this will mean any new products or rate plans only need made in one place and there is no chance of Salesforce acquisitions being forgotten when adding these.
- To fully supports promotions created via the promo code tool in Salesforce
- It will use the same product catalog used by all of our newer applications rather than the non standard custom pricing code in the old api
- It will migrate from the deprecated Subscribe / Amend Zuora API to the new Orders API.
- It moves the codebase from Scala to Typescript in line with our strategic goal of standardising on Typescript

## [Asana Card](https://app.asana.com/1/1210045093164357/project/1213112839739792/task/1213603435718294)


cdk/lib/new-subscription-api.ts
```diff
@@ -0,0 +1,33 @@
...
+
+		const lambda = new SrApiLambda(this, 'Lambda', {
+			legacyId: `${this.app}-lambda`,
+			lambdaOverrides: {
+				description:
+					'A lambda that creates new Zuora subscriptions, replacing the new-product-api',
+			},
+			monitoring: {
+				errorImpact:
+					'a user may not be able to create a new subscription through the checkout flow',
```
tjmw: Is this right? We won't use this from the checkout flow will we? (Maybe that means something else in this context?)

handlers/new-subscription-api/package.json
```diff
@@ -0,0 +1,25 @@
+{
+  "name": "new-subscription-api",
+  "scripts": {
+    "test": "jest --group=-integration",
+    "it-test": "jest --group=integration",
+    "type-check": "tsc --noEmit",
+    "lint": "eslint --cache --cache-location /tmp/eslintcache/ 'src/**/*.ts' 'test/**/*.ts'",
+    "check-formatting": "prettier --check \"**/*.ts\"",
+    "fix-formatting": "prettier --write \"**/*.ts\"",
+    "build": "esbuild --bundle --platform=node --target=node20 --outdir=target/ src/index.ts --sourcemap --source-root=/support-service-lambdas/handlers/new-subscription-api/target/",
```
tjmw: Could we target a later Node version here? Does this match the lambda runtime? (Node 20 is EOL soon).

handlers/new-subscription-api/README.md
```diff
@@ -0,0 +1,86 @@
+# new-subscription-api
+
+## Overview
+
+An AWS Lambda that creates new Zuora subscriptions for Guardian products. It exposes a single endpoint:
```
tjmw:
```suggestion
An API gateway in front of an AWS Lambda that creates new Zuora subscriptions for Guardian products. It exposes a single endpoint:
```

handlers/new-subscription-api/README.md
```diff
@@ -0,0 +1,86 @@
...
+POST /subscription
+```
+
+On receiving a request it:
+1. Validates the request body (product/rate plan, payment method, contact details, optional promo code) using Zod.
+2. Fetches the product catalogue.
+3. Optionally fetches and applies a promotion.
+4. Calls the Zuora Orders API to create the subscription.
+
+Supported payment methods: Credit/Debit card (`CreditCardReferenceTransaction`), Direct Debit (`Bacs`), PayPal (`PayPalNativeEC`).
```
tjmw: This is the old PayPal payment method - will it be supported?

cdk/lib/new-subscription-api.ts
```diff
@@ -0,0 +1,33 @@
...
+} from './cdk/policies';
+import { SrApiLambda } from './cdk/SrApiLambda';
+import type { SrStageNames } from './cdk/SrStack';
+import { SrStack } from './cdk/SrStack';
+
+export class NewSubscriptionApi extends SrStack {
+	constructor(scope: App, stage: SrStageNames) {
+		super(scope, { stage, app: 'new-subscription-api' });
+
+		const lambda = new SrApiLambda(this, 'Lambda', {
```
tjmw, line 15: Does `SrApiLambda` handle auth for us?

tjmw (approved):
This looks great. I left a few questions/suggestions to take or leave.

cdk/lib/new-subscription-api.ts
rupertbates: It handles adding an API key yes

handlers/new-subscription-api/package.json
rupertbates: Good spot. This is set by the build check template, I've made a PR to update it here: https://github.com/guardian/support-service-lambdas/pull/3533

cdk/lib/new-subscription-api.ts
```diff
@@ -0,0 +1,32 @@
...
+			legacyId: `${this.app}-lambda`,
```
johnduffell, line 16: not needed, it's only for existing lambdas that we don't want cloudformation to replace
```suggestion
```

handlers/new-subscription-api/src/index.ts
```diff
@@ -0,0 +1,43 @@
...
+// main entry point from AWS
+export const handler: Handler = Router([
+	{
+		httpMethod: 'POST',
+		path: '/subscription',
+		handler: withBodyParser(
+			createSubscriptionRequestSchema,
+			async (event, path, body: CreateSubscriptionRequest) => {
+				logger.log('Received POST /subscription request', body);
+				logger.mutableAddContext(body.createdRequestId);
```
johnduffell: if this is a UUID it might take up a lot of space on each log line.  Will be worth reviewing the logs in CODE just to get a feel for readability.
looking at `createSubscriptionRequestSchema` I'd probably use `identityId` instead, that's more likely to link together related calls in a useful way (e.g. bug report where a CSR is trying a few ways to create a sub for a user)

handlers/new-subscription-api/src/index.ts
```diff
@@ -0,0 +1,43 @@
...
+
+// main entry point from AWS
+export const handler: Handler = Router([
+	{
+		httpMethod: 'POST',
+		path: '/subscription',
+		handler: withBodyParser(
+			createSubscriptionRequestSchema,
+			async (event, path, body: CreateSubscriptionRequest) => {
+				logger.log('Received POST /subscription request', body);
```
johnduffell, line 31: IIRC this should be logged by the Router automcaitcally
```suggestion
```

johnduffell, line 31: is this not already logged by the router or withBodyParser?

handlers/new-subscription-api/src/requestSchema.ts
```diff
@@ -0,0 +1,49 @@
...
+	salesforceAccountId: z.string(),
+	salesforceContactId: z.string(),
+	identityId: z.string(),
+	currency: isoCurrencySchema,
+	paymentGateway: paymentGatewaySchema,
+	existingPaymentMethod: existingPaymentMethodSchema,
+	billToContact: contactSchema,
+	productPurchase: productPurchaseSchema,
+	giftRecipient: giftRecipientSchema.optional(),
+	appliedPromotion: appliedPromotionSchema.optional(),
```
johnduffell: we only use the promo code, do we need the rest of `appliedPromotionSchema` to be passed in from SF? i.e. the region id?  Or should we be checking that value as well in case of duplicate promo codes?

handlers/new-subscription-api/src/requestSchema.ts
```diff
@@ -0,0 +1,49 @@
...
+	runBilling: z.boolean().optional(),
+	collectPayment: z.boolean().optional(),
```
johnduffell: did we decide to miss these two out to simplify things?

handlers/new-subscription-api/src/requestSchema.ts
```diff
@@ -0,0 +1,49 @@
...
+	acquisitionCase: z.string().optional(),
+	acquisitionSource: z.string().optional(),
+	createdByCSR: z.string().optional(),
```
johnduffell: are these really optional? I think they should always be passed

handlers/new-subscription-api/src/requestSchema.ts
```diff
@@ -0,0 +1,49 @@
+import { isoCurrencySchema } from '@modules/internationalisation/schemas';
+import { productPurchaseSchema } from '@modules/product-catalog/productPurchaseSchema';
+import { appliedPromotionSchema } from '@modules/promotions/v2/schema';
+import { giftRecipientSchema } from '@modules/zuora/createSubscription/giftRecipient';
+import { paymentGatewaySchema } from '@modules/zuora/orders/paymentGateways';
+import { z } from 'zod';
+
+const existingPaymentMethodSchema = z.object({
+	id: z.string(),
+	requiresCloning: z.boolean(),
```
johnduffell: Not a funcionality issue but this name looks a bit strange, it doesn't quite capture why it might need cloning (it's not really a client side choice, it's helping the backend decide what to do).  From the outside I wonder whether we should be focussing on the fact that it's either already attached or freshly created, rather than the mechanics of whether this lambda should clone it or not?
Something along the lines of alreadyInUse ?
```suggestion
	alreadyInUse: z.boolean(),
```
  Or maybe we can cover the rules in docuemntation e.g.

```suggestion
	requiresCloning: z.boolean().describe("true if it is already attached and in use by an existing account"),
```
A final option would be to see if the backend has enough information to determine itself, ideally without any additional calls to zuora.

handlers/new-subscription-api/src/responseSchema.ts
```diff
@@ -0,0 +1,14 @@
+import { z } from 'zod';
+
+export const createSubscriptionResponseSchema = z.object({
+	orderNumber: z.string(),
+	accountNumber: z.string(),
+	subscriptionNumbers: z.array(z.string()),
```
johnduffell, line 6: why would there be more than one?

handlers/new-subscription-api/src/responseSchema.ts
```diff
@@ -0,0 +1,14 @@
...
+	invoiceNumbers: z.array(z.string()).optional(),
+	paymentNumber: z.string().optional(),
+	paidAmount: z.number().optional(),
```
johnduffell, line 9: I guess these are both either set or not, consider to wrap them to enforce that
```suggestion
	payment: z.object({
		number: z.string(),
		paidAmount: z.number(),
	}).optional(),
```

handlers/product-switch-api/test/changePlan/prepare/subscriptionInformation.test.ts
```diff
@@ -150,10 +150,10 @@ describe('getSubscriptionInformation', () => {
 			chargedThroughDate: '2026-03-10', // it uses the effective start date of the contribution charge
 			ratePlanId: subscription.ratePlan.id,
 			chargeIds: [
-				productCatalogCode.SupporterPlus.ratePlans.Monthly.charges.Contribution
-					.id,
 				productCatalogCode.SupporterPlus.ratePlans.Monthly.charges.Subscription
 					.id,
+				productCatalogCode.SupporterPlus.ratePlans.Monthly.charges.Contribution
+					.id,
```
johnduffell, line 156: was this intentional?

handlers/new-subscription-api/src/index.ts
rupertbates: Kind of but it's mixed in with the rest of the api gateway request object and escaped so hard to read and copy.

handlers/new-subscription-api/README.md
```diff
@@ -0,0 +1,146 @@
+# new-subscription-api
+
+## Overview
+The new-subscriptions-api is an API which Salesforce uses to enable Customer Service Representatives (CSRs) to purchase
```
johnduffell:
```suggestion
The new-subscription-api is an API which Salesforce uses to enable Customer Service Representatives (CSRs) to purchase 
```

handlers/new-subscription-api/README.md
```diff
@@ -0,0 +1,146 @@
+# new-subscription-api
+
+## Overview
+The new-subscriptions-api is an API which Salesforce uses to enable Customer Service Representatives (CSRs) to purchase 
+to new subscriptions on behalf of customers.
```
johnduffell:
```suggestion
new subscriptions on behalf of customers.
```

handlers/new-subscription-api/README.md
johnduffell: any other places lower down

handlers/new-subscription-api/src/index.ts
rupertbates: As above

handlers/new-subscription-api/src/requestSchema.ts
rupertbates: Checking with Graham

handlers/new-subscription-api/src/requestSchema.ts
rupertbates: They are never passed by support-workers so are optional in the shared model but you're right that for this endpoint they should always be present.

handlers/new-subscription-api/src/responseSchema.ts
rupertbates: That response comes from Zuora and the Orders API allows you to create more than one subscription in a request

handlers/new-subscription-api/src/responseSchema.ts
rupertbates: Again, this response comes directly from Zuora so it would require some mapping to change and I'm not sure it's worth it

handlers/new-subscription-api/src/requestSchema.ts
rupertbates: I like the describe idea 👍

handlers/product-switch-api/test/changePlan/prepare/subscriptionInformation.test.ts
rupertbates: Yes, unfortunately as a result of updating the catalog fixture these rate plans are now coming back in a different order so the test will fail otherwise.

handlers/new-subscription-api/src/responseSchema.ts
johnduffell: are we just passing back what zuora returned then so we can extend it in future?  given it has a privileged zuora credential I feel like we should be specific about what we return rather than just passing it all on. (e.g. in case of injection attacks)

handlers/new-subscription-api/src/index.ts
johnduffell: fair enough (for another PR: maybe it would make sense for withBodyParser to log the cleaned up version as we're always going to want to see it it in any new lambda)

johnduffell: no idea why I wrote this twice

handlers/new-subscription-api/README.md
```diff
@@ -0,0 +1,145 @@
+# new-subscription-api
+
+## Overview
+The nenew-subscription-api is an API which Salesforce uses to enable Customer Service Representatives (CSRs) to purchase new subscriptions on behalf of customers.
```
johnduffell:
```suggestion
The new-subscription-api is an API which Salesforce uses to enable Customer Service Representatives (CSRs) to purchase new subscriptions on behalf of customers.
```

handlers/new-subscription-api/README.md
```diff
@@ -0,0 +1,145 @@
...
+#### CreditCardReferenceTransaction (Stripe)
+CreditCardReferenceTransaction payment methods are slightly simpler to clone
+- We retrieve the `tokenId` and `secondTokenId` from the existing payment method and pass them into the new account API call in the `newAccount.paymentMethod` parameter [Docs](https://developer.zuora.com/v1-api-reference/api/orders/post_order#orders/post_order/t=request&path=newaccount/paymentmethod).
+
+#### CreditCard/PayPal
+Non-tokenised credit card payment methods (ones where the full credit card details are held in Zuora) are not supported as we are unable to retrieve the credit card number to clone the account.
+PayPal may be clone in the same way as a `CreditCardReferenceTransaction` but this is not implemented currently.
+
+### This flow is detailed in this flow chart:
+![img.png](img.png)
```
johnduffell: should this have a better name?

handlers/new-subscription-api/README.md
```diff
@@ -0,0 +1,145 @@
...
+
+Send a `POST` request to the CODE endpoint, for instance with cURL:
+
+```
+curl --location 'https://new-subscription-api-code.support.guardianapis.com/subscription' \
+--header 'Content-Type: application/json' \
+--header 'x-api-key: [REPLACE_THIS_WITH_A_REAL_API_KEY]' \
+--data-raw '{
+    "accountName": "Test Account",
+    "createdRequestId": "a1b2c3d4-e5f6-7890-abcd-ef1234567899",
```
johnduffell, line 108: should we make this a header rather than putting it in the body?  that more closely matches what other APIs do and in fact there's a standard in progress for exactly that
https://github.com/ietf-wg-httpapi/idempotency

handlers/new-subscription-api/README.md
```diff
@@ -0,0 +1,145 @@
...
+        "ratePlan": "Monthly",
+        "amount": 12
+    }
+}'
+```
+
+Check the CloudWatch logs for execution details:
+
+```
+/aws/lambda/new-subscription-api-CODE
```
johnduffell, line 139: direct link would be a nice to have but not essential, if not we probably don't need this part at all

johnduffell (approved):
looks great, thanks for breaking down as it is much easier to review 👏

handlers/new-subscription-api/src/requestSchema.ts
rupertbates: Ok I've agreed with Graham that SF will send the promo code only.

handlers/new-subscription-api/src/responseSchema.ts
rupertbates: It is parsed via a zod schema so we won't be passing on anything which is not in that schema

handlers/new-subscription-api/README.md
rupertbates: I think that's a good idea but I'll do it in another PR

PR #3498, opened 2026-03-27
Migrate supporter-product-data from support-frontend

rupertbates:
## What does this change?
This PR moves the state machine and associated lambdas, which are responsible for synchronising data between Zuora and the SupporterProductData Dynamo DB table, from support-frontend into support-service-lambdas.

## Benefits
- Migrates the code to Typescript
- Migrates the cloudformation to CDK
- Moves all code related to the supporter product data pipeline into the same repo (and support-service-lambdas is a more appropriate place for it)
- Uses the Zuora catalog to determine whether a rate plan is a discount rather than the previous hard coded list of Ids

## Details
- This is a new project in support-service-lambdas called supporter-product-data-lambdas
- It consists of 
  - one state machine which extracts any changed subscription records (based on an incremental timestamp) from Zuora and writes the details to an SQS queue
  - one lambda which is triggered by the queue and writes the details into the SupporterProductData[STAGE] Dynamo DB table
- The state machine has two modes it can run in, `incremental` which only syncs records which have changed since it last ran, or `full` which syncs every active subscription in Zuora. A full sync can take several hours to run in PROD and uses a large amount of lambda resource, although the concurrency is throttled to prevent it impacting other lambdas in the account by hitting the maximum concurrency limit.
- This PR creates new versions of all the resources in the supporter product data stack except for the Dynamo DB tables and the SQS queue which triggers the lambda - this has been kept the same so that any other applications which write directly to the queue do not need to be updated.

## Testing
The state machine runs on a schedule - daily in CODE and every 5 minutes in PROD. It can also be triggered manually to do either a full or incremental sync with the input
```json
{
  "queryType": "full"
}
```
or 
```json
{
  "queryType": "incremental"
}
```

It has been running in incremental mode in CODE since the 30th of March with no errors.

## Deploying to PROD
- First we will disable the old state machine scheduled trigger and the event source mapping which sends SQS messages to the old ProcessSupporterRatePlanItemLambda lambda. See [this PR](https://github.com/guardian/support-frontend/pull/8018)
- Next merge this PR and wait for deployment to PROD
- Wait for executions to begin and verify that they are successful (state machine executions, lambda logs and dynamo state)

## TODO: 
- Delete the old Scala stack (ensuring that the SQS queuea and Dynamo tables are retained)


.github/copilot-instructions.md
```diff
@@ -38,3 +38,7 @@ the high level calls should be implemented in separate files as per modules/zuor
 ### CDK
 
 Use SrCDK constructs from ./cdk/lib/cdk/ where possible.
+
+
+### Verification
+Always run build, fix-formatting and eslint with a fix flag as well as all relevant tests after making changes to Typescript code.
```
johnduffell, line 44: I'd forgotten about this file, thanks for updating it!

cdk/lib/supporter-product-data-lambdas.ts
```diff
@@ -0,0 +1,254 @@
...
+
+		const lambdaArtifact = Code.fromBucket(
+			artifactBucket,
+			`support/${this.stage}/supporter-product-data-lambdas/supporter-product-data-lambdas.zip`,
+		);
+
+		const lambdaRole = new Role(this, 'SupporterProductDataLambdaRole', {
+			assumedBy: new ServicePrincipal('lambda.amazonaws.com'),
+			inlinePolicies: {
+				LambdaPermissions: new PolicyDocument({
```
johnduffell: consider to use the ones from policies.ts - there's at least one for Zuora Auth.
https://github.com/guardian/support-service-lambdas/blob/19326a686004348618fbc871587391683a4ffac5/cdk/lib/cdk/policies.ts#L73

even if they aren't really reusable to put in SrCDK, it can help readability to extract and name them within the same file (I'd do them as a named subclass but here below are a named function)
https://github.com/guardian/support-service-lambdas/blob/d4c875be5f279d481a5e92882aad4ccc33280eef/cdk/lib/stripe-disputes.ts#L143-L168

cdk/lib/supporter-product-data-lambdas.ts
```diff
@@ -0,0 +1,254 @@
...
+	Wait,
+	WaitTime,
+} from 'aws-cdk-lib/aws-stepfunctions';
+import { LambdaInvoke } from 'aws-cdk-lib/aws-stepfunctions-tasks';
+
+interface SupporterProductDataLambdasProps extends GuStackProps {
+	processItemMaxConcurrency: number;
+}
+
+export class SupporterProductDataLambdas extends GuStack {
```
johnduffell: if you extends [SrStack](https://github.com/guardian/support-service-lambdas/blob/c8513600718d3dd62e5ead160ed1a88eb3e6eec5/cdk/lib/cdk/SrStack.ts#L17) you get the stack "support" for free and it makes "app" required

cdk/lib/supporter-product-data-lambdas.ts
```diff
@@ -0,0 +1,254 @@
...
+		const artifactBucket = Bucket.fromBucketName(
+			this,
+			'SupporterProductDataLambdasDistBucket',
+			'membership-dist',
+		);
+
+		const lambdaArtifact = Code.fromBucket(
+			artifactBucket,
+			`support/${this.stage}/supporter-product-data-lambdas/supporter-product-data-lambdas.zip`,
+		);
```
johnduffell: I think GuLambdaFunction gives you this for free and uses the built in guardian bucket discovery rather than hard coding it.
If you use [SrLambda](https://github.com/guardian/support-service-lambdas/blob/b45d39f77f72069fcb6c89e4554ebf3708ab7925/cdk/lib/cdk/SrLambda.ts#L59) you also get source maps enabled and various other things including log format

cdk/lib/supporter-product-data-lambdas.ts
```diff
@@ -0,0 +1,254 @@
...
+			code: lambdaArtifact,
+			timeout: Duration.minutes(5),
+			memorySize: 512,
+			role: lambdaRole,
+			environment: { STAGE: this.stage },
+		});
+
+		const queue = new Queue(this, 'SupporterProductDataQueue', {
+			queueName: `supporter-product-data-lambdas-${this.stage}`,
+			visibilityTimeout: Duration.seconds(600),
```
johnduffell:
```suggestion
			visibilityTimeout: Duration.minutes(10),
```

handlers/supporter-product-data-lambdas/src/lambdas/addSupporterRatePlanItemToQueueLambda.ts
```diff
@@ -0,0 +1,201 @@
...
+				[{ Name: 'Stage', Value: stage }],
+				'supporter-product-data',
+			),
+		triggerSqsWriteAlarm: () =>
+			putMetric(
+				'SqsWriteFailure',
+				stage,
+				[{ Name: 'Stage', Value: stage }],
+				'supporter-product-data',
+			),
```
johnduffell: I can't see these alarms (or any at all) in the CDK, are they missed?

cdk/lib/supporter-product-data-lambdas.ts
```diff
@@ -0,0 +1,254 @@
...
+			handler: 'fetchResultsLambda.handler',
+			code: lambdaArtifact,
+			timeout: Duration.minutes(5),
+			memorySize: 512,
+			role: lambdaRole,
+			environment: { STAGE: this.stage },
+		});
+
+		const queue = new Queue(this, 'SupporterProductDataQueue', {
+			queueName: `supporter-product-data-lambdas-${this.stage}`,
```
johnduffell: having "lambdas" in the queue name feels a bit odd (and in the name of the project itself)
perhaps the handler should be called something more specific e.g. "supporter-product-data-zuora-sync"? and the queue could be "supporter-product-data-zuora-buffer" or "supporter-product-data-writes" depending whether other systems might also write to it

handlers/supporter-product-data-lambdas/src/lambdas/addSupporterRatePlanItemToQueueLambda.ts
```diff
@@ -0,0 +1,201 @@
...
+		processedCount: state.processedCount,
+		remainingMillis: getRemainingTimeInMillis(),
+	});
+
+	let rowIndex = 0;
+	let processedCount = state.processedCount;
+	let seenAnyRow = false;
+	let batch: IndexedItem[] = [];
+
+	const flushBatch = async (): Promise<void> => {
```
johnduffell: I'm not keen on nested functions especially big ones as they make the original function big and split it in half so it's hard to get a feeling for what's going on.  Can this be moved to the top level?

cdk/lib/supporter-product-data-lambdas.ts
```diff
@@ -132,6 +143,12 @@ export class SupporterProductDataLambdas extends GuStack {
 		const queue = new Queue(this, 'SupporterProductDataQueue', {
 			queueName: `supporter-product-data-lambdas-${this.stage}`,
 			visibilityTimeout: Duration.seconds(600),
+			deadLetterQueue: {
```
johnduffell: I didn't mention SrSqsLambda [1] before as there were a lot of differences, but now you've added a DLQ and queue alarm the differences are fairly small.  Worth deciding whether to define it manually as you've done or prefer to use the above construct (you would need to parameterise the `batchSize` at least though)

[1] https://github.com/guardian/support-service-lambdas/blob/d4c875be5f279d481a5e92882aad4ccc33280eef/cdk/lib/cdk/SrSqsLambda.ts

cdk/lib/supporter-product-data-lambdas.ts
```diff
@@ -0,0 +1,382 @@
...
+			role: lambdaRole,
+			environment: { STAGE: this.stage },
+		});
+
+		const queue = new Queue(this, 'SupporterProductDataQueue', {
+			queueName: `supporter-product-data-lambdas-${this.stage}`,
+			visibilityTimeout: Duration.seconds(600),
+			deadLetterQueue: {
+				queue: new Queue(this, 'SupporterProductDataDeadLetterQueue', {
+					queueName: `dead-letters-supporter-product-data-lambdas-${this.stage}`,
```
johnduffell: I prefer to put the "dead-letter" marker near the end of the queue, this is because the AWS console queue search only looks at the prefix, and I more commonly want to find the queue and its DLQ, rather than all the DLQs.

```suggestion
					queueName: `supporter-product-data-lambdas-dlq-${this.stage}`,
```

handlers/supporter-product-data-lambdas/src/lambdas/addSupporterRatePlanItemToQueueLambda.ts
```diff
@@ -0,0 +1,201 @@
...
+		if (batch.length === 0) {
+			return;
+		}
+
+		if (getRemainingTimeInMillis() < timeoutBufferInMillis) {
```
johnduffell: tiny nitpick but worth considering if you're in anyway
```suggestion
		if (getRemainingTimeInMillis() < requiredRemainingTimeInMillis) {
```

cdk/lib/supporter-product-data-lambdas.ts
rupertbates: Yeah I was struggling to find something to call it, I couldn't just use supporter-product-data because that's what the old stack was called and -lambdas seemed like a good compromise because it's mostly the lambdas which relate to the SupporterProductData table. Maybe supporter-product-data-zuora-sync is better but it's such a lot to type!

handlers/supporter-product-data-lambdas/src/lambdas/addSupporterRatePlanItemToQueueLambda.ts
rupertbates: Yes they were well spotted. They're there now

cdk/lib/supporter-product-data-lambdas.ts
johnduffell: yeah totally agree it's a lot to type, given how much we use it we need a shorter or code word for supporter-product-data

spd?  user-holdings? holdings? holdings-db?
supporter and data are probably pretty implicit

The handler could be holdings-db-zuora-sync
If we want to rename, is now the perfect time to do it?

cdk/lib/supporter-product-data-lambdas.ts
```diff
@@ -35,151 +25,124 @@ import {
...
-			environment: { STAGE: this.stage },
+		// Policies shared across all lambdas — applied after lambda creation via addToRolePolicy
+		const ssmPolicy = new PolicyStatement({
+			effect: Effect.ALLOW,
+			actions: [
+				'ssm:GetParametersByPath',
+				'ssm:GetParameter',
+				'ssm:PutParameter',
+			],
+			// Non-standard path used by this app — not covered by the GuCDK auto-policy (which uses /${stage}/${stack}/${app})
```
johnduffell: consider to copy them to the standard place (and delete the old ones once the old one is unused)?

cdk/lib/supporter-product-data-lambdas.ts
```diff
@@ -35,151 +25,124 @@ import {
...
+			resources: [
+				`arn:aws:ssm:${this.region}:${this.account}:parameter/supporter-product-data/${this.stage}/*`,
+			],
+		});
+		const s3Policy = new PolicyStatement({
```
johnduffell: I'd call it something like s3ExportBucketPolicy

cdk/lib/supporter-product-data-lambdas.ts
```diff
@@ -35,151 +25,124 @@ import {
...
+			effect: Effect.ALLOW,
+			actions: ['s3:PutObject', 's3:GetObject'],
+			resources: [
+				`arn:aws:s3:::supporter-product-data-export-${this.stage.toLowerCase()}/*`,
+			],
 		});
+		const dynamoPolicy = new PolicyStatement({
```
johnduffell: mainDynamoTableRWPolicy?

cdk/lib/supporter-product-data-lambdas.ts
```diff
@@ -35,151 +25,124 @@ import {
...
-			functionName: `supporterProductData-QueryZuora-${this.stage}`,
-			runtime: Runtime.NODEJS_22_X,
-			handler: 'queryZuoraLambda.handler',
-			code: lambdaArtifact,
-			timeout: Duration.minutes(5),
-			memorySize: 512,
-			role: lambdaRole,
-			environment: { STAGE: this.stage },
+		// Policies shared across all lambdas — applied after lambda creation via addToRolePolicy
+		const ssmPolicy = new PolicyStatement({
```
johnduffell: ssmConfigReadPolicy

cdk/lib/supporter-product-data-lambdas.ts
```diff
@@ -35,151 +25,124 @@ import {
...
-			timeout: Duration.minutes(5),
-			memorySize: 512,
-			role: lambdaRole,
-			environment: { STAGE: this.stage },
+		const queryZuora = new SrLambda(this, 'QueryZuoraLambda', {
+			legacyId: `supporterProductData-QueryZuora-${this.stage}`,
+			lambdaOverrides: {
+				timeout: Duration.minutes(5),
+				memorySize: 512,
+				environment: { STAGE: this.stage },
```
johnduffell, line 103: I think this comes for free with GuCDK

cdk/lib/supporter-product-data-lambdas.ts
```diff
@@ -35,151 +25,124 @@ import {
...
+			resources: ['*'],
+		});
+		const zuoraPolicy = new AllowZuoraOAuthSecretsPolicy(this);
+
+		const addSharedPolicies = (lambda: SrLambda) => {
+			lambda.addToRolePolicy(ssmPolicy);
+			lambda.addToRolePolicy(s3Policy);
+			lambda.addToRolePolicy(dynamoPolicy);
+			lambda.addToRolePolicy(cloudwatchPolicy);
+			lambda.addPolicies(zuoraPolicy);
```
johnduffell: not a major issue but I would have thought most of the lambdas would need distinct permissions, rather than all of them having everything

cdk/lib/supporter-product-data-lambdas.ts
```diff
@@ -35,151 +25,124 @@ import {
...
-			functionName: `supporterProductData-FetchResults-${this.stage}`,
-			runtime: Runtime.NODEJS_22_X,
-			handler: 'fetchResultsLambda.handler',
-			code: lambdaArtifact,
-			timeout: Duration.minutes(5),
-			memorySize: 512,
-			role: lambdaRole,
-			environment: { STAGE: this.stage },
+		const queryZuora = new SrLambda(this, 'QueryZuoraLambda', {
+			legacyId: `supporterProductData-QueryZuora-${this.stage}`,
```
johnduffell: since this is a new lambda it shouldn't have a legacyId, that's just if we are converting an existing stack to SrCDK (although perhaps this is because you don't want to drop and recreate the stack/lambdas in CODE?)

rupertbates: I've removed this now

cdk/lib/supporter-product-data-lambdas.ts
rupertbates: Actually I realised that I need to use the existing queue because other applications may be writing to it

cdk/lib/supporter-product-data-lambdas.ts
rupertbates: I could do this after but I'm not doing it initially because the app writes to parameter store as well as reading and needs to share the location between the Scala and TS app

handlers/supporter-product-data-lambdas/src/lambdas/addSupporterRatePlanItemToQueueLambda.ts
rupertbates: Going to leave as is

cdk/lib/supporter-product-data-lambdas.ts
rupertbates: I'm going to leave it as it is. I want to keep supporter product data in the name to link it to the Dynamo table and I think that the lambdas on the end makes it (at least somewhat) clear that it is lambdas related to that table.

cdk/lib/supporter-product-data-lambdas.ts
```diff
@@ -0,0 +1,315 @@
...
+			this,
+			'SupporterProductDataStateMachine',
+			{
+				stateMachineName: `supporter-product-data-lambdas-${this.stage}`,
+				definitionBody: DefinitionBody.fromChainable(definition),
+			},
+		);
+
+		const scheduleExpression =
+			this.stage === 'PROD' ? 'cron(*/5 * ? * * *)' : 'cron(0 9 ? * MON-FRI *)';
```
tjmw: Worth adding a comment explaining the cron expressions here? I always struggle to remember how to parse these!

cdk/lib/supporter-product-data-lambdas.ts
```diff
@@ -0,0 +1,315 @@
...
+			dimensionsMap: { StateMachineArn: stateMachine.stateMachineArn },
+			statistic: 'Sum',
+			period: Duration.seconds(60),
+			unit: Unit.COUNT,
+		});
+		const executionFailureAlarm = new Alarm(this, 'ExecutionFailureAlarm', {
+			alarmName: `Supporter Product Data step function Failure in ${this.stage}`,
+			alarmDescription: `The supporter-product-data-lambdas-${this.stage} step function has failed. Check the Step Functions console for details.`,
+			metric: new MathExpression({
+				expression: 'IF(HOUR(errors) > 5, errors)',
```
tjmw, line 260: Would it be worth moving that comment about supressing alarms to this line for clarity?

handlers/supporter-product-data-lambdas/src/handlers/fetchResultsLambda.ts
```diff
@@ -0,0 +1,46 @@
...
+const lazyDependencies = new Lazy<FetchResultsDependencies>(
+	buildDependencies,
+	'Building dependencies',
+);
+
+export const handler: Handler<
+	FetchResultsState,
+	AddSupporterRatePlanItemToQueueState
+> = async (event) => {
+	const dependencies = await lazyDependencies.get();
```
tjmw, line 40: Because of the lazy/run once behaviour here, will we definitely always get an up to date `lastSuccessfulQueryTime` across lambda invocations? (I remember there was a weird issue with this in the old Scala code!)

handlers/supporter-product-data-lambdas/src/services/csvService.ts
```diff
@@ -0,0 +1,64 @@
+const parseCsvLine = (line: string): string[] => {
```
tjmw, line 1: Is there a library we could use for CSV parsing? https://www.npmjs.com/package/csv-parse looks well maintained and widely used.

handlers/supporter-product-data-lambdas/src/services/csvService.ts
```diff
@@ -0,0 +1,64 @@
...
+ * a time. The first line is treated as the header. Keeps only a single row in
+ * memory at once, making it suitable for very large files.
+ */
+export async function* parseCsvStreamWithHeader(
+	lines: AsyncIterable<string> | Iterable<string>,
+): AsyncGenerator<Record<string, string>> {
+	// Wrap in an async generator so we can uniformly await .next() on both
+	// sync (Generator) and async (AsyncGenerator) inputs
+	async function* toAsync(): AsyncGenerator<string> {
+		yield* lines;
```
tjmw, line 46: A real life usage of generators!

handlers/supporter-product-data-lambdas/src/services/queryZuora.ts
```diff
@@ -0,0 +1,110 @@
...
+	config: ZuoraQuerierConfig;
+	discountProductRatePlanIds: string[];
+	postQuery: (request: BatchQueryRequest) => Promise<{ id: string }>;
+};
+
+const localIsoForQueryName = (date: dayjs.Dayjs): string =>
+	date.utc().toISOString().replace('Z', '');
+
+const buildBatchQueryRequest = (
+	queryType: string,
```
tjmw: Very minor but could be worth typing this more strongly?

```ts
type QueryType = "full" | "incremental";
```

handlers/supporter-product-data-lambdas/src/services/selectActiveRatePlansQuery.ts
```diff
@@ -0,0 +1,35 @@
...
+      ProductRatePlan.Name,
+      Subscription.ContractEffectiveDate,
+      Subscription.TermEndDate,
+      Subscription.Status
+      FROM
+      rateplan
+      WHERE
+      (Subscription.Status = 'Active' OR Subscription.Status = 'Cancelled') AND
+      (RatePlan.AmendmentType is null OR RatePlan.AmendmentType = 'NewProduct' OR RatePlan.AmendmentType = 'UpdateProduct') AND
+      ${excludeDiscountProductRatePlans(discountProductRatePlanIds)} AND
```
tjmw, line 34: If `discountProductRatePlanIds` was ever empty I think that would break the query? The return value would be an empty string so we'd have two `AND`s side by side. Is that a scenario worth guarding against?

handlers/supporter-product-data-lambdas/test/addSupporterRatePlanItemToQueueLambda.test.ts
```diff
@@ -0,0 +1,209 @@
...
+			},
+			() => 120000,
+			{
+				streamCsvRows: () => streamCsvRows(csv),
+				sendMessagesToQueue: sendBatch,
+				putLastSuccessfulQueryTime: noopAsync as never,
+			},
+		);
+
+		// Only rows 1 and 2 (indices 1 and 2) should be sent
```
tjmw: What do indices 1 and 2 refer to here? Aren't they 0 and 1 in the array?

tjmw (approved):
This looks good to me.

What is the deployment strategy for prod? It'd be great if there was a way to run this alongside the existing Scala code, possibly in some kind of dry run mode, so build confidence that it's doing the same thing.

handlers/supporter-product-data-lambdas/src/services/selectActiveRatePlansQuery.ts
rupertbates: I just made a change to that actually!

handlers/supporter-product-data-lambdas/test/addSupporterRatePlanItemToQueueLambda.test.ts
rupertbates: They are the row numbers from the full batch of records, the first row (index 0) has already been 'sent' (because we've set the `processedCount` to 1 in the input). So this test is ensuring that we skip the first row and only send the remaining rows. I've added some more comments to make this clearer

handlers/supporter-product-data-lambdas/src/handlers/fetchResultsLambda.ts
rupertbates: Yes this was an issue actually! I've changed the queryZuoraLambda to resolve it

handlers/supporter-product-data-lambdas/src/services/csvService.ts
rupertbates: I'd probably rather avoid adding a new dependency, it's another security and upgrade burden and this code is pretty straightforward.

buildcheck/data/build.ts
```diff
@@ -804,6 +804,37 @@ const userSubscriptionsApi: HandlerDefinition = {
...
 
+const supporterProductDataLambdas: HandlerDefinition = {
+	name: 'supporter-product-data-lambdas',
+	functionNames: [
+		'supporterProductData-QueryZuora-',
+		'supporterProductData-FetchResults-',
+		'supporterProductData-AddToQueue-',
+		'supporterProductData-ProcessItem-',
+	],
+	entryPoints: ['src/lambdas/*.ts'],
```
johnduffell: other ones that use a directory use handlers instead of lambdas, do you intend to deviate from that?
```typescript
	entryPoints: ['src/handlers/*.ts'],
```

cdk/bin/cdk.ts
```diff
@@ -169,6 +170,17 @@ stacks.forEach((Constructor) => {
...
 
+new SupporterProductDataLambdas(app, {
+	stage: 'CODE',
+	app: 'supporter-product-data-lambdas',
+	processItemMaxConcurrency: 30,
+});
+new SupporterProductDataLambdas(app, {
+	stage: 'PROD',
+	app: 'supporter-product-data-lambdas',
+	processItemMaxConcurrency: 50,
```
johnduffell: did you switch to a "new" statement instead of using the `const stacks` in order to get the extra parameter?
If it's just that, you can have a stage mappings in the template to move the info out of this central file:
```typescript
const mappings = {
  CODE: {
    processItemMaxConcurrency: 30
  },
  PROD: {
    processItemMaxConcurrency: 50
  }
}
```
this also guarantees that the snapshots+tests are using the actual values (which was/is a problem for the ones that parameterise in the tests and in here)

cdk/lib/supporter-product-data-lambdas.ts
johnduffell: @rupertbates is that not the case?

cdk/lib/supporter-product-data-lambdas.ts
```diff
@@ -0,0 +1,316 @@
...
+			'SupporterProductDataStateMachine',
+			{
+				stateMachineName: `supporter-product-data-lambdas-${this.stage}`,
+				definitionBody: DefinitionBody.fromChainable(definition),
+			},
+		);
+
+		// Every 5 minutes 24/7 in PROD, once a day Mon-Fri at 9am UTC in CODE
+		const scheduleExpression =
+			this.stage === 'PROD' ? 'cron(*/5 * ? * * *)' : 'cron(0 9 ? * MON-FRI *)';
```
johnduffell:
```suggestion
		const every5Minutes = 'cron(*/5 * ? * * *)';
		const weekdays9am = 'cron(0 9 ? * MON-FRI *)';
		const scheduleExpression = this.stage === 'PROD' ? every5Minutes : weekdays9am;
```

cdk/lib/supporter-product-data-lambdas.ts
```diff
@@ -0,0 +1,316 @@
...
+		const alarmsTopic = Topic.fromTopicArn(
+			this,
+			'AlarmsHandlerTopic',
+			`arn:aws:sns:${this.region}:${this.account}:alarms-handler-topic-${this.stage}`,
+		);
+		const addAlarmAction = (alarm: Alarm) => {
+			if (this.stage === 'PROD') {
+				alarm.addAlarmAction(new SnsAction(alarmsTopic));
+			}
+		};
```
johnduffell: interesting approach, why not use 
`actionsEnabled: scope.stage == 'PROD', `

https://github.com/guardian/support-service-lambdas/blob/6c9d15358583e4381c8ea8434760d17a34ad2c7e/cdk/lib/cdk/SrLambdaAlarm.ts#L29

cdk/lib/supporter-product-data-lambdas.ts
```diff
@@ -0,0 +1,316 @@
...
+
+		// Step function execution failure alarm
+		const executionFailuresMetric = new Metric({
+			namespace: 'AWS/States',
+			metricName: 'ExecutionsFailed',
+			dimensionsMap: { StateMachineArn: stateMachine.stateMachineArn },
+			statistic: 'Sum',
+			period: Duration.seconds(60),
+			unit: Unit.COUNT,
+		});
```
johnduffell, line 253: is there a free convenience method like lambdas have e.g.
```
stateMachine.metricFailures({
				statistic: 'Sum',
				period: Duration.minutes(1),
			})
```

cdk/lib/supporter-product-data-lambdas.ts
```diff
@@ -0,0 +1,316 @@
...
+			alarmDescription:
+				`Check the ${dlq.queueName} SQS dead-letter queue for details of the record which failed, ` +
+				`and the ProcessSupporterRatePlanItemLambda CloudWatch logs for what went wrong.`,
+			metric: new Metric({
+				namespace: 'AWS/SQS',
+				metricName: 'ApproximateNumberOfMessagesVisible',
+				dimensionsMap: { QueueName: dlq.queueName },
+				statistic: 'Average',
+				period: Duration.seconds(60),
+			}),
```
johnduffell: dlq
					.metric('ApproximateNumberOfMessagesVisible')
					.with({ statistic: 'Average', period: Duration.minutes(1) }),

cdk/lib/supporter-product-data-lambdas.ts
```diff
@@ -0,0 +1,316 @@
...
+			comparisonOperator: ComparisonOperator.GREATER_THAN_THRESHOLD,
+			threshold: 1,
```
johnduffell: did you want GREATER than 0 rather than 1?

handlers/supporter-product-data-lambdas/README.md
```diff
@@ -0,0 +1,11 @@
+# supporter-product-data-lambdas
+Moves the Supporter Product Data sync pipeline into this repo from support-frontend. It runs a Step Functions state machine to:
```
johnduffell: I think "Moves the..." is more related to the PR than the handler.
The doc link is helpful though.
Might be worth adding a screenshot of the state machine structure in here, just to help people get their bearings?

handlers/supporter-product-data-lambdas/src/services/queryZuora.ts
```diff
@@ -0,0 +1,110 @@
...
+				logger.log(
+					'lastSuccessfulQueryTime could not be parsed as a date, ignoring',
+					{ lastSuccessfulQueryTime },
+				);
+			} else {
+				incrementalTime = formatZuoraDateTime(parsed);
+			}
+		} else {
+			logger.log(
+				'No lastSuccessfulQueryTime found in config, running without incrementalTime filter',
```
johnduffell: what does it mean if we run it without the filter? and why do we have a "20 year" subtract instead if you do a "full"?

handlers/supporter-product-data-lambdas/src/services/queryZuora.ts
```diff
@@ -0,0 +1,110 @@
...
+	if (queryType === 'full') {
+		incrementalTime = formatZuoraDateTime(now.subtract(20, 'year'));
+	} else {
+		const lastSuccessfulQueryTime =
+			await dependencies.getLastSuccessfulQueryTime();
+		if (lastSuccessfulQueryTime !== undefined) {
+			const parsed = parseLastSuccessfulQueryTime(lastSuccessfulQueryTime);
+			if (parsed === undefined) {
+				logger.log(
+					'lastSuccessfulQueryTime could not be parsed as a date, ignoring',
```
johnduffell: surely we should alarm at this point? what happens if we ignore, will it start from 1970?

handlers/supporter-product-data-lambdas/src/services/fetchResults.ts
```diff
@@ -0,0 +1,120 @@
...
+	getResultFileResponse: (fileId: string) => Promise<Response>;
+	uploadToS3: (
+		filename: string,
+		body: ReadableStream<Uint8Array>,
+		contentLength: number,
+	) => Promise<void>;
+	putLastSuccessfulQueryTime: (time: string) => Promise<void>;
+};
+
+const getValueOrThrow = <T>(value: T | undefined, errorMessage: string): T => {
```
johnduffell: would the existing getIfDefined not be good enough here?

handlers/supporter-product-data-lambdas/src/handlers/fetchResults.ts
```diff
@@ -0,0 +1,120 @@
...
+		jobId: event.jobId,
+		fileId: batch.fileId,
+		recordCount: batch.recordCount,
+		full: batch.full,
+	});
+
+	const fileId = getValueOrThrow(
+		batch.fileId,
+		`Batch.fileId was missing in jobId ${event.jobId}`,
+	);
```
johnduffell, line 58: we can probably simplify the business logic here - if we're calling a function called "getResults" it can check all these things in the deserialiser rather than in the business logic
the same goes for fileResponse.ok and .body etc, we can wrap it in a function or whatever so we just ask for it and it either gives us the data or throws an error. 
Then we can see this function is just
1. get the result 
2. download the file,
3. upload the file to s3,
4. put the last successful query time

cdk/lib/supporter-product-data-lambdas.ts
```diff
@@ -0,0 +1,316 @@
...
+		fetchResults.addPolicies(zuoraOAuthPolicy);
+
+		const addToQueue = new SrLambda(
+			this,
+			'AddSupporterRatePlanItemToQueueLambda',
+			{
+				lambdaOverrides: {
+					functionName: `supporterProductData-AddToQueue-${this.stage}`,
+					handler: 'addSupporterRatePlanItemToQueueLambda.handler',
+					timeout: Duration.minutes(10),
```
johnduffell, line 131: why not 15 minutes? (i.e. the maximum)

handlers/supporter-product-data-lambdas/src/handlers/processSupporterRatePlanItem.ts
```diff
@@ -0,0 +1,52 @@
...
+	) => Promise<MinimalZuoraSubscription>;
+	writeItem: (item: SupporterRatePlanItem) => Promise<void>;
+};
+
+export const processItem = async (
+	item: SupporterRatePlanItem,
+	dependencies: ProcessItemDependencies,
+): Promise<void> => {
+	logger.resetContext();
+	logger.mutableAddContext(item.subscriptionName);
```
johnduffell, line 22: this might not work ok, because this function is wrapped in Promise.all with a batch size of 10, and we use a global logger rather than passing it around, it may change to the next record while the previous is in an `await`
I think it would be necessary to create one logger per record, happy to pair on it some time.

handlers/supporter-product-data-lambdas/src/local/runQueryZuora.ts
```diff
@@ -0,0 +1,33 @@
+/// <reference types="node" />
+
```
johnduffell: assuming this is a way of running it locally, I've been putting these kinds of thing in folders called `runManual` alongside `src` for discoverability.  We should ideally standardise one way or another.

handlers/supporter-product-data-lambdas/src/model/supporterRatePlanItem.ts
```diff
@@ -0,0 +1,32 @@
+import type { SupporterRatePlanItem } from '@modules/supporter-product-data/supporterProductData';
+import dayjs from 'dayjs';
+
+export class CsvDecodeError extends Error {
+	constructor(cause: unknown) {
+		super(`Failed to decode CSV row: ${String(cause)}`);
```
johnduffell: is cause actually `message` or is it really the exception that caused it?  If it's an exception then you can add another parameter `, {cause}` but I think maybe this is actually message

handlers/supporter-product-data-lambdas/src/handlers/addItemsToQueue.ts
```diff
@@ -0,0 +1,113 @@
+import { logger } from '@modules/logger/logger';
```
johnduffell, line 1: I wouldn't say this is a `services` thing, I think this file is business logic, normally I'd put this in a package called `steps` or similar.
As an alternaitve I'd consider having a package for each lambda i,e 4 packages, each having something like `addSupporterItemHandler.ts` and `addItemsSteps.ts`.  The advantage of doing it that way is the packages are more cohesive where things aren't intended to be shared so it's easier to break things down further and still find things.  Anything that's used across lambdas can stay in services though.

handlers/supporter-product-data-lambdas/src/services/configService.ts
```diff
@@ -0,0 +1,59 @@
...
+		);
+
+		if (!parameter.Parameter?.Value) {
+			throw new Error(`Parameter ${fullPath} not found`);
+		}
+
+		return parameter.Parameter.Value;
+	}
+
+	async getLastSuccessfulQueryTime(): Promise<string | undefined> {
```
johnduffell: if we're expecting a time we should probably parse it and throw if it's not valid or undefined?

handlers/supporter-product-data-lambdas/src/services/configService.ts
```diff
@@ -0,0 +1,59 @@
...
+		);
+
+		if (!parameter.Parameter?.Value) {
+			throw new Error(`Parameter ${fullPath} not found`);
+		}
+
+		return parameter.Parameter.Value;
+	}
+
+	async putLastSuccessfulQueryTime(time: string): Promise<void> {
```
johnduffell, line 47: if it's supposed to be a time maybe we can take a dayjs

handlers/supporter-product-data-lambdas/src/services/contributions.ts
```diff
@@ -0,0 +1,54 @@
+import { logger } from '@modules/logger/logger';
+import type { Stage } from '@modules/stage';
+import type { SupporterRatePlanItem } from '@modules/supporter-product-data/supporterProductData';
+import type { ProcessItemDependencies } from './processSupporterRatePlanItem';
+import type { MinimalZuoraSubscription } from './zuoraSubscriptionService';
+
+export const contributionIdsForStage = (stage: Stage): string[] =>
+	stage === 'PROD'
+		? ['2c92a0fc5aacfadd015ad24db4ff5e97', '2c92a0fc5e1dc084015e37f58c200eea']
```
johnduffell: assuming we don't want to get these from the product catalog, we do have could give them const names so we know what they are.  We could move the copy from discount-api and reuse it as they are already named nicely https://github.com/guardian/support-service-lambdas/blob/2dc923e95b6317b2a4d7e2d656ba6f0aac4bb36c/handlers/discount-api/src/productToDiscountMapping.ts#L83

handlers/supporter-product-data-lambdas/src/services/contributions.ts
```diff
@@ -0,0 +1,54 @@
...
+		return undefined;
+	}
+
+	return {
+		amount: firstCharge.price,
+		currency: firstCharge.currency,
+	};
+};
+
+export const addContributionAmountIfNeeded = async (
```
johnduffell, line 35: // we can't get the charges(+price override) from the main rateplan query, so fetch the amount if needed

handlers/supporter-product-data-lambdas/src/services/contributions.ts
```diff
@@ -0,0 +1,54 @@
...
+
+	const subscription = await dependencies.getSubscription(
+		item.subscriptionName,
+	);
+	const contributionAmount = contributionAmountFromZuoraSubscription(
+		subscription,
+		dependencies.contributionIds,
+	);
+
+	logger.log('Resolved contribution amount');
```
johnduffell: logger.log('Resolved contribution amount ${contributionAmount}');

handlers/supporter-product-data-lambdas/src/services/dynamoService.ts
```diff
@@ -0,0 +1,91 @@
...
+		super(`Failed to write to Dynamo: ${String(cause)}`, { cause });
+		this.name = 'DynamoWriteError';
+	}
+}
+
+export class DynamoService {
+	constructor(
+		stage: Stage,
+		private readonly tableName = process.env
+			.SUPPORTER_PRODUCT_DATA_TABLE_NAME ?? `SupporterProductData-${stage}`,
```
johnduffell: what sets process.env.SUPPORTER_PRODUCT_DATA_TABLE_NAME?

handlers/supporter-product-data-lambdas/src/services/zuoraSubscriptionService.ts
```diff
@@ -0,0 +1,32 @@
...
+const ratePlanSchema = z.object({
+	id: z.string(),
+	ratePlanCharges: z.array(ratePlanChargeSchema),
+});
+
+const minimalZuoraSubscriptionSchema = z.object({
+	ratePlans: z.array(ratePlanSchema),
+});
+
+export type MinimalZuoraSubscription = z.infer<
```
johnduffell, line 21: PricesOnlyZuoraSubscription?

handlers/supporter-product-data-lambdas/src/services/contributions.ts
```diff
@@ -0,0 +1,54 @@
...
+	stage === 'PROD'
+		? ['2c92a0fc5aacfadd015ad24db4ff5e97', '2c92a0fc5e1dc084015e37f58c200eea']
+		: ['2c92c0f85a6b134e015a7fcd9f0c7855', '2c92c0f85e2d19af015e3896e824092c'];
+
+export const contributionAmountFromZuoraSubscription = (
+	subscription: MinimalZuoraSubscription,
+	contributionIds: string[],
+) => {
+	const contributionRatePlan = subscription.ratePlans.find((ratePlan) =>
+		contributionIds.includes(ratePlan.id),
```
johnduffell: if these are contribution product rate plan ids how would they match the individual rate plan id of the specific subscription version?  Should we be comparing with the productRatePlanId on the subscription instead?

handlers/supporter-product-data-lambdas/src/services/contributions.ts
```diff
@@ -0,0 +1,54 @@
...
+	);
+	const firstCharge = contributionRatePlan?.ratePlanCharges[0];
+	if (firstCharge?.price === undefined) {
+		return undefined;
```
johnduffell: why would it be undefined? should we fail/alarm at this point?

johnduffell: even if we're processing old records e.g. full import, and the person has switched away from contributions since then, it should still be in the get response, even if the amount has changed since the record we're processing now.

handlers/supporter-product-data-lambdas/src/services/queryZuora.ts
rupertbates: It syncs every active subscription in Zuora

handlers/supporter-product-data-lambdas/src/services/queryZuora.ts
rupertbates: I've changed it to fail

handlers/supporter-product-data-lambdas/src/services/dynamoService.ts
rupertbates: AI hallucination, I've removed it

handlers/supporter-product-data-lambdas/src/services/contributions.ts
rupertbates: I'm alarming now if it's missing

handlers/supporter-product-data-lambdas/src/services/contributions.ts
rupertbates: I've taken them from the product catalog instead

handlers/supporter-product-data-lambdas/src/handlers/addItemsToQueue.ts
rupertbates: I've moved these files alongside the lambdas

handlers/supporter-product-data-lambdas/src/services/configService.ts
rupertbates: It gets passed around through the state as a string so doing this would mean converting to a date before passing to this function and then converting back to a string within it.

handlers/supporter-product-data-lambdas/src/local/runQueryZuora.ts
rupertbates: I've removed it, it wasn't particularly useful

PR #3503, opened 2026-04-01
Create Subscription with existing payment method

rupertbates:
## What does this change?

This PR is the first part of piece of work to replace the new-product-api - the API which Salesforce uses to enable Customer Service Representatives (CSRs) to subscribe to new products on behalf of customers. 

The main reasons for this migration are:
- To use the same account creation code for both the website and Salesforce. This will reduce the amount of code we have to maintain, reduce the risk of differences resulting in bugs, and make it easier to add new features in the future. 
- Support promotions created via the promo code tool - the current discount system in Salesforce is very different to the system used by the website and is quite complicated to maintain.
- Migrate from the deprecated Subscribe / Amend Zuora API to the new Orders API.
- Migrate the codebase from Scala to Typescript in line with our strategic aims

A key requirement of this work is to allow users who already have an existing Guardian product to reuse the payment method associated with that product without re-entering their details.

## How it works
This PR adds a new function - `cloneSubscriptionWithExistingPaymentMethod` which will create a new Zuora account and subscription in a single Zuora Orders API call while reusing a payment method from an existing account. It does not introduce any API functionality yet, that will come in a later PR.

The `createSubscriptionWithExistingPaymentMethod` function takes the following input:

```typescript
{
  acquisitionCase: '1234',
  acquisitionSource: 'CSR',
  createdByCSR: 'John Smith',
  runBilling: true,
  collectPayment: true,
  accountName: '001UD00000Utt18YAB',
  createdRequestId: 'IT-createSubExistingPM-NationalDelivery-A01113215',
  salesforceAccountId: '001UD00000Utt18YAB',
  salesforceContactId: '003UD000010K3cfYAC',
  identityId: '200663850',
  currency: 'GBP',
  paymentGateway: 'GoCardless',
  existingPaymentMethod: {
    id: '8ad08ef39d670e4a019d6c9a762e1357',
    requiresCloning: true,
  },
  billToContact: {
    firstName: 'TestFirstName',
    lastName: 'TestLastName',
    workEmail: 'test@thegulocal.com',
    country: 'United Kingdom',
    state: undefined,
  },
  productPurchase: {
    product: 'NationalDelivery',
    ratePlan: 'EverydayPlus',
    firstDeliveryDate: '2026-04-10,
    deliveryContact: {
      firstName: 'TestFirstName',
      lastName: 'TestLastName',
      workEmail: 'test@thegulocal.com',
      country: 'United Kingdom',
      state: undefined,
      city: 'London',
      address1: 'Delivery address',
      address2: '1 Test Street',
      postalCode: 'N1 9GU',
    },
    deliveryInstructions: 'Leave with concierge',
    deliveryAgent: 123,
  },
}

```
The notable part of this is the existingPayment object:
```typescript
  existingPaymentMethod: {
    id: '8ad08ef39d670e4a019d6c9a762e1357',
    requiresCloning: true,
  },
```
This object contains an id which identifies a payment method object in Zuora and a boolean which tells the function whether it needs to be cloned before use.

### requiresCloning=false
If `requiresCloning` is false, the existing id can be passed directly to Zuora in the Orders API call we use to create the new account and subscription. It is passed in the `hpmCreditCardPaymentMethodId`
parameter - [Docs](https://developer.zuora.com/v1-api-reference/api/orders/post_order#orders/post_order/t=request&path=newaccount/hpmcreditcardpaymentmethodid)

### requiresCloning=true
If requiresCloning is true, the payment method is attached to another account and must be re-created first, how we do this depends on the type of payment method:
#### BankTransfer (GoCardless / BACS)
- We retrieve the existing payment method using the older [CRUD Payment Method API](https://developer.zuora.com/v1-api-reference/older-api/payment-methods/object_getpaymentmethod) - we need to use this API because it returns both the `MandateID` and `TokenId` which the newer API does not and both of these are needed for the clone to succeed.
- We then create a new 'orphaned' payment method unattached to an account. We need to use the CRUD API for this as well [docs](https://developer.zuora.com/v1-api-reference/older-api/payment-methods/object_postpaymentmethod) because it has an `ExistingMandate` parameter, which again is required for the clone to succeed.
- We then pass the id of the new payment method into the Orders API in the same way as with `requiresCloning=false` above.

#### CreditCardReferenceTransaction (Stripe) 
CreditCardReferenceTransaction payment methods are slightly simpler to clone
- We retrieve the `tokenId` and `secondTokenId` from the existing payment method and pass them into the new account API call in the `newAccount.paymentMethod` parameter [Docs](https://developer.zuora.com/v1-api-reference/api/orders/post_order#orders/post_order/t=request&path=newaccount/paymentmethod).

#### CreditCard/PayPal 
Non-tokenised credit card payment methods (ones where the full credit card details are held in Zuora) are not supported as we are unable to retrieve the credit card number to clone the account. 
PayPal may be clone in the same way as a `CreditCardReferenceTransaction` but this is not implemented currently.

### This flow is detailed in this flow chart:
image

## How to test

Unit tests:
```
pnpm --filter zuora test --testPathPattern=createSubscriptionWithExistingPaymentMethod
```

Integration tests (requires CODE Zuora credentials):
```
NODE_OPTIONS=--experimental-vm-modules pnpm --filter zuora it-test --testPathPattern=createSubscriptionWithExistingPaymentMethod
```

Integration tests cover CreditCardReferenceTransaction, BankTransfer, and a promotion code scenario. Each test cleans up by deleting the cloned account on completion.

### Checking custom fields are set correctly in Zuora
Screenshot 2026-04-10 at 10 44 25

### Checking delivery instructions are passed to Zuora correctly
Screenshot 2026-04-10 at 10 45 32

---
  - https://app.asana.com/0/0/1213995425643026


shtukas (approved):
Nice!

graham228221 (approved):
This looks great, looking forward to using this.

modules/zuora/src/createSubscription/clonePaymentMethod.ts
```diff
@@ -0,0 +1,102 @@
...
+			hpmCreditCardPaymentMethodId?: never;
+	  };
+
+const createPaymentMethodResponseSchema = z.object({
+	Id: z.string(),
+});
+
+export async function cloneBankTransfer(
+	bankTransfer: BankTransferPaymentMethod,
+	zuoraClient: ZuoraClient,
```
johnduffell: tiny nit: normally we pass the zuoraClient first and the data second
```suggestion
	zuoraClient: ZuoraClient,
	bankTransfer: BankTransferPaymentMethod,
```
same on `clonePaymentMethod` below

modules/zuora/src/createSubscription/clonePaymentMethod.ts
```diff
@@ -0,0 +1,102 @@
...
+	| {
+			paymentMethod: ClonedCreditCardReferenceTransaction;
+			hpmCreditCardPaymentMethodId?: never;
+	  };
+
+const createPaymentMethodResponseSchema = z.object({
+	Id: z.string(),
+});
+
+export async function cloneBankTransfer(
```
johnduffell: does it need to be exported?

```suggestion
async function cloneBankTransfer(
```

modules/zuora/src/createSubscription/clonePaymentMethod.ts
```diff
@@ -0,0 +1,102 @@
...
+// requiresCloning: true — the PM is attached to an existing account and must be cloned before use.
+export type ExistingPaymentMethod = {
+	id: string;
+	requiresCloning: boolean;
+};
+
+// The clonePaymentMethod function will return either an inline payment method
+// object which can be passed to Zuora or an id of an existing orphan payment
+// method but never both.
+export type ClonePaymentMethodResult =
```
johnduffell: happy to keep as-is if it's a convention but it looked a bit odd
```suggestion
export type ClonedPaymentMethod =
```

modules/zuora/src/createSubscription/createSubscription.ts
```diff
@@ -93,6 +100,62 @@ export function getPromotionInputFields(
...
+		productPurchase.product,
+	);
+	const createSubscriptionOrderAction = buildCreateSubscriptionOrderAction({
+		productRatePlanId: productRatePlan.id,
+		contractEffectiveDate,
+		customerAcceptanceDate,
+		chargeOverride,
+		promotionInputFields,
+		termType: productRatePlan.termType,
+		termLengthInMonths: productRatePlan.termLengthInMonths,
```
johnduffell, line 150: why do we need to get these from the catalog, won't zuora do that internally?

modules/zuora/src/createSubscription/createSubscription.ts
```diff
@@ -115,7 +178,7 @@ export function buildCreateSubscriptionRequest<T extends PaymentMethod>(
 ): CreateOrderRequest {
 	const { deliveryContact, deliveryAgent, deliveryInstructions } = {
 		deliveryContact: undefined,
-		deliveryAgent: '',
+		deliveryAgent: undefined,
 		deliveryInstructions: undefined,
 		...productPurchase,
```
johnduffell: is there a way to do this more type safely?  I'm nervous after the consents issue last year

modules/zuora/src/createSubscription/createSubscriptionWithExistingPaymentMethod.ts
```diff
@@ -0,0 +1,136 @@
...
+			'Created by createSubscriptionWithExistingPaymentMethod.ts in support-service-lambdas',
+		subscriptions: [
+			{
+				orderActions: [createSubscriptionOrderAction],
+				customFields: subscriptionCustomFields,
+			},
+		],
+		processingOptions: {
+			runBilling: runBilling ?? true,
+			collectPayment: collectPayment ?? true,
```
johnduffell, line 127: would we not run them for performance reasons?  if a payment isn't due, billing should be a no-op

modules/zuora/src/createSubscription/createSubscriptionWithExistingPaymentMethod.ts
```diff
@@ -0,0 +1,136 @@
...
+		...productPurchase,
+	};
+
+	const subscriptionCustomFields = {
+		DeliveryAgent__c: deliveryAgent?.toString(),
+		ReaderType__c: getReaderType(giftRecipient, appliedPromotion),
+		LastPlanAddedDate__c: zuoraDateFormat(contractEffectiveDate),
+		InitialPromotionCode__c: appliedPromotion?.promoCode,
+		PromotionCode__c: appliedPromotion?.promoCode,
+		CreatedRequestId__c: createdRequestId,
```
johnduffell, line 89: I don't think we set this in new-product-api and I don't think it's needed to set as we use the idempodency key instead now

modules/zuora/src/orders/newAccount.ts
```diff
@@ -16,71 +16,85 @@ export type Contact = {
...
+		name: input.accountName,
+		currency: input.currency,
+		crmId: input.salesforceAccountId,
 		customFields: {
-			sfContactId__c: salesforceContactId,
-			IdentityId__c: identityId,
-			CreatedRequestId__c: createdRequestId,
+			sfContactId__c: input.salesforceContactId,
+			IdentityId__c: input.identityId,
+			CreatedRequestId__c: input.createdRequestId,
```
johnduffell, line 77: didn't realise we had this on account as well as subscription! is it needed?

modules/zuora/src/types/objects/paymentMethod.ts
```diff
@@ -19,7 +19,7 @@ export const BasePaymentMethodSchema = z.object({
 	lastTransactionStatus: z.string().nullable(),
 	maxConsecutivePaymentFailures: z.number().nullable(),
 	numConsecutiveFailures: z.number(),
-	paymentRetryWindow: z.string().nullable(),
+	paymentRetryWindow: z.union([z.string(), z.number()]).nullable(),
```
johnduffell: I realise this is existing but why wouldn't it be a number? zuora docs suggest it's a nullable number
https://developer.zuora.com/v1-api-reference/api/accounts/get_acountpaymentmethods#accounts/get_acountpaymentmethods/t=response&c=200&path=returnedpaymentmethodtype/paymentretrywindow

modules/zuora/src/index.ts
```diff
@@ -5,6 +5,8 @@ export * from './types';
 export * from './utils';
 export * from './account';
 export * from './billingPreview';
+export * from './createSubscription/createSubscription';
+export * from './createSubscription/createSubscriptionWithExistingPaymentMethod';
```
johnduffell: what's this file for? It doesn't seem to affect how we have to do imports, and in fact I now realise this is what makes intellij auto-complete with broken imports that I have to manually correct every time (for me at least)

modules/zuora/src/getPaymentMethodObjectById.ts
```diff
@@ -0,0 +1,55 @@
+import { z } from 'zod';
```
johnduffell, line 1: rename this file to `modules/zuora/src/getPaymentMethodObjectById.ts` ?

modules/zuora/test/createSubscriptionWithExistingPaymentMethod.test.ts
```diff
@@ -0,0 +1,362 @@
...
+			expect(parsed.newAccount.paymentMethod.type).toBe(
+				'CreditCardReferenceTransaction',
+			);
+			expect(parsed.newAccount.paymentMethod.tokenId).toBe('tok_stripe_123');
+			expect(parsed.newAccount.paymentMethod.secondTokenId).toBe(
+				'cus_stripe_456',
+			);
+			expect(parsed.newAccount.autoPay).toBe(true);
+			expect(parsed.processingOptions.runBilling).toBe(true);
+			expect(result.accountNumber).toBe('A00099999');
```
johnduffell: is it possible to assert these all together, as then we get a full description of the difference in the assertion failure rather than quitting on the first difference?

modules/zuora/test/createSubscriptionWithExistingPaymentMethod.test.ts
```diff
@@ -0,0 +1,362 @@
...
+			};
+			expect(ordersRequest.newAccount.hpmCreditCardPaymentMethodId).toBe(
+				'new-pm-id',
+			);
+			expect(ordersRequest.newAccount.paymentMethod).toBeUndefined();
+			expect(ordersRequest.newAccount.autoPay).toBe(true);
+			expect(ordersRequest.processingOptions.runBilling).toBe(true);
+			expect(ordersRequest.processingOptions.collectPayment).toBe(true);
+
+			expect(mockPost).toHaveBeenCalledTimes(2);
```
johnduffell: could this ever fail? as we have only mocked two resolved values, and we've already looked inside `mockPost.mock.calls[1]`

modules/zuora/test/createSubscriptionWithExistingPaymentMethod.test.ts
```diff
@@ -0,0 +1,362 @@
...
+			expect(result.accountNumber).toBe('A00099999');
+		});
+
+		it('passes createdRequestId as idempotency key', async () => {
```
johnduffell:
```suggestion
		it('passes createdRequestId as idempotency key to the order call', async () => {
```

johnduffell (approved):
nice one, very well thought out and structured, thanks for breaking it down to a small(er) PR, I had a good look through and the only things that popped up were minor nit picks, all limitations are well documented and clear. 👏 

EDIT:
We probably need to make sure we're only cloning working payment methods - so the existing account shouldn't be in payment failure and auto pay should be on.  Not sure if that should be in this function, or a precondition.

modules/zuora/src/index.ts
rupertbates: Yes it does that for me too in Intellilj! I would like to get rid of this file, it was something that someone did in a refactor a while back but I don't think it adds much value and in fact as you say ends up causing more frustration if you use Intellij. That's a different PR though.

modules/zuora/test/createSubscriptionWithExistingPaymentMethod.test.ts
rupertbates: I've used `expect.toMatchObject` which should do that.

modules/zuora/src/createSubscription/createSubscription.ts
rupertbates: We override them for fixed term products, see this comment: https://github.com/guardian/support-service-lambdas/blob/fdaf15b5a220663cd4d1019c6c5ef90e275139cd/modules/zuora/src/orders/orderActions.ts#L173-L178

modules/zuora/src/createSubscription/createSubscriptionWithExistingPaymentMethod.ts
rupertbates: I'm not really sure when or if we wouldn't run them, I could remove these altogether and just hard code to true but it felt like it might be useful at some point. Maybe I should YAGNI it?

modules/zuora/src/createSubscription/createSubscriptionWithExistingPaymentMethod.ts
rupertbates: I think you're right that we don't need it, I'll create a PR after this one to remove it because it affects a few other things.

modules/zuora/src/orders/newAccount.ts
rupertbates: Same as above

modules/zuora/test/createSubscriptionWithExistingPaymentMethod.test.ts
rupertbates: I guess it could call one function multiple times? I'll remove it though I don't think it's very useful.

modules/zuora/src/createSubscription/createSubscription.ts
rupertbates: This should always be correct because of the way that the product type is bundled with the required fields in the `ProductPurchase` type https://github.com/guardian/support-service-lambdas/blob/0989a9f79c4469bcf3f3e540cb57955ea915ae10/modules/product-catalog/src/productPurchaseSchema.ts#L218

I could put a manual check in here if you think it's worth it? It's just double checking what the type system already has but arguably worth it?

modules/zuora/src/createSubscription/createSubscription.ts
johnduffell: fair enough maybe I should start reading comments!
Maybe there was some issue with term misalignment in the past.

modules/zuora/src/createSubscription/createSubscription.ts
johnduffell: my thinking was more if deliveryAgent is changed or moved in productPurchase in some way, it would quietly stop writing the value (notwithstanding any test coverage), so in my scala brain rather than spreading over the top of a default I would do this:
```
deliveryAgent: productPurchase.productType === 'NationalDelivery' ? productPurchase.deliveryAgent : undefined
```
then if e.g. we change the name or structure so `productPurchase.deliveryAgent` doesn't exist it would stop compiling.  Presumably at the moment it would just stop using the value.

If it would catch that already then fine, what kind of manual check would you do?

modules/zuora/src/createSubscription/createSubscriptionWithExistingPaymentMethod.ts
johnduffell: if you're on the fence I'd say simplify it for now

We do switch off runbilling for S+ cancellation lambda for performance reasons as it sometimes takes 30+ seconds but I guess if that's an issue for new products we can add the flag back in

johnduffell (approved):
great thanks for updating all the nit picks and I think I've answered all your questions back to  me inline

modules/zuora/src/createSubscription/createSubscription.ts
rupertbates: Ok I think this is the sort of thing you were thinking of is it? https://github.com/guardian/support-service-lambdas/pull/3503/changes/5fab5f988ad80aa1c0f82de43b4fd9ef430e0d95

modules/zuora/src/createSubscription/createSubscriptionWithExistingPaymentMethod.ts
rupertbates: Actually I'm going to leave it for now because the createSubscription input object which includes them is shared with the new createSubscriptionWithExistingPaymentMethod function so it's actually slightly more work to exclude them from the new function than it is to just keep them.

modules/zuora/src/createSubscription/createSubscription.ts
johnduffell: yes!  it does look very verbose, but it provides an explicit path for every field for every product - in conjunction with your lint rule in the other PR it will stop compiling if anything is missed or changed.

PR #3515, opened 2026-04-14
Fix CVE-2026-33671

andresilva-guardian:
## What does this change?

Addresses CVE-2026-33671, a high severity ReDoS vulnerability in `picomatch` affecting versions `< 2.3.2`, `>= 3.0.0 < 3.0.2`, and `>= 4.0.0 < 4.0.4`.

The vulnerability allows crafted extglob patterns (e.g. `+(a|aa)`, `+(+(a))`) to compile into regular expressions with catastrophic backtracking behaviour, causing multi-second event loop blocking on non-matching inputs. An attacker able to supply glob patterns to an application using picomatch could trigger a denial of service.

Three `pnpm.overrides` entries have been added to package.json to force all transitive dependents onto patched versions:

- `picomatch@>=4.0.0 <4.0.4` pinned to `4.0.4`
- `picomatch@>=3.0.0 <3.0.2` pinned to `3.0.2`
- `picomatch@<2.3.2` pinned to `2.3.2`

The main callers in this repo were `@stylistic/eslint-plugin` (via `4.0.3`) and `micromatch`/`anymatch`/`jest-util` (via `2.3.1`). All are now resolved to patched versions.

Reference: https://nvd.nist.gov/vuln/detail/CVE-2026-33671, comparable to CVE-2024-4067 (micromatch ReDoS).

This is a dependency version bump with no production code changes. After applying the overrides and running `pnpm install`, `pnpm list picomatch --depth=10` was used to verify that only patched versions (`2.3.2` and `4.0.4`) appear across the entire dependency tree.

Unit testing is sufficient for this change as no application logic has been modified.

The Dependabot alert for CVE-2026-33671 should be automatically dismissed once the patched versions are detected on the default branch.

The picomatch patch releases (`2.3.2`, `3.0.2`, `4.0.4`) contain only the security fix with no breaking API changes, so the risk of regression is low. These packages are dev/build-time dependencies (ESLint, Jest) and are not bundled into any deployed Lambda artifact, so there is no production runtime risk.

## Images

N/A, no UI changes.

## Accessibility

N/A, no UI changes.


project/Dependencies.scala
```diff
@@ -8,19 +8,6 @@ object Dependencies {
...
-  val nettyOverrides: Seq[ModuleID] = Seq(
-    "io.netty" % "netty-codec-http2" % nettyVersion,
-    "io.netty" % "netty-codec-http" % nettyVersion,
-    "io.netty" % "netty-codec" % nettyVersion,
-    "io.netty" % "netty-transport" % nettyVersion,
-    "io.netty" % "netty-common" % nettyVersion,
-    "io.netty" % "netty-buffer" % nettyVersion,
-    "io.netty" % "netty-handler" % nettyVersion,
-    "io.netty" % "netty-transport-classes-epoll" % nettyVersion,
-  )
```
johnduffell: this conflicts with the other PR and I don't think it's related.  If it's needed should it be done in its own PR?

andresilva-guardian: Those lines were a duplicate of the nettyVersion and nettyOverrides definitions that already exist lower in the file (added for CVE-2026-33871/CVE-2026-33870). The duplication was causing the CI compilation failure (nettyVersion is already defined as value nettyVersion) that blocked this PR's build, so they were removed here to unblock it, but the fix itself is unrelated to CVE-2026-4926. I will fix all of these once I get to merge these open PRs.

package.json
```diff
@@ -42,7 +42,10 @@
 			"brace-expansion@>=1.0.0 <=1.1.11": "^1.1.12",
 			"brace-expansion@>=2.0.0 <=2.0.1": "^2.0.2",
 			"form-data": "^2.5.5",
-			"flatted": ">=3.4.0"
+			"flatted": ">=3.4.0",
+			"picomatch@>=4.0.0 <4.0.4": "4.0.4",
+			"picomatch@>=3.0.0 <3.0.2": "3.0.2",
```
johnduffell, line 47: we don't have 3 installed, is that just for future protection?

johnduffell (approved):
looks good thanks for tidying up, just one non blocking q about version 3 (as that line wasn't present in the other PR for this library)

package.json
andresilva-guardian: Yes!

PR #3517, opened 2026-04-15
Fix CVE-2026-4926

andresilva-guardian:
## What does this change?

Upgrades `path-to-regexp` from `8.2.0` to `^8.4.0` to address CVE-2026-4926 (High severity).

The vulnerability causes a denial of service via exponentially growing regular expressions when multiple sequential optional groups are used in route patterns. The fix was released in `path-to-regexp@8.4.0`.

The package is a transitive dependency pulled in via: `aws-sdk-client-mock` -> `sinon@18.0.1` -> `nise@6.1.1` -> `path-to-regexp@^8.1.0`.

Since we don't directly depend on `path-to-regexp`, the fix is applied via a scoped `pnpm.overrides` entry in the root package.json, targeting only the vulnerable range (`>=8.0.0 <8.4.0`) to avoid inadvertently affecting other versions (e.g. `6.3.0` used by `@google-cloud/bigquery`).

The change is limited to bumping a transitive dev dependency (`path-to-regexp` is only used by `nise`, which is a test-double library used by `sinon`, which is used by `aws-sdk-client-mock` for mocking in tests). No production code is affected. Verifying that `pnpm install` resolves `path-to-regexp` to `8.4.2` is sufficient.

The Dependabot alert for CVE-2026-4926 should be resolved once this is merged and the lockfile is updated.

Low risk. `path-to-regexp` is a transitive test-only dependency, so it is not included in any deployed Lambda bundle. There is no impact on production behaviour. The override is scoped to the vulnerable range so other version lines are unaffected.

## Images

N/A

## Accessibility

N/A


project/Dependencies.scala
```diff
@@ -8,19 +8,6 @@ object Dependencies {
...
-  val nettyOverrides: Seq[ModuleID] = Seq(
-    "io.netty" % "netty-codec-http2" % nettyVersion,
-    "io.netty" % "netty-codec-http" % nettyVersion,
-    "io.netty" % "netty-codec" % nettyVersion,
-    "io.netty" % "netty-transport" % nettyVersion,
-    "io.netty" % "netty-common" % nettyVersion,
-    "io.netty" % "netty-buffer" % nettyVersion,
-    "io.netty" % "netty-handler" % nettyVersion,
-    "io.netty" % "netty-transport-classes-epoll" % nettyVersion,
-  )
```
johnduffell: is this related?

andresilva-guardian: Not directly, no. Those lines were a duplicate of the nettyVersion and nettyOverrides definitions that already exist lower in the file (added for CVE-2026-33871/CVE-2026-33870). The duplication was causing the CI compilation failure (nettyVersion is already defined as value nettyVersion) that blocked this PR's build, so they were removed here to unblock it, but the fix itself is unrelated to CVE-2026-4926.

johnduffell (approved):
looks good thanks for tidying it up

PR #3519, opened 2026-04-16
Remove index.ts from the Zuora module

rupertbates:
I believe this was added to simplify the process of adding imports to various files in the Zuora module, however it breaks automatic imports in IntelliJ, requiring the user to then manually fix them every time.

On top of this it has not actually been used, as can be seen from the lack of changes in this PR, and it is the only module which takes this approach making it inconsistent within the mono-repo.


johnduffell (approved):
thanks that will be a big help!

PR #3526, opened 2026-04-21
Add a lint exhaustiveness check for switch statements

rupertbates:
## What does this change?
This PR switches on the ESLint `switch-exhaustiveness-check` rule at an error level. This rule can be really useful for picking up unhandled options when we for instance add a new product or rate plan to ensure that we are treating it correctly throughout the codebase.

I have also fixed the two places where we had inexhaustive switches.

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -31,4 +31,9 @@ export default [
 			],
 		},
 	},
+	{
+		rules: {
+			'@typescript-eslint/switch-exhaustiveness-check': 'error',
+		},
+	},
 ];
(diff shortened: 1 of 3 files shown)


paul-daniel-dempsey (approved):
Yep, looks great @rupertbates

eslint.config.mjs
```diff
@@ -31,4 +31,9 @@ export default [
 			],
 		},
 	},
+	{
+		rules: {
+			'@typescript-eslint/switch-exhaustiveness-check': 'error',
```
johnduffell, line 36: it does seem a little strict overall - should we allow default with `considerDefaultExhaustiveForUnions` so that you can use a switch to only cover a subset?

https://typescript-eslint.io/rules/switch-exhaustiveness-check/#considerdefaultexhaustiveforunions

johnduffell (approved):
the more bugs we catch statically the better 👍

eslint.config.mjs
rupertbates: Discussed offline and we agreed to go with the stricter setting initially

PR #3527, opened 2026-04-22
Zod schemas for payment gateways

rupertbates:
## What does this change?

This PR adds Zod schemas for the Zuora payment gateway type. Up until now we have just had TS types defined but I am going to need Zod schemas to validate them in the new-subscription API which I am currently working on. 
To try to keep that PR manageable I am introducing this change in isolation. 

I have also moved them into their own file paymentGateways.ts - before this they were bundled in with payment methods in paymentMethod.ts.

### DIFF
--- a/modules/zuora/src/account.ts
+++ b/modules/zuora/src/account.ts
@@ -1,7 +1,5 @@
-import type {
-	PaymentGateway,
-	PaymentMethod,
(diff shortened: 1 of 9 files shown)


maloky (approved):
Looks good 👍

PR #3532, opened 2026-04-27
Send mobile-purchases alarms to Platform team

tomrf1:
These are now owned by the Platform team.


handlers/alarms-handler/src/alarmMappings.ts
```diff
@@ -9,21 +9,6 @@ export type Team =
 	| 'ENGINE'
 	| 'PUZZLES';
 
-const mobilePurchasesApps = [
```
tomrf1, line 12: defining the list in a separate const isn't particularly useful.

sookburt (approved):

shtukas (approved):

PR #3534, opened 2026-04-29
Remove duplication in the `createSubscriptionWithExistingPaymentMethod` module

rupertbates:
createSubscriptionWithExistingPaymentMethod

## What does this change?
I noticed that we still had quite a bit chunk of duplicated code between the `createSubscription` function and the `createSubscriptionWithExistingPaymentMethod` function. Specifically this was the code in the function `buildCreateSubscriptionRequest` which was just duplicated in `createSubscriptionWithExistingPaymentMethod` meaning that any changes would need to be made in two places.

This PR fixes that, the main changes required to do this were:
- Make the `paymentMethod` field on the `CreateSubscriptionInputFields` type into  `T | ClonedPaymentMethod` instead of just T;
- Set the custom fields associated with CSR acquisitions in the `buildCreateSubscriptionRequest` function, it wasn't previously doing this because they are not sent by support-frontend.

All unit and integration tests still pass

### DIFF
--- a/modules/zuora/src/createSubscription/clonePaymentMethod.ts
+++ b/modules/zuora/src/createSubscription/clonePaymentMethod.ts
@@ -1,4 +1,5 @@
 import { z } from 'zod';
+import type { ExistingPaymentMethodInput } from '@modules/zuora/createSubscription/createSubscriptionWithExistingPaymentMethod';
 import {
 	type BankTransferPaymentMethod,
 	type CreditCardReferenceTransactionPaymentMethod,
@@ -9,14 +10,6 @@ import type {
 	ExistingPaymentMethod,
 } from '@modules/zuora/orders/paymentMethods';
 import type { ZuoraClient } from '@modules/zuora/zuoraClient';
-
-// Represents a Zuora payment method ID provided by the caller.
-// requiresCloning: false — the PM exists but is not yet attached to any account;
-// requiresCloning: true — the PM is attached to an existing account and must be cloned before use.
-export type ExistingPaymentMethodInput = {
-	id: string;
-	requiresCloning: boolean;
-};
 
 // The clonePaymentMethod function will return either an inline payment method
 // object which can be passed to Zuora or an id of an existing orphan payment
--- a/modules/zuora/src/createSubscription/clonePaymentMethod.ts
+++ b/modules/zuora/src/createSubscription/clonePaymentMethod.ts
@@ -4,29 +4,25 @@ import {
 	type CreditCardReferenceTransactionPaymentMethod,
 	getPaymentMethodObjectById,
 } from '@modules/zuora/getPaymentMethodObjectById';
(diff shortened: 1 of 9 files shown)


modules/zuora/src/createSubscription/clonePaymentMethod.ts
```diff
@@ -1,32 +1,21 @@
...
+import type {
+	ClonedCreditCardReferenceTransaction,
+	ExistingPaymentMethod,
+} from '@modules/zuora/orders/paymentMethods';
 import type { ZuoraClient } from '@modules/zuora/zuoraClient';
 
-// Represents a Zuora payment method ID provided by the caller.
-// requiresCloning: false — the PM exists but is not yet attached to any account;
-// requiresCloning: true — the PM is attached to an existing account and must be cloned before use.
-export type ExistingPaymentMethod = {
```
rupertbates, line 13: This type represents input data passed in from Salesforce, it has been moved to the createSubscriptionWithExistingPaymentMethod module and renamed to `ExistingPaymentMethodInput`. 

A new type `ExistingPaymentMethod` type has been added in the paymentMethod module which represents the data which gets passed to the Zuora Orders API.

modules/zuora/src/createSubscription/createSubscription.ts
```diff
@@ -51,7 +55,7 @@ export type CreateSubscriptionInputFields<T extends PaymentMethod> = {
 	identityId: string;
 	currency: IsoCurrency;
 	paymentGateway: PaymentGateway<T>;
-	paymentMethod: T;
+	paymentMethod: T | ClonedPaymentMethod;
```
tjmw, line 58: Is it okay that `ClonedPaymentMethod` and the generic `PaymentGateway<T>` above aren't related type-wise? The generic `T` used to guarantee that `paymentMethod` and `paymentGateway` were of the same generic type. Can `ClonedPaymentMethod` be used with _any_ `PaymentGateway<T>`?

rupertbates: Yes it can, I think this is necessary because it could be of several different types, we're trusting Salesforce to know what payment method it is sending us.

modules/zuora/src/orders/paymentGateways.ts
```diff
@@ -33,6 +33,7 @@ type PaymentGatewayMap = {
 	Bacs: GoCardlessPaymentGateway;
 	PayPalNativeEC: PayPalPaymentGateway;
 	PayPalCP: PayPalCompletePaymentsPaymentGateway;
+	ExistingPaymentMethod: GoCardlessPaymentGateway;
```
tjmw: `ExistingPaymentMethod` is always GoCardless?

tjmw (approved):
Nice improvement!

modules/zuora/src/orders/paymentGateways.ts
rupertbates: No, that's not right actually. They can be stripe as well, I've added that now.

PR #3535, opened 2026-04-29
Push to SupporterProductData after identity backfill

andrade0:
## What does this change?

After updating Zuora and Salesforce, `identity-backfill` now publishes one
`SupporterRatePlanItem` per active rate plan to `supporter-product-data-${Stage}`.
The existing consumer in `support-frontend` writes the records into the
`SupporterProductData-${Stage}` DynamoDB table.

This removes the manual workaround where CSRs had to create a renewal amendment
in Zuora to make a backfilled subscription show up in MMA. SQS publish failures
are logged but do not fail the API.

- 71/71 unit tests pass; new tests cover the SQS push and the ZOQL fetch.
- Manual smoke test against the CODE queue confirmed the message publish and the
  DynamoDB write end to end.
- `sbt assembly` and `aws cloudformation validate-template` both pass.

### DIFF
--- a/handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/IdentityBackfillSteps.scala
+++ b/handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/IdentityBackfillSteps.scala
@@ -4,6 +4,7 @@ import com.gu.identityBackfill.PreReqCheck.PreReqResult
 import com.gu.identityBackfill.TypeConvert._
 import com.gu.identityBackfill.Types._
 import com.gu.identityBackfill.salesforce.UpdateSalesforceIdentityId.IdentityId
+import com.gu.identityBackfill.supporterProductData.ZuoraSubscription
 import com.gu.salesforce.TypesForSFEffectsData.SFContactId
 import com.gu.util.Logging
 import com.gu.util.apigateway.ApiGatewayResponse
@@ -24,6 +25,7 @@ object IdentityBackfillSteps extends Logging {
       createGuestAccount: EmailAddress => ClientFailableOp[IdentityId],
       updateZuoraAccounts: (Set[AccountId], IdentityId) => ApiGatewayOp[Unit],
       updateSalesforceAccount: (Option[SFContactId], IdentityId) => ApiGatewayOp[Unit],
+      updateSupporterProductData: (Set[AccountId], IdentityId) => ApiGatewayOp[Unit],
   )(request: DomainRequest): ApiResponse = {
 
     (for {
@@ -35,9 +37,32 @@ object IdentityBackfillSteps extends Logging {
       }).toApiGatewayOp("create guest identity account")
       _ <- updateZuoraAccounts(preReq.zuoraAccountIds, requiredIdentityId)
       _ <- updateSalesforceAccount(preReq.maybeBuyer, requiredIdentityId)
-      // need to remember which ones we updated?
+      _ <- updateSupporterProductData(preReq.zuoraAccountIds, requiredIdentityId)
     } yield ApiGatewayResponse.successfulExecution).apiResponse
 
+  }
+
+  def pushSupporterProductData(
+      getSubscriptionsForAccount: AccountId => ClientFailableOp[List[ZuoraSubscription]],
+      sendToSqs: (List[ZuoraSubscription], IdentityId) => Either[String, Unit],
+  )(accountIds: Set[AccountId], identityId: IdentityId): ApiGatewayOp[Unit] = {
+    val subscriptions = accountIds.toList.flatMap { accountId =>
+      getSubscriptionsForAccount(accountId) match {
+        case ClientSuccess(subs) => subs
+        case failure: ClientFailure =>
+          logger.warn(
+            s"Failed to fetch active subscriptions for account ${accountId.value} " +
+              s"while updating SupporterProductData: ${failure.message}",
+          )
+          Nil
+      }
+    }
+    sendToSqs(subscriptions, identityId) match {
+      case Right(_) => ContinueProcessing(())
+      case Left(err) =>
+        logger.warn(s"Failed to publish SupporterProductData messages for identity ${identityId.value}: $err")
+        ContinueProcessing(())
+    }
   }
 
   def dryRunAbort(request: DomainRequest): ApiGatewayOp[Unit] =
--- a/handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/supporterProductData/UpdateSupporterProductData.scala
+++ b/handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/supporterProductData/UpdateSupporterProductData.scala
@@ -21,30 +21,30 @@ class UpdateSupporterProductDataService(
       subscriptions: List[ZuoraSubscription],
       identityId: IdentityId,
   ): Either[String, Unit] = {
-    val items = subscriptions.flatMap { sub =>
-      sub.ratePlans.map { rp =>
-        SupporterRatePlanItem(
-          subscriptionName = sub.subscriptionName,
-          identityId = identityId.value,
-          productRatePlanId = rp.productRatePlanId,
-          productRatePlanName = rp.ratePlanName,
-          termEndDate = sub.termEndDate,
-          contractEffectiveDate = sub.contractEffectiveDate,
-        )
-      }
-    }
+    val items = for {
+      sub <- subscriptions
+      rp <- sub.ratePlans
+    } yield SupporterRatePlanItem(
+      subscriptionName = sub.subscriptionName,
+      identityId = identityId,
+      productRatePlanId = rp.productRatePlanId,
+      productRatePlanName = rp.ratePlanName,
+      termEndDate = sub.termEndDate,
+      contractEffectiveDate = sub.contractEffectiveDate,
+    )
 
     if (items.isEmpty) {
       logger.info(s"No active rate plans for identity ${identityId.value}; nothing to send to ${queueName.value}")
       Right(())
     } else {
       logger.info(s"Sending ${items.size} rate plan messages to ${queueName.value}")
-      val errors = items.flatMap { item =>
-        sendMessage(queueName, Payload(item.asJson.noSpaces)) match {
+      val errors = for {
+        item <- items
+        error <- sendMessage(queueName, Payload(item.asJson.noSpaces)) match {
           case Left(err) => Some(s"${item.subscriptionName}/${item.productRatePlanId}: $err")
           case Right(_) => None
         }
-      }
+      } yield error
       if (errors.isEmpty) Right(()) else Left(errors.mkString("; "))
     }
   }
--- a/handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/supporterProductData/UpdateSupporterProductData.scala
+++ b/handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/supporterProductData/UpdateSupporterProductData.scala
@@ -1,0 +1,60 @@
+package com.gu.identityBackfill.supporterProductData
+
+import com.gu.effects.sqs.AwsSQSSend.{Payload, QueueName}
+import com.gu.effects.sqs.SqsSync
+import com.gu.identityBackfill.salesforce.UpdateSalesforceIdentityId.IdentityId
+import com.typesafe.scalalogging.LazyLogging
+import io.circe.syntax.EncoderOps
+import software.amazon.awssdk.services.sqs.SqsClient
+
+trait UpdateSupporterProductData {
+  def update(subscriptions: List[ZuoraSubscription], identityId: IdentityId): Either[String, Unit]
+}
+
+class UpdateSupporterProductDataService(
+    queueName: QueueName,
+    sendMessage: (QueueName, Payload) => Either[String, Unit],
+) extends UpdateSupporterProductData
+    with LazyLogging {
+
+  override def update(
+      subscriptions: List[ZuoraSubscription],
+      identityId: IdentityId,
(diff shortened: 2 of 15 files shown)


handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/supporterProductData/UpdateSupporterProductData.scala
```diff
@@ -0,0 +1,60 @@
...
+      logger.info(s"Sending ${items.size} rate plan messages to ${queueName.value}")
+      val errors = items.flatMap { item =>
+        sendMessage(queueName, Payload(item.asJson.noSpaces)) match {
+          case Left(err) => Some(s"${item.subscriptionName}/${item.productRatePlanId}: $err")
+          case Right(_) => None
+        }
+      }
+      if (errors.isEmpty) Right(()) else Left(errors.mkString("; "))
+    }
+  }
```
johnduffell, line 50: this can be written as for comprehensions which flattens out the structure thus making it easier to read and also more idiomatic scala
```suggestion
  override def update(
      subscriptions: List[ZuoraSubscription],
      identityId: IdentityId,
  ): Either[String, Unit] = {
    val items = for {
      sub <- subscriptions
      rp <- ratePlans
    } yield SupporterRatePlanItem(
      subscriptionName = sub.subscriptionName,
      identityId = identityId.value,
      productRatePlanId = rp.productRatePlanId,
      productRatePlanName = rp.ratePlanName,
      termEndDate = sub.termEndDate,
      contractEffectiveDate = sub.contractEffectiveDate,
    )

    if (items.isEmpty)
      logger.info(s"No active rate plans for identity ${identityId.value}; nothing to send to ${queueName.value}")
    else
      logger.info(s"Sending ${items.size} rate plan messages to ${queueName.value}")
    
    val errors = for {
      item <- items
      error <- sendMessage(queueName, Payload(item.asJson.noSpaces)) match {
        case Left(err) => Some(s"${item.subscriptionName}/${item.productRatePlanId}: $err")
        case Right(_) => None
      }
    } yield error

    if (errors.isEmpty) Right(()) else Left(errors.mkString("; "))

  }

```

handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/supporterProductData/UpdateSupporterProductData.scala
```diff
@@ -0,0 +1,60 @@
...
+
+  override def update(
+      subscriptions: List[ZuoraSubscription],
+      identityId: IdentityId,
+  ): Either[String, Unit] = {
+    val items = subscriptions.flatMap { sub =>
+      sub.ratePlans.map { rp =>
+        SupporterRatePlanItem(
+          subscriptionName = sub.subscriptionName,
+          identityId = identityId.value,
```
johnduffell: nice to have: normally we would keep the values wrapped in their value class until serialisation - this gives extra type safety as you can't mix up an IdentityId with a String.
Then there should be an Encoder for IdentityId that unwraps it to a plain string as it writes it out, and vice versa.

```suggestion
          identityId = identityId,
```

handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/zuora/GetSubscriptionsForAccount.scala
```diff
@@ -0,0 +1,60 @@
...
+      .map(err => com.gu.util.resthttp.Types.GenericError(s"could not parse $field for $sub: ${err.getMessage}"))
+      .toClientFailableOp
+
+  private def traverse[A, B](as: List[A])(f: A => ClientFailableOp[B]): ClientFailableOp[List[B]] =
+    as.foldLeft[ClientFailableOp[List[B]]](ClientSuccess(Nil)) { (acc, a) =>
+      for {
+        bs <- acc
+        b <- f(a)
+      } yield bs :+ b
+    }
```
johnduffell: nit pick: I guess it's only dealing with a tiny amount of data, but it feels like it would be better to use foldRight and prepend, as `bs :+ b` has to rebuild the list on every iteration.

Also consider to put the traverse function alongside any similar functions with com.gu.util.resthttp.Types.ClientFailableOp

handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/zuora/GetSubscriptionsForAccount.scala
```diff
@@ -0,0 +1,60 @@
...
+      rpQuery <- zoql"SELECT ProductRatePlanId, Name FROM RatePlan where SubscriptionId=${sub.Id}"
+      rpResults <- zuoraQuerier[WireRatePlan](rpQuery)
+      ratePlans = rpResults.records.toList.map(rp => ZuoraRatePlan(rp.ProductRatePlanId, rp.Name))
+      termEndDate <- parseDate(sub.TermEndDate, "TermEndDate", sub.Name)
+      contractEffectiveDate <- parseDate(sub.ContractEffectiveDate, "ContractEffectiveDate", sub.Name)
+    } yield ZuoraSubscription(sub.Name, termEndDate, contractEffectiveDate, ratePlans)
+
+  private def parseDate(value: String, field: String, sub: String): ClientFailableOp[LocalDate] =
+    Try(LocalDate.parse(value)).toEither.left
+      .map(err => com.gu.util.resthttp.Types.GenericError(s"could not parse $field for $sub: ${err.getMessage}"))
```
johnduffell: add an import up the top and reduce the full package name
```suggestion
      .map(err => GenericError(s"could not parse $field for $sub: ${err.getMessage}"))
```

handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/supporterProductData/UpdateSupporterProductData.scala
andrade0: Good call, much nicer to read. Done.

handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/supporterProductData/UpdateSupporterProductData.scala
andrade0: Agreed, kept it as IdentityId and added a Circe encoder for it in Model.scala.

handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/zuora/GetSubscriptionsForAccount.scala
andrade0: Good point on the perf. Switched to foldRight + prepend. Also moved `traverse` into Types.scala alongside the other ClientFailableOp helpers.

handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/zuora/GetSubscriptionsForAccount.scala
andrade0: Done, thanks.

handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/IdentityBackfillSteps.scala
```diff
@@ -24,6 +25,7 @@ object IdentityBackfillSteps extends Logging {
       createGuestAccount: EmailAddress => ClientFailableOp[IdentityId],
       updateZuoraAccounts: (Set[AccountId], IdentityId) => ApiGatewayOp[Unit],
       updateSalesforceAccount: (Option[SFContactId], IdentityId) => ApiGatewayOp[Unit],
+      updateSupporterProductData: (Set[AccountId], IdentityId) => ApiGatewayOp[Unit],
```
johnduffell, line 28: this old style of passing functions has been deprecated as we found it hard to follow, in new code we would pass in an implementation of a named trait/abstract class.
Up to you if you want to write your new functionality in that way (I wouldn't recommend refactoring the existing though in this PR)

handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/IdentityBackfillSteps.scala
```diff
@@ -35,11 +37,34 @@ object IdentityBackfillSteps extends Logging {
       }).toApiGatewayOp("create guest identity account")
       _ <- updateZuoraAccounts(preReq.zuoraAccountIds, requiredIdentityId)
       _ <- updateSalesforceAccount(preReq.maybeBuyer, requiredIdentityId)
-      // need to remember which ones we updated?
```
johnduffell, line 38: not sure what this old comment meant!  fine to delete.

handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/IdentityBackfillSteps.scala
```diff
@@ -35,11 +37,34 @@ object IdentityBackfillSteps extends Logging {
...
+      _ <- updateSupporterProductData(preReq.zuoraAccountIds, requiredIdentityId)
     } yield ApiGatewayResponse.successfulExecution).apiResponse
 
   }
 
+  def pushSupporterProductData(
+      getSubscriptionsForAccount: AccountId => ClientFailableOp[List[ZuoraSubscription]],
+      sendToSqs: (List[ZuoraSubscription], IdentityId) => Either[String, Unit],
+  )(accountIds: Set[AccountId], identityId: IdentityId): ApiGatewayOp[Unit] = {
+    val subscriptions = accountIds.toList.flatMap { accountId =>
```
johnduffell, line 49: can also be a for comprehension (it's rare to use flatMap directly in scala)

handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/IdentityBackfillSteps.scala
```diff
@@ -35,11 +37,34 @@ object IdentityBackfillSteps extends Logging {
...
+      getSubscriptionsForAccount(accountId) match {
+        case ClientSuccess(subs) => subs
+        case failure: ClientFailure =>
+          logger.warn(
+            s"Failed to fetch active subscriptions for account ${accountId.value} " +
+              s"while updating SupporterProductData: ${failure.message}",
+          )
```
johnduffell, line 56: do we need to know about this?  would an alarm make sense, or should we rely on bug reports?

handlers/identity-backfill/src/main/scala/com/gu/identityBackfill/IdentityBackfillSteps.scala
```diff
@@ -35,11 +37,34 @@ object IdentityBackfillSteps extends Logging {
...
+          Nil
+      }
+    }
+    sendToSqs(subscriptions, identityId) match {
+      case Right(_) => ContinueProcessing(())
+      case Left(err) =>
+        logger.warn(s"Failed to publish SupporterProductData messages for identity ${identityId.value}: $err")
+        ContinueProcessing(())
+    }
```
johnduffell, line 65: nitpick, clarity improvement
```suggestion
    val sqsResult = sendToSqs(subscriptions, identityId)
    sqsResult.left.foreach { err =>
        logger.warn(s"Failed to publish SupporterProductData messages for identity ${identityId.value}: $err")
    }
    ContinueProcessing(())
```

handlers/identity-backfill/src/test/scala/com/gu/identityBackfill/supporterProductData/UpdateSupporterProductDataTest.scala
```diff
@@ -0,0 +1,118 @@
...
+    val service = new UpdateSupporterProductDataService(queueName, recorder.send)
+
+    val result = service.update(List(subscription("S-1", List(ratePlanA))), identityId)
+
+    result shouldBe Right(())
+    recorder.sent should have size 1
+    recorder.sent.head.value should include("\"subscriptionName\":\"S-1\"")
+    recorder.sent.head.value should include("\"identityId\":\"99999\"")
+    recorder.sent.head.value should include("\"productRatePlanId\":\"rp-A-id\"")
+    recorder.sent.head.value should include("\"termEndDate\":\"2026-04-29\"")
```
johnduffell, line 51: nitpick, I'd be tempted to json decode the value and assert the whole thing, rather than one by one, for clarity and better assertion failures

up with class Record you can put something like 
```scala
    case class ParsedPayload(subscriptionName: String, identityId: String, productRatePlanId: String, termEndDate: String)
```
and inside
```scala
    val parsedSent = sent.map(curSent => Json.fromJson(Json.parse(curSent.value)).get)
```
then the tests are cleaner
```suggestion
    recorder.parsedSent.head should be(ParsedPayload("S-1", "99999", "rp-A-id", "2026-04-29")
```

handlers/identity-backfill/src/test/scala/com/gu/identityBackfill/StepsTest.scala
```diff
@@ -18,6 +18,7 @@ class StepsTest extends AnyFlatSpec with Matchers {
 
     var zuoraUpdate: Option[(Set[AccountId], IdentityId)] = None // !!
     var salesforceUpdate: Option[(Option[SFContactId], IdentityId)] = None // !!
+    var supporterProductDataUpdate: Option[(Set[AccountId], IdentityId)] = None // !!
```
johnduffell, line 21: no idea what the `!!` were for, I think perhaps some marker of something that needed fixing and never was!

johnduffell (approved):
great I think it's a good idea and well implemented, I could only really see some nitpicks and other comments, thanks for doing the work 👍

PR #3541, opened 2026-05-06
Observer Digital Benefits Authorisation API

paul-daniel-dempsey:
## What does this change?
An API which advises Tortoise Media systems that a user has an active “Observer” or “Blended” Guardian subscription.

POST /is-active
Request : {SubscriptionId: subscriptionNumber, postCode: ukPostcode}
Response : {active: boolean, renews: isoDate}

POST : https://observer-benefits-api-code.support.guardianapis.com/is-active
BODY : {	"subscriptionId": "A-S01098933", "postCode": "N1 9GU"}

Note: 
This api authenticates using an `x-api-key` within the header and requires generating/supplying. 
The token can be found under AWS API GateWay/API Keys (observer-benefits-api-CODE/PROD).

Found: Response 200
{
  "isActive": true,
  "renews": "2027-05-07T00:00:00.000Z"
}
Not Found: Response 200
{
  "isActive": false
}

---
  - https://app.asana.com/0/0/1213910213963446


handlers/observer-benefits-api/src/isActiveEndpoint.ts
```diff
@@ -0,0 +1,81 @@
...
+	});
+}
+
+function isValid(
+	guardianSubscription: GuardianSubscription,
+	account: ZuoraAccount,
+	postCode: string,
+): boolean {
+	const matchPostCode =
+		account.billToContact.zipCode?.replaceAll(' ', '') ===
```
tjmw: I wonder if we should lower case both values before comparing, as well as stripping whitespace?

maloky: I agree 👍

maloky: I guess this was an older comment. anyway I just moved it into a sanitisation function and using regex to replace all spaces

paul-daniel-dempsey: Yep, updated

maloky (approved):
Looking good 🚀

rupertbates (approved):
Great stuff!

PR #3542, opened 2026-05-06
Add identity-backfill-batch CLI for bulk backfill of missing identity ids

andrade0:
## What does this change?

Adds a TypeScript CLI under `modules/identity-backfill-batch` to run the `identity-backfill` API in bulk against a CSV. Operational follow-up to #3535: that PR made the API propagate the new IdentityId all the way to SupporterProductData; this PR lets us fix the existing ~5,000 cases identified by the BigQuery query in the Trello card.

The CSV fed to the script is now produced by an **upstream pre-filter step** that handles all the data validation: it drops rows where the Salesforce and Zuora emails differ, where the shipping address (Zuora sold-to vs Salesforce Contact `other_*`) mismatches after intelligent normalisation (state abbreviations, country codes, postcode spacing), or where there is no email at all. The script itself no longer filters anything — it processes every row in the CSV it receives.

For each CSV row the script picks the contact email (Salesforce first, Zuora bill-to as fallback), calls the API with `dryRun: true` to validate the pre-reqs, then again with `dryRun: false`. Outputs go to `processed.csv`, `rejected.csv`, `errors.csv` and a `summary.txt`. State is persisted so the run can be resumed.

## How to run

The full README is in `modules/identity-backfill-batch/README.md`. Quick version:

1. Generate the pre-filtered CSV (done outside this repo, from the BigQuery query in the Trello card enriched with shipping-address fields).

2. Export the API URL and key for the target stage (read via `aws --profile membership` from API Gateway):

   ```sh
   export IDENTITY_BACKFILL_URL_CODE="https://$(aws --profile membership --region eu-west-1 \
     apigateway get-rest-apis --query "items[?name=='identity-backfill-api-CODE'].id" \
     --output text).execute-api.eu-west-1.amazonaws.com/CODE/identity-backfill"

   export IDENTITY_BACKFILL_API_KEY_CODE="$(aws --profile membership --region eu-west-1 \
     apigateway get-api-keys --name-query identity-backfill-api-key-CODE \
     --include-values --query 'items[0].value' --output text)"
   ```

   Same pattern with `PROD` for the production run.

3. Run the script against the pre-filtered CSV:

   ```sh
   pnpm --filter identity-backfill-batch backfill-batch \
     --stage CODE \
     --csv ~/Downloads/missing_identity_final.csv \
     --rps 5 \
     --dry-run-only
   ```

   Drop `--dry-run-only` once the dry-run pass looks clean. Outputs land in
   `~/Downloads/identity-backfill-batch-results/<stage>-<timestamp>/`. Re-running
   against the same output dir resumes from `state.json` and skips already-done emails.

- `pnpm --filter identity-backfill-batch type-check / lint / check-formatting` all pass.
- Tested live in CODE. A `--dry-run-only` pass on two test accounts produced the expected `rejected.csv` with the actual Salesforce reasons. A real run on a fixed-up account backfilled the IdentityId end to end — verified the entry appeared in Zuora, in Salesforce, and in DynamoDB `SupporterProductData-CODE`.

- `summary.txt` of the PROD run gives the count of fixed cases.
- DynamoDB `SupporterProductData-PROD` should gain entries matching `processed.csv`.
- Bug reports about "I can't see my subscription in MMA" tied to missing identity links should drop after the PROD run.

- The script only calls the existing API, so all mutations follow the same code path already running in PROD.
- Data correctness is enforced upstream by the pre-filter step (email and shipping address consistency between Salesforce and Zuora); the script trusts its input.
- Idempotent: re-running is safe — already-backfilled cases fail the existing pre-req check and are skipped.
- Rate-limited (default 5 req/s, configurable) to avoid pressure on Identity / Zuora / Salesforce.
- The CSV input contains PII (emails); kept local, never committed. The temporary BigQuery PII access will be revoked once the PROD run is done.
- The script is run manually, not deployed — no continuous prod blast radius.

## Images

N/A

## Accessibility

N/A — CLI tool, no UI.


johnduffell:
Hi Olivier I just spotted this PR -

Did you know there is a repo support-zuora-scripts which might be better for these kinds of manual scripts?

If it really should be in this repo,  I don't think they should be in modules, as it adds noise. I'd favour making a `manual` or `scripts` folder.  Happy to take other views/reasoning though!

PR #3547, opened 2026-05-12
add a discount for t2->3 switches (in CODE only)

johnduffell:
As an incentive to get people from T2 -> T3 we want to give them an initial discount.

This PR implements the basic sketch of the discount in CODE/sandbox.

Tested in CODE
Preview response looks like this
```json
{
    "amountPayableToday": 1.5,
    "proratedRefundAmount": 12,
    "targetCatalogPrice": 18,
    "nextPaymentDate": "2026-06-12",
    "discount": {
        "discountedPrice": 13.5,
        "discountPercentage": 0,
        "upToPeriods": 1,
        "upToPeriodsType": "Months"
    }
}
```
And this subscription has been switched
https://apisandbox.zuora.com/platform/subscriptions/71a1aa202549e1b06d6e1c0d1ed5235d

The UI and confirmation email correctly stated the payment today will be £1.50 but of course they will need to be refreshed
image

TODO (can be in later PRs)
- enable in PROD
- add PROD product ids
- confirm actual discount amount and timescale
- check confirmation email
- confirm eligibility criteria
- suitable tests


handlers/product-switch-api/src/changePlan/switchDefinition/discounts.ts
```diff
@@ -25,3 +25,18 @@ export const annualContribHalfPriceSupporterPlusForOneYear: Discount = {
...
+	},
+	productRatePlanChargeId: {
+		PROD: 'tbc',
+		CODE: '4a02ba3ebdcb4a469c402e09895cc716',
+	},
+	name: 'Upsell - Supporter Plus to Digital Plus Switch tbc off',
+	upToPeriods: 1,
+	upToPeriodsType: 'Months',
+	discountPercentage: 0,
+};
```
johnduffell, line 42: see https://apisandbox.zuora.com/apps/Product.do?method=view&id=2c92c0f85721ff7c01572940171363db
rate plan and charge are called `Upsell - Supporter Plus to Digital Plus Switch tbc off`

j-ruda-guardian:
@johnduffell Question, why is the discountPercentage in the example that you left in the PR description 0?

handlers/product-switch-api/src/changePlan/prepare/switchInformation.ts
```diff
@@ -20,6 +20,7 @@ export async function getSwitchInformation(
 	input: ProductSwitchTargetBody,
 	account: ZuoraAccount,
 	subscription: GuardianSubscription,
+	discountEnabled: boolean,
```
j-ruda-guardian, line 23: Seems like discountEnabled is required in getSwitchInformation but optional in getTargetInformation. Is this intended?

j-ruda-guardian (approved):
LGTM left two questions :)

johnduffell:
> @johnduffell Question, why is the discountPercentage in the example that you left in the PR description 0?

sorry just because it was zero at that point! It can be any number up to 100 in reality.  I think I left it as 25 so if you hit CODE that's what you should see.

handlers/product-switch-api/src/changePlan/prepare/switchInformation.ts
johnduffell: thanks good spot but yeah I think that's fine

I made it required at first and fixed all the non-test code, and then fixed the tests by making it optional only where necessary.  Clearly nothing is testing that directly so  I'd rather leave it required - optionality adds a small extra complexity.

This parameter will be removed again once it's live in PROD.

PR #3548, opened 2026-05-12
Multiple accounts: new invitation endpoint

rupertbates:
## What does this change?
This is the second in a series which will build the new multiple accounts functionality defined in the [technical specification](https://docs.google.com/document/d/1eqMn9zkg-iRK3Gi9ecmgGh8WWUYpLX6B3cOasTRYpns/edit?tab=t.0)

This PR adds a single endpoint to allow the creation of new secondary user invitations

## Request
POST request to /invitation endpoint
``` JSON
{
  "subscriptionName": "A-S00974337",
  "secondaryUserEmail": "integration-test2+multiple-account@theguardian.com"
}
```
## Response
On success:
``` JSON
{
  "invitationCode":"RpwR62kMnAxe"
}
```

## Implemented
Validation: 
- The subscription must be an active subscription with the `multipleAccounts` benefit (eg. Digital Plus, Newspaper etc.) and must belong to the user creating the invitation
- The number of existing invitations for the subscription must be less than the maximum allowed (currently 5)
- There must not be an existing invitation for the same secondary user email and subscription
Dynamo integration:
If validation checks are met, the API will write a record to the `multiple-account-invitation-{STAGE}` Dynamo table

## TODO
- Email integration - we still need to send an invitation email to the invitee
- Staff users - because of the way that multiple accounts is implemented it will be non-trivial to add this benefit for staff members (they do not have an actual account in the supporter-product-data table) so this has been left as a TODO for now


handlers/multiple-account-api/src/validation.ts
```diff
@@ -0,0 +1,74 @@
...
+	ProductCatalog,
+	ProductKey,
+} from '@modules/product-catalog/productCatalog';
+import { logger } from '@modules/routing/logger';
+import type { ZuoraSubscription } from '@modules/zuora/types';
+import type { ZuoraCatalog } from '@modules/zuora-catalog/zuoraCatalogSchema';
+import dayjs from 'dayjs';
+import type { InvitationRepository } from './invitationRepository';
+
+const MAXIMUM_NUMBER_OF_INVITES = 5;
```
tjmw: Would `MAXIMUM_NUMBER_OF_INVITES_PER_SUBSCRIPTION` be even clearer perhaps?

handlers/multiple-account-api/src/validation.ts
```diff
@@ -0,0 +1,74 @@
...
+
+function productHasMultipleAccountsBenefit(productKey: ProductKey) {
+	return productBenefitMapping[productKey].includes('multipleAccounts');
```
tjmw, line 19: Really nice to be getting this from the product benefits module 👍

tjmw (approved):

PR #3550, opened 2026-05-12
Add new functionality to Identity module

rupertbates:
## What does this change?
This PR adds three Identity related functions to the Identity module:
- getUserByEmail - retrieves a user given an email address, returns undefined if there is no such user
- createGuestAccount - creates a new guest account given an email address - throws if there is already an account with that email
- getOrCreateIdentityId - tries to retrieve a user by email and if there is no such user it creates a new guest account

I have also renamed the identity.ts file which already existed in this module to oktaTokenHelper.ts as that is a better description of what it does.
   
A new integration test idapiIntegration.test.ts

---
  - https://app.asana.com/0/0/1214747405190374


modules/identity/src/identityClient.ts
```diff
@@ -0,0 +1,36 @@
+import { getSSMParam } from '@modules/aws/ssm';
+import type { Stage } from '@modules/stage';
+import type { Authorisation, BearerTokenProvider } from '@modules/zuora/auth';
```
tjmw, line 3: Are these imports more generic than the file name suggests? I.e. they're coming from the zuora module but afaict there's nothing related to zuora here? If that's the case perhaps `@modules/zuora/auth` could be renamed?

modules/identity/src/identityClient.ts
```diff
@@ -0,0 +1,36 @@
...
+	PROD: 'https://idapi.theguardian.com',
+};
+
+class IdentityTokenProvider implements BearerTokenProvider {
+	constructor(
+		private readonly token: string,
+		private readonly stage: Stage,
+	) {}
+
+	getAuthorisation(): Promise<Authorisation> {
```
tjmw, line 17: Does this need to return a promise? Could it just return `Authorisation`?

modules/identity/src/idapi.ts
```diff
@@ -0,0 +1,76 @@
...
+		typeof guestAccountResponseSchema
+	>(
+		`/guest?accountVerificationEmail=true`,
+		JSON.stringify({ primaryEmailAddress: email }),
+		guestAccountResponseSchema,
+	);
+	return response.guestRegistrationRequest.userId;
+};
+
+export const getOrCreateIdentityId = async (
```
tjmw: Just out of interest - is there a reason why get or create by identity ID is wrapped in a single method, but get or create by email address is split across two separate methods?

tjmw (approved):
Looks good! I just left a few comments/questions to take or leave.

modules/identity/src/identityClient.ts
rupertbates: That's because this client inherits from RestClient and they come from that. I think it's right that that whole set of types could move to a non-zuora namespace but I'm not going to do that in this PR.

modules/identity/src/identityClient.ts
rupertbates: It has to override the function from `BearerTokenProvider`

modules/identity/src/idapi.ts
rupertbates: I'm not sure I understand the question, this PR only has `getUserByEmail` which retrieves a user or `getOrCreateIdentityId` which checks if a user exists and if not creates them. 

I think I might rename the second function to `getOrCreateUserFromEmail` though, it seems more descriptive.

tjmw: Oh sorry it was the name that through me off! Yep agreed that's a better name 👍

PR #3552, opened 2026-05-14
get rid of common build to speed the build up

johnduffell:
At the moment the TS is built by first checking all the modules, and then kicking off the individual lambda builds.
This adds about 3-4 minutes delay on to every build, but each build is then reasonably quick as it only checks its own code.
image

Many of our lambdas have few dependencies, and we are constantly adding more tests and code to modules, so the common build gradually gets slower.

This PR
- scraps the common build completely
- makes each lambda and module declare its workspace dependencies
- makes each lambda build check exactly the right code that it's declared to depend on.

The result is that every single handler build takes less elapsed time to produce an artifact since the last commit, and some lambdas with few dependencies have completed within around a minute.

side by side Comparison
image

The slight downside is that we are reprocessing some modules in several builds, so we would use more GHA minutes overall.  However since they are not paid at present, it's not an immediate concern.

## Why can't we just run the common build in parallel and make it a required check for merge?

The riffraff upload happens after the main build, so if we kick it off before common has completed, it will upload something for potential CD to PROD that could later fail due to an undetected merge conflict in common.

TODOs for later prs
- assert that there are no unused declared dependencies.  Perhaps use knip to do that? https://github.com/guardian/support-frontend/pull/6605
- consider optimising by passing the file expressions directly into prettier/lint/etc rather than rerunning the tool repeatedly?
- can we run more things in parallel? e.g. CDK and main build?
- consider using pnpm modules more properly rather than depending on the source files, apparently it has some benefits for incremental building


.github/workflows/ci-typescript.yml
```diff
@@ -87,8 +61,15 @@ jobs:
           node-version-file: '.nvmrc'
           cache: 'pnpm'
       - run: pnpm install
-      - run: pnpm --filter "./buildcheck" snapshot:assert handlers/${{ matrix.subproject }}
+      - run: pnpm --filter "./buildcheck" type-check
+      - run: pnpm --filter "./buildcheck" check-formatting
+      - run: pnpm --filter "./buildcheck" lint
+      - run: pnpm --filter "./buildcheck" snapshot:assert 'ROOT|modules|handlers/${{ matrix.subproject }}'
```
johnduffell, line 69: they all run buildcheck, but at a few seconds, it doesn't seem a problem.

.github/workflows/ci-typescript.yml
```diff
@@ -87,8 +61,15 @@ jobs:
...
       - run: pnpm --filter cdk package ${{ matrix.subproject }}
+      - run: pnpm --filter ${{ matrix.subproject }}... type-check
+      - run: pnpm --filter ${{ matrix.subproject }}... lint
+      - run: pnpm --filter ${{ matrix.subproject }}... check-formatting
+      - run: pnpm --filter ${{ matrix.subproject }}... test
```
johnduffell, line 74: the dots means it does the handler and only the modules that it depends on, which gives a speed boost to handlers with few dependencies.

buildcheck/data/snippets/buildPackageJson.ts
```diff
@@ -39,7 +40,12 @@ export function buildPackageJson(
 		},
 		NOTICE1: notice(filename),
 		NOTICE2: 'all dependencies are defined in buildcheck/data/build.ts',
-		dependencies: pkg.dependencies,
+		dependencies: {
+			...pkg.dependencies,
+			...recordFromEntries(
+				pkg.moduleDeps.map((module) => [module.name, 'workspace:*']),
+			),
```
johnduffell, line 48: adding the module dependencies to the package.json allows pnpm to correctly filter the modules when we do `---filter <project>...` with the three dots

buildcheck/data/snippets/buildTsConfigJson.ts
```diff
@@ -1,8 +1,37 @@
...
+				'@modules/*': ['../../modules/*'], // special case, everyone gets this for free
+				...recordFromEntries(
+					transitiveDeps.map(
+						(moduleName) =>
+							[
+								`@modules/${moduleName}/*`,
+								[`../../modules/${moduleName}/src/*`],
+							] as const,
+					),
+				),
```
johnduffell, line 33: we put the `paths` for the compilerOptions in each individual tsconfig, and they are deleted from the top level one.
This means that compilation fails if we use a file in a module that isn't a declared dependency

buildcheck/data/templates/handler/package.json.ts
```diff
@@ -5,12 +5,17 @@ export default (pkg: HandlerDefinition) => {
...
 		build:
 			'esbuild --bundle --platform=node --target=node22 --outdir=target/ ' +
 			entryPoints +
 			` --sourcemap --source-root=/support-service-lambdas/handlers/${pkg.name}/target/`,
-		package: `pnpm type-check && pnpm lint && pnpm check-formatting && pnpm test && pnpm build && cd target && zip -qr ${pkg.name}.zip ./*.js.map ./*.js`,
+		package: [
+			'pnpm build',
+			'cd target',
+			`zip -qr ${pkg.name}.zip ./*.js.map ./*.js`,
+		].join(' && '),
```
johnduffell, line 25: package is just the lightweight task now, the GHA calls each task one by one.  This gives us better logs as it shows each stage with a heading and a separate time.  It also makes sure that it runs all the checks on the transitive dependencies, but only runs packageQuick on the handler itself.

The old package script is maintained as before in case people are still using it, it still doesn't do transitive dependencies

buildcheck/data/build.ts
```diff
@@ -22,13 +22,279 @@ export interface ModuleDefinition {
 	tsConfigExtra?: Record<string, unknown>;
 	testTimeoutSeconds?: number;
 	jestClearMocks?: boolean;
+	moduleDeps: ModuleDefinition[];
```
johnduffell: now each handler and module must declare its module dependencies (as well as the normal library dependencies)

buildcheck/data/build.ts
```diff
@@ -22,13 +22,279 @@ export interface ModuleDefinition {
...
 }
 
 export interface BuildDefinition {
 	handlers: HandlerDefinition[];
 	modules: ModuleDefinition[];
 }
```
johnduffell, line 33: the modules need to be moved before the handlers so that there are no forward references when building the dependency graph
Some of the modules needed swapping around for the same reason.
I will comment on any significant lines.

buildcheck/data/templates/handler/package.json.ts
```diff
@@ -5,12 +5,17 @@ export default (pkg: HandlerDefinition) => {
...
 		'cdk:test': 'pnpm --filter cdk test ' + pkg.name,
 		'cdk:test-update': 'pnpm --filter cdk test-update ' + pkg.name,
 		'update-lambda': `../../update-lambda.sh "${pkg.name}"${pkg.functionNames ? ` ${pkg.functionNames.join(' ')}` : ''}`,
```
johnduffell, line 28: the `buildcheck/data/build.ts` is marked as a "large diff" but it's an important part of the change to review

Image

buildcheck/data/build.ts
```diff
@@ -22,13 +22,279 @@ export interface ModuleDefinition {
...
+		...dep['@aws-sdk/client-cloudwatch'],
+		...dep['@aws-sdk/client-lambda'],
+		...dep['@aws-sdk/client-s3'],
+		...dep['@aws-sdk/client-sqs'],
+		...dep['@aws-sdk/client-ssm'],
+		...dep['@aws-sdk/credential-provider-node'],
+		...dep['@aws-sdk/lib-storage'],
+		...dep.zod,
+	},
+	moduleDeps: [moduleLogger],
```
johnduffell: this is the new property that every handler and module will need to keep up to date from now on

buildcheck/data/build.ts
```diff
@@ -22,13 +22,279 @@ export interface ModuleDefinition {
...
+const moduleIdentity: ModuleDefinition = {
+	name: 'identity',
+	dependencies: {
+		...dep['@okta/jwt-verifier'],
+		...dep['zod'],
+	},
+	devDependencies: {
+		...devDeps['@types/aws-lambda'],
+	},
+	moduleDeps: [moduleAws, moduleZuora],
```
johnduffell: moduleZuora looks a bit odd as an identity dependency!

This is because the RestClient is there (and should be split into its own module)

buildcheck/data/build.ts
```diff
@@ -22,13 +22,279 @@ export interface ModuleDefinition {
...
+	name: 'salesforce',
+	dependencies: {
+		...dep['zod'],
+	},
+	devDependencies: {
+		...devDeps['@types/jest'],
+		...devDeps['jest'],
+		...devDeps['ts-jest'],
+	},
+	moduleDeps: [moduleSecretsManager, moduleZuora],
```
johnduffell: again this `moduleZuora` dependency is because it uses RestClient, to fix in another PR

handlers/alarms-handler/package.json
```diff
@@ -19,7 +19,10 @@
   "dependencies": {
     "@aws-sdk/client-cloudwatch": "^3.940.0",
     "@aws-sdk/credential-providers": "^3.940.0",
-    "zod": "catalog:"
+    "zod": "catalog:",
+    "aws": "workspace:*",
+    "routing": "workspace:*",
+    "zuora": "workspace:*"
```
johnduffell, line 26: this is what the new auto generated file looks like - it tells pnpm what the module dependencies are so it can process only transitive deps rather than everything

handlers/alarms-handler/tsconfig.json
```diff
@@ -1,3 +1,37 @@
 {
-  "extends": "../../tsconfig.json"
+  "extends": "../../tsconfig.json",
+  "compilerOptions": {
+    "paths": {
```
johnduffell, line 4: this is also auto generated - this means that code can only depend on declared dependencies. It might speed up the IDEs as well as the scope of auto complete is reduced.
unlike the above package.json, this is the transitive dependencies, because we're not using modules "properly"

handlers/alarms-handler/package.json
```diff
@@ -8,7 +8,7 @@
     "check-formatting": "prettier --check \"**/*.ts\"",
     "fix-formatting": "prettier --write \"**/*.ts\"",
     "build": "esbuild --bundle --platform=node --target=node22 --outdir=target/ src/index.ts src/indexScheduled.ts src/indexSummary.ts --sourcemap --source-root=/support-service-lambdas/handlers/alarms-handler/target/",
-    "package": "pnpm type-check && pnpm lint && pnpm check-formatting && pnpm test && pnpm build && cd target && zip -qr alarms-handler.zip ./*.js.map ./*.js",
+    "package": "pnpm build && cd target && zip -qr alarms-handler.zip ./*.js.map ./*.js",
```
johnduffell: these are removed from the package command so that the GHA build can have a separate timed section for each

tsconfig.json
```diff
@@ -31,31 +31,8 @@
...
-			"@modules/routing/*": ["./modules/routing/src/*"],
-			"@modules/logger/*": ["./modules/logger/src/*"],
-			"@modules/secrets-manager/*": ["./modules/secrets-manager/src/*"],
-			"@modules/guardian-subscription/*": [
-				"./modules/guardian-subscription/src/*"
-			],
-			"@modules/*": ["./modules/*"]
-		},
+		//		"paths": {
+		//		},
```
johnduffell, line 35: we don't have any global paths any more, each lambda has a defined set

update-lambda.sh
```diff
@@ -43,11 +43,12 @@ fi
...
-  pushd "handlers/$PROJECT_NAME"
-  pnpm build
-  cd target && zip -qr "$PROJECT_NAME.zip" ./*.js.map ./*.js
-  popd
+  pnpm --filter "$PROJECT_NAME" package
 else
+  pnpm --filter "$PROJECT_NAME"... type-check
+  pnpm --filter "$PROJECT_NAME"... lint
+  pnpm --filter "$PROJECT_NAME"... check-formatting
+  pnpm --filter "$PROJECT_NAME"... test
```
johnduffell, line 51: this is the same as the GHA - it has to do all the checks itself if it's not run in --quick mode.

shtukas (approved):

buildcheck/data/snippets/buildTsConfigJson.ts
tomrf1: it might be useful to have this sort of explanation as comments in the code

johnduffell: @tomrf1 would you recommend to have it appearing in the managed tsConfig.json files as well, or just in the template here?

tomrf1: I think just here in`buildTsConfigJson.ts`, to help people understand the logic here

johnduffell: done, on line 24:

> adding the module list here instead of at the top level means we can only use declared dependencies

handlers/contributions-only-countries-api/package.json
```diff
@@ -8,7 +8,8 @@
     "check-formatting": "prettier --check \"**/*.ts\"",
```
johnduffell, line 8: almost all the changes below here are automatic or generated, the next significant files are the last two:
[scroll down to the tsconfig.json](#diff-b55cdbef4907b7045f32cc5360d48d262cca5f94062e353089f189f4460039e0)
and update-lambda.sh below it

buildcheck/data/build.ts
rupertbates: could we call it `moduleDependencies`? I try to avoid abbreviations where possible

buildcheck/data/build.ts
```diff
@@ -23,13 +23,275 @@ export interface ModuleDefinition {
...
+const moduleZuoraCatalog: ModuleDefinition = {
+	name: 'zuora-catalog',
+	dependencies: {
+		...dep['@aws-sdk/client-s3'],
+		...dep['zod'],
+	},
+	extraScripts: {
+		'update-catalog-fixtures': 'bash runManual/updateCatalogFixtures.sh',
+	},
+	moduleDeps: [moduleAws],
```
rupertbates: So we don't need to include logger here? The fact that aws already declares that is sufficient?

buildcheck/data/build.ts
```diff
@@ -23,13 +23,275 @@ export interface ModuleDefinition {
...
+const moduleRouting: ModuleDefinition = {
+	name: 'routing',
+	dependencies: {
+		...dep['zod'],
+		...dep['dayjs'],
+	},
+	devDependencies: {
+		...devDeps['@types/aws-lambda'],
+	},
+	moduleDeps: [moduleAws, moduleLogger, moduleZuora],
```
rupertbates: Here we do declare logger as well as aws, should we prefer one approach over the others?

tsconfig.json
rupertbates: I imagine this would work ok when using support-service-lambdas as a dependency in support-frontend but it would be worth checking

buildcheck/data/build.ts
johnduffell: yes they are transitive, it would probably be good practice to declare them anyway, but we don't have the need to enforce at this stage

buildcheck/data/build.ts
johnduffell: Personally I think it would be nice to prefer only being able to reference direct dependencies, but in practice we just compile everything in one go.  As a future PR we can add an enforcement step as part of CI, or we could look into using modules more properly which would give incremental compilation and other benfits.

buildcheck/data/build.ts
johnduffell: I've added the moduleLogger now in this case

edit: although upon looking, this one uses console.log rather than the logger, so it was actually right before.

buildcheck/data/build.ts
johnduffell: done

tsconfig.json
johnduffell: I changed
https://github.com/guardian/support-frontend/blob/574ad437f2bdef0d13be7d002d9040643966fca7/pnpm-workspace.yaml#L22
locally to point to 7ddd87c9f4646c5d0c96e14f5fcb50aca3de3dc6
then ran `pnpm install` and `pnpm --filter support-workers package`
and it produced a file `support-workers/target/typescript/support-workers.zip` without errors

Would that have checked the right things?

johnduffell: I changed 
[pnpm-workspace.yaml line 22](https://github.com/guardian/support-frontend/blob/574ad437f2bdef0d13be7d002d9040643966fca7/pnpm-workspace.yaml#L22)

locally from
`'@guardian/support-service-lambdas': github:guardian/support-service-lambdas#1533432fc5bab3b75fb01765bc978ddce4934501`
 to point to
7ddd87c9f4646c5d0c96e14f5fcb50aca3de3dc6

then ran
```
pnpm install
pnpm --filter support-workers package
```

which produced a `support-workers/target/typescript/support-workers.zip` and reported no errors

Is that checking the right things?

tomrf1 (approved):
👍 nice

rupertbates (approved):
Really like this work, well done 👏

PR #3553, opened 2026-05-15
run buildcheck on modules

johnduffell:
When we added modules to buildcheck in https://github.com/guardian/support-service-lambdas/pull/3292 and related PRs, we didn't make it actually check them in the build.

This PR adds the required check to the common build.

It still takes a tiny amount of time to run 
image
And see the output so we can see the result
https://github.com/guardian/support-service-lambdas/actions/runs/25912532753/job/76161006573?pr=3553#step:10:23


rupertbates (approved):
👍

PR #3559, opened 2026-05-15
Standardise API gateway responses

rupertbates:
## What does this change?
We have a lot of API gateway projects in support-service-lambdas which all need to return appropriate response codes according to application state. This PR introduces four functions to simplilfy building the correct response object so that all our APIs do so in a standard way and with as little boilerplate as possible.

The functions are:
- `ok` - returns a 200 response with a json body
- `created` - returns a [201 response](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/201) with a json body and an optional location header for the created resource
- `badRequest` - returns a 400 response with a json body containing a custom error message
- `internalServerError` - returns a 500 error with json body with the message 'Internal server error'

There is also a utility function which takes an error and maps it to the correct API response based on error type - this already existed in the repo but was in a single handler so couldn't be used more widely. It has now been moved into the routing module alongside the new functions.


shtukas (approved):

tomrf1 (approved):

johnduffell (approved):
great idea 👍 
In future we could even pass an optional schema in to make sure we only return what we expect https://github.com/guardian/support-service-lambdas/blob/main/handlers/discount-api/src/stringify.ts

rupertbates:
> great idea 👍 In future we could even pass an optional schema in to make sure we only return what we expect https://github.com/guardian/support-service-lambdas/blob/main/handlers/discount-api/src/stringify.ts



> great idea 👍 In future we could even pass an optional schema in to make sure we only return what we expect https://github.com/guardian/support-service-lambdas/blob/main/handlers/discount-api/src/stringify.ts

Good idea, I've added that as well

johnduffell:
> > great idea 👍 In future we could even pass an optional schema in to make sure we only return what we expect https://github.com/guardian/support-service-lambdas/blob/main/handlers/discount-api/src/stringify.ts
> 
> Good idea, I've added that as well

Is there a commit missing as I can't see anything relating to that?

rupertbates:
> > > great idea 👍 In future we could even pass an optional schema in to make sure we only return what we expect https://github.com/guardian/support-service-lambdas/blob/main/handlers/discount-api/src/stringify.ts
> > 
> > 
> > Good idea, I've added that as well
> 
> Is there a commit missing as I can't see anything relating to that?

I made it on the wrong branch by accident! 🤦  It's there now

PR #3561, opened 2026-05-15
stop zuora module depending on discount-api

johnduffell:
zuora module tests depends on discount-api

No module should depend on a handler

This is only needed for IDs of things from sandbox catalog.

This PR makes it use productCatalog instead

### DIFF
--- a/handlers/discount-api/src/productToDiscountMapping.ts
+++ b/handlers/discount-api/src/productToDiscountMapping.ts
@@ -86,13 +86,6 @@ export const catalog = {
 	},
 };
 
-export const sandboxProductRatePlanChargeIds = {
-	recurringContribution: {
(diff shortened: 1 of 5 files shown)


handlers/discount-api/src/productToDiscountMapping.ts
```diff
@@ -86,13 +86,6 @@ export const catalog = {
 	},
 };
 
-export const sandboxProductRatePlanChargeIds = {
-	recurringContribution: {
-		Month: '2c92c0f85a6b1352015a7fcf35ab397c',
-		Annual: '2c92c0f85e2d19af015e3896e84d092e',
-	},
-};
```
johnduffell, line 94: this was only used by the tests

rupertbates (approved):
Nice 👍

PR #3562, opened 2026-05-15
break cyclical dep between routing and zuora modules

johnduffell:
There is a cyclical dependency between routing and zuora modules.

The routing module depends on zuora for the `withMMAIdentityCheck` logic.  It's arguable whether that should be the case, but it seems reasonable enough for now.

zuora module (along with most other things) depend on the logger.  This seems totally expected, however they is in the routing module.

This PR splits out the logging code into a new `logger` module, this breaking the cycle.


modules/logger/src/logger.ts
```diff
@@ -1,7 +1,7 @@
 import * as console from 'node:console';
```
johnduffell, line 1: MARKER: these are the actual changes

tsconfig.json
```diff
@@ -47,84 +49,90 @@
 			],
 			"@modules/identity/*": ["./modules/identity/src/*"],
 			"@modules/routing/*": ["./modules/routing/src/*"],
+			"@modules/logger/*": ["./modules/logger/src/*"],
```
johnduffell, line 52: MARKER: this is an actual change

tsconfig.json
```diff
@@ -1,35 +1,37 @@
...
-    "module": "commonjs",                                /* Specify what module code is generated. */
-    // "rootDir": "./",                                  /* Specify the root folder within your source files. */
-    // "moduleResolution": "node10",                     /* Specify how TypeScript looks up a file from a given module specifier. */
-    // "baseUrl": "./",                                  /* Specify the base directory to resolve non-relative module names. */
-    "paths": {
+		/* Modules */
+		"module": "commonjs" /* Specify what module code is generated. */,
+		// "rootDir": "./",                                  /* Specify the root folder within your source files. */
+		// "moduleResolution": "node10",                     /* Specify how TypeScript looks up a file from a given module specifier. */
+		// "baseUrl": "./",                                  /* Specify the base directory to resolve non-relative module names. */
```
johnduffell, line 33: for whatever reason our standard is tabs and intellij has helpfully updated the whole file

buildcheck/data/build.ts
```diff
@@ -491,6 +491,13 @@ const moduleInternationalisation: ModuleDefinition = {
 	extraScripts: srcOnly,
 };
 
+const moduleLogger: ModuleDefinition = {
+	name: 'logger',
+	devDependencies: {
+		...devDeps['@smithy/types'],
+	},
+};
+
```
johnduffell, line 500: MARKER: this is the new module

note: this PR has so much boring stuff that I've flagged up the interesting changes with "MARKER" so you can find them easily

buildcheck/data/dependencies.ts
```diff
@@ -45,7 +45,7 @@ export const devDeps = separateDepRecords({
 	'@types/stripe': '^8.0.417',
 	'@types/aws-lambda': '^8.10.147',
 	'@types/jest': '^29.5.14',
-	'@smithy/types': '^4.9.8',
+	'@smithy/types': '4.12.0',
```
johnduffell, line 48: MARKER: I had to use the exact version, as it was picking 4.13 which broke the CDK type checking

shtukas (approved):

PR #3563, opened 2026-05-15
New lambda template improvements

rupertbates:
## What does this change?
The support service lambdas repo has a [Hygen template](https://github.com/jondot/hygen) powered script to make it easy to generate new lambda backed API gateway APIs.
This PR makes some improvements to those templates:
- Refactor the generated code to move the endpoint function into a separate file as this has become a bit of a convention in the repo - this endpoint now also has a request zod schema to provide a slightly more realistic starting point for an API.
- Add a README.md file to the generated files with some minimal content including CODE and PROD urls for the API - all APIs should have one of these.
- Add an optional [Open API specification](https://swagger.io/specification) document for the generated endpoint and pnpm scripts to lint this and preview it `pnpm --filter my-api openapi:preview`. This provides a starting point for adding Open API docs to our APIs where useful. Whether this doc is generated or not depends on a new y/n prompt in the initial lambda generation.
- Add some default [throttle](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-request-throttling.html) settings for generate APIs to give some protection against DDOS by default.
- Rename the command from `new-lambda` to `new-api-lambda` - this is more accurate and would enable us to provide similar commands for other types of lambda in future.
- Make the example integration test implementation a bit more realistic.


buildcheck/data/build.ts
```diff
@@ -392,10 +393,7 @@ const multipleAccountApi: HandlerDefinition = {
 		...devDeps['@redocly/cli'],
 	},
 	extraScripts: {
-		'openapi:lint': 'redocly lint openapi.yaml',
-		'openapi:preview':
-			'redocly build-docs openapi.yaml --output target/docs/index.html && open target/docs/index.html',
-		package: `pnpm type-check && pnpm lint && pnpm openapi:lint && pnpm check-formatting && pnpm test && pnpm build && cd target && zip -qr multiple-account-api.zip ./*.js.map ./*.js`,
+		...openApiScripts,
```
tjmw, line 396: Does `package` belong in `openApiScripts`? It feels like maybe that should be defined separately, what do you think?

tjmw (approved):
Nice work! 👏

buildcheck/data/build.ts
rupertbates: I think you're right, I've made that change now

PR #3567, opened 2026-05-19
add basic user-subscriptions-api handler

johnduffell:
We want to display an upsell banner on the front page of manage account overview only if a user is eligible (they have a T2 subscription, they aren't in payment failure, the sub isn't cancelled, no current discount, etc)
See https://www.figma.com/design/iqt4KtIjj6cSDcSsApLib2/MMA-Self-Serve-Upsell-from-T2-to-T3?node-id=641-2135&t=tFn6wyKH5AwIN2Ze-0

At the moment the account overview information comes from MDAPI, but MDAPI doesn't understand discounts, and is slow.

As part of future work to improce performance and maintainability of MMA, we want to move towards smaller lambdas and progressive enhancement.  See here for full details: https://docs.google.com/document/d/1KSGN7JB9obi3vDnCE1mKLgOIOky4IqAZ3Zpr3mUGElg/edit?tab=t.0#heading=h.duljcbw0juko

The initial stage is a lambda that is called from manage, with an identity id, and it returns one record per subscription owned by that user.  It will eventually return eligibility information about the upsell.

I'm not sure about the name, any opinions apprecaited - it's a "read" rather than being a CRUD type api, and should cover all the progressive enhancement of the account overview.  But opinions on that are also appreciated.

This PR is the scaffolding only, generated by the new lambda generator script with minor tweaks and self-review

Tested in CODE.


cdk/lib/user-subscriptions-api.ts
```diff
@@ -0,0 +1,26 @@
...
+			},
+			monitoring: {
+				errorImpact:
+					'users may be receiving errors or missing information on manage their account',
+			},
+			// TODO work out what this means
+			// throttle: {
+			// 	rateLimit: 20,
+			// 	burstLimit: 10,
+			// },
```
johnduffell: I wasn't sure when I would need this and whether we would get an alarm if it was exceeded

shtukas (approved):

cdk/lib/user-subscriptions-api.ts
rupertbates: I think it is good to have this by default because it will prevent us reaching our lambda concurrency limit if something goes rogue and starts spamming the endpoint. Good question about the alarm, if the rate gets exceeded we return a 429 response to the client but I don't think we'd get an alarm unless we set one up for that response code. Might be worth adding one?

johnduffell: thanks, it is a little complex setting up an alarm because it has to either respond to all 4xx errors, or be driven by access logging filters.  Maybe a task for another day.

rupertbates (approved):

PR #3571, opened 2026-05-20
Move email fields information from support-frontend into support-service-lambdas

rupertbates:
## What does this change?
This PR moves the logic related to email data fields from support-frontend to support-service-lambdas.

There is a lot of logic related to what information needs to be passed to Braze when we are triggering an acquisition (day 0) email. Previously this has lived in support-frontend but it also needs to be used by the new-subscription-api to send acquisition emails after a CSR has sold a new product.

---
  - https://app.asana.com/0/0/1214403051060430


maloky (approved):

maloky:
@rupertbates what do you think about applying the Barrel Patterns export pattern in here. That way we could have cleaner import in support-frontend and it would serve as an interface to the directory.

rupertbates:
> @rupertbates what do you think about applying the Barrel Patterns export pattern in here. That way we could have cleaner import in support-frontend and it would serve as an interface to the directory.

I'm a bit wary because we used to have that for the Zuora module but it broke automatic imports in IntellilJ (not sure about VS code) so I removed it.

PR #3580, opened 2026-05-21
use buildcheck to generate/seed new lambdas

johnduffell:
Based on pre-refactoring PR https://github.com/guardian/support-service-lambdas/pull/3590
If this PR looks reasonable, be sure to review the above one to see the noise that's missing from this one.

This is a 10% time project

We use hygen to produce new lambdas, which uses ejs templating and a prompt to write some files needed to get the lambda boilerplate up and running.

I noticed that Hygen is unmaintained and the docs actually have redirected for a long time to somewhere doubtful that is blocked on my laptop.

Since we already have buildcheck to generate and manage build files across the repo, it seemed like it would make sense to have a "one time" write mode as a subset of the existing generation and control mechanism.

## This PR
1. adds a "seed" mode to buildcheck (moving the existing templates to "managed" mode)
1. adds an `insertion` template type for where we only want to add a snippet to a file.
1. migrates the templates for the new-api-lambda from the old EJS/hygen to plain typescript
1. other fixes and updates to match the new style

The advantages are that
- the templates are just normal, compiled typescript that we write every day
- consistency with existing buildcheck templates
- we don't have to learn EJS and (lack of) tooling
- we get normal highlighting/colour coding and navigation/autocomplete
- we can share code and logic where appropriate
- no external deps other than zod

## How to review

The diff is quite noisy, but to try it out:
1. check out this branch
1. run `pnpm --filter buildcheck seed api-lambda`
1. review the changes:
image

It should produce the same content as the old `pnpm new-api-lambda` when given the same options.

Learn about seeds in general: https://github.com/guardian/support-service-lambdas/blob/jd/buildcheck/new-lambda/buildcheck/data/seeds/README.md
Learn about the new-api-lambda seed https://github.com/guardian/support-service-lambdas/blob/jd/buildcheck/new-lambda/buildcheck/data/seeds/api-lambda/README.md

## TODO (possibly all in future PRs)

- delete the old _templates and package.json script
- add some kind of repeatable tests?
- a better way of generating .ts files than string concatenation!


buildcheck/data/seeds/api-lambda/templates/.github/workflows/ci-typescript.yml.inserts.ts
```diff
@@ -0,0 +1,9 @@
+import type { InsertChunks } from '../../../../../types';
```
johnduffell: these files are the new api-lambda seed files, they create/insert files in the repo based on their path, e.g. this one
`buildcheck/data/seeds/api-lambda/templates/.github/workflows/ci-typescript.yml.inserts.ts`
inserts chunks into this file
`REPO_ROOT/.github/workflows/ci-typescript.yml`

buildcheck/data/seeds/api-lambda/index.ts
```diff
@@ -0,0 +1,46 @@
...
+	),
+	includeApiKey: withPrompt(
+		booleanFlag(true),
+		'Should I generate an API key for this lambda?',
+	),
+	includeOpenApiDoc: withPrompt(
+		booleanFlag(true),
+		'Should I generate an Open API description for this lambda?',
+	),
+});
```
johnduffell, line 23: this defines the parameters that the seed will prompt for (or you can provide on the command line)

buildcheck/src/seed/args/getSeedEntry.ts
```diff
@@ -0,0 +1,15 @@
+import seedsDirConfig from '../../../data/seeds/_generated_dirIndex';
```
johnduffell, line 1: the files in src/seed/args/ are concerned with parsing whatever you pass in on the command line into a valid entry in the seeds data, including prompting for answers.  There are libraries for prompting and parsing, but I wasn't sure it's worth another (potentially unsupported) dependency for something relatively simple.

buildcheck/src/seed/args/recordFromFlags.ts
```diff
@@ -0,0 +1,18 @@
+/**
```
johnduffell, line 1: as an convenience for testing you can pass in the answers as command line flags.  This facility could be removed to simplify things as end users won't really use it.

buildcheck/src/dynamic/generate-template-list.sh
```diff
@@ -61,12 +65,56 @@ write_ts_index() {
...
+# generate TS file indexes for both dirs
+for parent_dir in "$data_dir/seeds" "$data_dir/managed"; do
+    while IFS= read -r -d '' subdir; do
+        [[ "$(basename "$subdir")" == _* ]] && continue
+        write_ts_index "$subdir"
+    done < <(find "$parent_dir" -mindepth 1 -maxdepth 1 -type d -print0 | sort -z)
+done
+
+# seeds index only needed for seeds
+write_dir_index "$data_dir/seeds"
```
johnduffell, line 118:
```typescript
// buildcheck/data/seeds/_generated_dirIndex.ts
// [imports etc....]

export default {
	'api-lambda': { index: api_lambda_index, templates: api_lambda_templates } satisfies DirConfig<api_lambda_TemplateParams>,
};
```

buildcheck/data/seeds/api-lambda/templates/.github/workflows/ci-typescript.yml.inserts.ts
rupertbates: I wonder whether we could add an @buildCheck/types style path mapping to reduce the length of all these relative imports, in the same way we do with @modules

buildcheck/data/seeds/api-lambda/templates/cdk/lib/_lambdaName_.ts.ts
```diff
@@ -0,0 +1,37 @@
+import { toPascalCase } from '../../../../../snippets/string';
+import type { MaybeTemplateContent } from '../../../../../types';
+import type { TemplateParams } from '../../../index';
+
+export default ({
+	lambdaName,
+	includeApiKey,
+}: TemplateParams): MaybeTemplateContent => `import type { App } from 'aws-cdk-lib';
```
rupertbates, line 8: It would definitely be nice to find a way to do this other than strings

buildcheck/data/seeds/api-lambda/templates/.github/workflows/ci-typescript.yml.inserts.ts
johnduffell: I see what you mean it's impossible to know at a glance where it ends up.  All the imports are from `buildcheck/data/....` so they kind of feel more relative rather than how modules are getting things from elsewhere
Let me have a play and a read up on use cases for the mappings.

buildcheck/data/seeds/api-lambda/templates/cdk/lib/_lambdaName_.ts.ts
johnduffell: yes I did research into it a bit and there was no obvious good option
- the "typescript compiler api" - is horrifically verbose
- [ts-morph](https://github.com/dsherret/ts-morph/blob/latest/packages/ts-morph) - better but still too verbose
- save the file as normal TS with placeholders and then use ts-morph or similar to just add or replace things - Probably the best of the above, but would involve two files, one for the basic file and one for the transforms.

buildcheck/data/seeds/api-lambda/templates/.github/workflows/ci-typescript.yml.inserts.ts
johnduffell: I've added
```
		"paths": {
			"@buildcheck/types": ["data/types.ts"],
			"@buildcheck/snippets/*": ["data/snippets/*"],
			"@buildcheck/seeds/*": ["data/seeds/*"],
			"@buildcheck/managed/*": ["data/managed/*"]
		},
```

in these commits
https://github.com/guardian/support-service-lambdas/pull/3580/changes/0ecabba037948b0645b823c9b007533adf26f071..4ff4cf2a6f1d717aaef17144d33ee38a58dd66fb

What do you think?  The PR is ready for another review now.

shtukas (approved):

PR #3588, opened 2026-05-29
Multiple accounts: Add new delete endpoint

maloky:
## What does this change?
1. Added a new delete invitation endpoint implementation in [deleteInvitationEndpoint.ts](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html).
2. Registered a new DELETE route for /invitation/{invitationCode} in [index.ts](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html).
3. Added an integration test that seeds DynamoDB, calls the endpoint, and verifies the invitation is deleted in [deleteInvitationIntegration.test.ts](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html).

## API Examples
DELETE /invitation/{invitationCode}

### Request
```
DELETE /invitation/RpwR62kMnAxe HTTP/1.1
Host: multiple-account-api-code.support.guardianapis.com
x-api-key: <your-api-key>
```

### Success Response
```
HTTP/1.1 204 No Content
Content-Type: application/json
```
### Not Found Response (404)
```
HTTP/1.1 404 Not Found
Content-Type: text/plain

Not found
```
### Error Response (500)
```
HTTP/1.1 500 Internal Server Error
Content-Type: text/plain

Internal server error
```

Deployed and tested on CODE with a invitation created for the subscriptionName `A-S00974337`.

### DIFF
--- a/handlers/multiple-account-api/src/deleteInvitationEndpoint.ts
+++ b/handlers/multiple-account-api/src/deleteInvitationEndpoint.ts
@@ -19,29 +19,24 @@ export const deleteInvitationEndpoint =
 		const { invitationCode } = path;
 		logger.mutableAddContext(invitationCode);
 
-		const invitation =
-			await invitationRepository.getByInvitationCode(invitationCode);
+		try {
+			const invitation = await invitationRepository.get(invitationCode);
 
-		if (invitation.length > 1) {
-			logger.error(
-				`Multiple invitations found for invitation code ${invitationCode}`,
-				{ invitationCount: invitation.length },
+			if (!invitation) {
+				return notFound();
+			}
+
+			await invitationRepository.delete(
+				invitation.subscriptionName,
+				invitationCode,
 			);
+
+			return {
+				body: '',
+				statusCode: 204,
+			};
+		} catch (error) {
+			logger.error('Error deleting invitation', error);
 			return internalServerError();
 		}
-
-		const [invitationToDelete] = invitation;
-
-		if (!invitationToDelete) {
-			return notFound();
-		}
-
-		const { subscriptionName } = invitationToDelete;
-
-		await invitationRepository.delete(subscriptionName, invitationCode);
-
-		return {
-			body: '',
-			statusCode: 204,
-		};
 	};
--- a/handlers/multiple-account-api/src/deleteInvitationEndpoint.ts
+++ b/handlers/multiple-account-api/src/deleteInvitationEndpoint.ts
@@ -1,0 +1,25 @@
+import { logger } from '@modules/logger/logger';
+import type { APIGatewayProxyResult } from 'aws-lambda';
+import { z } from 'zod';
+import { type InvitationRepository } from './invitationRepository';
+
+export const deleteInvitationBodySchema = z.object({
+	invitationCode: z.string(),
+	subscriptionName: z.string(),
+});
+
+export type DeleteInvitationBody = z.infer<typeof deleteInvitationBodySchema>;
+
+export const deleteInvitationEndpoint =
+	(invitationRepository: InvitationRepository) =>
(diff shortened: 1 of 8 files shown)


handlers/multiple-account-api/src/deleteInvitationEndpoint.ts
```diff
@@ -0,0 +1,25 @@
...
+export const deleteInvitationEndpoint =
+	(invitationRepository: InvitationRepository) =>
+	async (body: DeleteInvitationBody): Promise<APIGatewayProxyResult> => {
+		const { subscriptionName, invitationCode } = body;
+		logger.mutableAddContext(subscriptionName);
+
+		await invitationRepository.delete(subscriptionName, invitationCode);
+
+		return {
+			body: 'DELETED',
```
tjmw: According to https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/204 a 204 response should not contain any content, so I think we should remove the body (the status code conveys the outcome).

maloky: yeah, I think we might have to make the body optional in the existing Response type 👍

maloky: contrary to what I was think we can't drop the body as it's a mandatory field from `APIGatewayProxyResult` interface. should we pass an empty string? wdyt?
Screenshot 2026-06-02 at 12 19 25

tjmw: Oh right, that's surprising! Yeah empty string sounds good to me 👍

handlers/multiple-account-api/src/deleteInvitationEndpoint.ts
```diff
@@ -0,0 +1,47 @@
...
+export const deleteInvitationEndpoint =
+	(invitationRepository: InvitationRepository) =>
+	async (path: DeleteInvitationPath): Promise<APIGatewayProxyResult> => {
+		const { invitationCode } = path;
+		logger.mutableAddContext(invitationCode);
+
+		const invitation =
+			await invitationRepository.getByInvitationCode(invitationCode);
+
+		if (invitation.length > 1) {
```
rupertbates: I think I would move this check to the repository as it is business logic we want implemented in all cases. Do the check there and throw an exception if there is > 1 record

modules/routing/src/apiGatewayResponses.ts
```diff
@@ -11,6 +11,13 @@ export function badRequest(message: string): APIGatewayProxyResult {
 	};
 }
 
+export function notFound(): APIGatewayProxyResult {
```
rupertbates, line 14: 👍

handlers/multiple-account-api/src/invitationRepository.ts
```diff
@@ -61,6 +61,24 @@ export class InvitationRepository {
 		return invitationRecordSchema.parse(unmarshall(result.Item));
 	}
 
+	async getByInvitationCode(
```
rupertbates: I think it might be worth getting rid of the other `get` function and calling this one `get` as the other one isn't used. What do you think?

rupertbates (approved):
A couple of suggestions but looking really good 👍

handlers/multiple-account-api/src/invitationRepository.ts
maloky: if this is the default get behaviour I think so. I didn't want to bring much business logic into the CRUD functions as they should be as generic as possible for reusability but I think in here it's what we want

PR #3590, opened 2026-05-29
refactor to make space for new-lambda style seeds

johnduffell:
pre refactoring for https://github.com/guardian/support-service-lambdas/pull/3580

This PR makes contains all the "noise" from the above PR, to reduce the size of the above diff.  There shouldn't be any functionality differences.

It doesn't really add any value on its own so shouldn't be merged unless the above is also approved.

See inline comments for more details.


buildcheck/data/managed/handler/templates/jest.config.js.ts
```diff
@@ -0,0 +1,3 @@
+import { buildJestConfigJs } from '../../../snippets/buildJestConfigJs';
```
johnduffell, line 1: these files are the existing `buildcheck/data/templates/handler/*`
they have been moved down a level so `seeds` can be placed alongside in future

buildcheck/data/snippets/BUILDCHECK.md.ts
```diff
@@ -8,8 +8,9 @@ export function generateWarningFile(
 	pathToRoot: string,
 ): GeneratedFile {
 	const fileList = [...generatedFileNames, warningFileName];
+	const templateFilename = relativePath(__filename);
```
johnduffell, line 11: in general, I have switched the engine to make pathnames relative to the repo, and then prefixing with repoRoot only when needed. This makes them easier to deal with in files/messages/tests.

buildcheck/src/dynamic/generate-template-list.sh
```diff
@@ -1,70 +1,76 @@
 #!/bin/bash -e
```
johnduffell, line 1: To reduce complication here I've made this script a more simple "scan directories and write standard records for every .ts file found". To get type safety, it assumes the types are beside the generated file, in ./types.ts.

Obviously the only variation at this stage is HandlerDefinition vs ModuleDefinition so it seems overkill, but it makes more sense when seed specific types get passed to the various custom prompts.

Copilot was keen that this script too should be written in TS, but I'm reluctant to go too meta, and the script isn't too hard to follow in my opinion.

buildcheck/src/util/file-writer.ts
```diff
@@ -13,20 +14,6 @@ export function writeFiles(rootPath: string, files: GeneratedFile[]): void {
...
-	const resolvedTarget = path.resolve(fullPath);
-
-	if (!resolvedTarget.startsWith(resolvedBase + path.sep)) {
-		throw new Error(
-			`Path traversal detected: ${relativePath} escapes base directory ${basePath}`,
-		);
-	}
-
-	return fullPath;
-}
```
johnduffell, line 28: moved to its own file

handlers/alarms-handler/jest.config.js
```diff
@@ -1,4 +1,4 @@
-// MANAGED FILE: to push changes see buildcheck/README.md - template: buildcheck/data/templates/handler/jest.config.js.ts
+// MANAGED FILE: to push changes see buildcheck/README.md - template: buildcheck/data/managed/handler/templates/jest.config.js.ts
```
johnduffell, line 1: all the changes from here downwards are just reflecting the new location of the managed templates

buildcheck/src/dynamic/generate-template-list.sh
johnduffell: before - the script knew which imports to use to each folder

```typescript
// buildcheck/src/dynamic/generated/generatedMappings.ts
import type { HandlerDefinition, ModuleDefinition } from '../../../data/build';
import { type Template } from '../templater';
[....]

export const handlerTemplates: Array<Template<HandlerDefinition>> = [
  { targetPath: "jest.config.js", value: handler_jest_config_js, templateFilename: "buildcheck/data/templates/handler/jest.config.js.ts" },
  { targetPath: "package.json", value: handler_package_json, templateFilename: "buildcheck/data/templates/handler/package.json.ts" },
  { targetPath: "riff-raff.yaml", value: handler_riff_raff_yaml, templateFilename: "buildcheck/data/templates/handler/riff-raff.yaml.ts" },
  { targetPath: "tsconfig.json", value: handler_tsconfig_json, templateFilename: "buildcheck/data/templates/handler/tsconfig.json.ts" },
];

export const moduleTemplates: Array<Template<ModuleDefinition>> = [
  { targetPath: "jest.config.js", value: module_jest_config_js, templateFilename: "buildcheck/data/templates/module/jest.config.js.ts" },
  { targetPath: "package.json", value: module_package_json, templateFilename: "buildcheck/data/templates/module/package.json.ts" },
  { targetPath: "tsconfig.json", value: module_tsconfig_json, templateFilename: "buildcheck/data/templates/module/tsconfig.json.ts" },
];
```

johnduffell: after - the script just references relative imports from the folder's types.ts/index.ts etc rather than hard coding ModuleDefinition/HandlerDefinition.
You can also see it only stores relativeName, and a single copy of templateDir, rather than processing it in the shell script.

```typescript
// both files identical looking relative imports from different dirs
import type { TemplateEntry } from '../types';
import type { TemplateParams } from './index';
```
in handler:
```typescript
// buildcheck/data/managed/handler/_generated_tsIndex.ts
export default {
  templates: [
    { relativeName: 'jest.config.js.ts', value: jest_config_js },
    { relativeName: 'package.json.ts', value: package_json },
    { relativeName: 'riff-raff.yaml.ts', value: riff_raff_yaml },
    { relativeName: 'tsconfig.json.ts', value: tsconfig_json },
  ] satisfies Array<TemplateEntry<TemplateParams>>,
  templateDir: 'buildcheck/data/managed/handler/templates',
}
```
in modules:
```typescript
// buildcheck/data/managed/module/_generated_tsIndex.ts
export default {
  templates: [
    { relativeName: 'jest.config.js.ts', value: jest_config_js },
    { relativeName: 'package.json.ts', value: package_json },
    { relativeName: 'tsconfig.json.ts', value: tsconfig_json },
  ] satisfies Array<TemplateEntry<TemplateParams>>,
  templateDir: 'buildcheck/data/managed/module/templates',
}
```

buildcheck/src/util/argsParser.ts
```diff
@@ -9,39 +9,30 @@ export function parseArguments(argv: string[]): ParsedArgs {
...
-			'Mode (--generate or --clean) must be provided as first argument\n' +
-				syntax,
-		);
-	}
-
-	const mode = args[0];
-	if (mode !== '--generate' && mode !== '--clean') {
-		throw new Error('First argument must be --generate or --clean\n' + syntax);
-	}
+		'Syntax: cli.ts <REPO_ROOT> --generate|--clean\nActual: ' + argv.join(' ');
```
johnduffell, line 12: REPO root is moved to the start of the parameter list to make it easier to process variable parameter lists

buildcheck/src/util/argsParser.ts
```diff
@@ -9,39 +9,30 @@ export function parseArguments(argv: string[]): ParsedArgs {
...
+
+	const dashes = args[1].substring(0, 2);
+	const mode = args[1].substring(2);
+	if (dashes !== '--' || (mode !== 'generate' && mode !== 'clean')) {
+		throw new Error('Second argument must be --generate or --clean\n' + syntax);
 	}
 
-	return {
-		// eslint-disable-next-line @typescript-eslint/consistent-type-assertions -- todo use zod above
-		mode: mode.substring(2) as 'generate' | 'clean',
```
johnduffell, line 44: fixed `as` - now we process the `--` and `generate` part separately to keep the type checker on side

buildcheck/data/managed/module/index.ts
```diff
@@ -0,0 +1,2 @@
+// this is so the generated index knows what parameter the templates should take
```
rupertbates: I don't really understand this comment

buildcheck/data/managed/types.ts
```diff
@@ -0,0 +1,4 @@
+import type { FileTemplate } from '../types';
+
+// this ensures the generated file does not allow "insert" templates, only fully managed files
```
rupertbates: Or this one!

buildcheck/src/dynamic/generate-template-list.sh
rupertbates: I can definitely see the advantage of using TS to do this but happy to go with shell for now if you think it's better

rupertbates (approved):
I can't claim to really understand all of this but it looks ok to me!

buildcheck/data/managed/module/index.ts
johnduffell: good point, I've rewritten and added all the  background info

buildcheck/data/managed/types.ts
johnduffell: as above (and I also did the third instance of the similar comment)

johnduffell:
> I can't claim to really understand all of this but it looks ok to me!

That sounds like a vote against adding seed templating to buildcheck then - the idea of reducing down to one tool with one template format should be to make things easier to think about and maintain rather than making it not understandable.

PR #3597, opened 2026-06-03
Fix new lambda template

rupertbates:
## What does this change?

#3552 required all projects in this repo to declare their dependencies on other module projects using a `moduleDependencies` setting in the build.ts definition, to help with build time. 

However the template which generates new api lambdas was not updated to include this new setting so any lambdas created with the template will fail until it is added. This PR updates the template to ensure that it is there on creation.


paul-daniel-dempsey (approved):
LG

PR #3600, opened 2026-06-03
update it-test command 

maloky:
## What does this change?
update the build package.json generator `it-test` command to include a `--experimental-vm-modules` node flag. 

## Why
If `--experimental-vm-modules` flag is not present, tests that use dynamic imports will fail.
Dynamic import of modules fails if flag is not passed because the node api jest uses to enable ESM support is still not stable as of node 18.x.
- https://nodejs.org/api/vm.html#when-importmoduledynamically-is-a-function
- https://jestjs.io/docs/ecmascript-modules

### DIFF
--- a/buildcheck/data/snippets/buildPackageJson.ts
+++ b/buildcheck/data/snippets/buildPackageJson.ts
@@ -30,7 +30,9 @@ export function buildPackageJson(
 		name: `${pkg.name}`,
 		scripts: {
 			test: 'jest --group=-integration',
-			'it-test': 'jest --group=integration',
+			'it-test':
+				'NODE_OPTIONS="$NODE_OPTIONS --experimental-vm-modules" jest --group=integration',
+
 			'type-check': 'tsc --noEmit',
 			lint: "eslint --cache --cache-location /tmp/eslintcache/ 'src/**/*.ts' 'test/**/*.ts'",
 			'check-formatting': 'prettier --check "**/*.ts"',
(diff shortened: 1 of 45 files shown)


johnduffell:
thanks for the PR, please can you add something to the PR description to explain the "why" we need this and what happens if this isn't set? Maybe this is something we can remove once we bump the node version, or are we going to need it for ever?
Regarding the PR label I think this is "maintenance" rather than "fix" 
https://docs.google.com/document/d/1WX5_fn8ZEL6ifeuI02mqokvPz2kL28ZEbJtXpaetL5c/edit?tab=t.0#heading=h.7vsjnr32lv9y

maloky:
> thanks for the PR, please can you add something to the PR description to explain the "why" we need this and what happens if this isn't set? Maybe this is something we can remove once we bump the node version, or are we going to need it for ever? Regarding the PR label I think this is "maintenance" rather than "fix" https://docs.google.com/document/d/1WX5_fn8ZEL6ifeuI02mqokvPz2kL28ZEbJtXpaetL5c/edit?tab=t.0#heading=h.7vsjnr32lv9y

Updated @johnduffell. Do you think is clear enough? We don't seem to have a node version that fix this in sight.

johnduffell (approved):
Thanks @maloky yes I was unsure because I've previously had to add this locally and had the feeling that I shouldn't be doing it, but after chatting it through with @tjmw we think it's because jest interprets the code directly, rather than compiling down to a bundle, and perhaps something is doing a dynamic import e.g. in the AWS library.

PR #3601, opened 2026-06-04
Fix new lambda template - add required routing

paul-daniel-dempsey:
## What does this change?
https://github.com/guardian/support-service-lambdas/pull/3552 required all projects in this repo to declare their dependencies on other module projects using a moduleDependencies setting in the build.ts definition, to help with build time.

However the template which generates new api lambdas was not updated to include routing within this new setting so any lambdas created with the template fail the typescript compiler until it is added. 

This PR updates the template to ensure that routing is added on creation.


rupertbates (approved):
👍

PR #3606, opened 2026-06-05
Sales Tax Api : endpoint setup (Canadian)

paul-daniel-dempsey:
## What does this change?
Sets up a new API which will return tax rate percentages for a given country+productKey called `tax-rates`.

Initial Setup mimicks what we would retreive from Zuora eventually, ONLY Canada allowed.

Return is in format of type `CaState`

POST /tax-rates
Request : {productKey: productKeySchema, country: isoCountrySchema}
Response : {CaState}

POST : https://sales-tax-api-code.support.guardianapis.com/tax-rate
BODY : {
  "productKey": "SupportPlus",
  "country: "CA",
}

Note: 
This api authenticates using an `x-api-key` within the header and requires generating/supplying. 
The token can be found under AWS API GateWay/API Keys (sales-tax-api-CODE/PROD).

Found: Response 200
{
  "AB": 0.05,
  "BC": 0.12,
  "MB": 0.12,
  "NB": 0.15,
  "NL": 0.15,
  "NT": 0.15,
  "NS": 0.15,
  "NU": 0.05,
  "ON": 0.13,
  "PE": 0.15,
  "QC": 0.1498,
  "SK": 0.11,
  "YT": 0.05
}
Not Found: Response 404
{
}

---
  - https://app.asana.com/0/0/1215133618429652
  - https://app.asana.com/0/0/1215133618429653


handlers/sales-tax-api/src/salesTaxEndpoint.ts
```diff
@@ -0,0 +1,73 @@
+import { ValidationError } from '@modules/errors';
+import type { IsoCountry } from '@modules/internationalisation/country';
+import { logger } from '@modules/logger/logger';
+import { buildErrorResponse, ok } from '@modules/routing/apiGatewayResponses';
+import type { APIGatewayProxyResult } from 'aws-lambda';
+import type { SalesTaxRequest, SalesTaxResponse } from './schemas';
+import { salesTaxResponseSchema } from './schemas';
+
+export const countryStates: Partial<
+	Record<IsoCountry, Record<string, number>>
```
tjmw: I can't remember - when we were looking was there a type for the CA state codes we could use here (instead of `string`)?

handlers/sales-tax-api/src/salesTaxEndpoint.ts
```diff
@@ -0,0 +1,73 @@
...
+			country,
+			state,
+		});
+		return ok(
+			getSalesTaxRate({ productKey, country, state }),
+			salesTaxResponseSchema,
+		);
+	} catch (error) {
+		logger.error('Error fetching sales tax', error);
+		return Promise.resolve(buildErrorResponse(error));
```
tjmw: I don't think the `Promise.resolve` is needed here. Because the function is async the return value will always be wrapped in a promise.

```suggestion
		return buildErrorResponse(error);
```

handlers/sales-tax-api/src/salesTaxEndpoint.ts
```diff
@@ -0,0 +1,73 @@
...
+	if (!validProductKey) {
+		throw new ValidationError(`invalid productKey:${productKey}`);
+	}
+	return { taxRate: getLocationSalesTax(country, state) };
+}
+
+function getLocationSalesTax(country: IsoCountry, state: string) {
+	const salesTaxRate = countryStates[country]?.[state];
+	if (!salesTaxRate) {
+		const message = ['CA'].includes(country)
```
tjmw: Just a stylistic thing but I think I would pull this country check up to the top of this function and throw a `ValidationError` early if it's not in the list. Then below, if the tax rate can't be found for the state provided a different `ValidationError` would be thrown. I think that'd make it slightly more readable than rolling it all into a ternary.

handlers/sales-tax-api/test/index.test.ts
```diff
@@ -0,0 +1,108 @@
...
+				...baseEvent,
+				body: JSON.stringify({
+					state: 'XX',
+				}),
+			} as unknown as APIGatewayProxyEvent;
+
+			const response = await handler(requestEvent);
+			expect(response.statusCode).toEqual(400);
+		});
+		it('returns 404 for wrong HTTP method', async () => {
```
tjmw: I think _probably_ we don't need to test this here as it's the router's responsibility to handle this (which is tested elsewhere).

tjmw (approved):
Great work on this! I left a few minor comments/suggestions - do shout if any of them don't make sense.

handlers/sales-tax-api/src/salesTaxEndpoint.ts
paul-daniel-dempsey: Yep added `CAState` here

handlers/sales-tax-api/src/salesTaxEndpoint.ts
paul-daniel-dempsey: I think its needed otherwise I get an error saying I have not used `await` inside the `async ` salesTaxRequestEndpoint function

tjmw: Oh right - in that case maybe just remove the `async` for now (and the `Promise.resolve`) since there's no actual async/promise based code in here I don't think?

paul-daniel-dempsey: I think it needs to be called in an async from the index.ts handler Router above it however?

handlers/sales-tax-api/src/salesTaxEndpoint.ts
paul-daniel-dempsey: Ok, thought it looked neater this way. As you say, guess its more readable the way you suggest.

handlers/sales-tax-api/src/salesTaxEndpoint.ts
tjmw: Yeah, I think the piece that actually expects a Promise/async function is the second arg to `withBodyParser` in the Router. So perhaps we could wrap the return value there in `Promise.resolve` and remove all async/promise related code from this function for now.

paul-daniel-dempsey:
> Great work on this! I left a few minor comments/suggestions - do shout if any of them don't make sense.

@tjmw : I think we have changed this endpoint quite fundamentally, have added a further `tax-rateS` call for the moment to keep this seperate (will delete old one when happy). Am not 100% sure of the exact requirements now ->

1. Do we reduce this endpoint so its CAD specific only - if so, should CAD as a country be a parameter?
2. Do we return an object of cadState's only? Or something more country agnostic like an array of states (bearing in mind state names can overlap)?

At the moment am taking simple approach with Product/Country and return a CADState object. 

Would you mind taking a look and check if its what your expecting as Its very CAD specific but still checking the country?

tjmw:
>     1. Do we reduce this endpoint so its CAD specific only - if so, should CAD as a country be a parameter?

I would say continue to pass country as a param but anything other than `CAD` would result in a 400 (which I think is the behaviour for the existing endpoint). So the only difference in interface between this and the existing work is dropping the state code as a param.

>     2. Do we return an object of cadState's only? Or something more country agnostic like an array of states (bearing in mind state names can overlap)?

Yeah I think return an object with CAD state codes as keys and tax rates as values.

PR #3608, opened 2026-06-05
use buildcheck to manage the GHA build matrix

johnduffell:
At the moment, when you generate a new lambda, an entry has to be present in the GHA for each one.

Buildcheck already has a list of all the TS lambdas that exist but it has no way of checking that the GHA list is up to date.

This PR
- moves the list of subprojects to a JSON file
- adds a small new feature to buildcheck to manage files outside of a `handlers/*/` or `modules/*/` subfolder
- adds a template for the new JSON file so that buildcheck keeps the list in sync.

This reduces the amount of generated and developer maintained code, and makes the GHA shorter and only change when there are genuine changes in functionality.


buildcheck/data/templates/root/.github/workflows/handlerNames.managed.json.ts
```diff
@@ -0,0 +1,10 @@
+import type { BuildDefinition } from '../../../../build';
+import { notice } from '../../../../snippets/notices';
+
+// not in buildcheck but we do want to build it
+const cdkOnlyProject = 'salesforce-event-bus';
```
johnduffell, line 5: this one is a bit special https://github.com/guardian/support-service-lambdas/tree/main/handlers/salesforce-event-bus

tomrf1 (approved):

PR #3609, opened 2026-06-05
each handler check modules in parallel

johnduffell:
At the moment the build time is slow, could be 4-5 minutes for a handler.

We run each of the lint/type-check etc one after the other on each dependency.

This changes it to lint (etc) all modules and handler at the same time.


johnduffell:
first commit did help, but the later ones didn't.  Perhaps we're better to experiment with different runners instead

PR #3610, opened 2026-06-05
add validated self serve openapi, "try me" ui and local browser serving

johnduffell:
This PR gives the lambda (just update-supporter-plus-amount for now, but it can be rolled out to more or most in due course)
- type safe openapi docs generation
- local running of lambdas for testing
- clickable "try me" ui locally and in CODE

To do this it
- replaces the `Router` array with a more comprehensive type representing both the request and response types.
- adds extra dependency on hono
- adds pnpm script and boilerplate for running a lambda locally
- adds the scalar openapi library and route (although swagger is a more traditional option)

TODOs It's not polished, at the moment it's at "working POC" level.  I would expect a bit more work to reduce the amount of boilerplate and organise things in a clearer way, as well as usual naming improvements.  It may benefit from extracting into multiple PRs.

how to test it out:

Running locally
- check out branch
- cd handlers/update-supporter-plus-amount
- pnpm local
ui should appear
image

You can call the local endpoints using the "Test request" button on the right assuming you have suitable janus credentials, or use curl/postman against localhost:8787.

Testing on CODE
- visit https://update-supporter-plus-amount-code.support.guardianapis.com/docs
- if it doesn't work, check update-supporter-plus-amount is still deployed to CODE from this branch

Issues/limitations

Your browser won't be able to hit CODE (or PROD!) endpoint due to api key header requirement
This could be resolved with google auth via cognito [1] , but that opens a whole new vector for attacks - this would be ok for CODE but we would need to consider it more for PROD.

[1] see previous 10% time PR https://github.com/guardian/support-service-lambdas/pull/3141


cdk/lib/update-supporter-plus-amount.ts
```diff
@@ -30,5 +30,8 @@ export class UpdateSupporterPlusAmount extends SrStack {
 			new AllowZuoraOAuthSecretsPolicy(this),
 			AllowSqsSendPolicy.create(this, 'braze-emails', 'supporter-product-data'),
 		);
+
+		lambda.addPublicPath('openapi.json');
+		lambda.addPublicPath('docs');
```
johnduffell, line 35: https://update-supporter-plus-amount-code.support.guardianapis.com/docs

cdk/lib/update-supporter-plus-amount.ts
```diff
@@ -30,5 +30,8 @@ export class UpdateSupporterPlusAmount extends SrStack {
 			new AllowZuoraOAuthSecretsPolicy(this),
 			AllowSqsSendPolicy.create(this, 'braze-emails', 'supporter-product-data'),
 		);
+
+		lambda.addPublicPath('openapi.json');
```
johnduffell, line 34: assuming the branch is on CODE, visit
https://update-supporter-plus-amount-code.support.guardianapis.com/openapi.json

handlers/update-supporter-plus-amount/runManual/local.ts
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env -S STAGE=CODE pnpm exec tsx
```
johnduffell, line 1: I did have this as a `pnpm local` script but it added clutter and it actually seems nicer (with intellij at least) to run it using the green triangle.  It woudn't be a problem to add it back as a pnpm script though.

PR #3587, opened 2026-05-28
Remove aws-actions/configure-aws-credentials usage

aug24:
Removes workflow steps using aws-actions/configure-aws-credentials.

PR #3613, opened 2026-06-08
make sure to run tests when not using riffRaffUpload

johnduffell:
the old riffRaffUpload plugin used to automatically run tests before building and uploading the jar.

When we moved to the modern GHA way to do the upload in https://github.com/guardian/support-service-lambdas/pull/3587 , we forgot to explicitly run the tests

This PR adds the test task back in so they run again.

I also cleaned up some env vars that aren't needed any more.

### DIFF
--- a/.github/workflows/ci-scala.yml
+++ b/.github/workflows/ci-scala.yml
@@ -117,11 +117,8 @@ jobs:
       - name: Setup SBT
         uses: sbt/setup-sbt@v1.2.0
 
-      - name: Build
-        run: |
-          LAST_TEAMCITY_BUILD=6410
-          export GITHUB_RUN_NUMBER=$(( $GITHUB_RUN_NUMBER + $LAST_TEAMCITY_BUILD ))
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-scala.yml
```diff
@@ -117,11 +117,8 @@ jobs:
       - name: Setup SBT
         uses: sbt/setup-sbt@v1.2.0
 
-      - name: Build
-        run: |
-          LAST_TEAMCITY_BUILD=6410
-          export GITHUB_RUN_NUMBER=$(( $GITHUB_RUN_NUMBER + $LAST_TEAMCITY_BUILD ))
-          sbt "project ${{ matrix.subproject }}" assembly
+      - name: run tests and Build jar
+        run: sbt "project ${{ matrix.subproject }}" test assembly
```
johnduffell, line 121: the important bit is adding `test` here

shtukas (approved):

PR #3614, opened 2026-06-09
Add tax exclusive rate plans to the product-catalog-api schema

tjmw:
## What does this change?

Adds new tax exclusive rate plans to the product catalog API schema with charges in GBP and CAD. 

Deployed generateProductCatalog to CODE.


modules/product-catalog/src/productCatalogSchema.ts
```diff
@@ -127,6 +139,18 @@ export const productCatalogSchema = z.object({
...
 			}),
+			MonthlyTaxExclusive: z.object({
+				billingPeriod: z.literal('Month'),
+				charges: z.object({
+					Subscription: z.object({
+						id: z.string(),
+					}),
+				}),
+				id: z.string(),
+				pricing: z.object({ CAD: z.number(), GBP: z.number() }),
```
paul-daniel-dempsey, line 150: Thought it was just `CAD` to begin with? Is there an accountancy rason to keep `GBP`?

modules/product-catalog/src/productCatalogSchema.ts
paul-daniel-dempsey, line 72: Possibly we need `AnnualTaxExclusive` + `MonthlyTaxExclusive` sections under Contributions  to match Zuora?

modules/product-catalog/src/productCatalogSchema.ts
tjmw: Yeah it's required to add GBP pricing in Zuora as it's the base currency. However it does raise the question that perhaps we should be filtering it out in the product catalog as it won't be used.

modules/product-catalog/src/productCatalogSchema.ts
tjmw: I don't think we need those rate plans for contributions. Are you maybe thinking of the contribution charges for supporter plus?

rupertbates (approved):
Looks good to me 👍

PR #3586, opened 2026-05-28
Create accept invitation endpoint

rupertbates:
## What does this change?
This PR is part of the API work for the multiple accounts feature. It is described in more detail [in this doc](https://docs.google.com/document/d/1eqMn9zkg-iRK3Gi9ecmgGh8WWUYpLX6B3cOasTRYpns/edit?tab=t.0)

## Details
This adds one new endpoint which allows a user to accept an invitation to become a secondary user
`POST /invitation/[invitationCode]/accept`

Validation is done via an Okta token in the `Authorization` header - this is the standard way to identify signed in users.

The endpoint validates that the invitation exists and was sent to the currently signed in user

On successful acceptance, the endpoint creates a new record in the `multiple-account-secondary-user-[STAGE]` users Dynamo table, a new secondary subscription record in the `SupporterProductData-[STAGE]` Dynamo table, and then deletes the invitation record from the `multiple-account-invitation-[STAGE]` Dynamo table.

## Testing
Testing this endpoint is a little more complicated than the create invitation endpoint because it requires an Okta Authorization header. 

1. First you can create an invitation either via Postman/Hopscotch or using the following cURL command:
```shell
curl --location 'https://multiple-account-api-code.support.guardianapis.com/invitation' \
--header 'x-api-key: [GET_API_KEY_FROM_AWS]' \
--header 'x-identity-id: 112809589' \
--header 'Content-Type: application/json' \
--header 'Authorization: ••••••' \
--data-raw '{
  "subscriptionName": "A-S00974337",
  "secondaryUserEmail": "your.email@guardian.co.uk"
}'
```
This will return an invitation code you can use in your acceptance request:
```json
{
    "invitationCode": "8hpLZeTwHZj5"
}
```
2. You will then need to get an Okta authorisation header for your accept request. Follow the steps in [this README](https://github.com/guardian/identity/tree/main/docs/generating_tokens_locally) to generate a header in Postman or Hopscotch. 
3. Then you can finally make your accept request. In cURL this would look like:
```shell
curl --location --request POST 'https://multiple-account-api-code.support.guardianapis.com/invitation/8hpLZeTwHZj5/accept' \
--header 'x-api-key: [GET_API_KEY_FROM_AWS]' \
--header 'Authorization: Bearer [MY_GENERATED_OKTA_TOKEN]'
```

PR #3615, opened 2026-06-09
Delete secondary user endpoint

maloky:
## What does this change?

Adds support in multiple-account-api for revoking a secondary user from a subscription.

This introduces a new DELETE /secondaryUser endpoint that:

- deletes the secondary user record from the multiple-account table
- deletes the linked supporter product data record
- performs both deletes in a single DynamoDB transaction
- Adds lambda delete permission for supporterProductData table
- It also adds an integration test covering the full flow from invitation creation through revocation.

## How to Test
In order to test a secondary account deletion, one has to create it first.
You can use the guide available in the [Create accept invitation endpoint](https://github.com/guardian/support-service-lambdas/pull/3586) Testing section

After that you can use the secondaryIdentity id along with the subscription name to make the request as follows:
```
curl --location --request DELETE 'https://multiple-account-api-code.support.guardianapis.com/subscriptions/A-S00974337/secondary-users/{secondaryIdentityId}' \
--header 'x-identity-id: 112809589' \
--header 'x-api-key: [GET_API_KEY_FROM_AWS]' \
```

---
  - https://app.asana.com/0/0/1214265189400427

### DIFF
--- a/handlers/multiple-account-api/src/deleteSecondaryUserEndpoint.ts
+++ b/handlers/multiple-account-api/src/deleteSecondaryUserEndpoint.ts
@@ -11,13 +11,13 @@ import { z } from 'zod';
 import { buildComposedSubscriptionName } from './helpers';
 import type { SecondaryUserRepository } from './secondaryUserRepository';
 
-export const deleteSecondaryUserBodySchema = z.object({
+export const deleteSecondaryUserPathSchema = z.object({
 	subscriptionName: z.string(),
 	secondaryIdentityId: z.string(),
 });
 
-export type DeleteSecondaryUserBody = z.infer<
-	typeof deleteSecondaryUserBodySchema
+export type DeleteSecondaryUserPath = z.infer<
+	typeof deleteSecondaryUserPathSchema
 >;
 
 export const deleteSecondaryUserEndpoint =
@@ -26,9 +26,9 @@ export const deleteSecondaryUserEndpoint =
 		secondaryUserRepository: SecondaryUserRepository,
 		dynamoClient: DynamoDBClient,
 	) =>
-	async (body: DeleteSecondaryUserBody): Promise<APIGatewayProxyResult> => {
+	async (path: DeleteSecondaryUserPath): Promise<APIGatewayProxyResult> => {
 		try {
-			const { subscriptionName, secondaryIdentityId } = body;
+			const { subscriptionName, secondaryIdentityId } = path;
 			const composedSubscriptionName = buildComposedSubscriptionName(
 				subscriptionName,
 				secondaryIdentityId,
--- a/handlers/multiple-account-api/src/deleteSecondaryUserEndpoint.ts
+++ b/handlers/multiple-account-api/src/deleteSecondaryUserEndpoint.ts
@@ -1,0 +1,59 @@
+import {
+	type DynamoDBClient,
+	TransactWriteItemsCommand,
+} from '@aws-sdk/client-dynamodb';
+import { logger } from '@modules/logger/logger';
+import { buildErrorResponse } from '@modules/routing/apiGatewayResponses';
+import type { Stage } from '@modules/stage';
+import { getDeleteSupporterRatePlanTransaction } from '@modules/supporter-product-data/supporterProductData';
+import type { APIGatewayProxyResult } from 'aws-lambda';
+import { z } from 'zod';
+import type { SecondaryUserRepository } from './secondaryUserRepository';
+
+export const deleteSecondaryUserBodySchema = z.object({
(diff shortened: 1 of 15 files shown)


handlers/multiple-account-api/test/deleteSecondaryUserIntegration.test.ts
```diff
@@ -0,0 +1,156 @@
...
+beforeEach(async () => {
+	const zuoraClient = await ZuoraClient.create(stage);
+	const zuoraCatalog = await getZuoraCatalogFromS3(stage);
+	const productCatalog = await getProductCatalogFromApi(stage);
+	const identityClient = await IdentityClient.create(
+		stage,
+		`/${stage}/support/multiple-account-api/identity-client-access-token`,
+	);
+
+	const createEndpoint = createInvitationEndpoint(
```
paul-daniel-dempsey, line 46: think this test is failing as this point at the moment (possibly the invitation already exists?)

handlers/multiple-account-api/src/deleteSecondaryUserEndpoint.ts
```diff
@@ -0,0 +1,59 @@
...
+		secondaryUserRepository: SecondaryUserRepository,
+		dynamoClient: DynamoDBClient,
+	) =>
+	async (body: DeleteSecondaryUserBody): Promise<APIGatewayProxyResult> => {
+		try {
+			const { subscriptionName, secondaryIdentityId } = body;
+			const composedSubscriptionName = `${subscriptionName}-${secondaryIdentityId}`;
+			logger.mutableAddContext(composedSubscriptionName);
+
+			// Carry out the secondary user deletion and deletion of the support product data record
```
paul-daniel-dempsey, line 38: Nice comment to help understand this, so `TransactItems` rolls back if any fail

handlers/multiple-account-api/src/deleteSecondaryUserEndpoint.ts
```diff
@@ -0,0 +1,59 @@
...
+							secondaryIdentityId,
+						),
+						getDeleteSupporterRatePlanTransaction(
+							stage,
+							secondaryIdentityId,
+							composedSubscriptionName,
+						),
+					],
+				}),
+			);
```
paul-daniel-dempsey, line 54: perhaps a log.logger line here to indicate succesful deletion in secondaryUserRepository and suporterProductDate for this secondaryIdentityId?

paul-daniel-dempsey (commented):
Looks very neat and concise to me. Think worth Rupert checking this over first. Could not get the integration test to run successfully so may need to check that (or update description how to run if I have missed something). Amazing backend work on this api 👏 👏 👏

handlers/multiple-account-api/src/deleteSecondaryUserEndpoint.ts
```diff
@@ -0,0 +1,59 @@
...
+export const deleteSecondaryUserEndpoint =
+	(
+		stage: Stage,
+		secondaryUserRepository: SecondaryUserRepository,
+		dynamoClient: DynamoDBClient,
+	) =>
+	async (body: DeleteSecondaryUserBody): Promise<APIGatewayProxyResult> => {
+		try {
+			const { subscriptionName, secondaryIdentityId } = body;
+			const composedSubscriptionName = `${subscriptionName}-${secondaryIdentityId}`;
```
tjmw: Does this logic to create the `composedSubscriptionName` exist anywhere else? I'm just wondering if this should be in a shared function if so.

handlers/multiple-account-api/src/index.ts
```diff
@@ -91,4 +95,21 @@ export const handler: Handler = Router([
 			);
 		}),
 	},
+	{
+		httpMethod: 'DELETE',
+		path: '/secondary-users',
```
tjmw: For consistency with the other PR, should this be `DELETE /secondary-users/{subscriptionName}` with no body? MDN suggests deletes shouldn't have a body: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Methods/DELETE

handlers/multiple-account-api/src/deleteSecondaryUserEndpoint.ts
maloky: well spotted. it's being used when users accept the invitation for creating the `SupporterProductData` entry. will def move it out and call it from there in both places 👍

handlers/multiple-account-api/src/index.ts
maloky: yep, makes sense to me

rupertbates (approved):
👍

PR #3616, opened 2026-06-10
Sales Tax Api : retrieve CAD tax rates from Zuora

paul-daniel-dempsey:
What does this change?
modify existing API which to return CAD tax rate percentages for a given country+productKey from Zuora.

Return is in format of type CaState

POST /tax-rates
Request : {productKey: productKeySchema, country: isoCountrySchema}
Response : {CaState}

How has this change been tested?
POST : https://sales-tax-api-code.support.guardianapis.com/tax-rates
BODY : `{
"country: "CA",
}`

POST : https://sales-tax-api-code.support.guardianapis.com/tax-rates
BODY : `{
"productKey": "DigitalSubscription",
"country: "CA",
}`

Note:
This api authenticates using an x-api-key within the header and requires generating/supplying.
The token can be found under AWS API GateWay/API Keys (sales-tax-api-CODE/PROD).

How can we measure success?
Found: Response 200
```
{
  "AB": 0.05,
  "BC": 0.12,
  "MB": 0.12,
  "NB": 0.15,
  "NL": 0.15,
  "NT": 0.05,
  "NS": 0.15,
  "NU": 0.05,
  "ON": 0.13,
  "PE": 0.15,
  "QC": 0.1498,
  "SK": 0.11,
  "YT": 0.05
}
```

---
  - https://app.asana.com/0/0/1215133618429653


modules/zuora/src/tax.ts
```diff
@@ -0,0 +1,35 @@
...
+import type { ZuoraClient } from './zuoraClient';
+
+export const getZuoraTaxCodes = async <
+	T extends z.ZodType = typeof zuoraTaxCodeSchema,
+>(
+	zuoraClient: ZuoraClient,
+): Promise<z.infer<T>> => {
+	const path = `settings/tax-codes`;
+	return zuoraClient.get(path, zuoraTaxCodeSchema);
+};
```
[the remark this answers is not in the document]
paul-daniel-dempsey: will keep zod change as used like this elsewhere

modules/zuora/src/tax.ts
```diff
@@ -0,0 +1,35 @@
...
+
+export const getZuoraTaxPeriods = async <
+	T extends z.ZodType = typeof zuoraTaxPeriodsSchema,
+>(
+	zuoraClient: ZuoraClient,
+): Promise<z.infer<T>> => {
+	const path = `settings/tax-rate-periods`;
+	return zuoraClient.get(path, zuoraTaxPeriodsSchema);
+};
```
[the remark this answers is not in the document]
paul-daniel-dempsey: will keep zod change as used like this elsewhere

modules/zuora/src/tax.ts
```diff
@@ -0,0 +1,35 @@
...
+
+export const getZuoraTaxRates = async <
+	T extends z.ZodType = typeof zuoraTaxRateSchema,
+>(
+	zuoraClient: ZuoraClient,
+	id: string,
+): Promise<z.infer<T>> => {
+	const path = `settings/tax-rate-periods/${id}/tax-rates?page=1&size=500`;
+	return zuoraClient.get(path, zuoraTaxRateSchema);
+};
```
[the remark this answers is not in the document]
paul-daniel-dempsey: will keep zod change as used like this elsewhere

handlers/sales-tax-api/src/index.ts
```diff
@@ -12,7 +21,7 @@ export const handler = Router([
 			taxRatesRequestSchema,
 			async (event, path, body) => {
 				logger.log('Received POST /tax-rates request', body);
-				return taxRatesRequestEndpoint(body);
+				return taxRatesEndpoint(await lazyZuoraClient.get(), body);
```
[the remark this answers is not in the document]
paul-daniel-dempsey: url and description corrected

handlers/sales-tax-api/src/taxRatesEndpoint.ts
```diff
@@ -1,57 +1,147 @@
...
-	QC: 0.1498,
-	SK: 0.11,
-	YT: 0.05,
+type ProductKey = z.infer<typeof productKeySchema>;
+const taxExclusiveProductCodeNames: Partial<Record<ProductKey, string>> = {
+	SupporterPlus: `Supporter Plus`,
+	DigitalSubscription: `Supporter Plus`, // DigitalSubscription is a tax-exclusive product, but it uses the same tax code as Supporter Plus
+};
+
+const caStatesDefault: TaxRatesResponse = {
```
tjmw: I don't think we'd ever want to return a tax rate of 0 so maybe better to throw an error rather than use defaults if there's ever a problem getting the data from Zuora?

handlers/sales-tax-api/src/taxRatesEndpoint.ts
```diff
@@ -1,57 +1,147 @@
...
-	NU: 0.05,
-	ON: 0.13,
-	PE: 0.15,
-	QC: 0.1498,
-	SK: 0.11,
-	YT: 0.05,
+type ProductKey = z.infer<typeof productKeySchema>;
+const taxExclusiveProductCodeNames: Partial<Record<ProductKey, string>> = {
+	SupporterPlus: `Supporter Plus`,
+	DigitalSubscription: `Supporter Plus`, // DigitalSubscription is a tax-exclusive product, but it uses the same tax code as Supporter Plus
```
tjmw: What are we mapping to here? Is it a tax code? When I look in Zuora it looks like we're using a global tax code in both cases, or am I missing something?

handlers/sales-tax-api/src/taxRatesEndpoint.ts
```diff
@@ -1,57 +1,147 @@
...
+	const zuoraTaxPeriod = getZuoraTaxPeriod(
+		zuoraTaxCode.id,
+		zuoraTaxPeriods.taxRatePeriods,
+	);
+	if (!zuoraTaxPeriod) {
+		throw new ValidationError(`invalid period for productKey:${productKey}`);
+	}
+
+	const zuoraTaxRates = await getZuoraTaxRates(zuoraClient, zuoraTaxPeriod.id);
+	const cadZuoraTaxRates = getCadZuoraTaxRates(zuoraTaxRates.taxRates);
```
tjmw: Really minor but what do you think about calling this `extractCadZuoraTaxRates` instead of `get`? Get (as in the line above) makes me think we're actually fetching something, but I think we're just interrogating the data here?

modules/zuora/src/tax.ts
```diff
@@ -0,0 +1,35 @@
+import type { z } from 'zod';
+import {
+	zuoraTaxCodeSchema,
+	zuoraTaxPeriodsSchema,
+	zuoraTaxRateSchema,
+} from './types/objects/tax';
+import type { ZuoraClient } from './zuoraClient';
+
+export const getZuoraTaxCodes = async <
+	T extends z.ZodType = typeof zuoraTaxCodeSchema,
```
tjmw: Do these need to be generic? Could we give a name to the schema inferred TypeScript type and use that? (I might be missing something around what the zuoraClient expects, so no worries if not possible).

handlers/sales-tax-api/src/taxRatesEndpoint.ts
paul-daniel-dempsey: Have created a 'TaxCodeName` type now, possibly this makes it clearer?

modules/zuora/src/tax.ts
paul-daniel-dempsey: No, I think I incorectly followed a template fom elsewheer in the codebase, have removed all zod gerneic items from this file now. Updated.

handlers/sales-tax-api/src/taxRatesEndpoint.ts
paul-daniel-dempsey: Ah ok, have now refactored  to locate the missing states/key and list them in a validation error.  Updated. Still keeps the default 0's though, I could possibly somehow look through the CaState type via  keys I think, might look harder to read thats all (and will need to figure that out) - WDYT?

handlers/sales-tax-api/src/taxRatesEndpoint.ts
paul-daniel-dempsey: Ok refactored to use the newly created `Supporter Plus Global Tax` and `Digital Pack Global Tax` TaxCodeNames now, Updated.

handlers/sales-tax-api/src/taxRatesEndpoint.ts
paul-daniel-dempsey: Have refactored to locate the missing states/key and list them as a validation error. Updated. Still kept the default 0's though, I could possibly somehow look through the CaState type via keys I think, might look harder to read thats all (and will need to figure that out) - WDYT?

handlers/sales-tax-api/src/schemas.ts
```diff
@@ -24,3 +24,4 @@ export const taxRatesResponseSchema = z.object({
 	YT: z.number(),
 });
 export type TaxRatesResponse = z.infer<typeof taxRatesResponseSchema>;
+export type CaTaxRateState = keyof TaxRatesResponse;
```
rupertbates: We already have a type for this here: https://github.com/guardian/support-service-lambdas/blob/af5e733a86910e1a1417ccf47f0208cf02c299a5/modules/internationalisation/src/country.ts#L600

Better to use that one I think to keep it canonical

handlers/sales-tax-api/src/taxRatesEndpoint.ts
```diff
@@ -1,57 +1,161 @@
...
+import { type TaxRatesRequest, taxRatesResponseSchema } from './schemas';
+
+export class InternalServerError extends Error {
+	constructor(message: string) {
+		super(message);
+		this.name = 'InternalServerError';
+	}
+}
+
+type ProductKey = z.infer<typeof productKeySchema>;
```
rupertbates: Import this from productCatalog.ts

handlers/sales-tax-api/src/schemas.ts
tjmw: I did try using that but TS wasn't happy about mixing the two (i.e. indexing into the schema shape with `CaState`). Also as-is `CaState` is just `string` because `caStates` is `Record<string, string>`. But even removing that to tighten it up I still couldn't make it work without casting. If you're interested to pair to try and make it work I'd be up for that!

handlers/sales-tax-api/src/taxRatesEndpoint.ts
```diff
@@ -1,57 +1,161 @@
...
-			country,
-		});
 		return ok(
-			getSalesTaxRates({ productKey, country }),
-			taxRatesResponseSchema,
+			await getZuoraSalesTaxRates(zuoraClient, { productKey, country }),
 		);
 	} catch (error) {
-		logger.error('Error fetching sales tax', error);
+		logger.error('Error fetching Zuora taxes', error);
```
rupertbates: I don't think this try catch is really necessary, it will be caught and mapped to an error response in the router, so all this is adding is the log line

handlers/sales-tax-api/src/taxRatesEndpoint.ts
```diff
@@ -1,57 +1,161 @@
...
+	const zuoraTaxCodes = await getZuoraTaxCodes(zuoraClient);
+	if (!zuoraTaxCodes) {
+		throw new Error('no tax codes found');
+	}
+	const zuoraTaxCode = getProductZuoraTaxCode(
+		productKey,
+		zuoraTaxCodes.taxCodes,
+	);
+	if (!zuoraTaxCode) {
+		throw new ValidationError(`invalid productKey:${productKey}`);
```
rupertbates:
```suggestion
		throw new ValidationError(`No tax data found for productKey:  ${productKey}`);
```

tjmw: Good spot. I also wonder whether this should be a ValidationError. We already validated it was a valid product in the zod schema, so if we can't find tax data then that's feels like an internal server error. Any thoughts?

handlers/sales-tax-api/src/taxRatesEndpoint.ts
```diff
@@ -1,57 +1,161 @@
...
+	const zuoraTaxPeriod = getZuoraTaxPeriod(
+		zuoraTaxCode.id,
+		zuoraTaxPeriods.taxRatePeriods,
+	);
+	if (!zuoraTaxPeriod) {
+		throw new ValidationError(`invalid period for productKey:${productKey}`);
+	}
+
+	const zuoraTaxRates = await getZuoraTaxRates(zuoraClient, zuoraTaxPeriod.id);
+	const cadZuoraTaxRates = extractCadZuoraTaxRates(zuoraTaxRates.taxRates);
```
rupertbates: Wouldn't it be better to make this `extractZuoraTaxRatesForCountry(rates, country)` given that we have the country available in this function?

handlers/sales-tax-api/src/taxRatesEndpoint.ts
```diff
@@ -1,57 +1,161 @@
...
+function getZuoraTaxPeriod(
+	zuoraTaxCode: string,
+	zuoraTaxPeriods: ZuoraTaxPeriod[],
+) {
+	return zuoraTaxPeriods.find(
+		(zuoraTaxPeriod) => zuoraTaxPeriod.taxCodeId === zuoraTaxCode,
+	);
+}
+
+function extractCadZuoraTaxRates(
```
rupertbates: See comment above

handlers/sales-tax-api/src/taxRatesEndpoint.ts
```diff
@@ -1,57 +1,161 @@
...
+	zuoraTaxRates: ZuoraTaxRate[],
+): ZuoraTaxRate[] {
+	return zuoraTaxRates.filter(
+		(zuoraTaxRate) => zuoraTaxRate.country === getCountryNameByIsoCode('CA'),
+	);
+}
+
+function createCadStateTaxRates(
```
rupertbates, line 119: This could be generic for the country as well but up to you if it's worth putting the time in.

modules/zuora/src/types/objects/tax.ts
```diff
@@ -0,0 +1,64 @@
+import { z } from 'zod';
+
+export const zuoraTaxCode = z.object({
```
rupertbates:
```suggestion
export const zuoraTaxCodeSchema = z.object({
```

modules/zuora/src/types/objects/tax.ts
```diff
@@ -0,0 +1,64 @@
...
+	id: z.string(),
+	taxEngineId: z.string(),
+	active: z.boolean(),
+	name: z.string(),
+	description: z.string(),
+});
+export const zuoraTaxCodeSchema = z
```
rupertbates:
```suggestion
export const zuoraTaxCodesSchema = z
```

modules/zuora/src/types/objects/tax.ts
```diff
@@ -0,0 +1,64 @@
...
+	.object({
+		taxCodes: z.array(zuoraTaxCode),
+	})
+	.optional();
+export type ZuoraTaxCode = z.infer<typeof zuoraTaxCode>;
+export type ZuoraTaxCodes = z.infer<typeof zuoraTaxCodeSchema> | undefined;
+
+export const zuoraTaxPeriod = z.object({
```
rupertbates:
```suggestion
export const zuoraTaxPeriodSchema = z.object({
```

modules/zuora/src/types/objects/tax.ts
```diff
@@ -0,0 +1,64 @@
...
+});
+export const zuoraTaxPeriodsSchema = z
+	.object({
+		taxRatePeriods: z.array(zuoraTaxPeriod),
+	})
+	.optional();
+export type ZuoraTaxPeriod = z.infer<typeof zuoraTaxPeriod>;
+export type ZuoraTaxPeriods = z.infer<typeof zuoraTaxPeriodsSchema> | undefined;
+
+export const zuoraTaxRate = z.object({
```
rupertbates:
```suggestion
export const zuoraTaxRateSchema = z.object({
```

rupertbates: Also we could maybe omit a few of these fields? I can't imagine we need them all.

modules/zuora/src/types/objects/tax.ts
```diff
@@ -0,0 +1,64 @@
...
+	taxLocationCode2: z.string().nullable(),
+	taxRateDescription2: z.string().nullable(),
+	taxRate3: z.number(),
+	taxRateType3: z.string().nullable(),
+	taxName3: z.string().nullable(),
+	taxJursdiction3: z.string().nullable(),
+	taxLocationCode3: z.string().nullable(),
+	taxRateDescription3: z.string().nullable(),
+});
+export const zuoraTaxRateSchema = z.object({
```
rupertbates:
```suggestion
export const zuoraTaxRatesSchema = z.object({
```

handlers/sales-tax-api/src/taxRatesEndpoint.ts
rupertbates: I think you're right a 500 makes more sense

handlers/sales-tax-api/src/schemas.ts
rupertbates: Happy to give it a go!

modules/zuora/src/types/objects/tax.ts
tjmw: Yep definitely - we can omit almost all of them, we only use a few.

handlers/sales-tax-api/src/taxRatesEndpoint.ts
tjmw: I think I'll leave this one for now. Even if we do go on to do this for further countries I think the logic will be different enough for each (is it always state level?) that we'll need to figure out how it should work when we get to it.

modules/internationalisation/src/country.ts
```diff
@@ -57,7 +59,25 @@ export const usStates: Record<string, string> = {
 	WI: 'Wisconsin',
 	WY: 'Wyoming',
 };
-export const caStates: Record<string, string> = {
+
+export const caStateSchema = z.enum([
```
tjmw: We've switched to a "schema first" approach here, which is common in our projects. This was so that we've got a canonical schema and type for `CaState` which we can use elsewhere (the pre-existing `CaState` was just a `string`). We should follow up and make the other types in here consistent.

rupertbates (approved):
NIce 👍

PR #3617, opened 2026-06-10
Rename the supporter-product-data-lambdas execution failure alarm

rupertbates:
## What does this change?
https://github.com/guardian/support-service-lambdas/pull/3498 migrated the supporter product data state machine to this repo, however it failed to deploy in PROD because one of the alarms in the new stack had the same name as an alarm in the old stack. This PR changes the name of the new alarm so there is no collision


paul-daniel-dempsey (approved):
LG

PR #3618, opened 2026-06-10
Fix schema

rupertbates:
## What does this change?
Fix the schema of the supporter product rate plan item


paul-daniel-dempsey (approved):
Ok, currency and amount was not in its own object but held seperately, thanks for fixing 👍

PR #3620, opened 2026-06-11
Update supporter product data writes to include primarySubscriptionName

rupertbates:
## What does this change?
As part of the multiple accounts work described in [this doc](https://docs.google.com/document/d/1eqMn9zkg-iRK3Gi9ecmgGh8WWUYpLX6B3cOasTRYpns/edit?tab=t.0#heading=h.anbq3pjpq22s) a new field has been added to the SupporterProductData Dynamo DB table. This field `primarySubscriptionName` links a secondary subscription record to the original subscription which grants access for the secondary user.

As all writes to the SupporterProductData table are done via the `processSupporterRatePlanItemLambda` this needs to be updated to be aware of the new field. This PR makes that change and also adds some integration tests for the Dynamo write logic.

---
  - https://app.asana.com/0/0/1214265189400430

### DIFF
--- a/handlers/supporter-product-data-lambdas/src/services/dynamoService.ts
+++ b/handlers/supporter-product-data-lambdas/src/services/dynamoService.ts
@@ -55,6 +55,10 @@ export class DynamoService {
 		let updateExpression =
 			'SET productRatePlanId = :productRatePlanId, productRatePlanName = :productRatePlanName, termEndDate = :termEndDate, contractEffectiveDate = :contractEffectiveDate, expiryDate = :expiryDate';
 
+		// This is to prevent stale data in the data store if for instance the first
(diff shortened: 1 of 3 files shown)


handlers/supporter-product-data-lambdas/src/services/dynamoService.ts
```diff
@@ -64,6 +68,22 @@ export class DynamoService {
 			};
 			updateExpression +=
 				', contributionAmount = :contributionAmount, contributionCurrency = :contributionCurrency';
+		} else {
```
maloky, line 69: wouldn't it be simpler to add all values to `expressionValues ` and remove if not present?

maloky (approved):
looks good to me 👍

PR #3621, opened 2026-06-12
type safe zuora object query functionality

johnduffell:
folloiwng on from https://github.com/guardian/support-service-lambdas/pull/3567 and https://github.com/guardian/support-service-lambdas/pull/3584 

This PR
- adds a new object query library that has type safety/completion for input and output values
- adds support for "accounts" and "orders" to prove the concept
- gets the skeleton user-subscriptions-api handler to call object query for accounts and log the response

See inline comments for the interesting bits

The next PR will actually use the query response to produce the response from the lambda.


buildcheck/data/build.ts
```diff
@@ -794,14 +794,14 @@ const userSubscriptionsApi: HandlerDefinition = {
 	},
 	devDependencies: {
 		...devDeps['@types/aws-lambda'],
-
 		...devDeps['@redocly/cli'],
+		...devDeps['tsx'],
```
johnduffell, line 798: to run the runManual scripts - for some reason intellij 2026.2 has dropped the feature to use ts files directly in the run configuration.  I couldn't find an easy, single file, one click way round other than to make them a script calling `tsx` (see runMeEndpoint.ts below)

handlers/user-subscriptions-api/src/meEndpoint.ts
```diff
@@ -49,10 +52,18 @@ const staticResponse: MeResponse = {
...
+export class MeEndpoint {
+	constructor(private readonly zuoraClient: ZuoraClient) {}
+	async handle(identityId: string): Promise<APIGatewayProxyResult> {
+		const userSubscriptions = await objectQuery.accounts.execute(
+			this.zuoraClient,
+			['id', 'name', 'accountNumber', 'balance'],
+			['subscriptions.rateplans'],
+			[{ field: 'IdentityId__c', operator: 'EQ', value: identityId }],
+			99,
+		);
```
johnduffell, line 64: under the covers this is a type safe way of calling this
https://developer.zuora.com/v1-api-reference/api/object-queries/queryaccounts

modules/logger/src/logger.ts
```diff
@@ -74,19 +74,23 @@ export class Logger {
...
 		functionName: string | (() => string) = fn.name,
 		callerInfo: string = getCallerInfo(),
-		argsToLoggable: (args: TArgs) => LoggableInput,
-		responseToLoggable: (result: TReturn) => unknown = (result) => result,
-	): AsyncFunction<TArgs, TReturn> {
+		argsToLoggable: (args: Parameters<TFn>) => LoggableInput,
+		responseToLoggable: (result: Awaited<ReturnType<TFn>>) => unknown = (
+			result,
+		) => result,
+	): TFn {
```
johnduffell, line 97: before, wrapFn used to preserve the parameter and return types, but throw away the generic types.
Since we'd like to rely on the connection between the input schema and the output type, via generics, we need to preserve the type parameter to TFn too.

This function needed to change to treat `fn` as an unknown function, and cast it back at the end to get the generic type back.  That logic is in `aroundAsync` function.  The logic to actually do the logging is in `createFunctionLoggingHooks`  This tidies up this file to improve readability.

modules/zuora/src/objectQuery/expandSchemas/accountItemSchema.ts
```diff
@@ -0,0 +1,120 @@
+import z from 'zod';
```
johnduffell, line 1: I've generated the schemas from the object query specific documentation linked in the javadoc comments in each file.  There are arguments for putting these in a shared place, because they reflect zuora's object model rather than being specific to object queries and I'd be happy to do so. (contrast with the rest response model in `modules/zuora/src/types/objects/` where one call returns a whole hierarchy by design)

modules/zuora/src/objectQuery/queries/ordersObjectQuery.ts
```diff
@@ -0,0 +1,38 @@
+import { z } from 'zod';
```
johnduffell, line 1: we don't actually need an orders object query yet, but I wanted a second example in the file, so the type checker wouldn't narrow things too much.

modules/zuora/src/objectQuery/index.ts
```diff
@@ -0,0 +1,25 @@
+import { accountObjectQuery } from '@modules/zuora/objectQuery/queries/accountObjectQuery';
```
johnduffell, line 1: this is a convenience file, I am in two minds as to whether it's useful, or we should just reference the individual files from queries package in our code.  Happy to hear opinions.

modules/zuora/src/objectQuery/doesNotHaveDuplicateResponseKey.ts
```diff
@@ -0,0 +1,52 @@
...
+ * its literal key type unless its responseKey collides with another chosen key,
+ * in which case the element type becomes `HasDuplicateResponseKey` so the
+ * offending array entry fails to type-check exactly at the call site. Being
+ * homomorphic over the tuple (`[I in keyof TKeys]`), it preserves inference of
+ * the chosen keys.
+ *
+ * Example: `subscriptions`, `subscriptions.rateplans`, and
+ * `subscriptions.rateplans.rateplancharges` can all map to the same response
+ * key (`subscriptions`). Selecting more than one of those expands is invalid
+ * because the resulting response shape would collide on that key.
```
johnduffell, line 39: all that is to say that you'll get a compile time error if you use clashing keys.  It doesn't quite give you all the information, but it does stop you

Image

or
```
src/meEndpoint.ts:62:5 - error TS2322: Type '"subscriptions.rateplans.rateplancharges"' is not assignable to type '"subscriptions.rateplans.rateplancharges" & HasDuplicateResponseKey'.
  Type 'string' is not assignable to type 'HasDuplicateResponseKey'.

62     'subscriptions.rateplans.rateplancharges',
       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

src/meEndpoint.ts:64:5 - error TS2322: Type '"subscriptions.rateplans"' is not assignable to type '"subscriptions.rateplans" & HasDuplicateResponseKey'.
  Type 'string' is not assignable to type 'HasDuplicateResponseKey'.

64     'subscriptions.rateplans',
       ~~~~~~~~~~~~~~~~~~~~~~~~~
```


The other option is to have it as a run time only check (or just let one key overwrite the others)

In zuora itself, it uses the broadest version of all the ones you specify, which would probably require a bit more type machinery to union together the schemas dynamically rather than doing it up front.  We can look into that if it's necessary later.

modules/logger/src/logger.ts
```diff
@@ -25,11 +25,8 @@ export class Logger {
 		if (index !== -1) {
 			this.prefix = this.prefix.slice(0, index);
 		} else {
-			this.logFn(
-				this.getMessage(
-					getCallerInfo(),
-					`dropContext: context value "${value}" not found`,
-				),
+			this.emitLog(getCallerInfo())(
```
johnduffell, line 28: various changes in this file to have an emitLog/Error function that does the prefixing for free (prefixes are labels for all log messages to say what's being processed e.g. a subscription id, or identity id, to make it easier to search the logs)

modules/logger/test/logger.test.ts
```diff
@@ -275,6 +275,20 @@ Error: fail 42
...
+			logOnEntryOnly: ['some-arg'],
+		}));
+
+		await wrapped(7);
+
+		expect(logs[0]).toContain('ENTRY ARGS');
+		expect(logs[1]).toContain('EXIT');
+		expect(logs[1]).not.toContain('SHORT_ARGS');
+		expect(errors).toEqual([]);
+	});
```
johnduffell, line 291: this was a missing test spotted by copilot

modules/routing/src/lambdaHandler.ts
```diff
@@ -71,7 +73,7 @@ export function LambdaHandlerWithServices<ConfigType, Services, E>(
 
 	const handlerProps: Lazy<Services> = lazyEnv.then(
 		logger.wrapFn(
-			async ({ stage, stack, app }) => {
+			async ({ stage, stack, app }: GuEnv) => {
```
johnduffell, line 76: now that type inference of a wrapped function goes from the function outwards, this needs an explicit type, which is a loss (in exchange for the gain elsewhere), but it makes sense to be explicit anyway.

The other option is to have both a generic preserving and inwards inference preserving version of wrapFn, rather than just one or the other.

modules/zuora/src/creditBalanceAdjustment.ts
```diff
@@ -1,15 +1,25 @@
-import type { z } from 'zod';
+import type { z, ZodTypeDef } from 'zod';
```
johnduffell, line 1: these few files are updating the new the type inference, I don't think the changes are a bad thing, in fact it's arguable simpler/less repetition

modules/zuora/src/restClient.ts
```diff
@@ -26,19 +27,27 @@ export type RestResult = {
...
 
-	public async get<I, O, T extends z.ZodType<O, z.ZodTypeDef, I>>(
+	public async get<S extends ZodType<unknown, ZodTypeDef, unknown>>(
 		path: string,
-		schema: T,
-	): Promise<O> {
-		return await this.fetchWithLogging(getCallerInfo(1))(path, 'GET', schema);
+		schema: S,
+		urlSearchParams?: URLSearchParams,
+	): Promise<z.infer<S>> {
```
johnduffell, line 34: this is the important change to get type inference working - now the return type is inferred from the input schema rather than needing to be picked up as an explicit type parameter

modules/zuora/src/restClient.ts
```diff
@@ -95,25 +104,39 @@ export abstract class RestClient {
...
+		schema: S,
 		body?: string,
 		headers?: Record<string, string>,
-	): Promise<O> {
+		params?: URLSearchParams,
+	): Promise<z.infer<S>> {
 		const authorisation = await this.tokenProvider.getAuthorisation();
 		const pathWithoutLeadingSlash = path.replace(/^\//, '');
-		const url = `${authorisation.baseUrl}/${pathWithoutLeadingSlash}`;
+		const url = `${authorisation.baseUrl}/${pathWithoutLeadingSlash}${params === undefined ? '' : '?' + params.toString()}`;
```
johnduffell, line 139: now we support url params as a proper object rather than using string concat

modules/zuora/src/zuoraClient.ts
```diff
@@ -22,20 +21,31 @@ export class ZuoraClient extends RestClient {
...
-	override async fetch<I, O, T extends z.ZodType<O, ZodTypeDef, I>>(
+	override async fetch<S extends ZodType<unknown, ZodTypeDef, unknown>>(
 		path: string,
 		method: string,
-		schema: T,
+		schema: S,
 		body?: string,
 		headers?: Record<string, string>,
-	): Promise<O> {
+		params?: URLSearchParams,
```
johnduffell, line 30: changes here to allow url params to be passed through

shtukas (approved):

PR #3625, opened 2026-06-15
Add taxMode attribute to rate plans in product catalog api

tjmw:
## What does this change?

Surfaces a new attribute `taxMode` in the product catalog API at the rate plan level.

The `taxMode` is actually set on the charge, which is within the rate plan, but currently this should always be consistent for all charges within a rate plan. If that is ever not the case then the generate catalog step will fail and we'll receive an alarm.

The bulk of the changes in this PR are updating sandbox fixtures to make charge settings consistent within rate plans. These are already consistent in prod.

Deployed to CODE and can see the new attribute:

Screenshot 2026-06-15 at 15 55 42


modules/product-catalog/src/generateSchema.ts
```diff
@@ -119,6 +119,7 @@ const generateProductRatePlanSchema = (
 		${getBillingPeriodForRatePlan(productRatePlan)}
 		termType: termTypeSchema,
 		termLengthInMonths: z.number(),
+		taxMode: z.string().nullable().optional(),
```
tjmw: This is optional to avoid a race condition where the new schema is used as lambdas in this repo are deployed but the new attribute isn't yet present in the product catalog response because it hasn't been regenerated yet. I'll remove the `optional()` part in a follow up PR.

modules/zuora-catalog/src/zuoraCatalogSchema.ts
```diff
@@ -63,6 +63,7 @@ export const zuoraCatalogSchema = z.object({
 							billingPeriod: z.nullable(z.string()),
 							triggerEvent: z.string(),
 							description: z.nullable(z.string()),
+							taxMode: z.nullable(z.string()),
```
rupertbates: You think it's better to keep this as a string rather than a union type?

rupertbates (approved):
Just one question about types

modules/zuora-catalog/src/zuoraCatalogSchema.ts
tjmw: Yeah it's a good question, I also wondered about this. I don't know how likely it is that other tax modes could be created in future so decided to keep it flexible. What do you think? If we're confident these will be very stable then the enum approach sounds good.

tjmw: Made this change in 1204cec.

modules/product-catalog/src/generateSchema.ts
```diff
@@ -119,6 +119,7 @@ const generateProductRatePlanSchema = (
 		${getBillingPeriodForRatePlan(productRatePlan)}
 		termType: termTypeSchema,
 		termLengthInMonths: z.number(),
+		taxMode: z.enum(['TaxExclusive', 'TaxInclusive']).nullable().optional(),
```
johnduffell, line 122: why is this both nullable and optional?
it looks like in the zuora catalog it's nullable
https://developer.zuora.com/v1-api-reference/api/api/object-queries/queryproductrateplanchargebykey#object-queries/queryproductrateplanchargebykey/t=response&c=200&path=taxmode
but where does the optionality come from?

PR #3628, opened 2026-06-15
allow public paths on API gateway lambdas

johnduffell:
required for https://github.com/guardian/support-service-lambdas/pull/3610

This PR updates SrLambda CDK construct to allow the docs to be public on an otherwise API key'ed api gateway.

This is needed so lambda can serve their own openapi docs.


cdk/lib/cdk/SrApiLambda.ts
```diff
@@ -125,4 +129,14 @@ export class SrApiLambda extends SrLambda {
 			});
 		}
 	}
+
+	addPublicPath(path: string) {
```
rupertbates, line 140: Could you include a comment describing the usage of this function (the when and why rather than the how) please?

rupertbates (approved):
This is great, just one suggestion asking for an additional comment

cdk/lib/cdk/SrApiLambda.ts
johnduffell:
```suggestion
	/**
	  * if your API gateway normally requires an API key, you can use this function to make specific routes public
	  * 
	  * this can be useful you want a lambda to self-serve documentation for access via a browser
	  */
	addPublicPath(path: string) {
```

PR #3631, opened 2026-06-16
Fix buildcheck repo root validation to support bare repository worktree paths

maloky:
### What

Updates `parseArguments` in `buildcheck/src/util/argsParser.ts` to accept repository root paths from bare Git repositories with worktrees.

Previously the validation only accepted paths ending exactly with `support-service-lambdas`, which rejected paths in the form `/support-service-lambdas.git/<worktreeName>` (e.g. `/support-service-lambdas.git/main`).

The updated check accepts:
- `/support-service-lambdas` — standard clone
- `/support-service-lambdas.git/<worktreeName>` — bare clone with a worktree

### Why

Running `buildcheck` from a bare repository worktree (a common local dev setup) caused a hard failure at argument parsing with `Invalid repository root`, making the tool unusable in that setup.

### DIFF
--- a/buildcheck/src/util/argsParser.ts
+++ b/buildcheck/src/util/argsParser.ts
@@ -30,12 +30,13 @@ export function parseArguments(argv: string[]): ParsedArgs {
 	}
 
 	const repoRoot = args[1];
+	const isSupportedRepoPath =
+		repoRoot.startsWith('/') &&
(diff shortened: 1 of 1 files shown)


johnduffell (approved):
looks good thanks for the update.  This logic is just for a safety check, so relaxing it makes sense.
Also I might start using worktrees as they look useful! 👏

PR #3632, opened 2026-06-16
Migrate sf-move-subscriptions-api from CloudFormation to CDK

andrade0:
Migrates `sf-move-subscriptions-api` from a committed, frozen `cfn.yaml` to a native CDK stack built on `SrApiLambda`.

The point is to fix the "missing deploys" bug: because the template was a static snapshot, the API Gateway `Deployment` logical ID never changed, so method changes never triggered a fresh deployment and stages drifted to "undeployed" (403s). With native CDK the deployment is hashed and regenerates whenever the API definition changes.

Behaviour is preserved where it matters — same Lambda function name, same API key name, and the same SSM config path (`/<stage>/membership/sf-move-subscriptions-api`) so the handler still finds its configuration. The app also gains a stable custom domain (`sf-move-subscriptions-api.membership.guardianapis.com`) alongside the existing execute-api URL.

One thing to flag: the API key value is regenerated, so Salesforce (the only caller) will need its credential updated — ideally pointed at the new stable domain rather than the raw execute-api URL. Worth coordinating before the PROD deploy.

CI-wise the app moves from the legacy `build` job to `gu-cdk-build`, riff-raff now points at the synthesised templates, and the old `cfn.yaml` is removed.

## Deploying

Heads up for whoever runs this: it can't go out as an in-place stack update. Every logical ID changes under srCDK, so CloudFormation hits a physical-name collision (the Lambda keeps its `sf-move-subscriptions-api-<stage>` name) and `AWS::EarlyValidation::ResourceExistenceCheck` blocks the changeset. The old stack has to be deleted so the new one is created fresh — which is fine since we're regenerating the API URL and key anyway.

CODE is already done (stack live, Lambda active, API + custom domain up). PROD needs the same one-time steps, ideally in a coordinated window alongside the Salesforce credential swap:
1. Delete the existing `membership-PROD-sf-move-subscriptions-api` stack.
2. Seed the jar into `membership-dist` at `membership/PROD/sf-move-subscriptions-api/sf-move-subscriptions-api.jar` — the fresh create reads its code from there before riff-raff's own upload runs.
3. Deploy via riff-raff.
4. Update the Salesforce credential to the new API key and stable domain.

## Testing
- `pnpm --filter cdk package` passes (type-check, lint, prettier, snapshot test, synth).
- Synthesised templates checked against the deployed CODE stack: logical IDs preserved where needed, hashed deployment confirmed.


shtukas (approved):

handlers/sf-move-subscriptions-api/riff-raff.yaml
```diff
@@ -6,18 +6,20 @@ allowedStages:
   - CODE
   - PROD
 deployments:
-  cfn:
+  sf-move-subscriptions-api-cloudformation:
```
johnduffell, line 9: nitpick but I would have kept it as `cfn`

cdk/lib/sf-move-subscriptions-api.ts
```diff
@@ -0,0 +1,39 @@
...
+		});
+
+		const lambda = new SrApiLambda(this, 'Lambda', {
+			lambdaOverrides: {
+				runtime: Runtime.JAVA_21,
+				architecture: Architecture.ARM_64,
+				fileName: 'sf-move-subscriptions-api.jar',
+				handler: 'com.gu.sf.move.subscriptions.api.Handler::handle',
+				memorySize: 1536,
+				timeout: Duration.seconds(300),
```
johnduffell: nitpick again!

```suggestion
				timeout: Duration.minutes(5),
```

johnduffell (approved):
fantastic work, 419 lines of impenetrable yaml become 39 of easy to read cdk 👏 👏 

I can't see any obvious missing/differences (no alarms in the old one) so if it works, that's good enough!

Not sure if there may be issues deploying?  I'm not sure if the lambda/api gateway might clash - worth taking care when deploying, worst case you might have to drop the stack before deploying.

graham228221 (approved):
Just confirming I've tested end-to-end in CODE including updating the credentials in SF, and all works great, thanks!

PR #3637, opened 2026-06-16
add AI MCP server for common build/check commands

johnduffell:
todo


johnduffell:
close in favour of https://github.com/guardian/support-service-lambdas/pull/3639 as our org doesn't support alternative/custom MCP servers, and a script should be as good.

PR #3640, opened 2026-06-17
Migrate sf-emails-to-s3-exporter from CloudFormation to CDK

andrade0:
Migrates `sf-emails-to-s3-exporter` from a committed, frozen `cfn.yaml` (487 lines) to a native CDK stack on `SrScheduledLambda` / `SrStack`.

Same motivation as the `sf-move-subscriptions-api` migration: the static template pinned the API Gateway `Deployment` logical ID, so method changes never triggered a fresh deployment and stages drifted to "undeployed". The L2 `RestApi` construct hashes the deployment and regenerates it whenever the API definition changes.

Behaviour is preserved — the scheduled lambda (Java 21, runs 03:00 and 03:30 daily), the S3-backed API Gateway (GET reads an object, POST deletes), the IAM role, API key and usage plan, and the seven CloudWatch alarms all keep their original names. The S3 integration URIs are reproduced exactly (`s3:path/{folder}/{item}` for the read, `emails-from-sf-<stage>.s3:path/?delete` for the delete).

CI-wise the app moves from the legacy `build` job to `gu-cdk-build`, riff-raff now points at the synthesised templates, and the old `cfn.yaml` is removed.

## Deploying

Picking up John's point from the sf-move PR — this can't go out as an in-place update. Several resources keep fixed physical names (the IAM role `emails-to-sf-api-<stage>-api-gateway-role`, the API key, the usage plan), so CloudFormation hits a name collision on the changeset. The old stack has to be dropped so the new one is created fresh — and this applies to CODE as well as PROD, since neither stage is on CDK yet.

The catch specific to this app: recreating the API key gives it a new value, and there's no stable custom domain here, so the `execute-api` URL changes too. Salesforce calls this endpoint and stores both the key and the URL, so after each cutover the new key value and URL have to be copied into the Salesforce config — otherwise the writeback to SF starts failing.

PROD steps, ideally in one coordinated window:
1. Delete the existing `membership-PROD-sf-emails-to-s3-exporter` stack.
2. Seed the jar into `membership-dist` at `membership/PROD/sf-emails-to-s3-exporter/sf-emails-to-s3-exporter.jar`.
3. Deploy via riff-raff.
4. Copy the new API key value and endpoint URL into the Salesforce config.

## Testing
- `pnpm --filter cdk package` passes (type-check, lint, prettier, snapshot test, synth).
- Snapshot reviewed: alarms are PROD-only (CODE 17 resources, PROD 24) — matching the old template.


shtukas (approved):

cdk/lib/sf-emails-to-s3-exporter.ts
```diff
@@ -0,0 +1,338 @@
...
+				{
+					schedule: Schedule.cron({ minute: '00,30', hour: '03' }),
+					description: 'Runs SF Emails to S3 Exporter',
+				},
+			],
+			// The 7 custom-metric alarms below replace the standard error alarm.
+			monitoring: { noMonitoring: true },
+			lambdaOverrides: {
+				runtime: Runtime.JAVA_21,
+				architecture: Architecture.X86_64,
```
johnduffell: should this be arm? arm is generally cheaper, can't see where it's set in the old cfn though.

cdk/lib/sf-emails-to-s3-exporter.ts
```diff
@@ -0,0 +1,338 @@
...
+				fileName: 'sf-emails-to-s3-exporter.jar',
+				handler: 'com.gu.sf_emails_to_s3_exporter.Handler::handleRequest',
+				memorySize: 512,
+				timeout: Duration.seconds(900),
```
johnduffell: minutes

cdk/lib/sf-emails-to-s3-exporter.ts
```diff
@@ -0,0 +1,338 @@
...
+				actions: ['s3:ListBucket'],
+				resources: [bucketArn, `${bucketArn}/*`],
+			}),
+		);
+		lambda.addToRolePolicy(
+			new PolicyStatement({
+				sid: 'readDeployedArtefact',
+				effect: Effect.ALLOW,
+				actions: ['s3:GetObject'],
+				resources: ['arn:aws:s3::*:membership-dist/*'],
```
johnduffell: does gucdk give this for free?

cdk/lib/sf-emails-to-s3-exporter.ts
```diff
@@ -0,0 +1,338 @@
...
+					'secretsmanager:DescribeSecret',
+					'secretsmanager:GetSecretValue',
+				],
+				resources: [
+					'arn:aws:secretsmanager:eu-west-1:865473395570:secret:CODE/Salesforce/ConnectedApp/AwsConnectorSandbox-jaCgRl',
+					'arn:aws:secretsmanager:eu-west-1:865473395570:secret:PROD/Salesforce/ConnectedApp/SFEmailsToS3-6QJGTX',
+					'arn:aws:secretsmanager:eu-west-1:865473395570:secret:CODE/Salesforce/User/EmailsToS3APIUser-EbXFEb',
+					'arn:aws:secretsmanager:eu-west-1:865473395570:secret:PROD/Salesforce/User/EmailsToS3APIUser-kGtUDC',
+				],
+			}),
```
johnduffell, line 94: might be worth extracting these into named values so they can have a name, so we end ip with lambda.addToRolePolicy(salesforceCODEandPRODPolicy) which would be easier to read.

On that note, should both lambdas have both code and prod access?

cdk/lib/sf-emails-to-s3-exporter.ts
```diff
@@ -0,0 +1,338 @@
...
+		caseResource.addMethod(
+			'GET',
+			new AwsIntegration({
+				service: 's3',
+				path: '{folder}/{item}',
+				integrationHttpMethod: 'GET',
+				options: {
+					credentialsRole: apiRole,
+					passthroughBehavior: PassthroughBehavior.WHEN_NO_MATCH,
+					timeout: Duration.millis(29000),
```
johnduffell: nitpick but it would be good if all the Duration in the file could be using the closest units rather than requiring calculations

cdk/lib/sf-emails-to-s3-exporter.ts
```diff
@@ -0,0 +1,338 @@
...
+				{
+					id: 'FailedToAuthenticateWithSalesforceAlarm',
+					metricName: 'failed_to_authenticate_with_sf',
+					alarmName: 'emails-from-sf failed to authenticate with Salesforce',
+					alarmDescription:
+						'Something went wrong authenticating with Salesforce',
+				},
+			];
+
+			alarms.forEach((alarm) => {
```
johnduffell, line 322: nice way to make it clearer 👏

handlers/sf-emails-to-s3-exporter/riff-raff.yaml
```diff
@@ -6,18 +6,20 @@ allowedStages:
   - CODE
   - PROD
 deployments:
-  cfn:
+  sf-emails-to-s3-exporter-cloudformation:
```
johnduffell, line 9: I noticed you changed this key in the last PR too, why can't it stay as `cfn` which seems to be the convention (keys are only is unique within the file)

cdk/lib/sf-emails-to-s3-exporter.ts
andrade0: Great suggestion, thanks! Switched it to arm64. The artefact is a pure JVM fat jar with no native dependencies, so it runs unchanged on Graviton and it is cheaper. You are right that it was not set in the old cfn, it was just defaulting to x86. I will give it a quick re-test in CODE with the Salesforce side before we merge, just to be safe.

cdk/lib/sf-emails-to-s3-exporter.ts
andrade0: Good catch, done. Switched to Duration.minutes(15) so it reads properly.

cdk/lib/sf-emails-to-s3-exporter.ts
andrade0: Yes it does, nice spot. GuLambdaFunction already grants GetObject on the distribution bucket and the exact jar key (you can see it in the snapshot under the DistributionBucketName statement), so I removed the manual membership-dist policy. It was just leftover from the old cfn.

cdk/lib/sf-emails-to-s3-exporter.ts
andrade0: Love this, much more readable. I pulled the ARNs into a named per-stage map (salesforceSecretArns). And great question on the access: no, each lambda only needs its own stage. Secrets.scala keys off sys.env("Stage") and only ever reads that stage's connected app and user, so I scoped CODE to the CODE secrets and PROD to the PROD secrets.

cdk/lib/sf-emails-to-s3-exporter.ts
andrade0: Agreed, thanks. I went through the whole file and switched everything to the closest unit, so seconds(29) for the integration timeouts and hours(1) for the session duration and the alarm period.

cdk/lib/sf-emails-to-s3-exporter.ts
andrade0: Thanks, glad you like it! 🙏

handlers/sf-emails-to-s3-exporter/riff-raff.yaml
andrade0: Good question, and sorry for the confusion! I would actually like to keep this one as sf-emails-to-s3-exporter-cloudformation. The gu-cdk-build job in ci-scala.yml uploads the CloudFormation templates to a riff-raff content directory named <app>-cloudformation, and riff-raff matches a cloud-formation deployment to its content directory by the deployment key, so the key has to line up with that name or it will not find the templates. The plain cfn key works for the non-CDK handlers because their build job uploads to a cfn content directory instead. I followed the same convention the other migrated handlers use, batch-email-sender for example. Happy to revisit if you would prefer a different approach though!

johnduffell: I don't quite understand your comment but I looked into it and it seems like the name was set up centrally that way in the GHA back in 2023 and not flagged at the time as being a deviation.
https://github.com/guardian/support-service-lambdas/pull/1954

At this point the benefit of changing existing ones back to "cfn" again is limited, it would slightly simplify the setup and also future migration PRs, but it's up to you whether to go ahead and do it.

cdk/lib/sf-emails-to-s3-exporter.ts
johnduffell: thanks for addressing the CODE/PROD thing.  Did you also see my suggestion about giving each policy an named const or similar, and importing that way?  That means I don't have to read the whole policy to understand its intent/reasoning.

cdk/lib/sf-emails-to-s3-exporter.ts
```diff
@@ -102,20 +114,15 @@ export class SfEmailsToS3Exporter extends SrStack {
...
-			new PolicyStatement({
-				sid: 'readDeployedArtefact',
-				effect: Effect.ALLOW,
-				actions: ['s3:GetObject'],
-				resources: ['arn:aws:s3::*:membership-dist/*'],
-			}),
-		);
+		// Reading the deployed artefact from the distribution bucket is granted
+		// automatically by @guardian/cdk's GuLambdaFunction, so no extra policy
+		// is needed here.
```
johnduffell, line 119: I don't think we need this comment

johnduffell:
thanks for the updates, just one comment added as I think you missed one of my suggestions

--

Also, to help out reviewers being able to re-review changes, please can you avoid amending existing commits once a review has been done?   The reason is just it breaks the direct "view changes since last review" direct link and all the changes appear as "new" even if they are actually existing changes
image

In this case it's not  too hard as there are only two commits, but it makes things smoother especially on longer running PRs!

You can do merge commits instead to get up to date with main, and then once you "squash and merge" everything gets folded down cleanly into main regardless of the PR.  Let me know if I missed anything 🙏

PR #3649, opened 2026-06-22
Add a buffer to bearer token expiry check

rupertbates:
## What does this change?
We are still getting auth errors from Zuora in the process supporter item lambda. This PR adds a buffer to avoid any millisecond delays from taking us over the expiry limit as well as some slightly more verbose logging.

### DIFF
--- a/modules/zuora/src/auth/bearerTokenProvider.ts
+++ b/modules/zuora/src/auth/bearerTokenProvider.ts
@@ -14,9 +14,9 @@ export interface BearerTokenProvider {
 
 export class ZuoraBearerTokenProvider implements BearerTokenProvider {
 	private bearerToken: ZuoraBearerToken | null = null;
-	private lastFetchedTime: Date | null = null;
-	// avoid failures caused by millisecond timing issues
-	private bufferInMilliseconds = 1000;
+	private tokenExpiryTime: number | null = null;
+	// 5 minute buffer in case of early expiry
+	private bufferInMilliseconds = 5 * 60 * 1000;
 
 	constructor(
 		private stage: string,
@@ -25,39 +25,30 @@ export class ZuoraBearerTokenProvider implements BearerTokenProvider {
 
 	private tokenIsExpired = () => {
 		logger.log('Checking if Zuora bearer token is expired');
-		if (this.bearerToken === null) {
-			logger.log('No Zuora bearer token found, fetching a new one');
-			return true;
-		}
-		const now = new Date();
-		if (this.lastFetchedTime === null) {
+		if (this.tokenExpiryTime === null) {
 			logger.log(
-				'No last fetched time found, fetching a new Zuora bearer token',
+				'No token expiry time found, fetching a new Zuora bearer token',
 			);
 			return true;
 		}
-		if (
-			this.lastFetchedTime.getTime() + this.bearerToken.expires_in * 1000 <
-			now.getTime() + this.bufferInMilliseconds
-		) {
+		const now = new Date();
+		if (this.tokenExpiryTime < now.getTime() + this.bufferInMilliseconds) {
 			logger.log('Zuora bearer token is expired', {
-				lastFetchedTime: this.lastFetchedTime.getTime(),
-				expiresIn: this.bearerToken.expires_in * 1000,
+				tokenExpiryTime: this.tokenExpiryTime,
 				now: now.getTime(),
 			});
 			return true;
 		}
 		logger.log('Zuora bearer token is still valid', {
-			lastFetchedTime: this.lastFetchedTime.getTime(),
-			expiresIn: this.bearerToken.expires_in * 1000,
+			tokenExpiryTime: this.tokenExpiryTime,
 			now: now.getTime(),
 		});
 		return false;
 	};
 	public async getBearerToken(): Promise<ZuoraBearerToken> {
 		if (this.bearerToken === null || this.tokenIsExpired()) {
-			this.lastFetchedTime = new Date();
 			this.bearerToken = await this.fetchZuoraBearerToken();
+			this.tokenExpiryTime = Date.now() + this.bearerToken.expires_in * 1000;
 		}
 		return this.bearerToken;
 	}
--- a/modules/zuora/src/auth/bearerTokenProvider.ts
+++ b/modules/zuora/src/auth/bearerTokenProvider.ts
@@ -1,5 +1,5 @@
 import { logger } from '@modules/logger/logger';
-import type { OAuthClientCredentials, ZuoraBearerToken } from '../types';
+import type { OAuthClientCredentials } from '../types';
 import { zuoraBearerTokenSchema } from '../types';
 import { zuoraServerUrl } from '../utils';
 
@@ -12,11 +12,15 @@ export interface BearerTokenProvider {
 	getAuthorisation(): Promise<Authorisation>;
 }
 
+type BearerToken = {
+	bearerToken: string;
+	tokenExpiryTime: number;
+};
+
 export class ZuoraBearerTokenProvider implements BearerTokenProvider {
-	private bearerToken: ZuoraBearerToken | null = null;
-	private tokenExpiryTime: number | null = null;
 	// 1 minute buffer in case of early expiry
 	private readonly bufferInMilliseconds = 60 * 1000;
+	private bearerToken: BearerToken | null = null;
 
 	constructor(
 		private stage: string,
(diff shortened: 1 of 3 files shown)


modules/zuora/src/auth/bearerTokenProvider.ts
```diff
@@ -36,7 +38,7 @@ export class ZuoraBearerTokenProvider implements BearerTokenProvider {
 		}
 		if (
 			this.lastFetchedTime.getTime() + this.bearerToken.expires_in * 1000 <
-			now.getTime()
+			now.getTime() + this.bufferInMilliseconds
```
tjmw, line 41: Should this be:

```suggestion
			now.getTime() - this.bufferInMilliseconds
```

If we want to consider the token expired slightly sooner than it actually is expired?

johnduffell: I think it's actually right already ("pretend it's a second later than it actually is"), but it may be worth extracting a const for clarity
```ts
    const expectedExpiry = this.lastFetchedTime.getTime() + this.bearerToken.expires_in * 1000;
    const conservativeExpiry = expectedExpiry - this.bufferInMilliseconds
		if (conservativeExpiry < now.getTime()) {
```

tjmw: Ah yes you're right! My brain clearly not quite engaged on Monday morning 😄

johnduffell:
not sure this will work due to this async function not being synchronized
```
	public async getBearerToken(): Promise<ZuoraBearerToken> {
		if (this.bearerToken === null || this.tokenIsExpired()) {
			this.lastFetchedTime = new Date();
			this.bearerToken = await this.fetchZuoraBearerToken();
		}
		return this.bearerToken;
	}
```

1. existing (expired token)
3. update lastFetched time
4. call in progress, JS yields to other threads
5. other calls come in and find the (not yet fetched) token is not expired so go ahead and use it (edit: the existing token that's still there)
6. Errors..
7. new token comes back with fresh token
8. everything goes back to normal for now

modules/zuora/src/auth/bearerTokenProvider.ts
rupertbates: No I don't think so, it's saying 'in one second's time are we going to be past the expiry point'

johnduffell:
in case it helps, Tom F and I had a lot of fun pairing on the same problem in support-frontend with mparticle tokens.  We additionally had to handle the fact that tokens can actually be invalid well before their declared expiry
https://github.com/guardian/support-frontend/pull/7588

rupertbates:
> not sure this will work due to this async function not being synchronized
> 
> ```
> 	public async getBearerToken(): Promise<ZuoraBearerToken> {
> 		if (this.bearerToken === null || this.tokenIsExpired()) {
> 			this.lastFetchedTime = new Date();
> 			this.bearerToken = await this.fetchZuoraBearerToken();
> 		}
> 		return this.bearerToken;
> 	}
> ```
> 
>     1. existing (expired token)
> 
>     2. update lastFetched time
> 
>     3. call in progress, JS yields to other threads
> 
>     4. other calls come in and find the (not yet fetched) token is not expired so go ahead and use it (edit: the existing token that's still there)
> 
>     5. Errors..
> 
>     6. new token comes back with fresh token
> 
>     7. everything goes back to normal for now

Good spot, I've changed the logic so we're just storing the expiry time and that is set from now + the expires_in value at the point the request completes. That should avoid the issue you describe. 

I've also upped the buffer to 1 minute, hopefully that will avoid any issues with early expiry of tokens because the mParticle approach is quite a bit more complicated.

johnduffell:
edit: didn't look carefully, let me think about it again

johnduffell:
having looked at the code there are also efficiency issues with the first auth, in that any concurrent requests would come along and overwrite the fetch timestamp and then redo the fetch call, so there could be a flurry of duplicated auth requests at the initial lambda start.

With the changes to put the timestamp in after the request, this would now also happen at token expiry - we will allow all concurrent requests through to do the auth call until the first auth call returns from zuora and updates the expiry time value.

modules/zuora/src/auth/bearerTokenProvider.ts
```diff
@@ -14,7 +14,9 @@ export interface BearerTokenProvider {
 
 export class ZuoraBearerTokenProvider implements BearerTokenProvider {
 	private bearerToken: ZuoraBearerToken | null = null;
-	private lastFetchedTime: Date | null = null;
+	private tokenExpiryTime: number | null = null;
+	// 1 minute buffer in case of early expiry
+	private bufferInMilliseconds = 60 * 1000;
```
johnduffell: I don't like the word buffer as it makes me think about FIFO buffers and things but my mind is blank on a good alternative
```suggestion
	// gives us a better chance of fetching a new token before the old one expires
	private authExpiryToleranceOneMinute = 60 * 1000;
```

johnduffell: also this can be a top level const rather than a class property, then it's clear that it's read only and not a dynamic class level value

modules/zuora/src/auth/bearerTokenProvider.ts
```diff
@@ -23,35 +25,30 @@ export class ZuoraBearerTokenProvider implements BearerTokenProvider {
...
+				'No token expiry time found, fetching a new Zuora bearer token',
 			);
 			return true;
 		}
-		if (
-			this.lastFetchedTime.getTime() + this.bearerToken.expires_in * 1000 <
-			now.getTime()
-		) {
+		const now = new Date();
+		if (this.tokenExpiryTime < now.getTime() + this.bufferInMilliseconds) {
```
johnduffell: for ease of getting your head around it, I'd still rather work out the time we want something to happen, and then compare it with now, rather than pretenting we are in the future and comparing with the expiry time

```suggestion
		const conservativeExpiryTime = this.tokenExpiryTime - this.bufferInMilliseconds;
		if (conservativeExpiryTime < now.getTime()) {
```

modules/zuora/src/auth/bearerTokenProvider.ts
```diff
@@ -23,35 +25,30 @@ export class ZuoraBearerTokenProvider implements BearerTokenProvider {
...
+			tokenExpiryTime: this.tokenExpiryTime,
+			now: now.getTime(),
+		});
 		return false;
 	};
 	public async getBearerToken(): Promise<ZuoraBearerToken> {
 		if (this.bearerToken === null || this.tokenIsExpired()) {
-			this.lastFetchedTime = new Date();
 			this.bearerToken = await this.fetchZuoraBearerToken();
+			this.tokenExpiryTime = Date.now() + this.bearerToken.expires_in * 1000;
```
johnduffell: probably would be covered by the 1 minute, but could be worth to be conservative.  Alternatively we could use the Date header from the response, but it's possible their clock may not be in sync with ours.
```suggestion
			this.tokenExpiryTime = now.getTime() + this.bearerToken.expires_in * 1000;
```

johnduffell: we already check the tokenExpiryTime to be not null inside tokenIsExpired so we shouldn't have to redo effectively the same check on line 49.  Rather than two nullable fields it would be cleaner to have one nullable property that is only checked one
```suggestion
		this.bearerToken = {
			token: await this.fetchZuoraBearerToken(),
			expiryTime: Date.now() + this.bearerToken.expires_in * 1000,
		}
```

Then in tokenIsExpired it can just do all the checks rather than them also partly happening in getBearerToken

modules/zuora/src/auth/bearerTokenProvider.ts
rupertbates: I'd rather leave it in the class, I've made it readonly though

modules/zuora/src/auth/bearerTokenProvider.ts
```diff
@@ -12,51 +12,55 @@ export interface BearerTokenProvider {
...
-		if (this.bearerToken === null || this.tokenIsExpired()) {
-			this.lastFetchedTime = new Date();
-			this.bearerToken = await this.fetchZuoraBearerToken();
+
+	public async getBearerToken(): Promise<BearerToken> {
+		if (this.tokenIsNullOrExpired()) {
+			this.bearerToken = await this.fetchBearerToken();
 		}
-		return this.bearerToken;
+		return this.bearerToken!; //Checked in tokenIsNullOrExpired
```
tjmw, line 60: Is this comment accurate? We might have just regenerated in `this.fetchBearerToken()`.

tjmw (approved):

modules/zuora/src/auth/bearerTokenProvider.ts
rupertbates: Yes it is accurate

johnduffell: you could add a user defined type guard to tokenIsNullOrExpired to avoid the cast here although it might need the logic flipping as it operates off "true" not "false"

See https://www.w3schools.com/typescript/typescript_type_guards.php#:~:text=Try%20it%20Yourself%20%C2%BB-,User%2DDefined%20Type%20Guards,-For%20more%20complex

I'm not sure if they are any safer though, it might just be centralising the cast in effect

PR #3651, opened 2026-06-22
Add Canadian tax-exclusive plans to the new-product-api product list

andrade0:
## What does this change?

The product catalog that new-product-api returns to Salesforce now includes the Canadian tax-exclusive variants of Digital Plus and All Access Digital (Supporter Plus), so CSRs can see and pick them when setting up an acquisition. Each new plan only carries a CAD payment plan since these are Canada-only. The four new plan ids are `digipack_annual_tax_exclusive`, `digipack_monthly_tax_exclusive`, `annual_supporter_plus_tax_exclusive` and `monthly_supporter_plus_tax_exclusive`, wired to their Zuora rate plan ids for both CODE and PROD.

Each new plan also carries a new plan-level `enabledForBillingCountries` field set to `[ "Canada" ]`, which Graham asked for. It mirrors the existing product-level `enabledForDeliveryCountries`, but on the plan rather than the product. It's an Option, so it only appears on these four plans and stays absent on every other plan.

This is the first of two PRs for the CA sales tax CSR work. A follow-up will add a `taxMode` field to paymentPlans so Salesforce can tell which plans are tax-exclusive. The tax amount itself stays out of the catalog by design, Salesforce works it out per province using the existing sales-tax-api.

Unit tested. The catalog serialisation snapshot test was updated and the full new-product-api suite passes locally. Nothing deployed to CODE yet. The new plans are catalog-only and aren't meant to be created through this API's add-subscription endpoint, since acquisition happens via Zuora Quotes in Salesforce; the charge lookups there are Option-based, so an unsupported plan returns an error rather than failing.

Salesforce can read the new plans from the catalog, see they're billable in Canada, and CSRs can select the Canadian tax-exclusive plans. Once the follow-up taxMode change lands they'll be able to show the right tax messaging too.

Low risk, the change is additive. Existing plans and pricing are untouched, and the new plans only appear with CAD pricing and a Canada billing restriction. They do become valid plan names for the add-subscription endpoint, but that path isn't used for them and fails safely if it ever is.

### DIFF
--- a/handlers/new-product-api/src/main/scala/com/gu/newproduct/api/productcatalog/plans/DigitalPackPlans.scala
+++ b/handlers/new-product-api/src/main/scala/com/gu/newproduct/api/productcatalog/plans/DigitalPackPlans.scala
@@ -1,6 +1,11 @@
 package com.gu.newproduct.api.productcatalog.plans
 
-import com.gu.newproduct.api.productcatalog.PlanId.{DigipackAnnual, DigipackMonthly}
+import com.gu.newproduct.api.productcatalog.PlanId.{
+  DigipackAnnual,
+  DigipackAnnualTaxExclusive,
+  DigipackMonthly,
+  DigipackMonthlyTaxExclusive,
+}
 import com.gu.newproduct.api.productcatalog._
 
 import java.time.LocalDate
@@ -19,6 +24,8 @@ class DigitalPackPlans(today: LocalDate) {
   val planInfo: List[(PlanId, PlanDescription, StartDateRules, BillingPeriod)] = List(
     (DigipackAnnual, PlanDescription("Annual"), startRules, Annual),
     (DigipackMonthly, PlanDescription("Monthly"), startRules, Monthly),
+    (DigipackAnnualTaxExclusive, PlanDescription("Annual"), startRules, Annual),
+    (DigipackMonthlyTaxExclusive, PlanDescription("Monthly"), startRules, Monthly),
   )
 
 }
--- a/handlers/new-product-api/src/main/scala/com/gu/newproduct/api/productcatalog/plans/DigitalPackPlans.scala
+++ b/handlers/new-product-api/src/main/scala/com/gu/newproduct/api/productcatalog/plans/DigitalPackPlans.scala
@@ -24,8 +24,8 @@ class DigitalPackPlans(today: LocalDate) {
   val planInfo: List[(PlanId, PlanDescription, StartDateRules, BillingPeriod)] = List(
     (DigipackAnnual, PlanDescription("Annual"), startRules, Annual),
     (DigipackMonthly, PlanDescription("Monthly"), startRules, Monthly),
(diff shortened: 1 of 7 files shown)


shtukas (approved):

rupertbates:
Could you change the title of this PR to 'Add Canadian tax-exclusive plans to the new-product-api product list' please? The product catalog is a separate project so I'd like to avoid confusion 🙏

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/productcatalog/plans/DigitalPackPlans.scala
```diff
@@ -19,6 +24,8 @@ class DigitalPackPlans(today: LocalDate) {
   val planInfo: List[(PlanId, PlanDescription, StartDateRules, BillingPeriod)] = List(
     (DigipackAnnual, PlanDescription("Annual"), startRules, Annual),
     (DigipackMonthly, PlanDescription("Monthly"), startRules, Monthly),
+    (DigipackAnnualTaxExclusive, PlanDescription("Annual"), startRules, Annual),
+    (DigipackMonthlyTaxExclusive, PlanDescription("Monthly"), startRules, Monthly),
```
graham228221: Could we be a bit more explicit here just to help us out on the UI side: 

```suggestion
    (DigipackAnnualTaxExclusive, PlanDescription("Annual (Tax Exclusive)"), startRules, Annual),
    (DigipackMonthlyTaxExclusive, PlanDescription("Monthly Tax Exclusive)"), startRules, Monthly),
```

(note that this isn't customer facing, it's just something that our Contact Centre agents will see)

handlers/new-product-api/src/main/scala/com/gu/newproduct/api/productcatalog/plans/SupporterPlusPlans.scala
```diff
@@ -17,6 +22,8 @@ class SupporterPlusPlans(today: LocalDate) {
   val planInfo: List[(PlanId, PlanDescription, StartDateRules, BillingPeriod)] = List(
     (MonthlySupporterPlus, PlanDescription("Monthly"), rule, Monthly),
     (AnnualSupporterPlus, PlanDescription("Annual"), rule, Annual),
+    (MonthlySupporterPlusTaxExclusive, PlanDescription("Monthly"), rule, Monthly),
+    (AnnualSupporterPlusTaxExclusive, PlanDescription("Annual"), rule, Annual),
```
graham228221: Same suggestion as the Digital Pack plans:

```suggestion
    (MonthlySupporterPlusTaxExclusive, PlanDescription("Monthly (Tax Exclusive)"), rule, Monthly),
    (AnnualSupporterPlusTaxExclusive, PlanDescription("Annual (Tax Exclusive)"), rule, Annual),
```

graham228221 (approved):
Looks good and working in CODE, just need to update the `PlanDescription` for these (they are the plan names that appear in SF for users).

PR #3652, opened 2026-06-23
add t2->t3 upsell discount for PROD

johnduffell:
card https://app.asana.com/1/1210045093164357/project/1213134855566811/task/1214357149767966

We have previously done the work to add the upsell discount switch for CODE.
see main PR https://github.com/guardian/support-service-lambdas/pull/3547

Now we have the discount duration confirmed, and I have added the actual discount in PROD zuora.

This PR adds the suitable IDs and updates the duration accordingly.

It is still disabled in PROD, but the values are there ready.

### DIFF
--- a/handlers/product-switch-api/src/changePlan/action/preview.ts
+++ b/handlers/product-switch-api/src/changePlan/action/preview.ts
@@ -31,7 +31,7 @@ export interface SwitchDiscountResponse {
 	discountedPrice: number;
 	upToPeriods: number;
 	upToPeriodsType: 'Months' | 'Years';
-	discountPercentage: number;
+	discountPercentage?: number;
 }
 
 export const isRefundExpected = (
--- a/handlers/product-switch-api/src/changePlan/switchDefinition/discounts.ts
+++ b/handlers/product-switch-api/src/changePlan/switchDefinition/discounts.ts
@@ -8,10 +8,10 @@ export type Discount = {
 	name: string;
 	upToPeriods: number;
 	upToPeriodsType: 'Months' | 'Years';
-	discountPercentage: number;
+	discountPercentage?: number;
 };
 
-export const annualContribHalfPriceSupporterPlusForOneYear: Discount = {
+export const annualContribHalfPriceSupporterPlusForOneYear = {
 	productRatePlanId: {
 		PROD: '8a12994695aa4f680195ae2dc9d221d8',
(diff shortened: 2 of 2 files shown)


handlers/product-switch-api/src/changePlan/action/preview.ts
```diff
@@ -31,7 +31,7 @@ export interface SwitchDiscountResponse {
 	discountedPrice: number;
 	upToPeriods: number;
 	upToPeriodsType: 'Months' | 'Years';
-	discountPercentage: number;
+	discountPercentage?: number;
```
johnduffell, line 34: we now only return a percentage where it makes sense (i.e. when there is a fixed percentage for all currencies)
Where we have a specific price per currency, it doesn't make sense.
If we do need it, we can always use the currency from their subscription.

handlers/product-switch-api/src/changePlan/switchDefinition/discounts.ts
```diff
@@ -24,19 +24,18 @@ export const annualContribHalfPriceSupporterPlusForOneYear: Discount = {
 	upToPeriods: 1,
 	upToPeriodsType: 'Years',
 	discountPercentage: 50,
-};
+} satisfies Discount;
```
enesihsa, line 27: What is the difference between using `: Discount` vs `satisfies Discount`?

enesihsa (approved):

handlers/product-switch-api/src/changePlan/switchDefinition/discounts.ts
johnduffell: great question!  `: Discount` checks the type is compatible and then makes it forget the exact type of the value and treats it as only Discount, whereas `satisfies` only does the compatibility check.
The issue here is that I've made `discountPercentage` optional in the Discount type, but it's actually present here in `annualContribHalfPriceSupporterPlusForOneYear`.
This causes us to have to handle the fact that it might be undefined even when we reference it directly
image

If we keep the underlying shape, typescript knows the full type that for that particular one it's actually defined.
image

This does get more verbose, and it loses the ability to reference what uses discount directly, but it can be convenient in cases like this where we deal with discounts both generically and specifically.

enesihsa: Ah I see, thank you very much for the detailed explanation!

PR #3633, opened 2026-06-16
Update supporter product data to handle secondary users

rupertbates:
## What does this change?
This PR is the first part of the changes required to the supporter product data lambdas to support the multiple accounts feature. Requirements are outlined in more detail in [this document](https://docs.google.com/document/d/1eqMn9zkg-iRK3Gi9ecmgGh8WWUYpLX6B3cOasTRYpns/edit?tab=t.0)

Specifically this PR should handle updating the term end date and the expiry date (Dynamo TTL) of any secondary subscription records in the SupporterProductData Dynamo table whenever there is a change to the primary subscription they derive from.

## Changes
- Move secondary users repository and test to a module so that it can be used both by multiple-accounts-api and supporter-product-data-lambdas
- The processSupporterRatePlanItem lambda now checks in the secondary user Dynamo table for every supporter rate plan item it receives. If it finds one or more secondary user it upserts a secondary subscription record into the supporter product data Dynamo table for that user (worth noting that it does this by adding a record to the SQS queue, in effect calling itself recursively). 
- Add an Infrastructure diagram and flowchart for the process item lambda to the read me. I found this helpful while working through what needs doing.

## How has this change been tested?
With integration tests, unit tests and by running in CODE

#### Process Item Flowchart
process_item_flowchart

#### Infrastructure Diagram for supporter-product-data-lambdas
infrastructure_diagram

PR #3654, opened 2026-06-23
Set the expiryDate TTL of secondary user records from supporter product data

rupertbates:
## What does this change?
This PR is the second part of the changes to the supporter-product-data-lambdas project to support multiple accounts. This part deals with the lifecycle of secondary user records.

[The first part of this work is here](https://github.com/guardian/support-service-lambdas/pull/3633)

[The document describing this work is here](https://docs.google.com/document/d/1eqMn9zkg-iRK3Gi9ecmgGh8WWUYpLX6B3cOasTRYpns/edit?tab=t.0)

## Details 
Secondary user records are created by the accept invitation endpoint in multiple-accounts-api. When they are first created their expiryDate field (which is a Dynamo TTL recording when the record should be automatically deleted by Dynamo) is set to the term end date + 2 weeks of the primary subscription to which they belong.

Whenever the term end date of the primary subscription changes, either because the subscription has renewed or because it has been cancelled, we need to adjust the expiryDate field of the secondary user to match. That is what this PR does.

#### Flowchart
image

#### Infrastructure diagram
image

Unit tests, integration tests and running on CODE


maloky (approved):
Nice work 👍

handlers/supporter-product-data-lambdas/test/processSupporterRatePlanItemLambda.test.ts
```diff
@@ -130,5 +141,15 @@ describe('processSupporterRatePlanItemLambda', () => {
...
+		expect(updateSecondaryUserTTL).toHaveBeenCalledWith(
+			expect.stringContaining('sub-1'),
+			expect.stringContaining('secondary-id-1'),
+			secondaryUserTTLFromPrimarySubscriptionTTL(dayjs(termEndDate)),
+		);
+		expect(updateSecondaryUserTTL).toHaveBeenCalledWith(
+			expect.stringContaining('sub-1'),
+			expect.stringContaining('secondary-id-1'),
+			secondaryUserTTLFromPrimarySubscriptionTTL(dayjs(termEndDate)),
+		);
```
tjmw, line 153: Is this assertion duplicated?

handlers/supporter-product-data-lambdas/src/handlers/processSupporterRatePlanItem.ts
```diff
@@ -48,9 +54,19 @@ export const processItem = async (
 			subscriptionName: item.subscriptionName,
 		});
 		await Promise.all(
-			secondaryUsers.map((secondaryUser) =>
-				dependencies.writeSecondaryItem(itemWithContribution, secondaryUser),
-			),
+			secondaryUsers.map(async (secondaryUser) => {
+				await dependencies.writeSecondaryItem(
```
tjmw, line 58: I think this does an upsert? Could updating the TTL be done in the same Dynamo query? Or does it need to happen separately?

tjmw (approved):
Looks good, just a couple of questions.

handlers/supporter-product-data-lambdas/src/handlers/processSupporterRatePlanItem.ts
rupertbates: They are two separate tables, writeSecondaryItem writes the supporter product data table, updateSecondaryUserTTL updates the secondary user table.

PR #3655, opened 2026-06-24
Add the secondary user's email as a field in an invite

rupertbates:
## What does this change?
This PR is part of the work to build the multiple accounts feature [described here](https://docs.google.com/document/d/1eqMn9zkg-iRK3Gi9ecmgGh8WWUYpLX6B3cOasTRYpns/edit?tab=t.0)

When displaying a list of invites for a primary user we always need the email address that they have been sent to. Previously we didn't store this in the multiple-account-invitation Dynamo table which meant that we would always have to look it up with Identity whenever we wanted to display it to a user. 

This is very inefficient so this PR adds the email as a field on the Dynamo record. As invitations are short lived, the risk of a user changing their email address in that time is minimal and even if they do, any logic we have around invitation acceptance is driven by the identity ID, not the email address so will be unaffected.


tjmw (approved):
👍

PR #3657, opened 2026-06-24
fix type inference in object-query builder

johnduffell:
I noticed that when mapping over a subscription returned from object query, all the expands were available when using dot completion, but when mapping, none were available.

This PR tightens up the type inference of `pickKeys` so it doesn't widen the array when it shouldn't, it prefers to keep it as a const array.

### DIFF
--- a/modules/zuora/src/objectQuery/objectQueryBuilder.ts
+++ b/modules/zuora/src/objectQuery/objectQueryBuilder.ts
@@ -1,6 +1,6 @@
 import { mergeValues, pickKeys } from '@modules/objectFunctions';
-import type { ZodType, ZodTypeDef } from 'zod';
 import { z } from 'zod';
+import type { ZodType, ZodTypeDef } from 'zod';
 import type { DoesNotHaveDuplicateResponseKey } from '@modules/zuora/objectQuery/doesNotHaveDuplicateResponseKey';
 import type {
 	ObjectQueryExpandRegistry,
@@ -57,7 +57,7 @@ export class ObjectQueryBuilder<
 	 * @param expand
 	 */
 	buildItemSchema<
-		TFields extends keyof TFieldRegistry,
+		const TFields extends keyof TFieldRegistry,
 		const TExpandKeys extends ReadonlyArray<keyof TExpandRegistry>,
 	>(
 		fields: readonly TFields[],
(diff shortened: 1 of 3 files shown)


modules/zuora/test/objectQuery/buildSchema.test.ts
```diff
@@ -29,6 +29,11 @@ test('single item deserialiser', () => {
 	expect(actual.subscriptions[0]?.ratePlans[0]?.name).toEqual(
 		'Monthly Contribution',
 	);
+	// if the types aren't narrowed properly then it doesn't find the type in map as it may not exist on all
+	expect(actual.subscriptions.map((sub) => sub.ratePlans[0]?.name)).toEqual([
+		'Monthly Contribution',
+		'Monthly Contribution',
+	]);
```
johnduffell, line 36: although line 29 worked fine, this line was failing before the fix was done

```
  ● Test suite failed to run

    test/objectQuery/buildSchema.test.ts:32:47 - error TS2339: Property 'ratePlans' does not exist on type '{ status: "Active" | "Expired" | "Cancelled" | "Draft" | "Pending Activation" | "Pending Acceptance" | "Suspended"; id: string; createdById: string; createdDate: Date; updatedById: string; ... 36 more ...; orderId: string | null; }'.

    32  expect(actual.subscriptions.map((sub) => sub.ratePlans[0]?.name)).toEqual([
                                                     ~~~~~~~~~

```

modules/zuora/src/objectQuery/objectQueryBuilder.ts
```diff
@@ -1,6 +1,6 @@
 import { mergeValues, pickKeys } from '@modules/objectFunctions';
-import type { ZodType, ZodTypeDef } from 'zod';
 import { z } from 'zod';
+import type { ZodType, ZodTypeDef } from 'zod';
```
johnduffell, line 3: there's some disagreement between intellij and eslint on this, they keep getting swapped back and forth and failing builds.  Not sure if there's a version mismatch or something

modules/zuora/src/objectQuery/objectQueryBuilder.ts
```diff
@@ -57,7 +57,7 @@ export class ObjectQueryBuilder<
 	 * @param expand
 	 */
 	buildItemSchema<
-		TFields extends keyof TFieldRegistry,
+		const TFields extends keyof TFieldRegistry,
```
johnduffell, line 60: these two weren't needed to fix the issue but I added them for completeness

shtukas (approved):

PR #3659, opened 2026-06-25
Upgrade Zod to v4

rupertbates:
## What does this change?
This PR upgrades Zod from v3 to v4 using this [migration guide](https://zod.dev/v4/changelog) to gain the benefits mentioned in the [release announcement](https://zod.dev/v4) but particularly the [improved enum support](https://zod.dev/v4/changelog?id=improves-enum-support#improves-enum-support)

### DIFF
--- a/handlers/mparticle-api/src/routers/http/dataSubjectRequestCallback/data-subject-request-callback.ts
+++ b/handlers/mparticle-api/src/routers/http/dataSubjectRequestCallback/data-subject-request-callback.ts
@@ -29,6 +29,7 @@ export const dataSubjectRequestCallbackParser = {
 		results_url: z.string().url().nullable(),
 		extensions: z
 			.record(
+				z.string(),
 				z.object({
 					domain: z.string(),
 					name: z.string(),
--- a/handlers/product-switch-api/src/changePlan/action/switch.ts
+++ b/handlers/product-switch-api/src/changePlan/action/switch.ts
@@ -1,3 +1,4 @@
+import { getIfDefined } from '@modules/nullAndUndefined';
 import type { Stage } from '@modules/stage';
 import { sendToSupporterProductData } from '@modules/supporter-product-data/supporterProductData';
 import type {
@@ -79,6 +80,9 @@ export class DoSwitchAction {
 				zuoraSwitchResponseWithIdsSchema,
 			);
 
-		return switchResponse.invoiceIds[0];
+		return getIfDefined(
+			switchResponse.invoiceIds[0],
+			'No invoice id returned from switch order',
+		);
 	};
 }
--- a/handlers/product-switch-api/test/changePlan/action/pendingAmendments.test.ts
+++ b/handlers/product-switch-api/test/changePlan/action/pendingAmendments.test.ts
@@ -19,7 +19,7 @@ import { getAccountInformation } from '../../../src/changePlan/prepare/accountIn
 import { SwitchOrderRequestBuilder } from '../../../src/changePlan/prepare/buildSwitchOrderRequest';
 import { getSubscriptionInformation } from '../../../src/changePlan/prepare/subscriptionInformation';
 import type { TargetInformation } from '../../../src/changePlan/prepare/targetInformation';
-import type { ZuoraSwitchResponse } from '../../../src/changePlan/schemas';
+import type { ZuoraSwitchResponseWithIds } from '../../../src/changePlan/schemas';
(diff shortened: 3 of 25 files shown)


modules/product-benefits/src/schemas.ts
```diff
@@ -20,7 +20,10 @@ const trialSchema = z.object({
 	androidOfferTag: z.string(),
 });
 export type Trial = z.infer<typeof trialSchema>;
-const trialInformationSchema = z.record(productBenefitListSchema, trialSchema);
+const trialInformationSchema = z.partialRecord(
```
rupertbates, line 23: Previously Zod records with a union type as the key were not exhaustive in the same way as a Typescript Record is. In v4 they are so this now needs to be a Partial record

handlers/mparticle-api/src/routers/http/dataSubjectRequestCallback/data-subject-request-callback.ts
```diff
@@ -29,6 +29,7 @@ export const dataSubjectRequestCallbackParser = {
 		results_url: z.string().url().nullable(),
 		extensions: z
 			.record(
+				z.string(),
```
rupertbates, line 32: All records need to specify the key type now

handlers/product-switch-api/test/changePlan/action/pendingAmendments.test.ts
```diff
@@ -153,7 +153,7 @@ describe('pendingAmendments, e.g. contribution amount changes, are dealt with co
 		mockZuoraClient.post.mockResolvedValueOnce({
 			success: true,
 			invoiceIds: ['invoice-id'],
-		} as ZuoraSwitchResponse);
+		} as ZuoraSwitchResponseWithIds);
```
rupertbates, line 156: This was wrong before

shtukas (approved):

handlers/product-switch-api/src/changePlan/action/switch.ts
```diff
@@ -79,6 +80,9 @@ export class DoSwitchAction {
 				zuoraSwitchResponseWithIdsSchema,
 			);
 
-		return switchResponse.invoiceIds[0];
+		return getIfDefined(
+			switchResponse.invoiceIds[0],
+			'No invoice id returned from switch order',
+		);
```
johnduffell, line 86: has this check moved from the parser to the business logic?

handlers/ticket-tailor-webhook/src/apiGatewayToSqsEvent.ts
```diff
@@ -1,9 +1,9 @@
 import { z } from 'zod';
 
 export const apiGatewayToSqsEventSchema = z.object({
-	pathParameters: z.record(z.string()),
```
tjmw, line 4: So in Zod 3 you could get away with just providing the type of the values for records? Was the key always assumed to be a string?

johnduffell (approved):
fab thanks for going through all that, and it's not a bad diff for a major version update of a library we use literally everywhere!

handlers/ticket-tailor-webhook/src/apiGatewayToSqsEvent.ts
rupertbates: Yes

tjmw (approved):
I can't claim to fully understand all the changes, but if the compiler and tests are happy then I am too!

handlers/ticket-tailor-webhook/src/apiGatewayToSqsEvent.ts
johnduffell: looks like it was assumed to be a string
```ts
  static create<Value extends ZodTypeAny>(valueType: Value, params?: RawCreateParams): ZodRecord<ZodString, Value>;
  static create<Keys extends KeySchema, Value extends ZodTypeAny>(
    keySchema: Keys,
    valueType: Value,
    params?: RawCreateParams
  ): ZodRecord<Keys, Value>;
```

handlers/product-switch-api/src/changePlan/action/switch.ts
rupertbates: Yes the semantics of nonempty have changed: https://zod.dev/v4/changelog?id=changes-nonempty-type#changes-nonempty-type it validates that the array is non empty on parsing, but the inferred type is still a plain TS array which can return undefined as far as TS is concerned

johnduffell: wouldn't it have been better to use the example from the doc then? 
`z.tuple([z.string()], z.string());`

PR #3660, opened 2026-06-25
Fix Day 0 email first payment date for Direct Debit

tjmw:
## What does this change?

Before in Day 0 emails for Direct Debit we used to quote the later of the first date from the payment schedule or today + 10 days. When we showed today + 10 days this was quoted as an absolute date but in reality this padding is added to reflect the fact that the Direct Debit mandate can take up to 10 days to setup. So a customer could be surprised by being charged sooner than this.

Now we always show the first date from the payment schedule. If this is < 10 days in the future then we add a disclaimer for Direct Debit: `(Direct Debit may be up to 10 days after this)`.

Unit tests updated.

Verified on CODE: 

Screenshot 2026-06-25 at 13 43 06

### DIFF
--- a/modules/email/src/dataFields/dayZero/paymentEmailFields.ts
+++ b/modules/email/src/dataFields/dayZero/paymentEmailFields.ts
@@ -74,19 +74,14 @@ export function getPaymentMethodFieldsSupporterPlus(
 	created: Dayjs,
 	mandateId?: string,
 ):
-	| {
-			payment_method: 'credit / debit card' | 'PayPal';
-			first_payment_date: string;
-	  }
+	| Record<string, never>
 	| {
 			payment_method: 'Direct Debit';
 			account_name: string;
 			account_number: string;
 			sort_code: string;
 			Mandate_ID: string;
-			first_payment_date: string;
 	  } {
-	const DIRECT_DEBIT_LEAD_TIME_DAYS = 10;
 	switch (paymentMethod.Type) {
 		case 'BankTransfer':
 			return {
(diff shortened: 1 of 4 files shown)


modules/email/src/dataFields/dayZero/paymentEmailFields.ts
```diff
@@ -42,20 +42,22 @@ export function getPaymentFields(
 	mandateId?: string,
 ): EmailPaymentFields {
 	switch (paymentMethod.Type) {
-		case 'BankTransfer':
+		case 'BankTransfer': {
+			const isFirstPaymentLessThan10DaysAway =
+				firstZuoraPaymentDate < today.add(DIRECT_DEBIT_LEAD_TIME_DAYS, 'day');
+			const directDebitDisclaimer = isFirstPaymentLessThan10DaysAway
+				? ' (Direct Debit may be up to 10 days after this)'
+				: '';
```
rupertbates: Maybe extract all this and line 58 out to a getFirstPaymentDate function?

modules/email/src/dataFields/dayZero/paymentEmailFields.ts
```diff
@@ -74,19 +76,14 @@ export function getPaymentMethodFieldsSupporterPlus(
 	created: Dayjs,
 	mandateId?: string,
 ):
-	| {
-			payment_method: 'credit / debit card' | 'PayPal';
-			first_payment_date: string;
-	  }
+	| Record<string, never>
```
rupertbates: Might be simpler to return undefined for everything except the DD one?

rupertbates (approved):
A couple of small suggestions but looks good 👍

modules/email/src/dataFields/dayZero/paymentEmailFields.ts
tjmw: Yeah good point. I did it this way so that the result can be spread into the final object in the caller without having to care what was in it, but it's not hard to handle that case.

PR #3673, opened 2026-06-30
Sales Tax API: Fix an issue with DigitalSubscription tax rates in prod

tjmw:
## What does this change?

Previously we were returning the first tax period with a matching tax code, but there can be multiple of these so we need to make sure we get the current one.

This caused an issue in prod where the first matching period didn't have any tax rates for Canada and this caused the API to return a 500.

I've added an integration test for `DigitalSubscription` and have run this against prod (the committed code talks to CODE).

### DIFF
--- a/handlers/sales-tax-api/src/taxRatesEndpoint.ts
+++ b/handlers/sales-tax-api/src/taxRatesEndpoint.ts
@@ -75,9 +75,6 @@ async function getZuoraSalesTaxRates(
 		zuoraTaxCode.id,
 		zuoraTaxPeriods.taxRatePeriods,
 	);
-	if (!zuoraTaxPeriod) {
-		throw new Error(`invalid period for productKey:${productKey}`);
-	}
 
 	const zuoraTaxRates = await getZuoraTaxRates(zuoraClient, zuoraTaxPeriod.id);
 	const cadZuoraTaxRates = extractZuoraTaxRatesForCountry(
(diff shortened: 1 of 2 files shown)


handlers/sales-tax-api/src/taxRatesEndpoint.ts
```diff
@@ -102,9 +102,22 @@ function getZuoraTaxPeriod(
...
+	const periodMatchesTaxCode = (zuoraTaxPeriod: ZuoraTaxPeriod) =>
+		zuoraTaxPeriod.taxCodeId === zuoraTaxCode;
+	const periodHasNoEndDate = (zuoraTaxPeriod: ZuoraTaxPeriod) =>
+		zuoraTaxPeriod.endDate === null;
+
+	const periodsForTaxCodeWithNoEndDate = zuoraTaxPeriods
+		.filter((p) => periodMatchesTaxCode(p))
+		.filter((p) => periodHasNoEndDate(p));
+
+	if (periodsForTaxCodeWithNoEndDate.length > 1) {
```
rupertbates, line 111: I think we check at the call site that this method returns a record and throw if it doesn't. We could do that here and remove the `| undefined` from the return value.

rupertbates (approved):
With one suggestion

handlers/sales-tax-api/src/taxRatesEndpoint.ts
tjmw: Good point. I can move that check here. I think I'll keep it as a separate check within the function as I want the message to be slightly different.

PR #3674, opened 2026-06-30
Move currency schema into currency.ts

rupertbates:
This was a test to see whether moving zod schemas into the same file as the enumerated values are defined in would swell the support-frontend bundle size. It confirmed that it does because there are some pages which use the `currency` type but didn't previously import Zod.

Closing the PR as it has confirmed that this is not the recommended way to structure code, and that we should keep the schemas separate from the types.

### DIFF
--- a/modules/internationalisation/src/currency.ts
+++ b/modules/internationalisation/src/currency.ts
@@ -1,4 +1,5 @@
 import { isInList } from '@modules/arrayFunctions';
+import z from 'zod';
 
 export const CurrencyValues = [
 	'GBP',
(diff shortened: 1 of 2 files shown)


rupertbates:
This was a test to see whether moving zod schemas into the same file as the enumerated values are defined in would swell the support-frontend bundle size. It confirmed that it does because there are some pages which use the `currency` type but didn't previously import Zod.

Closing the PR as it has confirmed that this is not the recommended way to structure code, and that we should keep the schemas separate from the types.

PR #3675, opened 2026-06-30
use zuora billing preview for product switch confirmation email

johnduffell:
## Product switching email at present

after a product switch, we send a confirmation email including amongst other things,
- the payment today e.g. £2.65 (this comes from the invoice and is always correct)
- the next payment amount and date e.g. £12 on 30th July 2026 (see below)
- the plan that they are on e.g. £18 per month (see below)

The logic for working out the next payment amount and date is a bit of a "cross fingers and guess" situation which works based on the plans that we currently offer.

For S+ the next payment amount and plan are worked out by looking at the next payment amount - this works because
- we only allow discounts when it's a "save" which is Annual only - and the next payment on annual is always back to the full price, as the discount is 12 months.
- for all others, the switch is at full price, so we can just look at the catalog price (or the newAmount/current amount where provided)

The same applied when we started offering D+ switches because we didn't offer a discount, so the second and subsequent payments were always the same.

For both products, the next payment date is always intended to be the same as the billing period, so the logic just adds a billing period to today and quotes that date.

## Issues with the above

We had a setup issue with the D+ catalog so people's second payment was in less than a year although they were still being charged correctly in the long term.  However since the next payment date was hard coded to be today + 1 year, we showed the wrong date in the email.

For the new monthly D+ discount, the second(until sixth) payment is a price match discount unique to each currency.  This means that we would need to retrieve the ongoing price from the catalog, and also calculate the discounted amount.  This is complicated by the fact that the discount is currency specific.  Once worked out, the next payment and full price would both have to be passed through in the switch information.

Since zuora is the source of truth for how much and when someone will be charged, it doesn't make sense for us to try to replicate that logic ourselves.  The usual way that we do this is with a preview.

## This PR

This PR changes it so that we use a zuora billing preview to find out what the next payment will be, and use the date from that in the confirmation email.  The subscription_rate logic from the Day0 acquisition confirmation email is reused to turn the billing preview into a human readable sentence.

See inline PR comments for more information about the changes.


handlers/product-switch-api/src/changePlan/action/createSwitchOrder.ts
```diff
@@ -0,0 +1,32 @@
+import { getIfDefined } from '@modules/nullAndUndefined';
+import type {
+	CreateOrderRequest,
+	OrderRequest,
+} from '@modules/zuora/orders/orderRequests';
+import type { ZuoraClient } from '@modules/zuora/zuoraClient';
+import { zuoraSwitchResponseWithIdsSchema } from '../schemas';
+
+export class CreateSwitchOrder {
```
johnduffell, line 9: this and getNextPayment directly below are extracted as simple typed functions to make things more testable, rather than trying to directly mock a low level zuoraClient post/get etc in the tests we can directly mock  the right thing, making the test shorter, clearer and more type safe, at the cost of having a bit of wiring in the main code.

I didn't add tests directly for these wrappers as I avoided putting much logic in, but if it's worth it we can add it.

handlers/product-switch-api/src/changePlan/action/productSwitchEmail.ts
```diff
@@ -1,11 +1,14 @@
...
 import { getCurrencyInfo } from '@modules/internationalisation/currency';
-import type dayjs from 'dayjs';
+import type { SimpleInvoiceTotal } from '@modules/zuora/billingPreview';
+import dayjs from 'dayjs';
 import type { PaymentMethodType } from '../prepare/accountInformation';
 import type { SwitchInformation } from '../prepare/switchInformation';
 
 export const buildEmailMessage = (
 	firstPaymentAmount: number,
+	paymentSchedule: SimpleInvoiceTotal[],
```
johnduffell, line 11: this is the important change - we use the zuora payment schedule rather than calculating it ourselves.
If it's undefined (i.e. the preview is empty) we explicitly put "unknown date" in the email, as they can check in MMA, although perhaps we should have an alarm for that case as it indicates something else is wrong.

shtukas (approved):

handlers/product-switch-api/src/changePlan/action/productSwitchEmail.ts
```diff
@@ -44,12 +48,15 @@ export const buildEmailMessage = (
 					first_name: firstName,
 					last_name: lastName,
 					currency: getCurrencyInfo(currency).extendedGlyph,
+					price: switchInformation.target.ongoingPrice.toFixed(2), // superseded by subscription_rate - to remove
```
johnduffell, line 51: these fields are still being sent to keep the email working, once the emails are both over on subscription_rate, they should be removed to remove any confusion later as they don't tell the full story in discount cases

handlers/product-switch-api/src/changePlan/action/productSwitchEmail.ts
```diff
@@ -35,14 +48,17 @@ export const buildEmailMessage = (
...
-						.add(billingPeriodMonths, 'month')
-						.format('DD MMMM YYYY'),
-					payment_frequency: frequency,
+					next_payment_amount:
+						paymentSchedule[0]?.total.toFixed(2) ?? 'unknown amount', // superseded by subscription_rate - to remove
+					date_of_next_payment:
+						nextPaymentDate !== undefined
+							? dayjs(nextPaymentDate).format('DD MMMM YYYY')
+							: 'unknown date',
+					payment_frequency: frequency, // superseded by subscription_rate - to remove
```
johnduffell, line 60: this is actually going to be kept as it's used further down the email

PR #3676, opened 2026-06-30
reduce the amount of unnecessary async in product-switch-api

johnduffell:
pre-refactoring for https://github.com/guardian/support-service-lambdas/pull/3675

I noticed there's a thread of async all through the "get switch information" code, presumably we needed it at some point, but not now.

This PR removes the async from the above, and also extract the "to filtered guardian subscription" logic to further reduce the async nested in the code.

There are several inline PR comments to call out anything interesting.

### DIFF
--- a/handlers/product-switch-api/src/changePlan/changePlanEndpoint.ts
+++ b/handlers/product-switch-api/src/changePlan/changePlanEndpoint.ts
@@ -1,11 +1,6 @@
 import { ValidationError } from '@modules/errors';
 import type { GuardianSubscription } from '@modules/guardian-subscription/getSinglePlanFlattenedSubscriptionOrThrow';
-import { getSinglePlanFlattenedSubscriptionOrThrow } from '@modules/guardian-subscription/getSinglePlanFlattenedSubscriptionOrThrow';
-import type { GuardianSubscriptionMultiPlan } from '@modules/guardian-subscription/guardianSubscriptionParser';
-import { GuardianSubscriptionParser } from '@modules/guardian-subscription/guardianSubscriptionParser';
-import { SubscriptionFilter } from '@modules/guardian-subscription/subscriptionFilter';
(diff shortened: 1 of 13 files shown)


handlers/product-switch-api/src/changePlan/switchDefinition/supporterPlusTargetInformation.ts
```diff
@@ -43,15 +43,14 @@ export const supporterPlusTargetInformation: SwitchTargetInformation<
 			const isEligible =
 				productRatePlan.billingPeriod === 'Annual' &&
 				switchActionData.previousAmount <= discountedPrice;
-			if (isEligible) {
-				discount = discountDetails;
-				contributionAmount = 0;
-				actualTotalPrice = discountedPrice;
-			} else {
+			if (!isEligible) {
```
johnduffell, line 46: slightly easier to follow if we flip the logic to bail early if they aren't eligible?

handlers/product-switch-api/src/changePlan/changePlanEndpoint.ts
```diff
@@ -132,32 +128,13 @@ export class ChangePlanEndpoint {
...
+	gatherSwitchData(zuoraSubscription: ZuoraSubscription) {
 		const subscription: GuardianSubscription =
-			getSinglePlanFlattenedSubscriptionOrThrow(
-				guardianSubscriptionCurrentPlans,
+			this.toSingleGuardianSubscription.getSubscription(
+				this.today,
+				zuoraSubscription,
 			);
 
-		logger.log('guardian subscription', subscription);
```
johnduffell, line 159: rather than loading all the catalog etc here, easier to just offload it all to the new `ToSingleGuardianSubscription` class, it can load them up front and wrap all the product-switch specific logic into its method.
Then this function stops being async and should be clearer.

shtukas (approved):

PR #3677, opened 2026-07-01
Make eslint work correctly in editors for module directories

rupertbates:
## What does this change?
I noticed that for some files my editor whether VS Code or Intellij was showing lint errors for import ordering which were not picked up by on the command line, and if I let it 'fix' them, the command line highlighted the change as an error. 

The particular example I investigated was `modules/internationalisation/src/schemas.ts` where the editor wanted
```typescript
import { isoCountries } from '@modules/internationalisation/country';
import { SupportRegionId } from '@modules/internationalisation/countryGroup';
import { CurrencyValues } from '@modules/internationalisation/currency';
import { z } from 'zod';
```
but the cli wanted
```typescript
import { z } from 'zod';
import { isoCountries } from '@modules/internationalisation/country';
import { SupportRegionId } from '@modules/internationalisation/countryGroup';
import { CurrencyValues } from '@modules/internationalisation/currency';
```
After a bit of investigation it turned out that the reason for this was that our import ordering specifies that external package imports should always come before internal ones, but the editor and the CLI had a different understanding of whether @modules packages were internal or external.

- The CLI (pnpm -r run lint) runs ESLint once per package, with the package folder as the cwd. 
From there, @modules/* (pnpm-symlinked workspace packages) resolve to the real repo paths and classifies them as internal → correct order is zod first.

- VS Code and IntelliJ run ESLint from the workspace root by default. From there the same imports resolve via the root node_modules symlink and get classified as external → they sort before zod (because @ < z, which is why the editor flags the correct order and "fixes" it to the wrong one.

## Fix
To solve this inconsistency I have added some config which specifies that any code in @modules is an internal dependency. This has been added in  in eslint.shared.mjs so that it can be imported and used from each of the three separate eslint.config.mjs files in the project.

## Note
As a result of this change there has been a large reordering of imports across the repo because previously any @modules import from the _same directory_ was counted as internal whereas all others were external. Now we are explicitly saying that they are all internal that has affected ordering. I have done this update as a separate commit, so to see the relevant changes just look at [the first commit](https://github.com/guardian/support-service-lambdas/pull/3677/changes/4f5ff5d85f9110ca2aca855c91c50d1d345e3b68). Everything else is just reordering.

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -1,10 +1,12 @@
 import guardian from '@guardian/eslint-config';
+import { importOrderConfig } from './eslint.shared.mjs';
 
 export default [
 	...guardian.configs.recommended,
 	...guardian.configs.jest,
 	...guardian.configs.react,
 	...guardian.configs.storybook,
+	importOrderConfig,
 	{
 		rules: {
 			'@typescript-eslint/consistent-type-assertions': [
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -4,8 +4,6 @@ import { importOrderConfig } from './eslint.shared.mjs';
 export default [
 	...guardian.configs.recommended,
 	...guardian.configs.jest,
-	...guardian.configs.react,
-	...guardian.configs.storybook,
 	importOrderConfig,
 	{
 		rules: {
--- a/handlers/stripe-disputes/eslint.config.mjs
+++ b/handlers/stripe-disputes/eslint.config.mjs
@@ -1,8 +1,10 @@
 import guardian from '@guardian/eslint-config';
 import { defineConfig } from 'eslint/config';
+import { importOrderConfig } from '../../eslint.shared.mjs';
 
 export default defineConfig([
 	guardian.configs.recommended,
+	importOrderConfig,
 	{
 		extends: [guardian.configs.recommended],
 		files: ['test/**/*.ts', '**/*.test.ts'],
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -18,3 +18,6 @@ c7065a7fd22fc65c30d5c71f7ce819e84133be1b
 
 # Scala Steward: Reformat with scalafmt 3.10.4
 038ad7cbb7bcd2353cce4ec04273df16f7b6dd07
+
+# Reformat imports after setting all @modules imports to 'internal' for sort purposes
+b5401250344c270d473481f628c26c71699b713f
--- a/eslint.shared.mjs
+++ b/eslint.shared.mjs
@@ -1,0 +1,43 @@
+// Shared ESLint override that makes import ordering deterministic across
+// working directories.
+//
+// Workspace packages (@modules/*) are pnpm symlinks, so import/order's
+// external-vs-internal classification otherwise depends on the directory ESLint
+// runs from (the package dir on the CLI vs the workspace root in editors like
+// VS Code / IntelliJ). That flips their position relative to real external
+// packages. Pinning @modules/** to the "internal" group makes the order the
+// same everywhere.
+//
+// Spread this after `guardian.configs.recommended` in every eslint.config.mjs
+// so package-level configs (which don't inherit the root config) stay in sync.
+export const importOrderConfig = {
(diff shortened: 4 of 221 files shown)


handlers/stripe-disputes/eslint.config.mjs
```diff
@@ -1,8 +1,10 @@
 import guardian from '@guardian/eslint-config';
 import { defineConfig } from 'eslint/config';
+import { importOrderConfig } from '../../eslint.shared.mjs';
```
tjmw, line 3: Could this (and the product-catalog one below) use the same base eslint config?

rupertbates: Now that we are using eslint flat config the way it works is a bit different, I asked co-pilot about it and it said: 

The short version: it's not that it's *impossible*, it's that flat config gives you no *automatic* inheritance, and importing the root config file wholesale is the wrong tool here.

## 1. Flat config has no directory cascade

The old ESLint system (`.eslintrc`) walked up the directory tree and **merged** every config it found — true hierarchical inheritance. Flat config (eslint.config.mjs) deliberately removed that. For any file being linted, ESLint uses **exactly one** config file: the nearest eslint.config.mjs walking upward. Once eslint.config.mjs exists, the root eslint.config.mjs is **completely ignored** for files in that package — there's no merge with parent configs. So there's nothing to "inherit from" automatically.

## 2. `extends` in flat config extends *objects*, not parent files

You can see `extends` used in eslint.config.mjs:

```js
{ extends: [guardian.configs.recommended], files: ['test/**/*.ts'] }
```

That works because `guardian.configs.recommended` is a config **object that was imported**. `extends` (and plain spreading) only operate on values you explicitly `import` into the file. There is no syntax that says "extend the eslint.config.mjs in my parent directory." You always have to import it by path.

## 3. You *could* import the root config — but it's the wrong thing to import

Technically a child could do `import root from eslint.config.mjs'` and spread it. The reasons not to:

- **The root's default export is the whole repo ruleset** — `guardian.configs.recommended` **plus** `jest`, `react`, `storybook`, and several rule overrides. The package configs exist precisely to be *leaner* or *different* (product-catalog only wants `recommended` + `sort-keys-fix`; stripe-disputes relaxes rules for tests). Spreading the root in would drag all those extra layers back, defeating the reason those separate configs exist.
- **Role conflict.** The root config's job is "the config for every package that doesn't have its own." A package that *does* have its own importing it back is conceptually backwards, and because flat config is **last-match-wins**, layering the full root array underneath could reorder or re-enable rules in surprising ways.
- **Coupling/fragility.** It hard-links each package to the entire root file via a relative path, so any root change silently changes every package.

## Why a shared fragment is the right answer

That's exactly why I put the one thing that genuinely *should* be identical everywhere — the deterministic import ordering — into a small, single-purpose module, eslint.shared.mjs, and each config imports just that piece:

```js
import { importOrderConfig } from '../../eslint.shared.mjs';
```

This *is* inheritance — explicit and granular, which is the idiomatic flat-config approach. Each config shares the common rule without being forced to adopt the root's full (and intentionally different) rule set.

tjmw (approved):
Looks good 👍 

What do you think about adding the one time reordering to `.git-blame-ignore-revs` to keep git blames to meaningful/functional changes?

eslint.config.mjs
```diff
@@ -1,10 +1,12 @@
 import guardian from '@guardian/eslint-config';
+import { importOrderConfig } from './eslint.shared.mjs';
 
 export default [
 	...guardian.configs.recommended,
 	...guardian.configs.jest,
 	...guardian.configs.react,
 	...guardian.configs.storybook,
```
maloky: just curious why are we applying these. Do we even use react and storybook in support-service-lambdas?

maloky (approved):
LGTM 👍

eslint.config.mjs
rupertbates: Good point, they're not needed so I've removed them 👍

johnduffell (commented):
oh thanks for fixing that, it's been annoying me but I never got around to looking into it 👏

PR #3680, opened 2026-07-01
add tests for product switch email sending

johnduffell:
refactoring needed before https://github.com/guardian/support-service-lambdas/pull/3675

At the moment there aren't tests for what the email sends.

This PR uses dependency injection to let us pass in the email send logic in a type safer way.

See inline comments for significant or unrelated changes.


handlers/product-switch-api/src/changePlan/action/createSwitchOrder.ts
```diff
@@ -0,0 +1,32 @@
+import { getIfDefined } from '@modules/nullAndUndefined';
```
johnduffell, line 1: this is extracted as a simple typed function to make things more testable, rather than trying to directly mock a low level zuoraClient post/get etc in the tests we can directly mock the right thing, making the test shorter, clearer and more type safe, at the cost of having a bit of wiring in the main code.

I didn't add tests directly for this wrapper as I avoided putting much logic in, but if it's worth it we can add it.

handlers/product-switch-api/test/changePlan/action/switch/previewAction.test.ts
```diff
@@ -0,0 +1,116 @@
+import { generateProductCatalog } from '@modules/product-catalog/generateProductCatalog';
```
johnduffell, line 1: this file was extracted from handlers/product-switch-api/test/changePlan/action/pendingAmendments.test.ts which was in turn deleted

handlers/product-switch-api/test/changePlan/action/switch/switchAction.test.ts
```diff
@@ -0,0 +1,118 @@
...
+		createSwitchOrder,
+		sendEmail,
+	);
+
+describe('DoSwitchAction', () => {
+	beforeEach(() => {
+		jest.clearAllMocks();
+	});
+
+	test('next payment in the email is assumed to be the catalog price, one billing period after today', async () => {
```
johnduffell, line 90: this is a new test for the confirmation email, this was untested before

handlers/product-switch-api/test/changePlan/action/removePendingUpdateAmendments.test.ts
```diff
@@ -0,0 +1,117 @@
+import type { ZuoraClient } from '@modules/zuora/zuoraClient';
```
johnduffell, line 1: this file is extracted from pendingAmendments.test.ts above
there is a comment in the above "might be worth checking that it actually removed the pending amendments - need to mock the getLatestAmendment call"
This addresses that issue.

If we used dependency injection we could mock a cleaner method rather than zuoraClient.delete, but I don't want to touch too much at this stage.

shtukas (approved):

PR #3681, opened 2026-07-01
rename actualTotalPrice to ongoingPrice

johnduffell:
refactoring needed before https://github.com/guardian/support-service-lambdas/pull/3675

At the moment, when switching, you get one payment at a reduced price, followed by all the rest at a standard long term price.

The reduced price is mostly because we reduce it by however much you have paid for the existing product and not yet used.  However it's also possible for there to be a discount also, if it's an annual S+ cancellation save.

This means that the second payment is always the same as the long term payment amount.

This is going to change when we allow discounting of monthly subscriptions for 6 months, as there will be the initial small payment, then 5 months of discounted payments, followed by the base/ongoing payment amount.

At the moment the code calls the subsequent payment amount `actualTotalPrice`

This PR renames it to `ongoingPrice` instead.  A later PR (in draft here https://github.com/guardian/support-service-lambdas/pull/3675 ) will add next payment price, which may or may not differ from the ongoing price.

This is the only change in this PR, it's purely mechanical.

### DIFF
--- a/handlers/product-switch-api/src/changePlan/switchDefinition/supporterPlusTargetInformation.ts
+++ b/handlers/product-switch-api/src/changePlan/switchDefinition/supporterPlusTargetInformation.ts
@@ -33,7 +33,7 @@ export const supporterPlusTargetInformation: SwitchTargetInformation<
 
 		let discount: Discount | undefined;
 		let contributionAmount: number;
-		let actualTotalPrice: number;
+		let ongoingPrice: number;
 		if (switchActionData.mode === 'save') {
 			const discountDetails = annualContribHalfPriceSupporterPlusForOneYear;
 			const discountedPrice =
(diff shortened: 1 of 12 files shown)


shtukas (approved):

handlers/product-switch-api/src/changePlan/switchDefinition/supporterPlusTargetInformation.ts
```diff
@@ -50,15 +50,15 @@ export const supporterPlusTargetInformation: SwitchTargetInformation<
 			}
 			discount = discountDetails;
 			contributionAmount = 0;
-			actualTotalPrice = discountedPrice;
+			ongoingPrice = discountedPrice;
```
johnduffell, line 53: now you can see the issue fixed in the next PR https://github.com/guardian/support-service-lambdas/pull/3675
The ongoing price can't be the discounted price, it must be the catalog price.  The discounted price could be passed through separately, but for accuracy it will actually come from a zuora billing preview

PR #3684, opened 2026-07-02
Internationalisation refactor: Move state into separate file

rupertbates:
## What does this change?
This PR is the first of a few which will refactor the various types we have in the internationalisation module.

## Changes
- Move the lists of states and provinces into a separate file `state.ts`
- Change the type names from `UsState`, `CaState` & `AuState` to `UsStateCode` etc. to make it clearer that they represent just the two letter code, not an object. This will be a standard across all the internationalisation types, some of which have a more important distinction between `aThingCode` and `aThing`.
- Define the state code types _before_ the object which maps them to names so that we can make that object more strongly typed eg. `Record<UsStateCode, string>`
- Fix up any code broken by the code move


handlers/sales-tax-api/src/taxRatesEndpoint.ts
```diff
@@ -137,12 +133,12 @@ function createCadStateTaxRates(
...
 ): TaxRatesResponse {
 	const stateCodes = caStateSchema.options;
-	const missingStateCodes: CaState[] = [];
+	const missingStateCodes: CaStateCode[] = [];
 
 	const taxCodesByState = stateCodes.reduce<Partial<TaxRatesResponse>>(
 		(
 			memo: Partial<TaxRatesResponse>,
-			stateCode: CaState,
+			stateCode: CaStateCode,
```
tjmw, line 141: 👍 to this naming

tjmw (approved):

PR #3687, opened 2026-07-03
Augment promos with catalog information

rupertbates:
## What does this change?
This PR add some code to augment the Promo model with product catalog information. This will make it easier to 
- find all promotions for a given product
- check whether a given promotion applies to a particular product and rate plan

This is the first step towards creating an API which makes this promotion information available to support-frontend.

## Changes
Adds a new `PromoWithCatalogInformation` type where the `appliesTo` field contains a `catalogRatePlans` field of type `PromoCatalogInformation[]`.
Example Json:

```json
{
  "name": "Test Promotion",
  "promoCode": "TEST123",
  "campaignCode": "campaign",
  "startTimestamp": "2024-09-25T23:00:00.000Z",
  "endTimestamp": "2099-11-05T23:59:59.000Z",
  "discount": {
    "amount": 25,
    "durationMonths": 3
  },
  "appliesTo": {
    "countries": [
      "GB"
    ],
    "productRatePlanIds": [
      "8ad08cbd8586721c01858804e3275376"
    ],
    "catalogRatePlans": [
      {
        "productRatePlanKey": "Monthly"
      }
    ]
  }
}
```

### DIFF
--- a/modules/promotions/src/v2/addCatalogInformationToPromo.ts
+++ b/modules/promotions/src/v2/addCatalogInformationToPromo.ts
@@ -1,3 +1,4 @@
+import { logger } from '@modules/logger/logger';
 import { getIfDefined } from '@modules/nullAndUndefined';
 import type {
 	ProductCatalogHelper,
@@ -40,11 +41,38 @@ export const addCatalogInformationToPromo = (
 	};
 };
 
+/**
+ * The result of resolving catalog information for a batch of promotions:
+ * `succeeded` contains the enriched promos and `failed` contains the original
+ * promos whose rate plan ids could not be resolved against the catalog.
+ */
+export type AddCatalogInformationResult = {
+	succeeded: PromoWithCatalogInformation[];
+	failed: Promo[];
+};
+
 export const addCatalogInformationToPromos = (
 	promos: Promo[],
 	catalogHelper: ProductCatalogHelper,
-): PromoWithCatalogInformation[] =>
-	promos.map((promo) => addCatalogInformationToPromo(promo, catalogHelper));
+): AddCatalogInformationResult => {
+	const succeeded: PromoWithCatalogInformation[] = [];
+	const failed: Promo[] = [];
+
+	for (const promo of promos) {
+		try {
+			succeeded.push(addCatalogInformationToPromo(promo, catalogHelper));
+		} catch (error) {
+			// Collect promos whose rate plan ids can't be resolved against the
+			// catalog rather than failing the whole batch.
+			logger.log(
+				`Failed to resolve catalog information for promotion ${promo.promoCode}: ${String(error)}`,
+			);
+			failed.push(promo);
+		}
+	}
+
+	return { succeeded, failed };
+};
 
 export const promoAppliesTo = (
 	promo: PromoWithCatalogInformation,
--- a/modules/promotions/src/v2/addCatalogInformationToPromo.ts
+++ b/modules/promotions/src/v2/addCatalogInformationToPromo.ts
@@ -21,10 +21,13 @@ export const addCatalogInformationToPromo = (
 				catalogHelper.findProductDetails(productRatePlanId),
 				`Promotion ${promo.promoCode} references product rate plan id ${productRatePlanId} which does not exist in the product catalog`,
 			);
-			return {
-				productKey: productDetails.zuoraProduct,
-				productRatePlanKey: productDetails.productRatePlan,
-			};
+			// validateOrThrow returns a correlated GuardianCatalogKeys
+			// (productKey constrained to its rate plan key) rather than the widened union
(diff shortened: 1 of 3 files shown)


shtukas (approved):

johnduffell:
I assume the before in the description is the same as the "after" except without the `catalogRatePlans`?
```
"catalogRatePlans": [
      {
        "productRatePlanKey": "Monthly"
      }
    ]
```

modules/promotions/src/v2/schema.ts
```diff
@@ -50,6 +56,28 @@ export const promoSchema = z.object({
 
 export type Promo = z.infer<typeof promoSchema>;
 
+/**
+ * A product catalog rate plan a promotion applies to, resolved from a raw
+ * Zuora `productRatePlanId` stored in Dynamo.
+ */
+export type PromoCatalogInformation = {
```
johnduffell: why "Promo"CatalogInformation?  This looks like a pretty standard product-catalog plan reference?  Maybe we already have a type for this?

modules/promotions/src/v2/addCatalogInformationToPromo.ts
```diff
@@ -0,0 +1,64 @@
...
+ */
+export const addCatalogInformationToPromo = (
+	promo: Promo,
+	catalogHelper: ProductCatalogHelper,
+): PromoWithCatalogInformation => {
+	const catalogRatePlans = promo.appliesTo.productRatePlanIds.map(
+		(productRatePlanId) => {
+			const productDetails = getIfDefined(
+				catalogHelper.findProductDetails(productRatePlanId),
+				`Promotion ${promo.promoCode} references product rate plan id ${productRatePlanId} which does not exist in the product catalog`,
```
johnduffell, line 23: I'm sure it's ok but do you think this is safe enough?  e.g. would it be possible to have a bad promo in draft state? and would that break all checkouts

modules/promotions/test/v2/addCatalogInformationToPromo.test.ts
```diff
@@ -0,0 +1,92 @@
...
+	findPromosForProduct,
+	promoAppliesTo,
+} from '../../src/v2/addCatalogInformationToPromo';
+
+const catalogHelper = new ProductCatalogHelper(
+	generateProductCatalog(zuoraCatalogSchema.parse(codeZuoraCatalog)),
+);
+
+// SupporterPlus / Monthly in the code catalog
+const supporterPlusMonthlyId = '8ad08cbd8586721c01858804e3275376';
```
johnduffell: we have the product catalog loaded on line 14, can we move the comment into the code by using something like
productCatalog.SupporterPlus.ratePlans.Monthly.id

modules/promotions/test/v2/addCatalogInformationToPromo.test.ts
```diff
@@ -0,0 +1,92 @@
...
+	it('returns only promos whose catalog keys include the queried pair', () => {
+		const promos = addCatalogInformationToPromos(
+			[buildPromo([supporterPlusMonthlyId]), buildPromo([])],
+			catalogHelper,
+		);
+
+		const matches = findPromosForProduct(promos, 'SupporterPlus', 'Monthly');
+
+		expect(matches).toHaveLength(1);
+		expect(promoAppliesTo(matches[0]!, 'SupporterPlus', 'Monthly')).toBe(true);
```
johnduffell, line 80: not sure if toHaveLength logs the content of the array, but if you can assert the content and length in one it reduces the chance of an unhelpful "2 was not equal to 1" type of test failure

modules/promotions/test/v2/addCatalogInformationToPromo.test.ts
```diff
@@ -0,0 +1,92 @@
...
+		).toThrow(
+			'Promotion TEST123 references product rate plan id not-a-real-id which does not exist in the product catalog',
+		);
+	});
+});
+
+describe('findPromosForProduct', () => {
+	it('returns only promos whose catalog keys include the queried pair', () => {
+		const promos = addCatalogInformationToPromos(
+			[buildPromo([supporterPlusMonthlyId]), buildPromo([])],
```
johnduffell, line 73: nitpick but could have included a supporter plus annual or Contirbution monthly here as well

johnduffell (approved):
good idea, nice small PR easy to review 👏 Mostly nit picks in the comments

modules/promotions/src/v2/addCatalogInformationToPromo.ts
rupertbates: I've tweaked it so that we return two arrays from `addCatalogInformationToPromos` one for promos which we have successfully augmented and one for those that failed. Then we can return the successful ones via the API and alarm on the failed ones.

PR #3695, opened 2026-07-07
remove switch email fields that aren't used any more

johnduffell:


johnduffell:
already here https://github.com/guardian/support-service-lambdas/pull/3683

PR #3697, opened 2026-07-08
Create /secondary-user/me endpoint

rupertbates:
## What does this change?
This PR is part of the API work for the multiple accounts feature. It is described in more detail [in this doc](https://docs.google.com/document/d/1eqMn9zkg-iRK3Gi9ecmgGh8WWUYpLX6B3cOasTRYpns/edit?tab=t.0)

This PR adds a new endpoint to the multiple-accounts-api to enable MMA to display any secondary user access a user has.

## Authentication 
Authentication is via an Okta JWT token in the Authorization header. This is to prevent enumeration of identity ids - the signed in user will only be able to retrieve a list for their own identity id.

## Testing
Testing this endpoint is a little more complicated than some others because it requires an Okta Authorization header. 

To get one of these follow the steps in [this README](https://github.com/guardian/identity/tree/main/docs/generating_tokens_locally) to generate a header in Postman or Hopscotch. 

You can then make your request to secondary users endpoint:
```
## Example request 

```shell
curl --location --request GET 'https://multiple-account-api-code.support.guardianapis.com/secondary-user/me' \
```

```json
{
    "primaryUsers": [
        {
            "firstName": "Test",
            "lastName": "User
            "workEmail": "test.user@thegulocal.com"
        }
    ]
}
```

---
  - https://app.asana.com/0/0/1215969403983007

### DIFF
--- a/handlers/multiple-account-api/src/index.ts
+++ b/handlers/multiple-account-api/src/index.ts
@@ -10,6 +10,7 @@ import { Router } from '@modules/routing/router';
 import { withMMAIdentityCheck } from '@modules/routing/withMMAIdentityCheck';
 import { withBodyParser, withPathParser } from '@modules/routing/withParsers';
 import { stageFromEnvironment } from '@modules/stage';
+import { ZuoraClient } from '@modules/zuora/zuoraClient';
 import { getZuoraCatalogFromS3 } from '@modules/zuora-catalog/S3';
 import {
 	acceptInvitationEndpoint,
@@ -33,6 +34,7 @@ import {
 	listSecondaryUsersPathSchema,
 } from './listSecondaryUsersEndpoint';
 import { mmaPrimarySummaryEndpoint } from './mmaPrimarySummaryEndpoint';
+import { secondaryUserMeEndpoint } from './secondaryUserMeEndpoint';
 
 const stage = stageFromEnvironment();
 const authenticate = buildAuthenticate(stage, []);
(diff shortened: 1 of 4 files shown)


handlers/multiple-account-api/src/index.ts
```diff
@@ -149,4 +155,21 @@ export const handler: Handler = Router([
 			),
 		),
 	},
+	{
+		httpMethod: 'GET',
+		path: '/secondary-user/me',
```
maloky, line 160: For consistency should we keep the same urls pattens as used above? 
```suggestion
		path: '/secondary-users/{SubscriptionName}',
```

maloky:
In the request example we are passing a request body, which is not recommended for GET methods as per spec: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Methods/GET. And since we don't seem to be using the secondaryUserEmail, we can got fully restful and pass the subscriptionName in the path. wdyt?

handlers/multiple-account-api/src/index.ts
rupertbates: The secondary user doesn't know the subscription name, we have to look it up using their identity id, that's why it's a /me endpoint - that's the convention in Idapi and the user-benefits api for an endpoint which takes the Identity id of the currently signed in user and returns details about them.

rupertbates:
> In the request example we are passing a request body, which is not recommended for GET methods as per spec: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Methods/GET. And since we don't seem to be using the secondaryUserEmail, we can got fully restful and pass the subscriptionName in the path. wdyt?

Yes, you're right that body is unused we don't need the subscriptionName either. It's left over from me duplicating a request in Postman! I've updated the description.

maloky (approved):
LG 👍

PR #3666, opened 2026-06-26
Internationalisation refactor

rupertbates:
I have broken this down into [separate PRs](https://github.com/guardian/support-service-lambdas/pulls?q=is%3Apr+label%3A%22Internationalisation+refactor%22) so closing this

PR #3700, opened 2026-07-08
Internationalisation refactor: Countries

rupertbates:
## What does this change?
This PR follows on from #3666 it is the second part of rationalising the types in the internationalisation module. 

### Changes
- Rename `IsoCountry` type to `CountryCode` and change all references and schema names
- Fix the logic around Guardian Weekly delivery logic (see below)
- Sort the keys of the various countries objects alphabetically and add name comments where they are just codes to make them easier to scan
- Updated some country names to their current official names (see below)
- Added an `isCountryCode` and `countryCodeFromString` convenience function - these are used by support-frontend

## GW delivery countries changes
Previously the modelling of GW delivery countries was out of date and a bit odd. We had a separate `gwCountries` list which was all countries except Russia and a list called `gwNonDeliverableCountries` which listed 11 other countries we couldn't sell to. The `gwDeliverableCountries` list was then built by removing the second list from the first.

Since we introduced digital benefits to Guardian Weekly we no longer sell it to any country in the `contributionsOnlyCountries` list (which includes all those on the `gwNonDeliverableCountries` list), so I have removed the `gwCountries` list and the `gwNonDeliverableCountries` list and we now build up the `gwDeliverableCountries` list by
- Taking the canonical `countries` list
-  removing the `contributionsOnlyCountries` list from it
- removing a new `restrictedCountries` list from it - this list is countries to which we should not make any sales whatsoever and currently only contains Russia.

## Country names changed:
Code | Previous name | Current official name
-- | -- | --
MK | Macedonia | North Macedonia (renamed 2019)
SZ | Swaziland | Eswatini (renamed 2018)
TR | Turkey | Türkiye (renamed 2022)
TL | East Timor | Timor-Leste
CV | Cape Verde Islands | Cabo Verde


maloky (commented):
LGTM 👍

maloky (approved):

PR #3703, opened 2026-07-14
Migrate from jest to vitest for product-switch-api and its workspace dependencies

johnduffell:
Tests seem to take an inordinately long time given how little they do.  For example product-switch-api take 1:41.  This could be because jest is slow, or that it uses ts-jest which runs the TS directly.

I looked at several options to improve this
- SWC https://swc.rs/docs/usage/jest (see below)
- esbuild-jest (no updates in 2 years) https://github.com/etroynov/esbuild-jest but a home rolled version did get reasonable benefits
- vitest

Although there were improvements for all three, it seemed like SWC and esbuild broke the logger tests which make sure the stack traces were matching what we expect in PROD.  Obviously a stack trace is an implementation detail but we do currently lean on it for making the logger more helpful in cloudwatch.

After researching vitest, everyone seems to be raving about it.

This PR is the initial improvement PoC - product-switch-api and all its dependencies are updated so far.

As you can see, the main difference is replacing all the `jest.` with `vi.`.  This is because vitest closely replicates jest's interface.

before
https://github.com/guardian/support-service-lambdas/actions/runs/29317084721/job/87033507432
image

after 
https://github.com/guardian/support-service-lambdas/actions/runs/29337152560/job/87099195243?pr=3703
image

There are other benefits, including that the "test failed" summary is anecdotally a bit neater
image

If we like it, we can roll it out to the other lambdas and modules, and clean up the remaining jest code.

See inline comments for the significant things - it's quite a noisy diff otherwise.

----
### swc/jest

This was very simple to get running - the screenshot captures pretty much the notable change needed although there were more changes needed.

The main show stopper was that the the imoprtant logger test was invalidated by the changed stack trace.
image

### esbuild

Another small change, the notable part is shown below but there were a few more changes needed

The main worry was the fact that it hadn't been maintained in 2 years.  Copilot was happy to generate the required transform for us though but it felt like something we shouldn't be maintaining ourselves.

image


buildcheck/data/snippets/buildJestConfigJs.ts
```diff
@@ -1,6 +1,9 @@
 import type { ModuleDefinition } from '../build';
 
-export function buildJestConfigJs(pkg: ModuleDefinition) {
+export function buildJestConfigJs(pkg: ModuleDefinition): string | undefined {
+	if (pkg.testRunner !== 'jest') {
+		return undefined;
```
johnduffell, line 5: undefined (now) means "don't create this file"

buildcheck/data/snippets/buildPackageJson.ts
```diff
@@ -26,15 +26,27 @@ export function buildPackageJson(
 ) {
 	assertNoDisallowedLibs(pkg.name, pkg.dependencies);
 	assertNoDisallowedLibs(pkg.name, pkg.devDependencies);
+
+	const jestScripts = {
+		test: `jest --group=-integration${pkg.noTests ? ' --passWithNoTests' : ''}`,
```
johnduffell, line 31: --passWithNoTests was previously implemented in the `srcOnly` const which was spread over the top of the normal `test` script.
This brings it in as a proper functionality based on `noTests` which makes it easier to implement in vitest as well, and it's also less surprising.

buildcheck/data/build.ts
```diff
@@ -23,6 +23,10 @@ export interface ModuleDefinition {
 	tsConfigExtra?: Record<string, unknown>;
 	testTimeoutSeconds?: number;
 	jestClearMocks?: boolean;
+	/** Opt out of vitest (the default) back to jest. Remove once migrated. */
+	testRunner?: 'jest';
+	/** Module has no test files: passes with no tests found and lints src/ only. */
+	noTests?: boolean;
```
johnduffell, line 29: I'm not  keen on the double negative, but you can't seem to default a value in an object to `true`, only `undefined` which is falsy.  So whatever you want the default to be, it has to be falsy.
The other option is similar to testRunner have some kind of `tests?: 'none' | 'some';` and deal with it in the logic.
In reality it will either be "true" or omitted, so the double negative won't be so obvious.  Also, it's not a complex piece of code.

shtukas (approved):

buildcheck/data/build.ts
```diff
@@ -221,14 +229,13 @@ const moduleSupporterProductData: ModuleDefinition = {
 		...dep['dayjs'],
 		...dep['zod'],
 	},
-	extraScripts: {
-		test: 'NODE_OPTIONS="$NODE_OPTIONS --experimental-vm-modules" jest --group=-integration',
-	},
```
johnduffell, line 226: I think this was added in error in
https://github.com/guardian/support-service-lambdas/pull/3170/changes#diff-bcf42c911eeb9762ef487ffe11d4c7be8098f7dae423bb144098fe7a32da6a6aR7

handlers/product-switch-api/test/changePlan/action/switch/previewAction.test.ts
```diff
@@ -16,9 +16,9 @@ import pendingAmendmentsJson from '../../../fixtures/pendingAmendments.json';
 import { loadSubscription } from '../../prepare/subscriptionInformation.test';
 
 const mockZuoraClient = {
-	get: jest.fn(),
-	post: jest.fn(),
-	delete: jest.fn(),
+	get: vi.fn(),
```
johnduffell, line 19: all `jest` change to `vi`

handlers/user-subscriptions-api/test/meEndpoint.test.ts
```diff
@@ -4,7 +4,7 @@ import largeAccountResponse from '../../../modules/zuora/test/objectQuery/largeA
 import { MeEndpoint } from '../src/meEndpoint';
 
 test('me static endpoint', async () => {
-	mockZuoraClient.get = jest.fn().mockResolvedValueOnce(largeAccountResponse);
+	mockZuoraClient.get.mockResolvedValueOnce(largeAccountResponse);
```
johnduffell, line 7: we didn't need to overwrite the function that was already there, we can just add the mock to the existing one.  This means we don't hit issues during the migration where it's using vitest but we are on jest still.
It's possible there may be issues with clearAllMocks but we're ok for now and this is only a temporary state.

buildcheck/data/build.ts
```diff
@@ -23,6 +23,10 @@ export interface ModuleDefinition {
 	tsConfigExtra?: Record<string, unknown>;
 	testTimeoutSeconds?: number;
```
johnduffell, line 24: we can probably get rid of this - now we can globally set the it test timeout to 15 seconds in the specific it-test file and leave the normal one on normal default

buildcheck/data/snippets/buildVitestConfigTs.ts
```diff
@@ -0,0 +1,45 @@
...
+		return undefined;
+	}
+	return `import { defineConfig } from 'vitest/config';
+
+export default defineConfig({
+	resolve: { tsconfigPaths: true },
+	test: {
+		globals: true,
+		environment: 'node',
+		include: ['test/**/*Integration.test.ts', 'test/**/*.it.test.ts'],
```
johnduffell, line 36: this is now filename based instead of annotation based, which seems reasonable.  It would probably make sense to standardise on `.it.test.ts` in due course, but in pursuit of minimising changes in this PR we can allow both patterns.

modules/zuora/test/mocks/mockZuoraClient.ts
```diff
@@ -1,17 +1,24 @@
+import type { Mock } from 'vitest';
 import type { BearerTokenProvider } from '@modules/zuora/auth';
 import { ZuoraClient } from '@modules/zuora/zuoraClient';
 
+// This mock is shared across Vitest and Jest packages.
+// vi is available as a global in Vitest; jest is available in Jest.
+type FnFactory = { fn: () => Mock };
+const vitest = (globalThis as typeof globalThis & { vi?: FnFactory }).vi;
+const mockFn = (): Mock => (vitest ?? (jest as unknown as FnFactory)).fn();
```
johnduffell, line 9: this is just a bit of cleverness while some are using vi and some jest, it will go away in due course.

tsconfig.json
```diff
@@ -37,7 +37,8 @@
 		// "typeRoots": [],                                  /* Specify multiple folders that act like './node_modules/@types'. */
 		"types": [
 			"node",
-			"jest"
+			"jest",
+			"vitest/globals"
```
johnduffell, line 41: these can coexist for now, then jest can go away in the end.  This is easier than putting them into the individual tsconfigs based on the runner in the build.ts and taking it out again later.

PR #3704, opened 2026-07-14
Add more excluded rate plan ids to supporter product data aqua query

rupertbates:
## What does this change?
This PR extends the list of product rate plan ids that are ignored by the supporter-product-data-lambdas step function. Previously this list was just comprised of discount ids as these are not relevant when working out product benefits, now we are also including '6 for 6' rate plans in the excluded list.

'6 for 6' was a special type of rate plan which as added at acquisition to Guardian Weekly Quarterly subscriptions in the past, it gave a six week period of deliveries at the cost of £6 (or $6 etc.). All of these rate plans are long since expired and crucially do not exist in the newer Guardian [product catalog](https://product-catalog.guardianapis.com/product-catalog.json), however they do still exist on the subscription.

Because the SupporterProductData Dynamo DB table where we store supporter information is keyed on subscription number, we can only store one rate plan type per subscription in it, and the one which gets stored is a 'last write wins'  mechanism. If that last write is the '6 for 6' rate plan then the user-benefits API will ignore it because it does not exist in the product catalog and the user will not get their benefits.

By adding '6 for 6' rate plans to the excluded product rate plans list we make sure that we only write rate plans which are in the product catalog to SupporterProductData.

## Deployment
After this PR has deployed we will run a full sync to repair any subscriptions which are in an invalid state.

Unit testing and deployment to CODE

---
  - https://app.asana.com/0/0/1216524614170081


handlers/supporter-product-data-lambdas/src/handlers/processSupporterRatePlanItem.ts
```diff
@@ -34,8 +34,10 @@ export const processItem = async (
 ): Promise<void> => {
 	logger.log('Processing supporter rate plan item', item);
 
-	if (dependencies.isDiscountRatePlanItem(item)) {
-		logger.log('Supporter rate plan item is a discount and will be skipped');
+	if (dependencies.isExcludedRatePlanItem(item)) {
+		logger.log(
+			'Supporter rate plan item is an excluded rate plan so will be skipped',
```
tjmw, line 39: Would it be helpful to include an identifier for the rate plan in this message?

handlers/supporter-product-data-lambdas/src/services/excludedRatePlans.ts
```diff
@@ -0,0 +1,42 @@
...
+		throw new Error('Discounts product not found in catalog');
+	}
+	return discountProduct.productRatePlans.map((plan) => plan.id);
+}
+
+function getSixForSixProductRatePlanIds(zuoraCatalog: ZuoraCatalog): string[] {
+	const guardianWeeklyProducts = zuoraCatalog.products.filter((product) =>
+		product.name.startsWith('Guardian Weekly'),
+	);
+	if (guardianWeeklyProducts.length !== 5) {
```
tjmw: What's the significance of `5` here? Could it be extracted to variable and given a name?

handlers/supporter-product-data-lambdas/src/services/excludedRatePlans.ts
```diff
@@ -0,0 +1,42 @@
...
+	const discountProduct = zuoraCatalog.products.find(
+		(product) => product.name === 'Discounts',
+	);
+	if (!discountProduct) {
+		throw new Error('Discounts product not found in catalog');
+	}
+	return discountProduct.productRatePlans.map((plan) => plan.id);
+}
+
+function getSixForSixProductRatePlanIds(zuoraCatalog: ZuoraCatalog): string[] {
```
tjmw, line 24: I'm slightly confused about how this identifies the 6 for 6 rate plan - the filter below looks quite generic to GW, not just 6 for 6?

rupertbates: Sorry the last commit didn't push successfully, it's there now.

tjmw (approved):

PR #3711, opened 2026-07-17
The taxMode property in the product catalog is not optional

tjmw:
## What does this change?

Update the product catalog schema to remove the optionality of the `taxMode` property.

This property was added in #3625 where it was made optional to handle a race condition where the deployment happened before the catalog was regenerated. The optionality can now be removed, it will either be set or null but never undefined.

**This must go after #3713.**


rupertbates (approved):

tjmw:
Superseded by #3728

PR #3713, opened 2026-07-17
Set taxMode for Stripe products in product catalog

tjmw:
## What does this change?

Set `taxMode` (to `TaxInclusive`) for Stripe products (i.e. products which aren't in Zuora) in the product catalog. For consistency we want this to always be defined, but it will always be `TaxInclusive` for these products.


maloky (approved):
LG 👍

maloky (approved):

paul-daniel-dempsey (approved):
Looks like patrons and one-time conributions 👍

PR #3716, opened 2026-07-17
Show the amount without tax in day 0 emails (for tax exclusive rate plans)

tjmw:
## What does this change?

Before this change, we always use the total amount from the payment schedule (`amount`) in day 0 emails. For tax exclusive rate plans this isn't what we want, we want the amount without taxes. Separate fields were added in guardian/support-frontend#8104 to facilitate this.

We'd like to remove `amount` from the payment schedule as it duplicates other data (it can be derived from `amountWithoutTax` and `taxAmount` by summing them). So start doing this in some places instead of using `amount`.

There are some use cases outside of the support-workers day 0 email which have been affected by the change to the payment schedule, for example the product switch email confirmation. I might need some help testing these.

The day 0 email changes have been tested by taking out tax exclusive rate plans in CODE (see screenshots on guardian/support-frontend#8107). For example this is for a CAD tax exclusive rate plan:

Screenshot 2026-07-20 at 14 41 55

### DIFF
--- a/handlers/product-switch-api/src/changePlan/action/productSwitchEmail.ts
+++ b/handlers/product-switch-api/src/changePlan/action/productSwitchEmail.ts
@@ -25,14 +25,20 @@ export const buildEmailMessage = (
 
 	const subscriptionRate = describePayments(
 		{
-			payments: paymentSchedule.map(({ date, total }) => ({
-				date,
-				amount: total,
-			})),
+			payments: paymentSchedule.map(
+				({ date, total, amountWithoutTax, taxAmount }) => ({
+					date,
+					amount: total,
+					amountWithoutTax,
+					taxAmount,
+				}),
+			),
 		},
 		productRatePlanKey,
 		currency,
 		false,
+		// TODO: make this dynamic
+		'TaxInclusive',
 	);
 	const nextPaymentDate = paymentSchedule[0]?.date;
 
--- a/handlers/product-switch-api/src/changePlan/action/productSwitchEmail.ts
+++ b/handlers/product-switch-api/src/changePlan/action/productSwitchEmail.ts
@@ -37,8 +37,7 @@ export const buildEmailMessage = (
 		productRatePlanKey,
 		currency,
 		false,
-		// TODO: make this dynamic
-		'TaxInclusive',
+		switchInformation.target.taxMode,
 	);
 	const nextPaymentDate = paymentSchedule[0]?.date;
 
--- a/handlers/product-switch-api/src/changePlan/prepare/targetInformation.ts
+++ b/handlers/product-switch-api/src/changePlan/prepare/targetInformation.ts
@@ -5,6 +5,7 @@ import type {
 	GuardianCatalogKeys,
 	ProductCatalogHelper,
 	ProductKey,
+	TaxMode,
 } from '@modules/product-catalog/productCatalog';
 import type { ProductSwitchTargetBody, SwitchMode } from '../schemas';
 import type { Discount } from '../switchDefinition/discounts';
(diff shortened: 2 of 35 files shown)


handlers/product-switch-api/src/changePlan/action/productSwitchEmail.ts
```diff
@@ -25,14 +25,19 @@ export const buildEmailMessage = (
 
 	const subscriptionRate = describePayments(
 		{
-			payments: paymentSchedule.map(({ date, total }) => ({
-				date,
-				amount: total,
-			})),
+			payments: paymentSchedule.map(
```
tjmw, line 28: Apparently we won't support switching to any tax exclusive rate plans initially, but the changes in this file should mean it's ready to go when we do decide to enable this.

handlers/product-switch-api/src/changePlan/prepare/targetInformation.ts
```diff
@@ -25,6 +25,7 @@ export type TargetInformation = {
 	subscriptionChargeId: string; // adjust invoice, build response // used to find the price and service end date (next payment date) from the preview invoice, also to find and adjust out any charge less than 0.50
 	contributionCharge?: TargetContribution; // order // used to find the price from the preview invoice as above, also to do the chargeOverrides in the order to set the additional amount to take
 	discount?: Discount; // order (product rate plan id), return to client
+	taxMode: 'TaxExclusive' | 'TaxInclusive' | null | undefined;
```
rupertbates: Is the TaxMode type not defined somewhere else?

johnduffell: what does undefined mean distinct from null?  I think it comes from the original PR so I have asked there https://github.com/guardian/support-service-lambdas/pull/3625/changes#r3615108820

tjmw: Oops I added a comment about this on the PR but I must have not clicked the CTA. There's no difference and we shouldn't have both. I'm working on reducing it to just `null` with no `undefined` but that work is in a separate PR (the slightly longer story is that there are stripe products, as opposed to Zuora products, in the product catalog API response and these don't have a taxMode property, but I'm adding it for them).

modules/email/src/dataFields/dayZero/types.ts
```diff
@@ -2,6 +2,8 @@ import type { CountryCode } from '@modules/internationalisation/country';
 
 export type EmailBillingPeriod = 'Annual' | 'Monthly' | 'Quarterly';
 
+export type TaxMode = 'TaxInclusive' | 'TaxExclusive' | undefined | null;
```
rupertbates: Maybe move this type to the top level of @modules then everything can access it from there, it's not really email related.

handlers/product-switch-api/src/changePlan/prepare/targetInformation.ts
tjmw: Ah good spot, there's a named type in the email module, but I'd missed these cases in the product-switch-api. Perhaps the email module isn't the right place for it, the fact that it's here I think means that it's broader than that. Perhaps it should be in the product catalog and both can import from there.

modules/email/src/dataFields/dayZero/types.ts
tjmw: I think product catalog might be the right place (see comment above). What do you reckon? Or happy to go with top level modules if that's preferable.

handlers/product-switch-api/src/changePlan/prepare/targetInformation.ts
rupertbates: Yes or at the root of the modules alongside billingPeriod

modules/email/src/dataFields/dayZero/types.ts
rupertbates: I'm happy with either

rupertbates (approved):
Looks good to me

modules/email/src/dataFields/dayZero/types.ts
tjmw: I went with product catalog in the end, we can always revisit if that ends up feeling wrong.

PR #3717, opened 2026-07-20
block CAD switches to S+

johnduffell:
We currently switch people to the traditional tax inclusive plan regardless of currency.

Once we start selling Tax exclusive plans in CAD, we don't want to move people across onto the old plan.  This is because we want to cleanly turn off new acquisitions before the migration can happen, to avoid leaving people behind.

This PR blocks CAD switches as a temporary measure until we implement switches over to the new tax exclusive plan (along with any copy and email confirmations)


johnduffell:
not needed after all as the business don't consider it an issue

PR #3718, opened 2026-07-21
add new iam-policies handler for restricted janus permissions

johnduffell:
This is an implementation of a "developer policy" for daily development use in the membership account.
See this doc for the general concept: https://github.com/guardian/janus/blob/main/docs/developer-policies.md

The idea is this initial policy is enough for developers to develop and test locally without giving them any access to production resources or keys/secrets.

The existing read only access is no good because it gives read only access to everything (including PROD data and parameters) but not secrets manager CODE zuora creds.  So it's the worst of both worlds.

This PR adds the policy and enough permissions to get these working
- support service lambdas IT tests (for discount-api)
- manage-frontent
- support-admin-console

The idea is to roll this out and give the permission to a few more people [1] so we can test and iterate against support-frontend and payment-api (and anything else that's commonly used) then we can roll it out to the whole team and continue to iterate.

[1] to get the permission, give yourself `DeveloperPolicyGrants.membershipLocalDev` in janus, e.g.
https://github.com/guardian/janus/pull/5659

Only the CODE version exists at present and it looks like this
image

### DIFF
--- a/cdk/package.json
+++ b/cdk/package.json
@@ -16,7 +16,7 @@
 		"update-stack": "../update-stack.sh"
 	},
 	"devDependencies": {
-		"@guardian/cdk": "61.11.1",
+		"@guardian/cdk": "64.1.0",
 		"@types/jest": "^29.5.14",
 		"@types/node": "^22.15.14",
 		"aws-cdk": "2.1128.1",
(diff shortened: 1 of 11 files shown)


cdk/package.json
```diff
@@ -16,7 +16,7 @@
 		"update-stack": "../update-stack.sh"
 	},
 	"devDependencies": {
-		"@guardian/cdk": "61.11.1",
+		"@guardian/cdk": "64.1.0",
```
johnduffell, line 19: the existing version didn't have the `GuDeveloperPolicyExperimental` class

shtukas (approved):

PR #3728, opened 2026-07-24
Remove optionality from product catalog schema

tjmw:
## What does this change?

Following on from #3713 this PR removes `.optional()` from the product catalog schema.

This shouldn't be needed any longer, the property will always be there (but may be null).

I deployed this branch to CODE and re-generated the product catalog (which uses the schema to validate) by triggering the `generate-product-catalog-CODE` lambda.

I also tested the user-benefits-api, which fetches and parses the product catalog, and verified that it worked with both the version of the catalog api response before this PR and after (there shouldn't be any difference).


paul-daniel-dempsey (approved):
LG 👍

rupertbates (approved):
👍

PR #3731, opened 2026-07-24
IAM policies: Additions for local support-frontend development

tjmw:
following on from the intial PR https://github.com/guardian/support-service-lambdas/pull/3718

## What does this change?

Adds some additional policies required to run support-frontend locally.

Tested by deploying iam-policies to CODE and using the CODE credentials from Janus. With this set of permissions I was able to put through a transaction in my locally running support frontend.

### DIFF
--- a/cdk/lib/iam-policies.ts
+++ b/cdk/lib/iam-policies.ts
@@ -28,6 +28,7 @@ export class IamPolicies extends SrStack {
 				getSupportAdminConsoleBucketPolicy(),
 				new AllowDynamoTableFullAccessPolicy(this),
 				...getManageFrontendPolicies(this.region, this.account),
+				...getSupportFrontendPolicies(this.region, this.account),
 				new PolicyStatement({
 					actions: ['cloudwatch:PutMetricData'],
 					resources: ['*'],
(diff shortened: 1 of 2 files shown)


johnduffell (approved):
great, thanks for looking at that 👏

cdk/lib/iam-policies.ts
```diff
@@ -129,10 +131,33 @@ function getManageFrontendPolicies(region: string, account: string) {
 		fulfilmentDatesBucketPolicy,
 		allowListStackResources,
 		unsafeAllowGetAllApiKeys,
-		allowInvokeLambda,
+		allowInvokeApi,
 	];
 }
 
+function getSupportFrontendPolicies(region: string, account: string) {
```
johnduffell, line 138: interestingly the actual app has a few more policies than provided by the overall policy here
https://github.com/guardian/support-frontend/blob/cf81030821a6504a58e8372164cf2b5d64c3f610/cdk/lib/frontend.ts#L65-L134

Some of them are covered in the generic allows elsewhere in this role.

For the rest, it could be that some of the above are no longer used and should be removed, or we need to test that functionality and add the permissions.

tjmw: Yes I spotted that, I think it could be a mix of permissions which are no longer used and ones we'll need to add here. I think it's okay to build these up incrementally rather than get them all added in a single PR, what do you think?

johnduffell: yes incremental makes sense to me, having the basics work is much better than nothing!

PR #3734, opened 2026-07-27
Integrate support-service-lambdas with @guardian/devenv

rupertbates:
## What does this change?
This PR adds a dev container configuration managed by the [Guardian Devenv tool](https://github.com/guardian/devenv) to allow us to run support-service-lambdas in a dev container. This provides better isolation of the development environment and limits the impact of issues like supply chain attacks if they occur.

## Usage
See the [readme in this pull request](https://github.com/guardian/support-service-lambdas/blob/rb/devcontainer/.devcontainer/README.md)

## Additional optional setup
Provide a dotfiles repo to configure your devcontainer in the same way as your local machine.
This is described in the [devenv documentation](https://github.com/guardian/devenv#devenv) and I have an example of this [here](https://github.com/rupertbates/dotfiles).


.devcontainer/devenv.yaml
```diff
@@ -0,0 +1,85 @@
+# Devenv project configuration
+# Edit this file to configure your project's devcontainer
+# and then run `devenv generate` to create the devcontainer.json files
+
+# REQUIRED: Change this to your project name
```
johnduffell:
```suggestion
```

cdk/package.json
```diff
@@ -34,6 +34,7 @@
 		},
 		"setupFilesAfterEnv": [
 			"./jest.setup.js"
-		]
+		],
+		"workerIdleMemoryLimit": "1G"
```
johnduffell, line 38: is this intentional?

.devcontainer/README.md
```diff
@@ -0,0 +1,155 @@
...
+- **[mise](https://mise.jdx.dev/)** — needed on the host to generate the devcontainer config (`devenv generate`,
+  see below). Everything else the project needs runs inside the container.
+
+Apart from `mise` (used only to generate the config), all other tooling (Java, Node, the GitHub Copilot CLI, etc.)
+is installed automatically inside the container, so nothing else needs to be installed on your host.
+
+### Create the dev container (one time only)
+
+The standard flow uses a **local checkout only to hold the config**, then has the IDE **clone the sources
+into a container volume** — so there is no bind mount of your working tree.
```
johnduffell, line 29: what does this mean?

.devcontainer/README.md
```diff
@@ -0,0 +1,155 @@
...
+   ```
+3. Generate the devcontainer config, including your personal `user/devcontainer.json` (see
+   [`devenv.yaml`](#devenvyaml) below):
+   ```bash
+   devenv generate
+   ```
+4. Create the container from the generated **`user/devcontainer.json`**, using your IDE's
+   clone-into-volume action (this clones the sources into a container volume; your local checkout is
+   not bind-mounted):
+   - **IntelliJ IDEA**: open `.devcontainer/user/devcontainer.json` and use the gutter action
```
johnduffell, line 49:
```suggestion
   - **IntelliJ IDEA**: open `.devcontainer/user/devcontainer.json` and use the gutter action, then click through the wizard.
```

johnduffell: in case it helps someone else, I found I have to select the branch `rb/devcontainer` when cloning, not the default `main`, otherwise when it runs mise inside the container it can't find the tool versions file.

.devcontainer/README.md
```diff
@@ -0,0 +1,155 @@
...
+5. Wait for the first build to finish. The first run is slower because it provisions the shared Scala caches
+   (Ivy and coursier volumes), sets up the in-container tooling, and installs the GitHub Copilot CLI. Subsequent
+   starts reuse these and are much faster.
+
+> [!NOTE]
+> Using `user/devcontainer.json` (rather than the committed `shared/devcontainer.json`) is what lets your
+> personal overrides — extra extensions/plugins etc. — apply to the container, so keep the local checkout
+> around and re-run `devenv generate` after editing `devenv.yaml`. Bear in mind that with clone-in-volume
+> your **uncommitted work lives only in the container volume** — if that volume is removed (e.g. a
+> from-scratch re-clone or `docker rm -f`), the work is lost, so commit and push often.
```
johnduffell: this seems like two notes, one about overrides, and one about that data might be lost, so they should be separated.

Regarding the work being in the "container volume" I don't quite follow any additional risk - in the old world all your unpushed changes appear in your workspace e.g. `~/code/support-service-lambdas/` and if that's removed the work is lost.  What's special about docker volumes that makes them more likely to be lost than a normal workspace?

.devcontainer/README.md
```diff
@@ -0,0 +1,155 @@
...
+   mise install
+   ```
+3. Generate the devcontainer config, including your personal `user/devcontainer.json` (see
+   [`devenv.yaml`](#devenvyaml) below):
+   ```bash
+   devenv generate
+   ```
+4. Create the container from the generated **`user/devcontainer.json`**, using your IDE's
+   clone-into-volume action (this clones the sources into a container volume; your local checkout is
+   not bind-mounted):
```
johnduffell, line 48: it seems like "bind-mounted" means that the container uses the actual workspace directly rather than making an isolated copy.

```suggestion
4. Create the container from the generated **`user/devcontainer.json`**, using your IDE's
   clone-into-volume action (cloning into a volume means the AI agent cannot see/edit files visible from outside that container workspace):
```

.devcontainer/README.md
```diff
@@ -0,0 +1,155 @@
...
+> Using `user/devcontainer.json` (rather than the committed `shared/devcontainer.json`) is what lets your
+> personal overrides — extra extensions/plugins etc. — apply to the container, so keep the local checkout
+> around and re-run `devenv generate` after editing `devenv.yaml`. Bear in mind that with clone-in-volume
+> your **uncommitted work lives only in the container volume** — if that volume is removed (e.g. a
+> from-scratch re-clone or `docker rm -f`), the work is lost, so commit and push often.
+
+### Open an existing dev container
+
+Once the container has been created (see above), you don't need to recreate it — reconnect to the existing
+container instead (this reuses the same cloned-sources volume, so your work is preserved).
```
johnduffell, line 64: I think this paragraph says very little and could be removed.  What would be more useful is a short section on best practice for container reuse - do I use them like branches and create/trash them on each piece of work?  Or do I just have one or two and keep using them?

.devcontainer/README.md
```diff
@@ -0,0 +1,155 @@
...
+   starts reuse these and are much faster.
+
+> [!NOTE]
+> Using `user/devcontainer.json` (rather than the committed `shared/devcontainer.json`) is what lets your
+> personal overrides — extra extensions/plugins etc. — apply to the container, so keep the local checkout
+> around and re-run `devenv generate` after editing `devenv.yaml`. Bear in mind that with clone-in-volume
+> your **uncommitted work lives only in the container volume** — if that volume is removed (e.g. a
+> from-scratch re-clone or `docker rm -f`), the work is lost, so commit and push often.
+
+### Open an existing dev container
```
johnduffell:
```suggestion
## Open an existing dev container
```
"getting started" section says "These steps take you from a fresh checkout to a running dev container." which has happened prior to this heading.

I think this heading should be brought up a level and focus more on best practices/options etc

.devcontainer/README.md
```diff
@@ -0,0 +1,155 @@
...
+4. Create the container from the generated **`user/devcontainer.json`**, using your IDE's
+   clone-into-volume action (this clones the sources into a container volume; your local checkout is
+   not bind-mounted):
+   - **IntelliJ IDEA**: open `.devcontainer/user/devcontainer.json` and use the gutter action
+     **Create Dev Container and Clone Sources**.
+   - **VS Code**: run **Dev Containers: Clone Repository in Container Volume** from the command palette
+     (pointed at this repo), using the `user/devcontainer.json` configuration when prompted.
+5. Wait for the first build to finish. The first run is slower because it provisions the shared Scala caches
+   (Ivy and coursier volumes), sets up the in-container tooling, and installs the GitHub Copilot CLI. Subsequent
+   starts reuse these and are much faster.
```
johnduffell: I feel like it's mainly trying to excuse the slowness (but in my case it wasn't actually slow, maybe 30 seconds)
```suggestion
5. Wait for the build to finish.  In case of errors, see Troubleshooting below, otherwise dependencies are now installed so use your IDE and copilot CLI as normal.
```

.devcontainer/README.md
```diff
@@ -0,0 +1,155 @@
...
+> Avoid re-running the "create / clone sources" action to get back into an existing container — that provisions
+> a *new* container. Because the container has a fixed name (`--name` in `devenv.yaml`) Docker will refuse the
+> duplicate with a name conflict; use the reconnect actions above instead.
+
+### Working inside the container
+
+The `postCreateCommand` (see `devenv.yaml`) already runs `pnpm install` for you as part of the first
+build, so the Node dependencies are ready once the container has finished setting up. Open a terminal
+inside the container and follow the main [README](../README.md) for the Node.js and Scala development
+workflows (building, testing, linting, etc.).
```
johnduffell: this can probably be deleted once the suggested changes to step 5 above are made.

.devcontainer/README.md
```diff
@@ -0,0 +1,155 @@
...
+
+### Troubleshooting
+
+If the container fails to build or set up correctly, the setup output is logged inside the container at
+`/var/log/post-create.log`, which is a good starting point for debugging.
+
+#### IntelliJ: "Node.js is required" / test run says "no node executable found in path"
```
johnduffell: I did get this error but actually it was because I let it use "main" as the branch when I created the container, not realising it runs mise install in the post create script.

It miht have been worth merging the enabling changes e.g. to use `mise` in general first, and then doing this PR as a self contained one.

.devcontainer/devenv.yaml
```diff
@@ -0,0 +1,85 @@
...
+  # or the mise shims (~/.local/share/mise/shims), which are only added to
+  # interactive/login shells via ~/.bashrc / ~/.profile. Without this, IntelliJ
+  # reports "Node.js is required" and tests fail with "no node executable found in
+  # path". Symlink mise and its tool shims into /usr/local/bin (which IS on the
+  # default PATH) so node, java, npm etc. are always discoverable. The shims
+  # resolve tool versions from the project's mise config at runtime.
+  - cmd: 'sudo ln -sf "$HOME/.local/bin/mise" /usr/local/bin/mise && sudo ln -sf "$HOME"/.local/share/mise/shims/* /usr/local/bin/'
+    workingDirectory: '.'
+  - cmd: 'mise exec -- node --version'
+    workingDirectory: '.'
```
johnduffell, line 71: can we also give the commands a name, as they all come into the logs as "unknown" otherwise, e.g. I can only tell the indicated one is node version by the output

Image

.devcontainer/README.md
```diff
@@ -0,0 +1,155 @@
...
+> `/workspaces/<repo>`). If you **attach from a different IDE than the one that created the container** (e.g.
+> attach with VS Code to a container IntelliJ created), it opens the container's default directory rather than
+> the project, so you may have to open the folder manually. To make this predictable, the `postCreateCommand`
+> creates a fixed symlink **`~/project`** pointing at wherever the sources were cloned — so you can always open
+> `~/project` (i.e. `/home/vscode/project`) to get to the project regardless of which IDE created the container.
+
+> [!TIP]
+> Avoid re-running the "create / clone sources" action to get back into an existing container — that provisions
+> a *new* container. Because the container has a fixed name (`--name` in `devenv.yaml`) Docker will refuse the
+> duplicate with a name conflict; use the reconnect actions above instead.
```
johnduffell: this doesn't seem to be the case, in intellij it just happily lets me do it again, it seems to recreate the dev container but reuse the existing volume.  But it does run the postCreate tasks again which could possibly cause issues e.g. resymlinking the workspace
```
Reuse sources volume from the previous dev container.
Removing previous dev container…
Creating container…
Container Id: 5ac154c5f42eac457557ba3b41b8ff8f53845fb9b3cee619b91382d40b35ba48
Container name: '/support-service-lambdas-devcontainer'
Starting container '/support-service-lambdas-devcontainer'
....
```

.devcontainer/README.md
```diff
@@ -0,0 +1,155 @@
...
+shims resolve the correct versions from the project's [`.tool-versions`](../.tool-versions) at runtime. If you still see
+the warning after a rebuild, reopen the project (or restart the IDE) so IntelliJ re-detects the Node interpreter from
+`PATH`.
+
+## .devcontainer directory contents
+
+### `README.md`
+
+Edit this file as you see fit.  It is created on the first run of `devenv init` but it won't be overwritten
+by subsequent runs unless you delete it entirely.
```
johnduffell: is this just boilerplate to delete?

.devcontainer/README.md
```diff
@@ -0,0 +1,155 @@
...
+
+### `user/devcontainer.json`
+
+**This is the file you launch the dev container from** (see [Create the dev container](#create-the-dev-container)).
+It is generated by `devenv generate` from `devenv.yaml` **merged with your personal preferences** (extra VS Code
+extensions or IntelliJ plugins, personal settings/tweaks), and is **git-ignored** (see `.gitignore`) so those
+overrides don't affect other contributors.
+
+To add personal overrides, put them in your user devenv config (`~/.config/devenv/devenv.yaml`) and re-run
+`devenv generate` — do not hand-edit `user/devcontainer.json`, as `devenv generate` will overwrite it.
```
johnduffell:
```suggestion
### `shared/devcontainer.json` and `user/devcontainer.json`

Do not edit these - edit `devenv.yaml` and then run `devenv generate`.

The `user` file lets you create your dev container.  It includes personal overrides from your user devenv config (`~/.config/devenv/devenv.yaml`) and is gitignored.

The `shared` file is a reference snapshot created from devenv.yaml committed to version control.

johnduffell (approved):
thanks for doing this, I've been looking forward to having it.  I had a go it seems quite smooth and I'm quite impressed how little we need to add to the repo to enable it (it seems like the README is longer than the code changes needed!)

See inline comments for improvements/suggestions

cdk/package.json
rupertbates: Yes, it's a Jest option that sets a memory ceiling for each worker process. When a worker's memory usage exceeds this limit after finishing a test, Jest restarts that worker (child process) rather than reusing it.

I was getting OOM sigkill issues running the CDK tests in a container and this resolved it.

.devcontainer/README.md
rupertbates: When creating the container you can either mount the repo from your local machine or clone it from github, this is saying that in this case we are cloning.

PR #3715, opened 2026-07-17
Add cancelledBy field to multiple accounts invitations

rupertbates:
## What does this change?
This PR adds two new optional fields to multiple accounts Invitation records - `cancelledBy` and `cancelledDate` will be set when the invitation is either cancelled via MMA by the primary user or rejected by the secondary user.

The value will be determined in the delete invitation endpoint from the `x-identity-id` header which should match either the primary or secondary user identity id on the invitation record. If this header is not present the response will be a 400 bad request.

### Create a new invitation
Request:

```shell
curl --location 'https://multiple-account-api-code.support.guardianapis.com/invitation' \
--header 'x-api-key: [API_KEY]' \
--header 'x-identity-id: 112809589' \ # primary user
--header 'Content-Type: application/json' \
--header 'Authorization: ••••••' \
--data-raw '{
  "subscriptionName": "A-S00974337",
  "secondaryUserEmail": "test@thegulocal.com"
}'
```

Response:
`200`
```json
{
    "invitationCode": "FRcCyefqP3qf"
}
```

### Delete that invitation as the secondary user
Request:
```shell
curl --location --request DELETE 'https://multiple-account-api-code.support.guardianapis.com/invitation/FRcCyefqP3qf' \
--header 'x-api-key: [API_KEY]' \
--header 'x-identity-id: 21841960' # secondary user
```

Response:
`204 No Content`

### Try to delete the invitation again
Request:
```shell
curl --location --request DELETE 'https://multiple-account-api-code.support.guardianapis.com/invitation/FRcCyefqP3qf' \
--header 'x-api-key: [API_KEY]' \
--header 'x-identity-id: 21841960' # secondary user
```

Response:
`404`
```json
{
    "message": "Not Found"
}
```

### Try to accept the invitation
Request:
```shell
curl --location --request POST 'https://multiple-account-api-code.support.guardianapis.com/invitation/FRcCyefqP3qf/accept' \
--header 'x-api-key: [API_KEY]' \
--header 'Authorization: ••••••'
```

Response
`400`
```json
{
    "message": "Invitation has been cancelled by the secondary user"
}
```

### View the invitation list as the primary user
Request:
```shell
curl --location 'https://multiple-account-api-code.support.guardianapis.com/subscriptions/A-S00974337/mma-primary' \
--header 'x-api-key: [API_KEY]' \
--header 'x-identity-id: 112809589'
```

Request:
`200`
```Json
{
    "invitations": [],
    "secondaryUsers": []
}
```
[...]

PR #3736, opened 2026-07-29
Multiple accounts: Soft delete secondary users

rupertbates:
## What does this change?
This PR adds soft deletion to secondary users in the multiple accounts API.
It follows on from https://github.com/guardian/support-service-lambdas/pull/3715 which did the same for invitations.

## Details
Secondary users can be removed from a subscription with multiple accounts benefits in two ways;
either the primary user may remove them, or they may remove themselves. 

Previously when this happened the secondary user record was hard deleted from the data store, but this made it hard for the data team to provide analysis on which user had removed the record. 

To solve this we are introducing soft deletes for secondary users. When they are removed from the subscription now, the record is not deleted, instead a `cancelledBy` field is set to `primary` or `secondary` and a `cancelledDate` field is set to the current date time in ISO 8601 format. 

We also set the expiryDate field to now + 2 weeks, this is a Dynamo TTL field, meaning that the record will be hard deleted automatically on this date, meaning that we don't need to worry about retaining PII beyond that point but the data team have sufficient time to record the identity of the user who deleted the record.

The diff is slightly bigger than it would otherwise be because I refactored some names on the secondaryUserRepository class to make them consistent. Mostly ensuring that 'get' methods return a single items and 'list' methods return an array of items.

### Create a new invitation
Request:

```shell
curl --location 'https://multiple-account-api-code.support.guardianapis.com/invitation' \
--header 'x-identity-id: 112809589' \ # primary user
--data-raw '{
  "subscriptionName": "A-S00974337",
  "secondaryUserEmail": "test@thegulocal.com"
}'
```

Response:
`200`
```json
{
    "invitationCode": "FRcCyefqP3qf"
}
```

### Accept that invitation as the secondary user
Request:
```shell
curl --location --request POST 'https://multiple-account-api-code.support.guardianapis.com/invitation/twT95D1SFKBd/accept' \
--header 'Authorization: ••••••' # secondary user authenticated with Okta
```

Response:
```json
{
    "identityId": "21841960",
    "secondarySubscriptionName": "A-S00974337-21841960"
}
```

### Delete the invitation as the primary user
Request:
```shell
curl --location --request DELETE 'https://multiple-account-api-code.support.guardianapis.com/subscriptions/A-S00974337/secondary-users/21841960' \
--header 'x-identity-id: 112809589' # primary user
```

Response: `204 No Content` 

### Try to get the invitation
Request:
```shell
curl --location 'https://multiple-account-api-code.support.guardianapis.com/invitation/twT95D1SFKBd' \
--header 'x-api-key: [API_KEY]' 
```

Response
`404`
```json
{
    "message": "Not Found"
}
```

### View the secondary user list as the primary user
Request:
```shell
curl --location 'https://multiple-account-api-code.support.guardianapis.com/subscriptions/A-S00974337/mma-primary' \
--header 'x-identity-id: 112809589'
```

Request:
`200`
```Json
{
    "invitations": [],
    "secondaryUsers": []
}
```
Cancelled secondary users should not show up in the MMA response.

### DIFF
--- a/handlers/multiple-account-api/src/index.ts
+++ b/handlers/multiple-account-api/src/index.ts
@@ -155,17 +155,24 @@ export const handler: Handler = Router([
 		path: '/subscriptions/{subscriptionName}/secondary-users/{secondaryIdentityId}',
 		handler: withPathParser(
 			deleteSecondaryUserPathSchema,
-			withMMAIdentityCheck(
-				stage,
-				async (_body, _zuoraClient, subscription, _account, path) =>
-					deleteSecondaryUserEndpoint(
-						stage,
-						secondaryUserRepository,
-						dynamoClient,
-						path,
-					),
-				({ path }) => path.subscriptionName,
-			),
+			async (event, path) => {
+				// Both the primary (removing a secondary user) and the secondary
+				// user (removing themselves) send their identity id in the
+				// 'x-identity-id' header.
+				const loggedInUserIdentityId = event.headers['x-identity-id'];
+				if (!loggedInUserIdentityId) {
+					return badRequest('The x-identity-id header is required');
+				}
+				const { subscriptionName, secondaryIdentityId } = path;
+				return deleteSecondaryUserEndpoint(
+					stage,
+					secondaryUserRepository,
+					dynamoClient,
+					subscriptionName,
+					secondaryIdentityId,
+					loggedInUserIdentityId,
+				);
+			},
 		),
 	},
 	{
--- a/modules/multiple-account/src/secondaryUserRepository.ts
+++ b/modules/multiple-account/src/secondaryUserRepository.ts
@@ -1,6 +1,7 @@
 import {
 	DeleteItemCommand,
 	DynamoDBClient,
+	GetItemCommand,
 	PutItemCommand,
 	QueryCommand,
 	type TransactWriteItem,
@@ -61,7 +62,9 @@ export class SecondaryUserRepository {
 		);
 	}
 
-	async get(secondaryIdentityId: string): Promise<SecondaryUserRecord[]> {
+	async listByIdentity(
+		secondaryIdentityId: string,
+	): Promise<SecondaryUserRecord[]> {
 		const result = await this.client.send(
 			new QueryCommand({
 				TableName: this.tableName,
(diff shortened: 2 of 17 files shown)


handlers/multiple-account-api/src/index.ts
```diff
@@ -155,17 +155,24 @@ export const handler: Handler = Router([
 		path: '/subscriptions/{subscriptionName}/secondary-users/{secondaryIdentityId}',
 		handler: withPathParser(
 			deleteSecondaryUserPathSchema,
-			withMMAIdentityCheck(
```
tjmw, line 158: I can see we're now doing the identity check further down, but I wonder if there's a risk that we might accidentally remove/change this in a future PR since it's now a bit buried as opposed to right at the top via `withMMAIdentityCheck`. What do you think?

modules/multiple-account/src/secondaryUserRepository.ts
```diff
@@ -114,6 +179,29 @@ export class SecondaryUserRepository {
...
+			Update: {
+				TableName: this.tableName,
+				Key: {
+					subscriptionName: { S: subscriptionName },
+					secondaryIdentityId: { S: secondaryIdentityId },
+				},
+				UpdateExpression:
+					'SET expiryDate = :expiryDate, cancelledBy = :cancelledBy, cancelledDate = :cancelledDate',
+				ExpressionAttributeValues: {
+					':expiryDate': { N: secondaryUserCancellationTTL().toString() },
```
tjmw, line 197: I can't think how this would happen, but if there's already an `expiryDate` set for a record are we okay with just overwriting it?

tjmw (commented):
Looks good 👍

I left a couple of questions but no blockers.

tjmw (approved):

handlers/multiple-account-api/src/index.ts
rupertbates: I think moving it any higher would leak too much of the delete user endpoint logic up to this level. I'm more comfortable leaving it where it is I think.

modules/multiple-account/src/secondaryUserRepository.ts
rupertbates: Yes I think that's what we want to do here because we have a definitive deletion date which is today + 2 weeks.

PR #3739, opened 2026-07-30
Internationalisation Refactor: Support Regions

rupertbates:
## What does this change?
This PR is the final one in a [series](https://github.com/guardian/support-service-lambdas/issues?q=label%3A%22Internationalisation%20refactor%22) which standardises the way we define types and enumerated values in the internationalisation module.

Up until now we have had two separate types which captured the concept of a group of countries - CountryGroup and SupportRegion. These were more or less identical but used in different places, this PR removes the CountryGroup type and replaces all usages with SupportRegion. It also changes SupportRegion from being a Typescript enum, which is quite non-standard for our codebases into a standard union type derived from a list in the same way as all the other types in the internationalisation module.


modules/internationalisation/src/supportRegion.ts
```diff
@@ -309,43 +278,36 @@ export const countryGroups: CountryGroups = {
...
-		case SupportRegionId.INT:
-			return countryGroups.International;
-	}
 };
-export const supportRegionIdFromCountry = (country: CountryCode) => {
-	return Object.values(countryGroups).find((countryGroup) =>
-		countryGroup.countries.includes(country),
-	)?.supportRegionId;
+
+export const supportRegionIdFromCountryCode = (
```
rupertbates, line 295: This longer implementation is so we can return a SupportRegionId instead of an optional SupportRegionId

modules/internationalisation/src/currency.ts
```diff
@@ -10,38 +10,45 @@ export const currencyCodes = [
 export type CurrencyCode = (typeof currencyCodes)[number];
 
 export type Currency = {
+	code: CurrencyCode;
```
rupertbates, line 13: I have changed the `currency` property on a SupportRegion to be a full currency object rather than just a code, so this allows us to access the currency code via that route.

PR #3746, opened 2026-08-04
Add subscription name to secondary-user/me endpoint

rupertbates:
## What does this change?
This PR changes the structure of the response from the /secondary-users/me endpoint which was first introduced in: https://github.com/guardian/support-service-lambdas/pull/3697

Manage-frontend needs the subscription name (number) of the subscription which a secondary user belongs to so that it can remove them from it if needed. This field wasn't present in the previous response but this PR adds it.

## Authentication 
Authentication has not changed in this PR, it is via an Okta JWT token in the Authorization header. This is to prevent enumeration of identity ids - the signed in user will only be able to retrieve a list for their own identity id.

## Testing
Testing this endpoint is a little more complicated than some others because it requires an Okta Authorization header. 

To get one of these follow the steps in [this README](https://github.com/guardian/identity/tree/main/docs/generating_tokens_locally) to generate a header in Postman or Hopscotch. 

You can then make your request to secondary users endpoint:
```
## Example request 

```shell
curl --location --request GET 'https://multiple-account-api-code.support.guardianapis.com/secondary-user/me' \
```

The **OLD** version of this would have returned:

```json
{
    "primaryUsers": [
        {
            "firstName": "Test",
            "lastName": "User"
            "workEmail": "test.user@thegulocal.com"
        }
    ]
}
```

The **NEW** version returns:
```json
{
    "subscriptions": [
        {
            "subscriptionName": "A-S00000001",
            "firstName": "Test",
            "lastName": "User",
            "workEmail": "test.user@thegulocal.com"
        }
    ]
}
```

### DIFF
--- a/handlers/multiple-account-api/src/secondaryUserMeEndpoint.ts
+++ b/handlers/multiple-account-api/src/secondaryUserMeEndpoint.ts
@@ -10,14 +10,19 @@ import { getSubscription } from '@modules/zuora/subscription';
 import type { ZuoraAccount } from '@modules/zuora/types';
 import type { ZuoraClient } from '@modules/zuora/zuoraClient';
 
+const subscriptionWithPrimaryUserInformationSchema = z.object({
+	subscriptionName: z.string(),
+	firstName: z.string(),
+	lastName: z.string(),
+	workEmail: z.string(),
+});
+
+type SubscriptionWithPrimaryUserInformation = z.infer<
+	typeof subscriptionWithPrimaryUserInformationSchema
+>;
+
 const responseSchema = z.object({
-	primaryUsers: z
-		.object({
-			firstName: z.string(),
-			lastName: z.string(),
-			workEmail: z.string(),
-		})
-		.array(),
+	subscriptions: z.array(subscriptionWithPrimaryUserInformationSchema),
 });
 
 export async function secondaryUserMeEndpoint(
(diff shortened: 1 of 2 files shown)


handlers/multiple-account-api/src/secondaryUserMeEndpoint.ts
```diff
@@ -28,28 +33,32 @@ export async function secondaryUserMeEndpoint(
...
 	subscriptionName: string,
-) {
+): Promise<SubscriptionWithPrimaryUserInformation> {
 	const subscription = await getSubscription(zuoraClient, subscriptionName);
 	const account: ZuoraAccount = await getAccount(
 		zuoraClient,
 		subscription.accountNumber,
 	);
-	const { zipCode, ...billToContactWithoutZipCode } = account.billToContact;
+	const { zipCode, ...firstNameLastNameAndEmail } = account.billToContact;
```
maloky: this remaining object can change whiteout us knowing. Just wondering if we shouldn't be more explicit about what we want

maloky (approved):
looks goods 👍

handlers/multiple-account-api/src/secondaryUserMeEndpoint.ts
rupertbates: Yes probably right, I've updated it 👍

PR #3747, opened 2026-08-04
add script to drop iam policy for CODE

johnduffell:
just a quick convenience so we can clean up the "CODE" policy without the risk of accidentally dropping the wrong stack when clicking around in the console


handlers/iam-policies/README.md
```diff
@@ -16,15 +16,15 @@ Do a PR as normal and once in PROD, the policy will be updated for everyone gett
```
johnduffell, line 16: click here for a rendered diff 

Image

shtukas (approved):

PR #3748, opened 2026-08-04
Scrub non-tokenised payment methods on fully cancelled accounts

andrade0:
## What does this change?

Adds a scheduled job that strips the card data from non-tokenised card payment methods on Zuora accounts where every subscription is cancelled.

A `CreditCard` payment method in Zuora holds the card details themselves, unlike `CreditCardReferenceTransaction` which only holds a gateway token. There are around 39,000 of the former against 2.78 million of the latter. Once every subscription on an account is cancelled those details can never be used again, so there's no reason to keep them.

These aren't only a historical artefact. It's the same method type you get when card details are entered in Zuora directly, and around 200 to 250 are still created a month, with no sign of declining. About 88 percent of the recent ones land on an account that already existed, so they look like card updates rather than new signups. They only come into scope later, once the account they're on is fully cancelled, which is why this is a daily job rather than a one off cleanup.

Structure follows write-off-unpaid-invoices: a lambda queries BigQuery and drops the work list on S3, a distributed map feeds it one item at a time to a second lambda, and anything that fails ends up in an SNS message. Daily cron, PROD only.

## Scrub rather than delete

The card asked for deletion. Graham and I settled on Zuora's scrub endpoint instead so the payment records referencing the method stay intact, and testing in CODE turned up two harder reasons.

Delete is rejected on a default payment method, which about 11,000 of the targets are. And once the Zuora account itself is cancelled both the delete and the account update you'd use to detach it first are rejected, so there's no way through at all. Scrub works in both cases.

It also means we write nothing to any account. Where the scrubbed method was the default, Zuora clears `defaultPaymentMethodId` and turns `autoPay` off by itself, so an account is never left with auto pay on and nothing to charge. That's the largest group, 9,050 of them, and I'd assumed the opposite until I tested it.

Scrub does remove the payment method object, not just the card data. It stops being returned by the API and drops out of ZOQL. The payments survive and still carry the id, which is the bit we wanted.

Verified in CODE on A00187639 (cancelled account, default card with payments), A00181289 (audit trail), A00157083 (active account, default card) and A00078065 (autoPay true).

## Why the daily cap

The query is capped at 500 rows, oldest first. There are 19,744 in scope going back to 2015, most of them old, 6,000 from 2018 alone and only 121 created this year. On top of that a dozen to thirty a month keep arriving. So the same code drains the backlog over about six weeks and then handles the trickle, with no separate backfill.

The cap is not there for Zuora's sake: Zuora limits requests in flight, not requests per day, and the map runs one at a time. What it buys is a run short enough to finish inside the state machine's timeout, and small first PROD runs that are readable.

## How to test

`pnpm package` in the handler, `pnpm --filter cdk package` and the buildcheck snapshot all pass. knip is red on `main` already, but it reports nothing against this handler.

`DRY_RUN` is set to true on the lambda, so the first PROD run does every check and logs what it would do without touching Zuora. Worth watching one execution before flipping it in a follow-up.

Nothing is scrubbed on the strength of BigQuery alone. Fivetran syncs the Zuora tables once a day just after midnight, so by 6am the copy is already hours old. Every item is re-read from Zuora first and skipped if it no longer qualifies. That also makes re-runs safe, since a scrubbed method is no longer returned at all.

It is deployed on CODE and I ran it end to end, succeeded, 12m40s for 500 items. Worth knowing before you look at it: a CODE run can only exercise the plumbing. `datatech-fivetran.zuora` mirrors PROD Zuora and there's no sandbox equivalent, so the work list is PROD ids the sandbox has never heard of and every item skips. write-off-unpaid-invoices has the same constraint. Links to the state machine and that execution are in a comment below.

## Anything else

Needs one SSM parameter before it can run, at the standard config path so GuCDK grants the read access for free:

```
/<STAGE>/support/scrub-non-tokenised-payment-methods/gcpCredentialsConfig
```

It holds the workload identity federation config for BigQuery. Both CODE and PROD are already in place. `docs/bigquery-access.md` covers why the lambda's IAM role has a hand written short name and what breaks if it is renamed.

One deliberate difference from write-off-unpaid-invoices: this handler rethrows when items fail rather than swallowing everything. Swallowing leaves the distributed map's result file empty, so `CheckForFailures` never fires and neither does the alarm. Might be worth a ticket against the other lambda.


shtukas (approved):

johnduffell:
> Needs an SSM parameter /scrub-non-tokenised-payment-methods/<STAGE>/gcp-credentials-config

normally we use /STAGE/stack/app/* for config, the lambdas get that access for free with gucdk, and the appConfig.ts will load it into a typed (with zod) data structure.

cdk/bin/cdk.ts
```diff
@@ -298,6 +299,22 @@ new WriteOffUnpaidInvoices(app, 'write-off-unpaid-invoices-PROD', {
...
+	},
+);
+new ScrubNonTokenisedPaymentMethods(
+	app,
+	'scrub-non-tokenised-payment-methods-PROD',
+	{
+		stack: 'support',
+		stage: 'PROD',
+	},
+);
```
johnduffell: if you make it an SrStack instead of GuStack you can add it to the array on line 149, just one line instead of 15

cdk/lib/scrub-non-tokenised-payment-methods.ts
```diff
@@ -0,0 +1,359 @@
...
+				payload: TaskInput.fromObject({
+					filePath: JsonPath.format(
+						`executions/{}/${paymentMethodsFileName}`,
+						JsonPath.stringAt('$$.Execution.StartTime'),
+					),
+				}),
+			},
+		);
+
+		const scrubPaymentMethodsLambda = new GuLambdaFunction(
```
johnduffell: have you tried SrLambda instead here, it will give you some stuff for free

cdk/lib/scrub-non-tokenised-payment-methods.ts
```diff
@@ -0,0 +1,359 @@
...
+			{
+				lambdaFunction: new GuLambdaFunction(
+					this,
+					'GetPaymentMethodsToScrubLambda',
+					{
+						...lambdaDefaultConfig,
+						environment: {
+							GCP_CREDENTIALS_CONFIG_PARAMETER_NAME: `/${app}/${this.stage}/gcp-credentials-config`,
+							GCP_PROJECT_ID: `datatech-platform-${this.stage.toLowerCase()}`,
+							BUCKET_NAME: bucket.bucketName,
```
johnduffell: can these be constants in the code rather than the CDK?  then we find out about more issues at compile time rather than run time, and IDE support is better

cdk/lib/scrub-non-tokenised-payment-methods.ts
```diff
@@ -0,0 +1,359 @@
...
+		const bucket = new Bucket(this, 'Bucket', {
+			bucketName: `${app}-${this.stage.toLowerCase()}`,
+		});
+
+		const snsTopicArn = `arn:aws:sns:${this.region}:${this.account}:alarms-handler-topic-${this.stage}`;
+
+		const paymentMethodsFileName = 'payment-methods-to-scrub.json';
+
+		const lambdaRole = new Role(this, 'LambdaRole', {
+			roleName: `scrub-non-tok-pm-${this.stage}`, // Role name must be short to not break the authentication request to GCP
```
johnduffell: interesting - so we have somehow granted the lambda access to bigquery directly using the AWS IAM role?

cdk/lib/scrub-non-tokenised-payment-methods.ts
```diff
@@ -0,0 +1,359 @@
...
+						actions: ['secretsmanager:GetSecretValue'],
+						resources: [
+							`arn:aws:secretsmanager:${this.region}:${this.account}:secret:${this.stage}/Zuora-OAuth/SupportServiceLambdas-*`,
+						],
+					}),
+				],
+			},
+		);
+
+		const processPaymentMethods = new CustomState(
```
johnduffell: why CustomState rather than distributedmap https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_stepfunctions.DistributedMap.html

cdk/bin/cdk.ts
andrade0: Good shout, done - it's in the array now and it picks up the App tag too.

cdk/lib/scrub-non-tokenised-payment-methods.ts
andrade0: Hadn't spotted it, thanks. Both lambdas use it now, so the naming and source maps come for free and I only set the architecture and timeout.

cdk/lib/scrub-non-tokenised-payment-methods.ts
andrade0: Yeah that's better. They're all derived from the stage now, in constants the CDK imports as well, so there's one copy of each name and no more process.env with a bang after it.

cdk/lib/scrub-non-tokenised-payment-methods.ts
andrade0: No good reason, I'd copied the shape from write-off-unpaid-invoices. Swapped it for the L2 and it's much better - it works out the bucket, lambda and execution permissions itself, so the policy I was writing by hand is down to the SNS publish.

cdk/lib/scrub-non-tokenised-payment-methods.ts
andrade0: Yes, workload identity federation. GCP trusts the lambda's IAM role directly and hands back a short lived token, so there's no key to store. It's why the role has a hand written short name, the role name goes into the auth request and GCP rejects it past a certain length. The GCP side is set up in gcp-iac-terraform.

cdk/lib/scrub-non-tokenised-payment-methods.ts
```diff
@@ -0,0 +1,359 @@
...
+					'Key.$': JsonPath.stringAt('$.ResultWriterDetails.Key'),
+				},
+				ResultSelector: {
+					'Payload.$': JsonPath.stringToJson(JsonPath.stringAt('$.Body')),
+				},
+				OutputPath: '$.Payload',
+			},
+		});
+
+		const notifyTeam = new CustomState(
```
johnduffell: consider tasks.SnsPublish

cdk/lib/scrub-non-tokenised-payment-methods.ts
```diff
@@ -0,0 +1,359 @@
...
+
+		const stateMachine = new StateMachine(
+			this,
+			'ScrubNonTokenisedPaymentMethodsStateMachine',
+			{
+				stateMachineName: `${app}-${this.stage}`,
+				// 500 items processed one at a time, a few Zuora calls each. An hour
+				// is generous; past that something is stuck and we want to know
+				// rather than have the execution sit there until tomorrow's run.
+				timeout: Duration.hours(1),
```
johnduffell, line 290: how will we know if there's a timeout? it looks like there's an alarm for ExecutionsFailed but not ExecutionsTimedOut

cdk/lib/scrub-non-tokenised-payment-methods.ts
```diff
@@ -0,0 +1,359 @@
...
+
+		rule.addTarget(new SfnStateMachine(stateMachine));
+
+		new SrLambdaAlarm(
+			this,
+			'ScrubNonTokenisedPaymentMethodsStepFunctionFailureAlarm',
+			{
+				app,
+				metric: stateMachine.metricFailed({
+					period: Duration.minutes(5),
```
johnduffell: making it 1 minute means that the log link generated by the alarms handler will only cover the shorter time period, making it easier to find the offending execution in the logs

cdk/lib/scrub-non-tokenised-payment-methods.ts
andrade0: Done, and it pulled its weight - it grants the state machine sns:Publish itself, which was the last thing left in the policy I was writing by hand, so that whole block is gone now.

cdk/lib/scrub-non-tokenised-payment-methods.ts
```diff
@@ -0,0 +1,359 @@
...
+					statistic: 'Sum',
+				}),
+				threshold: 1,
+				evaluationPeriods: 1,
+				alarmDescription:
+					'The scheduled job that scrubs non-tokenised payment methods on fully cancelled accounts has failed. Login to the AWS console and debug the last execution.',
+				alarmName: `${this.stage}: ScrubNonTokenisedPaymentMethodsStepFunctionExecutionFailure`,
+				comparisonOperator:
+					ComparisonOperator.GREATER_THAN_OR_EQUAL_TO_THRESHOLD,
+				lambdaFunctionNames: scrubPaymentMethodsLambda.functionName,
```
johnduffell: this is used for the log links, the failure could also be getPaymentMethodsToScrub could it?  The property can also be an array e.g.:

```suggestion
				lambdaFunctionNames: [scrubPaymentMethodsLambda.functionName,getPaymentMethodsToScrub.functionName],
```

handlers/scrub-non-tokenised-payment-methods/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+Strips the card data from non-tokenised card payment methods on Zuora accounts
+where every subscription is cancelled.
+
+## Why
+
+A `CreditCard` payment method in Zuora holds the card details themselves. A
+`CreditCardReferenceTransaction` holds only a gateway token, and the card lives
+at Stripe or PayPal instead. We have roughly 39,000 of the former against 2.78
+million of the latter, so the non-tokenised ones are the tail end of an older
```
johnduffell: don't say former and latter, just say what they are

johnduffell: also correct the "older way of taking payments" as discussed - as this is still used

handlers/scrub-non-tokenised-payment-methods/README.md
```diff
@@ -0,0 +1,103 @@
...
+way of taking payments.
+
+Once every subscription on an account is cancelled, those card details can never
+be used again. Keeping them is pure risk with no upside.
```
johnduffell: Keeping them creates the risk of what?  accidentally charging the customer inappropriately?  Is this something special to non tokenised payment methods?
or is there a separate process for tokenised ones?

handlers/scrub-non-tokenised-payment-methods/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Scrub, not delete
+
+The card asked for deletion. We use Zuora's scrub endpoint instead, which strips
```
johnduffell: we don't need to talk about (asana) cards, this is a reference document to help people understand the current situation

handlers/scrub-non-tokenised-payment-methods/README.md
```diff
@@ -0,0 +1,103 @@
+# scrub-non-tokenised-payment-methods
+
+Strips the card data from non-tokenised card payment methods on Zuora accounts
+where every subscription is cancelled.
```
johnduffell: This handler is a state machine that runs daily to clear non-tokenised card data from Zuora accounts with no active subscriptions.

handlers/scrub-non-tokenised-payment-methods/README.md
johnduffell: I don't quite understand the relevance of CreditCardReferenceTransaction vs CreditCard, or the stats.  Do we need to mention it at all?

cdk/lib/scrub-non-tokenised-payment-methods.ts
andrade0: Good catch, that was a real hole. A timeout ends as TimedOut so ExecutionsFailed never sees it, which means the one case the hour timeout exists to catch was the one nobody would hear about. Alarm is now on failed + timedOut.

cdk/lib/scrub-non-tokenised-payment-methods.ts
andrade0: Makes sense, dropped it to a minute.

cdk/lib/scrub-non-tokenised-payment-methods.ts
andrade0: Yes it could, taken. I'd had the get lambda inline in the LambdaInvoke so I didn't have a handle on it, pulled it out into a variable and both names are in there now.

handlers/scrub-non-tokenised-payment-methods/README.md
```diff
@@ -0,0 +1,103 @@
...
+What scrub does not do is keep the payment method readable. Afterwards
+`GET /v1/payment-methods/{id}` answers `50000040 cannot be found` and the object
+disappears from ZOQL queries. The payment records survive and still carry the
+payment method id. Scrubbing twice fails with `50000020`.
+
+## What counts as a target
+
+- The payment method type is `CreditCard`, so not tokenised.
+- Its status is `Active`.
+- Every latest-version subscription on the account is `Cancelled`.
```
johnduffell, line 70: is that enough? CSRs can cancel subs forward dated to the end of the term, so we would still need to take payments.  I would recommend to wait until it's also after termEndDate for them all.

handlers/scrub-non-tokenised-payment-methods/README.md
```diff
@@ -0,0 +1,103 @@
...
+- The account has at least one subscription, so accounts that never had one are
+  left alone.
+
+Subscriptions are versioned in Zuora and each amendment creates a new version, so
+the query only looks at the latest version of each. An account with an active
+subscription and several old cancelled versions is not a target.
```
johnduffell: the old versions of the subscription are Expired, they can't be cancelled.
If an account has multiple subscriptions, you have covered it with "every [...] subscription on the account is cancelled"

handlers/scrub-non-tokenised-payment-methods/README.md
johnduffell:
```suggestion
- Every subscription on the account is `Cancelled`.
```

handlers/scrub-non-tokenised-payment-methods/README.md
```diff
@@ -0,0 +1,103 @@
...
+   message.
+
+### The daily cap
+
+The query is capped at 500 rows, oldest first. There is a historical backlog of
+around 20,000 payment methods going back to 2015, and an ongoing trickle of
+somewhere between four and fourteen a month. Capping each run means the same
+code drains the backlog over about six weeks and then quietly handles the
+trickle, with no separate backfill job. It also keeps the run comfortably inside
+Zuora's rate limit.
```
johnduffell: as far as I know zuora has a concurrency limit not a daily rate limit, so as long as you stick to one at a time that shouldn't be an limitation even running it constantly

handlers/scrub-non-tokenised-payment-methods/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+### Revalidation
+
+The work list comes from BigQuery, which lags Zuora by up to a sync interval. In
```
johnduffell:
```suggestion
The work list comes from BigQuery, which only contains Zuora data up to midnight. In
```

handlers/scrub-non-tokenised-payment-methods/README.md
andrade0: You're right and this was a real bug, thanks. I checked and there are 17,144 cancelled subs with a term end in the future, 147 of which are on accounts this job would have picked up today, so we'd have stripped the card while payments were still due. The query and the Zuora revalidation now both require the term to be over as well, with tests for the forward dated case.

handlers/scrub-non-tokenised-payment-methods/README.md
andrade0: Ah yes, I checked and every non-latest version is Expired, there isn't a single cancelled one. So my explanation was wrong. The filter is still needed but for the opposite reason: without it the Expired versions would fail the every-subscription-is-cancelled test and nothing would ever qualify. Reworded.

handlers/scrub-non-tokenised-payment-methods/README.md
andrade0: Taken as is, that's clearer than what I had.

handlers/scrub-non-tokenised-payment-methods/README.md
andrade0: Fixed, both types named.

handlers/scrub-non-tokenised-payment-methods/README.md
andrade0: Right, removed.

handlers/scrub-non-tokenised-payment-methods/README.md
andrade0: Rewritten to say it plainly: it's holding cardholder data we no longer have a use for, since a CreditCard method keeps the number and expiry inside Zuora. Not a wrong-charge risk, the account has nothing live left. And there's no equivalent for tokenised ones because there's nothing in Zuora to strip, only a gateway token. Whether those should be revoked at Stripe or PayPal is a separate question I haven't looked at.

cdk/lib/scrub-non-tokenised-payment-methods.ts
```diff
@@ -53,7 +55,7 @@ export class ScrubNonTokenisedPaymentMethods extends GuStack {
 			new PolicyStatement({
 				actions: ['ssm:GetParameter'],
 				resources: [
-					`arn:aws:ssm:${this.region}:${this.account}:parameter/${app}/${this.stage}/gcp-credentials-config`,
+					`arn:aws:ssm:${this.region}:${this.account}:parameter${gcpCredentialsConfigParameterName(stage)}`,
```
johnduffell: to keep it consistent with GuLambdafunction which gets it for free from ReadParametersByName‎:
https://github.com/guardian/cdk/blob/4f2cea261d10a225c34418d8155342efa52399cd/src/constructs/lambda/lambda.ts#L184-L187
you could reuse the standard policy rather than putting the config in a non standard name

andrade0: You're right, and it was already there even on the hand written role, I just hadn't put the parameter where GuCDK expects it. Moved it to /<STAGE>/support/<app>/gcpCredentialsConfig, dropped the policy, and switched to loadConfig so it's parsed through a zod schema rather than a process.env with a bang. Same answer to your comment on the description.

handlers/scrub-non-tokenised-payment-methods/README.md
andrade0: You're right, and the cap was never really about Zuora. The map runs one at a time so we're inside the in flight limit whatever the cap is. What it actually buys is a run that finishes inside the state machine's hour timeout, and small first PROD runs that are readable. Reworded here and in the two code comments that repeated it.

handlers/scrub-non-tokenised-payment-methods/README.md
andrade0: Taken. I checked the sync times too and it's exactly once a day just after midnight, so your wording is the accurate one.

andrade0:
On the SSM path, you were right, moved it to /<STAGE>/support/scrub-non-tokenised-payment-methods/gcpCredentialsConfig and it now goes through loadConfig with a zod schema, which also let me delete the policy I'd written by hand.

One thing to flag: the parameter already exists in CODE under the old path, so it needs recreating at the new one before the next CODE deploy, and creating in PROD.

cdk/lib/scrub-non-tokenised-payment-methods.ts
```diff
@@ -79,44 +81,35 @@ export class ScrubNonTokenisedPaymentMethods extends GuStack {
 					'logs:PutLogEvents',
 				],
 				resources: [
-					`arn:aws:logs:${this.region}:${this.account}:log-group:/aws/lambda/get-payment-methods-to-scrub-${this.stage}`,
-					`arn:aws:logs:${this.region}:${this.account}:log-group:/aws/lambda/get-payment-methods-to-scrub-${this.stage}:*`,
+					`arn:aws:logs:${this.region}:${this.account}:log-group:/aws/lambda/${getNameWithStage(this, 'get')}`,
```
johnduffell: the docs claim that you can add the actual built in roles yourself to get it
https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_lambda.Function.html#:~:text=The%20default%20Role,service%2Drole/AWSLambdaVPCAccessExecutionRole%22.

andrade0: Much better, done. AWSLambdaBasicExecutionRole attached to the role instead, and the hand written log group statements are gone. Slightly wider than what I had, it covers all log groups rather than just this lambda's, but it's the same policy CDK would have attached if I hadn't taken the role over.

handlers/scrub-non-tokenised-payment-methods/README.md
```diff
@@ -0,0 +1,157 @@
...
+
+The risk is holding cardholder data we have no use for. A `CreditCard` method
+keeps the card number and expiry inside Zuora, so they live in our systems, and
+in anything derived from them, for as long as the record does. Once the account
+is finished being billed that card can never be charged again, so there is
+nothing to weigh against keeping it.
+
+It is not about charging someone by mistake. The account has no live
+subscription left, and where the scrubbed method was the default Zuora switches
+auto pay off as well.
```
johnduffell: holding data itself is a hazard not  a risk.  the risk is what could happen as a result of that.  I think this bit can be a lot more concise.
The risk is surely that we either charge them when we shouldn't (this has happened in the past multiple times), that we could add a new subscription using old/defunct payment details, or worst case that there could be a breach of the data.
I don't know the answer why we're specifically doing this now, but presumably that came through in the original requirements.

handlers/scrub-non-tokenised-payment-methods/README.md
```diff
@@ -0,0 +1,157 @@
...
+
+There is no equivalent job for tokenised methods because there is nothing in
+Zuora to strip: the card sits at Stripe or PayPal and Zuora only holds a token.
+Whether those tokens should be revoked at the gateway is a separate question
+that this job does not answer.
```
johnduffell: I think we can just leave this out, unless we know of a job that we can point people at, it's redundant.

handlers/scrub-non-tokenised-payment-methods/README.md
```diff
@@ -0,0 +1,157 @@
...
+land on an account that already existed, so they look like card updates rather
+than new signups.
+
+What is old is the backlog currently in scope, because a card only becomes a
+target once every subscription on its account is cancelled, which happens long
+after the card was added. Hence 6,000 in scope from 2018 alone against 121
+created this year. A card added today simply becomes eligible later.
+
+That is why this is a scheduled job rather than a cleanup script: the stock is
+historical, but the source is live and will keep feeding it.
```
johnduffell:
```suggestion
`CreditCard` are added when details are entered directly into Zuora, which supports CSR assisted card updates.
Once established, this job will clean up the details as they become redundant, but before that there will be a backlog which will be cleared gradually over several months.
```

cdk/lib/scrub-non-tokenised-payment-methods.ts
```diff
@@ -0,0 +1,281 @@
...
+		const paymentMethodsFileName = 'payment-methods-to-scrub.json';
+
+		/*
+		 * This lambda is given an explicit role only so its name stays short
+		 * enough for the GCP auth request, which sends the role name. Passing a
+		 * role means CDK does not build one, so the basic execution policy it
+		 * would have attached is added here instead.
+		 */
+		const lambdaRole = new Role(this, 'LambdaRole', {
+			roleName: `scrub-non-tok-pm-${this.stage}`,
```
johnduffell, line 92: I think we need an explicit note here that this role name is referenced in the google cloud console, and where to check it/which team to contact, so that people know where to look if things break or something needs updating. 
If the same principle affects several lambdas then it can be a `.md` file somewhere and both places can link to it

handlers/scrub-non-tokenised-payment-methods/README.md
andrade0: Fair, hazard vs risk is the right distinction and I'd buried it in waffle. Cut down to the three actual risks you listed: charging someone we shouldn't, starting a subscription against defunct details, and the details being caught in a breach. I don't know why now either beyond the ticket asking for it, so I've not invented a reason.

handlers/scrub-non-tokenised-payment-methods/README.md
andrade0: Agreed, it was me answering a question nobody asked. Removed.

handlers/scrub-non-tokenised-payment-methods/README.md
andrade0: Taken as is, much better than mine.

cdk/lib/scrub-non-tokenised-payment-methods.ts
andrade0: Good point, and it does affect several: discount-expiry-notifier and negative-invoices-processor shorten their role name for exactly the same reason. So I've put it in docs/bigquery-access.md, saying the name is half a contract with the workload identity setup in gcp-iac-terraform, that renaming it breaks BigQuery access until that side matches, and to speak to Data Technology. All three now link to it.

cdk/lib/scrub-non-tokenised-payment-methods.ts
```diff
@@ -0,0 +1,281 @@
...
+
+		const getPaymentMethodsToScrub = new LambdaInvoke(
+			this,
+			'GetNonTokenisedPaymentMethodsOnCancelledAccounts',
+			{
+				lambdaFunction: getPaymentMethodsToScrubLambda,
+				payload: TaskInput.fromObject({
+					filePath: JsonPath.format(
+						`executions/{}/${paymentMethodsFileName}`,
+						JsonPath.stringAt('$$.Execution.StartTime'),
```
johnduffell, line 145: can you show a concrete screenshot/example of what the file key looks like?  I assume it's like `executions/2026-01-01T12:34:56Z/payment-methods-to-scrub.json` but it would be nice to know

cdk/lib/scrub-non-tokenised-payment-methods.ts
```diff
@@ -0,0 +1,281 @@
...
+				toleratedFailurePercentage: 100,
+				itemReader: new S3JsonItemReader({
+					bucket,
+					key: JsonPath.format(
+						`executions/{}/${paymentMethodsFileName}`,
+						JsonPath.stringAt('$$.Execution.StartTime'),
+					),
+				}),
+				itemBatcher: new ItemBatcher({ maxItemsPerBatch: 1 }),
+				resultWriter: new ResultWriter({
```
johnduffell: apparently this is deprecated https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_stepfunctions.ResultWriter.html#:~:text=aws_stepfunctions%20%C2%BB%20ResultWriter-,%E2%9A%A0%EF%B8%8F%20Deprecated%3A%20use%20%7B%40link%20ResultWriterV2%20%7D%20instead,-Configuration%20for%20writing
and we should use the V2 version now

cdk/lib/scrub-non-tokenised-payment-methods.ts
```diff
@@ -0,0 +1,281 @@
...
+						// follow-up once a PROD run has been eyeballed.
+						DRY_RUN: 'true',
+					},
+					initialPolicy: [
+						new PolicyStatement({
+							actions: ['secretsmanager:GetSecretValue'],
+							resources: [
+								`arn:aws:secretsmanager:${this.region}:${this.account}:secret:${this.stage}/Zuora-OAuth/SupportServiceLambdas-*`,
+							],
+						}),
```
johnduffell: you can reuse the existing one from policies.ts

cdk/lib/scrub-non-tokenised-payment-methods.ts
```diff
@@ -0,0 +1,281 @@
...
+					],
+				},
+			},
+		);
+
+		const processPaymentMethods = new DistributedMap(
+			this,
+			'ProcessPaymentMethodsInDistributedMap',
+			{
+				comment: `MaxConcurrency is 1 to stay inside Zuora's limit on requests in flight. ToleratedFailurePercentage is 100 so one bad account does not stop the rest of the run`,
```
johnduffell, line 198: nice, I didn't know you can do comments like that

cdk/lib/scrub-non-tokenised-payment-methods.ts
```diff
@@ -0,0 +1,281 @@
...
+						`executions/{}/${paymentMethodsFileName}`,
+						JsonPath.stringAt('$$.Execution.StartTime'),
+					),
+				}),
+				itemBatcher: new ItemBatcher({ maxItemsPerBatch: 1 }),
+				resultWriter: new ResultWriter({
+					bucket,
+					prefix: JsonPath.format(
+						`executions/{}`,
+						JsonPath.stringAt('$$.Execution.StartTime'),
```
johnduffell, line 213: presumably this is to put it alongside the input file in the same "directory".  Might be worth extracting that in some way.
Also it would be good to see an example of what the file looks like, it looks like there are separate files for successes/failures, so presumably we could use them for diagnosing issues if the alarm fires or payment methods have been removed when they shouldn't?

cdk/lib/scrub-non-tokenised-payment-methods.ts
```diff
@@ -0,0 +1,281 @@
...
+import type { SrStageNames } from './cdk/SrStack';
+import { SrStack } from './cdk/SrStack';
+
+export class ScrubNonTokenisedPaymentMethods extends SrStack {
+	constructor(scope: App, stage: SrStageNames) {
+		super(scope, { stage, app: APP });
+
+		const bucket = new Bucket(this, 'Bucket', {
+			bucketName: bucketName(stage),
+		});
```
johnduffell, line 74: is this bucket pre-created manually? what's the expiry/retention time set to, (for GDPR purposes)

handlers/scrub-non-tokenised-payment-methods/README.md
johnduffell: I don't suggest you have or should invent anything, I just mean we're trying to help customers, not complete tickets.  If we don't know the real reason now, we're even less likely to know in future, but now is the best time to find out and document it.
I did specifically say that I don't know the answer, so changing it to the "actual risks I listed" doesn't make any sense without confirming either way.

cdk/lib/scrub-non-tokenised-payment-methods.ts
andrade0: Your guess was right, it's the execution start time with millis: executions/2026-08-07T08:50:15.556Z/payment-methods-to-scrub.json. I've added a section to the README with the full layout from a real CODE run, including the map run subfolder, the manifest and the result files, plus what's in each. Ids in there are made up since this repo is public.

cdk/lib/scrub-non-tokenised-payment-methods.ts
andrade0: Good spot, moved to ResultWriterV2. It's behind a feature flag, so rather than turn it on in cdk.json for the whole repo I've set the context on this stack only. Worth saying the synthesised template comes out byte for byte identical, so GetDistributedMapResult still finds ResultWriterDetails and nothing changes at runtime.

cdk/lib/scrub-non-tokenised-payment-methods.ts
andrade0: Didn't know that existed, thanks. Swapped for AllowZuoraOAuthSecretsPolicy, same ARN, and it's now added via addPolicies rather than initialPolicy.

cdk/lib/scrub-non-tokenised-payment-methods.ts
andrade0: The comment prop on DistributedMap, it ends up in the definition so it shows in the console on the state itself, which felt like the right place for it.

modules/zuora/src/subscription.ts
```diff
@@ -71,17 +71,32 @@ export async function getSubscription<T extends z.ZodType>(
...
 	zuoraClient: ZuoraClient,
 	accountNumber: string,
-): Promise<ZuoraSubscription[]> => {
+): Promise<ZuoraSubscription[]>;
+export async function getSubscriptionsByAccountNumber<T extends z.ZodType>(
+	zuoraClient: ZuoraClient,
+	accountNumber: string,
+	schema: T,
+): Promise<Array<z.infer<T>>>;
+export async function getSubscriptionsByAccountNumber<T extends z.ZodType>(
```
andrade0, line 78: Self review, flagging because it's a shared module: this is the one existing signature I've touched. It's an overload, so the two arg call keeps the exact same code path and the same schema. Only caller outside this handler is modules/sync-supporter-product-data, which calls it with two args, and braze-acquisition-events-sync passes type-check/lint/test with the ... filter. Say if you'd rather I left the shared function alone and parsed narrowly on my side.

handlers/scrub-non-tokenised-payment-methods/src/constants.ts
```diff
@@ -0,0 +1,91 @@
+/*
+ * The CDK imports this file so the resource names are written once and both
+ * sides get the same ones. It cannot resolve the @modules aliases, so the stage
+ * is spelled out here rather than imported from @modules/stage. It is the same
+ * union, so a Stage from the lambda and an SrStageNames from the CDK both fit.
+ */
+type Stage = 'CODE' | 'PROD';
```
andrade0: Self review, this one's a wart. The CDK imports this file so the resource names are written once, but the cdk tsconfig can't resolve the @modules aliases, so Stage is spelled out here instead of imported from @modules/stage. Same union, so both a Stage and an SrStageNames fit. Open to a better idea.

cdk/lib/scrub-non-tokenised-payment-methods.ts
```diff
@@ -0,0 +1,293 @@
...
+				lambdaFunction: scrubPaymentMethodsLambda,
+				outputPath: '$.Payload',
+			}),
+			{
+				mode: ProcessorMode.DISTRIBUTED,
+				executionType: ProcessorType.STANDARD,
+			},
+		);
+
+		const getMapResult = new CustomState(this, 'GetDistributedMapResult', {
```
andrade0, line 228: Self review, since you asked about CustomState earlier: this one stays hand written because there's no L2 for reading the map's result file back. DistributedMap, SnsPublish and the rest are all constructs now, this is the only bit of raw state JSON left.

handlers/scrub-non-tokenised-payment-methods/src/services/scrubBatch.ts
```diff
@@ -0,0 +1,60 @@
...
+				error,
+			);
+		}
+	}
+
+	console.log(
+		`Batch finished: ${scrubbed} scrubbed, ${wouldScrub} would have been scrubbed, ${skipped} skipped, ${failures.length} failed`,
+	);
+
+	if (failures.length > 0) {
```
andrade0, line 53: Self review, deliberate difference from write-off-unpaid-invoices, which swallows failures. Swallowing leaves the map's result file empty, so CheckForFailures never fires and neither does the alarm. Might be worth a ticket against that one.

cdk/lib/scrub-non-tokenised-payment-methods.ts
```diff
@@ -0,0 +1,293 @@
...
+
+		const stateMachine = new StateMachine(
+			this,
+			'ScrubNonTokenisedPaymentMethodsStateMachine',
+			{
+				stateMachineName: `${APP}-${this.stage}`,
+				// 500 items processed one at a time, a few Zuora calls each. An hour
+				// is generous; past that something is stuck and we want to know
+				// rather than have the execution sit there until tomorrow's run.
+				timeout: Duration.hours(1),
```
andrade0, line 290: Self review, one of the bits I'm least sure of. The hour is a judgement call, not a measurement I had at the time. Since then a full 500 item run in CODE took 12m40s, so an hour is generous but not silly. Happy to bring it down if you'd rather it failed faster.

cdk/lib/scrub-non-tokenised-payment-methods.ts
```diff
@@ -0,0 +1,293 @@
...
+			{
+				nameSuffix: 'scrub',
+				lambdaOverrides: {
+					...lambdaDefaults,
+					memorySize: 512,
+					handler: 'scrubPaymentMethods.handler',
+					environment: {
+						// Log what would happen without touching Zuora. Flip to false in a
+						// follow-up once a PROD run has been eyeballed.
+						DRY_RUN: 'true',
```
andrade0: Self review, and the other thing I'm least sure of. Nobody asked for this, it's my own caution: scrub is irreversible and CODE proves nothing here, since the work list is PROD ids the sandbox doesn't have so every item skips. Plan was one PROD run to read the log, then a one line follow up. If you think that's overkill I'll drop it and just lower the cap for the first run instead.

andrade0:
Self reviewed the whole thing and left pre-emptive notes on the unusual bits, plus the two decisions I am least sure of (the state machine timeout and the dry run), so you can go straight at those.

If it helps to see it run rather than read it, it is deployed on CODE and I ran it end to end this morning, succeeded, 12m40s for 500 items:

- State machine: https://eu-west-1.console.aws.amazon.com/states/home?region=eu-west-1#/statemachines/view/arn:aws:states:eu-west-1:865473395570:stateMachine:scrub-non-tokenised-payment-methods-CODE
- That execution: https://eu-west-1.console.aws.amazon.com/states/home?region=eu-west-1#/v2/executions/details/arn:aws:states:eu-west-1:865473395570:execution:scrub-non-tokenised-payment-methods-CODE:4ba7da68-41d1-48cd-8854-b527b093667b

Worth knowing before you look: every item skips in CODE. The work list comes from the Fivetran mirror of PROD Zuora and there is no sandbox equivalent, so the ids are PROD ones the sandbox has never heard of and they come back 50000040. That is the intended skip path rather than a failure, and it is what the whole run will look like there. The 7 that skip with "subscription(s) that are not cancelled" are real sandbox accounts, so that is the revalidation doing its job.

The README now has the S3 layout and file contents from that run, and docs/bigquery-access.md covers the role name contract with GCP.

handlers/scrub-non-tokenised-payment-methods/src/services/stillBillingReason.ts
```diff
@@ -0,0 +1,44 @@
...
+ * Returns undefined when the account is finished, or what is still billing when
+ * it is not, so the caller can say why it skipped.
+ *
+ * The three clauses below are the same three the BigQuery query uses to build
+ * the work list: at least one subscription, all of them cancelled, none of them
+ * still in term. That query cannot call this function, so the rule does exist
+ * twice, once here and once as SQL in PAYMENT_METHODS_TO_SCRUB_QUERY. Change one
+ * and you have to change the other.
+ */
+export const stillBillingReason = (
```
andrade0, line 17: Self review, added after your comments today. The rule was spread over three checks in processPaymentMethod, and it is also the three HAVING clauses in the BigQuery query. They cannot share code across the language boundary so the duplication is real, but the TypeScript side has one home now and both sides point at each other. Existing tests passed unchanged, so the behaviour and the log lines are identical.

PR #3749, opened 2026-08-05
Allow sending emails with local dev role

rupertbates:
## What does this change?
The local development role used to run membership applications locally needs permission to access the braze emails SQS queue if we are going to be able to test sending emails locally.


shtukas (approved):

johnduffell (approved):
thanks for testing that and adding the policy statement 👍

PR #3752, opened 2026-08-06
Multiple accounts emails - invitation email

rupertbates:
## What does this change?
This PR follows on from #3750 and adds the email which is sent when an invitation is redeemed by a secondary user.

### Note that at this moment in time this email can't be sent because it is not completely set up in Braze.

---
  - https://app.asana.com/0/0/1217160435359543


modules/email/src/email.ts
```diff
@@ -97,6 +98,28 @@ type RecursiveValues<T> = T extends object
...
+): EmailMessageWithUserId {
+	return {
+		To: {
+			Address: emailAddress,
+			ContactAttributes: {
+				SubscriberAttributes: subscriberAttributes,
+			},
+		},
+		DataExtensionName: dataExtensionName,
+		...user,
```
tjmw, line 119: Is this what's used to target the user in Braze? Or is is the `To.Address` above?

handlers/multiple-account-api/test/invitation.test.ts
```diff
@@ -73,6 +86,9 @@ const mockIdentityClient = {} as IdentityClient;
 describe('createInvitationHandler', () => {
 	beforeEach(() => {
 		jest.clearAllMocks();
+		mockSendEmail.mockResolvedValue(
```
tjmw, line 89: Do we want to assert that this was invoked as expected?

handlers/multiple-account-api/src/multipleAccountEmails.ts
```diff
@@ -0,0 +1,42 @@
...
+			: 'https://support.code.dev-theguardian.com';
+	return {
+		acceptInvitationUrl: `${baseUrl}/invitation/accept/${invitationCode}`,
+		rejectInvitationUrl: `${baseUrl}/invitation/reject/${invitationCode}`,
+	};
+}
+
+export async function sendInvitationEmail(
+	stage: Stage,
+	identityId: string,
```
tjmw: I think this is the identity ID of the secondary user? Would it be worth naming it to make that clear, like we do for the email fields? I.e. `secondaryUserIdentityId`. Or is that obvious already?

tjmw (approved):
👍 This looks great. Just a few minor comments/questions to take or leave.

handlers/multiple-account-api/src/multipleAccountEmails.ts
```diff
@@ -0,0 +1,42 @@
+import {
+	buildEmailMessage,
+	DataExtensionNames,
+	sendEmail,
+} from 'email/src/email';
```
maloky: shouldn't we be getting this from '@modules/email'?

modules/email/src/email.ts
rupertbates: It's complicated. I've spent quite a while trying to understand this by looking at membership-workflow - I think essentially you need to have a valid identity id or sf contact id in the user details and then membership-workflow will update its email address to whatever you put in the To.Address field. 

However if it differs to what was there already it may or may not work based on whether we have an sf contact record for the user.

rupertbates: I've put a readme describing what I've found in this PR: https://github.com/guardian/membership-workflow/pull/716

PR #3757, opened 2026-08-10
Add primary user data to invitation table

maloky:
## What does this change?
Add two new fields to invitation record so we don't have to make further  requests to Zuora:
- Primary user first name
- Primary user email

Create a invitation and check dynamoDB table to see new fields being stored.

Since invitation records only last for 1 month, we do not expect this to be an issue.


rupertbates (approved):
Looks good 👍

PR #3758, opened 2026-08-11
Turn off dry run for the scrub job

andrade0:
### What does this change?

The scrub job has been in dry run since it was deployed. The first PROD run went out this morning and got through all 500 payment methods: 500 would have been scrubbed, none skipped, none failed. So this turns dry run off and lets it do the work.

CODE stays in dry run. The work list is read from the mirror of PROD, so none of it exists in the CODE Zuora tenant and every item is skipped there. Left live, the no-progress check would fail every CODE run for a reason that says nothing about the code.

There is also the tidy-up John asked for on #3748. The executions prefix was written out six times and the key the map reads back was built twice, so both are named once now. That is the first commit, and it is worth reading on its own: it changes no snapshot at all.

### How has this change been tested?

Gates pass locally. The synthesised template says the rest better than I can. The only difference in PROD is DRY_RUN on the three lambdas, and CODE is byte for byte unchanged, which is what makes the prefix change a pure rename.

Tomorrow's run should scrub 500, and the backlog should drop from 19,612 as it stood this morning to around 19,112. Scrubbed cards stop coming back from the query, so if that number does not move, the no-progress check fails the run and the alarm fires.

At 500 a day against 4 to 6 becoming eligible, the backlog should clear in about 40 days.

### Have we considered potential risks?

This is the first time the no-progress check is live, since it is skipped in dry run. It has never actually run, so tomorrow morning is its first real exercise.

The scrub is not reversible. What limits it is that every item is re-read from Zuora before it is touched, and the run is capped at 500 a day, so a bad day costs 500 rows rather than the whole backlog.

### DIFF
--- a/cdk/lib/__snapshots__/scrub-non-tokenised-payment-methods.test.ts.snap
+++ b/cdk/lib/__snapshots__/scrub-non-tokenised-payment-methods.test.ts.snap
@@ -1502,7 +1502,7 @@ exports[`The ScrubNonTokenisedPaymentMethods stack matches the snapshot 2`] = `
         "Environment": {
           "Variables": {
             "APP": "scrub-non-tokenised-payment-methods",
-            "DRY_RUN": "true",
+            "DRY_RUN": "false",
             "NODE_OPTIONS": "--enable-source-maps",
             "STACK": "support",
             "STAGE": "PROD",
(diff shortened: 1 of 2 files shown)


graham228221 (approved):
Logs look good, let's go!

PR #3759, opened 2026-08-11
Send invitation redeem confirmation email

maloky:
## What does this change?
Sends a redeem confirmation email after invitation acceptance to the secondary user.

Currently this email can't be tested  because it is not completely set up in Braze.


rupertbates (approved):
Nice 👍

PR #3760, opened 2026-08-11
Store the invitation code on the secondary user

rupertbates:
## What does this change?
The data design team have asked us to add the invitation code as an additional field on a secondary user record. This is so that they can distinguish between records if the user is removed and then reinvited.

## [Asana Ticket](https://app.asana.com/1/1210045093164357/project/1213112839739792/task/1217280945466071)


maloky (approved):
LG 👍

PR #3762, opened 2026-08-11
add permission to send email for multiple-accounts-api

maloky:
## What does this change?

Adds `braze-emails` to SQSSendPolicy in order to send email from multiple-accounts-api.
  
Tested on CODE and no errors were thrown. Got a 200 Status Code  with invitation code in the response and received the email.


rupertbates (approved):
👍

PR #3763, opened 2026-08-11
Move credit processors to Zuora Orders

andrade0:
## What does this change?

Holiday stop and delivery problem credits share the same credit processor and currently update subscriptions through the legacy subscription update endpoint. This moves that shared write path to the asynchronous Zuora Orders API, so both processors continue to create the same credit rate plans and write the resulting charge details back to Salesforce.

Each Order adds the existing one-time credit rate plan and preserves the credit amount, all three trigger dates, and the `HolidayStart__c` and `HolidayEnd__c` fields. When a credit falls beyond the current subscription term, a `TermsAndConditions` action extends the term before the `AddProduct` action.

The processor waits for both the asynchronous job and its nested Order to complete, then reloads the subscription before writing the charge details to Salesforce. A temporary failure while reading the job status is retried until the time budget expires. A submission is never retried because Zuora may have accepted it before the response was lost. When Zuora explicitly reports locking contention, the processor waits one minute then retries once if enough of the overall budget remains.

Before each serial credit on a subscription, the processor calculates how much Lambda time remains. It reserves time to prepare the Order and record the result, caps the whole Order workflow at eleven minutes, and leaves work in Salesforce for the next scheduled run when there is too little time to monitor it safely.

The handler descriptions and README files now refer to credits rather than amendments.

This is a representative Delivery credit Order payload:

```json
{
  "orderDate": "2026-08-12",
  "existingAccountNumber": "A0123456",
  "subscriptions": [{
    "subscriptionNumber": "A-0123456",
    "orderActions": [{
      "type": "AddProduct",
      "triggerDates": [
        {"name": "ContractEffective", "triggerDate": "2026-09-01"},
        {"name": "ServiceActivation", "triggerDate": "2026-09-01"},
        {"name": "CustomerAcceptance", "triggerDate": "2026-09-01"}
      ],
      "addProduct": {
        "productRatePlanId": "8ad08ad58aff62ad018aff6fbea80016",
        "chargeOverrides": [{
          "productRatePlanChargeId": "8ad08ad58aff62ad018aff6fc5fd005e",
          "customFields": {
            "HolidayStart__c": "2026-08-14",
            "HolidayEnd__c": "2026-08-14"
          },
          "pricing": {"oneTimeFlatFee": {"listPrice": -2.42}}
        }]
      }
    }]
  }],
  "processingOptions": {"runBilling": false, "collectPayment": false}
}
```

The shared Zuora code, holiday stop processor and delivery problem processor test suites passed locally with Java 21. The tests compare the complete Orders JSON and cover successful completion, failed and incomplete Orders, missing job IDs, unknown statuses, the wall-clock deadline, request timeouts, locking contention, a completed status response at the deadline, a temporary status-read failure and deferred work.

Earlier revisions of the branch were deployed to CODE. Holiday Stop covered a normal credit and a term extension before adding a credit. A fresh Delivery request created one -2.42 charge, populated its Salesforce record and marked it actioned. Running the Delivery processor again returned no results and created no duplicate charge.

The subscription REST response reports `amendmentType` as null for the test Holiday credit, while the ZOQL `RatePlan.AmendmentType` read by fulfilment is `NewProduct`. It therefore passes fulfilment’s existing `!= 'RemoveProduct'` condition.

The CI run has started. Once it is green, both processors will be redeployed to CODE and the end-to-end tests will be repeated.

The CODE credit tests should produce exactly one charge for each request, with the expected negative price, credit dates and Salesforce charge code. A second invocation should produce no charge. CloudWatch logs will show completed order jobs and any credits deferred because the run is close to its timeout.

The shared change affects both credit processors, so both will be redeployed to CODE. Unexpected terminal response states fail explicitly and the existing duplicate check remains in place. Only an explicit locking rejection is safe to retry as a new submission.

Requests remain sequential within a subscription and use up to 20 subscriptions in parallel, leaving room below Zuora’s default tenant-wide limit of 40 concurrent requests. Each HTTP request has an explicit maximum two-minute read timeout. A status read can be retried within the available budget, but an Order is never resubmitted after an ambiguous submission response.

There are no IAM or personal-data changes.

## Images

Not applicable, no UI changes.

## Accessibility

Not applicable, no UI changes.


shtukas (approved):

graham228221 (approved):
Looks good 👏

PR #3770, opened 2026-08-20
Remove name fields from mma-primary endpoint

tjmw:
## What does this change?

This PR removes `firstName`, `lastName` and `displayName` from the `secondaryUsers` list in the response from the mma-primary endpoint.  

This is for privacy reasons. Actually this behaviour was broken already because the token we use to auth with IDAPI does not allow inclusion of the privateFields property, which `firstName` and `lastName` are nested under.

Note: includes a change to update the IDAPI user response schema to make the `privateFields` property optional. I think this is the right way to model this since it may or may not be there depending on the permissions of the access token used to auth with IDAPI. This had one small knock on effect in braze-acquisition-events-sync, so I'll ask MarTech to check that to make sure they're happy.

This has been deployed to CODE and verified by calling the endpoint from MMA. Previously this was broken when the response contained secondary users who had accepted an invitiation.

### DIFF
--- a/modules/identity/src/idapi.ts
+++ b/modules/identity/src/idapi.ts
@@ -15,6 +15,13 @@ const identityUserByIdSchema = z.object({
 	id: z.string(),
 	primaryEmailAddress: z.string().optional(),
 	publicFields: z.object({ displayName: z.string() }),
+	privateFields: z
+		.object({
+			brazeUuid: z.string().optional(),
+			firstName: z.string().optional(),
+			secondName: z.string().optional(),
+		})
+		.optional(),
 });
 
 type IdentityUserById = z.infer<typeof identityUserByIdSchema>;
--- a/modules/identity/src/idapi.ts
+++ b/modules/identity/src/idapi.ts
@@ -3,38 +3,31 @@ import { z } from 'zod';
 import { RestClientError } from '@modules/zuora/restClient';
 import type { IdentityClient } from './identityClient';
 
-const identityUserSchema = z.object({
+const identityUserByEmailSchema = z.object({
 	id: z.string(),
 	primaryEmailAddress: z.string(),
 	publicFields: z.object({ displayName: z.string() }),
 });
 
-const identityUserWithPrivateFieldsSchema = z.object({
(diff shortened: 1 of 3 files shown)


modules/identity/src/idapi.ts
```diff
@@ -3,39 +3,32 @@ import { z } from 'zod';
...
+const identityUserByEmailSchema = z.object({
 	id: z.string(),
 	primaryEmailAddress: z.string(),
 	publicFields: z.object({ displayName: z.string() }),
 });
 
-const identityUserWithPrivateFieldsSchema = z.object({
+type IdentityUserByEmail = z.infer<typeof identityUserByEmailSchema>;
+
+const identityUserByIdSchema = z.object({
```
tjmw: I'm trying to establish if this could be merged with `identityUserByEmailSchema` above. The only difference is that `primaryEmailAddress` address is optional in this one which I'd like to confirm.

modules/identity/src/idapi.ts
```diff
@@ -67,7 +60,7 @@ export const getUserByEmail = async (
 export const getUserByIdentityId = async (
 	client: IdentityClient,
 	identityId: string,
-): Promise<IdentityUserWithPrivateFields | undefined> => {
+): Promise<IdentityUserById | undefined> => {
```
tjmw, line 70: This is a breaking change as the `privateFields` property was previously non-optional and it's now been removed. But the mma-primary endpoint appears to be the only caller so I think that's okay.

tjmw: > But the mma-primary endpoint appears to be the only caller so I think that's okay.

This is wrong - there's a build failure because `braze-acquisition-events-sync` expected privateFields to be available. For some reason my VSCode didn't find that usage.

tjmw: Fixed in 996f1358f6023baac44ec79e3797649c648d0b36

AnastasiiaBalenko (approved):
Looks good to me

graham228221 (approved):

PR #3771, opened 2026-08-20
Move Turkey (TR) from International to Europe country group

juabara:
Turkish visitors were getting inconsistent region cookies — one resolved to Turkey, the other fell back to the default (Germany). Business decided Turkey belongs in Europe, so this moves `TR` from `International` to `EURCountries` in `countryGroup.ts`.

`supportRegionIdFromCountry('TR')` now returns `SupportRegionId.EU` instead of `INT`, meaning Turkey gets EUR currency and EU support routing.


paulbrown1982 (approved):
👍

PR #3774, opened 2026-08-21
Fix Braze acquisition event print product field mapping

juabara:
Corrects the model to read `printOptions.product` (matching the actual EventBridge event shape) instead of the non-existent flat `printProduct` field, so Newspaper and Guardian Weekly print subscriptions are now differentiated correctly in Braze.


AnastasiiaBalenko (approved):
👍

PR #3775, opened 2026-08-24
Keep Orders within each Lambda's time budget

andrade0:
### What does this change?

The discount and product-move API Gateway handlers now submit synchronous Zuora Orders with an eighteen-second request timeout. This avoids holding an API request open while a slow asynchronous Zuora job completes.

Delivery-problem credits remain asynchronous background work, but can wait for an Order for up to thirteen minutes and fifty seconds. The existing seventy-second buffer is retained for finishing the run and writing the result.

### How has this change been tested?

Tested locally with the discount-api type check, lint, formatting and unit tests; product-move unit tests; delivery-problem unit tests and assembly; and the CDK test and synthesis suite.

The API Gateway timeout and 5XX logs should no longer show requests waiting on asynchronous Zuora jobs. The delivery-problem processor should complete slow Orders without exhausting its Lambda runtime.

### Have we considered potential risks?

A synchronous Order whose outcome is unknown is not retried, since Zuora may already have accepted it. Non-completed responses still fail clearly. Delivery-problem retains its single retry for explicit Zuora locking contention, within the overall Lambda budget.


handlers/discount-api/src/discountEndpoint.ts
```diff
@@ -30,7 +30,7 @@ const postOrderWorkBufferInMilliseconds = 10_000;
 const minimumOrderDurationInMilliseconds = 10_000;
 /**
  * API Gateway gives a REST API integration at most 29 seconds. The eighteen
- * second order budget leaves ten seconds for the billing preview and confirmation
+ * second synchronous order budget leaves ten seconds for the billing preview and confirmation
```
graham228221:
```suggestion
 * second order budget leaves ten seconds for the billing preview and confirmation
```

I wouldn't worry about mentioning this - synchronous requests are the norm

handlers/delivery-problem-credit-processor/src/main/scala/com/gu/deliveryproblemcreditprocessor/DeliveryCreditProcessor.scala
```diff
@@ -26,9 +26,14 @@ import zio.{RIO, Task, ZIO}
...
 import java.time.{DayOfWeek, LocalDate, LocalDateTime}
 import scala.concurrent.ExecutionContext
+import scala.concurrent.duration.{FiniteDuration, MINUTES, SECONDS}
 
 object DeliveryCreditProcessor extends Logging {
 
+  /** The 15-minute Lambda leaves the shared processor's 70-second completion and setup buffer.
+    */
+  private val MaximumOrderDuration: FiniteDuration = FiniteDuration(13, MINUTES) + FiniteDuration(50, SECONDS)
+
```
graham228221, line 36: This Lambda may make multiple requests within a single run. So ideally you might reduce this for each subsequent request depending on how long the previous request(s) took. 

E.g. 

- Request 1 has 13 minutes limit | completes in 10 seconds
- Request 2 has 12 minutes 50 seconds limit | completes in 12 seconds
- Request 3 has 12 minutes 38 seconds limit | completes in 5 seconds
- Request 4 has 12 minutes 33 seconds limit... etc etc

handlers/product-move-api/src/main/scala/com/gu/productmove/zuora/ZuoraCancel.scala
```diff
@@ -232,40 +134,7 @@ object CancellationPolicy {
...
-
-object AsyncJobStatus {
-  given JsonDecoder[AsyncJobStatus] = JsonDecoder[String].mapOrFail {
-    case "Processing" => Right(AsyncJobStatus.Processing)
-    case "Failed" => Right(AsyncJobStatus.Failed)
-    case "Completed" => Right(AsyncJobStatus.Completed)
-    case status => Left(s"Unknown Zuora async job status: $status")
-  }
-}
+case class SyncOrderResult(status: OrderStatus) derives JsonDecoder
```
graham228221:
```suggestion
case class OrderResult(status: OrderStatus) derives JsonDecoder
```

As mentioned in another comment, sync requests are the norm so just `OrderResult` is probably more appropriate. 

Someone looking at this in the future might be thinking "what is a SyncOrder?"

graham228221 (commented):
Generally looks good but just a couple of thoughts/suggestions

shtukas (approved):

handlers/delivery-problem-credit-processor/src/main/scala/com/gu/deliveryproblemcreditprocessor/DeliveryCreditProcessor.scala
andrade0: That is already how it behaves. The remaining Lambda time is read immediately before each credit, then capped at 13m50 and reduced by the 70-second setup and completion buffer. Once there is too little left, the remaining credits are deferred to the next run.

graham228221: Awesome, I wasn't sure - thanks for confirming!

handlers/delivery-problem-credit-processor/src/main/scala/com/gu/deliveryproblemcreditprocessor/DeliveryCreditProcessor.scala
```diff
@@ -26,9 +26,14 @@ import zio.{RIO, Task, ZIO}
 
 import java.time.{DayOfWeek, LocalDate, LocalDateTime}
 import scala.concurrent.ExecutionContext
+import scala.concurrent.duration.{FiniteDuration, MINUTES, SECONDS}
 
 object DeliveryCreditProcessor extends Logging {
 
+  /** The 15-minute Lambda leaves the shared processor's 70-second completion and setup buffer.
+    */
+  private val MaximumOrderDuration: FiniteDuration = FiniteDuration(13, MINUTES) + FiniteDuration(50, SECONDS)
```
johnduffell, line 35: for future, normally we use the implicit form by adding `import scala.concurrent.duration._`
```suggestion
  private val MaximumOrderDuration: FiniteDuration = 13.minutes + 50.seconds
```

Also I'd probably have just rounded to 13 minutes, for simplicity, as it's not that critical I assume!

PR #3778, opened 2026-08-24
Add supporter-product-data-CODE SQS queue to local dev policy

tjmw:
## What does this change?

Adds the `supporter-product-data-CODE` SQS queue to local dev policy.

This is needed when running the multiple-account-api integration tests locally.


rupertbates (approved):
👍

PR #3781, opened 2026-08-25
Add endpoint to get secondary users by identity id

rupertbates:
## What does this change?
This PR adds a new endpoint to the multiple-account API which enables the retrieval of secondary user subscriptions for any identity ID.

## Usage:
```shell
curl --location 'https://multiple-account-api-code.support.guardianapis.com/secondary-user/200712934' \
--header 'x-api-key: [API_KEY]'
```

```json
{
    "subscriptions": [
        {
            "subscriptionName": "A-S01115245",
            "firstName": "test",
            "lastName": "user",
            "workEmail": "test.user@theguardian.com"
        }
    ]
}
```

## Details
This endpoint shares almost all its code with the `secondary-user/me` endpoint. The only difference is that it takes the identity from the path rather than an Okta token.

### DIFF
--- a/handlers/multiple-account-api/openapi.yaml
+++ b/handlers/multiple-account-api/openapi.yaml
@@ -276,25 +276,27 @@ paths:
   /secondary-user/me:
     get:
       operationId: getSecondaryUserMe
-      summary: Get primary users for the signed-in secondary user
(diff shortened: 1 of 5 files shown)


tjmw (approved):
👍 Looks great.

Does the open api doc need updating to add the new endpoint?

PR #3782, opened 2026-08-25
Change error type for product-switch-api

enesihsa:
## Context
Since the Supporter Plus price rise started we have seen a big increase in "product-switch-api-PROD 5XX errors" alarms. 

These are being caused by an error which be found in Cloudwatch: `Handler returned 500 response due to unexpected error Error: Did not find a refund for every charge in the old subscription in the preview response: expected: ["8ad09ea0858682bb0185880ac57f4c4c","8ad08cbd8586721c01858804e3715378"] ... `

### Cause

The error is thrown when the invoice generated from a preview of the upgrade does not contain proration charges for all the existing charges in the subscription. 

This happens when a subscription has a pending price migration order - which removes the current rate plan and replaces it at the new price with a future effective date of the next billing date. An "upgrade" order by the product-switch-api does the same thing - removes the current rate plan - but is effective immediately. But from Zuora's perspective there is nothing to remove, the "current" version of the rate plan on the subscription doesn't start until the next billing date and the "active" version of the rate plan (the one that we are actually within the start and end dates of) has already been removed, so cannot be removed again. Zuora is therefore behaving correctly and product-switch-api is also correctly rejecting the result of the preview as not compatible with completing the upgrade.

If we were to skip the preview and continue with the upgrade, we would end up charging the customer twice without returning any funds for the old rate plan (technically from Zuora's perspective they'd hold both rate plans until the next billing date).

Finding a way to resolve this is non-trivial. The only way to complete the product switch as planned would be to delete the pending price rise order - but there is no way to do this as a "preview" that can be rolled back, so if the customer backed out of the upgrade we would need to reconstruct it again and this seems fraught with peril (not to mention it would be very slow). Better solutions would involve changes to how price migration orders are used, but this would be too late for the cohort currently active (and causing the alarms).

## What does this change?
We have instead chosen to suppress the alarm, while keeping the current behaviour in the UI.

This PR changes the error from a fatal 500 error to a ValidationError which defaults to 400, keeping the current behaviour of showing the CTA to the user but not firing the alarm. 

Behaviour and change in the error code is tested in CODE.

This does not change any customer-facing behaviour, it effectively just removes the alarm (which we have been pretty much ignoring for the past few weeks). There have been no customer complaints or bug reports.

## Images
image

### DIFF
--- a/handlers/product-switch-api/src/changePlan/action/preview.ts
+++ b/handlers/product-switch-api/src/changePlan/action/preview.ts
@@ -57,7 +57,7 @@ export const getRefundAmountMinorUnits = (
 		sourceSubscriptionReversalItems.length !== sourceChargeIds.length &&
 		(sourceSubscriptionReversalItems.length !== 0 || refundExpected)
 	) {
-		throw Error(
+		throw new ValidationError(
 			`getRefundAmountMinorUnits Did not find a refund for every charge in the old subscription in the preview response: expected: ${JSON.stringify(sourceChargeIds)} : received: ${JSON.stringify(itemsById)} : refundExpected: ${refundExpected}`,
 		);
 	}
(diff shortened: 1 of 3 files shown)


graham228221 (approved):
Thanks, quick work!

PR #3783, opened 2026-08-26
Reduce maximum number of invites per subscription to 3

rupertbates:
## What does this change?
This PR fixes two issues with the validation of the create invitation step in multiple-account API:
- Change the number of invitations allowed per subscription from 5 to 3 (this is the final agreed amount)
- Take account of already accepted invitations when calculating the number of invitations left. Previously we weren't doing this so a user could potentially continue creating more invitations after users had accepted them.

---
  - https://app.asana.com/0/0/1217869353043631

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ node_modules/
 handlers/*/coverage
 # PII query result downloads from supporter-product-data-lambdas download-query-results
 select-active-rate-plans-*.csv
+.pnpm-store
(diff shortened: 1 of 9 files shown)


.gitignore
```diff
@@ -14,3 +14,4 @@ node_modules/
 handlers/*/coverage
 # PII query result downloads from supporter-product-data-lambdas download-query-results
 select-active-rate-plans-*.csv
+.pnpm-store
```
tjmw, line 17: Should this be `.pnpm-store/` since it's a directory?

tjmw (approved):
👍

.gitignore
rupertbates: Seems to work as is?

tjmw: Oh weird - I just had to fix this on support-frontend as it didn't have the desired effect without the /: guardian/support-frontend#8189

PR #3784, opened 2026-08-27
Don't send an account verification email to secondary users

rupertbates:
## What does this change?
The invitation email we send to prospective secondary users will have a link to an onboarding process where the user will complete their identity account, so we don't need to send them a separate email to ask them to do this.

Previously when we created a new guest user account in the create invitation endpoint we were using the `accountVerificationEmail=true` query param to trigger an email send, this PR sets the param to false for this endpoint.


graham228221 (approved):
Thanks!

PR #3789, opened 2026-09-01
Send access removed email

maloky:
## What does this change?
Email setup and send an access removed email to secondary user  if a primary users removes it from their account.

Added test coverage  and run tests locally

[Access Removed - Triggered emails from Multiple account API](https://app.asana.com/1/1210045093164357/project/1213112839739792/task/1217160435359563?focus=true)


handlers/multiple-account-api/src/emails/accessRemovedEmail.ts
```diff
@@ -0,0 +1,28 @@
+import {
+	buildEmailMessage,
+	DataExtensionNames,
+	sendEmail,
+} from '@modules/email/email';
+import type { Stage } from '@modules/stage';
+
+export async function sendAccessRemovedEmail(
+	stage: Stage,
+	primaryUserFirstName: string,
```
tjmw: In #3792 I made a change to pass these fields as an object (other send email functions, but same concept) because I was getting slightly nervous about how easy it would be to unintentionally break things by getting the order wrong, especially since they're all typed as strings. Passing an object means we have to specify names at the call site. It's not perfect but I think it's a bit of an improvement, what do you reckon?

handlers/multiple-account-api/test/deleteSecondaryUserEndpoint.test.ts
```diff
@@ -100,105 +104,100 @@ const makeDynamoClient = (): {
...
+				Delete: {
+					TableName: 'SupporterProductData-CODE',
+					Key: {
+						subscriptionName: {
+							S: `${subscriptionName}-${secondaryIdentityId}`,
+						},
+						identityId: { S: secondaryIdentityId },
+					},
+				},
+			});
```
tjmw, line 173: Should we assert that `sendAccessRemovedEmail` was called in this test?

tjmw (approved):

handlers/multiple-account-api/src/emails/accessRemovedEmail.ts
maloky: I'm with with on this topic, also prefer to have an object being passed 👍

maloky:
@copilot resolve the merge conflicts in this pull request

tjmw (approved):
LGTM!

PR #3791, opened 2026-09-02
Remove unused taxEngineId property from Zuora response schema

tjmw:
## What does this change?

Remove unused taxEngineId property from Zuora response schema.

There's a case in Zuora sandbox where this isn't set and it's causing the sales-tax-api to return errors due to a schema validation failure. Looking at the code, I don't think we need it, so just remove it from the schema.

Ran the sales-tax-api integration tests before and after this change and they went from red to green.


paulbrown1982 (approved):
👍

PR #3766, opened 2026-08-14
Send leave subscription email to secondary user

maloky:
## What does this change?

When a secondary user leaves a primary subscription, send a confirmation email to the secondary user.

## How has this change been tested?

New tests added to verify the behaviour.

I've trigger the email to send to my inbox by temporarily modifying the secondary user email to my own in the integration test.

[Leave Shared Subscription Confirmation - Triggered emails from Multiple account API](https://app.asana.com/1/1210045093164357/project/1213112839739792/task/1217160312683324?focus=true)

PR #3792, opened 2026-09-02
Trigger email to primary user when secondary leaves subscription

tjmw:
## What does this change?

Following on from #3766. When a secondary user leaves a subscription, trigger an email to the primary user.

I've tested by running the integration tests.

### DIFF
--- a/handlers/multiple-account-api/test/deleteSecondaryUserIntegration.test.ts
+++ b/handlers/multiple-account-api/test/deleteSecondaryUserIntegration.test.ts
@@ -22,7 +22,7 @@ import { deleteSecondaryUserEndpoint } from '../src/deleteSecondaryUserEndpoint'
 import { InvitationRepository } from '../src/invitationRepository';
 
 const stage = 'CODE';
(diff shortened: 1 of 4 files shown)


handlers/multiple-account-api/test/deleteSecondaryUserIntegration.test.ts
```diff
@@ -22,7 +22,7 @@ import { deleteSecondaryUserEndpoint } from '../src/deleteSecondaryUserEndpoint'
 import { InvitationRepository } from '../src/invitationRepository';
 
 const stage = 'CODE';
-const subscriptionName = 'A-S00974337';
+const subscriptionName = 'A-S01117027';
```
tjmw, line 25: I've updated this to a subscription ID which belongs to a `thegulocal.com` email address, which will be ignored by membership-workflow.

In a subsequent PR, I'm planning on updating this integration test to not rely on data already existing, but instead create all of the data it needs (tearing down at the end).

maloky (approved):

PR #3797, opened 2026-09-03
send invitation declined email to primary user

maloky:
## What does this change?
Sends invitation declined email to primary user.

Added test coverage and run tests locally

[Trigger email to primary user when an invite is declined by Secondary user](https://app.asana.com/1/1210045093164357/project/1213112839739792/task/1218106672862361?focus=true)


handlers/multiple-account-api/src/deleteInvitationEndpoint.ts
```diff
@@ -39,6 +42,20 @@ export const deleteInvitationEndpoint = async (
 			cancelledBy,
 		);
 
+		if (cancelledBy === 'secondary') {
+			try {
```
tjmw, line 46: I think this is a nice approach. Wrapping the email send in a try/catch so the whole request doesn't fail just because the email send failed. I think we should probably be consistent and do the same thing elsewhere. Perhaps we should also introduce a Cloudwatch metric so that we could alarm on a failed email send? (This could happen in a future PR).

tjmw (commented):
👏

tjmw (approved):

handlers/multiple-account-api/src/deleteInvitationEndpoint.ts
maloky: Sounds like a good plan 👍

PR #3799, opened 2026-09-04
Return 410 Gone when accepting an already-accepted invitation

rupertbates:
## What
Previously, when a secondary user tried to accept a multiple-account invitation they had **already accepted**, the endpoint returned `404 Not Found` — because the invitation record is hard-deleted upon successful acceptance. This is misleading, since the invitation did exist and was successfully processed; a 404 implies the invitation code was invalid/unknown.

This PR changes that specific case to return `410 Gone` instead, while still returning `404` for genuinely unknown/invalid invitation codes.

## How
- Added a `gone(message)` helper (410) to `modules/routing/src/apiGatewayResponses.ts`, alongside the existing `notFound`/`badRequest` helpers.
- In `acceptInvitationEndpoint.ts`, when the invitation lookup returns nothing, we now check `secondaryUserRepository.listByIdentity(signedInUserId)` for an existing secondary-user record whose `invitationCode` matches the requested code. If found, we know the invitation was already accepted by this user, and return `410 Gone`. Otherwise we fall back to the existing `404 Not Found` behaviour.
- Added `test/acceptInvitationEndpoint.test.ts` with unit tests covering: 410 when already accepted, 404 when the invitation code is completely unknown, and 404 when the user has other secondary-user records but none matching this invitation code.

## Testing
- Added a new test to ensure correct behaviour

### DIFF
--- a/handlers/multiple-account-api/src/acceptInvitationEndpoint.ts
+++ b/handlers/multiple-account-api/src/acceptInvitationEndpoint.ts
@@ -8,6 +8,7 @@ import { getIfDefined } from '@modules/nullAndUndefined';
 import {
 	badRequest,
 	buildErrorResponse,
+	conflict,
 	notFound,
 	ok,
 } from '@modules/routing/apiGatewayResponses';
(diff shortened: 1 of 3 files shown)


handlers/multiple-account-api/src/acceptInvitationEndpoint.ts
```diff
@@ -28,6 +29,19 @@ export const acceptInvitationEndpoint = async (
...
+			// secondary user record already exists for this invitation code, the
+			// user has already accepted it previously.
+			const alreadyAccepted = (
+				await secondaryUserRepository.listByIdentity(signedInUserId)
+			).some(
+				(secondaryUser) => secondaryUser.invitationCode === invitationCode,
+			);
+
+			if (alreadyAccepted) {
+				return conflict('Invitation has already been accepted');
```
maloky: probably 410 would be more suited but not strongly opinionated: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/410

maloky (approved):
great work 👍

handlers/multiple-account-api/src/acceptInvitationEndpoint.ts
rupertbates: Yes I think you're probably right, I've updated 👍

PR #3800, opened 2026-09-04
DeliveryRecordsApi migration to CDK

enesihsa:
todo

### DIFF
--- a/cdk/lib/delivery-records-api.ts
+++ b/cdk/lib/delivery-records-api.ts
@@ -109,9 +109,17 @@ export class DeliveryRecordsApi extends SrStack {
 			DeliveryRecordsApiProxyResource: restApi.node
 				.findChild('Default')
 				.node.findChild('{proxy+}'),
+			DeliveryRecordsApiProxyAnyMethod: restApi.node
+				.findChild('Default')
+				.node.findChild('{proxy+}')
+				.node.findChild('ANY'),
 			DeliveryRecordsApiUsagePlan: restApi.node.findChild('UsagePlan'),
 			DeliveryRecordsApiLambda: lambda,
 			DeliveryRecordsApiRole: lambda.node.findChild('ServiceRole'),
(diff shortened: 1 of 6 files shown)


cdk/lib/delivery-records-api.ts
```diff
@@ -0,0 +1,147 @@
...
+			DeliveryRecordsApiAnyMethod: restApi.node
+				.findChild('Default')
+				.node.findChild('ANY'),
+			DeliveryRecordsApiProxyResource: restApi.node
+				.findChild('Default')
+				.node.findChild('{proxy+}'),
+			DeliveryRecordsApiProxyAnyMethod: restApi.node
+				.findChild('Default')
+				.node.findChild('{proxy+}')
+				.node.findChild('ANY'),
```
johnduffell: DeliveryRecordsApiAnyMethod didn't exist before so we don't really need to rename it (thus saving us to think of a name)
In the old CFN linked below, you can see the "AnyMethod" links to the ProxyResource
https://github.com/guardian/support-service-lambdas/blob/066bbfc2677d45ae97997c278e3117bfc1198b88/handlers/delivery-records-api/cfn.yaml#L137-L143

This should fix the Resource name conflict reported by cloudformation due to the logical and physical ids being the wrong way around.  See the thread for more context: https://chat.google.com/room/AAAAFNJwTZs/Ce84pI4V9p0/p-zCeiDU69Y?cls=10

```suggestion
			DeliveryRecordsApiProxyResource: restApi.node
				.findChild('Default')
				.node.findChild('{proxy+}'),
			// https://github.com/guardian/support-service-lambdas/blob/066bbfc2677d45ae97997c278e3117bfc1198b88/handlers/delivery-records-api/cfn.yaml#L137
			DeliveryRecordsApiAnyMethod: restApi.node
				.findChild('Default')
				.node.findChild('{proxy+}')
				.node.findChild('ANY'),
```

PR #3801, opened 2026-09-07
Return 410 for already accepted invitations from get invitation endpoint

rupertbates:
## What
Updates `getInvitationEndpoint.ts` (GET `/invitation/{invitationCode}`) to return `410 Gone` when an invitation has already been accepted, matching the existing behaviour of `acceptInvitationEndpoint.ts`.

Since this endpoint is not authenticated by user, it cannot check the signed-in user like the accept endpoint does. Instead, it looks up whether any secondary-user record exists for the given invitation code (via a new DynamoDB GSI) to determine "already accepted" status.

## Deployment notes
Adding the GSI is a standard, non-disruptive CloudFormation change — DynamoDB automatically backfills the new index for existing items, so no manual migration is required. The stack update may take slightly longer while the GSI becomes `ACTIVE`.


j-ruda-guardian (approved):
LGTM

PR #3802, opened 2026-09-07
Multiple Accounts: Setup our own test data in `deleteSecondaryUserIntegration.test.ts`

tjmw:
## What does this change?

In `deleteSecondaryUserIntegration.test.ts`, instead of relying on data existing in the code environment (which is error prone), delete everything we need as part of the test setup.

Also extract some of the setup to methods which we can call, to keep the main body of the test clearer.

### DIFF
--- a/handlers/multiple-account-api/test/deleteSecondaryUserIntegration.test.ts
+++ b/handlers/multiple-account-api/test/deleteSecondaryUserIntegration.test.ts
@@ -233,4 +233,4 @@ test('deleteSecondaryUserEndpoint soft deletes the secondary user and removes th
 		await deleteAccount(zuoraClient, accountNumber);
 		await deleteSupporterRatePlan(stage, primaryIdentityId, subscriptionName);
 	}
-}, 120000);
+});
--- a/handlers/multiple-account-api/test/deleteSecondaryUserIntegration.test.ts
+++ b/handlers/multiple-account-api/test/deleteSecondaryUserIntegration.test.ts
@@ -5,14 +5,21 @@
  * @group integration
  */
 import { DynamoDBClient } from '@aws-sdk/client-dynamodb';
+import dayjs from 'dayjs';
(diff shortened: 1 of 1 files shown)


handlers/multiple-account-api/test/deleteSecondaryUserIntegration.test.ts
```diff
@@ -82,88 +194,109 @@ beforeEach(async () => {
...
+		// async schedule which syncs to supporter product data?
+		// const supporterProductDataRecordsBeforeDelete =
+		// 	(await getSupporterRatePlans(stage, secondaryIdentityId)) ?? [];
+		// const subscriptionRecordBeforeDelete =
+		// 	supporterProductDataRecordsBeforeDelete.find(
+		// 		(record) =>
+		// 			record.subscriptionName ===
+		// 			`${subscriptionName}-${secondaryIdentityId}`,
+		// 	);
+		// expect(subscriptionRecordBeforeDelete).toBeDefined();
```
tjmw, line 254: I'm tempted to remove this check, which was present in this test from before I started working in here. It's quick flaky because it resolves on various things happening behind the scenes. Perhaps we could add another sleep here to make it more likely to pass?

handlers/multiple-account-api/test/deleteSecondaryUserIntegration.test.ts
```diff
@@ -5,47 +5,142 @@
...
+		termEndDate: dayjs(subscription.termEndDate),
+		contractEffectiveDate: dayjs(subscription.contractEffectiveDate),
+	});
+	cleanupTasks.push(() =>
+		deleteSupporterRatePlan(stage, primaryIdentityId, subscriptionName),
+	);
 
+	// Wait for the message above to be processed and written to DynamoDB
+	// Ugh why does this need to wait for so long?
+	await sleep(60000);
```
tjmw, line 127: I wonder how we could get away from this sleep? Poll the Dynamo table until the data is there perhaps?

PR #3803, opened 2026-09-07
Promotions API

rupertbates:
## What does this change?

Adds a new promotions API with a single `GET /promotions` endpoint that returns all v2 promotions from the `promotions` module, augmented with product catalog information (each promotion's `appliesTo.productRatePlanIds` is resolved into `catalogRatePlans` - the `ProductKey`/`ProductRatePlanKey` pairs it applies to - via `addCatalogInformationToPromos`). Promotions whose rate plan ids can't be resolved against the current product catalog are logged and excluded from the response rather than failing the whole request.

## How can I call the new endpoint?

```
curl --location 'https://promotions-api-code.support.guardianapis.com/promotions' \
--header 'x-api-key: [REPLACE_THIS_WITH_A_REAL_API_KEY]'
```

Example response:
```json
{
  "promotions": [
    {
      "promoCode": "TEST123",
      "name": "Test Promotion",
      "campaignCode": "CAMPAIGN1",
      "appliesTo": {
        "productRatePlanIds": ["2c92c0f8..."],
        "countries": ["GB"],
        "catalogRatePlans": [
          { "productKey": "SupporterPlus", "productRatePlanKey": "Monthly" }
        ]
      },
      "startTimestamp": "2024-01-01T00:00:00.000Z",
      "discount": { "amount": 50, "durationMonths": 3 }
    }
  ]
}
```

- Unit tests for `listPromotionsEndpoint` (success, partial-failure/exclusion, and error cases) and an integration test for `getPromotions` (`pnpm --filter promotions-api it-test` requires AWS credentials).


handlers/promotions-api/src/listPromotionsEndpoint.ts
```diff
@@ -0,0 +1,31 @@
...
+import { buildErrorResponse, ok } from '@modules/routing/apiGatewayResponses';
+import type { Stage } from '@modules/stage';
+
+export async function listPromotionsEndpoint(
+	stage: Stage,
+	catalogHelper: ProductCatalogHelper,
+) {
+	try {
+		const promotions = await getPromotions(stage);
+		const { succeeded, failed } = addCatalogInformationToPromos(
```
maloky, line 14: Not really related to this changes but one pattern i normally see is for functions like this to return a `{ data, errors}` object.

rupertbates: I don't think we want to expose the fact that there are some promos which have an inconsistent state to the outside world. It's mostly something that will happen in CODE and I think any promos which can't be parsed can be safely ignored.

maloky: ahhh, i meant the returned object from `addCatalogInformationToPromos` function in https://github.com/guardian/support-service-lambdas/blob/dc75aea21e3dd69bc36b5f52c98557ee25020b72/modules/promotions/src/v2/addCatalogInformationToPromo.ts#L74 not the actual response from this endpoint.

maloky: but a closer look made me understand that those are actually success and failed match for promos. you can discard what i said.

modules/promotions/src/v2/getPromotions.ts
```diff
@@ -0,0 +1,36 @@
...
+	if (result.Items === undefined) {
+		throw new ReferenceError(
+			`We were unable to retrieve promotions from ${tableName}`,
+		);
+	}
+
+	return result.Items.map((item) => {
+		const unmarshalledItem = unmarshall(item);
+		const parseResult = promoSchema.safeParse(unmarshalledItem);
+		if (!parseResult.success) {
```
maloky, line 27:
```suggestion
		if (parseResult.error) {
```
probably more inline with what we need to proxy to the logger

maloky (approved):
Nice work 👍

PR #3804, opened 2026-09-08
Add filtering to promotions API

rupertbates:
## What does this change?

Adds filtering support to the `promotions-api` `listPromotions` endpoint.

Consumers can now filter the returned promotions by:
- `active` — only return promotions whose active status (started, and not yet ended if it has an end date) matches the given value
- `productKey` — only return promotions that apply to a given catalog `ProductKey`
- `productRatePlanKey` — (used alongside `productKey`) only return promotions that apply to a given catalog `ProductRatePlanKey`

Unit tests were added/updated for:
- `listPromotionsEndpoint` covering the new `active`, `productKey`, and `productRatePlanKey` filters
- `isActivePromo`
- `getPromotions` (integration test)

Example request (CODE):

```bash
curl "https://promotions-api-code.support.guardianapis.com/promotions?active=true&productKey=SupporterPlus&productRatePlanKey=Monthly" \
  -H "x-api-key: <api-key>"
```

Example response:

```json
{
  "promotions": [
    {
      "promoCode": "PROMO1",
      "name": "A promotion",
      "campaignCode": "CAMPAIGN1",
      "appliesTo": {
        "productRatePlanIds": ["rate-plan-id"],
        "countries": ["GB"],
        "catalogRatePlans": [
          { "productKey": "SupporterPlus", "productRatePlanKey": "Monthly" }
        ]
      },
      "startTimestamp": "2024-01-01T00:00:00.000Z"
    }
  ]
}
```

### DIFF
--- a/handlers/promotions-api/src/index.ts
+++ b/handlers/promotions-api/src/index.ts
@@ -1,7 +1,9 @@
 import type { Handler } from 'aws-lambda';
+import { z } from 'zod';
 import { Lazy } from '@modules/lazy';
 import { getProductCatalogFromApi } from '@modules/product-catalog/api';
 import { ProductCatalogHelper } from '@modules/product-catalog/productCatalog';
+import { badRequest } from '@modules/routing/apiGatewayResponses';
 import { Router } from '@modules/routing/router';
 import { stageFromEnvironment } from '@modules/stage';
 import { listPromotionsEndpoint } from './listPromotionsEndpoint';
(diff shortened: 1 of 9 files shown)


handlers/promotions-api/src/index.ts
```diff
@@ -12,11 +17,48 @@ const lazyProductCatalogHelper = new Lazy(
 	'Get product catalog helper',
 );
 
+const listPromotionsQueryParamsSchema = z
+	.object({
+		active: z.enum(['true', 'false']).optional(),
+		productKey: z.string().optional(),
```
tjmw: Is there an argument for validating that this (and productRatePlanKey) are valid keys (and returning a 400 if not)? I think we have schemas for these.

Edit: Oh I can see that you're doing this validation below in code. Would it be preferable to do this with the zod schema?

tjmw (approved):
Nice!

PR #3806, opened 2026-09-09
Add payment failure comms exit API

andrade0:
## What does this change?

Adds a synchronous API endpoint that Customer Service can use to stop payment-failure communications for a customer. It resolves the Braze UUID from the supplied Identity ID and sends the `pf_csr_exit` event to Braze.

Asana: https://app.asana.com/1/1210045093164357/project/1213012175670365/task/1213521987720401?focus=true

Ran the handler checks and CDK checks locally. The tests cover the exact Braze payload, a missing Braze UUID, Braze HTTP errors and errors returned in an otherwise successful Braze response.

A successful call returns 200 only after Braze accepts the event. The Lambda 5XX alarm will show when the event cannot be sent.

The first deployment needs the CODE configuration and an upload-only artifact bootstrap before CloudFormation creates the stack. If IDAPI or Braze is unavailable, the endpoint returns an error rather than reporting false success.


johnduffell:
did you consider the pros/cons of making it more generic, e.g. braze-events-api or similar?  I haven't thought about it in detail but it does feel very single purpose on the face of it.

cdk/lib/payment-failure-comms-exit-api.ts
```diff
@@ -0,0 +1,31 @@
...
+				description:
+					'Sends a pf_csr_exit custom event to Braze for a customer Identity ID',
+				architecture: Architecture.ARM_64,
+				timeout: Duration.seconds(15),
+			},
+			monitoring: {
+				errorImpact:
+					'a customer may remain in payment-failure communications after a CSR attempts to stop them',
+				alarmDescription:
+					'payment-failure-comms-exit-api returned a 5XX response. Quick triage: check the Lambda logs for the Identity ID context, validate the IDAPI lookup returned a brazeUuid, then inspect the Braze /users/track response.',
```
johnduffell, line 23: nice this is good help for people investigating 👏

handlers/payment-failure-comms-exit-api/src/clients/brazeClient.ts
```diff
@@ -0,0 +1,50 @@
...
+		private readonly apiKey: string,
+		private readonly fetchFn: typeof fetch = fetch,
+	) {}
+
+	async sendCustomEvent(payload: BrazeTrackPayload): Promise<void> {
+		logger.log(
+			`Sending ${paymentFailureCommsExitEventName} custom event to Braze`,
+		);
+
+		const response = await this.fetchFn(`${this.apiUrl}${brazeTrackPath}`, {
```
johnduffell: please could you use the rest client as it provides all the standard logging etc for free
https://github.com/guardian/support-service-lambdas/blob/main/modules/zuora/src/restClient.ts

example here: https://github.com/guardian/support-service-lambdas/blob/main/modules/salesforce/src/sfClient.ts#L13

handlers/payment-failure-comms-exit-api/src/clients/index.ts
```diff
@@ -0,0 +1 @@
+export * from './brazeClient';
```
johnduffell: I have seen this pattern before and it seems to be to make the imports simpler, but we weren't sure what value it adds and it caused import errors in intellij.  Do we need it?

handlers/payment-failure-comms-exit-api/src/helpers/getBrazeClient.ts
```diff
@@ -0,0 +1,21 @@
...
+
+let brazeClientPromise:
+	| Promise<{ client: BrazeClient; appId: string }>
+	| undefined;
+
+export const getBrazeClient = async (): Promise<{
+	client: BrazeClient;
+	appId: string;
+}> => {
+	brazeClientPromise ??= (async () => {
```
johnduffell, line 12: oh that's clever, I've never seen `??=` but it's a bit like the `lazy val` in scala 👏

handlers/payment-failure-comms-exit-api/src/helpers/parseBrazeTrackResponse.ts
```diff
@@ -0,0 +1,16 @@
+import { brazeTrackResponseSchema } from '../schemas';
+import type { BrazeTrackResponse } from '../types';
+
+export const parseBrazeTrackResponse = (
+	responseText: string,
+): BrazeTrackResponse => {
+	try {
+		return brazeTrackResponseSchema.parse(
+			responseText === '' ? {} : JSON.parse(responseText),
```
johnduffell: why would the response be empty string and why would we want to pretend that it was actually an empty object rather than just failing/throwing?

handlers/payment-failure-comms-exit-api/src/schemas/brazeTrackResponseSchema.ts
```diff
@@ -0,0 +1,8 @@
+import { z } from 'zod';
+
+export const brazeTrackResponseSchema = z
+	.object({
+		message: z.string().optional(),
+		errors: z.array(z.string()).optional(),
+	})
+	.passthrough();
```
johnduffell: why pass through?  I think the fields we rely on should be parsed but maybe there's another benefit?

handlers/payment-failure-comms-exit-api/src/schemas/paymentFailureCommsExitRequestSchema.ts
```diff
@@ -0,0 +1,7 @@
+import { z } from 'zod';
+
+export const paymentFailureCommsExitRequestSchema = z
+	.object({
+		identityId: z.string().trim().min(1),
```
johnduffell: are we expecting salesforce to pad the identity id with whitespace?  I would prefer to be strict.
Good idea to validate with min(1) and we could even go further and check it only contains `[0-9]+`

handlers/payment-failure-comms-exit-api/src/services/getBrazeUuidFromIdapi.ts
```diff
@@ -0,0 +1,18 @@
+import { getUserByIdentityId } from '@modules/identity/idapi';
+import { logger } from '@modules/logger/logger';
+import { getIdentityClient } from '../helpers';
+
+export const getBrazeUuidFromIdapi = async (
+	identityId: string,
+): Promise<string | undefined> => {
+	const identityClient = await getIdentityClient();
+	const user = await getUserByIdentityId(identityClient, identityId);
+	const brazeUuid = user?.privateFields?.brazeUuid?.trim();
```
johnduffell: same as above, are we expecting whitespace to come back from identity?  it seems unlikely, I think we should not trim unneccesarily.

handlers/payment-failure-comms-exit-api/src/types/brazeTrackResponse.ts
```diff
@@ -0,0 +1,4 @@
+export type BrazeTrackResponse = {
+	message?: string;
+	errors?: string[];
+} & Record<string, unknown>;
```
johnduffell: why unknown?
normally we derive the types from the schema to ensure they match (although I know that has issues with the types looking quite complex - maybe this resolves that while keeping the link between the schema and type strict?

handlers/payment-failure-comms-exit-api/src/constants.ts
```diff
@@ -0,0 +1,7 @@
+export const paymentFailureCommsExitPath = '/exit';
+export const paymentFailureCommsExitEventName = 'pf_csr_exit';
+export const brazeTrackPath = '/users/track';
+export const brazeRequestTimeoutMillis = 10_000;
+
+export const identityClientAccessTokenPath = (stage: string): string =>
+	`/${stage}/support/payment-failure-comms-exit-api/identity-client-access-token`;
```
johnduffell: good idea to reuse the IdentityClient, it looks like it uses an old way where it reads SSM itself rather than just taking the token as a parameter.

It could be a separate PR but it might be worth adding a new create method to the identity client so we can read the value in via our normal getConfig and then pass it over, rather than having to replicate the config path in SSM here.
This is just a bit neater e.g. the schema will be the source of truth of what config is needed in SSM, rather than having most of it in the config schema and then one bit refereneced here directly.
What do you think?

handlers/payment-failure-comms-exit-api/test/index.test.ts
```diff
@@ -0,0 +1,8 @@
+import { handler } from '../src';
+import { handler as entrypointHandler } from '../src/handlers/paymentFailureCommsExitHandler';
+
+describe('package entrypoint', () => {
+	it('exports the Lambda handler', () => {
+		expect(handler).toBe(entrypointHandler);
+	});
+});
```
johnduffell: it seems like about half of the PR is just tests, and e.g. this one I don't think it's even testing anything - it looks like it's just testing that something is equal to itself.  It's just going to end up being noise/ignored and using resources to run every time, as well as maintaining them unnecessarily.

I think a reasonable level of testing for such a one dimensional lambda would be testing end to end, so mock the identity and braze clients, pass in some suitable config, pass in the various expected and unexpected CSR inputs, mock the braze and identity client to either both pass or one to fail, and then check that the result is OK/internal server/not found as appropriate and that the right call was made to braze.

Then the tests are a useful documentation of the key behaviour, should be readable (it looks like you've used dependency injection in the key places but may need tweaking) and will provide value as things change over time.

handlers/payment-failure-comms-exit-api/src/handlers/paymentFailureCommsExitController.ts
```diff
@@ -0,0 +1,26 @@
...
+export async function paymentFailureCommsExitController(
+	{ identityId }: PaymentFailureCommsExitRequest,
+	deps: RuntimeDeps = defaultDeps,
+): Promise<APIGatewayProxyResult> {
+	logger.mutableAddContext(identityId);
+
+	try {
+		const brazeUuid = await deps.getBrazeUuidFromIdapi(identityId);
+		if (!brazeUuid) {
+			return notFound();
```
johnduffell: can we add a helpful body e.g. "identity id was not found" or "identity id had no braze uuid found"?
no braze uuid I think should never happen should it?  so this should probably be an internalServerError so that we get an alarm.  We can always change it to 4xx later if we realise it's a common case that we don't care about.

handlers/payment-failure-comms-exit-api/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+`POST /exit` accepts a Guardian Identity ID, gets the associated `brazeUuid` from IDAPI, and sends Braze one `pf_csr_exit` custom event. The event uses `_update_existing_only`, so it cannot create a Braze profile. A `200` response means Braze accepted the event; a `404` means that IDAPI did not return a Braze UUID for the supplied Identity ID.
+
+## URLs
+
+#### CODE - https://payment-failure-comms-exit-api-code.support.guardianapis.com/exit
+
+#### PROD - https://payment-failure-comms-exit-api.support.guardianapis.com/exit
+
+Both environments require the API key generated by `SrApiLambda` in the `x-api-key` header.
```
johnduffell: nitpick but it might be more useful to mention where the API key is visible in the aws console rather than what generated it?

handlers/payment-failure-comms-exit-api/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+It also reads the Identity client token from:
+
+`/{stage}/support/payment-failure-comms-exit-api/identity-client-access-token`
+
+These parameters must be configured in CODE before the first functional test, and in PROD before deployment. They are intentionally not committed here.
+
+## First deployment
+
+For the initial deployment, upload the Riff-Raff artifacts before deploying the CloudFormation stack. The stack references the Lambda artifact, which does not exist until that first upload has completed.
```
johnduffell: I don't think we need this in the readme - by all means add it to the PR description as a reminder, but it's generic advice and it's in the main HOWTO new lambda Docs of this repo already.

cdk/lib/payment-failure-comms-exit-api.ts
andrade0: Thanks, I wanted the alarm to point straight to the likely failure points.

handlers/payment-failure-comms-exit-api/src/clients/brazeClient.ts
andrade0: Good point. It now uses RestClient. I kept the standard wrapper but suppress the request body in its log hook because it contains the Braze UUID.

handlers/payment-failure-comms-exit-api/src/clients/index.ts
andrade0: No. It was unused, so I removed it and import the client directly.

handlers/payment-failure-comms-exit-api/src/helpers/getBrazeClient.ts
andrade0: Yes, that was the intention: cache the promise once per Lambda instance.

handlers/payment-failure-comms-exit-api/src/helpers/parseBrazeTrackResponse.ts
andrade0: Agreed. An empty or malformed successful response now fails schema validation instead of being treated as an empty object.

handlers/payment-failure-comms-exit-api/src/schemas/brazeTrackResponseSchema.ts
andrade0: No benefit here. I removed it and validate the success fields the client depends on.

handlers/payment-failure-comms-exit-api/src/schemas/paymentFailureCommsExitRequestSchema.ts
andrade0: Agreed. The request is no longer trimmed and rejects any whitespace. I kept it non-numeric because we do not have a documented contract that guarantees Identity IDs are digits only.

handlers/payment-failure-comms-exit-api/src/services/getBrazeUuidFromIdapi.ts
andrade0: Agreed, the UUID is now returned unchanged.

handlers/payment-failure-comms-exit-api/src/types/brazeTrackResponse.ts
andrade0: Agreed. I removed the duplicate response type; the client uses the schema result directly.

handlers/payment-failure-comms-exit-api/src/constants.ts
andrade0: Agreed. I added IdentityClient.createWithAccessToken and this handler now loads identity.accessToken through its normal app config schema.

handlers/payment-failure-comms-exit-api/test/index.test.ts
andrade0: Agreed. I removed the atomic tests and added a handler factory with flow-level tests covering valid input, invalid input, no Identity user, missing Braze UUID and Braze failures.

handlers/payment-failure-comms-exit-api/src/handlers/paymentFailureCommsExitController.ts
andrade0: Agreed. A missing Identity user now returns 404 with a specific message. A user without a Braze UUID now throws and becomes a 500, so it is alarmed.

handlers/payment-failure-comms-exit-api/README.md
andrade0: Updated to point to API Gateway > API keys and the generated key name.

handlers/payment-failure-comms-exit-api/README.md
andrade0: Removed from the README; it is generic repository guidance rather than handler documentation.

andrade0:
I did consider it. I kept this endpoint purpose-specific because it has a constrained CSR contract: it only emits pf_csr_exit after resolving an Identity ID to an existing Braze user. A generic event API would need an explicit allow-list, access policy and consumer contract before it was safe to expose arbitrary Braze events. If a second concrete use case appears, extracting the shared Braze client while keeping purpose-specific endpoints would be straightforward.

graham228221:
>> did you consider the pros/cons of making it more generic, e.g. braze-events-api or similar? I haven't thought about it in detail but it does feel very single purpose on the face of it.

> I did consider it. I kept this endpoint purpose-specific because it has a constrained CSR contract: it only emits pf_csr_exit after resolving an Identity ID to an existing Braze user. A generic event API would need an explicit allow-list, access policy and consumer contract before it was safe to expose arbitrary Braze events. If a second concrete use case appears, extracting the shared Braze client while keeping purpose-specific endpoints would be straightforward.

I was also thinking about this, I'm not sure what the answer is but we do have another lambda in this repo that is sending Braze custom events - [braze-acquisition-events-sync](https://github.com/guardian/support-service-lambdas/tree/12dbc81e2441251eb860a924aa4589fbbe232c86/handlers/braze-acquisition-events-sync). We also use them for apps metering as well as payment failure comms.

It is attractive to have one generic "thing" to emit custom events but I'm not 100% sure it's worth working out what that would look at this point.

andrade0:
> > > did you consider the pros/cons of making it more generic, e.g. braze-events-api or similar? I haven't thought about it in detail but it does feel very single purpose on the face of it.
> 
> > I did consider it. I kept this endpoint purpose-specific because it has a constrained CSR contract: it only emits pf_csr_exit after resolving an Identity ID to an existing Braze user. A generic event API would need an explicit allow-list, access policy and consumer contract before it was safe to expose arbitrary Braze events. If a second concrete use case appears, extracting the shared Braze client while keeping purpose-specific endpoints would be straightforward.
> 
> I was also thinking about this, I'm not sure what the answer is but we do have another lambda in this repo that is sending Braze custom events - [braze-acquisition-events-sync](https://github.com/guardian/support-service-lambdas/tree/12dbc81e2441251eb860a924aa4589fbbe232c86/handlers/braze-acquisition-events-sync). We also use them for apps metering as well as payment failure comms.
> 
> It is attractive to have one generic "thing" to emit custom events but I'm not 100% sure it's worth working out what that would look at this point.

That makes sense. I agree there is probably a shared Braze custom-events capability to explore, especially with acquisition events, app metering and payment failure comms all using them.

I think it is safer to keep this CSR endpoint purpose-specific for now, then look at extracting the shared client/configuration once we have a clearer picture of the contracts and access controls needed for each use case.

handlers/payment-failure-comms-exit-api/src/clients/brazeClient.ts
```diff
@@ -1,49 +1,61 @@
...
+}
+
+export class BrazeClient extends RestClient {
+	constructor(apiUrl: string, apiKey: string) {
+		super(new BrazeTokenProvider(apiUrl, apiKey));
+	}
 
-		const parsedResponse = parseBrazeTrackResponse(await response.text());
+	// The standard RestClient logs request bodies. Omit it here because it contains
+	// the customer's Braze UUID.
```
johnduffell: that doesn't sound right, we want to log all the information so that we can diagnose any issues.  The logs don't get revealed publically, they are only accessible within our AWS console

handlers/payment-failure-comms-exit-api/src/clients/brazeClient.ts
```diff
@@ -1,49 +1,61 @@
...
-			throw new Error(
-				`Braze ${brazeTrackPath} failed with status ${response.status}: ${JSON.stringify(parsedResponse.errors ?? parsedResponse.message ?? {})}`,
-			);
-		}
+		const response = await this.post(
+			brazeTrackPath,
+			JSON.stringify(payload),
+			brazeTrackResponseSchema,
+			undefined,
+			brazeRequestTimeoutMillis,
```
johnduffell: I notice you have added a timeout here, and you introduced the timeout functionality recently for the discount-api orders changes https://github.com/guardian/support-service-lambdas/blame/26c00c6dd727b59a8fe164f1d14a64e530d0be5d/modules/zuora/src/restClient.ts#L167

What's the benefit of this timeout functionality rather than letting the lambda timeout and fail with a 500 error?  (I can't see anywhere obvious where it's being trapped and dealt with)

By letting the lambda timeout only we don't have to keep them in step, and we use the full lambda timeout rather than potentially giving up early.

handlers/payment-failure-comms-exit-api/src/handlers/paymentFailureCommsExitController.ts
```diff
@@ -9,18 +8,12 @@ export async function paymentFailureCommsExitController(
 	{ identityId }: PaymentFailureCommsExitRequest,
 	deps: RuntimeDeps = defaultDeps,
 ): Promise<APIGatewayProxyResult> {
-	logger.mutableAddContext(identityId);
```
johnduffell: is the context getting added elsewhere? I think it's helpful to have the identity id prefixed in the log lines

johnduffell: if you've tested it in CODE then it would useful to have a link to the cloudwatch logs for an example execution, just to get an idea whether anything obvious would help with investigations later

handlers/payment-failure-comms-exit-api/src/handlers/paymentFailureCommsExitController.ts
```diff
@@ -9,18 +8,12 @@ export async function paymentFailureCommsExitController(
...
-
-	try {
-		const brazeUuid = await deps.getBrazeUuidFromIdapi(identityId);
-		if (!brazeUuid) {
-			return notFound();
-		}
+	const brazeUuid = await deps.getBrazeUuidFromIdapi(identityId);
+	if (!brazeUuid) {
+		return notFound('Identity ID was not found');
```
johnduffell: I think this should say that the braze uuid wasn't found for the identity id.  Unless it covers both cases?

handlers/payment-failure-comms-exit-api/src/handlers/paymentFailureCommsExitHandler.ts
```diff
@@ -3,15 +3,23 @@ import { Router } from '@modules/routing/router';
...
+			path: paymentFailureCommsExitPath,
+			handler: withBodyParser(
+				paymentFailureCommsExitRequestSchema,
+				async (_event, _path, body) =>
+					paymentFailureCommsExitController(body, deps),
+			),
+		},
+	]);
+
+export const handler = buildPaymentFailureCommsExitHandler();
```
johnduffell: nitpick but I'd probably pass in `defaultDeps` explicitly rather than defaulting it, for clarity

handlers/payment-failure-comms-exit-api/src/helpers/index.ts
```diff
@@ -4,4 +4,3 @@ export * from './getBrazeClient';
 export * from './getCurrentIsoTime';
 export * from './getEnvironmentVariable';
 export * from './getIdentityClient';
-export * from './parseBrazeTrackResponse';
```
johnduffell: I notice you agreed that the other index.ts wasn't needed, does the same apply to the others as there are a couple more still there?

handlers/payment-failure-comms-exit-api/src/schemas/paymentFailureCommsExitRequestSchema.ts
```diff
@@ -2,6 +2,6 @@ import { z } from 'zod';
 
 export const paymentFailureCommsExitRequestSchema = z
 	.object({
-		identityId: z.string().trim().min(1),
+		identityId: z.string().min(1).regex(/^\S+$/),
```
johnduffell: I think \S is non whitespace but identity ids are exclusively 0-9.  I think \d can do that, or [0-9] to be more explicit

handlers/payment-failure-comms-exit-api/src/services/getBrazeUuidFromIdapi.ts
```diff
@@ -2,16 +2,26 @@ import { getUserByIdentityId } from '@modules/identity/idapi';
...
+	if (!user) {
+		logger.log('No Identity user found for the supplied Identity ID');
+		return undefined;
+	}
 
+	const brazeUuid = user.privateFields?.brazeUuid;
 	if (!brazeUuid) {
-		logger.log('No Braze UUID found for the supplied Identity ID');
-		return undefined;
+		throw new BrazeUuidMissingError();
```
johnduffell, line 24: a bit inconsistent to use undefined and to throw in the same function, I like the idea of using an exception to make the error clearer and we can't recover from it anyway so perhaps we should throw for both cases?

handlers/payment-failure-comms-exit-api/src/types/index.ts
```diff
@@ -1,5 +1,4 @@
 export * from './appConfig';
 export * from './braze';
-export * from './brazeTrackResponse';
```
johnduffell: here is the other potentially redundant index.ts as mentioned above

handlers/payment-failure-comms-exit-api/src/services/getBrazeUuidFromIdapi.ts
johnduffell: OK I can see now the distinction- the undefined means 404 and the throw means 500.
This assumes that getUserByIdentityId itself returns undefined for a user not found, which is indeed true 
https://github.com/guardian/support-service-lambdas/blob/26c00c6dd727b59a8fe164f1d14a64e530d0be5d/modules/identity/src/idapi.ts#L81

modules/identity/src/identityClient.ts
```diff
@@ -26,11 +26,18 @@ class IdentityTokenProvider implements BearerTokenProvider {
...
+		return new IdentityClient(new IdentityTokenProvider(accessToken, stage));
+	}
+
 	static async create(
 		stage: Stage,
 		ssmParamName: string,
 	): Promise<IdentityClient> {
 		const token = await getSSMParam(ssmParamName);
-		return new IdentityClient(new IdentityTokenProvider(token, stage));
+		return IdentityClient.createWithAccessToken(stage, token);
```
johnduffell: the parameter order of createWithAccessToken isn't consistent with the constructor, ideally stick with `token, stage` (or change it in the constructor)

handlers/payment-failure-comms-exit-api/src/clients/brazeClient.ts
graham228221: The brazeUUID is logged in other places like payment-failure-comms and membership-workflow, it's fine to log it here

handlers/payment-failure-comms-exit-api/src/clients/brazeClient.ts
andrade0: Updated again following the later logging feedback: this now uses the standard RestClient logging unchanged.

handlers/payment-failure-comms-exit-api/src/schemas/paymentFailureCommsExitRequestSchema.ts
andrade0: Followed up on the later comment: the endpoint now accepts digits-only Identity IDs.

handlers/payment-failure-comms-exit-api/src/clients/brazeClient.ts
andrade0: Restored the standard RestClient request logging, so the Braze payload is available in the Lambda logs.

handlers/payment-failure-comms-exit-api/src/clients/brazeClient.ts
andrade0: Removed the client timeout. The Lambda timeout is now the single timeout governing this request.

handlers/payment-failure-comms-exit-api/src/handlers/paymentFailureCommsExitController.ts
andrade0: Added a scoped Identity ID log context. The CODE log group is [here](https://eu-west-1.console.aws.amazon.com/cloudwatch/home?region=eu-west-1#logsV2:log-groups/log-group/%252Faws%252Flambda%252Fpayment-failure-comms-exit-api-CODE).

handlers/payment-failure-comms-exit-api/src/handlers/paymentFailureCommsExitController.ts
andrade0: The 404 only covers a missing Identity user. A user without a Braze UUID throws and returns a 500, so the 404 message now says that specifically.

handlers/payment-failure-comms-exit-api/src/handlers/paymentFailureCommsExitHandler.ts
andrade0: Changed to pass defaultDeps explicitly when constructing the exported handler.

handlers/payment-failure-comms-exit-api/src/helpers/index.ts
andrade0: Removed all remaining local barrel files and changed imports to point to the implementation files directly.

handlers/payment-failure-comms-exit-api/src/schemas/paymentFailureCommsExitRequestSchema.ts
andrade0: Updated this to a digits-only validation, with coverage for a non-numeric value.

handlers/payment-failure-comms-exit-api/src/services/getBrazeUuidFromIdapi.ts
andrade0: Exactly. The distinction is intentional: no Identity user is a 404; a user without a Braze UUID is unexpected and becomes a 500.

handlers/payment-failure-comms-exit-api/src/types/index.ts
andrade0: Removed this alongside the other local barrel files.

modules/identity/src/identityClient.ts
andrade0: Aligned createWithAccessToken with the constructor: token first, then stage.

graham228221 (approved):
Nice work 👏  tested on CODE and working great:

Image

Thanks for this and thanks to @johnduffell for checking it

PR #3813, opened 2026-09-09
promotions-api: add promoCodes filter to listPromotions

### LINKED ISSUE #3812: promotions-api: add promoCodes filter to listPromotions for exact-match lookups
## Context

Follow-up from guardian/support-frontend#8208 (deciding the scope of promotions fetched server-side once support-frontend replaces its legacy `ProductPrices`/`PriceSummaryService` mechanism with `promotions-api`).

## Problem

The current legacy mechanism in support-frontend (`PriceSummaryService.getPrices` / `DefaultPromotionService`) never needs "all active promotions for a product" — it resolves a small, deterministic candidate set of promo codes:

```scala
val defaultPromos = getDefaultPromoCodes(product)   // curated allow-list from S3 (default-promos.json), per product — typically a handful of codes
val promotions = promotionService.findPromotions(promoCodes ++ defaultPromos)  // + whatever's in ?promoCode=, typically 0-1 codes
```

The current `listPromotions` filters (`active`, `productKey`, `productRatePlanKey` — added in #3804) support "give me everything active for product X", but don't support "give me exactly these codes". Using `active`+`productKey` as a substitute for the above would mean:
- Fetching/scanning a potentially much larger set than needed (most active promotions for a product are irrelevant to any given page — e.g. referral, partner, or soft-launch codes not on the default list).
- Support-frontend having to re-implement "is this code one of my curated defaults / the one from the query string" filtering client- or server-side, duplicating logic and creating a safety-critical dependency on that filtering being correct (a bug could surface a promo never meant to be shown/auto-applied to the public).

## Proposal

Add an optional `promoCodes` filter (array) to `listPromotions`, e.g.:

```
GET /promotions?promoCodes=PROMO1,PROMO2&active=true&productKey=SupporterPlus&productRatePlanKey=Monthly
```

- Combinable with the existing `active`/`productKey`/`productRatePlanKey` filters (AND semantics) so callers can ask "these specific codes, but only if active and applicable to this rate plan" in one call.
- Given `getPromotion.ts` already does a `GetItem` by `promoCode` (the table's partition key), this maps naturally onto a `BatchGetItem` (or a loop of key lookups) rather than a full scan — should be cheap.
- Promotions codes not found, inactive (if `active=true` is set), or not matching the product/rate-plan filters are simply omitted from the response — same behaviour as the current `active`/`productKey` filters, just narrowed to only the requested codes.

## Why this is better than "load all active and filter in support-frontend"

- **Matches actual usage** — support-frontend only ever needs a small, known set of codes (curated defaults + optional query-string code), not a full active list.
- **Efficient** — direct key lookups instead of scanning/filtering "all active for product X".
- **Safer by construction** — callers can only ever get back what they explicitly asked for; there's no risk of a filtering bug on the consumer side accidentally surfacing a non-default/private promo code (e.g. referral/partner/QA codes) to the public.
- **Cleaner separation of concerns** — "which codes are eligible for default/automatic display" stays a support-frontend/business concern (wherever the curated list lives); `promotions-api` just validates/enriches the specific codes it's asked about.

## Related
- guardian/support-frontend#8207 (parent tracking issue: remove legacy ProductPrices mechanism)
- guardian/support-frontend#8208 (scope of promotions fetched/injected server-side)
- guardian/support-service-lambdas#3804 (added active/productKey/productRatePlanKey filters to listPromotions)

rupertbates:
Adds an optional promoCodes filter (comma-separated list in the query string) to GET /promotions, so callers that only need a small, known set of promo codes (e.g. support-frontend's curated default list plus an optional ?promoCode= from the query string) can fetch exactly those codes instead of scanning/filtering the full active list.

- modules/promotions: add getPromotionsByCodes, using DynamoDB's BatchGetItem (chunked to the 100-key limit, retrying any UnprocessedKeys) instead of a full table Scan. Promo codes that don't exist are simply omitted, same as an inactive/non-matching promo under the existing active/productKey filters.
- handlers/promotions-api: parse the promoCodes query parameter, wire it into listPromotionsEndpoint (combinable with the existing active, productKey and productRatePlanKey filters), and document it in openapi.yaml.
- Tests: unit tests for the new filter and its combination with the existing filters, plus an integration test for getPromotionsByCodes.

Fixes https://github.com/guardian/support-service-lambdas/issues/3812

## Example

Fetching only two specific promo codes (one existing, one not), filtered to only active ones:

```bash
curl -s 'https://promotions-api-code.support.guardianapis.com/promotions?promoCodes=SPRING26,SPRING25&active=true' \
  -H 'x-api-key: <api-key>'
```

Response (`SPRING26` exists and is active, `SPRING25` is simply omitted):

```json
{
  "promotions": [
    {
      "promoCode": "SPRING26",
      "name": "Spring 25% off",
      "campaignCode": "SPRING_CAMPAIGN",
      "appliesTo": {
        "productRatePlanIds": [
          "8a128ed885fc6ded018602296ace3eb8"
        ],
        "countries": [
          "GB"
        ],
        "catalogRatePlans": [
          {
            "productRatePlanKey": "Monthly"
          }
        ]
      },
      "startTimestamp": "2026-01-01T00:00:00.000Z",
      "endTimestamp": "2099-01-01T00:00:00.000Z",
      "discount": {
        "amount": 25,
        "durationMonths": 3
      },
      "description": "25% off for 3 months"
    }
  ]
}
```


modules/promotions/src/v2/getPromotions.ts
```diff
@@ -21,16 +45,53 @@ export const getPromotions = async (stage: Stage): Promise<Promo[]> => {
...
+
+	const tableName = getTableName(stage);
+	logger.log(
+		`Batch getting ${uniquePromoCodes.length} promo code(s) from ${tableName}`,
+		{
+			promoCodes: uniquePromoCodes,
+		},
+	);
+
+	const chunks = chunkArray(uniquePromoCodes, batchGetItemMaxKeys);
```
tjmw: This looks good, but wondering if it might be easier to put a limit on the number of promo codes you can supply in the filter. Would anyone ever need to filter > 100? It might help simplify the code a little bit.

tjmw (approved):
👍

modules/promotions/src/v2/getPromotions.ts
rupertbates: Yes it's a good point 👍

PR #3815, opened 2026-09-10
Soft-accept invitations

rupertbates:
## What does this change?

Changes how invitation acceptance is handled: instead of hard-deleting the invitation record when it's accepted, we now **soft-accept** it — the record is kept for a short retention period (2 weeks, via the existing DynamoDB `expiryDate` TTL) with an `acceptedDate` set, the same pattern already used for cancelled/rejected invitations. This gives the data platform time to import the record before it's removed by TTL.

### Key changes
- `InvitationRepository.getAcceptTransaction` (an `Update` that sets `acceptedDate` and shortens `expiryDate`) replaces the previous `getDeleteTransaction` (hard `Delete`) call in `acceptInvitationEndpoint`.
- Renamed `invitationCancellationTTL` → `dataRetentionTTL` since it's now used for both cancellation and acceptance retention.
- Renamed `listNonCancelled*` / `getNonCancelledBySubscriptionAndIdentity` methods on `InvitationRepository` and `SecondaryUserRepository` to `listActive*` / `getActiveBySubscriptionAndIdentity`, since "active" now also excludes soft-accepted invitations, not just cancelled ones.
- `getInvitationEndpoint` and `acceptInvitationEndpoint` now check `invitation.acceptedDate` directly to return `410 Gone` for already-accepted invitations, with the existing secondary-user-record lookup kept as a fallback for the case where the invitation's retention period has since expired and the record has been removed by DynamoDB's TTL.
- `deleteInvitationEndpoint` now also treats an already-accepted invitation as not found (in addition to already-cancelled).

### Related PRs
- guardian/support-service-lambdas#3799 (Return 410 Gone when accepting an already-accepted invitation) — this PR changes the mechanism used to detect an already-accepted invitation, from relying on hard-deletion to using `acceptedDate`.
- guardian/support-service-lambdas#3801 (Return 410 for already accepted invitations from get invitation endpoint) — same mechanism update applied here.
- guardian/support-service-lambdas#3736 (Multiple accounts: Soft delete secondary users) — precedent for the soft-delete/soft-accept retention pattern used here.

## External documentation

See [this document](https://docs.google.com/document/d/1eqMn9zkg-iRK3Gi9ecmgGh8WWUYpLX6B3cOasTRYpns/edit?tab=t.0#heading=h.n725ldbma47t) for background on the multiple accounts feature.

### DIFF
--- a/handlers/multiple-account-api/src/acceptInvitationEndpoint.ts
+++ b/handlers/multiple-account-api/src/acceptInvitationEndpoint.ts
@@ -28,10 +28,18 @@ export const acceptInvitationEndpoint = async (
 	try {
 		const invitation = await invitationRepository.get(invitationCode);
 
+		// Invitations are soft-accepted, so they may still exist with an
+		// acceptedDate set.
+		if (invitation?.acceptedDate) {
+			return gone('Invitation has already been accepted');
+		}
+
 		if (!invitation) {
-			// The invitation is hard-deleted once accepted, so if it's missing but a
-			// secondary user record already exists for this invitation code, the
-			// user has already accepted it previously.
+			// The invitation may be missing because a soft-accepted invitation's
+			// retention period (used to allow time for export to the data
+			// platform) has expired and the record has been removed by DynamoDB's
+			// TTL. If a secondary user record already exists for this invitation
+			// code, the user has already accepted it previously.
 			const alreadyAccepted = (
 				await secondaryUserRepository.listByIdentity(signedInUserId)
 			).some(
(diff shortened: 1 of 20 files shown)


handlers/multiple-account-api/src/getInvitationEndpoint.ts
```diff
@@ -8,11 +8,40 @@ import {
...
+	invitationCode: string,
+) => {
+	// Invitations are soft-accepted, so they may still exist with an acceptedDate set.
+	if (invitation?.acceptedDate) {
+		return true;
+	}
+
+	if (invitation) {
+		return false;
+	}
```
johnduffell, line 29: is this bit just an optimisation? I would have thought the below code would cover it fine if we accept the extra call to dyanmo.

handlers/multiple-account-api/src/acceptInvitationEndpoint.ts
```diff
@@ -28,10 +28,18 @@ export const acceptInvitationEndpoint = async (
...
+
 		if (!invitation) {
-			// The invitation is hard-deleted once accepted, so if it's missing but a
-			// secondary user record already exists for this invitation code, the
-			// user has already accepted it previously.
+			// The invitation may be missing because a soft-accepted invitation's
+			// retention period (used to allow time for export to the data
+			// platform) has expired and the record has been removed by DynamoDB's
+			// TTL. If a secondary user record already exists for this invitation
+			// code, the user has already accepted it previously.
```
johnduffell, line 42: should we reuse `invitationAlreadyAccepted` here?  (having said that we do have an extra check on the signed in identity id here, which may be important)

johnduffell (approved):
looks good,, see minor comments, and I like that it's caused the naming "nonCancelled" to change to "active" which avoids the double negatives

handlers/multiple-account-api/src/getInvitationEndpoint.ts
rupertbates: Yeah it's just to avoid the call to Dynamo, could leave it out but I thought it would be clearer

PR #3816, opened 2026-09-10
Add regex for subscription id validation in digital-subscription-expiry

shtukas:
Add regex for subscription id validation, among other things will guard against people submitting emails addresses and some passwords

### DIFF
--- a/handlers/digital-subscription-expiry/src/test/scala/com/gu/digitalSubscriptionExpiry/DigitalSubscriptionExpiryStepsTest.scala
+++ b/handlers/digital-subscription-expiry/src/test/scala/com/gu/digitalSubscriptionExpiry/DigitalSubscriptionExpiryStepsTest.scala
@@ -16,8 +16,6 @@ import org.scalatest.flatspec.AnyFlatSpec
 import org.scalatest.matchers.should.Matchers
 
 class DigitalSubscriptionExpiryStepsTest extends AnyFlatSpec with Matchers {
-
-  // Date: September 2026
   
   // While adding server side validation for the subscription ids
   // https://github.com/guardian/support-service-lambdas/pull/3816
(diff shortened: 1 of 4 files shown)


handlers/digital-subscription-expiry/src/test/scala/com/gu/digitalSubscriptionExpiry/DigitalSubscriptionExpiryStepsTest.scala
```diff
@@ -17,6 +17,14 @@ import org.scalatest.matchers.should.Matchers
 
 class DigitalSubscriptionExpiryStepsTest extends AnyFlatSpec with Matchers {
 
+  // Date: September 2026
```
andrade0: Tiny nitpick: I don't think we need the dated PR-reference comment. It will age quickly, and the fixture change is clear from the code.

handlers/digital-subscription-expiry/src/test/scala/com/gu/digitalSubscriptionExpiry/DigitalSubscriptionExpiryStepsTest.scala
```diff
@@ -251,3 +259,17 @@ class DeserialiserTest extends AnyFlatSpec with Matchers {
     Json.parse(json).validate[UrlParams] should be(JsSuccess(UrlParams(true)))
   }
 }
+
+class SubscriptionIdValidationTest extends AnyFlatSpec with Matchers {
```
andrade0, line 293: Could we also prove the endpoint-level early exit? These tests cover the predicate, but a stub that fails if `getSubscription` is reached would pin the new protection against invalid IDs reaching Zuora.

andrade0 (approved):
Looks good overall. Two minor comments below.

handlers/digital-subscription-expiry/src/test/scala/com/gu/digitalSubscriptionExpiry/DigitalSubscriptionExpiryStepsTest.scala
shtukas: https://github.com/guardian/support-service-lambdas/pull/3816/commits/c9343ea19668f2a91ee4cbdf10cfb80f8db2a828

handlers/digital-subscription-expiry/src/test/scala/com/gu/digitalSubscriptionExpiry/DigitalSubscriptionExpiryStepsTest.scala
shtukas: Done here: https://github.com/guardian/support-service-lambdas/pull/3816/changes/ec96ae55409e3a25e44755fae793e92bce3c8c15

PR #2349, opened 2024-06-26
Update new-product-api SupporterPlusLimits

rupertbates:
## What does this change?
new-product-api contains a list of minimum prices for Supporter Plus as a validation, these need to change to coincide with the Supporter Plus acquisition price rise which is happening shortly.

PR #3818, opened 2026-09-10
PLAT-870/Bug I-229 - Update Minimum prices for Supporter Plus following Price Rise- #1588

JoeMitchellGuardian:
## What does this change?
A [bug report](https://app.asana.com/1/1210045093164357/project/1213012175670365/task/1218229109488311) raised by our contact centre indicated that they were unable to add a discount to a newly created Supporter Plus monthly subscription because there were no dates available to start the discount, irregardless of how many times the term date was extended using the "Add more dates" button in the Discount UI.

_Figure 1 - replication of the Discount UI warning seen by CSRs_
Screenshot 2026-09-10 at 16 12 54

The root cause of this issue, however, is not an issue with the Discount UI, but the configuration of the subscription. It was noted that for the impacted subscription, the rate plan charges had a unique issue. Previously, the Supporter Plus V2 rate plan was introduced to ensure that any variable amount added over the base subscription price (the subscription charge) was viewed as a contribution (the contribution charge) and be taxed accordingly. This was introduced in https://github.com/guardian/salesforce/pull/789 & https://github.com/guardian/salesforce/pull/793. It was noted that the impacted subscription had a negative contribution charge, as shown in figure 2. How is this possible?

_Figure 2 - Negative Contribution charge on a Subscription_
Screenshot 2026-09-10 at 16 30 35

The reason this has occurred is because of a mismatch between the minimum amount that can be input when acquiring a new Supporter Plus Subscription in Salesforce and Zuora's product catalog. In Salesforce, the minimum amount for a UK based Supporter Plus subscriber paying monthly is £12 per month (as per the work done in https://github.com/guardian/salesforce/pull/1046), whereas in Zuora the minimum amount is £14! Therefore, the CSR was able to create a new subscription at £12 per month, which Zuora interpreted as "Create a subscription with a subscription charge of £14 per month and a contribution of £-2 per month". The discrepency is caused by Salesforce not being updated as part of the recent S+ price rise, like we had done previously in #2349 .

We have since fixed the issue for the impacted subscription mentioned in bug 229 and analysis is ongoing to find other impacted subscriptions. In this PR, we're updating relevant config and code to account for the new base prices for Supporter Plus.

## How to test this change?

1. Deploy this branch to CODE
2. Use the Salesforce acquisition functionality to validate that the new minimum amounts are being recognised.

## Notes
To be deployed at the same time as https://github.com/guardian/salesforce/pull/1588


graham228221 (approved):
👍

PR #3821, opened 2026-09-13
Add Identity deletion cleanup handler

andrade0:
When an Identity account is deleted, its old ID can remain on Salesforce Contacts and Zuora Customer Accounts. This creates data-quality problems if the person later returns with the same email address and receives a new Identity ID.

This PR adds `identity-deletion-cleanup`. Identity will add a subscription from its SNS topic to a dedicated SQS queue. The Lambda clears `IdentityID__c` from every matching Salesforce Contact and `IdentityId__c` from every matching Zuora Customer Account. The cleanup is safe to retry, and zero matches are successful.

The handler uses the existing Salesforce and Zuora REST clients, handles paging in both systems, and has a DLQ with a PROD alarm. I ran the handler, Salesforce and CDK checks locally.

For the CODE end-to-end test, Identity needs to add the queue subscription and the dedicated Salesforce Connected App credential needs to be provisioned. The first deployment also needs the normal artifact bootstrap.

Asana: https://app.asana.com/1/1210045093164357/project/1213012175670365/task/1218244474515969?focus=true


shtukas (approved):

cdk/lib/identity-deletion-cleanup.ts
```diff
@@ -0,0 +1,61 @@
...
+					'Clears deleted Identity IDs from matching Salesforce Contacts and Zuora Customer Accounts',
+				timeout: Duration.minutes(5),
+			},
+		});
+
+		lambda.addPolicies(
+			new AllowGetSecretValuePolicy(
+				this,
+				'Allow Salesforce Identity deletion cleanup secret',
+				'Salesforce/ConnectedApp/IdentityDeletionCleanup-*',
```
johnduffell, line 42: just want to check with you - is this our recommended way to store SF credentials (rather than in the normal parameter store appConfig for the lambda)?

johnduffell:
the handler is missing a README.md from its root, which the contents are up to your discretion but usually should be provided with the basic "why" it exists, how to test/set up (although not needed if it's only the standard pnpm test) and any other surprising or unusual information of note.  Bascially whatever you think would be most useful concise information for future developers to know when they are having to deal with this code in future.

cdk/lib/identity-deletion-cleanup.ts
```diff
@@ -0,0 +1,61 @@
...
+import type { SrStageNames } from './cdk/SrStack';
+import { SrStack } from './cdk/SrStack';
+
+export class IdentityDeletionCleanup extends SrStack {
+	constructor(scope: App, stage: SrStageNames) {
+		super(scope, { stage, app: 'identity-deletion-cleanup' });
+
+		const identityMmaSnsDeletionRequestTopicArn = new SrAppConfigKey(
+			this,
+			'identityMmaSnsDeletionRequestTopicArn',
```
johnduffell, line 20: would be useful to have information about this parameter, presumably it is provided by identity team and placed into our config.  Either can be an inline comment or in the readme.

handlers/identity-deletion-cleanup/src/handlers/identityDeletionHandler.ts
```diff
@@ -0,0 +1,45 @@
+import type { SQSEvent, SQSRecord } from 'aws-lambda';
+import { logger } from '@modules/logger/logger';
+import {
+	identityDeletionEventSchema,
+	identityDeletionSnsEnvelopeSchema,
+} from '../schemas/identityDeletionEventSchema';
+import { cleanDeletedIdentity } from '../services/cleanDeletedIdentity';
+import type { IdentityDeletionCleanupDependenciesFactory } from '../types/identityDeletionCleanup';
+
+export async function handleIdentityDeletionEvent(
```
johnduffell: there's a ready made handler that should cover the standard logging etc for SQS handlers and call your function for each message, which is worth looking at for this case
https://github.com/guardian/support-service-lambdas/blob/main/modules/routing/src/sqsHandler.ts

handlers/identity-deletion-cleanup/src/helpers/chunk.ts
```diff
@@ -0,0 +1,9 @@
+export function chunk<T>(items: readonly T[], size: number): T[][] {
```
johnduffell: is this the same as this?
https://github.com/guardian/support-service-lambdas/blob/066bbfc2677d45ae97997c278e3117bfc1198b88/modules/arrayFunctions.ts#L90-L99

handlers/identity-deletion-cleanup/src/schemas/identityDeletionEventSchema.ts
```diff
@@ -0,0 +1,12 @@
+import { z } from 'zod';
+
+export const identityDeletionEventSchema = z.object({
```
johnduffell, line 11: is there any reference to this spec? e.g. in identity's repo, maybe it could be linked either here or in the readme.  So it's easier to find the producer rather than searching the org.

handlers/identity-deletion-cleanup/src/schemas/identityDeletionEventSchema.ts
```diff
@@ -0,0 +1,12 @@
...
+	userId: z.string().regex(/^[0-9]+$/),
+	eventType: z.literal('DELETE'),
+	brazeId: z.string().nullable().optional(),
```
johnduffell: does it need to be both nullable and optional? do they mean different things or is this just in case?
What do we actually use it for, as I can't see any references to it in the code.  If not, can we just ignore the field completely for now?

handlers/identity-deletion-cleanup/src/schemas/identityDeletionEventSchema.ts
```diff
@@ -0,0 +1,12 @@
+import { z } from 'zod';
+
+export const identityDeletionEventSchema = z.object({
+	userId: z.string().regex(/^[0-9]+$/),
+	eventType: z.literal('DELETE'),
```
johnduffell, line 13: if we get something else e.g. "CREATE", will it end up on the DLQ?  Is that what we want?

cdk/lib/identity-deletion-cleanup.ts
andrade0: Yes. This uses the existing `SfClient.createWithClientCredentials` convention: OAuth credentials stay in Secrets Manager. The SSM app config only holds the non-sensitive Identity topic ARN. I have documented that distinction in the README.

cdk/lib/identity-deletion-cleanup.ts
andrade0: Added this to the README. The parameter contains Identity’s account-deletions topic ARN, is stored under the standard Supporter Revenue app-config path, and scopes the queue policy. Identity creates the SNS subscription separately.

handlers/identity-deletion-cleanup/src/handlers/identityDeletionHandler.ts
andrade0: Good call. This now uses `SQSHandler`, so the standard wrapper handles batch iteration, record logging and error propagation. The Salesforce and Zuora clients are created once per cold start.

handlers/identity-deletion-cleanup/src/helpers/chunk.ts
andrade0: Yes. It now uses `chunkArray`; the local helper and its test have been removed.

handlers/identity-deletion-cleanup/src/schemas/identityDeletionEventSchema.ts
andrade0: Added a link to Identity’s `EventPublisher` in the README, alongside the event contract.

andrade0:
Added a README covering the purpose, flow, SSM and secret configuration, Identity event source, SNS subscription setup, and automated and CODE end-to-end testing.

handlers/identity-deletion-cleanup/src/services/findSalesforceContactIds.ts
```diff
@@ -0,0 +1,21 @@
...
+
+const salesforceContactSchema = z.object({
+	Id: z.string(),
+});
+
+export async function findSalesforceContactIds(
+	sfClient: SfClient,
+	identityId: string,
+): Promise<string[]> {
+	const query = `SELECT Id FROM Contact WHERE IdentityID__c = '${identityId}'`;
```
johnduffell: would be good to quickly check here that identityId is only matching one or more of 0-9, to tighten up safety on the string
It might also be worth making a branded type with a specific zod parser for identityid, similar to we do for the zuora product ids
https://github.com/guardian/support-service-lambdas/blob/066bbfc2677d45ae97997c278e3117bfc1198b88/modules/zuora-catalog/src/zuoraCatalogSchema.ts#L11-L29

handlers/identity-deletion-cleanup/src/services/cleanDeletedIdentity.ts
```diff
@@ -0,0 +1,20 @@
...
+	dependencies: IdentityDeletionCleanupDependencies,
+): Promise<IdentityDeletionCleanupOutcome> {
+	const salesforceContactIds =
+		await dependencies.findSalesforceContactIds(identityId);
+	const salesforceContactsCleared =
+		await dependencies.clearSalesforceContactIds(salesforceContactIds);
+
+	const zuoraAccountIds = await dependencies.findZuoraAccountIds(identityId);
+	const zuoraAccountsCleared =
+		await dependencies.clearZuoraAccountIds(zuoraAccountIds);
```
johnduffell, line 18: how many do we expect? it might be worth querying bigquery to see how many we would expect, and refuse (DLQ) and that have too many.  This would avoid it going rogue and clearing all of them out if the query somehow gets muddled.

handlers/identity-deletion-cleanup/src/services/findZuoraAccountIds.ts
```diff
@@ -0,0 +1,49 @@
+import { z } from 'zod';
+import type { ZuoraClient } from '@modules/zuora/zuoraClient';
+
+const zuoraAccountsPageSchema = z.object({
```
johnduffell: there is already an object-query with provided schemas/type safety, not sure if it's reusable here, it looks like it is on the face of it
https://github.com/guardian/support-service-lambdas/tree/066bbfc2677d45ae97997c278e3117bfc1198b88/modules/zuora/src/objectQuery/queries

handlers/identity-deletion-cleanup/test/services/defaultDependencies.test.ts
```diff
@@ -0,0 +1,79 @@
...
+		);
+		expect(mockFindZuoraAccountIds).toHaveBeenCalledWith(
+			expect.anything(),
+			'1234567',
+		);
+		expect(mockClearZuoraAccountIds).toHaveBeenCalledWith(expect.anything(), [
+			'account-1',
+		]);
+	});
+});
```
johnduffell: I'm struggling to tell what this tests
The tests are pretty important as documentation as well as testing logic.  
What I'd recommend is having a small set of clear, understandable tests where we pass in the mocks at the deps level (or even zuora/sf client level if appropriate) and then a few it-tests or a runManual folder alongside src in order to exercise the deps themselves.
Each test has a maintenance cost so we need to make sure the value is greater than the cost.

handlers/identity-deletion-cleanup/src/schemas/identityDeletionEventSchema.ts
andrade0: Agreed. We do not use `brazeId`, so it is no longer modelled. Zod will ignore the producer’s extra field.

handlers/identity-deletion-cleanup/src/schemas/identityDeletionEventSchema.ts
andrade0: Yes. This handler has no safe interpretation of a non-DELETE event, so it rejects `CREATE` explicitly. `SQSHandler` will retry it and then send it to the DLQ for investigation. I added a handler test for that path.

handlers/identity-deletion-cleanup/src/services/findSalesforceContactIds.ts
andrade0: Added an `IdentityId` branded type parsed from a digits-only schema. It is now passed through the lookup dependencies rather than widening back to `string`.

handlers/identity-deletion-cleanup/src/services/cleanDeletedIdentity.ts
andrade0: Good point. Both lookups now request at most eleven results and reject the message if more than ten match, before any update. I could not query the BigQuery distribution because my local credentials require reauthentication, so I have not treated ten as an evidenced expectation; we can tighten it once that data is checked.

handlers/identity-deletion-cleanup/src/services/findZuoraAccountIds.ts
andrade0: Good point. This now uses the typed `objectQuery.accounts` builder and its account query schema instead of defining a local response schema.

handlers/identity-deletion-cleanup/test/services/defaultDependencies.test.ts
andrade0: Agreed. I removed this wiring test. It only tested implementation plumbing; the remaining tests exercise each lookup and update operation at the client boundary, and the handler orchestration with its dependencies.

handlers/identity-deletion-cleanup/src/services/findSalesforceContactIds.ts
```diff
@@ -1,21 +1,29 @@
 import { z } from 'zod';
 import { executeSalesforceQueryAll } from '@modules/salesforce/query';
 import type { SfClient } from '@modules/salesforce/sfClient';
+import { MAXIMUM_MATCHING_RECORDS_PER_IDENTITY } from '../constants';
```
johnduffell: not a major issue but the number isn't fundamentally the same for zuora and salesforce.
I'd expect more for zuora than salesforce - we only create a single SF account per identity user (although I'm sure we could have 2-3 if there have been migrations) whereas we have one per subscription in zuora, so there could be 10 or more
Also is there a benefit of keeping the constants together? would it be cleaner to just put each const in the file that uses it

handlers/identity-deletion-cleanup/src/services/findSalesforceContactIds.ts
```diff
@@ -1,21 +1,29 @@
...
+	const query = `SELECT Id FROM Contact WHERE IdentityID__c = '${identityId}' LIMIT ${MAXIMUM_MATCHING_RECORDS_PER_IDENTITY + 1}`;
 	const contacts = await executeSalesforceQueryAll(
 		sfClient,
 		query,
 		salesforceContactSchema,
 	);
 
+	if (contacts.length > MAXIMUM_MATCHING_RECORDS_PER_IDENTITY) {
+		throw new Error(
+			`Identity ID matched more than ${MAXIMUM_MATCHING_RECORDS_PER_IDENTITY} Salesforce Contacts`,
```
johnduffell: need some guidance what to do, either try to include it in the message, or refer to the readme doc.  E.g. `This may indicate a query issue so should be checked manually and if it's legitimately found more than the limit ...` and then put the action to take, maybe it would be done manually or we could add a facility (if and when it's needed) to pass in a higher limit on an exceptional basis.

handlers/identity-deletion-cleanup/README.md
```diff
@@ -0,0 +1,57 @@
+# identity-deletion-cleanup
+
+When an Identity account is deleted, this handler clears the matching numeric
+Identity ID from Salesforce Contacts and Zuora Accounts. It leaves the records
+themselves in place, so both systems retain their existing audit history.
```
johnduffell: need some information on why, what's the requirement/value in doing this e.g. gdpr or whatever/if it's not working what would be the issue?

handlers/identity-deletion-cleanup/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Flow
+
+1. Identity publishes a `DELETE` event to its account-deletions SNS topic.
```
johnduffell: I think there must be some stuff happens before that, e.g. a user goes to some system to delete their account? or do they have to request it manually via userhelp?  This information will be useful if something goes wrong or someone is trying to manually test this in future.
The flow illustrated here pretty much just describes what's already in the repo, but the wider context is the valuable part.

handlers/identity-deletion-cleanup/README.md
```diff
@@ -0,0 +1,57 @@
...
+3. The Lambda processes each SQS record, then clears `IdentityID__c` from every
+   matching Salesforce Contact and `IdentityId__c` from every matching Zuora
+   Account.
+
+No match in either system is a successful outcome. Multiple Salesforce Contacts
+or Zuora Accounts for the same Identity ID are all updated.
+
+Each lookup is limited to ten matches. More than ten indicates an unexpected
+data condition, so the message fails before any records are updated and can be
+investigated from the DLQ.
```
johnduffell: why do we have the limit and what do people need to do about it if it happens?

handlers/identity-deletion-cleanup/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+The CloudFormation stack reads this non-sensitive SSM parameter:
+
+| Key | Description |
+| --- | --- |
+| `identityMmaSnsDeletionRequestTopicArn` | The ARN of Identity's account-deletions SNS topic. It is stored at `/<STAGE>/support/identity-deletion-cleanup/identityMmaSnsDeletionRequestTopicArn` and is used to restrict the SQS queue policy to that topic. |
+
+The Identity team owns the SNS subscription. On a first deployment or queue
+recreation, ask them to subscribe the queue to the topic and confirm the
+subscription message delivered to the queue.
```
johnduffell, line 38: good this is helpful to capture the dependency 👍

handlers/identity-deletion-cleanup/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+The handler reads Salesforce OAuth credentials from Secrets Manager at
+`<STAGE>/Salesforce/ConnectedApp/IdentityDeletionCleanup`. The credential stays
+in Secrets Manager; it is not application configuration. It reuses the standard
```
johnduffell: I still don't understand why it's separate, if it's specific to this lambda anyway?  I can understand the zuora one since it's shared.  If it's just convention then that's fine, it's not something to address today, but it would be good to write that.

handlers/identity-deletion-cleanup/README.md
```diff
@@ -0,0 +1,57 @@
...
+Support Service Lambdas Zuora OAuth secret.
+
+## Event contract
+
+The producer is Identity's [EventPublisher](https://github.com/guardian/identity/blob/main/identity-api/src/main/scala/com/gu/identity/api/integration/sns/EventPublisher.scala).
```
johnduffell: I'd change the link to a permalink, just in case the code is moved in main
https://github.com/guardian/identity/blob/4f09e19357f1bde61a967538ab9fb4a43bf30d5a/identity-api/src/main/scala/com/gu/identity/api/integration/sns/EventPublisher.scala#L70

handlers/identity-deletion-cleanup/README.md
```diff
@@ -0,0 +1,57 @@
...
+Run the automated tests with:
+
+```bash
+pnpm test
+```
+
+For an end-to-end CODE test, create a non-Guardian-domain Identity test account,
+add its numeric Identity ID to test records in Salesforce and Zuora, then delete
+the Identity account through Manage My Account. Confirm both fields were cleared
+and that the Lambda processed the queue message successfully.
```
johnduffell: good this is really helpful 👏

handlers/identity-deletion-cleanup/src/schemas/identityDeletionEventSchema.ts
johnduffell: thanks, I don't think we need a test for that, I was just checking that it was considered

handlers/identity-deletion-cleanup/src/services/cleanDeletedIdentity.ts
johnduffell: ok I have added a separate comment as I think the limit will be different between SF and zuora, I was imagining something like 5 for sf and more like 15 for zuora, subject to what the actual data tell us

handlers/identity-deletion-cleanup/test/services/defaultDependencies.test.ts
johnduffell: Thanks have you also looked through the other tests and considered adding one or two manual/it-tests or is that still to come?

handlers/identity-deletion-cleanup/src/schemas/identityDeletionEventSchema.ts
andrade0: Removed the duplicate handler-level test. The schema test still documents the DELETE-only contract.

handlers/identity-deletion-cleanup/src/services/findSalesforceContactIds.ts
andrade0: Added the manual handling guidance to the README. The limit fails before any update; after the message reaches the DLQ, the query and matches should be checked manually. Legitimate exceptional matches can be cleared manually and the limit raised in a reviewed change if needed.

handlers/identity-deletion-cleanup/README.md
andrade0: Changed this to the suggested permalink.

handlers/identity-deletion-cleanup/src/services/cleanDeletedIdentity.ts
andrade0: I checked the current BigQuery mirrors and split the limits. Salesforce has at most one matching Contact across 2.14m numeric Identity IDs. Zuora has a longer tail: 109 IDs have more than 15 matching Accounts, but only 7 have more than 50. The limits are now five for Salesforce and fifty for Zuora, each next to its lookup.

handlers/identity-deletion-cleanup/test/services/defaultDependencies.test.ts
andrade0: I reviewed the remaining tests and added a CODE manual runner which creates the real clients and runs the handler against an explicit test Identity ID. The README now distinguishes that dependency check from the deployed end-to-end test through Manage My Account. It is deliberately opt-in because it clears matching test records.

handlers/identity-deletion-cleanup/src/services/findSalesforceContactIds.ts
andrade0: I checked the current BigQuery mirrors and split the limits. Salesforce has at most one matching Contact across 2.14m numeric Identity IDs, so five allows for migrations. Zuora has a longer tail: 109 IDs have more than 15 Accounts, but only 7 have more than 50. I have therefore used fifty for Zuora, keeping the guard without creating routine DLQs. The constants now live alongside their use.

handlers/identity-deletion-cleanup/README.md
andrade0: Added the rationale and action for the limits to the README. They protect against a broad or malformed query match, and an exceeded limit goes to the DLQ before any record is changed.

handlers/identity-deletion-cleanup/README.md
andrade0: Added the data-quality rationale to the README. Without this cleanup, the deleted Identity ID remains as a stale link on Salesforce and Zuora records. The handler removes that link while retaining the existing audit history.

handlers/identity-deletion-cleanup/README.md
andrade0: Added the wider flow and the manual end-to-end route to the README: the user deletes their account through Manage My Account, Identity publishes the event, and the handler clears the linked records.

handlers/identity-deletion-cleanup/README.md
andrade0: Documented that this is a dedicated Connected App credential for this Lambda, held in Secrets Manager and independently rotatable from application configuration. The shared Zuora credential remains unchanged.

PR #3824, opened 2026-09-14
Add a skill for buildCheck

rupertbates:
## What does this change?
[Copilot agent skills](https://learn.microsoft.com/en-us/microsoft-copilot-studio/agents-experience/skills-overview) are self-contained packages of instructions, scripts, and resources that teach AI agents how to perform specialized, repeatable tasks. 

This PR adds a skill for using the build check tool, the aim being that it will prevent every new agent from having to work out how to use the tool from scratch, saving on time and tokens.


shtukas (approved):

.github/skills/buildcheck/SKILL.md
```diff
@@ -0,0 +1,40 @@
...
+Buildcheck generates and checks the build files (`package.json`, `tsconfig.json`, jest
+config, etc.) for every handler in `handlers/*` and module in `modules/*` from a
+programmatic definition, similar to how CDK generates CloudFormation. **Never hand-edit a
+generated file** - each one starts with a `MANAGED FILE` header comment; edit the
+definition instead and regenerate. The generated file list lives in the root
+`BUILDCHECK.md`.
+
+For the basic "add a dependency / bump a version / fix a failing CI check" steps, see
+[`buildcheck/README.md`](../../../buildcheck/README.md) - don't duplicate them here, follow
+that doc directly.
```
johnduffell, line 17: "don't duplicate them here" sounds like your instruction to write this rather than the skill's instruction to the future AI

johnduffell (approved):
I'm honestly not sure what the benefits of replicating things here as opposed than improving the readme/inline comments as this is all the same repo.  If it's good for AI it's good for humans and vice versa?
I feel like skills might make more sense for information not otherwise represented in the repo e.g. how zuora stuff is defined (the sort of thing that would be spread around PR descriptions and chat messages otherwise)
However if it helps it then go for it, we might learn something in the process!

rupertbates:
> I'm honestly not sure what the benefits of replicating things here as opposed than improving the readme/inline comments as this is all the same repo. If it's good for AI it's good for humans and vice versa? I feel like skills might make more sense for information not otherwise represented in the repo e.g. how zuora stuff is defined (the sort of thing that would be spread around PR descriptions and chat messages otherwise) However if it helps it then go for it, we might learn something in the process!

I think the difference between this and the readme is that skills are loaded in each new session, so it will have that context without having to go off and work it out every time.

johnduffell:
if the readme has sufficient information (as it should) then why does AI still have to work it out every time?  Do you mean that AI is less likely to trust a README and therefore double check anyway, whereas a skill is treated as the last word?

I don't think the skill actually has sufficient information without looking at the build.ts and checking the actual types etc.

It seems like skills are only loaded when triggered, but all the generated files link to the README already, so now it would read the skill instead of/as well as the readme?

I'm not objecting to trying it out, I just don't understand the benefit here.

johnduffell:
After copilot adding a buildcheck module on something I'm doing, I then asked it to make a skill that would have saved it time and tokens and this is what it came up with.  It is very comprehensive but it does seem to have replicated a lot more than your PR, which would require a lot of maintenance.  When it was understanding buildcheck it just read the actual build.ts etc.

It seems like the benefit of skills is that it loads the name and description in as context at all times, meaning it is primed to know about it even when it doesn't discover it by itself.

I think if I was going to go for one it would be yours as it has far less duplication, although maybe also less saving.


---
name: buildcheck
description: Guide for working with this repo's buildcheck tool, which generates and CI-checks package.json/tsconfig.json/jest.config.js/BUILDCHECK.md for every handler and module from a central TypeScript definition. Use when scaffolding a new module or handler under modules/ or handlers/, adding or bumping a dependency, changing a module's/handler's scripts, or fixing a failing buildcheck check in CI.
---

# Buildcheck

`buildcheck` generates and CI-checks the boilerplate build files for every
handler/module in this monorepo (`package.json`, `tsconfig.json`, `jest.config.js`,
`BUILDCHECK.md`), from one central TypeScript definition - conceptually like CDK,
but for `package.json`/`tsconfig.json` instead of CloudFormation. It guarantees every
handler/module has consistent scripts and pinned dependency versions, and lets CI
detect drift (a manually-edited generated file, or a stale one after the definition
changed).

**Golden rule: never hand-edit a generated file.** Files under handler/module
directories that start with a "MANAGED FILE" notice comment are generated - edit the
definition in `buildcheck/data/` instead and regenerate.

## Key files

| File | Purpose |
|---|---|
| `buildcheck/data/build.ts` | The main definition: one `ModuleDefinition`/`HandlerDefinition` object per module/handler, plus the `build.modules`/`build.handlers` arrays that register them. **This is what you edit to add a module or handler, or change its dependencies/scripts.** |
| `buildcheck/data/dependencies.ts` | Central, version-pinned list of allowed dependencies (`dep`, `devDeps`, `deprecatedDeps`), referenced from `build.ts` via `...dep['some-package']`. AWS SDK v3 clients are added via the `awsClients([...])` helper (single shared version, just list the client ids e.g. `'client-eventbridge'`, prefixed with `@aws-sdk/` and versioned automatically). **Add a dependency here before it can be referenced in `build.ts`.** |
| `buildcheck/data/scripts.ts` | Shared script snippets (e.g. `openApiScripts`, `srcOnly`) reusable across definitions. |
| `buildcheck/data/snippets/*.ts` | Templates for the actual generated file contents (`buildPackageJson.ts`, `buildTsConfigJson.ts`, `buildJestConfigJs.ts`, `BUILDCHECK.md.ts`, `notices.ts`). Rarely need editing - only if changing the *shape* of generated files, not their content per-handler/module. |
| `buildcheck/README.md` | The canonical (short) how-to; this file expands on it with more mechanism detail. |
| `BUILDCHECK.md` (repo root) | Auto-generated index of every managed file in the repo - itself a generated file, don't hand-edit. |

## `ModuleDefinition` shape (from `build.ts`)

```ts
interface ModuleDefinition {
	name: string; // becomes the folder name under modules/ or handlers/
	extraScripts?: Record<string, string>; // extra/overriding pnpm scripts
	dependencies?: Record<string, string>; // spread from `dep[...]`
	devDependencies?: Record<string, string>; // spread from `devDeps[...]`
	testTimeoutSeconds?: number;
	jestClearMocks?: boolean;
	moduleDependencies: ModuleDefinition[]; // other modules this one depends on -
	  // drives both the generated package.json ("workspace:*" deps) and the
	  // tsconfig.json project references, so it must match real imports.
}

interface HandlerDefinition extends ModuleDefinition {
	stack?: 'support' | 'membership';
	functionNames?: string[];
	entryPoints?: string[];
	extraStages?: Array<'CSBX'>;
}
```

Handlers are modules plus deployment-specific fields (they also get a generated
`riff-raff.yaml`; modules don't, since only handlers deploy).

## Procedure: add a new module (e.g. `modules/subscription-events/`)

1. If it needs a dependency not already in `buildcheck/data/dependencies.ts`, add it
   there first (either a plain entry in `dep`/`devDeps`, or add the AWS SDK client id
   to the `awsClients([...])` array if it's `@aws-sdk/client-*`).
2. In `buildcheck/data/build.ts`, add a new `const moduleXyz: ModuleDefinition = {
   name: 'xyz', dependencies: {...dep['zod']}, moduleDependencies: [moduleAws,
   moduleLogger] }` near the other module definitions (order doesn't matter
   functionally, but definitions must be declared before they're referenced by
   another module's `moduleDependencies`, since it's plain TS object references, not
   string names).
3. Add `moduleXyz` to the `modules: [...]` array in the exported `build`
   (`BuildDefinition`) object at the bottom of `build.ts`.
4. Run `pnpm snapshot:update` from the repo root (or `cd buildcheck && pnpm run
   snapshot:update`). This generates `modules/xyz/package.json`,
   `modules/xyz/tsconfig.json`, `modules/xyz/jest.config.js`, and
   `modules/xyz/BUILDCHECK.md` - creating the `modules/xyz/` directory if it doesn't
   exist yet.
5. Run `pnpm install` (new workspace package/deps need linking).
6. Write the actual source code under `modules/xyz/src/` by hand - buildcheck never
   generates source files, only the build/config boilerplate above.
7. Run the module's own `pnpm --filter xyz build|lint|test` (or the repo-wide
   equivalents) to verify.

## Procedure: add a new handler

Handlers additionally use a `hygen` generator for initial scaffolding (see the root
`package.json` script `new-api-lambda`: `pnpm hygen new-lambda api-gateway &&
pnpm snapshot:update && pnpm install && ...`) - modules have no equivalent generator,
just steps 1-3 above (manually add the `ModuleDefinition`, no hygen template).

## Procedure: add a dependency to an existing module/handler

1. Open `buildcheck/data/build.ts`, find the module/handler's definition, add the
   dependency to its `dependencies`/`devDependencies` (spreading from
   `dep`/`devDeps` in `dependencies.ts` - add it there first if not already present).
2. Run `pnpm snapshot:update`.

## Procedure: bump a dependency version

1. Open `buildcheck/data/dependencies.ts`, edit the version number.
2. Run `pnpm snapshot:update`.

## Procedure: fix a failing buildcheck check in CI

1. Run `pnpm snapshot:update` to refresh the full set of build files.
2. Review the diff; if necessary, update `buildcheck/data/build.ts`/`dependencies.ts`
   and re-run `pnpm snapshot:update` until the diff looks correct.
3. Commit and push.

## Common commands

| Command | What it does |
|---|---|
| `pnpm snapshot:update` (repo root) | Regenerates all managed files from the current `build.ts`. Run this after any edit to `buildcheck/data/*.ts`. |
| `pnpm snapshot:clean` (inside `buildcheck/`) | Removes generated files (rarely needed directly). |
| `pnpm snapshot:assert` (inside `buildcheck/`) | Runs the Jest check that generated files match what's on disk - this is what CI runs to catch drift. |
| `pnpm --filter <module-name> <script>` | Run a generated script (`build`, `lint`, `test`, etc.) for one module/handler. |

## Gotchas / things to double check

- **Declare dependencies before referencing them**: `dep['@aws-sdk/client-eventbridge']`
  will be a TypeScript error in `build.ts` until the client id is added to the
  `awsClients([...])` list in `dependencies.ts`.
- **`moduleDependencies` must reflect real imports**: it's not just documentation -
  it drives the generated `package.json` workspace dependencies and `tsconfig.json`
  project references, so if module A's code imports from module B but `A`'s
  `moduleDependencies` doesn't include `moduleB`, the build/type-check will fail with
  a missing-project-reference or module-resolution error, not a buildcheck-specific
  one.
- **No cycles**: since `moduleDependencies` is a plain array of object references
  (not lazy/string-based), a true circular dependency between two modules isn't
  representable at all - TypeScript would fail to compile `build.ts` itself
  (`moduleA` can't reference `moduleB` if `moduleB`'s definition also needs to
  reference `moduleA` before either is declared). This forces dependency direction
  decisions to be made explicit at the `build.ts` level, not just by convention.
- **Never hand-edit generated files** - look for the "MANAGED FILE" notice comment at
  the top; if present, edit `buildcheck/data/*.ts` instead and re-run
  `pnpm snapshot:update`.
- **Modules don't get a `riff-raff.yaml`** - only `HandlerDefinition`s do, since only
  handlers are deployed as lambdas.

PR #3826, opened 2026-09-16
new subscriptions event bus infra

johnduffell:
As part of the work for multiple accounts, we want to send cancellation confirmation emails to secondary users, so they know that their access will be/has ended.  As part of that, we are mediating cancellation events via a new eventbridge event bus.
See the spike doc for full details: https://docs.google.com/document/d/1soYr7Ok0ew041hm4V4C-o7i2roRPo4IvFU5u0myDtpI/edit?usp=sharing

This is the first PR, this adds a new event bus for subscription events.  No producers or listeners are added yet — the bus currently has no traffic.

## This PR...

- Creates a new CDK-only stack, `subscription-events-bus` (`subscription-events-CODE`/`PROD`), owning an EventBridge bus for subscription lifecycle events.
- Adds a log destination to log all messages
- Exports the bus ARN and adds a reusable `AllowPutSubscriptionEventPolicy` IAM helper, so future producer lambdas can be granted `events:PutEvents` access.

## Testing
Tested in CODE, deployed ok: https://riffraff.gutools.co.uk/deployment/history?projectName=support-service-lambdas::subscription-events-bus&stage=CODE

https://eu-west-1.console.aws.amazon.com/events/home?region=eu-west-1#/eventbus/subscription-events-CODE

## Logging options

I have tried two approaches for logging - firstly logging all events with a normal rule, and secondly the built in event bus INFO logging.  My preference is the former, as we are interested in the business messages rather than the bus operation, but opinions welcome.

### Option 1 for logging is similar to the existing acquisitions event bus - a catch all rule that targets a log group:
Event sent and received
image

image

https://eu-west-1.console.aws.amazon.com/cloudwatch/home?region=eu-west-1#logsV2:log-groups/log-group/$252Faws$252Fevents$252Fsubscription-events-CODE/log-events/edb312ef-2f34-39a4-8273-9dca4257e658

### Option 2 for logging is the built in logging:
Event sent and received
image

image

https://eu-west-1.console.aws.amazon.com/cloudwatch/home?region=eu-west-1#logsV2:log-groups/log-group/$252Faws$252Fvendedlogs$252Fevents$252Fevent-bus$252Fsubscription-events-CODE/log-events/EventBridgeEventBusLogs

## Next steps (future PRs)

- Define the `Cancellation` event payload schema using zod (see spike doc's proposed envelope).
- membership-workflow PR to add the relevant Braze api trigger ID for the secondary user email
- Build the SQS+lambda listener that looks up secondary users and pushes to `braze-emails`.
- Update producers to send events: Zuora auto cancel, Supporter Plus cancel, MDAPI, Salesforce (needs an SF dev).


cdk/lib/subscription-events-bus.ts
```diff
@@ -0,0 +1,63 @@
...
+			logGroupName: `/aws/events/${busName}`,
+			retention: RetentionDays.TWO_WEEKS,
+		});
+
+		const rule = new Rule(this, 'LogAllEventsRule', {
+			description: `Log all events on the ${busName} bus to CloudWatch Logs for debugging`,
+			eventBus: bus,
+			eventPattern: { account: [this.account] },
+			targets: [
+				// can't use CloudWatchLogGroup as it needs custom/cdk lambdas
```
johnduffell: I got `Missing parameters for AssetParameters...` errors (see https://riffraff.gutools.co.uk/deployment/view/2a0b6439-3e8c-4434-b66a-894896eb6897?verbose=1 ) as CDK expects there to be some special properties which riffraff doesn't set.  I know some places the CDK lambdas are inlined into the json, but in this case it wanted them provided.

In this case the workaround was pretty simple - target the ARN and add a resource policy below, but if it were more complicated we could look into getting the lambda working by inlining or other techniques.

cdk/lib/subscription-events-bus.ts
```diff
@@ -0,0 +1,63 @@
...
+				resources: [logGroup.logGroupArn],
+				principals: [new ServicePrincipal('events.amazonaws.com')],
+				conditions: {
+					ArnEquals: { 'aws:SourceArn': rule.ruleArn },
+				},
+			}),
+		);
+
+		new CfnOutput(this, 'BusArn', {
+			exportName: SubscriptionEventsBus.exportName(stage),
```
johnduffell: it will be exported as `subscription-events-bus-${stage}-arn`, this has to be unique per AWS account and can be imported by other stacks.
This isn't essential - apps and CDK permissions can also look the bus up by name, which is likely to be more convenient in most cases, but it may be useful to reference the arn directly sometimes.

shtukas (approved):

PR #3827, opened 2026-09-16
Add the aws-cli to .tool-versions

tjmw:
## What does this change?

Add the AWS CLI tool to `.tool-versions` so that it is automatically available in a dev container.

I need to use it to set AWS credentials in order to run integration tests.

Re-built my dev container and `aws` is available.


rupertbates (approved):
👍

.tool-versions
```diff
@@ -1,3 +1,4 @@
 github:guardian/devenv 20260721-123412
 java corretto-21.0.7.6.1
 node 22.18.0
+aqua:aws/aws-cli 2
```
johnduffell, line 4: it seems like the recommendation is to use `aws-cli 2` for clarity although it maps under the covers to aqua at present.
https://mise.jdx.dev/registry.html#registry

tjmw: Oh that's interesting. Using the shorthand feels slightly more risky as IIUC it's effectively a pointer and you may not notice if that pointer is updated to point to something else? Using the full name feels more deterministic to me.

johnduffell: yeah I couldn't quite get to the bottom if it - actually I was mostly checking whether "aqua" was trustworthy/official, it seems it is.  I just noticed along the way that the pointer exists - it seems to be recommended by AI but I couldn't find any specific recommendation from mise docs.

PR #3828, opened 2026-09-16
Multiple Account API: Handle send invitation email failures

tjmw:
## What does this change?

When calling the create invitation endpoint we trigger an email to the secondary user. If that email fails then the response will be a 500 error, so the client will consider it failed. However, the invite itself was created. So instead if this ever happens (which is unlikely, because we're just adding an item to an SQS queue) then catch the error and emit a Cloudwatch metric. I've defined a new Cloudwatch alarm for this metric.

If we're happy with this approach I'll extend it to other multiple account API emails.

Added a unit test to cover the failure scenario.

I also tested the happy path manually in CODE to verify that the email send itself was still working as expected.

I tested the failure scenario by temporarily breaking email sending in CODE (I removed the relevant SQS permission from multiple-account-api) - the API request returned a successful response code, I saw the expected message in the logs and the alarm triggered.


cdk/lib/multiple-account-api.ts
```diff
@@ -51,6 +54,21 @@ export class MultipleAccountApi extends SrStack {
 		lambda.addPolicies(
 			AllowSqsSendPolicy.create(this, 'supporter-product-data', 'braze-emails'),
 		);
+		const putMetricPolicy: Policy = new Policy(this, 'Put Metric Policy', {
```
rupertbates: Could maybe make this reusable by putting it in cdk/lib/cdk/policies.ts? Don't feel strongly about it though.

rupertbates (approved):
Looks good

cdk/lib/multiple-account-api.ts
tjmw: Oh yeah good point. Do you think `metricNamespace` should be parameterized? Or would we be happy to always assume the value it takes here (`support-service-lambdas`). I.e. should the caller be able to control the `metricNamespace` or is that unnecessary?

tjmw: Done in d95c1d6.

PR #3834, opened 2026-09-18
Add mParticle acquisition publisher

juabara:
Adds the EventBridge → SQS → Lambda pipeline, CDK infrastructure, tests, packaging, and CI registration for the mParticle paid-media daily feed.


handlers/mparticle-acquisitions-publisher/src/mparticleClient.ts
```diff
@@ -0,0 +1,60 @@
+import { logger } from '@modules/logger/logger';
+import type { MParticleBatch } from './acquisitions';
+import type { AppConfig } from './config';
+
+export class MParticleClient {
```
tomrf1, line 5: could we reuse https://github.com/guardian/support-service-lambdas/blob/main/handlers/mparticle-api/src/services/mparticleClient.ts#L18 ?

juabara: thanks, I will extract the shared client into a common module in a follow-up PR

tomrf1 (approved):

PR #3843, opened 2026-09-21
Multiple Account API: Handle send email failures

tjmw:
## What does this change?

Following on from #3828, this PR adds the same approach that was applied to the invitation email to all other emails. I.e. in the multiple account API if an email trigger fails to be added to the SQS queue then put a metric which we'll alarm on, but return a successful HTTP response code to the caller.

I've tested that the email sending still works by going through the scenarios which result in an email to either the primary or secondary in CODE.

I also used the same approach as the previous PR where I temporarily revoked SQS put permissions and verified that the log messages looked good (I tested the invitation redeemed emails) and the metric was correctly emitted.


handlers/multiple-account-api/src/deleteSecondaryUserEndpoint.ts
```diff
@@ -109,6 +109,10 @@ export const deleteSecondaryUserEndpoint = async (
 			}),
 		);
 
+		// What should happen if this or the below lines fail? Right now this
+		// would result in an HTTP 500 being returned, but these only relate to
+		// emails so aren't strictly essential. Therefore should we tread this
+		// like and email failure and return a 200?
```
tjmw, line 115: I'm interested in opinions on the right way to handle this!

rupertbates: I think we probably should return a 200 yes

rupertbates (approved):
👍

handlers/multiple-account-api/src/deleteSecondaryUserEndpoint.ts
tjmw: Thanks - I'll make the change in this PR.

tjmw: Changed in f42ad69

handlers/multiple-account-api/src/deleteSecondaryUserEndpoint.ts
```diff
@@ -110,51 +112,26 @@ export const deleteSecondaryUserEndpoint = async (
...
-		const [account, secondaryUserDetails] = await Promise.all([
-			getZuoraAccount(zuoraClient, subscriptionName),
-			getUserByIdentityId(identityClient, secondaryIdentityId),
-		]);
-
-		if (!secondaryUserDetails?.primaryEmailAddress) {
-			throw new Error('Secondary user does not have email address');
-		}
-
 		const sendSoftOptInCancelEventPromise = sendSoftOptInCancelEvent(
```
rupertbates: This doesn't need to be a const now it's only being used in one place, it can be inlined below

PR #3846, opened 2026-09-23
Add a consent mapping for secondary users

rupertbates:
## What does this change?
This PR adds a mapping for multiple accounts secondary users, to enable us to set consents for them.

See [this doc](https://docs.google.com/document/d/1_BaaN1Z2QS_RGlovG6YguKczZeYQLJakCk9fmsY31Ms/edit?tab=t.0#task=KJpSwouKtyEuLuXZ) for more information


handlers/soft-opt-in-consent-setter/src/main/scala/com/gu/soft_opt_in_consent_setter/models/ConsentsMapping.scala
```diff
@@ -96,6 +97,11 @@ object ConsentsMapping {
       similarGuardianProducts,
       supporterNewsletter,
     ),
+    "Secondary User" -> Set(
+      yourSupportOnboarding,
+      similarGuardianProducts,
```
graham228221: We need to remove 
```suggestion
```

For secondary users, `similar_guardian_products` will be set by Identity directly from the iframe. @theemilyturner confirmed (this isn't working at the moment but they are investigating).

We should not set it here, as this would set it to `true` even if they opted out in the onboarding.

For cancellations, we already ignore `similar_guardian_products` and never switch it off from the cancellation event ([see this change for details](https://github.com/guardian/support-service-lambdas/pull/2130))

graham228221 (approved):
👍

PR #3847, opened 2026-09-23
Send soft opt-in consents for secondary users

rupertbates:
## What does this change?
We need to set the correct soft opt-in settings for secondary users as described in [this document](https://docs.google.com/document/d/1_BaaN1Z2QS_RGlovG6YguKczZeYQLJakCk9fmsY31Ms/edit?tab=t.0#task=OmlFaGuh7JVwAydZ)

Essentially this means sending an acquisition event when an invitation is successfully accepted and a cancellation event when a secondary user is deleted.


handlers/multiple-account-api/src/acceptInvitationEndpoint.ts
```diff
@@ -119,6 +120,12 @@ export const acceptInvitationEndpoint = async (
 			secondaryUserIdentityId: invitation.secondaryIdentityId,
 		});
 
+		await sendSoftOptInAcquisitionEvent(
```
tjmw: Should this block returning the response to the client? I guess it's not very expensive so should be fine.

handlers/multiple-account-api/src/softOptinConsents.ts
```diff
@@ -0,0 +1,57 @@
...
+	const queueName = `soft-opt-in-consent-setter-queue-${stage}`;
+	const messageBody = JSON.stringify(
+		{
+			productName: 'SECONDARY_USER',
+			eventType,
+			identityId,
+			subscriptionId,
+		},
+		null,
+		2,
```
tjmw: Is the spacing to make the log message below prettier? I always forget what the signature is for `JSON.stringify` so maybe a comment or assign the other args to consts with meaningful names?

handlers/multiple-account-api/test/deleteSecondaryUserEndpoint.test.ts
```diff
@@ -41,6 +41,10 @@ jest.mock('../src/emails/leaveSubcriptionEmail', () => ({
 	sendLeaveSubscriptionEmailToPrimary: jest.fn(),
 }));
 
+jest.mock('../src/softOptinConsents', () => ({
```
tjmw: Is this mock required in any other tests to prevent real SQS events from being sent (not that it's a big deal if we do accidentally)?

tjmw (approved):
Looks great. I left a few minor comments/questions to take or leave.

handlers/multiple-account-api/src/acceptInvitationEndpoint.ts
rupertbates: We have to await it to make sure it is sent but I've added it in to a Promise.all with the email to reduce the blocking time

handlers/multiple-account-api/src/softOptinConsents.ts
rupertbates: I've used the prettyPrint function to make it clearer hopefully

